PR #73, opened 2012-02-04
NH-2821 - another approach to finding log4net 

whut:
https://nhibernate.jira.com/browse/NH-2821

Here is an updated pull request, like in #15. This time I just check if log4net is already loaded in current AppDomain, so now it does not matter if someone uses log4net or nhibernate from GAC or not!


hazzik:
Merged 8692a5932d65ba0d8fbb60e10401c3fadf4a51fb

PR #171, opened 2012-11-14
NH-1262: Port of one-to-one cascade delete orphan functionality from Hibernate 

mickfold:
https://nhibernate.jira.com/browse/NH-1262

This is a direct port of the cascade delete orphan functionality for one-to-one mappings from Hibernate to NHibernate.

This commit supports the following cases:

1) Bidirectional primary key based one-to-one.
2) Unidirectional shared primary key based one-to-one
3) Bidirectional foreign key based one-to-one with orphan delete declared on the non-constrained (&lt;one-to-one/>) side.
4) Bidirectional foreign key based one-to-one with orphan delete declared on the constrained (&lt;many-to-one unique="true"/>) side.
5) Unidirectional foreign key based one-to-one with orphan delete declared on the constrained (&lt;many-to-one unique="true"/>) side.
6) Unidirectional foreign key based one-to-one with orphan delete declared on the constrained side (&lt;many-to-one unique="true"/>) where the foreign key is composite (multiple columns).

Further information can be found in the original work for Hibernate https://hibernate.atlassian.net/browse/HHH-4726


hazzik:
Merged, thanks

PR #203, opened 2013-04-16
Multi-Level Distinct Entity Transformer 

manarhusrieh:
This transformer is similar to DistinctRootEntityTransformer but removes
duplicates from all the details of the entity.

### DIFF
--- a/src/NHibernate.Test/NHibernate.Test.csproj
+++ b/src/NHibernate.Test/NHibernate.Test.csproj
@@ -1137,7 +1137,12 @@
     <Compile Include="TestDialects\MsSql2008TestDialect.cs" />
     <Compile Include="TestDialects\PostgreSQL82TestDialect.cs" />
     <Compile Include="TestDialects\SQLiteTestDialect.cs" />
+    <Compile Include="TransformTests\MultiLevelDistinctEntityTransformer\Child.cs" />
+    <Compile Include="TransformTests\MultiLevelDistinctEntityTransformer\Employee.cs" />
     <Compile Include="TransformTests\ImplementationOfEqualityTests.cs" />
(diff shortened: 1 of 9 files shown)


hazzik:
@manarhusrieh what is the difference to #228?

fredericDelaporte:
I think I have understood the differences from its description (not read the code). This one not only perform a distinct on listed entities, but also on any children entity which may be listed twice in an collection belonging to the same root entity.

This could serve as a workaround for hiding the effects of a Cartesian product occurring when eager fetching more than one children collections. But well, this will not solve the performance trouble such a Cartesian product causes.

hazzik:

fredericDelaporte:
Thinking again about this result transformer, I do not think we should accept it.
It is a result transformer which mutates entity states, by altering the contents of entity collections. It may cause undesired updates. I do not think a result transformer which mutates entity collections should be accepted.

If the aim is to use it as a workaround for the Cartesian product troubles occurring when fetching multiple collections, another solution should be chosen instead, like using multiple queries or futures to separate the fetches while batching them in a single roundtrip to the database, as illustrated in the last example of multi-queries [here](http://nhibernate.info/doc/nhibernate-reference/performance.html#performance-multi-query).

fredericDelaporte:
The number of elements in the collection does change, right? That is a mutation of the collection, and this will cause the collection to be persisted on flush. By the way another workaround would be to map them as `set`. But performance wise it is not a very good workaround since the Cartesian product will still occur: separating the fetches of different collections would still be better.

manarhusrieh:
I didn't about the working of the collections in sessions. I usually use this transformer with read-only result sets, that is why I didn't run into this problem before.
Thanks

PR #204, opened 2013-04-17
Fix NH-3252

darrenkopp:
This is similar to the fix proposed on jira for NH-3252. Basically size is handled the same way as string. Without this, can't do varchar(max), only nvarchar(max).


hazzik:
Merged @ bea9bc9c1b98ae1c8b48d9dd0e16df079e7abff1

Thanks, @darrenkopp

PR #210, opened 2013-06-15
NH-2790 - Added support for EagerFetching using Inner Joins using Linq

eallegretta:
https://nhibernate.jira.com/browse/NH-2790

4 new extesion methods had been added to support this functionality on Linq

InnerFetch
InnerFetchMany
ThenInnerFetch
ThenInnerFetchMany

### DIFF
--- a/src/NHibernate.ncrunchsolution
+++ b/src/NHibernate.ncrunchsolution
@@ -1,0 +1,13 @@
+<SolutionConfiguration>
+  <FileVersion>0</FileVersion>
+  <AutoEnableOnStartup>Default</AutoEnableOnStartup>
+  <AllowParallelTestExecution>false</AllowParallelTestExecution>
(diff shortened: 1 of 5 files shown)


razvan-panda:
@hazzik Any update on this?

renatoguarilhadeploy:
@hazzik Greetings, Have this issue been implemented?

felipeoriani:
Would be great to have this feature implemented on the next version of NHibernate. We have queries here where Inner join perform better than left join on the database. I will use the extension.

My suggestion is to define an overload for Fetch method to define the JoinType using the existent enum. The default could be the left join as it already works. For sample:

`
var query = Session.Query<Product>().Fetch(x => x.Category, JoinType.Inner).Fetch(x => x.Supplier).ToList();
`

Perform a inner join on the Category and a left join on the Supplier.

Thank you.

PR #211, opened 2013-06-19
Fixes for NH-3480 and NH-3453

JKeyser:
http://nhibernate.jira.com/browse/NH-3453

Hello, all.

This is my first attempt at a proposed contribution to NHibernate, so please be patient with me.  I have created proposed fixes for JIRA issues NH-3480 and NH-3453 related to one-to-many and many-to-one mappings using property references.

Basically, the problem with both of these issues is that in some parts of NHibernate, the values used to join entities together is assumed to be the keys of the entities, and "forgets" that those values could be property reference values instead.  For both of these issues, the property values were attempted to be used as the entity key for the purposes of determining if an entity was dirty and for finding the owner of the collection from the first-level cache.  In my proposed solution, I attempted to make a distinction between these two values so that one would not be used in place of the other.

The one thing I has to do to make this work was make significant changes to the one-to-many join walker and persister.  In order to get the actual ID of the collection's owning entity into the dataset read from the database so owner could then be gotten from the first-level cache, I decided to join in the owning entity's table and return its ID when retrieving the collection when a property reference was used.  This seemed to be the best solution to this problem.

Although I have been using NHibernate for a couple of years and have dug into the code many times during that time, but this is the deepest I have dug into it, much less attempted to made changes, so I am open to suggestions or pointers to things I have missed.

Thank you for you attention.

### DIFF
--- a/src/NHibernate/Loader/Collection/OneToManyJoinWalker.cs
+++ b/src/NHibernate/Loader/Collection/OneToManyJoinWalker.cs
@@ -58,7 +58,7 @@ namespace NHibernate.Loader.Collection
 			int collectionJoins = CountCollectionPersisters(associations) + 1;
 			CollectionSuffixes = BasicLoader.GenerateSuffixes(joins + 1, collectionJoins);
 
-			SqlStringBuilder whereString = WhereString(oneToManyPersister.CollectionType.UseLHSPrimaryKey ? alias : alias + "owner_", oneToManyPersister.KeyColumnNames, subquery, batchSize);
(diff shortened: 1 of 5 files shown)


fubar-coder:
The commit 6b33184a2301469d562ebd170a31de7514d95f73 seems dangerous, because some databases like Firebird don't support ANSI JOIN for UPDATE. Please keep at least both paths for databases with different capabilities.

hazzik:
Thanks, @fubar-coder. You can rise a JIRA issue and make a pull request with tests.

hazzik:
In fact the tests seems to work on firebird.

PR #217, opened 2013-07-25
Makes the session available to ISQLExceptionConverter.

alex-davidson:
Enables more advanced exception translations based on the persistence
context.

When using batching, details about the entity name and ID which caused a failure are unavailable to the exception converter. Given the name of a violated constraint, however, it is often possible to determine the offending entity by type if we have access to the persistence context. This commit makes the relevant information available to exception converters.

### DIFF
--- a/src/NHibernate.Test/ExceptionsTest/NullQueryTest.cs
+++ b/src/NHibernate.Test/ExceptionsTest/NullQueryTest.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Collections;
 using System.Data;
+using NHibernate.Engine;
 using NHibernate.Exceptions;
+using NHibernate.Impl;
 using NUnit.Framework;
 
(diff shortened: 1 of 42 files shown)


fredericDelaporte:
I do not see any change in this PR for improving exception messages. The session is available to the sql exception converters, but none of them does anything of it. Correct me if I am wrong.

Then it is a lot of changes for no added feature, or only for enabling custom converter implementation to have the session.

And it looks like you were intending to infer the entities from the violated constraint name and some inspection of dirty entities in session I guess: it can only work if constraints are named according to entity names (or to table names at least).

alex-davidson:
Indeed. This PR is rather obsolete now and should probably be discarded. I've since learned that it was really the wrong approach anyway: these days we just try, rollback, infer failure reason from database state, maybe retry.

fredericDelaporte:
Ok, thanks for your answer, closing.

PR #227, opened 2013-10-04
NH-2520 - Add TimestampUtc type

matomira:
This adds the TimestampUtc type

This is the only the patch from the original author, without modifications.

This was applied to the 3.3.x branch of 4th October 2013

### DIFF
--- a/src/NHibernate/NHibernateUtil.cs
+++ b/src/NHibernate/NHibernateUtil.cs
@@ -242,6 +242,11 @@ namespace NHibernate
 		
 		public static readonly NullableType DbTimestamp = new DbTimestampType();
 
+        /// <summary>
+        /// NHibernate timestamp utc type.
+        /// </summary>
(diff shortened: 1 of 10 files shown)


src/NHibernate/NHibernateUtil.cs
```diff
@@ -242,6 +242,11 @@ public static IType GuessType(System.Type type)
 		
 		public static readonly NullableType DbTimestamp = new DbTimestampType();
 
+        /// <summary>
+        /// NHibernate timestamp utc type.
+        /// </summary>
+        public static readonly NullableType TimestampUtc = new TimestampUtcType();
```
hazzik: Can you please change this to `public static readonly TimestampUtcType TimestampUtc = new TimestampUtcType()`;?

hazzik:
@matomira can you please rebase on top of master, fix indents (tabs instead of spaces), and squash commits?

hazzik:
@matomira ?

PR #228, opened 2013-10-11
[NH-3547] New feature: DistinctEntitiesResultTransformer

matomira:
This integrates the code I posted to JIRA into the project.

https://nhibernate.jira.com/browse/NH-3547

### DIFF
--- a/src/NHibernate/Transform/DistinctEntitiesResultTransformer.cs
+++ b/src/NHibernate/Transform/DistinctEntitiesResultTransformer.cs
@@ -1,0 +1,128 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+using Iesi.Collections.Generic;
+
+namespace NHibernate.Transform
+{
+	[Serializable]
+	public class DistinctEntitiesResultTransformer : IResultTransformer
+	{
+		private static readonly IInternalLogger log = LoggerProvider.LoggerFor(typeof(DistinctEntitiesResultTransformer));
+		private static readonly object Hasher = new object();
+		private readonly bool[] returns;
+
+		internal sealed class Identity
+		{
+			internal readonly object entity;
+
+			internal Identity(object entity)
+			{
+				this.entity = entity;
+			}
+
+			public override bool Equals(object other)
+			{
+				Identity that = (Identity) other;
+				return ReferenceEquals(entity, that.entity);
+			}
+
+			public override int GetHashCode()
+			{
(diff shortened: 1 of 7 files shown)


hazzik:

fredericDelaporte:
This transformer could be renamed `DistinctPivotResultTransformer`. It does not care whether the result row contains entities or scalar values. It transforms each column in a new list (so pivoting each column into rows), removing duplicates from it by the way, then yielding the resulting list of lists.

src/NHibernate/Transform/DistinctEntitiesResultTransformer.cs
```diff
@@ -0,0 +1,128 @@
...
+
+			internal Identity(object entity)
+			{
+				this.entity = entity;
+			}
+
+			public override bool Equals(object other)
+			{
+				Identity that = (Identity) other;
+				return ReferenceEquals(entity, that.entity);
```
fredericDelaporte: My proposed renaming would not suit this implementation, as it most likely to not handle properly equality of value types. Maybe the result transformer constructor should accept an `IEqualityComparer`.

src/NHibernate/Transform/DistinctEntitiesResultTransformer.cs
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		public override bool Equals(object obj)
+		{
+			if (obj == null)
+			{
+				return false;
+			}
+			
+			return obj.GetHashCode() == Hasher.GetHashCode();
```
fredericDelaporte: I guess this code is from `DistinctRootEntityResultTransformer`. Considering that objects having the same hashcode are equal is a bug, since hashcode collisions may occur. This has been fixed in `DistinctRootEntityResultTransformer`, and should be fixed here too. Here is how the code is now in `DistinctRootEntityResultTransformer`:

```c#
if (obj == null || obj.GetHashCode() != Hasher.GetHashCode())
{
	return false;
}
// NH-3957: do not rely on hashcode alone.
// Must be the exact same type
return obj.GetType() == typeof(DistinctRootEntityResultTransformer);
```

But that is indeed not enough for this new result transformer, since this result transformer has some state: the `returns` field must be compared for equality. And the hashcode logic should be amended to generate different hashcode values for different `returns` field.

fredericDelaporte:
What does actually this new transformer is explained [here](https://github.com/nhibernate/nhibernate-core/pull/228#issuecomment-403486536).

It looks to me very specific. I do not think it should be added into NHibernate. Closing for now.

PR #238, opened 2013-11-19
 NH-3573  - Added query cache statistics in MultiCriteria, and cache put only if enabled

whut:
Added query cache statistics when using MultiCriteria, and putting to cache only when put is enabled. 

Statistics handling is based on code from Loader.cs.

This fix can be easly backported to 3.x, could that happen?


hazzik:
Merged in 569b49c4ead3f026f0a161c1b9b6833fb92e7216

PR #241, opened 2013-12-06
NH-3431 - Replace usage of System.Data classes to System.Data.Common

hazzik:
https://nhibernate.jira.com/browse/NH-3431

Need to consider include this changes to 4.0 release or not


ngbrown:
@hazzik I've rebased your [NH-3431](https://nhibernate.jira.com/browse/NH-3431) work to the most recent HEAD here https://github.com/ngbrown/nhibernate-core/tree/NH-3431 

I'm close to having [NH-3807 (Support for .NET Core 1.0)](https://nhibernate.jira.com/browse/NH-3807) ready to show, so if we can get dependencies like this in, that would be great.

PR #243, opened 2013-12-10
NH-3578 - Subcriteria.UniqueResult<T> for value types should return default(T), same as CriteriaImpl.UniqueResult<T> when result is null

whut:
Subcriteria.UniqueResult<T> for value types should return default(T), same as CriteriaImpl.UniqueResult<T> when result is null


hazzik:
Hi @whut, can you please enable [Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) feature?

PR #259, opened 2014-03-06
RegisterCustomProjection with MethodInfo

stackcoder:
Allow calling RegisterCustomProjection and RegisterCustomMethodCall directly with MethodInfo like RegisterGenerator in DefaultLinqToHqlGeneratorsRegistry. This is useful to register your custom extensions more dynamically without accessing _customMethodCallProcessors by reflection.


stackcoder:
Currently I'm using a workaround which dynamicly creates a new lamda expression from the MethodInfo to pass it to RegisterCustomProjection<T>. So I close this pull request.

PR #260, opened 2014-03-14
NH-3470 - Allow Session.Query to load entities as read-only

rjperes:
https://nhibernate.jira.com/browse/NH-3470

### DIFF
--- a/src/NHibernate.Test/Linq/QueryReadOnlyTests.cs
+++ b/src/NHibernate.Test/Linq/QueryReadOnlyTests.cs
@@ -1,0 +1,68 @@
+﻿using System.Linq;
+using NHibernate.AdoNet;
+using NHibernate.Cfg;
+using NHibernate.Engine;
+using NHibernate.Linq;
+using NUnit.Framework;
+
+namespace NHibernate.Test.Linq
+{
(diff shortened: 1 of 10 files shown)


Sebazzz:
@hazzik Extension methods are consistent with how the LINQ API is designed though, and are consistent with how NHibernate currently works (`Cacheable` for example). It is also consistent with other ORM.

rjperes:
Also, why is the PR broken?

rjperes:
@hazzik hello?

hazzik:
> @hazzik Extension methods are consistent with how the LINQ API is designed though
I don't like this pollution with extension methods. And it makes harder to analyze expression tree.

>@hazzik hello?

Moving house and don't have internet connection at home.

hazzik:
>Also, your proposal doesn't allow flexible setting the query options later on in the query, for example what Cacheable allows (or the criteria API for example).

This is a fair point, but the way this is implemented is wrong, it should just call a method on INhQueryable to set required option, but not produce additional node types, result operators and unnecessary complications to the expression tree.

hazzik:
>Also, why is the PR broken?

I don't know why

hazzik:
Replaced by #418

PR #282, opened 2014-07-11
Request for adding missing dialect methods Per

jeffreyabecker:
Resolves:
NH-3633 - Add Support for Dropping Columns on Dialect
NH-3632 - Add Support for AlteringColumns on Dialect
NH-3631 - Add support for BatchTerminator on Dialect


hazzik:
Won't fix. Please use a full-feature migration framework

PR #284, opened 2014-08-01
NH-3648 addition of entity listener annotations

wgroeneveld:
See https://nhibernate.jira.com/browse/NH-3648
This is the implementation of [PreInsert], [PreUpdate], ... attributes you can place on entities which get automatically triggered through a NHibernate event listener and some reflection. It's missing from NH and the java version has this feature for years (@PrePersist, @PreUpdate). This enables developers to leverage the power of DDD instead of writing logic in data access layers. 

The most common usage for this would be auto-updating audit fields (user create, timestamping).
Please let me know if it's sufficient or something is missing, I'm willing to make changes :-) I implemented a caching mechanism to speed up the reflection part - as entity type metadata simply cannot change during runtime.


wgroeneveld:
Any updates on this one? Is there anything I can do to help?

gliljas:
I'll review it again. Thanks for your contribution.

src/NHibernate/Event/EntityListener.cs
```diff
@@ -0,0 +1,211 @@
...
+        }
+    }
+
+    [Serializable]
+    public class EntityListener : IPreUpdateEventListener, IPreInsertEventListener, IPreDeleteEventListener,
+        IFlushEntityEventListener, IPostUpdateEventListener, IPostInsertEventListener, IPostDeleteEventListener
+    {
+        [NonSerialized]
+        private Dictionary<System.Type, Dictionary<System.Type, IEnumerable<MethodInfo>>>
+            _methodsWithListenerAttributeCache;
```
fredericDelaporte: Instances of this class are potentially shared among many threads. This dictionary must be changed for a `ConcurrentDictionary`.

src/NHibernate/Event/EntityListener.cs
```diff
@@ -0,0 +1,211 @@
...
+                UpdateDirtyFieldValuesInState(config, initialFieldValues);
+            }
+        }
+
+        private void InvokeMethodsWithAttribute(EventConfig config, System.Type attributeType)
+        {
+            var entityType = config.Entity.GetType();
+            if (!_methodsWithListenerAttributeCache.ContainsKey(entityType))
+            {
+                _methodsWithListenerAttributeCache.Add(entityType,
```
fredericDelaporte: Atomic `ConcurrentDictionary` methods should be used, like `GetOrAdd`.

src/NHibernate/Event/EntityListener.cs
```diff
@@ -0,0 +1,211 @@
...
+                                                               attributeType,
+                                                               GetMethodsWithListenerAttribute(attributeType, entityType)
+                                                           }
+                                                       });
+            }
+
+            Dictionary<System.Type, IEnumerable<MethodInfo>> cacheDict = _methodsWithListenerAttributeCache[entityType];
+            if (!cacheDict.ContainsKey(attributeType))
+            {
+                cacheDict.Add(attributeType, GetMethodsWithListenerAttribute(attributeType, entityType));
```
fredericDelaporte: Same here.

src/NHibernate/Event/EntityListener.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            cacheDict[attributeType].ForEach(method => InvokeAndApplyDirtyFields(config, method));
+        }
+
+        private static IEnumerable<MethodInfo> GetMethodsWithListenerAttribute(System.Type attributeType,
+                                                                               System.Type entityType)
+        {
+            return entityType
+                .GetMethods()
+                .Where(method => method.GetCustomAttributes(attributeType, false).Length > 0);
```
fredericDelaporte: The `Where` part yields a deferred enumerable, which will be re-executed at each usage, seriously reducing the benefits of caching it into `_methodsWithListenerAttributeCache`. A `ToList()` or `ToArray()` should be added.

It will be enumerated by concurrent threads by the way, but enumerations of these types are known to be thread safe, even if not documented as such.

src/NHibernate/Event/EventListeners.cs
```diff
@@ -99,12 +101,12 @@ static EventListeners()
...
-		private IPreInsertEventListener[] preInsertEventListeners = new IPreInsertEventListener[] {};
-		private IPostDeleteEventListener[] postDeleteEventListeners = new IPostDeleteEventListener[] {};
-		private IPostUpdateEventListener[] postUpdateEventListeners = new IPostUpdateEventListener[] {};
-		private IPostInsertEventListener[] postInsertEventListeners = new IPostInsertEventListener[] {};
+        private IPreDeleteEventListener[] preDeleteEventListeners = new IPreDeleteEventListener[] { defaultEventEntityListener };
+        private IPreUpdateEventListener[] preUpdateEventListeners = new IPreUpdateEventListener[] { defaultEventEntityListener };
+        private IPreInsertEventListener[] preInsertEventListeners = new IPreInsertEventListener[] { defaultEventEntityListener };
+        private IPostDeleteEventListener[] postDeleteEventListeners = new IPostDeleteEventListener[] { defaultEventEntityListener };
+        private IPostUpdateEventListener[] postUpdateEventListeners = new IPostUpdateEventListener[] { defaultEventEntityListener };
+        private IPostInsertEventListener[] postInsertEventListeners = new IPostInsertEventListener[] { defaultEventEntityListener };
```
fredericDelaporte: Sorry, the code base uses tabs. Identation should be fixed here, and new .cs files should re-idented with tabs.

hazzik:
I believe the analysis for the entity lifecycles annotations has to be performed at a configuration stage and cached then, not on a first occurrence.

PR #287, opened 2014-08-05
NH-2016 - Fix joining to association twice using criteria

### LINKED ISSUE #990: NH-2016 - Duplicate Association Path when creating multiple aliases
**Anup Marwadi** created an [issue](https://nhibernate.jira.com/browse/NH-2016) — <time datetime="2009-11-12T13:32:15.727Z" title="2009-11-12T13:32:15.727Z">12th November 2009, 13:32:15</time>:
> Hello All,
> This is a bug I have noticed on my current NHibernate project. I tried to search for it on this forum and don't know if it has been rectified as of yet or not. However, the explanation for this bug is provided here:
> http://derek-says.blogspot.com/2008/06/duplicate-association-path-bug-in.html
> 
> I would like to know if there is a way using criteria API to create multiple aliases that point to the same table. I would highly appreciate it if you could share any information on the same.
> My criteria is pretty simple:
> 
> 
> ```
> 
>  LeadManager manager = new LeadManager();
> ISession mySession = manager.Session.GetISession();
> ICriteria criteria = mySession.CreateCriteria(typeof(Lead));
> criteria.SetProjection(Projections.Distinct(Projections.Property("Id")));
> criteria.Add(Expression.Eq("CustomerId", 950));
> criteria.CreateAlias("CandidateMember", "candidate");
> criteria.CreateAlias("candidate.ExtensionValues", "extensions");
> criteria.CreateAlias("extensions.ExtensionValueTypeMember", "extensionTypes");
> Conjunction c = new Conjunction();
> c.Add(Expression.Eq("extensionTypes.Id", 1));
> c.Add(Expression.In("extensions.ExtensionDatum", new object[] { 'A', 'B' }));
> criteria.Add(c);
> criteria.CreateAlias("candidate.ExtensionValues", "extensions1");
> criteria.CreateAlias("extensions1.ExtensionValueTypeMember", "extensionTypes1");
> Conjunction c1 = new Conjunction();
> c1.Add(Expression.Eq("extensionTypes.Id", 2));
> c1.Add(Expression.In("extensions.ExtensionDatum", new object[] { 'A', 'B', 'C', 'D' }));
> criteria.Add(c1);
> criteria.SetMaxResults(10);
> long[] leadIds = criteria.List<Int64>().ToArray();
> 
> ```
> 
> 
---
**Robert Dal Santo** added a comment — <time datetime="2011-09-17T15:44:15.464Z" title="2011-09-17T15:44:15.464Z">17th September 2011, 15:44:15</time>:
> This issue is related to https://nhibernate.jira.com/browse/NHLQ-43 and a duplicate of the Hibernate defect/limitation https://hibernate.onjira.com/browse/HHH-879.
> This limitation causes me to create ICriteria queries using subqueries, which will almost certainly result in sub-optimal execution plans. I'd rather not, but this might force me to abandon ICriteria and use native-SQL instead, which will result in much more verbose logic.
>

gmalyshev:
Joining to association twice using criteria.

I'm added **criteria alias** for unique identification join statement. This makes it possible to declare complex expressions join.

Deviations in the standard behavior of the rest of the functionality I hadn't noticed

Jira issue [NH-3472](https://nhibernate.jira.com/browse/NH-3472) (dup of [NH-2016](https://nhibernate.jira.com/browse/NH-2016), #990).

Fixes #990


hazzik:

gmalyshev:
Done

fredericDelaporte:
Rebased, async code generated, and cleaned up. Especially I have made some renaming to improve code readability.

The change does the job, but maybe it could be done in a simpler manner. @bahusoid, you are more used than me to the criteria code, may you have a look?

bahusoid:
Will be able to check code during next week.

But it seems those changes might lead to unexpected behavior if query is mixed with `Fetch` and `JoinAlias/JoinQueryOver` calls for the same properties. I mean something like this:

```C#
Child child = null;
session.QueryOver<Obj>
.Fetch(o => o.Child).Eager
.Fetch(o => o.Child.SubChild).Eager

.JoinAlias(o => o.Child, () => child)
.Where(x => x.Child.IsActive == true);
```

It is my understanding the query above after this change will start to generate two joins to Child table. Am I right?

Also after this change it seems query can contain multiple joins to fetch the same collection persister (with added multiple left join aliases for  the same collection). I wonder how it will be handled.

P.S. Actually the desired behavior is already possible via "Entity Join" - `JoinEntityAlias/JoinEntityQueryOver` (but yeah you need to specify explicitly full join condition).

fredericDelaporte:
> It is my understanding the query above after this change will start to generate two joins to Child table. Am I right?

No I do not think so. And testing it, it does not, see the new commit.

> Also after this change it seems query can contain multiple joins to fetch the same collection persister (with added multiple left join aliases for the same collection). I wonder how it will be handled.

I think this will somewhat "union" the fetches, with de-duplication in case of a `set`, an error in case of a `bag` (Cannot simultaneously fetch multiple bags). But that is a somewhat weird use case, I am not sure we should specify the expected output in such case. Normally it could always be rewritten for using a single alias with a more complex restriction. Should really a test for it be added? So I have set the test as `Explicit` only.

bahusoid:
Additional unexpected join was my main concern regarding this PR - if it's not a problem then it looks good to me. It can be useful.

Regarding "Entity Join" changes -  they seem unrelated to this PR. Am I wrong?

And there is another catch - only one fetch mode (`SelectMode`) can be applied for such joins. Otherwise it seems alias support needs to be added for `CriteriaImpl.GetSelectMode`. Not sure if it's worth it.

bahusoid:
> EntityJoinInfo had to change to avoid casting ICriteria to Subcriteria 

I see no such places related to this PR. It seems this change was done only to remove the  following line:
https://github.com/nhibernate/nhibernate-core/blob/8649f9ac52cfc0d022a9553f41a860873a0d1450/src/NHibernate/Loader/Criteria/CriteriaJoinWalker.cs#L83

If you want to make such refactoring maybe it's better to do as another pr to better see actual changes for this PR.

fredericDelaporte:
https://github.com/nhibernate/nhibernate-core/blob/8649f9ac52cfc0d022a9553f41a860873a0d1450/src/NHibernate/Loader/Criteria/CriteriaJoinWalker.cs#L83

This line has not been removed. It has been changed to the correct call, required for robustly supporting the change since the `Alias` may be something else than the `Path`.

bahusoid:
So just to be clear I think that `EntityJoinInfo` -> `EntityJoinInformation` refactoring is unrelated to this PR.  See suggested clean up (feel free to drop it if you are not convinced) - but  I think this change gives the wrong idea of the scope of this PR.

fredericDelaporte:
I find this a bit overstated, but well, so be it.

PR #292, opened 2014-08-13
NH-3402 - Refactor ThreadLocalSessionContext

rjperes:
NH-3402

### DIFF
--- a/src/NHibernate/Context/ThreadLocalSessionContext.cs
+++ b/src/NHibernate/Context/ThreadLocalSessionContext.cs
@@ -19,7 +19,7 @@ namespace NHibernate.Context
 	/// has been called. If <tt>Close()</tt> is called on a session managed by
 	/// this class, it will be automatically unbound.
 	/// <p/>
-	/// Additionally, the static <see cref="Bind"/> and <see cref="Unbind"/> methods are
+	/// Additionally, the static <see cref="CurrentSessionContext.Bind"/> and <see cref="CurrentSessionContext.Unbind"/> methods are
 	/// provided to allow application code to explicitly control opening and
 	/// closing of these sessions.  This, with some from of interception,
 	/// is the preferred approach.  It also allows easy framework integration
--- a/src/NHibernate/Context/ThreadLocalSessionContext.cs
+++ b/src/NHibernate/Context/ThreadLocalSessionContext.cs
@@ -1,12 +1,11 @@
 using System;
 using System.Collections.Generic;
-
+using System.Threading;
 using NHibernate;
 using NHibernate.Engine;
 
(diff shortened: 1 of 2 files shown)


src/NHibernate/Context/ThreadLocalSessionContext.cs
```diff
@@ -20,164 +19,26 @@ namespace NHibernate.Context
...
 	/// this class, it will be automatically unbound.
 	/// <p/>
-	/// Additionally, the static <see cref="Bind"/> and <see cref="Unbind"/> methods are
+	/// Additionally, the static <see cref="CurrentSessionContext.Bind"/> and <see cref="CurrentSessionContext.Unbind"/> methods are
 	/// provided to allow application code to explicitly control opening and
 	/// closing of these sessions.  This, with some from of interception,
 	/// is the preferred approach.  It also allows easy framework integration
 	/// and one possible approach for implementing long-sessions.
 	/// <p/>
 	/// </summary>
```
fredericDelaporte: With this change, most of this xml comment is obsolete. The previous implementation was generating a session if there was no current session, which this new implementation does not.
The comment should be updated accordingly.

src/NHibernate/Context/ThreadLocalSessionContext.cs
```diff
@@ -1,12 +1,11 @@
 using System;
 using System.Collections.Generic;
-
+using System.Threading;
 using NHibernate;
 using NHibernate.Engine;
 
 namespace NHibernate.Context
 {
-	//TODO: refactoring on this class. Maybe using MapBasedSessionContext.
```
fredericDelaporte: Good move to have ignored and removed this comment, since the `MapBasedSessionContext` implementation is flawed (thread safety issues, going to Jira that), and unneeded with your implementation.

fredericDelaporte:
Another changes may be missing:

- `SessionFactoryImpl.BuildCurrentSessionContext` handles short configuration names for built-in session contextes. It would probably be nice to give a short name to the re-implemented `ThreadLocalSessionContext`.
- doc/reference/modules/architecture.xml documents build-in session contextes. Now that the `ThreadLocalSessionContext` would behave as the others, it should be documented here too I think.

fredericDelaporte:
Completely forgotten all that when doing #644. If #644 is ok, I think this partially done port from Hibernate should be either completely removed, or finished with its original intended semantic. If we agree on that alternative, then this would mean this #292 PR is to be closed.

fredericDelaporte:
With #644, this PR in its current state would just change `ThreadLocalSessionContext` to a duplicate of `ThreadStaticSessionContext` with more code. Closing.

PR #293, opened 2014-08-13
NH-3607 - Resilient Connection Provider

rjperes:
https://nhibernate.jira.com/browse/NH-3607


lnu:
Could it be part of the next release?

src/NHibernate.Test.VisualBasic/NHibernate.Test.VisualBasic.vbproj
```diff
@@ -145,6 +145,9 @@
     <EmbeddedResource Include="Issues\NH2966\Issues.NH2966.Mappings.hbm.xml" />
     <EmbeddedResource Include="Issues\NH3302\Issues.NH3302.Mappings.hbm.xml" />
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
```
fredericDelaporte: This is a VS change that should very likely be undone, as it looks already done in master. (This change is a sort of "marking as test project" [thing](http://stackoverflow.com/q/18614342/1178314) by Microsoft.)

src/NHibernate.TestDatabaseSetup/NHibernate.TestDatabaseSetup.csproj
```diff
@@ -115,6 +115,9 @@
       <Install>true</Install>
     </BootstrapperPackage>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
```
fredericDelaporte: Same as previous comment.

src/NHibernate/Connection/ResilientDriverConnectionProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+		public Int32 MaxTries { get; set; }
+
+		public TimeSpan DelayBetweenTries { get; set; }
+
+		public override void Configure(IDictionary<String, String> settings)
+		{
+			String maxTries;
+			String delayBetweenTries;
+
+			if (settings.TryGetValue(ConnectionMaxTries, out maxTries) == true)
```
fredericDelaporte: Is it a NHibernate accepted practice to "explicit" boolean conditions like this and the three next conditions?  
(I personally quite dislike it. Even on nullable boolean I would rather do `if (thingYieldingNullableBoolean ?? false)` instead of comparing to true. By the way, I dislike non mandatory usages of `this.` too. Even more than now VS signal them as useless.)

src/NHibernate/Connection/ResilientDriverConnectionProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+				{
+					this.DelayBetweenTries = timespan;
+				}
+				else if (Int32.TryParse(delayBetweenTries, out delay) == true)
+				{
+					this.DelayBetweenTries = TimeSpan.FromSeconds(delay);
+				}
+				else
+				{
+					throw (new Exception("Invalid delay"));
```
fredericDelaporte: We should throw a `HibernateException` or a more specialize one instead of the general `Exception`.

src/NHibernate/Connection/ResilientDriverConnectionProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+				{
+					if (i == this.MaxTries - 1)
+					{
+						log.Error(String.Concat("Could not get connection after {0} tries", this.MaxTries), ex);
+						throw;
+					}
+					else
+					{
+						Thread.Sleep(this.DelayBetweenTries);
+					}
```
fredericDelaporte: Should we really silently eat away the failure on first tries? I would rather have a warning here.

src/NHibernate/Connection/ResilientDriverConnectionProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+			}
+
+			base.Configure(settings);
+		}
+
+		public override IDbConnection GetConnection()
+		{
+			IDbConnection con = null;
+
+			for (var i = 0; i < this.MaxTries; ++i)
```
fredericDelaporte: If `MaxTries <= 0`, this will yield `null` instead of a connection or an exception. Either we should test against `Math.Max(1, MaxTries)`, or setting `MaxTries` to zero or less should yield an exception.

src/NHibernate/Connection/ResilientDriverConnectionProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+						throw;
+					}
+					else
+					{
+						Thread.Sleep(this.DelayBetweenTries);
+					}
+				}
+			}
+
+			return (con);
```
fredericDelaporte: Forgotten parenthesis?

hazzik: No, it's his coding style.

src/NHibernate/Connection/ResilientDriverConnectionProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+				catch (Exception ex)
+				{
+					if (i == this.MaxTries - 1)
+					{
+						log.Error(String.Concat("Could not get connection after {0} tries", this.MaxTries), ex);
+						throw;
+					}
+					else
+					{
+						Thread.Sleep(this.DelayBetweenTries);
```
stefan-steinegger: In case of an error consider clearing the connection pools. Our server became unusable and required retarting after the SQL Server had been swapped by some MS load balancing or availability technology. The reason was that all connections in the pool got broken. Trying to get a new connection didn't help, because it came from the pool again. It probably won't hurt to clear the pools here. When no connection can be established, it's very unlikely that the pool has anything useful. Unfortunately, I've never got the chance to test whether clearing the pool really solves the problem.

hazzik:
I don't think this is very "resilient" connection provider. What it does is only retries to open the connection. But what would happen if the connection breaks in the middle of an operation?

rjperes:
@hazzik : you are right, this is a duplicate of the functionality that exists for EF pre-Core, with all its limitations. Not much we can do in those scenarios, other than wrap all ADO.NET calls in a try...catch loop.

hazzik:
As per my last comment.

PR #294, opened 2014-08-13
NH-2920 - NHibernateContext for WCF Data Service

rjperes:
http://nhibernate.jira.com/browse/NH-2920

Includes Microsoft.Data.Services.dll from WCF Data Services 5.6 (http://www.microsoft.com/en-us/download/details.aspx?id=39373)


hazzik:
Closed as per JIRA.

PR #304, opened 2014-08-17
NH-2285 - Support for LockMode in linq provider

OnurGumus:
reimplemented the ffix

### DIFF
--- a/src/NHibernate.Test/Linq/QueryLock.cs
+++ b/src/NHibernate.Test/Linq/QueryLock.cs
@@ -1,0 +1,47 @@
+﻿using System.Linq;
+using NHibernate.AdoNet;
+using NHibernate.Cfg;
+using NHibernate.Engine;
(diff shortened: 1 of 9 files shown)


oskarb:
What bothers me a bit with this is that while the test cases hypothetically will execute the code and implicitly verify that there are no exceptions, they don't actually verify that the requested locking has been applied.

oskarb:
It's unfortunately also vulnerable to caching issues:

```csharp
var result1 = (from e in db.Customers
                     select e).SetLockMode(LockMode.Upgrade).ToList();
var result2 = (from e in db.Customers
                     select e).SetLockMode(LockMode.UpgradeNoWait).ToList();
```

If the above two statements are executed, the second query will reuse LockMode.Upgrade from the first query.

oskarb:
I've rebased and submitted a version with improved test cases as PR #530. It shows some more work is required.

PR #308, opened 2014-08-22
NH-1440 - Common base type for ISession and IStatelessSession

rjperes:


oskarb:
The ISharedSession interface proposed by this PR contains more methods than the corresponding SharedSession interface in Hibernate. I think this should be adressed, so that we don't deviate unless there is good reason.

Some notable differences:
* In Hibernate, some methods have been pushed even furter up to another interface called QueryProducer.
* In Hibernate, methods such as Update() and Delete() are NOT present in SharedSession. Probably because despite having the same name and parameters, the semantics of Session.Update() and StatelessSession.Update() are quite different. I think that makes sense.

gliljas:
I agree.

hazzik:
Fixes #858

hazzik:
This would be easier to re-do from scratch as a lot of things has changed since the PR.

alexbk66:
Any news on that?

fredericDelaporte:
As far as I know, no one works on this currently.

alexbk66:
I created a simple wrapper class - because I only need two functions, but seriously it's disgusting

`public class SessionWrapper
{
    IStatelessSession statelessSession;
    ISession session;

    public SessionWrapper(IStatelessSession session)
    {
        this.statelessSession = session;
    }

    public SessionWrapper(ISession session)
    {
        this.session = session;
    }

    public T Get<T>(object id)
    {
        if (statelessSession != null)
            return statelessSession.Get<T>(id);
        else
            return session.Load<T>(id);
    }

    public IQueryOver<T, T> QueryOver<T>() where T : class
    {
        if (statelessSession != null)
            return statelessSession.QueryOver<T>();
        else
            return session.QueryOver<T>();
    }
}
`

fredericDelaporte:
Here you are on an abandoned pull-request. The issue is still opened and up for grab, see #858.

PR #315, opened 2014-09-02
NH-3619 - Make default value of FlushMode configurable

rjperes:
Fixed: default flush mode is now Auto, as the NHibernate default. Was set to Unspecified.
https://nhibernate.jira.com/browse/NH-3619


hazzik:
@rjperes I've changed unconfigured flush mode back to "Auto".

Also, could you please document this new configuration setting?

rjperes:
Will do.

PR #346, opened 2014-09-27
NH-2401 - Method for specifying IType of LINQ parameter

rjperes:
Implementation and simple unit test
https://nhibernate.jira.com/browse/NH-2401
@oskarb @hazzik: Debugging the code, it is clear that the parameter type is being correctly set, but I'm not sure how to do it in a unit test, have you got any suggestions?


hazzik:
I'm going to remove `MappedAsGenerator` and other stuff. Leave only changes to `ExpressionParameterVisitor`, `LinqExtensionMethods`, and tests.

hazzik:
It seems that `MappedAs` from this pull request does not work on parameters at all, because it is being stripped by `NhPartialEvaluatingExpressionTreeVisitor` before `ExpressionParameterVisitor` have a chance to process it.

PR #347, opened 2014-10-02
NH-3714 - Query by Example in Linq

rjperes:
https://nhibernate.jira.com/browse/NH-3714


hazzik:
That weekend never came. Closing as per JIRA.

PR #357, opened 2014-10-13
NH-3719

rjperes:
https://nhibernate.jira.com/browse/NH-3719


hazzik:
It just works with Accessor.Field:

	case Accessor.Field:
		string partialFieldNamingStrategyName = GetNamingFieldStrategy();
		if (partialFieldNamingStrategyName != null)
		{
			setAccessor("field." + partialFieldNamingStrategyName);
		}
		else
		{
			setAccessor("field");
		}
		break;

PR #363, opened 2014-10-18
NH-3724 - Added support for notification handlers

rjperes:
Added support for notification handlers (InfoMessage event in most ADO.NET drivers)


hazzik:
Replaced by #2113

PR #364, opened 2014-10-20
NH-3727 - Unit tests showing bug and bugfix proposition

itmagination:


hazzik:
@itmagination can you please make usual method overload instead of default parameter?

PR #365, opened 2014-10-22
Fixed PropertyExpression.ToString when RHS is a projection

LordJZ:
Code to reproduce:

```
Restrictions.EqProperty("A", Projections.Property("B")).ToString()
```

Expected output:

```
"A = B"
```

Actual output:

```
"A = "
```


hazzik:
Can someone create/find a jira issue for this?

PR #371, opened 2014-11-12
NH-3735 - Delete in Criteria and QueryOver

rjperes:
https://nhibernate.jira.com/browse/NH-3735


hazzik:
Need to implement as a native feature, not as a hack on top of `select`.

PR #372, opened 2014-11-18
NH-3448 - Fix GuessEntityName

rjperes:
https://nhibernate.jira.com/browse/NH-3448


hazzik:
@rjperes am I correct that you are trying to pass a proxy into a stateless session? As far as I remember stateless sessions do not support proxies.

hazzik:
Simplest fix to this issue (if it were an issue*) would be to update `StatelessSessionImpl.GetEntityPersister` method to call `BestGuessEntityName` instead of `GuessEntityName`.

PR #373, opened 2014-11-19
Fixed both dynamic-update cases for NH-3512

DmitryNaumov:
This is rebased and squashed copy of #220.
There is still issue is narrow base class (for example with Id only) when we can simulate fake base class property update to force version increment...


hazzik:
@DmitryNaumov can you please take a look at PostgreSQL fails?

> ERROR: 42883: operator does not exist: timestamp without time zone = bytea

DmitryNaumov:
Does it fail in tests?

hazzik:
@DmitryNaumov  Yes, the problem is with version column sql-type

PR #375, opened 2014-11-19
NH-2319 - Add ability to query collections with AsQueryable()

hazzik:


src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -95,26 +95,30 @@ public IQueryExpressionPlan GetHQLQueryPlan(IQueryExpression queryExpression, bo
...
+			var key = new FilterQueryPlanKey(queryExpression.Key, collectionRole, shallow, enabledFilters);
+			var plan = (IQueryExpressionPlan) planCache[key];
 
 			if (plan == null)
 			{
 				if (log.IsDebugEnabled)
 				{
-					log.Debug("unable to locate collection-filter query plan in cache; generating (" + collectionRole + " : "
-							  + filterString + ")");
+					log.Debug(string.Format("unable to locate collection-filter query plan in cache; generating ({0} : {1})", collectionRole, queryExpression.Key));
```
fredericDelaporte: I guess at the time of this change, $"" formatting was not available!

I am frequently a bit surprised by those `if (log.IsDebugEnabled)`: they are pointless if they are just about avoiding formatting without any value evaluation, aren't they? `log.DebugFormat("", args)` already handle that.

src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
```diff
@@ -749,7 +759,10 @@ IASTNode CreateFromFilterElement(IASTNode filterEntity, IASTNode alias)
 			{
 				log.Debug("createFromFilterElement() : processed filter FROM element.");
 			}
-	
+
+			if (fromElementFound)
+				return (IASTNode) adaptor.Nil();
```
fredericDelaporte: Has this been done because some of the calls above would have side effects? If yes, maybe a comment would be welcome. Otherwise I would tend to move this `return` in the `if (fromElement == null)` clause.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -604,13 +604,60 @@ public override void CloseSessionFromDistributedTransaction()
...
+		}
+
+		protected override void ListFilter(object collection, IQueryExpression queryExpression, QueryParameters queryParameters, IList results)
+		{
+			using (new SessionIdLoggingContext(SessionId))
+			{
+				CheckAndUpdateSessionStatus();
+				queryParameters.ValidateParameters();
+				var plan = GetFilterQueryPlan(collection, queryExpression, queryParameters, false);
+
```
fredericDelaporte: Almost dup code here. I think this need to be refactored with above method.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -753,27 +810,86 @@ public void Lock(string entityName, object obj, LockMode lockMode)
...
+				{
+					parameters.PositionalParameterValues[0] = entry.LoadedKey;
+					parameters.PositionalParameterTypes[0] = entry.LoadedPersister.KeyType;
+				}
+
+				return plan;
+			}
+		}
+
+		private IQueryExpressionPlan GetFilterQueryPlan(object collection, string filter, QueryParameters parameters, bool shallow)
```
fredericDelaporte: Same here, lot of dup code with above method, should probably be refactored.

src/NHibernate/Linq/NhQueryable.cs
```diff
@@ -46,11 +46,17 @@ public NhQueryable(IQueryProvider provider, Expression expression, string entity
 			EntityName = entityName;
 		}
 
+		public NhQueryable(ISessionImplementor session, object collection)
+			: base(new DefaultQueryProvider(session, collection))
```
fredericDelaporte: Shouldn't it use the `QueryProviderFactory`?

src/NHibernate/Collection/Generic/PersistentGenericBag.cs
```diff
@@ -162,6 +167,26 @@ public bool IsReadOnly
...
+		}
+
+		public IQueryProvider Provider
+		{
+			get { return InnerQueryable.Provider; }
+		}
+
+		private IQueryable<T> InnerQueryable
+		{
+			get { return queryable ?? (queryable = new NhQueryable<T>(Session, this)); }
```
hazzik: `Lazy<>` is not really required. What's required is a proper implementation of `IQueryable<>`, and it turns out that `NhQueryable<>` is doing almost everything what we need.

src/NHibernate/Collection/Generic/PersistentGenericBag.cs
```diff
@@ -46,6 +49,8 @@ public class PersistentGenericBag<T> : AbstractPersistentCollection, IList<T>, I
 		 */
 		private IList<T> _gbag;
 
+		private IQueryable<T> queryable;
```
hazzik: It's a hundreds years old PR, so I don't remember/know why I didn't put underscore in front of the field name.,

src/NHibernate/Engine/Query/QueryPlanCache.cs
hazzik: >I guess at the time of this change, $"" formatting was not available!
True. However, It seems that it shall be changed to `log.DebugFormat(..)`

>`if (log.IsDebugEnabled)`

It seems just to avoid doing string concatenations and win some milliseconds and object allocations (in some cases it's actually a huge win)  

Haven't found anything on C#, so the links are about Java, but internals are almost the same, so 

- http://stackoverflow.com/questions/963492/in-log4j-does-checking-isdebugenabled-before-logging-improve-performance
- https://www.igorkromin.net/index.php/2015/05/12/are-guard-statements-like-isdebugenabled-necessary-when-using-log4j/
- http://learnertobeginner.blogspot.co.nz/2010/06/how-to-use-isdebugenabled-effectly.html

src/NHibernate/Linq/NhQueryable.cs
hazzik: Yes. But it wasn't here when this code was written.

hazzik:
>Is there a reason for not including idbags (PersistentIdentifierBag) and lists (PersistentGenericList) in the feature?

It's because this feature is really just a strongly-typed wrapper on a filter collection feature which does not support these collections

hazzik:
Also, Hibernate is moving towards this direction as well.

They are going to:

- deprecate session#createFilter in [HHH-11390](https://hibernate.atlassian.net/browse/HHH-11390)
- shift support for "collection filtering" to Java 8 Streams (a java version of LINQ) [HHH-10962](https://hibernate.atlassian.net/browse/HHH-10962)

src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
hazzik: As I understand after quick look-around. There are some differences in the AST between HQL and LINQ. It seems that LINQ AST already contains `fromElement` in the tree, and HQL does not. 

We still need to do all these manipulations with the `fromElement`, but we do not want to add it to the tree (because it is already there).

hazzik:
About the differences in LINQ vs HQL:

Linq generates following AST:

```
( query ( select_from ( from ( range NHibernate.Test.NHSpecificTest.NH2319.Child x ) ) ( select x ) ) ( where ( == ( . x Name ) ( : p1 ) ) ) )
```

And HQL:

```
"( query ( SELECT_FROM {filter-implied FROM} ) ( where ( = ( . this Name ) 'Jack' ) ) )"
```

So, it's better to eliminate the initial difference in the AST tree instead of modifying CreateFromFilterElement

src/NHibernate/Engine/Query/QueryPlanCache.cs
fredericDelaporte: I am used to rarely need the explicit check, because I am mostly using `Format` suffixed methods when having to format a log. Of course sometime we need to format with computed values, or to have a more complex message construction. Then I use the explicit check. I am sensible to this because log messages are almost bloat code in my view, and I rather keep them as compact as possible. (Actually I tend to only log exceptions. I fully agree with [Jeff Atwood blog](https://blog.codinghorror.com/the-problem-with-logging/) on this.)

fredericDelaporte:
I never needed to use filters so I am not sure, but are filter supporting batch loading of collection? I tend to think yes, and then, deprecating filters for queryable collections may cause a significant performances loss, unless we find a way to maintain batching with queryable collections.

fredericDelaporte:
Hibernate team seems to want to only remove the special query on collection case. I was thinking about the more global filters functionality, declarable in mappings. Well, I nether find this query on collection feature very useful, it is so easy to just write a usual query doing the same job.

hazzik:
>Hibernate team seems to want to only remove the special query on collection case.

Yes, only this `Session#createFilter` feature. They want to replace HQL version with Java Streams API.

>global filters functionality, declarable in mappings

Not these filters:) I know it confusing to have 2 distinct features share similar names

> Well, I neither find this query on collection feature very useful, it is so easy to just write a usual query doing the same job.

Actually, [some] people are desperate to have this feature implemented. It's one of the highest voted features in the JIRA.

hazzik:
Later `SELECT_FROM {filter-implied FROM}` is replaced by `HqlFilterPreprocessor` with `SELECT_FROM ( FROM NHibernate.DomainModel.Many this )` which makes the queries identical except the alias name (`this` for HQL, and whatever for LINQ)

fredericDelaporte:
Rebased. Once rebased some tests where failing when run in sequence with other, due to a trouble when a filtered query plan was retrieved from cache.
And refactored.

src/NHibernate/Collection/AbstractPersistentCollection.cs
```diff
@@ -79,7 +79,7 @@ public void Reset()
 				}
 			}
 		}
-
+
```
fredericDelaporte: Cleanup glitch, going to fix and repush forced.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -676,70 +707,87 @@ public void Lock(string entityName, object obj, LockMode lockMode)
...
+				role => Factory.QueryPlanCache.GetFilterQueryPlan(queryExpression, role, shallow, EnabledFilters));
+		}
+
+		private IQueryExpressionPlan GetFilterQueryPlan(object collection, string filter, QueryParameters parameters, bool shallow)
+		{
+			return GetFilterQueryPlan(collection, parameters,
+				role => Factory.QueryPlanCache.GetFilterQueryPlan(filter, role, shallow, EnabledFilters));
+		}
+
+		private IQueryExpressionPlan GetFilterQueryPlan(object collection, QueryParameters parameters, Func<string, IQueryExpressionPlan> getPlanForRole)
```
fredericDelaporte: I usually avoid this kind of pattern, putting code in common thanks to function pointers passed around. But here, with the function being defined for all cases right above, I think it is ok.

src/NHibernate/Linq/QueryProviderFactory.cs
```diff
@@ -9,16 +9,20 @@ static class QueryProviderFactory
...
-				return new DefaultQueryProvider(session);
+				return new DefaultQueryProvider(session, collection);
 			}
 			else
 			{
-				return Activator.CreateInstance(session.Factory.Settings.LinqQueryProviderType, session) as INhQueryProvider;
+				// For backward compatibility, prioritize using the version without collection.
+				return (collection == null
+					? Activator.CreateInstance(session.Factory.Settings.LinqQueryProviderType, session)
+					: Activator.CreateInstance(session.Factory.Settings.LinqQueryProviderType, session, collection)) as INhQueryProvider;
```
fredericDelaporte: Minimal backward compatibility handling.

src/NHibernate/Impl/SessionImpl.cs
fredericDelaporte: Does not seem compatible with Async generator, causes the removal of `GetFilterQueryPlanAsync`. Now I wonder what IO are there in that function, going to check.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -676,70 +707,88 @@ public void Lock(string entityName, object obj, LockMode lockMode)
...
 			{
 				if (collection == null)
-				{
-					throw new ArgumentNullException("collection", "null collection passed to filter");
-				}
+					throw new ArgumentNullException(nameof(collection), "null collection passed to filter");
+				if (filter != null && queryExpression != null)
+					throw new ArgumentException($"Either {nameof(filter)} or {nameof(queryExpression)} must be specified, not both.");
+				if (filter == null && queryExpression == null)
+					throw new ArgumentException($"{nameof(filter)} and {nameof(queryExpression)} were both null.");
```
fredericDelaporte: I do not like this kind of parameter semantic but do not see a better solution for avoiding code duplication and not using a pattern not handled by async generator.

src/NHibernate.Test/NHSpecificTest/Logs/LogsFixture.cs
```diff
@@ -73,6 +74,7 @@ public TextLogSpy(string loggerName, string pattern)
 				};
 				loggerImpl = (Logger)LogManager.GetLogger(typeof(LogsFixture).Assembly, loggerName).Logger;
 				loggerImpl.AddAppender(appender);
+				previousLevel = loggerImpl.Level;
 				loggerImpl.Level = Level.All;
```
fredericDelaporte: This spy was used only for `NHibernate.SQL`, but that is already an heavily logging logger. Better restore its previous log level.

src/NHibernate.Test/SystemTransactions/ResourceManagerFixture.cs
```diff
@@ -684,7 +684,6 @@ void Clone_TransactionCompleted(object sender, TransactionEventArgs e)
 		[OneTimeSetUp]
 		public void TestFixtureSetUp()
 		{
-			((Logger)_log.Logger).Level = log4net.Core.Level.Info;
 			_spy = new LogSpy(_log);
```
fredericDelaporte: Undue log level manual change, since `LogSpy` set it to `Debug` anyway.

src/NHibernate.Test/Async/NHSpecificTest/NH2319/Fixture.cs
```diff
@@ -68,12 +77,41 @@ public abstract class FixtureBaseAsync : TestCaseMappingByCode
 		}
 
 		[Test]
+		public async Task ShouldBeAbleToReuseQueryPlanAsync()
+		{
+			await (ShouldBeAbleToFindChildrenByNameAsync());
+			using (var spy = new LogSpy(typeof(QueryPlanCache)))
+			{
+				Assert.That(ShouldBeAbleToFindChildrenByNameAsync, Throws.Nothing);
+				AssertPlanCacheHit(spy);
```
fredericDelaporte: Now checking the cache was actually hit, demonstrating using it works.

src/NHibernate.Test/Async/NHSpecificTest/NH2319/Fixture.cs
```diff
@@ -68,12 +77,41 @@ public abstract class FixtureBaseAsync : TestCaseMappingByCode
...
+			}
+		}
+
+		private static void AssertPlanCacheHit(LogSpy spy) =>
+			// Each query currently ask the cache two times, so asserting reuse requires to check cache has not been missed
+			// rather than only asserting it has been hit.
+			Assert.That(spy.GetWholeLog(),
+				Contains.Substring("located collection-filter query plan in cache (")
+				.And.Not.Contains("unable to locate collection-filter query plan in cache"));
```
fredericDelaporte: The "not contains" is not much robust: if the log message in cache is changed, this will defeat this assert. So having a cache miss test would be nice, but this has another trouble: how to be sure no other test has put an adequate plan in cache?

src/NHibernate.Test/NHSpecificTest/NH2319/Fixture.cs
```diff
@@ -56,12 +66,34 @@ public void ShouldBeAbleToPerformComplexFiltering()
 		}
 
 		[Test]
+		public void ShouldBeAbleToReuseQueryPlan()
+		{
+			ShouldBeAbleToFindChildrenByName();
+			using (var spy = new LogSpy(typeof(QueryPlanCache)))
+			{
+				Assert.That(ShouldBeAbleToFindChildrenByName, Throws.Nothing);
+				AssertFilterPlanCacheHit(spy);
```
fredericDelaporte: Now checking the cache was actually hit, demonstrating using it works.

src/NHibernate.Test/NHSpecificTest/NH2319/Fixture.cs
```diff
@@ -118,6 +150,35 @@ public void TestFilter()
...
+		}
+
+		private const string _filterPlanCacheMissLog = "unable to locate collection-filter query plan in cache";
+
+		private static void AssertFilterPlanCacheHit(LogSpy spy) =>
+			// Each query currently ask the cache two times, so asserting reuse requires to check cache has not been missed
+			// rather than only asserting it has been hit.
+			Assert.That(spy.GetWholeLog(),
+				Contains.Substring("located collection-filter query plan in cache (")
+				.And.Not.Contains(_filterPlanCacheMissLog));
```
fredericDelaporte: The "not contains" is not much robust: if the log message in cache is changed, this will defeat this assert silently. Thus the additional `TestPlanCacheMiss()` above.

src/NHibernate.Test/Async/NHSpecificTest/NH2319/Fixture.cs
```diff
@@ -0,0 +1,428 @@
...
+				AssertFilterPlanCacheMiss(spy);
+			}
+		}
+
+		private static void AssertFilterPlanCacheHit(LogSpy spy) =>
+			// Each query currently ask the cache two times, so asserting reuse requires to check cache has not been missed
+			// rather than only asserting it has been hit.
+			Assert.That(spy.GetWholeLog(),
+				Contains.Substring("located collection-filter query plan in cache (")
+				.And.Not.Contains(_filterPlanCacheMissLog));
```
fredericDelaporte: [#49](https://github.com/maca88/AsyncGenerator/issues/49), missing const in async file.

doc/reference/modules/query_linq.xml
```diff
@@ -54,6 +54,14 @@ using NHibernate.Linq;]]></programlisting>
       This will be executed as a query on that <literal>cat</literal>'s kittens without loading the
       entire collection.
     </para>
+    <para>
+      If the collection is a map, call <literal>AsQueryable</literal> on its <literal>Values</literal>
+      property.
+    </para>
```
fredericDelaporte: Added support for map.

src/NHibernate.Test/NHSpecificTest/NH1612/NativeSqlCollectionLoaderFixture.cs
```diff
@@ -39,8 +39,8 @@ public void LoadCompositeElementsWithWithSimpleHbmAliasInjection()
 			Country country = LoadCountryWithNativeSQL(CreateCountry(stats), "LoadAreaStatisticsWithSimpleHbmAliasInjection");
 
 			Assert.That(country, Is.Not.Null);
-			Assert.That((ICollection) country.Statistics.Keys, Is.EquivalentTo((ICollection) stats.Keys), "Keys");
-			Assert.That((ICollection) country.Statistics.Values, Is.EquivalentTo((ICollection) stats.Values), "Elements");
+			Assert.That(country.Statistics.Keys, Is.EquivalentTo(stats.Keys), "Keys");
+			Assert.That(country.Statistics.Values, Is.EquivalentTo(stats.Values), "Elements");
```
fredericDelaporte: Those casts were likely required for an older NUnit. Since I have not additionally implemented `ICollection` on the map values wrapper, they were failing. We will probably have to document that as a possible breaking change.

src/NHibernate.Test/NHSpecificTest/NH2319/MapFixture.cs
```diff
@@ -0,0 +1,286 @@
...
+			using (session.BeginTransaction())
+			{
+				var parent = session.Get<Parent>(_parent1Id);
+
+				Assert.NotNull(parent);
+
+				var filtered = parent.ChildrenMap.Values
+					.AsQueryable()
+					.Where(x => x.Name == "Piter")
+					.SelectMany(x => x.GrandChildren)
```
fredericDelaporte: Not a map here. Supporting maps inside lambda would be another feature. (Currently already doable by pretending the map is a `ICollection<TValue>` with a cast, but a bit hackish.)

src/NHibernate.Test/NHSpecificTest/NH2319/MapFixture.cs
```diff
@@ -0,0 +1,286 @@
...
+
+		protected override HbmMapping GetMappings()
+		{
+			var mapper = new ModelMapper();
+			mapper.Class<Parent>(rc =>
+			{
+				rc.Id(x => x.Id, m => m.Generator(Generators.GuidComb));
+				rc.Property(x => x.Name);
+				rc.Map(x => x.ChildrenMap,
+					map => map.Cascade(Mapping.ByCode.Cascade.All | Mapping.ByCode.Cascade.DeleteOrphans),
```
fredericDelaporte: I have not found how to specify the index, although we can specify it in hbm. Missing feature of by code mapping? Anyway, works with defaults for the test.

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -1,26 +1,21 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../../build-common/NHibernate.props" />
-
```
fredericDelaporte: Hum, probable Rider glitch... Bad point for that IDE.

src/NHibernate/Collection/Generic/PersistentGenericMap.cs
```diff
@@ -568,5 +578,95 @@ public void Operate()
...
 		#endregion
+
+		[Serializable]
+		private class ValuesWrapper : ICollection<TValue>, IQueryable<TValue>
+		{
+			private readonly PersistentGenericMap<TKey, TValue> _map;
+
+			public ValuesWrapper(PersistentGenericMap<TKey, TValue> map)
+			{
+				_map = map;
```
fredericDelaporte: Just storing the persistent map, not the values, since the underlying map may still be changed at that point.

src/NHibernate/Collection/Generic/PersistentGenericMap.cs
```diff
@@ -22,15 +25,22 @@ namespace NHibernate.Collection.Generic
 	[DebuggerTypeProxy(typeof(DictionaryProxy<,>))]
 	public partial class PersistentGenericMap<TKey, TValue> : AbstractPersistentCollection, IDictionary<TKey, TValue>, ICollection
 	{
-		protected IDictionary<TKey, TValue> WrappedMap;
+		protected IDictionary<TKey, TValue> WrappedMap { get; private set; }
```
fredericDelaporte: Protected only for one test file, looks like it should have been private but experienced a probable "rogue" change. At least switched to property with private setter.

src/NHibernate/Collection/Generic/PersistentGenericMap.cs
```diff
@@ -22,15 +25,22 @@ namespace NHibernate.Collection.Generic
...
+		private readonly ICollection<TValue> _wrappedValues;
 
-		public PersistentGenericMap() { }
+		public PersistentGenericMap()
+		{
+			_wrappedValues = new ValuesWrapper(this);
```
fredericDelaporte: x3, one for each constructor: they cannot call each over and the field cannot be initialized in place when using `this`.

src/NHibernate/Collection/Generic/PersistentGenericMap.cs
```diff
@@ -347,8 +358,7 @@ public ICollection<TValue> Values
 		{
 			get
 			{
-				Read();
-				return WrappedMap.Values;
+				return _wrappedValues;
```
fredericDelaporte: `Read()` moved to `ValuesWrapper` `ICollection<TValue>` members. Otherwise, reaching `Values` for calling `AsQueryable` was loading the map.

fredericDelaporte:
I will not merge it myself since it would be a bit auto-approval of my own non cosmetic changes.

src/NHibernate.Test/NHSpecificTest/NH2319/MapFixture.cs
hazzik: It is supposed that you use the `Map` overload which takes 4 lambdas, but `IMapKeyRelation` is missing the index part.

hazzik: Or probably ` Element` will behave as you need.

src/NHibernate/Collection/Generic/PersistentGenericMap.cs
```diff
@@ -568,5 +578,95 @@ public void Operate()
...
+				return _map.WrappedMap.Values.GetEnumerator();
+			}
+
+			IEnumerator IEnumerable.GetEnumerator()
+			{
+				_map.Read();
+				return GetEnumerator();
+			}
+
+			public void Add(TValue item)
```
hazzik: Dictionary's "Values" collection is read-only.

src/NHibernate/Collection/Generic/PersistentGenericMap.cs
hazzik: Can this please be reverted? And the test needs to change to use `.Entries(...)` method instead.

src/NHibernate/Collection/Generic/PersistentGenericMap.cs
fredericDelaporte: Must still implement the interface, and I am just letting the throw be done by the underlying collection. But just for doing that, it triggers the read. Would you rather have a direct throw not reading at all? Since it is an error case anyway, I do not tend to optimize it.

hazzik: I don't think this actually matters much, so let it be this way.

fredericDelaporte: It does not matter much but I have already changed it anyway, just not yet pushed.

hazzik: But how about "I do not tend to optimize it."?

PR #377, opened 2014-11-20
NH-2140 - Consolidate common functionality of ICriteria, IQuery, ISQLQuery, etc

rjperes:
Unifies all query APIs (LINQ, SQL, HQL, Criteria and QueryOver) under a common interface (IQueryOptions).
Includes implementation of NH-3299, NH-3470 and NH-2285.

### DIFF
--- a/src/NHibernate.Test/CommonInterfaceTest.cs
+++ b/src/NHibernate.Test/CommonInterfaceTest.cs
@@ -1,0 +1,142 @@
+using System;
+using System.Collections;
+using NHibernate.DomainModel;
+using NHibernate.Linq;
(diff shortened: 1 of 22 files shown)


teamneusta:
Great, we really like this.

gliljas:
I disagree with this implementation.

1. The List, List<T>, UniqueResult etc. methods don't belong on an IQueryOptions interface
2. IQueryable (Linq) should not be included. While it does indeed need Cachable etc. it should not provide SetFetchSize etc. Not even explicitly, since supporting it makes assumptions about the internals.

hazzik:
Closing as per @gliljas

PR #378, opened 2014-11-21
Fixes for NH-3480 and NH-3453

JKeyser:
@hazzik, I have rebased the previous pull request (#211) for these issues.

### DIFF
--- a/src/NHibernate/Persister/Collection/OneToManyPersister.cs
+++ b/src/NHibernate/Persister/Collection/OneToManyPersister.cs
@@ -64,69 +64,31 @@ namespace NHibernate.Persister.Collection
 		/// <returns></returns>
 		protected override SqlCommandInfo GenerateDeleteString()
 		{
+			var update = new SqlUpdateBuilder(Factory.Dialect, Factory)
+				.SetTableName(qualifiedTableName)
+				.AddColumns(JoinColumnNames, "null");
+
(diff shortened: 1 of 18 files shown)


src/NHibernate/Persister/Collection/OneToManyPersister.cs
```diff
@@ -66,8 +66,18 @@ protected override SqlCommandInfo GenerateDeleteString()
...
-				.AddColumns(KeyColumnNames, "null")
-				.SetIdentityColumn(KeyColumnNames, KeyType);
+				.AddColumns(JoinColumnNames, "null");
+
+			if (CollectionType.UseLHSPrimaryKey)
+			{
+				update.SetIdentityColumn(KeyColumnNames, KeyType);
+			}
+			else
+			{
```
hazzik: @JKeyser Why do we need this join?

It fails on SQLite, because it does not support joins at the update statements.

hazzik: Can it be just

```csharp
.AddColumns(JoinColumnNames, "null");
.SetIdentityColumn(KeyColumnNames, KeyType);
```
??

hazzik:
Also, tests are failing on PostgreSQL and MySQL, and this is a regression.

hazzik:
@JKeyser can you please review #404?

PR #384, opened 2014-11-26
NH-3739 - Fix usage of SqlString.Append method

lesnyh:
I develop NHibernate Driver for DBMS Linter SQL and found a problem in 88 line of AbstractStatementExecutor.cs. The problem is in this code:
whereJoinFragment.Append(" and ");
it should be:
whereJoinFragment = whereJoinFragment.Append(" and ");

Because of this bug I have problems in 6 tests. For example, NHibernate.Test.Hql.Ast.BulkManipulation.DeleteOnJoinedSubclass. In this test I have the following query with error:
insert into $$$TMP_Mammal SELECT mammal0_.animal as animal FROM Mammal
mammal0_, Animal mammal0_1_ WHERE
mammal0_.animal=mammal0_1_.idbody_weight>150


hazzik:
Merged in 569b49c4ead3f026f0a161c1b9b6833fb92e7216

PR #385, opened 2014-11-28
NH-3741 - Many to many mapping with interfaces.

neilcrowther:
Fix mapping bug when generic collection contains interfaces instead of entities (Jira #NH-3741)

### DIFF
--- a/src/NHibernate.Test/MappingByCode/IntegrationTests/NH3741/MapFixture.cs
+++ b/src/NHibernate.Test/MappingByCode/IntegrationTests/NH3741/MapFixture.cs
@@ -1,0 +1,80 @@
+﻿using System.Linq;
+using NHibernate.Cfg.MappingSchema;
+using NHibernate.Mapping.ByCode;
+using NUnit.Framework;
+
+namespace NHibernate.Test.MappingByCode.IntegrationTests.NH3741
+{
+	[TestFixture]
+	public class MapFixture
+	{
+		[Test]
+		public void TestMapManyToManyGenericCollectionBasedOnInterface()
+		{
+			var mapper = new ModelMapper();
+
+			mapper.Class<ParentWithInterfaceChild>(c =>
+			{
+				c.Id(x => x.Id, x => x.Generator(Generators.Identity));
+				c.IdBag(x => x.Children, bag => { }, rel => rel.ManyToMany());
+			});
+
+			mapper.Class<Child>(c =>
+			{
+				c.Id(x => x.Id, x => x.Generator(Generators.Identity));
+				c.Property(x => x.Description);
+			});
+
+			var hbmMapping = mapper.CompileMappingFor(new[] { typeof(ParentWithInterfaceChild), typeof(Child) });
+			var hbmClass = hbmMapping.RootClasses[0];
+			var hbmIdbag = hbmClass.Properties.OfType<HbmIdbag>().SingleOrDefault();
+
+			Assert.That(hbmIdbag, Is.Not.Null);
+			Assert.That(hbmIdbag.ElementRelationship, Is.InstanceOf<HbmManyToMany>());
+		}
+
+		[Test]
+		public void TestMapManyToManyGenericCollectionBasedOnEntity()
+		{
+			var mapper = new ModelMapper();
+
+			mapper.Class<ParentWithEntityChild>(c =>
+			{
+				c.Id(x => x.Id, x => x.Generator(Generators.Identity));
+				c.IdBag(x => x.Children, bag => { }, rel => rel.ManyToMany());
+			});
+
+			mapper.Class<Child>(c =>
+			{
+				c.Id(x => x.Id, x => x.Generator(Generators.Identity));
+				c.Property(x => x.Description);
+			});
+
+			var hbmMapping = mapper.CompileMappingFor(new[] { typeof(ParentWithEntityChild), typeof(Child) });
+			var hbmClass = hbmMapping.RootClasses[0];
+			var hbmIdbag = hbmClass.Properties.OfType<HbmIdbag>().SingleOrDefault();
+
+			Assert.That(hbmIdbag, Is.Not.Null);
+			Assert.That(hbmIdbag.ElementRelationship, Is.InstanceOf<HbmManyToMany>());
+		}
+
+		[Test]
+		public void TestMapManyToManyGenericCollectionBasedOnItem()
+		{
+			var mapper = new ModelMapper();
+
+			mapper.Class<ParentWithItemChild>(c =>
+			{
+				c.Id(x => x.Id, x => x.Generator(Generators.Identity));
+				c.IdBag(x => x.Children, bag => { }, rel => rel.ManyToMany());
+			});
+
+			var hbmMapping = mapper.CompileMappingFor(new[] { typeof(ParentWithItemChild) });
+			var hbmClass = hbmMapping.RootClasses[0];
+			var hbmIdbag = hbmClass.Properties.OfType<HbmIdbag>().SingleOrDefault();
+
+			Assert.That(hbmIdbag, Is.Not.Null);
+			Assert.That(hbmIdbag.ElementRelationship, Is.InstanceOf<HbmManyToMany>());
+		}
+	}
+}
--- a/src/NHibernate/Mapping/ByCode/AbstractExplicitlyDeclaredModel.cs
+++ b/src/NHibernate/Mapping/ByCode/AbstractExplicitlyDeclaredModel.cs
@@ -15,7 +15,8 @@ namespace NHibernate.Mapping.ByCode
 		private readonly HashSet<MemberInfo> dictionaries = new HashSet<MemberInfo>();
 		private readonly HashSet<MemberInfo> idBags = new HashSet<MemberInfo>();
 		private readonly HashSet<MemberInfo> lists = new HashSet<MemberInfo>();
-		private readonly HashSet<MemberInfo> manyToManyRelations = new HashSet<MemberInfo>();
+		private readonly HashSet<MemberInfo> keyManyToManyRelations = new HashSet<MemberInfo>();
+		private readonly HashSet<MemberInfo> itemManyToManyRelations = new HashSet<MemberInfo>();
 		private readonly HashSet<MemberInfo> manyToAnyRelations = new HashSet<MemberInfo>();
 		private readonly HashSet<MemberInfo> manyToOneRelations = new HashSet<MemberInfo>();
 		private readonly HashSet<MemberInfo> naturalIds = new HashSet<MemberInfo>();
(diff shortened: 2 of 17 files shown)


src/NHibernate/Mapping/ByCode/FakeModelExplicitDeclarationsHolder.cs
```diff
@@ -15,6 +15,8 @@ public class FakeModelExplicitDeclarationsHolder : IModelExplicitDeclarationsHol
 		private readonly IEnumerable<MemberInfo> idBags = Enumerable.Empty<MemberInfo>();
 		private readonly IEnumerable<MemberInfo> lists = Enumerable.Empty<MemberInfo>();
 		private readonly IEnumerable<MemberInfo> manyToManyRelations = Enumerable.Empty<MemberInfo>();
+		private readonly IEnumerable<MemberInfo> entityManyToManyRelations = Enumerable.Empty<MemberInfo>();
+		private readonly IEnumerable<MemberInfo> keyManyToManyRelations = Enumerable.Empty<MemberInfo>();
```
stefan-steinegger: Why does it need three lists? These lists are only used to know that a list item is many to many or a map key is many to many. As far as I can see, it is not used anywhere else.

neilcrowther: Needed to be able to tell the difference between key based many-to-many and entity based many-to-many relationships.  I could have used only the two lists and made the ManyToManyRelations property return the union of them both, but this seemed to make the Fake simpler to use for anyone that wanted to set the ManyToManyRelations list directly in other tests.

stefan-steinegger: I mean, you do not need the ManyToManyRelations property which returns the union at all.

stefan-steinegger:
These are valuable changes I would like to have in NH. It makes collection mapping by code a bit more stable and solves a couple of issues (particularly the one that makes it impossible for us to use NH 4). What is happening to these changes now?

neilcrowther:
OK, removed the extra collection and property.  should be good now.

src/NHibernate/Mapping/ByCode/FakeModelExplicitDeclarationsHolder.cs
neilcrowther: I've removed the extra property now.

stefan-steinegger:
Where can I see the most actual changes?

neilcrowther:
all files affected are here: https://github.com/nhibernate/nhibernate-core/pull/385/files

stefan-steinegger:
Ok, I probably misunderstood. Or you did. I mean, in ExplicitDeclarationsHolder, the three lists could be reduced to two. And shouldn't the KeyManyToManyRelations be used somewhere? To check whether the key of a map is a many-to-many relation. Otherwise there are similar issues with many-to-many keys as we just solved for the values. 
By the way, "EntityManyToManyRelations" is probably not a very good naming, because every many-to-many is referencing entities. I would prefer "ItemManyToManyRelations" in contrast to "KeyManyToManyRelations". It would be used for the items of lists, bags and sets and also for the values of dictionaries (maps).

neilcrowther:
OK, I see what you mean now (I think), I'm happy to change the names to "ManyToManyRelations" and "ItemManyToManyRelations".  Yeah, the "ItemManyToManyRelations" should be used, not sure where though.  I can't find any existing tests covering "items of lists, bags and sets and also for the values of dictionaries (maps)", do you know of any that exist?  Or any idea how I can test mapping for these would be great (not mapped directly to items myself before, so I want to ensure that I'm testing it correctly).

neilcrowther:
I've done the name changes (not updated the pull request yet), and looking into where the ItemManyToMany is used, its being set in the MapKeyManyToManyCustomizer.  Any pointers on how to set up a test to use this in mapping?  I've does a quick test just using 

    c.IdBag(x => x.Children, bag => { }, rel => rel.ManyToMany());

Where Children is a string, but that still uses the ManyToManyCustomizer.  I'm obviously missing something, but don't have enough knowledge to see what it is.  Any pointers would be appreciated.

stefan-steinegger:
The real problem comes with maps (dictionaries). There are four cases to test: map a primitive value to a primitive value, a primitive value to an entity, an entity to a primitive value and an entity to an entity. All these cases need to separately decide whether the key or the entity is many-to-many.

In ModelMapper.MapDictionary you can see these two calls: DetermineMapKeyRelationType and DetermineCollectionElementRelationType. The first is making a call to IsEntity, which should be changed to IsManyToManyKey. The second one had been fixed by you.

neilcrowther:
OK, name changes done and DetermineMapKeyRelationType  is now calling IsManyToManyKey.

src/NHibernate.Test/MappingByCode/IntegrationTests/NH3741/MapFixture.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            var hbmMapping = mapper.CompileMappingFor(new[] { typeof(ParentWithItemChild) });
+            var hbmClass = hbmMapping.RootClasses[0];
+            var hbmIdbag = hbmClass.Properties.OfType<HbmIdbag>().SingleOrDefault();
+
+            Assert.That(hbmIdbag, Is.Not.Null);
+            Assert.That(hbmIdbag.ElementRelationship, Is.InstanceOf<HbmManyToMany>());
+        }
+	}
+}
```
stefan-steinegger: Please add tests with dictionaries. e.g. Dictionary<string, Child> and Dictionary<Child, string>. This way you test whether the model mapper can distinguish between the key and the item.

src/NHibernate/Mapping/ByCode/AbstractExplicitlyDeclaredModel.cs
```diff
@@ -16,6 +16,7 @@ public abstract class AbstractExplicitlyDeclaredModel : IModelExplicitDeclaratio
 		private readonly HashSet<MemberInfo> idBags = new HashSet<MemberInfo>();
 		private readonly HashSet<MemberInfo> lists = new HashSet<MemberInfo>();
 		private readonly HashSet<MemberInfo> manyToManyRelations = new HashSet<MemberInfo>();
+		private readonly HashSet<MemberInfo> itemManyToManyRelations = new HashSet<MemberInfo>();
```
stefan-steinegger: Now I'm confused. The many to many keys are gone. At the end you have to distinguish between many-to-many items and many-to-many keys. Just many-to-many and nothing more is not clear enough.

src/NHibernate/Mapping/ByCode/ModelMapper.cs
```diff
@@ -1282,19 +1282,12 @@ protected virtual ICollectionElementRelationMapper DetermineCollectionElementRel
 			{
 				return new OneToManyRelationMapper(propertyPath, ownerType, collectionElementType, modelInspector, customizerHolder, this);
 			}
-			//NH-3667 & NH-3102
-			//check if property is really a many-to-many: as detected by modelInspector.IsManyToMany and also the collection type is an entity
-			if (modelInspector.IsManyToMany(property) == true)
+			//NH-3667 & NH-3102 && NH-3741
+			// many to many split from key many to many so that XML mappings and Interfaces work with many to many.
+			// MapKeyManyToManyCustomizer now registers itself as KeyManyToMany, so will return false for IsManyToMany(property).
+			if (modelInspector.IsManyToMany(property))
```
stefan-steinegger: Almost perfect. Should be IsManyToManyItem.

src/NHibernate/Mapping/ByCode/ModelMapper.cs
```diff
@@ -1309,10 +1302,7 @@ protected virtual ICollectionElementRelationMapper DetermineCollectionElementRel
 
 		private IMapKeyRelationMapper DetermineMapKeyRelationType(MemberInfo member, PropertyPath propertyPath, System.Type dictionaryKeyType)
 		{
-			// Perhaps we have to change IModelInspector with IsDictionaryKeyManyToMany(member), IsDictionaryKeyComponent(member) and so on
-
-			//if (modelInspector.IsManyToMany(member) || modelInspector.IsOneToMany(member))
-			if (modelInspector.IsEntity(dictionaryKeyType))
+			if (modelInspector.IsManyToManyKey(dictionaryKeyType))
```
stefan-steinegger: Exactly. That's great.

src/NHibernate/Mapping/ByCode/SimpleModelInspector.cs
```diff
@@ -73,6 +72,11 @@ public bool IsManyToMany(MemberInfo member)
 				return ManyToManyRelations.Contains(member);
 			}
 
+			public bool IsManyToManyKey(MemberInfo member)
+			{
+				return ItemManyToManyRelations.Contains(member);
```
stefan-steinegger: There is a mistake. Should be KeyManyToManyRelations (which had been gone. See comment above.)

src/NHibernate.Test/MappingByCode/IntegrationTests/NH3741/MapFixture.cs
neilcrowther: This is the bit I'm really not sure about.  I'd be OK writing a mapping for a dictionary property, but cannot get my head around how to do the same in the structure of a test (using the Model Mapper directly).  Any pointers would be appreciated.

src/NHibernate/AssemblyInfo.cs
```diff
@@ -0,0 +1,28 @@
+using System;
```
hazzik: Please remove this file. It is not intended to be committed.

hazzik:
Also tests are failing, please investigate

hazzik:
@neilcrowther can you please investigate tests? http://teamcity.codebetter.com/viewLog.html?buildId=178709&buildTypeId=bt7&guest=1

neilcrowther:
Yes, will do.  Matter of fitting this around life at the moment.

stefan-steinegger:
I've fixed it on my machine. There are also some small errors in the code. How can I share it?

neilcrowther:
Pull request against https://github.com/neilcrowther/nhibernate-core would be fine.  Also happy to do the changes again myself if you want to email me the details (depends on how many changes I suppose).  Whatever is easiest for you.  Just made my email public if that's what you want to use.

stefan-steinegger:
I try to send you a pull request. I working the first time with git, so please bear with me. I fixed two smaller issues in the code and changed the unit tests a bit. They are similar to others which test the generated Hbm classes. This seems to be working quite well, because the tests turn red pretty reliably when you change anything in the code.

neilcrowther:
Thanks to Stefan, the tests now pass (and I've removed the AssemblyInfo.cs file I included in error).

nptomlin:
Is it possible to find out when this PR will be merged?

stefan-steinegger:
IMHO, this pull request could and should be merged for the next NH release. I don't have any clue who is deciding about and merging changes like this. I helped a little bit with the changes and reviewed them and it looks perfect. By the way, it's the blocker which makes it impossible for me to update NH 3.3.0 in our large project to the latest and greatest version of NHibernate.

hazzik:
@neilcrowther this has been cleaned up and merged to master as 09d5f5da7b4817b7eff9f0adfb31718bbbde5f40

Thanks a lot.

PR #391, opened 2015-01-01
NH-3488 - Strongly typed Inserts, Updates and Deletes

gliljas:
https://nhibernate.jira.com/browse/NH-3488

The bulk of the code in the commit is the unit tests, which is a copy of the tests in Hql/Ast/BulkManipulation.cs. 

It would have been nice to reuse the same code but almost all the mappings in the original have casing problems, i.e they refer to lower case properties, but the properties are not lower cased. The HQL tests work, since they refer to the mapped properties, not the real ones. One solution would have been to check against the mapping, but I found the existing behavior strange and other parts of the Linq implementation doesn't do this, so I'll leave it as is for now.

The goal was to provide the new functionality by deriving from the existing Linq system, but that was easier said than done. Lots of code in NHibernate is very hard to derive from, with static calls and massive amounts of logic in the constructors. I decided to create derived types of NHLinqExpression, and handle most of the initialization in the constructor call. I could have created a derived QueryModelVisitor and IntermediateHqlTree too, but opted not to.

The syntax for the update and insert is worth discussing, but I wanted to support both a member initialization version and an explicit one. They could not be implemented with the same method name, since the compiler then sometimes will not be able to distinguish between the overloads.

For inserts we have:

`session.Query<Cat>().Insert().As(x=>new Dog{Name=x.Name});`

or

`session.Query<Cat>().Insert().Into<Dog>(x=>x.Set(y=>y.Name,y=>y.Name));`

For updates we have:

`session.Query<Cat>().Update().As(x=>new Cat{Name=x.Name + " the cat"});`

or

`session.Query<Cat>().Update().Assign(x=>x.Set(y=>y.Name,y=>y.Name + " the cat"));`

A better name than "Assign" is welcome. Or maybe for "Set".


gliljas:
A bit of clarification.

While the delete part was easy enough - it just throws away the select and replaces it with a delete - the update and insert required a bit more. The concept used was to embed the UPDATE SET clause and the INSERT INTO [..] SELECT [..] in a projected SELECT clause.

`session.Query<Cat>().Update().As(x=>new Cat{Name=x.Name + " the cat"});`

becomes

`session.Query<Cat>().Select(x=>new Dictionary<string,object>{{"Name",x.Name + " the cat"}});`

By doing this, it is ensured that anything inside the clauses, such as subqueries, parameters etc., is going the same rewriting process as everything else. It maybe feels like bit of a hack, and a full refactoring of the query modeling could open up for a more explicit path, but right now I think this is OK.

Also, regarding the syntax. The reason the "fluent" stuff was needed was that without it...

`session.Query<Cat>().Insert().Into<Dog>(x=>x.Set(y=>y.Name,y=>y.Name));`

would have to be 

session.Query<Cat>().Insert<Cat,Dog>(x=>x.Set(y=>y.Name,y=>y.Name));

since the compiler can't infer just Dog, and if the base query contains a projection to an anonynous type, type inference is a must.

Also, supporting both..

`session.Query<Cat>().Update(x=>new Cat{Name="Missy"});`

and

`session.Query<Cat>().Update(x=>x.Set(y=>y.Name,"Missy"));`

with the same method name is not possible, since the compiler can't distinguish between the two signatures when the lambda parameter is not used, unless the first one is qualified like this:

`session.Query<Cat>().Update((Cat x)=>new Cat{Name="Missy"});`

A minor nuisance maybe, but also something that could unnecessarily become a frequently asked question.

The resone to use a fluent syntax for the Update and not only the Insert, which required it, was simply concistency.

jogibear9988:
Any News to this Pull Req???

gliljas:
No, but any input you might have would be welcome. Especially regarding the syntax. I feel quite confident that the implementation itself is solid.

jogibear9988:
I think the Syntax is quite good.

We're using NHibernate quite long now in a Project. But not supporting Linq for Deletes and Updates, is a real show stopper.

What are the conflicts wich are not yet resolved?

jogibear9988:
Is there any active discussion to merge this?

gliljas:
Sorry for leaving you hanging there. Not much discussion, since this was aimed for 5.0.

hazzik:
Hi @gliljas, can you please enable [Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) feature?

src/NHibernate.Test/NHSpecificTest/NH3438/BaseFixture.cs
```diff
@@ -0,0 +1,46 @@
+using System.Collections;
+using System.Collections.Generic;
+using NHibernate.Hql.Ast.ANTLR;
+using NHibernate.Util;
+
+namespace NHibernate.Test.NHSpecificTest.NH3438
```
fredericDelaporte: Typo, it should be 3488. (Already renamed on my local branch.)

src/NHibernate/Linq/Assignments.cs
```diff
@@ -0,0 +1,153 @@
...
+		/// Sets the specified property.
+		/// </summary>
+		/// <typeparam name="TProp">The type of the property.</typeparam>
+		/// <param name="property">The property.</param>
+		/// <param name="expression">The expression that should be assigned to the property.</param>
+		/// <returns></returns>
+		public Assignments<TInput, TOutput> Set<TProp>(Expression<Func<TOutput, TProp>> property, Expression<Func<TInput, TProp>> expression)
+		{
+			ArgumentUtility.CheckNotNull("property", property);
+			ArgumentUtility.CheckNotNull("expression", expression);
```
fredericDelaporte: This utility seems to be no more available, I have re-coded it "old style".

src/NHibernate/Linq/InsertSyntax.cs
```diff
@@ -0,0 +1,55 @@
...
+			_query = query;
+		}
+
+		/// <summary>
+		/// Executes the insert, using the specified assignments.
+		/// </summary>
+		/// <typeparam name="TOutput">The type of the output.</typeparam>
+		/// <param name="assignmentActions">The assignments.</param>
+		/// <returns></returns>
+		public int Into<TOutput>(Action<Assignments<TInput, TOutput>> assignmentActions)
```
fredericDelaporte: Maybe renaming `TInput` and `TOutput` as `TSource` and `TTarget` would be more explicit.
And comment doc for TTarget could be "The type of the entities to insert."

src/NHibernate/Linq/LinqExtensionMethods.cs
```diff
@@ -98,5 +98,45 @@ public static IFutureValue<T> ToFutureValue<T>(this IQueryable<T> query)
...
 		}
+
+		
+		/// <summary>
+		/// Deletes all entities in the specified query. The delete operation is performed in the database.
+		/// </summary>
+		/// <typeparam name="T"></typeparam>
+		/// <param name="query">The query.</param>
+		/// <returns></returns>
+		public static int Delete<T>(this IQueryable<T> query)
```
fredericDelaporte: Contrary to `Update` and `Insert`, this one causes immediate execution. Maybe should we require an additional `Execute` call for aligning with `Update` and `Insert`.

src/NHibernate/Linq/UpdateSyntax.cs
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+
+		/// <summary>
+		/// Specify the assignments and execute the update.
+		/// </summary>
+		/// <param name="assignments">The assignments.</param>
+		/// <param name="versioned">if set to <c>true</c> [versioned].</param>
+		/// <returns></returns>
+		public int Assign(Action<Assignments<T, T>> assignments, bool versioned = false)
```
fredericDelaporte: Shouldn't it be versioned by default for entity having versioning, if possible?  (I guess there are some hurdles indeed.)
The xml doc should maybe be written : `<c>true</c> for having the update increment the entities version, <c>false</c> otherwise.`

Instead of `Assign`, maybe `Values` would do.

src/NHibernate.Test/NHSpecificTest/NH3438/BaseFixture.cs
fredericDelaporte: Typo, it should be 3488. All files and the older of the test cases for NH3438 have this typo. (Already renamed on my local branch.)

src/NHibernate/Linq/Assignments.cs
fredericDelaporte: This utility seems to be no more available after rebasing, I have re-coded it "old style" on my local branch.

src/NHibernate/Linq/InsertSyntax.cs
fredericDelaporte: Maybe renaming everywhere `TInput` and `TOutput` as `TSource` and `TTarget` would be more explicit. I have also thought about `TSelected` and `TInserted` but well, I am not sure.
And xml doc for `TTarget` could be "The type of the entities to insert."
Returns: "Number of inserted rows/entities"? What about insert into entities dispatched into join tables?

src/NHibernate/Linq/LinqExtensionMethods.cs
fredericDelaporte: Contrary to `Update` and `Insert`, this one causes immediate execution. Maybe should we require an additional `Execute` call for aligning with `Update` and `Insert`.

src/NHibernate/Linq/InsertSyntax.cs
gliljas: Good points. I agree fully.

src/NHibernate/Linq/LinqExtensionMethods.cs
gliljas: Possibly. I think that Delete().Execute() would look a bit strange, with a double verb.

src/NHibernate.Test/NHSpecificTest/NH3438/BaseFixture.cs
gliljas: Oops!

gliljas:
"Allow edits from maintainers" has been enabled.

src/NHibernate/Linq/LinqExtensionMethods.cs
fredericDelaporte: Yes, maybe would it be too "noisy" just for consistency.

fredericDelaporte:
May I force push my rebase?

gliljas:
Yes!

fredericDelaporte:
The tests are in fact the old `BulkManipulationFixture` tests ported to Linq, thus some old pattern. A bunch of those tests are failing for some dialects.

Either we add them to the list of failings tests, or we fix them for being ignored on unsupported dialects.

Some attempts at that were present, such as a test on `Dialect.SupportsSubqueryOnMutatingTable`, but this property appears to be never set to false for any dialect. (In Hibernate, it is, see [`MySQLDialect`](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java#L511) by example.)

I am in the process of modernizing those Linq tests and fixing their conditions.

src/NHibernate.Test/NHSpecificTest/NH3488/LinqBulkManipulationFixture.cs
```diff
@@ -866,6 +910,11 @@ public void UpdateSetNullOnJoinedSubclass()
 		[Test]
 		public void DeleteWithSubquery()
 		{
+			if (Dialect is MsSqlCeDialect)
+			{
+				Assert.Ignore("Test failing on Ms SQL CE.");
```
fredericDelaporte: Fails too in original `BulkManipulationFixture`, with a slightly different SQL query, but same error message. As this DB is deprecated since 2013, I do not intend to investigate more what is going on for this case.

fredericDelaporte:
Under Firebird `InsertToComponent` takes 54 minutes and gets timed-out. 
Trying that locally, the test gets stuck at executing:
```
insert into SimpleClassWithComponent ( id, name_first )
	select gen_id(hibernate_sequence, 1 ), cast(@p0 as VARCHAR(255)) as col_0_0_
		from SimpleClassWithComponent simpleclas0_
```
And the db size endlessly grows... It looks like this kind of query under Firebird causes a recursive insert, as if newly inserted rows where taken into account in the `select` and re-inserted... I am going to ignore that test for Firebird.

Under SqlServerCe, there is a TeamCity build failure I do not understand. The log is huge due to the numerous known failing tests (700+), and at the end we can find:

```
[16:17:21][NHibernate.Test.VisualBasic.dll] NHibernate.Test.VisualBasic.Issues.NH3302.Fixture.LinqStringLikeQuestion
[16:17:21][NHibernate.Test.VisualBasic.Issues.NH3302.Fixture.LinqStringLikeQuestion] Test ignored: NHibernate.Test.VisualBasic.Issues.NH3302.Fixture.LinqStringLikeQuestion, ignore reason: OneTimeSetUp: Not fixed yet.
[16:17:21][call] common.remove-connection-settings-from-app-config
[16:17:21][common.remove-connection-settings-from-app-config] xmlpoke
[16:17:21][xmlpoke] No matching nodes were found with XPath expression '/configuration/nhibernate/add[@key='hibernate.connection.connection_string']/@value'.
[16:17:21][call] common.run-database-tests
[16:17:21][Step 1/1] verify-test-results (3s)
[16:17:21][verify-test-results] if (3s)
[16:17:21][if] copy
[16:17:21][copy] Copying 1 file to 'D:\BuildAgent\work\b0ecfa5dcda2234\build\NHibernate-5.0.0.Alpha1-debug\bin\net-4.0\test-results'.
[16:17:24]
[Step 1/1] D:\BuildAgent\work\b0ecfa5dcda2234\teamcity.build(30,5):
Function call failed.
Expression: ${testResult::CompareResults(teamcity.current.result, teamcity.last.result)}
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[16:17:24][Step 1/1] Process exited with code 1
[16:17:24][Step 1/1] NAnt output
[16:17:24][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.EntityModeToTuplizerPerf.Fixture.VerifyEntityModeNotFound
[16:17:24][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryComplexEqualToComplexThentUseTheCaseConstructorForBoth
[16:17:24][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantEqualToMemberThenDoesNotUseTheCaseConstructor
[16:17:24][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantEqualToNullableMemberThenUseTheCaseConstructorForMember
[16:17:24][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantEqualToNullableMemberValueThenDoesNotUseTheCaseConstructorForMember
[16:17:24][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantNotEqualToMemberThenDoesNotUseTheCaseConstructor
[16:17:24][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantNotEqualToNullableMemberThenUseTheCaseConstructorForMember
[16:17:24][NAnt output]     
[16:17:24][NAnt output]     *** Tests ignored since last recorded results ***
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteOnJoinedSubclass
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteOnMappedJoin
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteRestrictedOnManyToOne
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteUnionSubclassAbstractRoot
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteUnionSubclassConcreteSubclass
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.SimpleDeleteOnAnimal
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateMultiplePropertyOnAnimal
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateOnAnimal
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateOnComponent
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateOnMammal
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateOnManyToOne
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateSetNullOnJoinedSubclass
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateSetNullUnionSubclass
[16:17:24][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateWithWhereExistsSubquery
[16:17:24][NAnt output]     
[16:17:24][NAnt output]     *** Tests broken since last recorded results ***
[16:17:24][NAnt output]     None
[16:17:24][NAnt output]     
[16:17:24][NAnt output] 
[16:17:24][NAnt output] Total time: 453.4 seconds.
[16:17:24][NAnt output] 
[16:17:35][Step 1/1] Process exited with code 1
[16:17:35][Step 1/1] Step NAnt failed
```

fredericDelaporte:
I have changed some xml comments according to earlier discussion, with some minor code cleanup. And made a third try at fixing the tests for accounting dialect unsupported features.

fredericDelaporte:
SqlServerCe TeamCity build still failing, but looks like a test tooling failure.

```
[19:14:09][Step 1/1] verify-test-results (2s)
[19:14:09][verify-test-results] if (2s)
[19:14:10][if] copy
[19:14:10][copy] Copying 1 file to 'D:\BuildAgent\work\b0ecfa5dcda2234\build\NHibernate-5.0.0.Alpha1-debug\bin\net-4.0\test-results'.
[19:14:12]
[Step 1/1] D:\BuildAgent\work\b0ecfa5dcda2234\teamcity.build(30,5):
Function call failed.
Expression: ${testResult::CompareResults(teamcity.current.result, teamcity.last.result)}
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[19:14:12][Step 1/1] Process exited with code 1
```

hazzik:
No, it is not. It means that the comparison produced results which has
failing tests

hazzik:
No, it is not. It means that the comparison produced the result file which
has new failing tests.

fredericDelaporte:
But following lines seem to state otherwise:

```
[19:14:12][Step 1/1] NAnt output
[19:14:12][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.EntityModeToTuplizerPerf.Fixture.VerifyEntityModeNotFound
[19:14:12][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryComplexEqualToComplexThentUseTheCaseConstructorForBoth
[19:14:12][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantEqualToMemberThenDoesNotUseTheCaseConstructor
[19:14:12][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantEqualToNullableMemberThenUseTheCaseConstructorForMember
[19:14:12][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantEqualToNullableMemberValueThenDoesNotUseTheCaseConstructorForMember
[19:14:12][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantNotEqualToMemberThenDoesNotUseTheCaseConstructor
[19:14:12][NAnt output]     PASS    - NHibernate.Test.NHSpecificTest.NH2505.Fixture.WhenQueryConstantNotEqualToNullableMemberThenUseTheCaseConstructorForMember
[19:14:12][NAnt output]     
[19:14:12][NAnt output]     *** Tests ignored since last recorded results ***
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteOnJoinedSubclass
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteOnMappedJoin
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteRestrictedOnManyToOne
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteUnionSubclassAbstractRoot
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.DeleteUnionSubclassConcreteSubclass
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.SimpleDeleteOnAnimal
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateMultiplePropertyOnAnimal
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateOnAnimal
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateOnComponent
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateOnMammal
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateOnManyToOne
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateSetNullOnJoinedSubclass
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateSetNullUnionSubclass
[19:14:12][NAnt output]     IGNORED - NHibernate.Test.Hql.Ast.BulkManipulation.UpdateWithWhereExistsSubquery
[19:14:12][NAnt output]     
[19:14:12][NAnt output]     *** Tests broken since last recorded results ***
[19:14:12][NAnt output]     None
[19:14:12][NAnt output]     
[19:14:12][NAnt output] 
[19:14:12][NAnt output] Total time: 422.5 seconds.
[19:14:12][NAnt output] 
[19:14:23][Step 1/1] Process exited with code 1
[19:14:24][Step 1/1] Step NAnt failed
```
https://teamcity.jetbrains.com/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=1030470&sizeLimit=0#_focus=70102

fredericDelaporte:
Otherwise we have to find among more than seven hundred usually failing tests which ones are "new". The TeamCity interface seems to no more showcase them after a first build. And I have checked the new test cases of this pull request, they are no more in error.

Ok I forgot one of the tests added to the old `BulkManipulationFixture`. Still very inconvenient to have so many flaky tests.

hazzik:
Check the Comparison.txt file in build artifacts

hazzik:
```
FAIL    - NHibernate.Test.Hql.Ast.BulkManipulation.SimpleInsertFromAggregate
```

hazzik:
>Otherwise we have to find among more than seven hundred usually failing tests which ones are "new". 

The [comparision.txt](https://teamcity.jetbrains.com/repository/download/bt1175/1030470:id/NHibernate-5.0.0.Alpha1-debug/bin/net-4.0/test-results/Comparison.txt) here to do that for us.

fredericDelaporte:
Ok, I was not knowing we had this file for finding out which test causes trouble.

fredericDelaporte:
I have updated the Linq reference documentation for documenting this new feature.
And fourth try at conditionally ignoring tests incompatible with some dialects.

fredericDelaporte:
I have squashed together some of my commits and reworded them for helping in reviewing my changes.

gliljas:
Massive thanks, Frederic!

gliljas:
You mentioned the  `Future`  thing. Perhaps it's something to consider now. How does it work now? 

session.Query<Cat>().ToFuture().Insert().As(x=>new Dog{Name=x.Name});

Maybe that works...

But, even if it does, there no natural point of execution for the hidden multiquery. That could be a separate feature to add, but we better think about it now, so that we don't have to modify the API too much.

fredericDelaporte:
As was thinking as something like:

    session.Query<Cat>().Insert().AsFuture(x=>new Dog{Name=x.Name});
    session.Query<Cat>().DeleteFuture();

Currently future is called instead of the executing call, it looks to me more natural to keep it that way. In such case it could be added later, it should not change anything to non future calls.

I do not think future to be much needed for bulk inserts/updates/deletes. It would be just a nice to have.

fredericDelaporte:
Added a missing null check in `QueryModelVisitor.VisitUpdateClause`, discovered by testing what it gives disabling the `IIf` in `Assignments.ConvertToDictionaryExpression`.

fredericDelaporte:
I wonder if I should add a warning about `Insert` in the reference documentation: what about entities having an id generator running on application side? In most cases it would fail. I guess some cases may be tricked by assigning the id with some expression (probably involving a custom function) which would calculate a sound one in database, but well, that would be quite hackish.

fredericDelaporte:
I have added some informations and links for these operations limitations, already documented in the [“DML-style operations”](http://nhibernate.info/doc/nhibernate-reference/batch.html#batch-direct) section.

fredericDelaporte:
Is anyone wanting to additionally review it? Or may I proceed and squash&merge?

hazzik:
Yes, I want to review, I'll probably do it later on the weekends.

src/NHibernate.Test/NHSpecificTest/NH3488/LinqBulkManipulationFixture.cs
```diff
@@ -0,0 +1,1272 @@
...
+				s.Query<EntityWithCrazyCompositeKey>().Where(x => x.Id.Id == 1 && x.Id.OtherId == 2).Delete();
+
+				t.Commit();
+				s.Close();
+			}
+		}
+
+		#endregion
+
+		private class TestData
```
hazzik: Can this be implemented as a "normal" test setup / teardown?

fredericDelaporte: Sure. That is just the original `BulkManipulation` pattern not completely rewritten. There are a bunch of other tests in this Linq version which still do their own setup/teardown and may get put in normal pattern too.

fredericDelaporte:
Tests moved out of NHSpecificTest, being tests of a whole feature, not only a bug or improvement test. (I have tried putting them in Linq or BulkManipulation existing folders, but that was conflicting with the tests already there and defining other classes with same names.)

Tests modernized further for putting all their setup and tear-down inside corresponding methods.

fredericDelaporte:
Rebased for resolving conflicts, most commits squashed.

fredericDelaporte:
The new SQL Server CE failures are due to [this change](https://github.com/nhibernate/nhibernate-core/pull/615/files#diff-b7ae926346dceb1cb85564e3f5f1d3c1R128). It causes bulk inserts without id specified and not guarding against incompatible id generator to fail. I have added required checks.

hazzik:
The pat I do not like here, is "two-phase" inserts. They look weird.

Probably we need to require user to always do a `Select` the desired type first.

So,

```
var count = s.Query<Car>().Insert().As(x => new Pickup { Id = x.Id, Vin = x.Vin, Owner = x.Owner });
```
would become
```
var count = s.Query<Car>().Select(x => new Pickup { Id = x.Id, Vin = x.Vin, Owner = x.Owner }).Insert();
```
```
var count = s.Query<Car>()
	.GroupBy(x => x.Id)
	.Select(x => new { Id = x.Key, Vin = x.Max(y => y.Vin), Owner = x.Max(y => y.Owner) })
	.Insert().As(x => new Pickup { Id = x.Id, Vin = x.Vin, Owner = x.Owner });
```
Would become:
```
var count = s.Query<Car>()
	.GroupBy(x => x.Id)
	.Select(x => new Pickup { Id = x.Key, Vin = x.Max(y => y.Vin), Owner = x.Max(y => y.Owner) })
	.Insert();
```

Also, .Insert() at the end looks unnatural. Maybe something like this:

```
var count = s.Insert(
	s.Query<Car>()
		.GroupBy(x => x.Id)
		.Select(x => new Pickup { Id = x.Key, Vin = x.Max(y => y.Vin), Owner = x.Max(y => y.Owner) })
	);
```
This potentially can be a lambda:
```
var count = s.Insert((IQueryable<Cat> cats) => 
		cats.GroupBy(x => x.Id)
		.Select(x => new Pickup { Id = x.Key, Vin = x.Max(y => y.Vin), Owner = x.Max(y => y.Owner) })
	);
```

Or, maybe replace `Select` by `Insert` alltogether.

hazzik:
What will happen if such query is executed:
```
s.Query<Animal>().Where(x => x.Mother == _butterfly).Select(x => new Car {Id = x.Id}).Delete()
```

hazzik:
White space issues.

hazzik:
Tested following syntax, and it works perfectly:

```
var count = s.Query<Car>()
	.GroupBy(x => x.Id)
	.Insert(x => new Pickup { Id = x.Key, Vin = x.Max(y => y.Vin), Owner = x.Max(y => y.Owner) });
```

I think we shall adopt it, and remove As/Assign altogether.

gliljas:
I'm not fully convinced by the terser syntax, for a few reasons.

`cats.Insert().Into<Dog>`

..actually reads better as code with an intention.

`cats.Insert(x=>new Dog`

is so terse that it almost looks a dog is inserted into the cats list.

'cats.InsertInto<Dog>(x=>new Dog...)`

would be preferrable, but it's not possible.

As to the Set syntax, I agree that it's bit clunky, but there's some value in be able to programmatically add the columns that should be included in the insert/update, which is not possible with an anonymous object. It also allows statically typed insert into columns exposed as readonly properties.
A dictionary could be used, but it's less typed and personally I think it's uglier.

The Update is easier. Both Update(x=>new Dog{...}) and Update(x=>x.Set(... can be implemented, without overload collision.

An option could be to have both query.Insert(x=>.....) and query.Insert().Into(x=>....), but as I mentioned I'm not completely sold on the terse version.

I found this:
http://docs.telerik.com/data-access/feature-reference/api/context-api/bulk-operations/featureref-contextapi-bulk-ud-operations-update

gliljas:
And it's the Insert which makes things tricky, since it requires both TSource and TTarget, in some way. Doing the Update in a similar way is just to keep things somewhat consistent.

fredericDelaporte:
Good points Gunnar. Reconsidering both, I would rather stay with your syntax. Some refactoring done in Alexander branch could be done here too, without changing the syntax.

About:

> What will happen if such query is executed?
> ```C#
> s.Query<Animal>().Where(x => x.Mother == _butterfly).Select(x => new Car {Id = x.Id}).Delete()
> ```

It sound likely to me it will fail to parse once translated to HQL. It could be added in the tests. And it should fail in my opinion.

fredericDelaporte:
Rebased, Alexander's refactoring cherry picked, test case for the invalid delete added and fixed. (It was not failing, it was ignoring the select completely, deleting the animal.)

hazzik:
I've been able to find 3 libraries offering bulk operations: [Telerik OpenAccess](http://docs.telerik.com/data-access/data-access-introduction) (full featured ORM, discontinued now),  [Entity Framework Extensions](http://entityframework-extensions.net/) and [EntityFramework Plus](http://entityframework-plus.net/) (from the same vendor as Entity Framework Extensions).

1. None of them offer insert from Linq query, so I believe we should not offer it as well.
2. Currently our syntax is more complicated and different to all of the mentioned implementations. I think we should harmonize the syntax with some of them.

gliljas:
@hazzik 

1. Why not? It's quite a powerful function, when it's possible to use instead of retrieve many + insert many.
2. I'm all for simplification and inspiration from others.

fredericDelaporte:
> None of them offer insert from Linq query, so I believe we should not offer it as well.

I have a hard time understanding the argument. NHibernate does many things others .Net ORM do not do, and I think this is what cause it to be still used. Otherwise everyone would have switched to the Microsoft backed ORM instead.

The Microsoft community choice tends to always go first for Microsoft solutions, and considers others only if there are too much features lacking for them in Microsoft solutions. Choosing to align on other solutions "because everyone do that and do not support that" is like signing NHibernate death for me.

I do not know why the Telerik OpenAccess has been discontinued, but if it was not offering much more compared to Entity Framework, that is no surprise.

Moreover, NHibernate already supports bulk insert through HQL. This PR just aims at exposing this feature with Linq. NHibernate users are likely to have a hard time understanding why we would have done that only for update and delete.

This PR current syntax still looks very reasonable to me. Others maybe propose a slightly simpler one, but with less features, especially lacking the insert.

I do not really see what is the point about aligning with other syntaxes when it forces to removes features, while by the same time there is no actual common interface, no actual standard. Those interfaces are all custom extensions over Linq. 

If Linq was incorporating a native interface for such feature, of course it would be better to strive for using it instead of adding our own. But that is not the case.

hazzik:
R.I.P.

jogibear9988:
Linq2Db also supports insert & update & delete via linq

hazzik:
thanks @jogibear9988 

https://github.com/linq2db/linq2db

```
using (var db = new DbNorthwind())
{
  db.Product
    .Value(p => p.Name, product.Name)
    .Value(p => p.UnitPrice, 10.2m)
    .Value(p => p.Added, () => Sql.CurrentTimestamp)
    .Insert();
}

```

fredericDelaporte:
So it does not really look related to what is DML in NHibernate. Here that is a `insert ... values ...` with all values provided from the client, whereas the DML insert in NHibernate is inserting from data selected in the database without a roundtrip to the client.

gliljas:
It's not quite the same thing, but it does hint of another way to go. 

`
session.Query<Product>()
    .Where(p => p.UnitsInStock == 0)
    .Set(p => p.Discontinued, true)
    .Update();
`

or 

`
session.Update<Product>()
    .Where(p => p.UnitsInStock == 0)  //Just a predicate, not a standard Where
    .Set(p => p.Discontinued, true)
    .Execute();
`

and

session.InsertInto<Product>()
    .From<Car>(p => p.UnitsInStock == 0)  // or From<TSource>(IQueryable<TSource> queryable)
    .Value(p => p.Discontinued, c=>c.Discontinued)
    .Execute();
`

fredericDelaporte:
In order to compare things:

# Insert

## Current syntax

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Insert()
    .As(c => new Pickup { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Insert()
    .Into<Pickup>(
        a => a // assignments object
            .Set(p => p.Id, c => c.Id)
            .Set(p => p.Vin, c => c.Vin)
            .Set(p => p.Owner, "The owner"));
```

## linq2db alike

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .InsertInto(c => new Pickup { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    // Maybe add here an .InsertValues() or .Insert()
    .Value(y => y.Id, y => y.Id)
    .Value(y => y.Vin, y => y.Vin)
    .Value(y => y.Owner, "The owner"))
    .Insert(); // Replace by .Execute() if we have added an insert call above.
```

# Update

## Current syntax

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update() // .UpdateVersioned() available too
    // Have to check exact type here, should throw if putting new Pickup.
    .As(c => new Car { Vin = c.Vin + "b", Owner = c.Owner }); 
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update() // .UpdateVersioned() available too
    .Assign<Pickup>(
        a => a // assignments object
            .Set(p => p.Vin, c => p.Vin + "b")
            .Set(p => p.Owner, "The owner"));
```

## linq2db alike

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    // Have to check exact type here, should throw if putting new Pickup.
    .Update(c => new Car { Vin = c.Vin + "b", Owner = c.Owner }); // Should have UpdateVersioned too
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    // Maybe add here an .Update() or UpdateVersioned()
    .Set(y => y.Vin, y => y.Vin + "b")
    .Set(y => y.Owner, "The owner"))
    .Update(); // Or UpdateVersioned(); replace by .Execute() if we have added an update call above.
```

# Delete

Same with both.

---

I have not taken Gunnar two last syntaxes because I do not see a benefit in having to code our own `Where`.

The linq2db inspired syntax looks nicer to me, thanks @jogibear9988 for having pointed it.

About whether we should choose:

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Set(y => y.Vin, y => y.Vin + "b")
    .Set(y => y.Owner, "The owner"))
    .Update();
```
or 

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update()
    .Set(y => y.Vin, y => y.Vin + "b")
    .Set(y => y.Owner, "The owner"))
    .Execute();
```

Well, both have advantages and drawbacks. The former is more consistent with the "by selector" syntax, but having `Set` or `Value` methods directly on queryables looks a bit weird. The later avoids the weirdness, but is less consistent with the "by selector" syntax. I am more for the later, but then maybe should the insert/update by selector be renamed `ExecuteInsertInto`/`ExecuteUpdate` for smoothing the inconsistency.

hazzik:
@fredericDelaporte can you also add EF+ and Telerik OpenAccess syntax to the comparison?

hazzik:
I think we need to drop complicated Insert.As and Update.As syntax in favor of simpler InsertInto and Update. I don't mind keeping the "Set" syntax as is.

fredericDelaporte:
I still disagree. This syntax is a must for conditionally building updates or inserts, omitting/adding some more columns based on conditions. Doing that with only the selector syntax would force to redefine all the other assignments for each case. And in some dynamic query building cases, it would be unusable.

hazzik:
I dont understand what you disagree with. I'm saying about simplifying 2 methods calls `Update().As(...)` into one call `.Update(...)`. 

This syntax:

```
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update() // .UpdateVersioned() available too
    .Assign<Pickup>(
        a => a // assignments object
            .Set(p => p.Vin, c => p.Vin + "b")
            .Set(p => p.Owner, "The owner"));

```

Stays the same.

Same for insert.

hazzik:
Also, I don't see how it is possible to build anything "dynamically" with the current syntax. Can you provide an example?

hazzik:
Also, IMO, linq2db syntax uses the grammar which is closely related to the actual SQL syntax as it does not have unnatural words as "execute", "assign" and "as"

fredericDelaporte:
Ok, I have misunderstood your comment. I had thought you were willing only the "by selector" syntax.

So this comment was about at least replacing the current "by selector" syntax by the linq2db alike syntax. Agreed with that.

Having the set syntax replaced by something more natural still would be better IMO. Is your last comment meaning you are more for the first linq2db alike "individual setter" proposal than for the second, which still have an `Execute`? Why not. After all, user may be surprised when seing at first  `Value` and `Set` method on queryables, but once having checked what they do, that should not be hard for them to remember what is their purpose.

fredericDelaporte:
> can you also add EF+ and Telerik OpenAccess syntax to the comparison?

EF+ seems to be the same than all others for delete, and the same than linq2db alike "by selector" for update, not supporting inserts, not supporting individual setters. So nothing more to see with it.

Telerik DataAccess seems to be the same than all others for delete, and the same than this PR current "by individual setters" for update but collapsing `.Update().Assign` into `.UpdateAll`, not supporting inserts, not supporting "by selector". So nothing more to see with it either. (Excepted for #601: it uses a `WithOption` method for setting options on queryable, instead of having as many extensions as options.)

fredericDelaporte:
The "update on subclass" case was failing with a hard to understand error (null ref exception in some very remote internal method of NHibernate). I have added an explicit safeguard.

hazzik:
> Is your last comment meaning you are more for the first linq2db alike "individual setter" proposal than for the second, which still have an Execute? 

No, just about taking linq2db approach with `.Update()` at the end.

> After all, user may be surprised when seing at first Value and Set method on queryables, but once having checked what they do, that should not be hard for them to remember what is their purpose.

I think we can hide DML operation into a specific namespace (`NHibernate.Linq.Dml` or NHibernate.Linq.Update`/`...Insert`/`...Delete`).

fredericDelaporte:
I have added "by anonymous selector". It works, but the way it is implemented does not suits me: it relies on implementation details of the compiler. Maybe we should forget about them.

More specifically, with .Net Framework, anonymous objects creation are compiled to a constructor call with arguments and without member initialization. So I detect that constructor call, and parse its arguments. Since there are no guarantee of matching ordering between anonymous type properties and constructor arguments, the code has to rely on constructor argument names to be the same than the property they will initialize. That is currently the case...

hazzik:
>I detect that constructor call, and parse its arguments. Since there are no guarantee of matching ordering between anonymous type properties and constructor arguments, the code has to rely on constructor argument names to be the same than the property they will initialize. 

The framework actually provides the binding information.

fredericDelaporte:
For normal class initialization, yes, but for anonymous type, I have not seen anything providing that.

hazzik:
Especially for anonymous classes. The framework havely relies on anonymous classes when you write LINQ expression (in query form). I'll take a look in a couple of hours.

hazzik:
There is "Members" property.

>Gets the members that can retrieve the values of the fields that were initialized with constructor arguments.

For a normal class this would be empty.

src/NHibernate/Linq/DmlExpressionRewriter.cs
```diff
@@ -38,6 +38,30 @@ void AddSettersFromBindings(IEnumerable<MemberBinding> bindings, string path)
 				}
 		}
 
+		void AddSettersFromConstructorArguments(NewExpression newExpression, string path)
+		{
+			var argumentDefinitions = newExpression.Constructor.GetParameters();
```
fredericDelaporte: Thanks for the pointer @hazzik ,
So instead it should be `newExpression.Members` here. At least it is documented as matching constructor arguments, but that logic still looks quite brittle.

src/NHibernate/Linq/Assignments.cs
```diff
@@ -0,0 +1,60 @@
...
+		internal IReadOnlyDictionary<string, Expression> List => _assignments;
+
+		/// <summary>
+		/// Sets the specified property.
+		/// </summary>
+		/// <typeparam name="TProp">The type of the property.</typeparam>
+		/// <param name="property">The property.</param>
+		/// <param name="expression">The expression that should be assigned to the property.</param>
+		/// <returns>The current assignments list.</returns>
+		public Assignments<TSource, TTarget> Set<TProp>(Expression<Func<TTarget, TProp>> property, Expression<Func<TSource, TProp>> expression)
```
hazzik: We would need an additional method which accepts `Expression<Func<TProp>> expression` to be able to write following:

```
.Set(x=>x.CreatedAt, () => SqlFunctions.GetDate() /* DateTime.Now in future */)
```

fredericDelaporte: That would be for avoiding using a lambda if an unused parameter? Like:

    .Set(x=>x.CreatedAt, x => SqlFunctions.GetDate() /* DateTime.Now in future */)

hazzik: Yes

src/NHibernate/Linq/DmlExpressionRewriter.cs
```diff
@@ -38,6 +38,35 @@ void AddSettersFromBindings(IEnumerable<MemberBinding> bindings, string path)
 				}
 		}
 
+		void AddSettersFromConstructorArguments(NewExpression newExpression, string path)
+		{
+			// See Members documentation, this property is specifically designed to match constructor arguments values
+			// in the anonymous object case. It can be null otherwise, or non-matching.
+			var argumentMatchingMembers = newExpression.Members;
+			if (argumentMatchingMembers == null || argumentMatchingMembers.Count != newExpression.Arguments.Count)
```
fredericDelaporte: There are constructor paths setting it to null (seen with IlSpy). About non-matching, I have not seen an explicit case, but nothing seems to be there for preventing building manually a `NewExpression` with a non-matching members list. So I have put at least a check for not failing with an index out of range.

src/NHibernate/Linq/UpdateSyntax.cs
```diff
@@ -0,0 +1,65 @@
...
+			return ExecuteUpdate(DmlExpressionRewriter.PrepareExpression<T>(_sourceExpression, assignments.List));
+		}
+
+		/// <summary>
+		/// Specify the assignments and execute the update.
+		/// </summary>
+		/// <param name="expression">The assignments expressed as a member initialization, e.g.
+		/// <c>x => new Dog { Name = x.Name, Age = x.Age + 5 }</c>. Unset members are ignored and left untouched.</param>
+		/// <returns>The number of updated entities.</returns>
+		public int As(Expression<Func<T, T>> expression)
```
fredericDelaporte: From an API standpoint, it better documents that member initialization has to be done on the same type, since the other one tell it is only for anonymous object.

From the implementation standpoint, it avoids reflecting the actual resulting type and rebuilding the lambda for being able to call `PrepareExtension<TSource, TTarget>` when the resulting object is indeed a `TSource`.

hazzik: I know. But it still feels not needed.

Also, the example `session.Query<Car>().Update().As(c=>new Pickup(){})` resolves the strongly types version instead of object version becaus pickup is an child class of a Car.

fredericDelaporte: Yes, as anticipated, and handled by this [check](https://github.com/nhibernate/nhibernate-core/pull/391/commits/63d659e6178420e132e505d4cc1689109939561e#diff-db3b5a31ee1bd41be64a8166f7d27707R160), added in ["Protect against invalid use cases" commit](https://github.com/nhibernate/nhibernate-core/pull/391/commits/6aa9c1807716958e59739b612e7feb38db188f08).

This invalid usage case correctly errors out, telling this strongly typed method expect the exact same type.

If doing `session.Query<Car>().Update().As(c=>new Animal(){})`, it will go through the anonymous object version and there the anonymous object method will complains it accepts only anonymous objects.

If we merge both, it will complicate the checking and the error message. Where we do the checking, we do not know if it is called for an update or an insert, so we may end-up with some cumbersome message trying to cover all cases. (Strongly typed update? Accept only exact same type, but need to not imply that if raising the error for an insert. Anonymous insert? Need an explicitly specified target type, while update do not care.)
Otherwise we would have to move up the checking, meaning starting to analyze the expression in `InsertSyntax` and `UpdateSyntax`, instead of keeping it together in `DmlExpressionRewriter`. (Or tell to `DmlExpressionRewriter` if it is called for update or insert in order to adapt the error message.)

I personally do not feel that having both methods is unneeded. It eases implementation and understanding of this API. The later alone seems enough to me for having both methods.

fredericDelaporte:
> I think we can hide DML operation into a specific namespace (`NHibernate.Linq.Dml` or `NHibernate.Linq.Update/...Insert/...Delete`).

Great for whose who do not wish to see that everywhere they use queryables and Nhibernate.Linq, but quite bad for discover-ability.

`NHibernate.Linq` IMO already illustrates the pain of poor discover-ability due to being hidden in a namespace, causing many people to not easily find how to perform a Linq query with NHibernate. (And with some trying to use queryover as a link provider instead, as frequently seen on Stack Overflow.) We all knows most developers first try to use the software, and eventually go seeking answers only if failing using it. Discover-ability is a very important aspect IMO, we should no more add features in a poorly discover-able way I think.

src/NHibernate/Linq/Assignments.cs
fredericDelaporte: Am I wrong or `SqlFunctions` are currently completely unsupported? They do not fit well NHibernate anyway, as they seems specifically meant for SQL Server. NHibernate does not promote using database specific feature usually, it rather supplies database agnostic equivalent.

At this subject, I have put in my todo list since long: NH- : a static class like `HqlFunctions` supplying a way to call with Linq built-in HQL function supported by dialects.
At least having a `HqlFunctions.Call(string name, params object[] args)` would be handy. But there are many other things in my list.

Overall, is there currently a supported case where the "no args" lambda would be useful? If none, I would rather not add a "not yet useful method".

fredericDelaporte:
The linq2db alike insert syntax with individual setter was not working, I have updated my [comparison comment](https://github.com/nhibernate/nhibernate-core/pull/391#issuecomment-316046550).

We are back to initial issue with inserts indeed.

fredericDelaporte:
I have pushed an alternate syntax proposal. Commit to be dropped if preferring the previous one.

# Insert

## Old syntax

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Insert()
    .As(c => new Pickup { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By single anonymous selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Insert()
    .As<Pickup>(c => new { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Insert()
    .Into<Pickup>(
        a => a // assignments object
            .Set(p => p.Id, c => c.Id)
            .Set(p => p.Vin, c => c.Vin)
            .Set(p => p.Owner, "The owner"));
```

## linq2db alike

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .InsertInto(c => new Pickup { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By single anonymous selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    // Sadly, have to re-specify the source type.
    .InsertInto<Car, Pickup>(c => new { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    // Value are not yielding a queryable anyway, so better be explicit
    // with the fact we are toggling to a builder mode. Allows not re-specifying source type.
    .InsertBuilder().Into<Pickup>()
    .Value(y => y.Id, y => y.Id)
    .Value(y => y.Vin, y => y.Vin)
    .Value(y => y.Owner, "The owner"))
    .Insert();
```

# Update

## Current syntax

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update() // .UpdateVersioned() available too
    // Have to check exact type here, should throw if putting new Pickup.
    .As(c => new Car { Vin = c.Vin + "b", Owner = c.Owner }); 
```

### By single anonymous selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update() // .UpdateVersioned() available too
    // Have to check exact type here, should throw if putting new Pickup.
    .As<Car>(c => new { Vin = c.Vin + "b", Owner = c.Owner }); 
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update() // .UpdateVersioned() available too
    .Assign<Pickup>(
        a => a // assignments object
            .Set(p => p.Vin, c => p.Vin + "b")
            .Set(p => p.Owner, "The owner"));
```

## linq2db alike

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update(c => new Car { Vin = c.Vin + "b", Owner = c.Owner }); // Or UpdateVersioned
```

### By single anonymous selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update(c => new { Vin = c.Vin + "b", Owner = c.Owner }); // Or UpdateVersioned
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    // For consistency with insert case, and for telling explicitly we are switching away from
    // queryable to a builder mode.
    .UpdateBuilder()
    .Set(y => y.Vin, y => y.Vin + "b")
    .Set(y => y.Owner, "The owner"))
    .Update();
```

# Delete

Same with both.

fredericDelaporte:
I have pushed an alternate syntax proposal. Commit to be dropped if preferring the previous one.

# Insert

## Old syntax

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Insert()
    .As(c => new Pickup { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By single anonymous selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Insert()
    .As<Pickup>(c => new { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Insert()
    .Into<Pickup>(
        a => a // assignments object
            .Set(p => p.Id, c => c.Id)
            .Set(p => p.Vin, c => c.Vin)
            .Set(p => p.Owner, "The owner"));
```

## New linq2db alike

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .InsertInto(c => new Pickup { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By single anonymous selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    // Sadly, have to re-specify the source type.
    .InsertInto<Car, Pickup>(c => new { Id = c.Id, Vin = c.Vin, Owner = c.Owner });
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    // Value are not yielding a queryable anyway, so better be explicit
    // with the fact we are toggling to a builder mode. Allows not re-specifying source type.
    .InsertBuilder().Into<Pickup>()
    .Value(y => y.Id, y => y.Id)
    .Value(y => y.Vin, y => y.Vin)
    .Value(y => y.Owner, "The owner"))
    .Insert();
```

# Update

## Old syntax

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update() // .UpdateVersioned() available too
    .As(c => new Car { Vin = c.Vin + "b", Owner = c.Owner }); 
```

### By single anonymous selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update() // .UpdateVersioned() available too
    .As<Car>(c => new { Vin = c.Vin + "b", Owner = c.Owner }); 
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update() // .UpdateVersioned() available too
    .Assign<Pickup>(
        a => a // assignments object
            .Set(p => p.Vin, c => p.Vin + "b")
            .Set(p => p.Owner, "The owner"));
```

## New linq2db alike

### By single selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update(c => new Car { Vin = c.Vin + "b", Owner = c.Owner }); // Or UpdateVersioned
```

### By single anonymous selector

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    .Update(c => new { Vin = c.Vin + "b", Owner = c.Owner }); // Or UpdateVersioned
```

### By individual setters

```c#
session
    .Query<Car>()
    .Where(c => c.UnitsInStock == 0)
    // For consistency with insert case, and for telling explicitly we are switching away from
    // queryable to a builder mode.
    .UpdateBuilder()
    .Set(y => y.Vin, y => y.Vin + "b")
    .Set(y => y.Owner, "The owner"))
    .Update(); // Or UpdateVersioned
```

# Delete

Same with both.

jogibear9988:
I think in linq2db you could use some set-functions, and after that you could add an additional where before you update (i've not tested, but i think it will work).
will thos work with your syntax also?

fredericDelaporte:
No, we no more have a queryable from the point we start using one of these new methods. And that looks to be the same with [linq2b](https://github.com/linq2db/linq2db): calling [`Set`](https://github.com/linq2db/linq2db/blob/master/Source/LinqExtensions.cs#L316) yields an [`IUpdatable`](https://github.com/linq2db/linq2db/blob/master/Source/Linq/IUpdatable.cs) which is not a queryable either, and seems to only allow specifying additional [`Set`](https://github.com/linq2db/linq2db/blob/master/Source/LinqExtensions.cs#L338) and call a final [`Update`](https://github.com/linq2db/linq2db/blob/master/Source/LinqExtensions.cs#L260).

Anyway, maybe this would look more natural from a SQL standpoint to be able to put the `Where` after, but well, Linq is not SQL and users are already used to this kind of clauses reversal, like putting `Select` last.

fredericDelaporte:
Rebased, bulk manipulation (non linq) refactoring dropped.

asd-and-Rizzo:
This stuff will make NH live in enterprise much longer. Thanks!

PR #392, opened 2015-01-03
NH-3747 - Unit tests and fixes for aggregate predicates

gliljas:


src/NHibernate/Linq/ExpressionExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+using Remotion.Linq.Clauses.ResultOperators;
+
+namespace NHibernate.Linq
+{
+	public static class ExpressionExtensions
+	{
+		public static bool IsGroupingKey(this MemberExpression expression)
+		{
+			return expression.Member.Name == "Key" && expression.Member.DeclaringType!=null &&
+			       expression.Member.DeclaringType.GetGenericTypeDefinition() == typeof (IGrouping<,>);
```
hazzik: Need to check that DeclaringType is generic type

gliljas: True! Will fix

PR #394, opened 2015-01-26
NH-3750: Use Remotion.Linq from NuGet instead of ILMerging it.

patearl:


hazzik:
How will it work with relinq 1.15?

hazzik:
I think the best solution will be continue to ilmerge ReLinq, but rewrite the namespace

PR #396, opened 2015-02-10
uuid.hex id generator fix in mapping by code. Parameters added and retur...

lnu:
We use uuid.hex in the NHibernate.AspNet.Identity project. Right now a custom mapper is implemented in the project. It could be avoided with this fix.

### DIFF
--- a/src/NHibernate/Mapping/ByCode/Generators.cs
+++ b/src/NHibernate/Mapping/ByCode/Generators.cs
@@ -91,6 +91,28 @@ namespace NHibernate.Mapping.ByCode
 	{
 		#region Implementation of IGeneratorDef
 
+		private readonly object param;
+
+		public UUIDHexGeneratorDef()
+		{
+		}
+
+		public UUIDHexGeneratorDef(string format)
+		{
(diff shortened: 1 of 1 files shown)


src/NHibernate/Mapping/ByCode/Generators.cs
```diff
@@ -98,12 +120,12 @@ public string Class
...
 		public object Params
 		{
-			get { return null; }
+			get { return param; }
 		}
 
 		public System.Type DefaultReturnType
 		{
-			get { return typeof(Guid); }
+			get { return typeof(string); }
```
hazzik: Why this?

lnu: the uuid.hex generator returns a string not a guid. Am I missing something?

hazzik:
@lnu do we have JIRA for this? If not, could you please create one?

lnu:
https://nhibernate.jira.com/browse/NH-3759

hazzik:
Thanks

PR #400, opened 2015-02-19
Entity loaders lazy initialization

git-heavy:
Using NH in our own project we found out that Batching Entity Loaders need a lot of time to initalize. It was critical for us, so we decided to make this process a bit lazy.


DmitryNaumov:
What if just replace Loader[] with Lazy<Loader>[]? Wouldn't it be simpler change? BTW I wonder why creating Loader takes lot of time...

hazzik:
I've checked and it seems the solution with `Lazy<>` works best.

PR #404, opened 2015-02-25
NH-3480 - Use ANSI "join" for UPDATE statement

hazzik:
@JKeyser could you please review?

### DIFF
--- a/src/NHibernate/Persister/Collection/OneToManyPersister.cs
+++ b/src/NHibernate/Persister/Collection/OneToManyPersister.cs
@@ -75,8 +75,7 @@ namespace NHibernate.Persister.Collection
 			else
 			{
 				var ownerPersister = (IOuterJoinLoadable)OwnerEntityPersister;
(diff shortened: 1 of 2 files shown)


JKeyser:
I'll try to look at this later today.

JKeyser:
My only minor comment on this is that in many circles where I've seen the EXISTS function used this way, the SELECT statement reads "SELECT 1" (or some other constant) instead of "SELECT *".  The reasoning is that no data need to be pulled from the table since the only concern is testing the WHERE clause, assuming the compiled execution plan doesn't figure this out on its own.  But it's a small suggestion, and if it stays the way it is, it's fine by me.  Functionally, it looks completely consistent with the original code.

Thanks!

hazzik:
@JKeyser "SELECT 1" vs "SELECT *" does not really matter, please see this SO answer: http://stackoverflow.com/a/1597487/259946

And this is explicitly told by SQL'92 standard

> 3) Case:
>
>    a) If the <select list> "*" is simply contained in a <subquery>
>      that is immediately contained in an <exists predicate>, then
>      the <select list> is equivalent to a <value expression> that
>      is an arbitrary <literal>.

PR #410, opened 2015-03-04
NH-2176 - Add test case

hazzik:


hazzik:
Need to cleanup whitespaces

hazzik:
Related to #532

hazzik:
@fredericDelaporte I need someone with fresh eyes to look into this

fredericDelaporte:
I may be completely wrong, but I fear NHibernate tries to support an invalid programming model about transaction scope currently.

1. The session detects ambiant transaction and enlist with them.
2. A transaction scope in which a session is enlisted needs the session for completing the transaction (distributed or not).
3. If the session is closed (`ISession.Close()`) before transaction scope disposal, the transaction scope will fail.
4. What I see as a cursed hack causes `ISession.Dispose` to not actually dispose the session if it is enlisted with an ongoing transaction scope. It just set a flag for transaction completion to cause the session disposal. This allows "disposing" the session although it participates in a an ongoing transaction scope.

It looks to me that point 4 is an attempt at supporting a model that violates `IDisposable` pattern. We should not try supporting the disposal of a ressource which is still in use.

If we look at how this is handled in [EF 6 (doc updated on October 2016)](https://msdn.microsoft.com/en-us/library/dn456843(v=vs.113).aspx), it dodges the trouble in a quite more valid way:

1. The context to be disposed inside a transaction scope has to be provided with a SQL connection it will not handle.
2. The context has to be flushed to db before being disposed.

That way, the context is indeed no more needed by the transaction scope for completing the transaction.

Point 1. is important for allowing other contextes nested inside the transaction scope to access the changes done on the data by a previous context while still in the ongoing transaction scope. (Otherwise they have [this issue](https://stackoverflow.com/q/21202982/1178314).)

I think we need to do breaking changes for having a more reasonable implementation:

1. Cease delaying the session disposal on dispose, even when nested in a transaction scope.
2. Cease throwing in connection manager when closing a connection while in a transaction scope: just pass along the call to the connection provider (or de-reference non-owned connection).
3. Mandates session Flush before disposal for changes to be persisted.
4. Change the enlistment: if the session has been disposed, reply `Done` on prepare, no-op on session.

Well ideally, we should still perform post-transaction operations required for things like the second level cache, although the session is disposed. I do not know if those operations can be moved outside session responsibility easily or not.

An additional point could be added for an alternate to point 1. of EF case:

5. Render connection provider transaction scope aware: it will need to enlist too (at least receive transaction completed event), and will give the same connection to sessions executing in the same scope. It will no more close them if in a transaction scope, but will wait for the transaction scope completion.

fredericDelaporte:
Checking the state of Hibernate, for the JPA case, they do something similar to NHibernate delayed disposal of the session. But for the JDBC case, they actually close the session, but not the JdbcCoordinator if a shared transaction is ongoing. The connection is handled by the JdbcCoordinator.

I think we should replicate the JDBC handling.

fredericDelaporte:
Still more thoughts on the subject: currently on NHibernate side, it looks like people try to use the old entity mode switching capability (`GetSession(EntityMode mode)`) for sharing the connection among sessions. And when doing [NH-3722](https://nhibernate.jira.com/browse/NH-3722) as a follow up to [HHH-6330](https://hibernate.atlassian.net/browse/HHH-6330), instead of completely removing the functionality as Hibernate has done, we have just removed the entity mode argument and renamed it `GetChildSession()`.

We are quite diverging from Hibernate by doing so. They do not need this because they have decoupled connexion handling from the session.

The more I look to all that, the more I think we should re-align our session/transaction/connection handling with Hibernate.

hazzik:
Agree

src/NHibernate.Test/NHSpecificTest/NH2176/Fixture.cs
```diff
@@ -99,8 +105,42 @@ public ITransaction CreateTransaction(ISessionImplementor session)
...
+				// multiple calls with the same ongoing transaction, but some others may not.)
+				var current = System.Transactions.Transaction.Current;
+				var connection = session.Connection;
+				System.Transactions.Transaction previous;
+				if (!_sessionsTransaction.TryGetValue(connection, out previous) || previous != current)
+				{
+					_sessionsTransaction.AddOrUpdate(connection, current, (s, t) => current);
+					if (current == null &&
+						// This will need an ad-hoc property on Dialect base class instead.
+						(session.Factory.Dialect is SQLiteDialect || session.Factory.Dialect is MsSqlCeDialect))
```
hazzik: I think it shall be a dialect setting

hazzik:
I think that the current approach can be safely combined with the new approach.

hazzik:
I'm going to run non managed oracle on these changes

hazzik:
Nice,

```
PASS    - NHibernate.Test.NHSpecificTest.NH2176.Fixture.MultipleConsecutiveTransactionScopesCanBeUsedInsideASingleSession
```

src/NHibernate.Test/NHSpecificTest/NH2176/Fixture.cs
fredericDelaporte: Yes, as per comment above.

>      // This will need an ad-hoc property on Dialect base class instead.

hazzik: I saw the comment only after I commented:)

fredericDelaporte: By the way it looks like a bunch of connections does not need that, they are able of "unenlisting" from ended transaction by themselves. But not all.

hazzik: Actually, it should be driver setting, not dialect. Because SqlServer via ODBC requires this change, and SqlServer via native does not.

fredericDelaporte:
> I think that the current approach can be safely combined with the new approach.

As I write in [NH-4011](https://nhibernate.jira.com/browse/NH-4011), I do not think so. The current approach triggers connection releases from the `TransactionCompleted` event, which may run concurrently to following scope start. In such case that means the following transaction has started using the previous connection, breaking the `AfterStatement` contract: it should have acquired a new connection instead. At that point that can still work (helped by the explicit enlistment), if the connection is no more interacted with by the previous transaction events. But then this `TransactionCompleted` causes its release...

I thought about moving the release from that `TransactionCompleted` event to the 2 PC callback instead: the scope disposal is supposed to block until the transaction is committed or rollbacked. So those 2PC callbacks, even if executed in another thread, are never supposed to be able to run concurrently with a scope created after current scope disposal.

But I think that will just swap a bug for another as stated in NH-4011 and illustrated by NH-3968: during the 2PC session callbacks, in most case, the session connection is already enlisted itself in the transaction and receives its own 2PC calls from the transaction (which is illustrated by NH-3968, where the connection 2PC prepare phase seems to have been executed before the session 2PC prepare phase). Releasing it there is likely to cause issues too.

The custom transaction factory used here also need some internal API changes for avoiding breaking things like serialization. Currently, the way it works cause it to force session to acquire a connection almost all the time. Due to this, serialization tests fail because a connection is acquired at unexpected session states.

src/NHibernate.Test/NHSpecificTest/NH2176/Fixture.cs
```diff
@@ -99,8 +105,42 @@ public ITransaction CreateTransaction(ISessionImplementor session)
...
 			// commits. Explicit Flush calls required.
+			// Still make sure the session connection is enlisted, in case it was acquired before 
+			// transaction scope start.
+			// Will not support nested transaction scope. (Will throw, while current NHibernate
+			// just stay in previous scope.)
+			// Will cause an "earlier than required" connection acquisition.
+			// It is required to enlist with null when the scope is ended, otherwise using
+			// the transaction without a new scope will fail by attempting to use it inside
+			// the completed scope.
+			// If an explicit transaction is ongoing, we must not enlist.
```
hazzik: >If an explicit transaction is ongoing, we must not enlist.

Why?

fredericDelaporte: Because it throws with a bunch of connection. Firebird connection seems to accept that, but not `SqlConnection`, `SQLiteConnection`, and probably a bunch of others.

fredericDelaporte: At first I had locally tested the custom transaction factory without that safeguard, and it had succeeded. Only to immediately fail on teamcity build. I had not realized I had left my local test config on Firebird. Switching it back to Sql Server caused it to fail locally too.

hazzik:
It seems that `session.Connection.EnlistTransaction(System.Transactions.Transaction.Current);` at Line 33 of AdoNetWithDistributedTransactionFactory fixes this issue.

src/NHibernate.Test/NHSpecificTest/NH2176/Fixture.cs
fredericDelaporte: Well, it is not a trouble for Sql Server since it supports it in both case. But that is a trouble for SQLite since it will require it if accessed through ODBC (I have just checked, there is an ODBC driver for that) but it will not support it if accessed through its managed driver.

fredericDelaporte:
> It seems that `session.Connection.EnlistTransaction(System.Transactions.Transaction.Current);` at Line 33 of AdoNetWithDistributedTransactionFactory fixes this issue.

It will very likely fix [NH-2176](https://nhibernate.jira.com/browse/NH-2176), but it will not fix:

- [NH-2237](https://nhibernate.jira.com/browse/NH-2237) - unlocking of entities occurring while next transaction already started: due to `TransactionCompleted` executing in DTC cases after transaction scope completion in separated and concurrent thread.
- [NH-3968](https://nhibernate.jira.com/browse/NH-3968) - connection already prepared during session prepare phase: due to trying to re-use the session connection for flushing from session prepare phase while this connection is itself enlisted for its own 2PC in the same transaction.

Unlikely to fix:

- [NH-2107](https://nhibernate.jira.com/browse/NH-2107) - connection not released on rollback. This bug lacks information for me to actually understand how it gets triggered. But I highly suspect this is bounded to the current pattern of delaying session disposal and connection release to transaction completion events.
 - [NH-2238](https://nhibernate.jira.com/browse/NH-2238) - prepare phase failure. Again lack of informations, but looks bounded to trying to flush from session prepare phase with a connection already enlisted itself for its own 2PC.
- [NH-3023](https://nhibernate.jira.com/browse/NH-3023) - deadlock causing connection pool corruption: very likely same root cause than NH-2107.
- [NH-3227](https://nhibernate.jira.com/browse/NH-3227) - rollback causing connection pool corruption: almost duplicate of NH-3023.

That is why I consider this change is not enough to fix current approach.

hazzik:
NH-2238 is exactly the same as NH-3968

fredericDelaporte:
(Well, the time for me to consolidate my previous comment, it seems you have met NH-3968. A comment  notification on #532 indicates that in my mailbox, but I do not see it online.)

hazzik:
>but I do not see it online.

image

hazzik:
>but I do not see it online.

I deleted it along with the changes.

fredericDelaporte:
> NH-2238 is exactly the same as NH-3968

Having the same root cause, very likely. But the former writes about a closed connection exception while the later is about a timeout trying to use it. 

> but I do not see it online.

I'd guessed you have deleted it of course, but sometime I prefer tell facts (not seeing it) rather than guesses (being deleted) even when obvious. (And sometime that's Github auto-loading of new comments which does not always trigger when I would have expected.)

src/NHibernate.Test/NHSpecificTest/DtcFailures/DtcFailuresFixture.cs
```diff
@@ -19,6 +19,11 @@ public class DtcFailuresFixture : TestCase
 	{
 		private static readonly ILog log = LogManager.GetLogger(typeof(DtcFailuresFixture));
 
+		protected override void Configure(Configuration configuration)
+		{
+			configuration.SetProperty(Cfg.Environment.TransactionStrategy, "NHibernate.Test.NHSpecificTest.NH2176.CustomAdoNetTransactionFactory, NHibernate.Test");
```
fredericDelaporte: Testing the custom factory with DTC testes.

src/NHibernate.Test/NHSpecificTest/DtcFailures/DtcFailuresFixture.cs
```diff
@@ -87,8 +92,8 @@ public void WillNotCrashOnDtcPrepareFailure()
 		[Test]
 		public void Can_roll_back_transaction()
 		{
-			if (Dialect is FirebirdDialect)
-				Assert.Ignore("Firebird driver does not support distributed transactions");
+			/*if (Dialect is FirebirdDialect)
+				Assert.Ignore("Firebird driver does not support distributed transactions");*/
```
fredericDelaporte: In my local test Firebird supports them, so just letting them run to test with teamcity too.

src/NHibernate.Test/NHSpecificTest/DtcFailures/DtcFailuresFixture.cs
```diff
@@ -113,8 +118,8 @@ public void Can_roll_back_transaction()
 		[Description("Another action inside the transaction do the rollBack outside nh-session-scope.")]
 		public void RollbackOutsideNh()
```
fredericDelaporte: This one showcase the roadblock: session close within an ongoing transaction scope is rejected by connection manager.

The connection manager close does close any pending explicit transaction, then call disconnect which checks there are no pending transaction, explicit or ambient. The ambient check on disconnect should be removed, ambient transaction supports connection disconnect.

This need changes in NHibernate project even for testing:  the connection manager is not pluggable.

src/NHibernate.Test/NHSpecificTest/NH2176/Fixture.cs
```diff
@@ -0,0 +1,188 @@
...
+			catch (InvalidOperationException)
+			{
+				// This damn thing may yield an invalid operation exception (instead of null
+				// or of a completed transaction) if we are between scope.Complete() and
+				// scope.Dispose(). This happen when having completed the scope then disposing
+				// the session and only after that disposing the scope.
+				// Instead of testing System.Transactions.Transaction.Current here, storing in
+				// connection manager the ambient transaction associated to the connection
+				// (and updating it when enlisting) then checking that stored transaction would
+				// reduce testes on Transaction.Current.
```
fredericDelaporte: Another bug fixed in custom transaction factory: accessing to Transaction.Current may throw, need to catch.

src/NHibernate.Test/NHSpecificTest/NH2176/Fixture.cs
```diff
@@ -0,0 +1,188 @@
...
+			try
+			{
+				return System.Transactions.Transaction.Current;
+			}
+			catch (InvalidOperationException)
+			{
+				// This damn thing may yield an invalid operation exception (instead of null
+				// or of a completed transaction) if we are between scope.Complete() and
+				// scope.Dispose(). This happen when having completed the scope then disposing
+				// the session and only after that disposing the scope.
```
fredericDelaporte: Another bug fixed in custom transaction factory: accessing Transaction.Current may throw, need to catch.

fredericDelaporte:
Doomed attempt, I have already seen locally that the second phase of 2PC can be executed after transaction scope disposal. (Illustrated by NH-2420.) I need to do some for edits on NH-4011!

I may still add one more attempt, if I find a way through lock to ensure second phase is done before any new session usage. But it will probably be a clutch reducing risks but not eliminating them.

src/NHibernate.Test/NHSpecificTest/NH1632/Fixture.cs
```diff
@@ -84,43 +84,51 @@ public void When_commiting_items_in_DTC_transaction_will_add_items_to_2nd_level_
...
-				using (var s = sessions.OpenSession(connection))
+				using (var tx = new TransactionScope())
 				{
-					var nums = s.Load<Nums>(29);
-					Assert.AreEqual(1, nums.NumA);
-					Assert.AreEqual(3, nums.NumB);
+					using (var s = sessions.OpenSession(connection))
+					{
+						Nums nums = null;
+						Assert.DoesNotThrow(() => nums = s.Load<Nums>(29), "Failed loading entity from second level cache.");
```
fredericDelaporte: Leaving an explicit message instead of caching failure, instead of failing, then failing to tear down, and letting no much clues.

src/NHibernate.Test/NHSpecificTest/NH1632/Fixture.cs
```diff
@@ -84,43 +84,51 @@ public void When_commiting_items_in_DTC_transaction_will_add_items_to_2nd_level_
...
+						Assert.AreEqual(3, nums.NumB);
+					}
+					tx.Complete();
 				}
-				tx.Complete();
 			}
-
-			using (var s = sessions.OpenSession())
-			using (var tx = s.BeginTransaction())
+			finally
```
fredericDelaporte: Avoiding additional tear down failure.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -24,52 +25,36 @@ public ITransaction CreateTransaction(ISessionImplementor session)
 
 		public void EnlistInDistributedTransactionIfNeeded(ISessionImplementor session)
 		{
+			// Ensure the session does not run on a thread supposed to be blocked, waiting
+			// for transaction completion.
+			session.TransactionContext?.WaitOne();
```
fredericDelaporte: Transaction completion may concurrently nullify it, but `?.` handles that (correctly evaluate it only once, guaranteeing it will not try to call `WaitOne` on null).

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -94,90 +79,196 @@ public class DistributedTransactionContext : ITransactionContext, IEnlistmentNot
 		{
 			public System.Transactions.Transaction AmbientTransation { get; set; }
 			public bool ShouldCloseSessionOnDistributedTransactionCompleted { get; set; }
-			private readonly ISessionImplementor sessionImplementor;
+
+			private readonly ISessionImplementor _sessionImplementor;
+			private readonly ManualResetEvent _waitEvent = new ManualResetEvent(true);
+			private readonly AsyncLocal<bool> _bypassWait = new AsyncLocal<bool>();
```
fredericDelaporte: Not already needing async but well, better avoid adding new thread locals.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -94,90 +79,196 @@ public class DistributedTransactionContext : ITransactionContext, IEnlistmentNot
...
+						_sessionImplementor.TransactionContext = null;
+					}
+				}
+				finally
+				{
+					_afterTransactionActionDone = true;
+					// Dispose releases blocked threads by the way.
+					// Must dispose in case !ShouldCloseSessionOnDistributedTransactionCompleted, since
+					// we nullify session TransactionContext, causing it to have nothing still holding it.
+					Dispose();
```
fredericDelaporte: This was missing and is a transaction leak in NHibernate 4.1. When scopes are disposed inside session life-span, the `TransactionContext` was nullified on session without being disposed, causing the `AmbientTransaction` clone to not be disposed of explicitly.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -94,90 +79,196 @@ public class DistributedTransactionContext : ITransactionContext, IEnlistmentNot
...
+						_waitEvent.Set();
+						throw new HibernateException(
+							"Synchronization timeout for transaction completion. This is very likely a bug in NHibernate.");
+					}
+				}
+				catch(Exception ex)
+				{
+					_logger.Warn(
+						"Synchronization failure, assuming it has been concurrently disposed and do not need sync anymore.",
+						ex);
```
fredericDelaporte: I have not witnessed this, but it may happen if between the test on `_isDisposed` and the call to `WaitOne`, the transaction completion event dispose the wait event.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -94,90 +79,196 @@ public class DistributedTransactionContext : ITransactionContext, IEnlistmentNot
...
+			{
+				if (disposing)
+				{
+					if (AmbientTransation != null)
+					{
+						AmbientTransation.Dispose();
+						AmbientTransation = null;
+					}
+					_waitEvent.Set();
+					_waitEvent.Dispose();
```
fredericDelaporte: In my tests, disposing right after signaling does not cause an exception for thread potentially waiting. (I have run all test first without the catch around `WaitOne`, and do not got any failures, all tests passed.)

src/NHibernate.Test/NHSpecificTest/NH2420/Fixture.cs
```diff
@@ -55,41 +56,56 @@ public void ShouldBeAbleToReleaseSuppliedConnectionAfterDistributedTransaction()
...
+				connection?.Dispose();
 			}
 
+			// It appears neither the second phase of the 2PC nor TransactionCompleted 
+			// event are guaranteed to be executed before exiting transaction scope disposal.
+			// When having only 2PC, the second phase tends to occur after reaching that point
+			// here. When having TransactionCompleted event, this event and the second phase
+			// tend to occur before reaching here. But some other NH cases demonstrate that
+			// TransactionCompleted may also occur "too late".
+			((ISessionImplementor) s).TransactionContext?.WaitOne();
```
hazzik: shall be `s.GetSessionImplementor()`

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -94,90 +79,196 @@ public class DistributedTransactionContext : ITransactionContext, IEnlistmentNot
...
+
 					enlistment.Done();
-					IsInActiveTransaction = false;
 				}
 			}
 
-			void IEnlistmentNotification.Rollback(Enlistment enlistment)
+			#endregion
+
+			public void TransactionCompleted(object sender, TransactionEventArgs e)
```
hazzik: TransactionCompleted shall be incorporated to IEnlistmentNotification.Commit/IEnlistmentNotification.Rollback/IEnlistmentNotification.InDoubt

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -94,90 +79,196 @@ public class DistributedTransactionContext : ITransactionContext, IEnlistmentNot
...
 			{
-				using (new SessionIdLoggingContext(sessionImplementor.SessionId))
+				e.Transaction.TransactionCompleted -= TransactionCompleted;
+				// This event may execute before second phase, so we cannot try to get the success from second phase.
+				// Using this event is required in case the prepare phase failed and called force rollback: no second
+				// phase would occur for this ressource.
```
hazzik: This shall be handled on force rollback then

src/NHibernate.Test/NHSpecificTest/NH2420/Fixture.cs
fredericDelaporte: Probably not the first time... Maybe should we remove all the other alike catches which still liter the tests.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
fredericDelaporte: Documentation on which conditions the second phase will not be called at all lacks. At first I have put it all in second phase callbacks, just to realize thanks to some tests there were cases for which it was not good.

So the forced rollback from session volatile prepare phase is maybe only one case in which second phase do not occur. For staying on the safe side, we would better keep that handling in `TransactionCompleted`.

The fact the completion actions were called from in-doubt too made me wonder whether in in-doubt case, the transaction completed call be not called either. There too, for staying on the safe side, I have decided to keep that call. This transaction factory is the result of many years of fiddling with many failing case for trying to support them all. During my own fiddling with it, I have seen that many design decisions in it, lacking comments and which were looking as overly complicating the code, where indeed needed for some reasons. So I had to kept them, but at least adding comments for most of them.

Moreover, handling transaction completion actions from the prepare phase in case of force rollback may cause other issues, since the session connection state at that point is unknown: has it already processed its own prepare phase? Is it processing it concurrently? Raising completion actions there would trigger its release, and I do not know if all connection implementations would support that.

This is also why I consider we should still provide other transaction factories even if we appear to get this one working: there too many points where we still act on the connection with concurrency risks. Providing other transaction factories would allow whoever wishing it to not take those risks, at the cost of some more explicit flushes calls.

PR #411, opened 2015-03-06
NH-3765 - Prevent updating timestamp cache more than once for the same space.

TheCloudlessSky:
Right now, when the timestamp cache is updated it does not account for multiple updates to the same space. This causes multiple unnecessary writes to the second level cache. We have a case where there are ~20 inserts and each causes a timestamp cache put even though the value (the new timestamp) is the same.

Ideally `PreInvalidate(object[] spaces)` and `Invalidate(object[] spaces)` should be changed to accept `ISet<string>` like `IsUpToDate()`, but that would be a breaking change.

### DIFF
--- a/src/NHibernate/Cache/UpdateTimestampsCache.cs
+++ b/src/NHibernate/Cache/UpdateTimestampsCache.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using System.Linq;
 
 using NHibernate.Cfg;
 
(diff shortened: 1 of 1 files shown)


hazzik:
@TheCloudlessSky we need to have JIRA issue linked. Please follow [CONTRIBUTING](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md) guide

hazzik:
@TheCloudlessSky can you please add some tests?

TheCloudlessSky:
Right sorry. I added it to the Jira issue but not here: https://nhibernate.jira.com/browse/NH-3765. I'll see what I can come up with tests for this.

TheCloudlessSky:
Well, I realized that multiple PreInvalidates are occurring *outside* the loop, so this PR is moot. I'm going to close the associated issue as well. There's no way to easily fix this like I thought...

hazzik:
@TheCloudlessSky but, can you still provide tests?

TheCloudlessSky:
Not sure what you mean? I'm abandoning this PR because it makes sense that the timestamp is being pre-invalidated many times.

The main reason `PreInvalidate()` why is called many times is because it's called by `ActionQueue.ExecuteActions()` which can be called at *many* different points and could therefore trigger *many* different `PreInvalidate()` calls. I thought it was all called in one place, but after digging into this more, I was wrong.

Also, after investigating this, we're abandoning using the QueryCache and using higher-level caching because it causes too many cache misses (the whole table's timestamp is invalidated each time any entity is updated -- which is often for our application).

hazzik:
Ok

PR #414, opened 2015-03-16
NH-3767 - Wrong aliases when querying a table with the name starting with Select, From or As in Oracle

lnu:
Description in jira
Workaround for the bug, should be reviewed


hazzik:
Is this only Oracle's thing?

lnu:
I think so. The parsed used is used only in the Oracle8iDialect class. That's why I think this is a workaround and maybe a refactoring/cleaning of some stuffs could be done.

PR #415, opened 2015-04-03
Add check to ensure that IUserCollectionType.Instantiate returns uninitialized collection

Sebazzz:
The documentation on `IUserCollection.Instantiate` already says that the collection needs to be uninitialized: https://github.com/nhibernate/nhibernate-core/blob/dc71c7df5c22d82991bc1b009ecdfe2ec9902b45/src/NHibernate/UserTypes/IUserCollectionType.cs#L10-L13

It would be nice to have a programmatic check that this assumption shall be true to prevent possible invalid or error-prone behaviour downstream.


no149:
Interestingly , I ran into this issue a short time ago and I in my case it was because a generic IList could not be cast to a non-generic IList. The reference code lies In PersistentGenericList.cs: 42. That is from the source code for Build 3.3.4.GA.

Sebazzz:
After some debugging I found out that this line, in the test, is the cause:

    return new PersistentGenericSet<T>(session , new HashSet<T>(EqualityComparer<T>.Default));

Because the persistent collection is already given a wrapped collection, the persistent collection considers itself as initialized. Later on it is asked to check if it dirty, where it nullrefs because it hasn't got a snapshot (because it is considered initialized). 

This is confusing API behavior, and probably a check needs to written that `IUserCollectionType.Instantiate(ISessionImplementor, ICollectionPersister)` doesn't return an initialized collection?

Sebazzz:
Added the proposed check to the pull request and amended the test for this check.

hazzik:
@Sebazzz can you please fix the whitespace changes and also rebase on top of master?

Sebazzz:
No problem, I will do it tonight.

Sebazzz:
Took a long night :) Rebased on top of master and squashed the changes.

hazzik:
Took even longer to return here.

The documentation on `IUserCollection.Instantiate` already says that the collection needs to be un-initialized: https://github.com/nhibernate/nhibernate-core/blob/dc71c7df5c22d82991bc1b009ecdfe2ec9902b45/src/NHibernate/UserTypes/IUserCollectionType.cs#L10-L13

Sebazzz:
Done

src/NHibernate.Test/NHSpecificTest/NH3772/Fixture.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System.Linq;
+using NHibernate.Cfg.MappingSchema;
+using NHibernate.Mapping.ByCode;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3772
+{
+	public class Fixture : TestCaseMappingByCode
```
fredericDelaporte: Lacking `TestFixture` attribute, causing it to not generate an async counterpart.

hazzik: True

hazzik: It does not generate async code because `action` delegate declared outside of `Assert.That(...)`

fredericDelaporte: Ok, but it still should have the attribute for consistency with other tests.

hazzik: I've added the attribute

PR #301, opened 2014-08-14
NH-3470 - Allow Session.Query to load entities as read-only (NH-3658)

rjperes:
Cleanup with tests passing.

PR #418, opened 2015-04-16
NH-3470 - Allow Session.Query to load entities as read-only

rjperes:
https://nhibernate.jira.com/browse/NH-3470


hazzik:
Why?

hazzik:
@rjperes why? There are 3 PR with the same functionality.

fredericDelaporte:
If #601 get merged, this change could be done on top of it with less code and without a new extension method/result operator/....

hazzik:
Replaced by #1500

hazzik:
Duplicate of #301

PR #419, opened 2015-04-16
- Replaced NHibernate.Loader.TopologicalSorter by a more performant impl...

TAkkerman:
...ementation.

- Because of a slight change in the (internal) Interface of TopologicalSorter, JoinWalker had to be adjusted (note that the removed parameter i was identical to the return value for the current use of the sorter)


TAkkerman:
Well, it seems that the topological sort itself was alright, but there must be more restrictions on the order than those that are exposed to the sort algorithm. Will try to update to a variant that matches to the order the old algorithm produces when there are multiple valid solutions.

hazzik:
Hi @TAkkerman, can you please enable [Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) feature?

TAkkerman:
'allow edits from maintainers' should be active now.

hazzik:
@TAkkerman I've force pushed with rebase on top of master. I've cleaned up indents & namings.

PR #425, opened 2015-05-01
NH-3783 - Expose the existing IsOrderUpdatesEnabled property via configuration.

graemefoster:
Property already exists in code and is implemented, but was never set.
Setting it to true can improve update batching on more complex graphs.

Jira link here:
https://nhibernate.jira.com/browse/NH-3783


hazzik:
@graemefoster could you please create/find a jira issue for it?

graemefoster:
@hazzik jira issue is here
https://nhibernate.jira.com/browse/NH-3783

colewerner:
Shouldn't the default value for this and `order_inserts` be true? It seems to me if they aren't true by default you could run into potential deadlocks at least for updates anyway. http://stackoverflow.com/questions/24175257/maintaining-order-of-updates-across-transactions-to-avoid-deadlocks

PR #432, opened 2015-06-08
NH 3797

PleasantD:
See https://nhibernate.jira.com/browse/NH-3797

Modify the QueryModelVisitor to treat group by keys as first-class HQL candidates for select clause rewriting.

PR #433, opened 2015-06-17
NH-3800 and NH-3681

PleasantD:
Fixes NH-3681 by allowing the GroupBySelectClauseRewriter to correctly match the group by elements.
Fixes NH-3800 by unwrapping constant ArrayIndex expressions to their inner expressions.

### DIFF
--- a/src/NHibernate/Linq/ExpressionExtensions.cs
+++ b/src/NHibernate/Linq/ExpressionExtensions.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Linq.Expressions;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
@@ -30,6 +31,26 @@ namespace NHibernate.Linq
 			var query = fromClause.FromExpression as SubQueryExpression;
 			if (query == null) return false;
 	
+			return query.QueryModel.ResultOperators.Contains(groupBy);
+		}
+
+		public static bool IsGroupingElementOf(this QuerySourceReferenceExpression expression, GroupResultOperator groupBy)
+		{
+			var fromClause = expression.ReferencedQuerySource as MainFromClause;
+			if (fromClause == null) return false;
+
+			var innerQuerySource = fromClause.FromExpression as QuerySourceReferenceExpression;
+			if (innerQuerySource == null) return false;
+
+			if (innerQuerySource.ReferencedQuerySource.ItemName != groupBy.ItemName
+				|| innerQuerySource.ReferencedQuerySource.ItemType != groupBy.ItemType) return false;
+
+			var innerFromClause = innerQuerySource.ReferencedQuerySource as MainFromClause;
+			if (innerFromClause == null) return false;
+
+			var query = innerFromClause.FromExpression as SubQueryExpression;
+			if (query == null) return false;
+
 			return query.QueryModel.ResultOperators.Contains(groupBy);
 		}
 	}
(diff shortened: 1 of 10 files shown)


src/NHibernate/Linq/ReWriters/ArrayIndexExpressionFlattener.cs
```diff
@@ -0,0 +1,33 @@
...
+            var visitedExpression = base.VisitBinaryExpression(expression);
+
+            if (visitedExpression.NodeType != ExpressionType.ArrayIndex)
+                return visitedExpression;
+
+            var index = expression.Right as ConstantExpression;
+            if (index == null)
+                return visitedExpression;
+
+            var expressionList = expression.Left as NewArrayExpression;
```
hazzik: Please add check that this is a `NewArrayInit` and not `NewArrayBounds`

hazzik:
@PleasantD can you please update the pull request according to [CONTRIBUTING.md](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md). Eg, tabs instead of spaces

PleasantD:
Done and done! Is there anything else I can do to expedite the release of this and https://github.com/nhibernate/nhibernate-core/pull/432?

hazzik:
@PleasantD 

[TransparentIdentifierRemovingExpressionTreeVisitor](http://www.symbolsource.org/Public/Metadata/NuGet/Project/Remotion.Linq/1.13.183.0/Release/.NETFramework,Version=v3.5/Remotion.Linq/Remotion.Linq/Parsing/ExpressionTreeVisitors/TransparentIdentifierRemovingExpressionTreeVisitor.cs) (doing the same as you did with arrays, but for anonymous types) also visits SubQueryExpression, can you investigate, if we need so?

hazzik:
@PleasantD also, can you squash your commits?

src/NHibernate/Linq/GroupBy/GroupBySelectClauseRewriter.cs
```diff
@@ -31,7 +31,7 @@ private GroupBySelectClauseRewriter(GroupResultOperator groupBy, QueryModel mode
 
 		protected override Expression VisitQuerySourceReferenceExpression(QuerySourceReferenceExpression expression)
 		{
```
hazzik: Probably need to add 

```
			if (!IsMemberOfModel(expression))
			{
				return base.VisitQuerySourceReferenceExpression(expression);
			}
```
If necessary

src/NHibernate/Linq/ExpressionExtensions.cs
```diff
@@ -32,5 +33,30 @@ public static bool IsGroupingKeyOf(this MemberExpression expression,GroupResultO
 	
 			return query.QueryModel.ResultOperators.Contains(groupBy);
 		}
+
+		public static bool IsGroupingElementOf(this QuerySourceReferenceExpression expression, GroupResultOperator groupBy)
+		{
+			if (expression.ReferencedQuerySource == groupBy)
+			{
+				return true;
```
hazzik: This never happens

PleasantD: Odd... that was the original (and only) check that GroupBySelectClauseRewriter.VisitQuerySourceReferenceExpression was doing before I modified it.
No wonder it never worked.

PleasantD:
@hazzik regarding the ArrayIndexExpressionFlattener needing to visit SubQueryExpressions
It seems like it could, but it may not need to.

I wrote an example test and found that the initial pass wasn't flattening the array access, but then QueryModelVisitor.GenerateHqlQuery gets called on the subquery by HqlGeneratorExpressionTreeVisitor, causing the array indexes to be flattened on the second pass.

If the SubQueryExpression query model is flattened by the initial pass, then the same result occurs.

PR #435, opened 2015-07-01
Added SybaseASE nhibernate driver for the sybase ado.net 4 connector …

nrubino:
…(currently nhibernate only supported sybase ado .net 2 connector)

Also fixed the trim function in the sybase ase 15 dialect.


hazzik:

nrubino:
All set

fredericDelaporte:
@nrubino, it seems SAP ASE has no free test version available without an expiry. We now tend to check changes to dialects with at least local version of the database, but we do not expect any contributor to buy a licence for that.
Unfortunately, contrary to SAP HANA and SAP SQL Anywhere, it seems SAP ASE has only a 90 days trial version, which is unsuitable. (I do not want to install it, check some PRs, then be no more able to do so after three months.)

But maybe I have missed something. Do you know if it is possible to have a SAP ASE for free, without an expiration, for testing purposes?

nrubino:
@fredericDelaporte - SAP used to have an ASE developer edition for windows but it seems they have taken this away.  There is an expression version of ASE for Linux that can be found here.  https://www.sap.com/cmp/syb/crm-xu15-int-asexprdm/index.html

PR #434, opened 2015-06-24
NH-3474 - Fix support of GroupBy with constant keys

PleasantD:
* Remove the GROUP BY clause entirely if all keys are constant
* Re-write the GROUP BY clause using only non-constant keys

PR #436, opened 2015-07-07
NH-3801

PleasantD:
- Conditional statements are now generated in SQL for select clauses
- Case statements in the select clause don't generate unexpected or implicit joins for the test expression
- Combined the logic from SelectJoinDetector and ResultOperatorAndOrderByJoinDetector, also handles case statement join logic
- Removed most of the changes from NH-3797 (pull request #432) as they are now unnecessary because the select clause now does the right thing


PleasantD:
Please include in milestone 4.1.0, and as usual, let me know if there are any problems or missing tests

PleasantD:
I think this pull request will have merge issues with #434, which I can fix once one of them has been merged.

hazzik:
@PleasantD which should be merged first?

hazzik:
In GroupByTests.GroupByComputedValue instead of having that SQL generated 

```
select 
    cast(case when order0_.CustomerId is null then @p0 else @p1 end as INT) as col_0_0_, 
    cast(count(*) as INT) as col_1_0_
from Orders order0_ 
group by cast(case when order0_.CustomerId is null then @p0 else @p1 end as INT) 
```  

We should wrap case into subselect, so this will allow this query to work on some databases (SQL Server via ODBC, Firebird, Sql Server CE). What do you think?

Also there are few test failing for Oracle:

FAIL    - NHibernate.Test.Linq.ByMethod.GroupByTests.GroupByComputedValueInObjectArrayWithJoinInRightSideOfCase
FAIL    - NHibernate.Test.Linq.JoinTests.OrderLinesWithSelectingCustomerNameInCaseShouldProduceTwoJoins
FAIL    - NHibernate.Test.Linq.JoinTests.OrderLinesWithSelectingCustomerNameInCaseShouldProduceTwoJoinsAlternate

*** Tests broken since last recorded results ***
FAIL    - NHibernate.Test.NHSpecificTest.NH2378.Fixture.ShortEntityCanBeQueryCorrectlyUsingLinqProvider

PleasantD:
I'd maybe recommend merging this first, then I'll fix #434. That's what I did locally to build a prerelease.

PleasantD:
Yes, ideally we need to wrap the case into a subselect.
How do I do this? Do I have to modify the QueryModel?

On Tue, Jul 7, 2015 at 2:47 PM, Alexander Zaytsev <notifications@github.com>
wrote:

> In GroupByTests.GroupByComputedValue instead of having that SQL generated
>
> select
>     cast(case when order0_.CustomerId is null then @p0 else @p1 end as INT) as col_0_0_,
>     cast(count(*) as INT) as col_1_0_
> from Orders order0_
> group by cast(case when order0_.CustomerId is null then @p0 else @p1 end as INT)
>
> We should wrap case into subselect, so this will allow this query to work
> on some databases (SQL Server via ODBC, Firebird, Sql Server CE). What do
> you think?
>
> Also there are few test failing for Oracle:
>
> *** Tests new (failed) since last recorded results ***
> FAIL -
> NHibernate.Test.Linq.ByMethod.GroupByTests.GroupByComputedValueInObjectArrayWithJoinInRightSideOfCase
> FAIL -
> NHibernate.Test.Linq.JoinTests.OrderLinesWithSelectingCustomerNameInCaseShouldProduceTwoJoins
> FAIL -
> NHibernate.Test.Linq.JoinTests.OrderLinesWithSelectingCustomerNameInCaseShouldProduceTwoJoinsAlternate
>
> *** Tests broken since last recorded results ***
> FAIL -
> NHibernate.Test.NHSpecificTest.NH2378.Fixture.ShortEntityCanBeQueryCorrectlyUsingLinqProvider
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nhibernate/nhibernate-core/pull/436#issuecomment-119333700>
> .
>



-- 
*Duncan Munro*, Software Developer
duncan@pleasantsolutions.com | PleasantSolutions.com
<http://www.PleasantSolutions.com>
[image: Pleasant Solutions]

hazzik:
>Yes, ideally we need to wrap the case into a subselect.
>How do I do this? Do I have to modify the QueryModel?

I did not check the solution yet, and do not really know how exactly does it work, but, probably, in that case we shall not flatten subquery if we have `case` in a group by.

hazzik:
>- Combined the logic from SelectJoinDetector and ResultOperatorAndOrderByJoinDetector, also handles case statement join logic

There might be a reason why I left them separated.

PleasantD:
The oracle issue seems to be that the parameter doesn't match the charset of the DB. Although that seems to only affect the Oracle Managed build. Could that be a configuration issue?

PleasantD:
The SelectJoinDetector and ResultOperatorAndOrderByJoinDetector used to have (slightly) different logic. But as I was adding the modifications to allow the SelectJoinDetector to bypass joins for the conditional test expression, I found that the logic basically began to overlap. By unifying it you get a JoinDetector that can decide on a more granular level if it needs to create joins for entities.

hazzik:
>Could that be a configuration issue?

No

hazzik:
>>Could that be a configuration issue?

>No

The problem that case got casted to the default type of string, which is NVARCHAR2(255) in this case.

hazzik:
Merged, thanks

hazzik:
@PleasantD the oracle fails are related to https://nhibernate.jira.com/browse/NH-3787

PR #438, opened 2015-07-20
NH-3811 Added [ThreadStatic] to table counter 

BrettMcGirr:
so that table aliases can be generated consistently in a multi thread environment.

### DIFF
--- a/src/NHibernate.Test/MappingTest/TableFixture.cs
+++ b/src/NHibernate.Test/MappingTest/TableFixture.cs
@@ -1,3 +1,5 @@
+using System.Linq;
+using System.Threading;
 using NHibernate.Dialect;
 using NHibernate.Mapping;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


hazzik:
@BrettMcGirr 

Unfortunately we can not accept pull requests without proper tests. Can you please log a JIRA issue and add some tests?

BrettMcGirr:
Hi Alexander,

Thanks for the quick response. I’ve amended the pull request with some tests and created a Jira issue.

https://nhibernate.jira.com/browse/NH-3811

https://github.com/nhibernate/nhibernate-core/pull/438


Thanks,

Brett

BrettMcGirr:
Hi, any update on this?

oskarb:
In fact, the table counter is updated using:

```
public Table()
{
	uniqueInteger = tableCounter++;
}
```

And the increment operator is not guaranteed to be thread safe so potentially the problem (in the current code) is even worse. Though in practice no-one seems to have been bothered by that.

oskarb:
Effectively this limits the "unique table counter" to be unique only within a session factory scope, not within process scope. Not entirely sure this is proper, but we can try it.

oskarb:
Merged to master (7fc042419d2857af95eefb638b6edc5dfc4cc26e).

PR #441, opened 2015-07-27
Fix for NH-3795

gerritsen:
Compiling query over queries with the Roslyn compiler of Visual Studio breaks resolving alias names because Roslyn compiles lambda's differently. It now sometimes generates a compiler generated class. This compiler generated class has a member of the type of the class declaring the alias. The ExpressionProcessor is recursively searching for members and does not stop at the alias field but also returns the member of the compiler generated class. Members of compiler generated classes can never be used in a query over so they should never be returned

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3795/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3795/Fixture.cs
@@ -4,80 +4,76 @@ using NUnit.Framework;
 
 namespace NHibernate.Test.NHSpecificTest.NH3795
 {
-    /// <summary>
-    /// Tests in this class only failed when the code was build with the Roslyn compiler which is included in Visual Studio 2015
-    /// </summary>
-    [TestFixture]
-    public class Fixture : TestCase
-    {
-        protected Child childAliasField = null;
-        protected A aAliasField = null;
-
-        protected override IList Mappings
-        {
-            get
-            {
-                return new[] { "ParentChild.hbm.xml", "ABC.hbm.xml" };
-            }
-        }
-
-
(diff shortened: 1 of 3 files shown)


oskarb:
Thanks Marc,

Would you mind fixing the indentation style to match surrounding code (space vs.tabs). You may want to consider using EditorConfig.

Also, can you very briefly describe the problem and why this fix is correct?

gerritsen:
@oskarb I think it is correct now. Could you please check?

gerritsen:
The member of the compiler generated class should also have a compiler generated name.
The only way to check this is detecting if the name starts with the < character.
I don't like this but found this was the only way

hazzik:
I have an discussion of the similar issue on [DelegateDecompiler](https://github.com/hazzik/DelegateDecompiler/issues/54) and it has a link to [this SO answer](http://stackoverflow.com/a/2509524/259946) which describes how the compiler generates names for such members

hazzik:
Your fix does not work, if you move aliases to method.

hazzik:
Fix is here https://github.com/nhibernate/nhibernate-core/tree/NH-3795-4.0.x

gerritsen:
@hazzik I checked your tree and works in my cases. Looks like we're good to go.

danspam:
Is there a plan to release this fix soon?

hazzik:
@danspam  yes, there is a plan to release it by the end of this week.

hazzik:
Thanks. This fix has been released as 3.3.5, 3.4.1 and 4.0.4

mcollins4551:
any chance of getting this issue fixed in 3.1.0?

PR #442, opened 2015-08-12
NH-3816 - Fix Select with Conditional

PleasantD:
Went back to a process of nominating expressions from the Group By key selector.
These nominated expressions are wrapped in a custom expression type so that they may be transformed and remain nominated. The nominator then strips the wrapper from them (and the wrapper is ignored elsewhere) and nominates the internal expression.

### DIFF
--- a/src/NHibernate/Linq/Visitors/SelectClauseNominator.cs
+++ b/src/NHibernate/Linq/Visitors/SelectClauseNominator.cs
@@ -1,7 +1,9 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq.Expressions;
 using NHibernate.Linq.Functions;
 using NHibernate.Linq.Expressions;
+using NHibernate.Util;
 using Remotion.Linq.Parsing;
 
 namespace NHibernate.Linq.Visitors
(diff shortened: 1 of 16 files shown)


src/NHibernate/Linq/Visitors/SelectClauseNominator.cs
```diff
@@ -109,6 +112,29 @@ public override Expression VisitExpression(Expression expression)
 			return expression;
 		}
 
+		private bool DbCanCast(System.Type type)
```
hazzik: I think that probably needs to be decided by Driver/Dialect, as, for example, MySQL support only few types in cast function. 

https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Dialect/MySQLDialect.cs#L317
http://dev.mysql.com/doc/refman/5.7/en/cast-functions.html

PleasantD: @hazzik I don't think this is true.
At this point we are trying to nominate expressions that can be converted to HQL.
I don't think that conditionals in where clauses are checking dialect casting capabilities, and yet they all generate a case statement wrapped in a cast statement, using the IDENT node with the type to cast as.

I assume the dialect is responsible for converting the IDENT node into an appropriate DB type. If an IDENT type isn't supported, what does the dialect do?

PleasantD:
@hazzik Any comments on this? Any progress?

hazzik:
@PleasantD merged. Thanks

PR #444, opened 2015-09-24
 Update ThreadStaticSessionContext.cs 

twtja1:
The thread static session context should also be able to cache multiple sessions as well as the web session context.
So, the C/S application will be able to support multiple databases at the same time.It will return the correct cached session(if have) according to the factory of the model object.

### DIFF
--- a/src/NHibernate/Context/ThreadStaticSessionContext.cs
+++ b/src/NHibernate/Context/ThreadStaticSessionContext.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections;
+using NHibernate.Engine;
 
 namespace NHibernate.Context
 {


oskarb:
Why not use ThreadLocalSessionContext instead?

twtja1:
Because when we set up the property "current_session_context_class" in the NHibernate configuration file to "thread_static", it cannot support multiple databases at the same time.I think the values("call","thread_static","web" and "wcf_operation") of the property should be consistent with the performance of NHibernate.

fredericDelaporte:
`ThreadLocalSessionContext` is not the same as `ThreadStaticSessionContext`, and does support multiple factories. But it has some other issues, like being only partially implemented in regards to its xml documentation and advertised features, not being bindable from `CurrentSessionContext`, lacking a short configuration name, ...

This PR here re-implements the logic we already have in `MapBasedSessionContext`, meant to be derived from, and do not test the changes.

Closing in favor of #644.

PR #445, opened 2015-10-01
NH-3831 - NullableDictionary does not set _gotNullValue when using Add(TKey, TValue) method

cjzness:
[NH-3831](https://nhibernate.jira.com/browse/NH-3831)

Set _gotNullValue when using the NullableDictionary Add(TKey, TValue) method with a null key.
Added unit tests to verify all methods of adding a null key do so correctly.


hazzik:
Merged, thanks!

PR #447, opened 2015-10-08
Flush child sessions when parent is flushed.

czema:
For years I used GeneratedBy.Assigned() with a incremental PK for audit logging.  As soon as Save is called it issues an INSERT to the database, so it didn't matter if Flush was called on the child session.  I switched to guid.comb and now Save does not issue an INSERT (which is the correct behavior) but since the child session isn't flushed my audit record is never written.

### DIFF
--- a/src/NHibernate/Impl/SessionImpl.cs
+++ b/src/NHibernate/Impl/SessionImpl.cs
@@ -1462,6 +1462,15 @@ namespace NHibernate.Impl
 				{
 					flushEventListener[i].OnFlush(new FlushEvent(this));
 				}
+				
+				// Flush children when parent is flushed.
+				if (childSessionsByEntityMode != null)
+				{
+					foreach (var childSession in childSessionsByEntityMode)
+					{
+						childSession.Value.Flush();
+					}
+				}
 			}
 		}
 


hazzik:
@czema whould you able to add some tests?

hazzik:
Also, please create a jira issue here: http://nhibernate.jira.com

czema:
https://nhibernate.jira.com/browse/NH-3960

fredericDelaporte:
"needs work": no tests.

hazzik:
The implementation needs to be adjusted according to the master (childSessionsByEntityMode is replaced by single _childSession) and accompanied with tests.

fredericDelaporte:
The notion of child session does no more actually exist with [NH-4003](https://nhibernate.jira.com/browse/NH-4003), port of modern Hibernate way of building sessions. There is still a way to open a session from an existing one, but it has no "child" semantic. It is a mechanism for opening an "alike" session, I mean, using the same settings. It has still an option for sharing the transaction, but that does not go more far.

czema:
I just upgraded to NHibernate 5.  I get there are no child sessions any more. 

If I create an "alike" session, which shares the transaction, during the transaction's commit phase will it pick up this new session and flush it?

My use case: I call session.Flush().  Internal to NHibernate it calls my OnPostUpdate implementation.  I create a new session which shares the transaction and save some new audit records.  FlushMode is left at Auto (I tried Commit as well).  Later, when the transaction commits, it never flushes this "alike" session.  Is this the expected behavior?

fredericDelaporte:
`FlushMode.Auto` or `Commit` should apply to any session in the transaction. Please fill a bug report or directly PR a test case if there is an issue with this.

PR #448, opened 2015-11-12
NH-1285 - Drop/Create script with default_schema/default_catalog fix(SqlServer)

lnu:
https://nhibernate.jira.com/browse/NH-1285
related also to https://nhibernate.jira.com/browse/NH-1443 (both already closed) and #987.
In case the Schema is set using the Default_Schema property, the drop script for the generated constraint is  wrong: the schema is missing(Sql Server).


hazzik:
@lnu can you please avoid changing method signatures of `Dialect` class? Preferable create new virtual methods and use them, make old methods using new methods (or vice verse) and mark old methods as obsolete.

lnu:
Ok i'll check for this

lnu:
I did some changes. Is it what you were thinking about?

hazzik:
>Is it what you were thinking about?

Almost,

You forgot about this part: "make old methods using new methods (or vice verse)". The idea is that if you use not updated provider, it should still work as expected.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -752,12 +756,63 @@ public virtual string GetIfExistsDropConstraint(Table table, string name)
...
 		/// <summary>
+		/// The syntax that is used to check if a constraint does not exists before creating it
+		/// </summary>
+		/// <param name="table">The table.</param>
+		/// <param name="name">The name.</param>
+		/// <param name="defaultSchema">The defaultSchema.</param>
+		/// <returns></returns>
+		public virtual string GetIfNotExistsCreateConstraint(Table table, string name, string defaultSchema)
+		{
+			return "";
```
hazzik: Here you need to call `GetIfExistsDropConstraintEnd(table, name)`, for example

hazzik:
@lnu does it actually work for SQL2000? Shouldn't it be the same?

lnu:
There is no schema in SqlServer 2000. It was introduced in 2005. no?
How does it work if you use a schema with sql 2000?

hazzik:
@lnu true, it does not have a schema.

Also, the same issue exists with `defaultCatalog`, so, i think the code of `ForeignKey` and other constraints should be changed to something like this:

```csharp
public override string SqlDropString(Dialect.Dialect dialect, string defaultCatalog, string defaultSchema)
{
	var tableName = Table.GetQualifiedName(dialect, defaultCatalog, defaultSchema);

	var ifExists = dialect.GetIfExistsDropConstraint(tableName, Name);
	string drop = string.Format("alter table {0} {1}", tableName, dialect.GetDropForeignKeyConstraintString(Name));
	string end = dialect.GetIfExistsDropConstraintEnd(tableName, Name);
	return ifExists + System.Environment.NewLine + drop + System.Environment.NewLine + end;
}
```

lnu:
Ok...
I've checked multiple scenarios with catalog and schema and there are definitely more bugs than my initial one.

lnu:
@hazzik Does it look for you now?
Now everything is quoted correctly including the catalog (default or not). The catalog is also used when querying, which was not the case before.

src/NHibernate/Mapping/Constraint.cs
```diff
@@ -101,10 +101,10 @@ public virtual string SqlDropString(Dialect.Dialect dialect, string defaultCatal
...
 			if (IsGenerated(dialect))
 			{
-				string ifExists = dialect.GetIfExistsDropConstraint(Table, Name);
-				string drop =
-					string.Format("alter table {0} drop constraint {1}", Table.GetQualifiedName(dialect, defaultCatalog, defaultSchema), Name);
-				string end = dialect.GetIfExistsDropConstraintEnd(Table, Name);
+				var tableName = Table.GetQualifiedName(dialect, defaultCatalog, defaultSchema);
+				string ifExists = dialect.GetIfExistsDropConstraint(Table, Name, defaultCatalog, defaultSchema);
+				string drop = string.Format("alter table {0} drop constraint {1}", Table.GetQualifiedName(dialect, defaultCatalog, defaultSchema), Name);
+				string end = dialect.GetIfExistsDropConstraintEnd(Table, tableName, defaultCatalog, defaultSchema);
```
hazzik: Hi @lnu can we move all this to the Dialect? We really need to have one method, which will generate the constraint. Reason for that, is that MS SQL does support "if exist" statement natively. Also the `GetIfExistsDropConstraintEnd` method looks more like a workaround than a solution.

hazzik:
Looks good. I will review the PR shortly. Please ignore my last comment for now, as it is just a loud thinking.

hazzik:
NH-2288 tests are failing now

src/NHibernate/Cfg/Configuration.cs
```diff
@@ -2479,6 +2482,25 @@ private IEnumerable<IPersistentIdentifierGenerator> IterateGenerators(Dialect.Di
 			return generators.Values;
 		}
 
+		/// <summary>
+		/// Returns a quoted name(schema/catalog) if needed
+		/// </summary>
+		/// <param name="name">The schema name</param>
+		/// <param name="dialect">The instance of dicalect to use</param>
+		/// <returns></returns>
+		private string GetQuotedName(string name, Dialect.Dialect dialect)
```
hazzik: Is this necessary to quote everything?

src/NHibernate/Dialect/MsSql2005Dialect.cs
```diff
@@ -63,15 +63,22 @@ public override bool SupportsVariableLimit
...
 
-		protected override string GetSelectExistingObject(string name, Table table)
+		protected override string GetSelectExistingObject(string name, Table table, string defaultCatalog, string defaultSchema)
 		{
-			string schema = table.GetQuotedSchemaName(this);
+			// catalog defined on table takes precedence over defaultSchema
+			string catalog = table.Catalog == null ? defaultCatalog : table.GetQuotedCatalog(this);
+			if (catalog != null)
+			{
+				catalog += ".";
```
hazzik: Please use `StringHelper.Dot` instead of `"."`

src/NHibernate/Dialect/MsSql2005Dialect.cs
```diff
@@ -63,15 +63,22 @@ public override bool SupportsVariableLimit
...
+			}
+			// schema defined on table takes precedence over defaultSchema
+			string schema = table.Schema == null ? defaultSchema : table.GetQuotedSchema(this);
 			if (schema != null)
 			{
 				schema += ".";
```
hazzik: Please use `StringHelper.Dot` instead of `"."`

src/NHibernate/Mapping/ForeignKey.cs
```diff
@@ -86,10 +86,10 @@ public bool CascadeDeleteEnabled
 		/// </returns>
 		public override string SqlDropString(Dialect.Dialect dialect, string defaultCatalog, string defaultSchema)
 		{
-			string ifExists = dialect.GetIfExistsDropConstraint(Table, Name);
-			string drop = string.Format("alter table {0} {1}", Table.GetQualifiedName(dialect, defaultCatalog, defaultSchema),
-																	dialect.GetDropForeignKeyConstraintString(Name));
-			string end = dialect.GetIfExistsDropConstraintEnd(Table, Name);
+			var tableName = Table.GetQualifiedName(dialect, defaultCatalog, defaultSchema);
```
hazzik: `tableName` is not used here

src/NHibernate/Mapping/UniqueKey.cs
```diff
@@ -93,6 +93,7 @@ public override string SqlCreateString(Dialect.Dialect dialect, IMapping p, stri
 		/// </returns>
 		public override string SqlDropString(Dialect.Dialect dialect, string defaultCatalog, string defaultSchema)
 		{
+			var tableName = Table.GetQualifiedName(dialect, defaultCatalog, defaultSchema);
```
hazzik: `tableName` is not used here

src/NHibernate/Mapping/PrimaryKey.cs
```diff
@@ -75,12 +75,13 @@ public override string SqlConstraintString(Dialect.Dialect d, string constraintN
 		/// </returns>
 		public override string SqlDropString(Dialect.Dialect dialect, string defaultCatalog, string defaultSchema)
 		{
-			string ifExists = dialect.GetIfExistsDropConstraint(Table, Name);
+			var tableName = Table.GetQualifiedName(dialect, defaultCatalog, defaultSchema);
```
hazzik: `tableName` is not used here

src/NHibernate/Mapping/Table.cs
```diff
@@ -553,6 +570,42 @@ public string GetQuotedSchemaName(Dialect.Dialect dialect)
...
+		/// Gets the catalog for this table in quoted form if it is necessary.
+		/// </summary>
+		/// <param name="dialect">
+		/// The <see cref="Dialect.Dialect" /> that knows how to quote the table name.
+		/// </param>
+		/// <returns>
+		/// The catalog name for this table in a form that is safe to use inside
+		/// of a SQL statement. Quoted if it needs to be, not quoted if it does not need to be.
+		/// </returns>
+		public string GetQuotedCatalogName(Dialect.Dialect dialect)
```
hazzik: This method is not used anywhere

hazzik: Ok, let's keep it consistent. Please do not remove it.

src/NHibernate/Cfg/Configuration.cs
lnu: if the catalog/schema is not quoted and a reserved keyword is used it will fail. Don't you think?

src/NHibernate/Mapping/Table.cs
lnu: It's just a copy paste from the equivalent schema methods

src/NHibernate/Mapping/Constraint.cs
lnu: It could be part of a refactoring. I think we should just close this one with the current solution.

oskarb:
I kind of wish this had been done as a series of commits - I get the impression there are multiple issues and quoting of catalog doesn't really concern if schema is missing or not, does it? Not saying it's worth the effort to split it up now though.

I'm unsure what the repercussions of changing the signature of the overloads in the concrete dialects with regards to API stability. Any idea if that could be an issue?

Does Hibernate have the same bugs, or how have they been fixed there?

hazzik:
There are some divergences between NHibernate and Hibernate. This feature was heavily modified to enable it for SQL Server. Another issue, that we need to refactor the feature to enable Sql Server 2016 native `if exist` clause.

hazzik:
@lnu I did a quick test of the proposed changes. Let's assume I have following:

1. DBs (catalogues): `nhibernate` & `test`
2. table `Blog` with PK `PK__Blogs__2975AA28C884A705` in `nhibernate` catalogue in `dbo` schema

When I query following from within test catalogue:

```sql
select * from 
sys.objects where object_id = OBJECT_ID(N'nhibernate.dbo.PK__Blogs__2975AA28C884A705') 
AND parent_object_id = OBJECT_ID('nhibernate.dbo.Blogs')
```
I get nothing.

lnu:
I'll check again and add some unit tests. We are using those changes for one year now and never noticed any particular issues(sqlite, sqlserver and oracle)
Thanks for pointing this.

hazzik:
This issue is described in NH-1509. In short, the rule is: if catalog is specified.

hazzik:
@lnu can you pleas review my branch? https://github.com/hazzik/nhibernate-core/tree/NH-1285

lnu:
@hazzik At first look it seems ok. I did not want to change too many things in my previous commit but the new signature is cleaner. I'll run some tests today.

lnu:
@hazzik this pull request concerns this issue too:
https://github.com/nhibernate/nhibernate-core/pull/543/files

hazzik:
>One question, what do you mean by:
>>    When I query from nhibernate I get the same result regardles object_id is fully qualified or not.

Just ignore this. 

When query objects from different catalog it requires the catalog specification for `sys.objects` and `sysobjects` (SqlServer 2000) (so `catalog.sys.objects` and `catalog.dbo.sysobjects`)

lnu:
@hazzik I've incorporated your changes and squashed/rebased. Everything looks fine.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -723,15 +724,21 @@ public virtual string GetDropForeignKeyConstraintString(string constraintName)
...
 		/// <returns></returns>
+		[Obsolete("Can cause issues when a custom schema is defined(https://nhibernate.jira.com/browse/NH-1285). The new overload with the defaultSchema parameter should be used instead")]
 		public virtual string GetIfNotExistsCreateConstraint(Table table, string name)
 		{
-			return "";
+			var catalog = table.Catalog == null ? null : table.GetQuotedCatalog(this);
+			var schema = table.Schema == null ? null : table.GetQuotedSchema(this);
+			var tableName = table.GetQuotedName(this);
+
+			return GetIfNotExistsCreateConstraint(catalog, schema, tableName, name);
```
fredericDelaporte: Since the called method is public too, it will have to check if handling quoting is required too, for cases where it is called by some other code. So quoting here seems redundant. Same for next methods.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -1723,6 +1799,24 @@ public virtual string QuoteForSchemaName(string schemaName)
...
+		/// If the catalogName is already enclosed in the OpenQuote and CloseQuote then this 
+		/// method will return the catalogName that was passed in without going through any
+		/// Quoting process.  So if catalogName is passed in already Quoted make sure that 
+		/// you have escaped all of the chars according to your DataBase's specifications.
+		/// </p>
+		/// </remarks>
+		public virtual string QuoteForCatalogName(string catalogName)
+		{
+			return IsQuoted(catalogName) ? catalogName : Quote(catalogName);
+		}
```
fredericDelaporte: This looks like a general quoting implementation. I definitely will have to check with local sources what is going on with quoting.

src/NHibernate/Cfg/Configuration.cs
hazzik: It's kind-of correct behavior. On the Hibernate side they started conversion to "quotable identifier" value object for names (but did not finish, as usual)

hazzik:

lnu:
Done

fredericDelaporte:
Well thinking again about it, changes have been made specifically for SQL 2000 but without some dedicated tests on them. So it may be better to add a `MsSql2000DialectFixture` for checking them (especially the `if exists` with `sysobjects`).

That said, SQL 2000 is also quite an old thing, no more supported, not actually tested. So well, I will not mandate this fixture to be added.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -1829,6 +1926,66 @@ public virtual string[] UnQuote(string[] quoted)
 			return unquoted;
 		}
 
+		/// <summary>
+		/// Convert back-tilt quotes in a name for being used as an aliasname.
+		/// </summary>
+		/// <param name="aliasName">Name of the alias.</param>
+		/// <returns>A name with back-tilt quotes converted if any.</returns>
+		public virtual string ConvertQuotesForAliasName(string aliasName)
```
hazzik: Having X methods doing the same isn't necessary.

hazzik:
Could the unnecessary whitespace changes be reverted (especially in Configuration.cs)? 
Also if we want #956 then this would require more systematic approach.

src/NHibernate/Dialect/Dialect.cs
fredericDelaporte: This is done for consistency with the `QuoteFor` series, which does exactly the same. I think the idea is to allow a dialect to override some but not all of them with a different implementation, in case a database may have different quoting rules depending of the quoted object type.

If we consider this as not necessary, then we should obsolete the `QuoteFor` series by the way and keep only a general purpose `QuoteFor`.

We may improve the pattern by calling the corresponding `QuoteFor` from `ConvertQuotesFor`, in order to allow to override only one method instead of both. (But this will cause the convert to do the additional check done in quote for, which it does not need.)

fredericDelaporte:
> Could the unnecessary whitespace changes be reverted (especially in Configuration.cs)?

I have taken the time to review them all. I am favorable to let them.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -799,7 +818,67 @@ public virtual string GetIfExistsDropConstraint(Table table, string name)
...
+		/// The syntax that is used to check if a constraint does not exists before creating it
+		/// </summary>
+		/// <param name="catalog">The catalog.</param>
+		/// <param name="schema">The schema.</param>
+		/// <param name="table">The table.</param>
+		/// <param name="name">The name.</param>
+		/// <returns></returns>
+		public virtual string GetIfNotExistsCreateConstraint(string catalog, string schema, string table, string name)
+		{
+			return "";
```
fredericDelaporte: I realize a bit late that we have a breaking change here.
`virtual string GetIfNotExistsCreateConstraint(Table table, string name)` has been obsoleted in favor of this new `virtual string GetIfNotExistsCreateConstraint(string catalog, string schema, string table, string name)` method, which is now used instead of the old in NHibernate code base.

So dialects overriding the obsolete method will cease working as expected, because their override will now just be ignored. It means that such dialects will return an empty string for the new `GetIfNotExistsCreateConstraint`, since they do not override it. And so such dialects will no more be able to create constraints.

So when obsoleting overridable methods, we cause a breaking change by ceasing using them.

In the present case, since the new default implementation is indeed "no-op", we could try avoiding the breaking change by calling the old obsolete method instead. The old method should then, of course, be reverted to its old "no-op" behavior instead of calling the new implementation. So furthermore all its removed overrides would have to be put in place again for avoiding another breaking change.

But this does not stops here. If we do that, how should we then consider overrides of the new method? These overrides would then be somewhat breaking change too, since they would cause another dialect further down the inheritance chain and overriding itself the obsolete method to cease achieving its override...
But in the case of this PR, removing these new overrides cannot be done without undoing the PR.

So either we fully revert this PR to avoid this kind of breaking change, or we add this as possible breaking change in 5.1, or we bump version to 6.0 (and still add this as possible breaking change).

An additional conclusion to this is that obsoleting overridable methods is in most cases a breaking change. (It could be done without breaking changes only if the new method call the old one and is not overridden in an exposed descendant, unless the new override also call the old method.)

fredericDelaporte:
Possible breaking changes:

- Overridable methods have been obsoleted and their overrides will no more have effects. Overrides of their replacement must be added in the impacted deriving classes. Obsoleted overridable methods are:
  - `Dialect.GetIfNotExistsCreateConstraint(Table table, string name)`, replaced by `GetIfNotExistsCreateConstraint(string catalog, string schema, string table, string name)`.
  - `Dialect.GetIfNotExistsCreateConstraintEnd(Table table, string name)`, replaced by `GetIfNotExistsCreateConstraintEnd(string catalog, string schema, string table, string name)`.
  - `Dialect.GetIfExistsDropConstraint(Table table, string name)`, replaced by `GetIfExistsDropConstraint(string catalog, string schema, string table, string name)`.
  - `Dialect.GetIfExistsDropConstraintEnd(Table table, string name)`, replaced by `GetIfExistsDropConstraintEnd(string catalog, string schema, string table, string name)`.
  - `MsSql2000Dialect.GetSelectExistingObject(string name, Table table)`, replaced by `GetSelectExistingObject(string catalog, string schema, string table, string name)`.

PR #450, opened 2015-12-22
3.4.x

hailtondecastro:
Test case for NH-3079 on branch 3.4 before the fix


hailtondecastro:
Test case for NH-3079 on branch 3.4 before the fix

hailtondecastro:
Sorry, I opened and then closed accidentally

PR #452, opened 2015-12-22
Fix custom sql loader with composite id

hailtondecastro:
Test case and fix for [NH-3079](https://nhibernate.jira.com/browse/NH-3079) (#1117) on branch master


hailtondecastro:
Pull request posted on 
https://nhibernate.jira.com/browse/NH-3079?focusedCommentId=39554&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-39554

src/AsyncGenerator.yml
```diff
@@ -104,6 +104,9 @@
     - conversion: Ignore
       name: GetEnumerator
       containingTypeName: IFutureEnumerable
+    - conversion: Ignore
+      name: ComputeFlattenedParameters
+      containingTypeName: SqlQueryImpl
```
fredericDelaporte: Not ignoring it would cause an interface change by adding an async counterpart to a method belonging to an interface.

Moreover it generates an async counterpart due to a second level cache check, which would incurs IO only if a distributed cache is enabled.
There is also an `IType.GetPropertyValue` call which has an async counterpart, but it is called only on component types which currently have a fully synchronous implementation for it.

src/NHibernate.Test/NHSpecificTest/NH3079/Fixture.cs
```diff
@@ -0,0 +1,235 @@
...
+		[Test]
+		public void SqlPersonNoComponent()
+		{
+			using (var session = OpenSession())
+			{
+				var personList =
+					session
+						.GetNamedQuery("personNoComponentSql")
+						.SetParameter(0, new PersonNoComponent { IdA = 17, IdB = 18 })
+						.SetParameter(1, new PersonNoComponent { IdA = 19, IdB = 20 })
```
fredericDelaporte: The tests implying custom sql loaders only have one parameter. This test allows to check the change works also when multiple parameters needs flattening.

src/NHibernate/Async/Impl/AbstractQueryImpl.cs
```diff
@@ -20,6 +20,7 @@
 using NHibernate.Type;
 using NHibernate.Util;
 using System.Linq;
+using System.Text.RegularExpressions;
```
fredericDelaporte: Unrelated change due to a missing async regeneration in #1923.

src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -79,11 +79,11 @@ protected internal virtual void VerifyParameters()
...
 
 		/// <summary>
-		/// Perform parameter validation.  Used prior to executing the encapsulated query.
+		/// Perform parameters validation. Flatten them if needed. Used prior to executing the encapsulated query.
 		/// </summary>
 		/// <param name="reserveFirstParameter">
-		/// if true, the first ? will not be verified since
-		/// its needed for e.g. callable statements returning a out parameter
+		/// If true, the first positional parameter will not be verified since
+		/// its needed for e.g. callable statements returning an out parameter.
```
fredericDelaporte: Current code base always calls this method with `false`. The `true` path looks like dead code. And maybe it was always dead, if that can be useful only for callable statement, as I do not think callable statements have been ported.

src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -95,11 +95,15 @@ protected internal virtual void VerifyParameters(bool reserveFirstParameter)
 				throw new QueryException("Not all named parameters have been set: " + CollectionPrinter.ToString(missingParams), QueryString);
 			}
 
-			int positionalValueSpan = 0;
-			for (int i = 0; i < values.Count; i++)
+			var positionalValueSpan = 0;
+			// Values and Types may be overriden to yield refined parameters, check them
+			// instead of the fields.
+			var values = Values;
+			var types = Types;
```
fredericDelaporte: Trying to use here `ValueArray()` and `TypeArrays()` instead wrecks `CollectionFilterImpl` and `ExpressionFilterImp`, because they override these methods and cause them to yield values failing the validation. So I had to use the `Values` and `Types` properties instead.

src/NHibernate/Impl/SqlQueryImpl.cs
```diff
@@ -296,7 +299,75 @@ protected internal override void VerifyParameters()
...
+
+		protected override IList Values => _flattenedValues ??
+			throw new InvalidOperationException("Flattened parameters have not been computed");
 
+		protected override IList<IType> Types => _flattenedTypes ??
+			throw new InvalidOperationException("Flattened parameters have not been computed");
+
+		public override object[] ValueArray()
+		{
+			return _flattenedValues?.ToArray() ??
```
fredericDelaporte: It could also be `Values.Cast<object>().ToArray()`. In which case it could be moved in the base class instead.

fredericDelaporte:
I have rebased and refactored this.

src/NHibernate/Impl/SqlQueryImpl.cs
maca88: Added a TODO for 6.0 for changing `Values` type to `IList<object>`.

bahusoid:
> Simplified FlattenTypesAndValues to use a for statement and a single list for types/values.

Hm.. I believe I've been pushing something very similar some time ago. It seems those changes were lost by force push.. Let me check tomorrow If I also refactored something else.

fredericDelaporte:
Sorry if I have missed getting your own changes before resolving the conflicts this PR had.

PR #455, opened 2016-01-28
NH-3844 - Tests and fix

PleasantD:
I found that for some cases we need to flatten array index before we start rewriting the query for group bys and sometimes we need to do it after. Not sure if there is a case where both is required, but running the array flattener twice should have no ill effects on the query.

### DIFF
--- a/src/NHibernate.Test/Linq/ByMethod/GroupByTests.cs
+++ b/src/NHibernate.Test/Linq/ByMethod/GroupByTests.cs
@@ -715,6 +715,20 @@ namespace NHibernate.Test.Linq.ByMethod
 			Assert.AreEqual(2155, orderGroups.Sum(g => g.Count));
 		}
 
+		[Test(Description = "NH-3844")]
+		public void GroupByComputedValueFromNestedArraySelect()
+		{
+			var orderGroups = db.OrderLines.Select(o => new object[] { o }).GroupBy(x => new object[] { ((OrderLine)x[0]).Order.Customer == null ? 0 : 1 }).Select(g => new { Key = g.Key, Count = g.Count() }).ToList();
+			Assert.AreEqual(2155, orderGroups.Sum(g => g.Count));
+		}
+
+		[Test(Description = "NH-3844")]
+		public void GroupByComputedValueFromNestedObjectSelect()
+		{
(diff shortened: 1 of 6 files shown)


src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -27,6 +27,9 @@ public static ExpressionToHqlTranslationResults GenerateHqlQuery(QueryModel quer
 			// Remove unnecessary body operators
 			RemoveUnnecessaryBodyOperators.ReWrite(queryModel);
 
+			// Flatten array index access to query references
+			ArrayIndexExpressionFlattener.ReWrite(queryModel);
```
hazzik: Can we flatten only once?

PleasantD:
I don't think so.

In the case I'm trying to fix, the array create and access exist in the groupby key clause inside a conditional. The key nominator sees the new array expression and will only nominate from inside that, instead of nominating a larger conditional expression. Flattening array access early bypasses this problem. The array access would be flattened later anyway, but since only the inner part is nominated the conditional doesn't get promoted to the select correctly.

However, if I move the array flattening before the SubQueryFromClauseFlattener then NH-3800 reappears. In this case the group by key contains a QuerySourceReference with an array accessor on it. The early array flattening ignores it because its not an array access on a new array until after the SubQueryFromClauseFlattener re-writes it.

That being said... I can attempt to flatten array access during key nomination directly, if encountered.

PleasantD:
I updated the pull request :-)

PleasantD:
Any chance of this getting merged? And made part of 4.1.0?

Is there an official release timeline for 4.1.0? We've been using a local build for months now and it would be really nice to be able to update to an official release.

hazzik:
No timelines yet. I will do my best to do a release soon.

hazzik:
@PleasantD can you please rebase on top of master?

PleasantD:
@hazzik Done :-)

hazzik:
@PleasantD one more time, please

PleasantD:
Done again

PleasantD:
@hazzik Any chance this will be merged into master in the near future?
Any idea when the next official release will happen?

PR #456, opened 2016-02-14
Make OfType work better with polymorphic mappings

abrobston:
Addresses NH-3845.  The ProcessOfType Linq visitor class now handles interfaces and base classes by checking which mapped types that could be in the result set implement the class, then filtering accordingly.  If no results are possible, any remaining filters are removed from the query tree to avoid problems with invalid property names.

Because the HQL tree for OfType can now mimic the ".class IN" construct, the In method of HqlTreeBuilder was inadequate to express the tree.  Allowing an array of nodes to be passed as the source solves this problem.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3845/Concrete/MainEntity.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3845/Concrete/MainEntity.cs
@@ -1,0 +1,13 @@
+﻿using System.Collections.Generic;
+using NHibernate.Test.NHSpecificTest.NH3845.Interfaces;
+
+namespace NHibernate.Test.NHSpecificTest.NH3845.Concrete
+{
+	public class MainEntity : IMainEntity
+	{
+		public virtual int MainEntityId { get; set; }
+		public virtual string Text { get; set; }
+		public virtual ISet<IPropertyEntityBase> Properties { get; protected set; } = new HashSet<IPropertyEntityBase>();
+		public virtual ISet<ISeparateEntity> SeparateEntities { get; protected set; } = new HashSet<ISeparateEntity>();
+	}
+}
--- a/src/NHibernate/Hql/Ast/HqlTreeBuilder.cs
+++ b/src/NHibernate/Hql/Ast/HqlTreeBuilder.cs
@@ -416,7 +416,7 @@ namespace NHibernate.Hql.Ast
 			return new HqlFalse(_factory);
 		}
 
-		public HqlIn In(HqlExpression itemExpression, HqlTreeNode source)
+		public HqlIn In(HqlExpression itemExpression, params HqlTreeNode[] source)
 		{
 			return new HqlIn(_factory, itemExpression, source);
 		}
--- a/src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessOfType.cs
+++ b/src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessOfType.cs
@@ -1,5 +1,10 @@
-﻿using System.Linq.Expressions;
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
 using NHibernate.Hql.Ast;
+using NHibernate.Persister.Entity;
 using Remotion.Linq.Clauses.ResultOperators;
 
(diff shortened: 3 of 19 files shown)


hazzik:
I believe that you need to use ExpressionList instead of modifying In

abrobston:
I tried using ExpressionList, and it causes an exception at parse time.  When using the HQL construct `where EntityName.class in (Some.Class, Some.Other.Class)`, the parse tree places the two class names (as Ident nodes) as direct children of the InList node, which in turn is the child of the In node.  The changes in this commit replicate that behavior.

src/NHibernate.Test/NHSpecificTest/NH3845/Concrete/MainEntity.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
+using NHibernate.Test.NHSpecificTest.NH3845.Interfaces;
+
+namespace NHibernate.Test.NHSpecificTest.NH3845.Concrete
+{
+	public class MainEntity : IMainEntity
+	{
+		public virtual int MainEntityId { get; set; }
+		public virtual string Text { get; set; }
+		public virtual ISet<IPropertyEntityBase> Properties { get; protected set; } = new HashSet<IPropertyEntityBase>();
```
hazzik: C# 6 is not yet supported by our build engine, please change to using C#5

hazzik:
Ok, understand.
Please remove usage of C#6 language features

abrobston:
I found another facet of the same bug, and I'm likely to be amending the pull request once I can create a minimal test case for it.

hazzik:
You will also need to check the caching issues. Please check that query do
process correctly with regardless the type you specify.

abrobston:
@hazzik, I'm not entirely sure what you mean when referring to "check the caching issues."  Do you have a test case or example of some kind?  Or, can you tell me where to look for a similar unit test?

hazzik:
Start exploring from NH-2500 in JIRA. I can provide more resources later

abrobston:
I added two unit test methods to check for cache issues similar to those NH-2500 describes.  The tests pass.  If there are other scenarios I should check, please let me know.

abrobston:
I'm having some issues with another test case that I found does not pass.  I posted to the nhibernate-developers Google Group, so see discussion there once my post is released from moderation.

abrobston:
The [Google Group thread](https://groups.google.com/forum/#!topic/nhibernate-development/hzIw2esVHL8) has one reply, but more information would be very helpful.

hazzik:
Related to #553. The #553 contains more clear logic, which, I think shall be ported here.

hazzik:
Happy birthday, PR!

src/NHibernate/Hql/Ast/HqlTreeBuilder.cs
```diff
@@ -416,7 +416,7 @@ public HqlFalse False()
 			return new HqlFalse(_factory);
 		}
 
-		public HqlIn In(HqlExpression itemExpression, HqlTreeNode source)
+		public HqlIn In(HqlExpression itemExpression, params HqlTreeNode[] source)
```
hazzik: I've found the proper solution for not changing this. You need to use `HqlTreeBuilder.ExpressionSubTreeHolder`.

src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessOfType.cs
```diff
@@ -8,15 +11,112 @@ public class ProcessOfType : IResultOperatorProcessor<OfTypeResultOperator>
...
+			while (fromItemImplementorQueue.Any())
+			{
+				var currentImplementor = fromItemImplementorQueue.Dequeue();
+				if (fromItemTypeNamesProcessed.Add(currentImplementor))
+				{
+					var persister = queryModelVisitor.VisitorParameters.SessionFactory.TryGetEntityPersister(currentImplementor);
+					if (persister != null)
+					{
+						fromItemTypePersisters.Add(persister);
+						persister.EntityMetamodel.SubclassEntityNames.ToList().ForEach(fromItemImplementorQueue.Enqueue);
```
hazzik: Please do not use `.ForEach` use the proper loop.

hazzik:
I've extracted the issue with unmapped base classes and interfaces to a new feature request NH-3947.

hazzik:

PR #457, opened 2016-02-17
NH3820

osoykan:
- Added new LockMode -> Nolock
- Added join hints like merge,loop, hash


oskarb:
Hi,

If you don't mind, it would be great if you could rework the commits (rebase) to remove the mistaken commits etc. We don't want those to end up in the official repository.

osoykan:
I see, i'm going to  do this.

osoykan:
@oskarb I created new pull request. Thanks for warning.

PR #459, opened 2016-02-20
NH-802 - Investigate possible use of MSBuild to build the project

jrauber:
Use MsBuild instead of csc directly


hazzik:
Thanks. I'll use your changes to make a new pull request.

PR #460, opened 2016-02-23
Fix Criteria caching filtered collections

zalewskil:
When a cacheable criteria, or query-over, eager-loads a collection with restrictions, causing the collection to be partially loaded, the collection is nonetheless put in the second level cache. This causes lazy-loads of the collection done by other sessions to retrieve an invalid state for the collection.

This is a partial fix of #1341, for QueryOver and Criteria.


hazzik:

/cc @itmagination @zalewskil

zalewskil:
@hazzik done :)

hazzik:
Thanks

src/NHibernate.Test/NHSpecificTest/NH3848/CriteriaTestFixture.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        protected override IList<Customer> GetCustomersWithOrdersEagerLoaded(ISession session)
+        {
+            return session.CreateCriteria<Customer>()
+					      .Fetch("Orders")
+						  .SetResultTransformer(new DistinctRootEntityResultTransformer())
+                          .List<Customer>();
+        }
+
+		protected override async Task<IList<Customer>> GetCustomersWithOrdersEagerLoadedAsync(ISession session)
```
fredericDelaporte: In most cases, no async code should be written. That is the job of the async generator, included for tests. See [contributing](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md). Is there any special reason here for hand coding them?

By the way, indentation is mixed between spaces and tabs in this file. It should be re-formated to tabs, as instructed in [contributing](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md#submit-a-pull-request).

(And other .cs files fully spaces indented added by this PR should be switched to tabs.)

fredericDelaporte:
As this is a partial fix of #1341, I have re-tagged it as a stand-alone issue and I have change the description for avoiding GitHub auto-closing #1341.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
```diff
@@ -152,6 +154,14 @@ public QueryParameters GetQueryParameters()
 				{
 					lockModes[GetSQLAlias(subcriteria)] = lm;
 				}
+
+				if (subcriteria.HasRestrictions && subcriteria.JoinType == JoinType.LeftOuterJoin)
+				{
+					var collectionName = $"{rootCriteria.EntityOrClassName}.{subcriteria.Path}";
```
bahusoid: This place looks hacky to me:
- We already have place where we collect all query collection persisters (see `CreateCriteriaCollectionPersisters`). I think it's a good place to collect this skip cache collection flags. And maybe collect them similarly to`criteriaCollectionPersisters' in `uncacheableCollectionPersisters`
- We should not rely on Left join anymore as now we support fetching collections with inner join too (with `SelectMode.Fetch`).

src/NHibernate/Engine/QueryParameters.cs
```diff
@@ -139,6 +157,11 @@ public bool HasRowSelection
 
 		public bool Callable { get; set; }
 
+		/// <summary>
+		/// Indicates if we can add loaded child collections to second level cache.
+		/// </summary>
+		public HashSet<string> UncacheableCollections { get; set; }
+
```
bahusoid: Why can't we store this information right inside `Loader`? Looks like less changes to me (no need to modify `InitializeEntitiesAndCollections`)

src/NHibernate/Engine/Loading/CollectionLoadContext.cs
```diff
@@ -145,7 +145,20 @@ public IPersistentCollection GetLoadingCollection(ICollectionPersister persister
...
+		}
+
+		/// <summary> 
+		/// Finish the process of collection-loading for this bound result set.  Mainly this
+		/// involves cleaning up resources and notifying the collections that loading is
+		/// complete. 
+		/// </summary>
+		/// <param name="persister">The persister for which to complete loading. </param>
+		/// <param name="uncacheableCollections">Indicates if collections can be put in cache</param>
+		public void EndLoadingCollections(ICollectionPersister persister, HashSet<string> uncacheableCollections)
```
bahusoid: I think here we can simplify signature to provide bool flag instead of HashSet:
`public void EndLoadingCollections(ICollectionPersister persister, bool skipCache)`

fredericDelaporte: It seems initially it was only a `bool` all the way long, but on the JIRA:

>**Alexander Zaytsev** added a comment — <time datetime="2017-04-07T01:37:44.330Z" >title="2017-04-07T01:37:44.330Z">7th April 2017, 1:37:44</time>:
>> <~Lukasz> your pull request(s) do not consider 2 separate fetches. When there is a collection *with* restrictions and another one *without* then I expect that the collection *without* restrictions would be cached.
>> 
>> So, the property "CanAddCollectionsToCache" shall probably be a dictionary.
>> 

But well, I am not valuing much handling this case, because two collection fetches causes a Cartesian product, so in most cases it should be avoided.

bahusoid: @fredericDelaporte I'm not sure it's related to this particular place. From what I see here we provide exact collection `persister` to fetch.  So for multiple fetched collections it's multiple calls and it's up to caller to calculate `skipCache` flag for this particular `persister`. So with my suggestion it's still possible to mix cached and uncached collections.

To be more precise I think we can have in `Loader` optional array 
`protected virtual bool[] UncachedCollectionPersisters`

And check it something like:
`EndCollectionLoad(resultSetId, session, collectionPersisters[i], skipCache: UncachedCollectionPersisters?[i] == true);`

src/NHibernate/Engine/Loading/CollectionLoadContext.cs
```diff
@@ -145,7 +145,20 @@ public IPersistentCollection GetLoadingCollection(ICollectionPersister persister
 		/// complete. 
 		/// </summary>
 		/// <param name="persister">The persister for which to complete loading. </param>
+		[Obsolete("Please use EndLoadingCollections(ICollectionPersister, HashSet<string>) instead.")]
```
fredericDelaporte: On a method with a generated async counterpart, specifying the exact signature causes the message to be inadequate. (I will a push a commit addressing this point.)

fredericDelaporte:
Oracle was still failing for another reason. I have solved this locally and tested it with my local Oracle instance. I have also pushed some minor adjustments. (They do not take into account @bahusoid review.)

src/NHibernate.Test/NHSpecificTest/NH3848/Fixture.cs
```diff
@@ -0,0 +1,392 @@
...
+			var rootEntityTypeFullPath = typeof(T).FullName;
+			var memberExpression = pathToCollection.Body as MemberExpression;
+			if (memberExpression == null)
+				throw new ArgumentException("pathToCollection should be member expression");
+
+			var role = $"{rootEntityTypeFullPath}.{memberExpression.Member.Name}";
+			Sfi.EvictCollection(role);
+		}
+
+		protected abstract IList<Customer> GetCustomersWithFetchedOrdersWithoutRestrictions(ISession session);
```
fredericDelaporte: This method and its overrides are never used.

fredericDelaporte:
Since first commits are quite old, I have prepared a rebase locally. Tell me if it is ok for you if I force push it. (Of course if some more commits are added here before my push, I will pick them on my rebase before pushing.)

fredericDelaporte:
Rebase pushed.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
IwanowM: It seems to me, `ShouldFetchCollectionPersister()` is exactly what we need to determine whether collection will be fetched or not. 
From other side, we can map `HasRestrictions` flags from subcriteries to assotiations, to determine if persister should be cached or not. 
That is why I have moved initialization of `uncacheableCollectionPersisters` to `CriteriaJoinWalker`. 
Is it ok for you?

bahusoid: By "hacky" I meant we shouldn't use "magic strings" like `$"{rootCriteria.EntityOrClassName}.{subcriteria.Path}"`.  Will it work for multilevel fetch like `obj.Child.Collection` (it seems we need a test for this case, maybe it will work but it's not obvious to me)?

So instead I think it's better to collect directly all collection persisters `ICollectionPersister` with restrictions. That's why I suggested to look in to `CreateCriteriaCollectionPersisters` as that's the place where actual collection persisters are enumerated (though I'm not saying it's the only and best place to do the job)

To me this task looks like the following:
1) In translator (or walker) we need to find which collection persisters have restrictions;
2) And for loader we need to map "restricted" collection perstisters with persisters in `collectionPerstisters` and prepare `UncachedCollectionPersisters`

So to me it looks like we don't need to bother  "to determine whether collection will be fetched or not" - as long as we can obtain 'ICollectionPersister` for given criteria. And I don't see why `OuterJoinableAssociation` needs to be modified - it seems all information can be obtained from translator/walker.

IwanowM: Ok, if I understood well, we shouldn't add collection into cache if:
1) There are restrictions and `subcriteria.JoinType == JoinType.LeftOuterJoin`
2) There are restrictions and `subcriteria.JoinType == JoinType.InnerJoin` and  `SelectMode == SelectMode.Fetch`, right?

And it wasn't clear for me how to get selection mode for subcriteria, but it seems that I can get it in this way - `rootCriteria.GetSelectMode(subcriteria.Path)`, it isn't?

IwanowM: Could you have a look at my last commit, please? Is it right direction?

hazzik:
I want to do another refactoring.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
bahusoid: Yep. But as I already said I think that logic that checks `Join` types and `SelectMode` is unnecessary here. No need to duplicate code - those checks will be executed by the routine that prepares collection persisters to fetch.
 
I've refactored and made a few clean ups to it (see my latest [commit](https://github.com/nhibernate/nhibernate-core/pull/460/commits/c25658e19988ea5178d82b82b8ef99c52fed5d3d)). Let me know if you disagree or see any issues.

bahusoid:
@hazzik Ok. No more commits from my side for today.

src/NHibernate/Loader/Criteria/CriteriaLoader.cs
```diff
@@ -46,6 +50,10 @@ public partial class CriteriaLoader : OuterJoinLoader
 
 			InitFromWalker(walker);
 
+			if (translator.UncacheableCriteriaCollectionPersisters.Any())
```
hazzik: I don't quite agree with this check because this obfuscates things by making stuff dependant on a index. Although it looks as an optimization it produces a bigger array.

bahusoid: Actually I didn't mean it as optimization but merely to make it processed the same way as others properties (see  `CollectionAliases`, 'CollectionOwners', `CollectionPersisters`,...).  And that's collection fetches -  it seems big arrays should not be an issue.

hazzik:
@bahusoid I've done with my changes

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
```diff
@@ -41,10 +41,11 @@ public class EntityJoinInfo
 		private readonly IDictionary<String, ICriteriaInfoProvider> nameCriteriaInfoMap =
 			new Dictionary<string, ICriteriaInfoProvider>();
 
+		private readonly HashSet<ICollectionPersister> uncacheableCollectionPersisters = new HashSet<ICollectionPersister>();
 		private readonly ISet<ICollectionPersister> criteriaCollectionPersisters = new HashSet<ICollectionPersister>();
 		private readonly IDictionary<ICriteria, string> criteriaSQLAliasMap = new Dictionary<ICriteria, string>();
 		private readonly IDictionary<string, ICriteria> aliasCriteriaMap = new Dictionary<string, ICriteria>();
-		private readonly IDictionary<string, ICriteria> associationPathCriteriaMap = new LinkedHashMap<string, ICriteria>();
+		private readonly Dictionary<string, CriteriaImpl.Subcriteria> associationPathCriteriaMap = new Dictionary<string, CriteriaImpl.Subcriteria>();
```
fredericDelaporte: `LinkedHashMap` guarantees the dictionary elements will be iterated in the same order they were added. This is not the case of `Dictionary`, which, with current implementations, will have a different iteration order as soon as some elements were removed.

So this change is not neutral. But it looks like preserving the order has no purpose here, so this is ok. (Similarly `associationPathJoinTypesMap` does not need to be a `LinkedHashMap`.)

PR #461, opened 2016-02-23
NH-3046 Potential Memory leak

Tikuna:
Found a potential kind of memory leak.

Memory used by entities is not released when Session.Clean is called. This is due a reference that is kept from delegates used in before/after commit. 

This delegates are necessary in case we are using cache's but totally unnecessary in other escenarios.

The proposed solution is simple, is just not create this delegates when not needed. 

In our environment, we needed a long running transaction... we discover the issue when the process took 8Gb of memmory and 4 hours of processing. After the change, it took 200mb and 15 minutes...

We add a test case where we can detect the issue when delegates are always created.

https://nhibernate.jira.com/browse/NH-3046

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3046/Model.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3046/Model.cs
@@ -1,0 +1,21 @@
+﻿namespace NHibernate.Test.NHSpecificTest.NH3010
+{
+	using System;
+	using System.Collections.Generic;
+
+	public class Parent
+	{
+		public Parent()
+		{
+			Childs = new List<Child>();
+		}
+
+		public virtual Guid Id { get; set; }
+		public virtual IList<Child> Childs { get; set; }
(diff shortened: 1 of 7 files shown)


src/NHibernate.Test/NHSpecificTest/NH3046/Model.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace NHibernate.Test.NHSpecificTest.NH3010
```
hazzik: Wrong namespace here, it fails the build

Tikuna: Sorry, made a mistake with the files (I'm new on Git).

Will upload the changes soon.

Tikuna:
Fixed !

hazzik:
@Tikuna thanks. I'm going to close this pull request and open a clean one

PR #462, opened 2016-02-24
NH-802 - use msbuild instead of csc/vbc

hazzik:


jrauber:
Thanks for improving it and pushing it really forward. I failed to fix the breaking nunit-runner and wasted time on different possibilities to ilrepack the assemblies.

PR #463, opened 2016-02-24
NH-3848 - Tests and fix for HQL

### LINKED ISSUE #1341: NH-3848 - Child collection fetched using left outer join with on clause or where clause restrictions on fetched collection shouldn’t be stored in second level cache.
**Lukasz** created an [issue](https://nhibernate.jira.com/browse/NH-3848) — <time datetime="2016-02-22T11:36:30.196Z" title="2016-02-22T11:36:30.196Z">22nd February 2016, 11:36:30</time>:
> Reproductions Steps:
> 1.	Enable second level cache in nh configuration.
> 2.	Create entity with collection.
> 3.	Enable second level cache on collection (on collection entity also ofc).
> 4.	Open session.
> 5.	Create QueryOver (Criteria or HQL) which fetch root entities with children collections using left outer join with some restrictions on collection. 
> E.g.
> 
> ```csharp
>  firstSession.QueryOver<Customer>()
>     .JoinAlias(n => n.Orders,
>                () => ordersAlias,
>                JoinType.LeftOuterJoin,
>                Restrictions.Eq("Number", orderNumber))
>     .List();
> ```
> 
> 6.	Open another session.
> 7.	Create QueryOver on second session for all root entities. E.g. 
> 
> ```csharp
> secondSession.QueryOver<Customer>().List();
> ```
> 
> 8.	Iterate over entities fetched in point 7. and fetch child collections.
> 9.	**All collection fetched in point 8. Will be filtered like in point 5.**
> 
> 
---
**Lukasz** added a comment — <time datetime="2016-02-23T13:08:49.516Z" title="2016-02-23T13:08:49.516Z">23rd February 2016, 13:08:49</time>:
> Pull request with tests and fix for Criteria and QueryOver: https://github.com/nhibernate/nhibernate-core/pull/460
> 
---
**Lukasz** added a comment — <time datetime="2016-02-24T14:01:27.897Z" title="2016-02-24T14:01:27.897Z">24th February 2016, 14:01:27</time>:
> Pull request with tests and fix for HQL: https://github.com/nhibernate/nhibernate-core/pull/463
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2017-04-07T01:37:44.330Z" title="2017-04-07T01:37:44.330Z">7th April 2017, 1:37:44</time>:
> <~Lukasz> your pull request(s) do not consider 2 separate fetches. When there is a collection *with* restrictions and another one *without* then I expect that the collection *without* restrictions would be cached.
> 
> So, the property "CanAddCollectionsToCache" shall probably be a dictionary.
>

zalewskil:
This pull should be treated as a hint for fix or tests rather than actual fix (but maybe I'm wrong).
It can harm performance (traversing Where clause tree or using String.StartsWith). 
Actually we aren't using HQL in our application so this fix had to be done quick.

[NH-3848](https://nhibernate.jira.com/browse/NH-3848).

Fixes #1341


hazzik:
Rebased to resolve conflicts

hazzik:
Tests need to be adapted from the base fixture on the master branch.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -283,9 +284,29 @@ public bool ContainsCollectionFetches
...
 			get
 			{
-				ErrorIfDML();
-				IList<IASTNode> collectionFetches = ((QueryNode)_sqlAst).FromClause.GetCollectionFetches();
-				return collectionFetches != null && collectionFetches.Count > 0;
+				return CollectionFetches.Count > 0;
+			}
+		}
+
+		public bool CanAddFetchedCollectionToCache
```
hazzik: This probably needs to be moved to a loader, however, I'm not entirely sure.

fredericDelaporte: And shouldn't it be `CanAddFetchedCollectionsToCache`? (Collection is switched to plural.)

If I understand well this PR, contrary to #460, it would disable collection caching for the whole query, not just for the one having restrictions.

> This probably needs to be moved to a loader, however, I'm not entirely sure.

I am not sure either. It may require to expose some internal state of the translator for the loader to analyse it, and then, better keep it in the translator.

src/NHibernate.Test/NHSpecificTest/NH3848/HqlTestFixture.cs
```diff
@@ -0,0 +1,76 @@
...
+		{
+			throw new NotImplementedException();
+		}
+
+		protected override IList<Customer> GetCustomersByOrderNumberUsingWhereClause(ISession session, int orderNumber)
+		{
+			var numberParameterName = "number";
+			var query = session.CreateQuery(string.Format("from Customer as c " +
+														   "left join fetch c.Orders as o " +
+														   "where o.Number = :{0}", numberParameterName));
```
fredericDelaporte: A formatting like
```c#
var query =
	session.CreateQuery(
		string.Format(
			"from Customer as c " +
			"left join fetch c.Orders as o " +
			"where o.Number = :{0}",
			numberParameterName));
```
Would be more readable.

But in fact I would rather write it as:
```c#
var query =
	session.CreateQuery(
		"from Customer as c " +
		"left join fetch c.Orders as o " +
		$"where o.Number = :{numberParameterName}");
```

Or even:		
```c#
var query =
	session.CreateQuery(
		"from Customer as c " +
		"left join fetch c.Orders as o " +
		"where o.Number = :number");
```
To keep it simple.

And do the same for following queries.

src/NHibernate.Test/NHSpecificTest/NH3848/TestFixture.cs
```diff
@@ -0,0 +1,266 @@
...
+using System.Linq.Expressions;
+using NHibernate.Cache;
+using NHibernate.Cfg;
+using NHibernate.Cfg.MappingSchema;
+using NHibernate.Mapping.ByCode;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3848
+{
+	public abstract class TestFixture : TestCaseMappingByCode
```
fredericDelaporte: Shouldn't this one be dropped in favour of the `Fixture` class added by #460?

Ok, I guess this what you were referring to with your comment:

> Tests need to be adapted from the base fixture on the master branch.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -283,9 +284,29 @@ public bool ContainsCollectionFetches
...
+		{
+			get
+			{
+				foreach (var collectionFetch in CollectionFetches)
```
fredericDelaporte: Before iterating the collections, it would be nice if the overall cacheability of the query was checked first. (But maybe that check should be done in the loader.)

src/NHibernate.Test/NHSpecificTest/NH3848/TestFixture.cs
```diff
@@ -0,0 +1,266 @@
...
+
+				session.Flush();
+				transaction.Commit();
+			}
+		}
+
+		[Test]
+		public virtual void ChildCollectionsFromLeftOuterJoinWithOnClauseRestrictionOnCollectionShouldNotBeInSecondLevelCache()
+		{
+			var firstSession = OpenSession();
```
fredericDelaporte: I have left this pattern go into master for the `Fixture`, I shouldn't had. With some overrides throwing an ignore in this PR, this causes sessions to remain opened, triggering a test clean-up failure. `using` needs to be used.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -430,6 +451,50 @@ private void ErrorIfDML()
 				throw new QueryExecutionRequestException("Not supported for DML operations", _queryIdentifier);
 			}
 		}
+
+		private bool ContainsRestrictionOnTable(string tableAlias)
```
bahusoid: Can't this method be simplified to find child node with 
`child.FromElement == fromElement`

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
fredericDelaporte: Checking the global cacheability of the query does not seem doable where this property is used, because it requires the queryParameters, which are not available there.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
bahusoid: Oups. Sorry - I made a quick debug and didn't actually check if it's really interface method. It seems casting to `FromReferenceNode' is required. See https://github.com/nhibernate/nhibernate-core/pull/463/commits/2f66408bafe95c9a45e7c9d6137912c5a713ed95 to give the idea. As I know nothing about HQL code -  I'm not sure it's the best way to traverse through children elements.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -292,17 +292,14 @@ public bool CanAddFetchedCollectionToCache
 		{
 			get
 			{
-				foreach (var collectionFetch in CollectionFetches)
+				foreach (FromElement fromElement in CollectionFetches)
```
fredericDelaporte: If it is guaranteed that they are all always `FromElement` instances like above code implies, it would be better to retype `CollectionFetches` and do the casting in its computing.

After analyzing the code, indeed, it can only be `FromElement` items.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -280,30 +281,25 @@ public string[] ReturnAliases
...
-			}
-		}
+		public bool ContainsCollectionFetches => CollectionFetches.Count > 0;
 
-		public bool CanAddFetchedCollectionToCache
+		public ISet<ICollectionPersister> UncacheableCollectionPersisters
 		{
 			get
 			{
+				var result = new HashSet<ICollectionPersister>();
```
bahusoid: It seems `UncacheableCollectionPersisters` and `ContainsRestrictionOnTable` can be merged:
```
//Find all restrictions in Where clause for cached fetched collections
return new HashSet<ICollectionPersister>(
	ASTUtil.IterateChildrenOfType<FromReferenceNode>(
		((QueryNode) _sqlAst).WhereClause,
		skipSearchInChildrenWhen: node => node.FromElement != null)
	.Select(rn => rn.FromElement)
	.Where(fr => fr?.IsFetch == true && fr.QueryableCollection?.HasCache == true)
	.Select(fr => fr.QueryableCollection)
	);
```

hazzik: Ok, I've applied your suggestion (it got me some time to understand the logic).

src/NHibernate/Hql/Ast/ANTLR/Util/ASTUtil.cs
```diff
@@ -103,5 +103,28 @@ public static IList<IASTNode> CollectChildren(IASTNode root, FilterPredicate pre
...
+		/// Iterates over all children and sub-children and finds elements of required type.
+		/// </summary>
+		internal static IEnumerable<TRequiredType> IterateChildrenOfType<TRequiredType>(IASTNode root, Func<TRequiredType, bool> skipSearchInChildrenWhen)
+		{
+			foreach (var child in root)
+			{
+				var searchInChildren = true;
+				if (child is TRequiredType typedChild)
+				{
+					searchInChildren = !skipSearchInChildrenWhen(typedChild);
```
hazzik: Why do we need this skip logic? Can we just iterate to the end?

hazzik: Ok

bahusoid:
But it seems to fully fix NH-3848 filter case (when collection is filtered by `session.EnableFilter`) is also needs to be handled (both in criteria and hql)

fredericDelaporte:
> But it seems to fully fix NH-3848 filter case (when collection is filtered by `session.EnableFilter`) is also needs to be handled (both in criteria and hql).

Let's call it another bug, if confirmed.

PR #467, opened 2016-03-07
Nh 3771 - Batch Update with Optimistic Locking control.

montagutcarlos:
https://nhibernate.jira.com/browse/NH-3771

This is a suggestion for batch update when optimistic control. Now, when
version control is active, batch update is not done, see the fixture (3771).
This is a down side in performance.

We made the change, and run the tests. The diference is that
StaleStateException is thrown instead of StaleObjectStateException, in
which case we loose the information regarding the entity and it's id
that comes with the exception. In our application, this is not a big problem, compare with the
possibility of doing a batch operation instead a single one.

We have been looking at the source code, and will be a possibility to
still get a class and a group of id's on the batcher to generate a more
accurate exception.

Please, let me know if this will be of intereset, so i can go deeper in
that aproach, or if you see another aproach that you will suggest.


oskarb:
The 7290c0da0e163d780511bbfd2bc6e6101de60015 commit should not really be a port of this should it? At least, that change appears to already be on master under a different commit id. Rebase?

hazzik:
Hi @Tikuna, can you please enable [Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) feature?

Tikuna:
Hi @hazzik . Done it. Sorry for the delay with the request of @oskarb . I'm pretty new to git and take me lot of time to understand what exactly should I done.

hazzik:
I would prefer this to be implemented as "isJdbcBatchVersionedData" (the better name is welcome).

hazzik:
Replaced by #606

PR #446, opened 2015-10-02
Projection performance improvements [NH-3486]

JesperD60:
https://nhibernate.jira.com/browse/NH-3486

I have implemented the changes suggested by reporter.

PR #468, opened 2016-03-09
NH-3486 - Performance: For projections, identical metadata for rows is recalculated for every row

lnu:
related to https://github.com/nhibernate/nhibernate-core/pull/446
Changes done in CriteriaLoader instead of CriteriaQueryTranslator.

### DIFF
--- a/src/NHibernate/Loader/Criteria/CriteriaLoader.cs
+++ b/src/NHibernate/Loader/Criteria/CriteriaLoader.cs
@@ -30,6 +30,8 @@ namespace NHibernate.Loader.Criteria
 		private readonly string[] userAliases;
 		private readonly bool[] includeInResultRow;
 		private readonly int resultRowLength;
+		// caching NH-3486
+		private readonly string[] cachedProjectedColumnAliases;
 
 		public CriteriaLoader(IOuterJoinLoadable persister, ISessionFactoryImplementor factory, CriteriaImpl rootCriteria,
 							  string rootEntityName, IDictionary<string, IFilter> enabledFilters)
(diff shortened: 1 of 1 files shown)


src/NHibernate/Loader/Criteria/CriteriaLoader.cs
```diff
@@ -48,6 +51,8 @@ public class CriteriaLoader : OuterJoinLoader
 			resultTypes = walker.ResultTypes;
 			includeInResultRow = walker.IncludeInResultRow;
 			resultRowLength = ArrayHelper.CountTrue(IncludeInResultRow);
+			cachedProjectedTypes = translator.ProjectedTypes;
```
hazzik: this will be the same as `resultTypes`

hazzik:
I get a lot of test failing

lnu:
Ok i'll check it but I ran everything locally and it seemed ok. 
thanks

hazzik:
@lnu you need `if(translator.HasProjection)` around variable assignments

lnu:
I've just run it again and you're right :+1:

src/NHibernate/Loader/Criteria/CriteriaLoader.cs
lnu: @hazzik I can safely use resultTypes instead of cachedProjectedTypes?

hazzik: @lnu yes, it just set in walker using the same `translator.HasProjection` condition

PR #469, opened 2016-04-06
Fix false positive warning about non-serializable interfaces

EvK:
See this issue in NHibernate.Spatial: https://github.com/nhibernate/NHibernate.Spatial/issues/9.
Also see this discussion on StackOverflow: http://stackoverflow.com/q/35983333/5311735.

All geometry types return IGeometry interface as their ReturnedType, and they cannot do otherwise, because there are several concrete types which implement this interface. This produces a false positive warning about IGeometry type being not serializable, which floods the logs. It seems there is no need to produce non-serializable warning for interface and abstract types, because they cannot be serializable anyway.

### DIFF
--- a/src/NHibernate/Type/CompositeCustomType.cs
+++ b/src/NHibernate/Type/CompositeCustomType.cs
@@ -49,7 +49,7 @@ namespace NHibernate.Type
 				throw new MappingException(name + " must implement NHibernate.UserTypes.ICompositeUserType", ice);
 			}
 			TypeFactory.InjectParameters(userType, parameters);
-			if (!userType.ReturnedClass.IsSerializable)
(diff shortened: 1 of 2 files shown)


hazzik:
@EvK I can not accept pull requests which do not have tests

EvK:
I can create a test for it, but I'm a bit confused about test name. As I understand, test should be put under NHSpecificTest folder with the name of JIRA entry. Does that mean I should first create an issue in JIRA, or I can somehow reference this github issue?

hazzik:
Yes, it does mean that you need to create a JIRA issue first

EvK:
Root issue in NHibernate.Spatial was resolved in another way.

PR #470, opened 2016-04-08
Additional check if child session has already been closed.

itmagination:
This simple fix prevents unnecessary `SessionException("Session was already closed")` throw when both parent session and child session are closed separately (order does not matter). Before invoking close on child session (during parent session close) we should check if child session has not been closed already. Unfortunately to check if session was closed we have to use `SessionImpl` class instead of `ISession` interface, so we changed `SessionImpl`’s `childSessionsByEntityMode` private field type `IDictionary<EntityMode, ISession>` into `IDictionary<EntityMode, SessionImpl>`.

### DIFF
--- a/src/NHibernate/Engine/ISessionImplementor.cs
+++ b/src/NHibernate/Engine/ISessionImplementor.cs
@@ -306,7 +306,12 @@ namespace NHibernate.Engine
 		/// </returns>
 		bool IsClosed { get; }
 
-		void Flush();
+        /// <summary>
+        /// Close the session and release all resources.
+        /// </summary>
+        IDbConnection Close();
+
+        void Flush();
 
 		/// <summary> 
 		/// Does this <tt>Session</tt> have an active Hibernate transaction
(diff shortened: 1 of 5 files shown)


hazzik:
Please use `ISessionImplementor` instead of SessionImpl

itmagination:
I used `ISessionImplementor` instead of `SessionImpl` as requested. Unfortunately I had to add `Close` method in `ISessionImplementor` interface and make other necessary changes in dependent types (add abstract `Close` method in `AbstractSessionImpl` etc.). In `GetSession` method of `SessionImpl` class I had to cast result to `ISession` interface, because there is no formal connection between `ISession` and `ISessionImplementator` interfaces.

hazzik:
Hi @itmagination.

There is `ISession.IsOpen` property which implementation is exactly `!IsClosed`, so, this change shall be only one line:

```
if (pair.Value.IsOpen)
{
     pair.Value.Close();
}
```

No other changes are required.

itmagination:
Oh my! You're right! I've closed this knotty PR and opened a new one #542 containing this simple fix. Thank you!

PR #471, opened 2016-04-15
Fix discriminator in where clause on outer joins

ithielnor:
This fix requires a change to the `IJoinable.FilterFragment` method. A third parameter is added for `excludeDiscriminator`. When `true`, implementations of this method should not include any entity discriminator in the fragment returned to the caller.

In general, the argument should be true when called for a joined table fragment, and false for a root table fragment.

This is partially related to NH-3506


hazzik:
This seems to be working, but it feels to be wrong. Instead of excluding the discriminator, you effectively disable filters when you pass "false"

hazzik:
I reopen #268 instead of this.

PR #473, opened 2016-05-02
NH-3414 - Fixes ORDER BY HQL grammar

PleasantD:
The HqlSqlWalker grammar was not supporting more than a single query in the ORDER clause.
It would handle multiple expressions because it was looking for them with a recursive lookup, but the parsing of a query was only being handled by the master ORDER clause.

So a single query would work, or any number of expression. But combine an expression with a query or multiple queries and it would fail.

I moved the query check into the inner part of the ORDER grammar so that it becomes part of the recursive parsing, like expressions.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3414/Entity.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3414/Entity.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+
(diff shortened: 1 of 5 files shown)


hazzik:
@PleasantD new tests are failing for Firebird with "Dynamic SQL Error" in SetUP. Can you please take a look?

PleasantD:
Fixed and squashed. Firebird didn't like that the column name was "Value".

hazzik:
@PleasantD rebase?

PleasantD:
Done

PR #474, opened 2016-05-18
Added compatibility for db2 schema update.

fabiankoebel:
This pull request contains the compatibility for schema updates for db2 database systems.

### DIFF
--- a/src/NHibernate/Type/IntBooleanType.cs
+++ b/src/NHibernate/Type/IntBooleanType.cs
@@ -1,0 +1,74 @@
+﻿using System.Data;
+using NHibernate.SqlTypes;
+using NHibernate.UserTypes;
+
+namespace NHibernate.Type
+{
+    public class IntBooleanType : BooleanType, IUserType
+    {
+        public override SqlType SqlType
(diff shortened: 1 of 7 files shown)


oskarb:
Please link to an issue in Jira with description of the problem or feature.

Also, there are issues with the indenting, I suspect tabs/versus spaces. Consider installing EditorConfig (and ReSharper if you can) to get correct settings and fix this.

oskarb:
Also all builds are failing, which seems to be because of problems with the IntBooleanType. This class also lacks documentation similar to e.g. BooleanType. When is IntBooleanType required?

src/NHibernate/Type/IntBooleanType.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System.Data;
+using NHibernate.SqlTypes;
+using NHibernate.UserTypes;
+
+namespace NHibernate.Type
+{
+    public class IntBooleanType : BooleanType, IUserType
```
hazzik: This should be done with query substitutions.

PR #476, opened 2016-06-03
Eliminated double Persister resolution in Loader.InstanceNotYetLoaded flow

usarskyy:
Eliminated double Persister resolution when Loader.LoadFromResultSet(...) and session.Instantiate(...) get called from Loader.InstanceNotYetLoaded(...)


hazzik:
Hi, can you please enable [Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) feature?

usarskyy:
Enabled

fredericDelaporte:
I will probably review it today.

Question unrelated with review: instead of merging, I would tend to rebase instead. Is it considered bad practice to rewrite the commits from a branch we are not the author?  
(Of course I understand we should not do that when it appears others are currently working on it. Otherwise that would causes conflicts for their pushes. Or worst, if they push just before the rebase forced push, causes overrides of their commits.)

src/NHibernate/Loader/Loader.cs
```diff
@@ -992,7 +992,7 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
 			}
 			else
 			{
-				obj = session.Instantiate(instanceClass, key.Identifier);
+				obj = session.Instantiate(instanceClass.EntityName, key.Identifier);
```
hazzik: This should not have changed

usarskyy:
@frederic for me, as an author, it is absolutely OK. I will do new checkout anyway.

@hazzik: I don't have my laptop around right now. Did I change that? Or, is it something that was not correctly merged?

hazzik:
>Did I change that?

Yes. I think you just forgot to use the newly created method on ISessionImplementor.

src/NHibernate/Loader/Loader.cs
fredericDelaporte: Indeed, this change defeats the purpose of your patch: avoiding to resolve again the persister. The naming could be improved for code clarity: `instanceClass` should be renamed `concretePersister` and `GetRowPersister` would be renamed `GetConcretePersister`.

src/NHibernate/Util/EnumerableExtensions.cs
```diff
@@ -6,6 +6,16 @@ namespace NHibernate.Util
 {
 	public static class EnumerableExtensions
 	{
+		public static HashSet<T> ToHashSet<T>(this IEnumerable<T> collection)
```
fredericDelaporte: It is likely that others projects define such an extension too. This could cause conflicts for users. Maybe should it be set as `internal` instead of `public`. @hazzik, what do think of this?

hazzik: I think this chunk of code needs to be reverted.

fredericDelaporte: Indeed, its sole purpose is exploiting type inference for avoiding having to explicitly specify the element type by calling HashSet<> constructor.

usarskyy: I am sorry but I disagree with the statement that ToHashSet() has to be revered just because it is so simple. For example, ToList() and ToArray() methods from .NET Framework lib do exactly same. It is simply convenient.

Also, if you scroll a bit down you will see ForEach() extension method. Its sole purpose is to avoid writing foreach-statement :) And there are a few more methods like that in different *Extensions.

hazzik: >Also, if you scroll a bit down you will see ForEach() extension method.

We are discouraged of using these extensions, and the plan to eliminate their usage.

usarskyy: Only ForEach or also ToList()/ToArray()/etc?

fredericDelaporte: ... It is about custom utilities, which NHibernate is not focused at providing, rather than about standard built-in functionalities.

usarskyy: Personally I don't see any problem in having utility extension methods. But I will update my pull request if this is NHibernate's guidelines.

usarskyy: Removed

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1859,23 +1870,10 @@ public override void List(CriteriaImpl criteria, IList results)
 
 				string[] implementors = Factory.GetImplementors(criteria.EntityOrClassName);
 				int size = implementors.Length;
-
```
hazzik: What is the point replacing this loop with 2 loops?

fredericDelaporte: It looks to me there was already two loops with previous implementation, but one (the `UnionWith`) buried inside the other (the `for`).
I find this new implementation cleaner than previous one which was using a `for` loop. Now it causes more deviation from java impl and could cause harder port of Hibernate changes: is-it a trouble?

hazzik: It's not a problem, I just don't see what is the point of this irrelevant change.

hazzik: I think this is a [failed] attempt to remove double persister resolution here. `GetImplementors` gets the persister from by a class name and then returns `EntityName`, and `GetOuterJoinLoadable` gets the persister again.

usarskyy: Hi!
Unfortunately now I don't remember why I didn't refactor everything. Most probably I was hoping that my pull request can be included in current version of NH and I tried to minimize amount of (API) changes.
But since this pull request will make only in NH 5, I think, I could make a few more adjustments.

usarskyy: I reviewed the code once more. It will require some public API changes on ISessionFactoryImplementor level.
Shall I continue to refactor rest of the code?

usarskyy: I made necessary modifications but I would rather push it as a separate pull request.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -189,6 +189,7 @@ public IList ListFilter(object collection, IQueryExpression queryExpression, Que
 		public abstract void AfterTransactionCompletion(bool successful, ITransaction tx);
 		public abstract object GetContextEntityIdentifier(object obj);
 		public abstract object Instantiate(string clazz, object id);
+		public abstract object Instantiate(IEntityPersister persister, object id);
```
hazzik: Adding a new abstract member

src/NHibernate/Engine/ISessionImplementor.cs
```diff
@@ -203,6 +203,12 @@ public partial interface ISessionImplementor
 		/// </summary>
 		object Instantiate(string entityName, object id);
 
+		/// <summary>
+		/// Instantiate the entity class, initializing with the given identifier
+		/// </summary>
+		object Instantiate(IEntityPersister persister, object id);
```
hazzik: A new interface member

bahusoid:
I believe I've removed all breaking changes

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -1395,7 +1395,7 @@ public virtual object InitializeLazyProperty(string fieldName, object entity, IS
 		}
 
 		public void InitializeLazyProperties(
-			DbDataReader rs, object id, object entity, ILoadable rootPersister, string[][] suffixedPropertyColumns,
+			DbDataReader rs, object id, object entity, string[][] suffixedPropertyColumns,
```
bahusoid: Unreleased public API change (removed unused rootPersister param)

PR #477, opened 2016-06-05
merged entity's properties related arrays (propertyNames, propertyTyp…

usarskyy:
…es, etc): all arrays alias main "properties" array

Public interface of EntityMetamodel remains unchanged.

### DIFF
--- a/src/NHibernate/Tuple/Entity/EntityMetamodel.cs
+++ b/src/NHibernate/Tuple/Entity/EntityMetamodel.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Runtime.CompilerServices;
 
 using NHibernate.Engine;
@@ -7,7 +8,7 @@ using NHibernate.Intercept;
 using NHibernate.Mapping;
 using NHibernate.Type;
 using NHibernate.Util;
-using Array = NHibernate.Mapping.Array;
+
 
 namespace NHibernate.Tuple.Entity
 {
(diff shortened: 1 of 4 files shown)


usarskyy:
Hi everyone,

can somebody review my changes? It's been awhile since I pushed them.
As you can see, they are not so sophisticated. I was mainly focused on simplifying debugging and refactoring process. All public interfaces remain unchanged.
BTW, I can also remove outdated variables, that I keep for backward compatibility, if you decide to include those changes in NH v5.

gliljas:
Hi and thanks!

Have you evaluted the performance implications of using this approach, since it uses a lot more iterations over the properties? I agree that it makes the code clearer, which may be worth a bit of an impact in the right phase.

usarskyy:
Hi!

Yes, I did. Actually, current implementation does not have performance issues because all arrays are still there. They are just lazy-evaluated. My target is to remove them at all and I even did application profiling to check the performance without arrays but unfortunately did not record the results.
If you give me a bit of time, I can run profiling once more and prepare a small report.

fredericDelaporte:
We need a [Jira](https://nhibernate.jira.com) for keeping track of changes and releasing, could please add one?

bahusoid:
I don't quite understand the title as I don't see any  "merged entity's properties related arrays".

And I don't think that making all arrays lazy can be considered as improvement - as it involves locking on initialization and in general works slower. IMHO it's not worth it here.  Shall we close it?

fredericDelaporte:
It seems the aim was to remove all these arrays, but the PR was not there yet.

hazzik:
Ok, closing.

PR #478, opened 2016-06-09
Remove unnecessary ID increment in SequenceHiLoGenerator

livioc:
fixed  using the same implementation of TableHiLoGenerator and of hibernate-java seqhilo


hazzik:
Any tests?

livioc:
i will add them but i need to execute them only with databases that support sequences, how can i do that? there are some tests with the same constraint that i can check?  thanks

hazzik:
You can add tests only for TableHiLoGenerator itself

livioc:
thanks, i already did the test for sqhilo.
is it better to add a specific test for nh3879 to check that the first id is the first available number in current hilo range or to add a generic test in a seqhilo specic feature as [HiLoTableGeneratorInt64Fixture](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate.Test/IdTest/HiLoTableGeneratorInt64Fixture.cs) to check that we will use all the available ids in the range?

hazzik:

livioc:
Allowed

PR #484, opened 2016-07-25
NH-3726: Added escape character support in SqlMethods.Like

gliljas:
Did not yet follow my own suggestion to fix StartsWIth, Contains and EndWith, since that could be considered a breaking change. But it is a bug, IMO.

### DIFF
--- a/src/NHibernate.Test/Linq/FunctionTests.cs
+++ b/src/NHibernate.Test/Linq/FunctionTests.cs
@@ -23,6 +23,38 @@ namespace NHibernate.Test.Linq
 			Assert.That(query[0].FirstName, Is.EqualTo("Margaret"));
 		}
 
+		[Test]
+		public void LikeFunctionWithEscapeCharacter()
+		{
+			using (var tx = session.BeginTransaction())
+			{
+				var employeeName = "Mar%aret";
+				var escapeChar = '\\';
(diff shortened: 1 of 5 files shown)


hazzik:
>All checks have failed

gliljas:
Hmm...

I'm on a shaky rural connection, but will try to fix ASAP

gliljas:
Fixed and squashed

src/NHibernate.Test/Linq/FunctionTests.cs
```diff
@@ -23,6 +23,38 @@ where NHibernate.Linq.SqlMethods.Like(e.FirstName, "Ma%et")
...
+		{
+			using (var tx = session.BeginTransaction())
+			{
+				var employeeName = "Mar%aret";
+				var escapeChar = '\\';
+				var employeeNameEscaped = employeeName.Replace("%", escapeChar + "%");
+
+				//This entity will be flushed to the db, but rolled back when the test completes
+
+				session.Save(new Employee { FirstName = employeeName, LastName = "" });
```
hazzik: Oracle does not like `""` as `LastName` because in oracle `""` is `NULL`

oskarb:
I suspect it's the choice of escape character that is causing the fail in MySql. I will test.

oskarb:
I've fixed the issue on MySql, but decided to also rebase to get rid of unrelated formatting changes. Merged separately.

PR #486, opened 2016-07-28
NH-3891 - Add [SecurityCritical] attributes to ISerializable.GetObjectData overrides

ngbrown:
Specifically, `NHibernate.Engine.CascadeStyle`, was missing the attributes and it was causing errors when I ran the unit tests.


oskarb:
Thanks, I've merge a reindented version in 35135126864c6f405cf59cbe1e280944583806db.

Consider using EditorConfig to have VS follow NHibernate indenting rules.

PR #487, opened 2016-07-29
NH3500 - Unwrap TIE errors for classes with lazy properties

MarkPerryBV:


MarkPerryBV:
OK so what's wrong with these tests?

If someone can let me know how to resolve that would be great.

oskarb:
Unfortunately some of the builds have persistent issues. They don't indicate any new failures though so it should be ok.

oskarb:
I've made some adjustment, but will proceed with merging it later. Some unrelated tech trouble right now.

oskarb:
Thanks for the pull request. I've moved the test to where another test for the same class was to keep it a bit more together.

I see you made your commit on master. For future reference it's usually better to start a topic branch as a basis for pull requests - doing it on master will eventually lead to wrong commits being included if you later on want to submit another change.

PR #490, opened 2016-08-03
Fix many odbc test failures

oskarb:
Fixes for datetime handling with MSSQL over ODBC.


oskarb:
Eh, no... I messed it up with regards to which dialect is being used.

PR #491, opened 2016-08-04
[WIP] Use nuget package instead of local libraries

hazzik:


ngbrown:
I think using NuGet is a great idea.  I noticed that `src/NHibernate/packages.config` doesn't have all of the database providers that NHibernate supports.  Since the current NHibernate loads drivers that exist ambient in the directory by reflection, it doesn't **need** to include them.  However, I think this is a good point to introduce splitting the different database providers into separate packages so each provider NuGet package can then depend on specific version ranges of the database provider NuGet packages.

For example, there would be a `NHibernate.SQLServer`, `NHibernate.SQLite`, `NHibernate.Oracle`, and so on.  Each of these would then depend on the main `NHibernate` package and a provider package such as `Oracle.ManagedDataAccess`.

This way, reflection would no longer be needed and developers could be sure that the provider packages were pulled down along with NHibernate.

hazzik:
Superseded by #549

PR #493, opened 2016-08-05
Team City

hazzik:
- Previously the Postgres build was treated as successfully finished, even if there were failing tests
- Use TeamCity NUnit AddIn to track tests online
- Fix tests for PostgreSQL
 - The BulkManipulation.UpdateMultiplePropertyOnAnimal and BulkManipulation.UpdateOnAnimal tests were using Double parameter types to update the Single column. It seems that now Npgsql calculates parameter types according to the columns they touches and fails to set the parameters in this case.
 - The CriteriaQueryTest.TestSubcriteriaBeingNull were using the Int32 parameter type instead of expected Double.

### DIFF
--- a/build-common/common-project.xml
+++ b/build-common/common-project.xml
@@ -99,7 +99,7 @@
 	<target name="common.run-tests"
 		description="Run NUnit tests">
 		<call target="common.find-nunit" unless="${property::exists('nunit.found')}" />
-		<property name="common.run-tests.failonerror" value="${not property::exists(project::get-name() + '.IgnoreFail')}"/>
+		<property name="common.run-tests.failonerror" value="${not property::exists(test.file + '.IgnoreFail')}"/>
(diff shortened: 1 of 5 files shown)


oskarb:
There is now  three new failing tests on ODBC, but I'm not sure why. Those tests aren't touched directly by this commit, and furthermore according to teamcity test history they did not run before, and even now they have only run on the ODBC build:
http://teamcity.codebetter.com/viewLog.html?buildId=224380&tab=buildResultsDiv&buildTypeId=bt876

Strange.

oskarb:
Btw, I decided to merge my fixes for ODBC but I don't think that will affect these three tests.

oskarb:
Oh, I see now. It's actually a failure in TestFixtureSetup, so I guess those just weren't being reported/detected by teamcity before. And assuming they will only be reported as a _fixture_ if there is an actual failure, that would explain why they, according to the test history, wrongly appear to run only on ODBC.

PR #494, opened 2016-08-06
Makes DebugConnectionProvider.connections thread safe.

ngbrown:
Fixes NH-3436 failing for me when multiple threads try to extend
the HashSet at the same time.


oskarb:
Please open a new issue in Jira for this. This problem may trouble the NH-3436 test case, but it is really just coincidental to it should be reported separately.

ngbrown:
@oskarb Sorry about that.  I created a new pull request: #495.  I'll close this one and let the other one be merged.  (yes I know I can force push, but I wanted the branch name to match the Jira ticket number).

PR #496, opened 2016-08-09
NH-3856 - Improve performance of SqlClientSqlCommandSet

gliljas:
Moved reflection discovery from instance constructor to static constructor.


hazzik:
Hi, I have a changes in a branch which move reflection to the static constructor. I'll use them.

hazzik:
Also. PropertyInfo is _extremely_ slow. My code compiles Action<T>/Func<T> which is fast. Also, it is faster then just a delegate.

gliljas:
Yes, the previous delegates should be about the same as propertyinfo. Please update JIRA to reflect that you're working on it.

PR #498, opened 2016-08-10
NH-3885 - Replace ThreadSafeDictionary wrapper with Framework's ConcurrentDictionary

ngbrown:
This should address [NH-3885](https://nhibernate.jira.com/browse/NH-3885) by replacing all uses of the custom `ThreadSafeDictionary` wrapper with the .NET Framework provided `ConcurrentDictionary`.

In most places the `IDictionary<,>` interface is kept as it provides the expected exception throwing when adding existing keys, otherwise the caller would be responsible for error handling on the result of `TryAdd()`.

In `SessionFactoryImpl` specifically, I update the code to use the thread-safe `GetOrAdd()` as that was what the intention of the code was.

This removes `ThreadSafeDictionary` from the project.


src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -247,7 +248,7 @@ public SessionFactoryImpl(Configuration cfg, IMapping mapping, Settings settings
 					if (cache != null)
 					{
 						caches.Add(cacheRegion, cache);
-						allCacheRegions.Add(cache.RegionName, cache.Cache);
+						((IDictionary<string, ICache>)allCacheRegions).Add(cache.RegionName, cache.Cache);
```
stefan-steinegger: I think the explicit interface implementations of the `ConcurrentDictionary` are not thread safe (that's why they are explicitly implemented). I had a similar solution on my machine. Currently I'm in holidays. I'll compare yours to mine when I'm back.

ngbrown: @stefan-steinegger I had that same concern too, but I looked at the [Microsoft Reference Source](http://referencesource.microsoft.com/#mscorlib/system/Collections/Concurrent/ConcurrentDictionary.cs,1218) and the `IDictonary` implementation is a very simple wrapper around the main `TryAdd` and it also throws if not added (`TryAdd` returns `false`).  This throw code could easily be moved into the application, but since it is in the library, in the explicit `IDictionary` definitions, I left it there.

oskarb: The documentation, however, states:
All public and protected members of ConcurrentDictionary<TKey, TValue> are thread-safe and may be used concurrently from multiple threads. However, members accessed through one of the interfaces the ConcurrentDictionary<TKey, TValue> implements, including extension methods, are not guaranteed to be thread safe and may need to be synchronized by the caller. 

We shouldn't rely on behavior the documentation doesn't promise.

oskarb: Shouldn't we simply use TryAdd() instead?

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -248,7 +248,10 @@ public SessionFactoryImpl(Configuration cfg, IMapping mapping, Settings settings
 					if (cache != null)
 					{
 						caches.Add(cacheRegion, cache);
-						((IDictionary<string, ICache>)allCacheRegions).Add(cache.RegionName, cache.Cache);
+						if (!allCacheRegions.TryAdd(cache.RegionName, cache.Cache))
+						{
+							throw new HibernateException("Key already existed");
```
oskarb: As per guidelines, exception messages are supposed to be proper sentences, which includes a period at the end. (Multiple locations.)

ngbrown:
@oskarb I updated the messages to a bit more than the standard .Net framework Dictionary exception sentence, including the dictionary variable name.

lnu:
I'm doing some tests with the new version and it seems NHibernate.Envers [relies on the ThreadSafeDictionnary class](https://bitbucket.org/RogerKratz/nhibernate.envers/src/fb8a65e8141e1a194c29d5033621c53c48fc1a6b/Src/NHibernate.Envers/Synchronization/AuditProcessManager.cs?at=default&fileviewer=file-view-default).

PR #499, opened 2016-08-12
Test case and fix to NH-3898

hailtondecastro:
Test case and fix to [NH-3898](https://nhibernate.jira.com/browse/NH-3898)
Configuring a property with generated="insert" it makes
"Property.IsUpdatable" "false" even using update="true" in the xml
mapping file.
Please consider keep this issue as Critical, as it does not break the
application, it makes the applications migrated from NHibernate version
2 to 3 or 4 generates inconsistent data within the database.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ hibernate.cfg.xml
 available-test-configurations
 current-test-configuration
 NHibernate.VisualState.xml
+/.project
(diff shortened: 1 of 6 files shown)


oskarb:
Replaced by PR #500.

oskarb:
Oh, sorry it was different branches actually.

oskarb:
However, every build failed, and the indenting doesn't match NHibernate's style.

fredericDelaporte:
@hazzik, are we still releasing patches to v3.4? I guess no and we should close this PR.

hazzik:
There will be no support for 3.4.x branch except regressions caused by the .NET framework changes.

PR #500, opened 2016-08-12
Test case and fix to NH-3898

hailtondecastro:
Test case and fix to [NH-3898](https://nhibernate.jira.com/browse/NH-3898)
Configuring a property with generated="insert" it makes
"Property.IsUpdatable" "false" even using update="true" in the xml
mapping file.
Please consider keep this issue as Critical, as it does not break the
application, it makes the applications migrated from NHibernate version
2 to 3 or 4 generates inconsistent data within the database.


fredericDelaporte:
If this patch get merged, it would be probably more likely to happen on master or 4.1 branch.  If you port your PR on master, now it requires VS2017 (Community edition is enough).

The change looks legit, but you need to follow [contributing guidelines](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md). Notably, use tabs instead of spacing.

fredericDelaporte:
Re-based for fixing the release build, small additional fix added to test.

src/NHibernate.Test/NHSpecificTest/NH3898/Fixture.cs
```diff
@@ -0,0 +1,61 @@
...
+			}
+
+			using (var session = OpenSession())
+			using (var tx = session.BeginTransaction())
+			{
+				var employee = session.Get<Employee>(id);
+				Assert.That(employee.PromotionCount, Is.EqualTo(0));
+				employee.Name = "Employee 1 changed";
+				employee.PromotionCount++;
+				Assert.That(employee.PromotionCount, Is.EqualTo(1));
```
hazzik: That asert does not make sense anymore.

hazzik:
Thanks

PR #492, opened 2016-08-04
Test for NH-3874

LordJZ:
https://nhibernate.jira.com/browse/NH-3874

PR #501, opened 2016-08-13
NH-3874 - Fix NullReferenceException

hazzik:
If user was trying to remove or evict a collection which key is of `IUserType` when the `ReturnedType` property of the `IUserType` implementation returns a base type of an actually persisted type (eg. `object`) the `MessageHelper.CollectionInfoString` could throw an `NullReferenceException` due to incorrect type check.

Replaces #492


hazzik:
@nhibernate/committers please review

PR #503, opened 2016-08-20
Nh 3899 Fixes to avoid too long column aliases

oskarb:
https://nhibernate.jira.com/browse/NH-3899

### DIFF
--- a/src/NHibernate.Test/MappingTest/ColumnFixture.cs
+++ b/src/NHibernate.Test/MappingTest/ColumnFixture.cs
@@ -39,5 +39,53 @@ namespace NHibernate.Test.MappingTest
 			column.Length = 100;
 			Assert.AreEqual("NVARCHAR(100)", column.GetSqlType(_dialect, null));
 		}
+
+
+		[TestCase("xxxxyyyyz")]
+		[TestCase("xxxxyyyyzz")]
+		[TestCase("xxxxyyyyzzz")]
+		[TestCase("xxxxyyy4z", Description = "Non-letter digit character would be cut, make sure we don't skip length check.")]
+		[TestCase("xxxxyyyz4z", Description = "Non-letter digit character would be cut, make sure we don't skip length check.")]
+		[TestCase("xxxxyyyzz4z", Description = "Non-letter digit character would be cut, make sure we don't skip length check.")]
+		[TestCase("xxxxyyyy4", Description = "Non-letter digit character would be cut, make sure we don't skip length check.")]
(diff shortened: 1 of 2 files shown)


hazzik:
This, for some reasons, led to a duplicate alias generation. 

Before:
```sql
SELECT
    this_.Name as Name2_2_,
    morechildr4_.parentName as parentName,
    child1_.Name as childName,
    morechildr4_.loc as loc,
    child1_.Name as Name0_0_,
    friends6_.childName1 as childName1_,
    child2_.Name as childName2_,
    child2_.Name as Name0_1_ 
FROM
    Parent this_ 
	inner join ParentChild morechildr4_ on this_.Name=morechildr4_.parentName 
	inner join Child child1_ on morechildr4_.childName=child1_.Name 
	inner join ChildChild friends6_ on child1_.Name=friends6_.childName1 
	inner join Child child2_ on friends6_.childName2=child2_.Name 

ORDER BY 
    this_.Name desc
```
After:
```sql
SELECT 
	this_.Name as Nam1_3915_2_, 
	morechildr4_.parentName as parentNa1_, 
	child1_.Name as childNam2_, 
	morechildr4_.loc as loc3_, 
	child1_.Name as Nam1_3913_0_, 
	friends6_.childName1 as childNam1_, 
	child2_.Name as childNam2_, 
	child2_.Name as Nam1_3913_1_ 

FROM Parent this_ 
	inner join ParentChild morechildr4_ on this_.Name=morechildr4_.parentName 
	inner join Child child1_ on morechildr4_.childName=child1_.Name 
	inner join ChildChild friends6_ on child1_.Name=friends6_.childName1 
	inner join Child child2_ on friends6_.childName2=child2_.Name 

ORDER BY 
	this_.Name desc
```

oskarb:
I've been looking at this earlier today actually.

The reason it worked before the change is mostly by accident: 
* In line 4 (of the SQL as above), ParentChild.childName is returned as-is from GetAlias() because it contains no digits and is less than 10 characters.
* Inline 7 and 8, ChildChild.childName1 and childName2 both gets the digit cut off and because the remainder is clearly not equal to the full column name, the suffix is added (which happens to contain the same digit as the one cut off).

If for example ParentChild.childName had instead been named childName7 in the mapping, the digit would have been cut and the 2_ suffix added, causing the same issue.

In Hibernate, the AbstractCollectionPersister also includes the table in two of four calls to GetAlias(), which NHibernate doesn't. Adding them fixes this. Working on it.

PR #506, opened 2016-08-24
NH-3903: TableGenerator quotes table and column names.

jan-schubert:
https://nhibernate.jira.com/browse/NH-3903

### DIFF
--- a/src/NHibernate.Test/IdTest/TableGeneratorFixture.cs
+++ b/src/NHibernate.Test/IdTest/TableGeneratorFixture.cs
@@ -25,5 +25,35 @@ namespace NHibernate.Test.IdTest
 			Assert.That(selectSql.GetValue(tg).ToString(), Is.StringContaining(customWhere));
 			Assert.That(updateSql.GetValue(tg).ToString(), Is.StringContaining(customWhere));
 		}
+
+		[Test]
+		public void GenerateSelectAndUpdateWithoutQuotes()
+		{
+			var dialect = new MsSql2005Dialect();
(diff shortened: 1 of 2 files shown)


oskarb:
Looks good, but can you please fix the indenting in TableGenerator to match the surrounding indent style?

NHibernate generally uses tabs for indent, and space for alignment. Consider using Resharper and EditorConfig to have VS automatically pick up on NHibernate's configured style.

oskarb:
The corresponding change in Hibernate is bit more full-featured:
Corresponding issue for Hibernate: https://hibernate.atlassian.net/browse/HHH-4553
Their commit: https://github.com/hibernate/hibernate-orm/commit/a2bf14ae7c2ed7b2b0a6eb558dad08bcfc69c489

Later on, they've also deprecated and then removed the TableGenerator and there are lots of changes around the id generators. Would be interesting to port a larger piece.

jan-schubert:
Sorry but I have not enough time to develop the full featured change, currently.

I hope you integrate my pull request for the next release. We need it to deploy our next software release.

oskarb:
The patch, as is, means that the NHibernate.Id modules takes a dependency on NHibernate.Mapping, which it doesn't currently have. Futhermore, NHibernate.Mapping does already have a dependency on NHibernate.Id.

Not sure which direction is best, but making it cyclic isn't good.

oskarb:
Sorry, but I don't want to accept this as is. I will nominate the bug for for NH5.0 in the hopes that a better fix can be provided. In the meantime you can use a custom build.

The test cases from this PR should be useful.

PR #507, opened 2016-09-06
Fix for NH-3904

jeffreyabecker:
Fixes custom types not being properly detected during linq expression tree translation to IASTNode.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/EntityWithUserTypeCanHaveLinqGenerators/ExampleUserType.cs
+++ b/src/NHibernate.Test/NHSpecificTest/EntityWithUserTypeCanHaveLinqGenerators/ExampleUserType.cs
@@ -1,0 +1,94 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Text;
+using NHibernate.SqlTypes;
+using NHibernate.UserTypes;
+
+namespace NHibernate.Test.NHSpecificTest.EntityWithUserTypeCanHaveLinqGenerators
+{
+	public class ExampleUserType : IUserType
+	{
+		public new bool Equals(object a, object b)
+		{
+			IExample ga = a as IExample;
+			IExample gb = b as IExample;
+			if (ga == null && gb == null)
+			{
+				return true;
+			}
+			if (ga != null && gb != null && (ga.GetType() == gb.GetType()))
+			{
+				return ga.Value == gb.Value;
+			}
+			return false;
+		}
+
+		public int GetHashCode(object x)
+		{
+			return ((IExample)x).Value.GetHashCode();
+		}
+
+		public object DeepCopy(object value)
+		{
+			if (value == null) return null;
+			if (value.GetType() == typeof(BarExample))
+			{
+				return new BarExample { Value = ((IExample)value).Value };
+			}
+			return new FooExample { Value = ((IExample)value).Value };
+		}
+
+		public object Replace(object original, object target, object owner)
+		{
+			return original;
+		}
+
+		public object Assemble(object cached, object owner)
+		{
+			return cached;
+		}
+
+		public object Disassemble(object value)
+		{
+			return value;
+		}
+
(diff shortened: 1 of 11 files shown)


src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -88,6 +93,14 @@ protected override Expression VisitConstantExpression(ConstantExpression express
...
 					type = NHibernateUtil.GuessType(expression.Type);
 
+
+			    var customType =
+			        _allMappedCustomTypes.FirstOrDefault(ct => ct.UserType.ReturnedType.IsAssignableFrom(expression.Type));
+			    if (customType != null)
+			    {
+			        type = customType;
+			    }
+
```
oskarb: If use  usertype for 1 int property because I need to store that in some special way, won't this then affect ALL integer constants in LINQ queries?

src/NHibernate.Test/NHSpecificTest/EntityWithUserTypeCanHaveLinqGenerators/ExampleUserType.cs
```diff
@@ -0,0 +1,95 @@
...
+            var index = rs.GetOrdinal(names[0]);
+            if (rs.IsDBNull(index))
+            {
+                return null;
+            }
+            var buff = new byte[1024 * 8];
+            var len = (rs).GetBytes(index, 0, buff, 0, buff.Length);
+            var val = Encoding.UTF8.GetString(buff, 0, (int)len);
+            var parts = val.Split(':');
+            if (parts[0] == "Bar")
```
oskarb: Do we really need all these technicalities in the user type to expose the issue, or can it be minimized?

oskarb:
No.... REBASE you branch on current master please, instead of merging master to your branch. The latter causes the new commits on master to appear as part of your PR.

jeffreyabecker:
I'm confused, I dont believe I've resubmitted anything yet.

jeffreyabecker:
Do I need to rebase on upstream/master rather than origin/master?

oskarb:
Yeah I looked to and I'm also confused to exactly what happened.

Note that when you add commits to the branch you submitted as a PR, they are automatically included in the PR if that's still open. Same thing happens when you rebase and overwrite the branch.

jeffreyabecker:
I merged upstream/master into origin/master then rebased on origin/master. Is that possibly not doing what I'd expected?

oskarb:
You should rebase on something equivalent to the official master (don't use your local master if that has additional changes).

Something like:
$ checkout your branch
$ git rebase master (or nhibernate/master, or whatever you call it that's exactly the same as the true upstream master).

oskarb:
I think you're rebase was OK, but then you somehow merged you old branch to your new rebased branch or something like that which makes it very confused. Suggest you rebase again, interactively, and just skip the strange merge commit.

jeffreyabecker:
Ok, those issues seem to have been fixed by a slightly different rebase.

jeffreyabecker:
On a separate note are there a set of Resharper settings available anywhere so I don't have the formatting issues?

jeffreyabecker:
Ok, I think that covers everything.  I'm only seeing two commits on my branch now which looks correct.

oskarb:
Settings for both Resharper and EditorConfig are in the project and should take effect as needed. You may need the EditorConfig extension to properly switch between space and tabs if you have other space-based projects.

oskarb:
Yes, seems much better. Would be great if you could also squash the two commits into just one - keeping the formatting changes separate isn't useful to long-term history.

jeffreyabecker:
Done.

oskarb:
Please fix the failure on SqlServerCe:
http://stackoverflow.com/questions/17053885/error-the-specified-data-type-is-not-valid-data-type-if-known-varchar

jeffreyabecker:
There's also a failure on Oracle 32 but I cant get into teamcity to see what it is. Is that something that I need to address?

oskarb:
The Oracle32 failures are irrelevant.

PR #509, opened 2016-10-09
Added generic IResultTransformer interface

mwpowellhtx:
I am also working on a _ResultTransformerBase_ that helps automate ad hoc database queries that would otherwise not require a full mapping, but I see that `ResultTransformer` already has a significant footprint, so I want to tread carefully where that's concerned. Will be content with the interface for now; future direction for this feature is to automate the transformation, whether via delegated `Func`, `Expression`, `Reflection`, or what have you.

Possibly use cases:

``` Sql
SELECT [dbo].[fn_does_my_feature_work](N'blahblahblah',1) AS [Working]
```

And which would respond to the data set including `[Working]`, whether that's a first class object `T Result`, or a simple `System.Boolean`, matters not.

So far it's working well in my situation, and it is **FAST**, which is _CRITICAL_ for this purpose.

### DIFF
--- a/src/NHibernate/Transform/IResultTransformer.cs
+++ b/src/NHibernate/Transform/IResultTransformer.cs
@@ -1,10 +1,9 @@
-using System;
 using System.Collections;
 
 namespace NHibernate.Transform
 {
 	/// <summary>
-	/// Implementors define a strategy for transforming criteria query
+	/// Implementers define a strategy for transforming criteria query
 	/// results into the actual application-visible query result list.
 	/// </summary>
 	/// <seealso cref="NHibernate.ICriteria.SetResultTransformer(IResultTransformer)" />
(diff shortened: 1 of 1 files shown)


oskarb:
Hmm... I don't think we want to add an interface that isn't actually used inside NHibernate. Why would that be required?

mwpowellhtx:
@oskarb

> Why would that be required?

I illustrated a couple of use cases in the description above.

oskarb:
Why is this interface required _in NHibernate_ when NHibernate doesn't actually do anything with it?

mwpowellhtx:
@oskarb IMO, because it is more closely related to a service NHibernate is providing, which happens to be useful in the scenarios I described. It's a suggestion and I will leave it there.

hazzik:
As per @oskarb

PR #510, opened 2016-10-12
Rename connection.provider_class to connection.provider

JornWildt:
The documentation refers to "connection.provider_class" but the example shows "connection.provider". On the system I have been working on it is "connection.provider" that works.


hazzik:
Thanks

PR #511, opened 2016-10-17
NH-3911 test

LordJZ:


oskarb:
Please try to give the test case method a self-descriptive name, and add some comments to explain the scenario. Doesn't need to be long. Also, could you have a look and see if this test case would fit in with some existing topic-based test cases instead of under NHSpecificTest? There should be existing tests already for the classes affected by the fix. If it's not too awkward, it's better to keep it organized by topic.

Since you already have a proposed fix, please include the tests in the PR #512 instead, so they would go in together. Currently the build server is down, but when it's working it would normally build and test all PRs and then this PR would come up with failing tests, while the checks for the PR containing the fix wouldn't actually test the fix until after both PRs have been merged.

PR #513, opened 2016-10-19
NH-3912 Oracle IBatcher impl instance is not reusable after failed operation.

waynesayonara:
Batch operations with the same IBatcher instance fail on expected rows count after single failed operation.
Also, passed DbException message for oracle batcher to outer message.

This issue regards OracleDataClientBatchingBatcher. It seems that in the scenario where two batch commands are performed using the same instance of OracleDataClientBatchingBatcher, second batch command will nearly always fail on expected rows count assert. This will only happen if first batch command failed before the second one (any error will do). The reason is that internal counter _totalExpectedRowsAffected isn't reset after first failed batch command.
Example:
BATCH1:
_totalExpectedRowsAffected += 1 (now 1, was 0)
Insert into table and fail on unique constaint violation for instance
_totalExpectedRowsAffected is not reset.
BATCH2:
_totalExpectedRowsAffected += 1 (now 2, was 1)
Insert into table and succeed
Assert Expectations.VerifyOutcomeBatched fails because _totalExpectedRowsAffected and only one row was actually affected.
Every consecutive batch command will also fail.

Pull requests contains following changes:
- _totalExpectedRowsAffected is reset now regardless of the batch query execution outcome (in finally block)
- DbException.Message is passed to the AdoExceptionHelper, so diagnostic information will be included in the exception propagated to user code.

### DIFF
--- a/src/NHibernate/AdoNet/OracleDataClientBatchingBatcher.cs
+++ b/src/NHibernate/AdoNet/OracleDataClientBatchingBatcher.cs
@@ -125,21 +125,27 @@ namespace NHibernate.AdoNet
 				// this value is not a part of the ADO.NET API.
 				// It's and ODP implementation, so it is being set by reflection
 				SetArrayBindCount(arraySize);
-				int rowsAffected;
 				try
 				{
-					rowsAffected = _currentBatch.ExecuteNonQuery();
+					int rowsAffected;
+					try
+					{
+						rowsAffected = _currentBatch.ExecuteNonQuery();
+					}
+					catch (DbException e)
+					{
+						throw ADOExceptionHelper.Convert(Factory.SQLExceptionConverter, e, "could not execute batch command.");
+					}
+
+					Expectations.VerifyOutcomeBatched(_totalExpectedRowsAffected, rowsAffected);
 				}
-				catch (DbException e)
+				finally
 				{
-					throw ADOExceptionHelper.Convert(Factory.SQLExceptionConverter, e, "could not execute batch command.");
+					// Cleaning up even if batched outcome is invalid
+					_totalExpectedRowsAffected = 0;
+					_currentBatch = null;
+					_parameterValueListHashTable = null;
 				}
-
-				Expectations.VerifyOutcomeBatched(_totalExpectedRowsAffected, rowsAffected);
-
-				_totalExpectedRowsAffected = 0;
-				_currentBatch = null;
-				_parameterValueListHashTable = null;
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


hazzik:
@waynesayonara I've restricted tests to the Oracle specific dialects only, fixed the indents. So I had to force push your changes with rebase on current master.

waynesayonara:
@hazzik Do I need to fix unsuccessful checks at this point (3 of them) or you will merge it anyway?
PS Sorry about indents.

hazzik:
They are unrelated, these tests are fluffy.

src/NHibernate/AdoNet/OracleDataClientBatchingBatcher.cs
```diff
@@ -125,21 +125,32 @@ protected override void DoExecuteBatch(IDbCommand ps)
...
-					rowsAffected = _currentBatch.ExecuteNonQuery();
+					int rowsAffected;
+					try
+					{
+						rowsAffected = _currentBatch.ExecuteNonQuery();
+					}
+					catch (DbException e)
+					{
+						// Message from DbException could be useful
+						var additionalMessage = string.IsNullOrEmpty(e.Message) ? "." : string.Concat(": ", e.Message);
```
oskarb: Not sure why this is required. I've not run it, but looking at the code is seems the exception ``e`` should eventually end up as inner exception inside the "converted" exception and the message can and should be read from that.

hazzik: @oskarb this is how it was already implemeted. The real change here is additional try-finally around the execution block. I think it shall go to 4.1

oskarb: additionalMessage in string.Format("could not execute batch command{0}", additionalMessage) is new. That's what I'm referring to.

hazzik: In MinimalSQLExceptionConverter and SQLStateConverter the actual exception will end up as an inner exception of GenericADOException. So, I think it's fine. Or, we can amend the PR to remove this change.

oskarb: Yes, that's my point, it's already there. Based on the comment in the PR and also in the code, the additionalMessage seems to have been added based on the idea that without it, the message from the original exception was somehow lost. As far as I can tell, the inner exception will be there as expected and then for anything that does log the exception chain properly, this will just be unfortunate duplication. t's not a big issue and I don't care too much either way, but I think it should be removed.

hazzik:
@oskarb I've changed the PR

oskarb:
Sure we can have this in 4.1.

hazzik:
I've missed one point on this pull request...

**The user shall never reuse session after an exception. Ever.**

PR #514, opened 2016-10-24
Making test NH2420 less brittle

AndrewShepherd:
The test for defect NH2420 creates a connection independently of the NHibernate framework.

To create this connection, it first queries the NHibernate configuration for the connection string.
'
There are actually multiple ways to set an NHibernate connection string:

- Via the property `connection.connection_string`
- Via the property `connection.connection_string_name`

I had set up my app.config file to use the `connection.connection_string_name` property, and this test fails as a result.

This commit improves the test so that if it is unable to find the `connection.connection_string` property it checks for the `connection.connection_string_name` property.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH2420/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH2420/Fixture.cs
@@ -2,11 +2,14 @@
 using System.Data;
 using System.Data.Odbc;
 using System.Data.SqlClient;
+using System.Configuration;
 using System.Transactions;
 using NHibernate.Dialect;
 using NHibernate.Driver;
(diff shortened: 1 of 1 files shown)


oskarb:
Oh, now you've gone and done something unfortunate. Please keep changes for pull requests on separate topic branches, instead of your master-branch. Now that you've merged from upstream master to your master, that merge commit is also part of this pull request, and would pollute history.

Please create a topic branch containing the intended change only, then submit that as a new pull request.

AndrewShepherd:
@oskarb Thank you for the feedback.

I've created a new fork, and performed my change on a dedicated branch. This is now pull request [517](https://github.com/nhibernate/nhibernate-core/pull/517)

PR #516, opened 2016-10-28
NH-3100 - Implement correct semantic for equal and not equal considering null behaviours

hazzik:


hazzik:
@oskarb @gliljas @rjperes can you please review? This still WIP, but I don't think that there will be much changes.

gliljas:
Ok, will do.

hazzik:
I've decided to remove NH-3225 & NH-3915 changes as they require more work due dialect-specific issues.

PR #517, opened 2016-10-30
NH-2420 Supports connection.connection_string_name config property

AndrewShepherd:
The test for defect NH2420 creates a connection independently of the NHibernate framework.

To create this connection, it first queries the NHibernate configuration for the connection string.

There are actually multiple ways to set an NHibernate connection string:

- Via the property `connection.connection_string`
- Via the property `connection.connection_string_name`

I had set up my `app.config` file to use the `connection.connection_string_name` property, and this test fails as a result.

This commit improves the test so that if it is unable to find the `connection.connection_string property` it checks for the `connection.connection_string_name` property.


oskarb:
Thanks.

PR #482, opened 2016-07-21
NH-3889 - Fixed weird extra joins in subqueries

PleasantD:
See https://nhibernate.jira.com/browse/NH-3889

PR #522, opened 2016-11-08
NH-3918 - Nominate equality expressions in SELECT clauses

PleasantD:
with constants in them

* Allow for entity equality to be correctly translated to HQL
* Prevents unneeded joins and caching of entities in the query plan

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3918/FixtureByCode.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3918/FixtureByCode.cs
@@ -1,0 +1,127 @@
+﻿using System;
+using System.Linq;
+using System.Linq.Expressions;
+using NHibernate.Cfg.MappingSchema;
+using NHibernate.Linq;
+using NHibernate.Mapping.ByCode;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3918
+{
+	public class ByCodeFixture : TestCaseMappingByCode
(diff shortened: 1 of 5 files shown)


PleasantD:
Rebased on master

PleasantD:
Can this be included in 4.1.0?

oskarb:
I prefer to proceed with 4.1 as is rather than risk further delays by going into reviewing something. There is always going to be "one more PR". Better we try to get back to more frequent releases.

PleasantD:
How soon is the 4.1 release? How often will you be doing updates after that?
We are already using a custom build that includes this patch as well as #505 and #482, as these are blocking issues for our use of NH.

src/NHibernate.Test/NHSpecificTest/NH3918/FixtureByCode.cs
```diff
@@ -0,0 +1,127 @@
...
+		[Test]
+		public void EntityComparisonTest()
+		{
+			using (ISession session = OpenSession())
+			using (session.BeginTransaction())
+			{
+				var bob = session.Query<Owner>().Single(o => o.Name == "Bob");
+
+				var queryWithWhere = session.Query<Entity>()
+										.Where(WhereExpression(bob))
```
oskarb: Note for future - this indenting looks wrong depending on tab-size, because tab has been used also for alignment within a statement. Tabs are for indenting, space for alignment.

PR #526, opened 2016-11-14
NH-3874 - Make test pass for SqlServer CE.

hazzik:
SqlServer CE could not handle tables with single column which is an identity primary key.


hazzik:
@oskarb this fixes 
```text
FAIL    - NHibernate.Test.NHSpecificTest.NH3874.Fixture.EvictShallNotThrowWhenLoggingIsEnabled
```

PR #528, opened 2016-11-14
Driver/DriverBase: When cloning a parameter, also copy size/scale/pre…

oskarb:
…cision.

### DIFF
--- a/src/NHibernate/Driver/DriverBase.cs
+++ b/src/NHibernate/Driver/DriverBase.cs
@@ -267,6 +267,12 @@ namespace NHibernate.Driver
 			var clone = cmd.CreateParameter();
 			clone.DbType = originalParameter.DbType;
 			clone.ParameterName = originalParameter.ParameterName;
+			if (originalParameter.Size != 0)
+				clone.Size = originalParameter.Size;
+			if (originalParameter.Scale != 0)
+				clone.Scale = originalParameter.Scale;
+			if (originalParameter.Precision != 0)
+				clone.Precision = originalParameter.Precision;
 			clone.Value = originalParameter.Value;
 			return clone;
 		}


oskarb:
This causes NH3567.NH3567Tests.TestFlushModeAuto() to fail on ODBC with MSSQL Server 2016, with error:
[SQL: SELECT this_.Id as Id1_6548_0_, this_.Content as Con2_6548_0_, this_.PostId as Pos3_6548_0_ FROM Post this_ WHERE this_.Content = ?] ----> System.Data.Odbc.OdbcException : ERROR [42000] [Microsoft][SQL Server Native Client 11.0][SQL Server]The data types nvarchar and ntext are incompatible in the equal to operator.

The test uses an entity with a string property defined as length 2000 characters. Initially the parameter for the above SQL is defined with a size of 2000, but without the submitted change to CloneParameter that size is lost in the parameter instance that is actually used (so it will calculate the length as 6 characters itself). This works.

With the change, the size of 2000 is carried over to the cloned parameter. The parameter is initially treated as OdbcType.NVarChar, but eventually we end up in OdbcParameter's private method PrepareForBind() which does this:

```csharp
            int cbParameterSize = GetParameterSize(value, offset, ordinal);      // count of bytes for the data, for SQLBindParameter
            // here we upgrade the datatypes if the given values size is bigger than the types columnsize
            //
            switch(_bindtype._sql_type) {
                case ODBC32.SQL_TYPE.VARBINARY: // MDAC 74372
                    // Note: per definition DbType.Binary does not support more than 8000 bytes so we change the type for binding
                    if ((cbParameterSize > 8000))
                        { _bindtype = TypeMap._Image; } // will change to LONGVARBINARY
                    break;
                case ODBC32.SQL_TYPE.VARCHAR: // MDAC 74372
                    // Note: per definition DbType.Binary does not support more than 8000 bytes so we change the type for binding
                    if ((cbParameterSize > 8000))
                        { _bindtype = TypeMap._Text; }  // will change to LONGVARCHAR
                    break;
                case ODBC32.SQL_TYPE.WVARCHAR : // MDAC 75099
                    // Note: per definition DbType.Binary does not support more than 8000 bytes so we change the type for binding
                    if ((cbParameterSize > 4000))
                        { _bindtype = TypeMap._NText; }  // will change to WLONGVARCHAR
                    break;
}
```

Notice it calls GetParameterSize() to convert the size of 2000 characters to 4002 bytes (including two null bytes), which in the third case is or course larger than 4000 so it switches to the NText type, which cannot in fact be used quite like that in the query.

I don't know where exactly the limits of 8000 and 4000 used above come from. I cannot find them in any ODBC document, but of course we can recognize them as MSSQL servers maximum lengths in characters for varchar and nvarchar respectively. If that is the source of the constants it seems like a bug in OdbcParameter to compare the size in *bytes* of the unicode string to the maxmum size in *characters*.

Even if the above is a bug and that was fixed (which it probably never will be), we would get the same problem again if the string property is defined with a length of 4000 characters or more. To handle that case there would either need to be a way to make the database interpret the parameter as nvarchar(max), or a cast added in the SQL.

fredericDelaporte:
The `CloneParameter` logic has been changed as part of #703, in 612ea3653e8fee56c3d98805b02897966210f656. This renders this PR obsolete.

PR #439, opened 2015-07-21
Ingres9 sequences

gamblen:
Support for Sequence use within Ingres

PR #529, opened 2016-11-20
NH-3921 - Sequence support for Ingres9

### LINKED ISSUE #768: NH-3921 - Support sequences in Ingres9Dialect
**Oskar Berggren** created an [issue](https://nhibernate.jira.com/browse/NH-3921) — <time datetime="2016-11-20T10:40:54.661Z" title="2016-11-20T10:40:54.661Z">20th November 2016, 10:40:54</time>:
> Apparently Ingres supports sequences. This should be added to the dialect.
> 
> 5.0 since it might cause some identifier strategies to change behavior.
> 
---
**Oskar Berggren** added a comment — <time datetime="2016-11-20T10:41:17.593Z" title="2016-11-20T10:41:17.593Z">20th November 2016, 10:41:17</time>:
> A pull request is at https://github.com/nhibernate/nhibernate-core/pull/439
> 
---
**Oskar Berggren** added a comment — <time datetime="2016-11-20T10:47:08.961Z" title="2016-11-20T10:47:08.961Z">20th November 2016, 10:47:08</time>:
> Cleaned up pull request at https://github.com/nhibernate/nhibernate-core/pull/529
>

oskarb:
Cleaned up version of PR #439.
https://nhibernate.jira.com/browse/NH-3921
Fixes #768

### DIFF
--- a/src/NHibernate/Dialect/Ingres9Dialect.cs
+++ b/src/NHibernate/Dialect/Ingres9Dialect.cs
@@ -30,7 +30,12 @@ namespace NHibernate.Dialect
 		{
 			get { return true; }
 		}
-		
+
+		/// <summary>
+		/// Does this dialect support sequences?
+		/// </summary>
+		public override bool SupportsSequences => true;
+
 		/// <summary>
 		/// Attempts to add a <c>LIMIT</c> clause to the given SQL <c>SELECT</c>.
 		/// Expects any database-specific offset and limit adjustments to have already been performed (ex. UseMaxForLimit, OffsetStartsAtOne).
@@ -65,8 +70,28 @@ namespace NHibernate.Dialect
 			return pagingBuilder.ToSqlString();
 		}
 
+		public override string GetSequenceNextValString(string sequenceName)
+		{
+			return "select " + GetSelectSequenceNextValString(sequenceName) + " as seq";
+		}
+
+		public override string GetSelectSequenceNextValString(string sequenceName)
+		{
+			return "next value for " + sequenceName;
+		}
+
+		public override string GetCreateSequenceString(string sequenceName)
+		{
+			return "create sequence " + sequenceName;
+		}
+
+		public override string GetDropSequenceString(string sequenceName)
+		{
+			return "drop sequence " + sequenceName;
+		}
+
 		#region Overridden informational metadata
-		
+
 		public override bool DoesRepeatableReadCauseReadersToBlockWriters => true;
 
 		#endregion
--- a/src/NHibernate/Dialect/Ingres9Dialect.cs
+++ b/src/NHibernate/Dialect/Ingres9Dialect.cs
@@ -31,10 +31,14 @@ namespace NHibernate.Dialect
 			get { return true; }
 		}
 
-		/// <summary>
-		/// Does this dialect support sequences?
-		/// </summary>
+		/// <inheritdoc />
 		public override bool SupportsSequences => true;
+
+		/// <inheritdoc />
+		public override bool SupportsPooledSequences => true;
+
+		/// <inheritdoc />
+		public override string QuerySequencesString => "select seq_name from iisequences";
 
 		/// <summary>
 		/// Attempts to add a <c>LIMIT</c> clause to the given SQL <c>SELECT</c>.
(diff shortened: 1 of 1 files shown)


oskarb:
For 5.0 since it might change behaviour.

oskarb:
The failing test is DialectSupportingSequencesMustFullfillSequenceContract().

hazzik:
Dialect is missing implementations for `GetCreateSequenceString` and `GetDropSequenceString`

hazzik:
@oskarb shall we install Ingres on a build agent?

oskarb:
@hazzik I think the main concern about testing on Ingres or not is how much we should care about test failures, particularly in new test cases which are not specifically about Ingres.

It does slow us down to fix failures for specific dialects, given that often it's not really a problem in NHibernate itself but rather that the test (accidentally) tries to use something not supported in a specific dialect or database engine, or there is something silly like a rounding difference. On the other hand, it does obviously lead to better quality if the dialects are tested and problems fixed immediately.

What's the state of Ingres? With regards to what we test on today, it seems we are more likely to have cumbersome failures on platforms with a limited feature set (quite naturally).

src/NHibernate/Dialect/Ingres9Dialect.cs
```diff
@@ -65,10 +70,30 @@ public override SqlString GetLimitString(SqlString queryString, SqlString offset
...
 
+		public override string GetSequenceNextValString(string sequenceName)
+		{
+			return "select " + GetSelectSequenceNextValString(sequenceName) + " as seq";
+		}
+
+		public override string GetSelectSequenceNextValString(string sequenceName)
+		{
+			return "next value for " + sequenceName;
+		}
```
fredericDelaporte: Hibernate use the alternate syntax: `sequenceName + ".nextval"`. No indication in [Ingres 11 documentation](http://docs.actian.com/ingres/11.0/index.html#page/SQLRef/04_ElementsSQLStatements.04.085.htm) (nor in [10](http://docs.actian.com/ingres/10S/index.html#page/SQLRef/SQLRef_Body.1.0155.htm)) if one of these syntaxes is supported since longer.

I am not sure Hibernate is good source for this case though, since their `GetSequenceNextValString` syntax is not even documented in previous link (they do not call `GetSelectSequenceNextValString`, they return `"select nextval for " + sequenceName`

The way Ingres documentation is written let me think the "main" syntax is the one currently in this PR.

src/NHibernate/Dialect/Ingres9Dialect.cs
```diff
@@ -65,10 +70,30 @@ public override SqlString GetLimitString(SqlString queryString, SqlString offset
...
+
+		public override string GetCreateSequenceString(string sequenceName)
+		{
+			return "create sequence " + sequenceName;
+		}
+
+		public override string GetDropSequenceString(string sequenceName)
+		{
+			return "drop sequence " + sequenceName;
+		}
```
fredericDelaporte: Hibernate add after `sequenceName` the keyword `restrict`. Though [documented](http://docs.actian.com/ingres/11.0/index.html#page/SQLRef/08_SQLStatements.08.051.htm) only for other types of objects, it would be there for ignoring the drop if the sequence is referenced by another object in the database.

A bit conservative I think, and not done for other dialects, so I think above syntax is ok as currently is.

fredericDelaporte:
Rebased, and added a commit with my proposed minimal changes.

fredericDelaporte:
I realize a bit late this one should have been re-planed for 6.0, not 5.1, because it changes the `native`, `identity` and `enhanced-sequence` generator behavior. With Ingres, they will all start using a sequence. I may avoid the change for `native` and `identity` by PR an override of `IdentityStyleIdentifierGeneratorClass` and `NativeIdentifierGeneratorClass` in `Ingres9Dialect`, with a "6.0 TODO: remove". But for the `enhanced-sequence` generator, the only solution seems to tell users to add in its mapping the `force_table_use` parameter. So it will still be a possible breaking change.

So either we revert this on master, or we accept this possible breaking change.

hazzik:
>So either we revert this on master

It's not possible to undo mince.

I'm voting to the option of overriding `native` and `identity` and possible breaking changes with `force_table_use`

fredericDelaporte:
>>So either we revert this on master
>
>It's not possible to undo mince.

It is still doable with a revert PR, but adding noise in history. I was thinking about this option.

> I'm voting to the option of overriding `native` and `identity` and possible breaking changes with `force_table_use`

Ok, will do.

PR #530, opened 2016-11-20
NH-2285 - Support for LockMode in Linq provider

oskarb:
This is a rebased and updated version of PR #304 to implement support for setting lock mode on LINQ queries.

A notable difference is that the provided test cases have been improved to verify that the locking actually happens. This seems to show that they in fact do NOT take effect on certain dialects (at least MSSQL2008) when paging is also used. That may not be a flaw of this patch though, perhaps it's a deeper problem that affects HQL too?


oskarb:
@hazzik What would be the way to fix the caching issue exposed by QueryLock.CanChangeLockModeForQuery()?

oskarb:
It seems the locking is not applied when paging is also used in the query on MSSQL 2008 dialect.

hazzik:
I don't think that any of the databases support locking with paging.

gliljas:
I'd like to throw in this old thread into the discussion:
https://groups.google.com/d/topic/nhibernate-development/XqhDhDrILhc/discussion

gliljas:
I'll have another go at this.

src/NHibernate.Test/Linq/QueryLock.cs
```diff
@@ -0,0 +1,235 @@
...
+		{
+			using (session.BeginTransaction())
+			{
+				var result = (from e in db.Customers
+				              select e).Skip(5).Take(5).WithLock(LockMode.Upgrade).ToList();
+
+				Assert.That(result, Has.Count.EqualTo(5));
+				Assert.That(session.GetCurrentLockMode(result[0]), Is.EqualTo(LockMode.Upgrade));
+
+				if (!(Dialect is MsSql2000Dialect) && !(Dialect is MsSqlCeDialect))
```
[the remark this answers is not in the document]
hazzik: This would require 2 separate test dialect properties because those dialects do support locks in general, but they do not  support them in some limited circumstances (n subqueries, for example).

src/NHibernate.Test/Linq/QueryLock.cs
```diff
@@ -17,6 +19,11 @@ protected override bool AppliesTo(Dialect.Dialect dialect)
 			return TestDialect.SupportsSelectForUpdate;
 		}
 
+		protected override bool AppliesTo(ISessionFactoryImplementor factory)
+		{
+			return factory.ConnectionProvider.Driver is OdbcDriver;
```
fredericDelaporte: Only ODBC supports this? I guess a `!` is missing.

src/NHibernate.Test/Linq/QueryLock.cs
```diff
@@ -0,0 +1,248 @@
...
+			}
+		}
+
+		private void AssertSeparateTransactionIsLockedOut(string customerId)
+		{
+			using (new TransactionScope(TransactionScopeOption.Suppress))
+			using (var s2 = OpenSession())
+			using (s2.BeginTransaction())
+			{
+				// TODO: We should try to verify that the exception actually IS a locking failure and not something unrelated.
```
fredericDelaporte: It would be better, but it does seem critical to me.

src/NHibernate.Test/Linq/QueryLock.cs
hazzik: Yes, hahah

src/NHibernate.Test/Async/Linq/QueryLock.cs
```diff
@@ -29,7 +29,7 @@ public class QueryLockAsync : LinqTestCase
 	{
 		protected override bool AppliesTo(Dialect.Dialect dialect)
 		{
-			return TestDialect.SupportsSelectForUpdate;
+			return TestDialect.SupportsSelectForUpdate && !(dialect is SapSQLAnywhere17Dialect);
```
fredericDelaporte: Why not overriding the `TestDialect` property for it?

fredericDelaporte: Ok, I have now read the additional commit message. So, that is yet another property which would be required. But still this is strange: if it is waiting, it should end up with a timeout, shouldn't it?

hazzik: This behaviour is specific to this particular test case, so I did not want to add a property (initially I was going to add it, but decided against). The dialect itself claims that it supports "SELECT FOR UPDATE", however it does not support "nowait", which makes the check on this tests useless.

fredericDelaporte:
Oracle fails on the paging test because:
- With Oracle 10g dialect, used currently by TeamCity tests, the paging is done with a subquery, and the `for update of` references a column in the subquery while being out of the subquery. It does not support being into the subquery anyway, so it cannot be supported.
- With Oracle 12i dialect, the paging is done with `Fetch` syntax, avoiding a subquery, but as stated somewhere in the middle of [this huge wall of text](https://docs.oracle.com/en/database/oracle/oracle-database/12.2/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6):
  > **Restrictions on the row_limiting_clause**
  > This clause is subject to the following restrictions:
  > - You cannot specify this clause with the `for_update_clause`.

  So it cannot be supported too.

hazzik:
But I have Oracle10gTestDialect. What's wrong with it...

src/NHibernate.Test/Linq/QueryLock.cs
```diff
@@ -158,6 +158,7 @@ public void CanSetLockOnBothJoinAndMainComplex()
 		[Test]
 		public void CanSetLockOnLinqPagingQuery()
 		{
+			Assume.That(TestDialect.SupportsSelectForUpdateInSubQueries, Is.True, "Dialect does not support locking in subqueries");
```
fredericDelaporte: On my side I was adding a `TestDialect.SupportsSelectForUpdateWithPaging`. The fact that some dialects do implement paging with subqueries is an implementation detail which is not true for all dialects.

fredericDelaporte: In fact that property `SupportsSelectForUpdateInSubQueries` has been introduced just for this test case, so it should be renamed to something like `SupportsSelectForUpdateWithPaging`.

hazzik: Done

src/NHibernate.Test/Async/Linq/QueryLock.cs
fredericDelaporte: I have debugged it. And locking in `SybaseSQLAnywhere10Dialect` is actually bugged. I think we should flag `SupportsSelectForUpdate` as false for SAP SQL anywhere.

The trouble is, it uses the `for update by lock` syntax. This syntax is cursor specific, and silently results in "no-op" (no intent/update/exclusive locks acquired) as soon as it is used in a context where the engine considers there are no updatable cursor involved.

And that is what happen: when I breakpoint after the `for update` query, there are only shared locks in the db (seen with `SELECT * FROM sa_locks();`).
To actually get `intent` lock, the query has to be changed for using `with (updlock)` hint instead of the `for update`. The hint is not bound to cursor usage and do acquire locks whatever the query or cursor used. When using such a query, then I can see locks in the database.

So the database can support locking rows, but its dialects currently fail to do so.

hazzik: Ok

hazzik: >So the database can support locking rows, but its dialects currently fail to do so.

Needs to be fixed sometime. Can you create a ticket? (I'm going to sleep now).

fredericDelaporte: #1915, good night.

fredericDelaporte:
@hazzik, I am not merging this as I do not know how you intent this to be merged. I would tend to squash and add Oskar and you as co-authors, but as you have rebased this to three commits, I wonder if you were willing to merge this as-is.

hazzik:
I'm just not sure who will be the main author of the commit when it's got squashed: the PR is made by Oskar with the changes originated from Onur.

fredericDelaporte:
It would surprise me if it will not be the first commit author, so Onur.

fredericDelaporte:
Well looking into a past example, #1492 merging seems to have taken into account the PR author, not the first commit author. It is the same for #460. So the author would be Oskar in fact.

hazzik:
Yes, I was going to investigate this first (what you've already done).

PR #533, opened 2016-11-22
Changed license file name

baderbuddy:
Changed license file name in order for GitHub to be able to auto detect the license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,15 +1,14 @@
                   GNU LESSER GENERAL PUBLIC LICENSE
-                     Version 2.1, February 1999
+                       Version 2.1, February 1999
 
-       Copyright (C) 1991, 1999 Free Software Foundation, Inc.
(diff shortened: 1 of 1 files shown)


oskarb:
Thanks for the patch.

Build log error:
[18:54:50][Step 1/1] Could not find file 'D:\BuildAgent\work\b0ecfa5dcda2234\lgpl.txt' to copy.

Please do a full file content source over all source code to find any references to the lgpl.txt file and adjust accordingly.

oskarb:
Actually there is also the gfdl.txt. By the name of the license we can of course guess about the intention, but I'm not sure if anything explicitly references that license. Perhaps that should be clarified in various places as necessary.

Maybe that means the LICENSE file should summarize what licenses are used, rather than just replicate one of them?

hazzik:
Closed as per @oskarb

PR #534, opened 2016-11-27
NH-3913 Component Property Customizers Not Applied

AndrewShepherd:
Fix for [NH-3913](https://nhibernate.jira.com/browse/NH-3913)

The data relationship is Entity => ParentComponent => Bag => ChildComponent.

The ChildComponent had a customizer on one of its properties.

During the mapping specification, the class mapper stored the customizer against the full path (Entity.ParentComponent.Bag.ChildComponent)

The generation process was only searching for the customizer against a partial path (ParentComponent.Bag.ChildComponent)

This pull request changes the generation process so it uses the full path to search for a property's customizers.


src/NHibernate/Mapping/ByCode/ModelMapper.cs
```diff
@@ -1327,19 +1327,19 @@ private IMapKeyRelationMapper DetermineMapKeyRelationType(MemberInfo member, Pro
...
 			private readonly System.Type ownerType;
 
-			public ComponentRelationMapper(MemberInfo collectionMember, System.Type ownerType, System.Type componentType, ICandidatePersistentMembersProvider membersProvider,
+			public ComponentRelationMapper(PropertyPath propertyPath, System.Type componentType, ICandidatePersistentMembersProvider membersProvider,
 			                               IModelInspector domainInspector, ICustomizersHolder customizersHolder, ModelMapper modelMapper)
 			{
-				this.collectionMember = collectionMember;
-				this.ownerType = ownerType;
+				this.propertyPath = propertyPath;
+				this.ownerType = propertyPath.LocalMember.DeclaringType;
```
hazzik: This changes the behaviour. Above ownerType initialized with ReflectedType, and here the DeclaringType is used. Is this change necessary?

AndrewShepherd: I don't believe it changes the behavior. Before the change, the `ComponentRelationParameter` constructor was being passed
- A `property` parameter of type `MemberInfo`
- An `ownerType` parameter of type 'System.Type'
I've replaced both of these parameters with
- A propertyPath parameter of type `PropertyPath`

I was under the impression that `PropertyPath` parameter already had the ownerType in the field `propertyPath.DeclaringMember.OwnerType`, so it was redundant to pass this in as a parameter.
There doesn't seem be a scenario where `propertyPath.DeclaringMember.OwnerType` differs from the `ownerType`.
Can you think of a possible scenario? If so, we should capture it in an automated test.

On another note, I'm trying to read the tests I wrote last October, and I think they're too hard to read. I refactored them too much. 

I should change them - introduce some duplication into each test, just so it's easier to follow through the steps that are being taken.

AndrewShepherd: I have found a scenario where `propertyPath.DeclaringMember.OwnerType` and `ownerType` is different: when you are mapping a member that's inherited from a base class.

I tried to write a test to reveal a defect that comes about when they are different, and I was successful. Unfortunately, this newly failing test didn't get fixed by passing the `ownerType` to `ComponentRelationMapper`.

I've pushed the failing test to this pull request - obviously I should come up with a fix now.

AndrewShepherd: I have figured out where `propertyPath.DeclaringMember.OwnerType` differs from the `ownerType`, and that is where you are mapping inherited members.
I've changed the code so it now passes the ownerType explicitly

PR #535, opened 2016-11-28
NH-3877 Target Net 4.6.2 and upgrade solution files to VS2015

oskarb:


oskarb:
I take it you mean ilrepack?

hazzik:
Yes, that one

oskarb:
I added a little something for ilrepack - not really sure if it's correct or the best way to do it.

hazzik:
Looks legit

igitur:
Hope it's OK if I add on to this issue. I see this is the issue that removed `net40` support. For me, that's OK, but is supporting for `net452` not possible with multi-targeting in VS2017?

fredericDelaporte:
The code uses some classes not available in `net452`, like `AsyncLocal<T>`.

PR #537, opened 2016-12-04
NH-3978 - Extract IDatabaseMetadata from DatabaseMetadata

CodeWithCream:
Extracting IDatabaseMetadata from DatabaseMetadata

https://nhibernate.jira.com/browse/NH-3978


oskarb:
Hi,

Is there an issue in Jira?
What's the motivation for this?
Other changes have been done in Hibernate, which doesn't even have this file since March 2015. Might be useful to look into if those changes can be ported.

CodeWithCream:
Hi Oscar, 
no there is no JIRA issue for this change. My motivation for doing this is a little selfish ;)
About 2 years ago we made a custom implementation of upgrading database schema because we didn't want to depend on database connection. Our solution uses an old model .dll and mapping xml and a new model .dll and using the nhibernate implementation for generating  update scripts we generate them without connecting to the database because we have migrations (with fluent migrator) and we cannot generate database upgrade script only between current and the last version because some migrations could use columns and tables that doesn't exist anymore. So we decided to make update scripts for each model change and when we upgrade the database we execute that scripts one by one combining them with the migration of the same version.
To do that we had to download nhibernate source code and make our own CustomDatabaseMetadata implementation so we can fill its data manually after getting them from the mapping file. I looked at the code and realized that the only thing I have to change is extracting IDatabaseMetadata from DatabaseMetadata and change the implementation of the GenerateSchemaUpdateScript so that it uses interface instead of the concrete class because scripts generation doesn't care of how DatabaseMetadata was filled.
I taught that maybe more people would like to have their custom utilities without having to maintain their own nhibernate source (we had a problem when we had to upgrade to nhibernate 4 because we had to implement it one more time) and I saw that ITableMetadata was already extracted so I decided to change a code a little bit because it means a lot and don't change a main implementation at all. 

I can show you our update script generator implementation (I have a plan to write a blog post about it because I thing it's a really great thing) but I'll have to simplify it a little bit or show you only an important part of the whole solution...

Can you explain a little bit what your last sentence means? I've only worked with nhibernate, so I don't know a Hibernate code...

fredericDelaporte:
NHibernate is a fork from Hibernate and a lot of its code is ported from Hibernate. The corresponding Hibernate file for DatabaseMetaData was removed in [this commit](https://github.com/hibernate/hibernate-orm/commit/9caca0ce37d5a2763d476c6fa2471addcca710ca). Prior to that removal, it had received changes which were not ported to NHibernate. Its latest version is [here](https://github.com/hibernate/hibernate-orm/blob/4bf3b162ad00ee977d65c02ef0f6db6fe29bde85/hibernate-core/src/main/java/org/hibernate/tool/hbm2ddl/DatabaseMetadata.java). Maybe there was other interesting thing to port from it. As you look quite involved in using those features, it mat be of interest to you.

We need a Jira for tracking changes and managing our releases. Could please add one?
https://nhibernate.jira.com

hazzik:
I created a JIRA. [NH-3978](https://nhibernate.jira.com/browse/NH-3978)

PR #540, opened 2016-12-04
NH-3900 - Prework for NUnit 3, but still on NUnit 2.x

ngbrown:
I have another pull request, #508 for updating to NUnit 3 for [NH-3900](https://nhibernate.jira.com/browse/NH-3900), this is as much pre-changes that can be done without updating all the way to NUnit 3.

I am going to push the Linq tests split after the build server starts so that there is a baseline without the split, because a bunch of stuff will be "new" failures because it's in a different place.


oskarb:
Please hold the LINQ test split to a separate PR after this one is merged.

PR #542, opened 2016-12-07
Additional check if child session has already been closed during parent session closing.

itmagination:
This simple fix prevents unnecessary SessionException("Session was already closed") throw when both parent session and child session are closed separately (order does not matter). Before invoking close on child session (during parent session close) we should check if child session has not been closed already.


gliljas:
Thanks!

PR #543, opened 2016-12-12
NH-4022 - Patch for Deletion of SQL Server Sequences

Eustachius:
[NH-4022](https://nhibernate.jira.com/browse/NH-4022) - Deletion of a sequence fails if the sequence does not reside in the default schema, since schema and name are stored in different columns of sys.sequences.

### DIFF
--- a/src/NHibernate/Dialect/MsSql2012Dialect.cs
+++ b/src/NHibernate/Dialect/MsSql2012Dialect.cs
@@ -30,7 +30,7 @@ namespace NHibernate.Dialect
 
 		public override string GetDropSequenceString(string sequenceName)
 		{
-			string dropSequence = "IF EXISTS (select * from sys.sequences where name = N'{0}') DROP SEQUENCE {0}";
+			string dropSequence = "if object_id('{0}') is not null drop sequence {0}";
 
 			return string.Format(dropSequence, sequenceName);
 		}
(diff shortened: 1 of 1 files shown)


fredericDelaporte:
May you provide a Jira please? And test cases would be great too.

hazzik:
Actually, this wouldn't work. There are 2 options: leave it as is, or re-implement together with #448

src/NHibernate/Dialect/MsSql2012Dialect.cs
```diff
@@ -30,7 +30,7 @@ protected override string GetCreateSequenceString(string sequenceName, int initi
 
 		public override string GetDropSequenceString(string sequenceName)
 		{
-			string dropSequence = "IF EXISTS (select * from sys.sequences where name = N'{0}') DROP SEQUENCE {0}";
+			string dropSequence = "if object_id(N'{0}') is not null drop sequence {0}";
```
hazzik: This should read as `IF EXISTS (select * from {catalog}.sys.sequences where object_id =object_id(N'{0}')) ...`

Taking into the consideration catalog+default_catalog, schema+default_schema.

hazzik:
Replaced by #647

Eustachius:
@hazzik: Thanks for including the change and updating the test suite ...

PR #545, opened 2017-01-09
NH-3932 - Merge() may fire unnecessary updates

RogerKratz:
[NH-3932](https://nhibernate.jira.com/browse/NH-3932)

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3932/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3932/Fixture.cs
@@ -1,0 +1,209 @@
+﻿using System.Collections;
+using NHibernate.Test.NHSpecificTest.NH3932.Model;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3932
+{
+	public abstract class Fixture : BugTestCase
+	{
+		private IParent storedParent;
+		protected abstract bool CareAboutOrder { get; }
+
+		[Test]
+		public void ShouldKeepDirtyCollectionDirtyAfterMergingClone()
+		{
+			using (var s = OpenSession())
+			{
+				using (s.BeginTransaction())
+				{
+					s.Lock(storedParent, LockMode.None);
+					storedParent.ClearChildren();
+					Assert.IsTrue(s.IsDirty());
+
+					var loadedClone = storedParent.Clone();
+					s.Merge(loadedClone);
+					Assert.IsTrue(s.IsDirty());
+				}
+			}
+		}
+
+		[Test]
+		public void ShouldCareAboutOrder_MergeObjectNotInSession()
+		{
+			var parent = CreateParent(2);
+			using (var s = OpenSession())
+			{
+				using (var tx = s.BeginTransaction())
+				{
+					s.Save(parent);
+					tx.Commit();
+				}
+			}
+			var parentClone = parent.Clone();
+			parentClone.ReverseChildren();
+			using (var s = OpenSession())
+			{
+				s.SessionFactory.Statistics.Clear();
+				using (var tx = s.BeginTransaction())
+				{
+					s.Merge(parentClone);
+					tx.Commit();
+				}
+				Assert.That(s.SessionFactory.Statistics.EntityUpdateCount, CareAboutOrder ? Is.EqualTo(1) : Is.EqualTo(0));
+			}
+		}
+
+		[Test]
+		public void ShouldCareAboutOrder_MergeObjectInSession()
+		{
+			var parent = CreateParent(2);
+			using (var s = OpenSession())
+			{
+				using (var tx = s.BeginTransaction())
+				{
+					s.Save(parent);
+					tx.Commit();
+				}
+			}
+			var parentClone = parent.Clone();
+			parentClone.ReverseChildren();
+			using (var s = OpenSession())
+			{
+				s.SessionFactory.Statistics.Clear();
+				using (var tx = s.BeginTransaction())
+				{
+					s.Lock(parent, LockMode.None);
+					s.Merge(parentClone);
+					tx.Commit();
+				}
+				Assert.That(s.SessionFactory.Statistics.EntityUpdateCount, CareAboutOrder ? Is.EqualTo(1) : Is.EqualTo(0));
+			}
+		}
+
+		[Test]
+		public void MergeCleanCloneShouldNotResultInUpdate()
+		{
+			var parentClone = storedParent.Clone();
+			using (var s = OpenSession())
+			{
+				s.SessionFactory.Statistics.Clear();
+				using (var tx = s.BeginTransaction())
+				{
+					s.Merge(parentClone);
+					tx.Commit();
+				}
+				Assert.That(s.SessionFactory.Statistics.EntityUpdateCount, Is.EqualTo(0));
+			}
+		}
+
+		[Test]
+		public void MergeCleanCloneShouldNotMakeSessionDirty()
+		{
+			var parentClone = storedParent.Clone();
+			using (var s = OpenSession())
+			{
+				s.SessionFactory.Statistics.Clear();
+				using (var tx = s.BeginTransaction())
+				{
+					s.Merge(parentClone);
+					Assert.That(s.IsDirty(), Is.False);
+					tx.Commit();
+				}
+			}
+		}
+
+		[Test]
+		public void MergeCloneOnNonCloneShouldNotResultInUpdate()
+		{
+			var parentClone = storedParent.Clone();
+			using (var s = OpenSession())
+			{
+				s.SessionFactory.Statistics.Clear();
+				using (var tx = s.BeginTransaction())
+				{
+					s.Lock(storedParent, LockMode.None);
+					s.Merge(parentClone);
+					tx.Commit();
+				}
+				Assert.That(s.SessionFactory.Statistics.EntityUpdateCount, Is.EqualTo(0));
+			}
+		}
+
+		[Test]
+		public void MergeCloneOnNonCloneShouldNotMakeSessionDirty()
+		{
+			var parentClone = storedParent.Clone();
+			using (var s = OpenSession())
+			{
+				s.SessionFactory.Statistics.Clear();
+				using (var tx = s.BeginTransaction())
+				{
+					s.Lock(storedParent, LockMode.None);
+					s.Merge(parentClone);
+					Assert.That(s.IsDirty(), Is.False);
+					tx.Commit();
+				}
+			}
+		}
+
+		[Test]
+		public void MoreElementsInTargetShouldBeTreatedAsDirty()
+		{
+			var parent = CreateParent(2);
+			using (var s = OpenSession())
+			{
+				using (var tx = s.BeginTransaction())
+				{
+					s.Save(parent);
+					tx.Commit();
+				}
+			}
+			var parentClone = parent.Clone();
+			parentClone.RemoveLastChild();
+			using (var s = OpenSession())
+			{
+				s.SessionFactory.Statistics.Clear();
+				using (var tx = s.BeginTransaction())
+				{
+					s.Merge(parentClone);
+					tx.Commit();
+				}
+				Assert.That(s.SessionFactory.Statistics.EntityUpdateCount, Is.EqualTo(1));
+			}
+		}
+
+		protected abstract IParent CreateParent(int numberOfChildren);
+
+		protected override void OnSetUp()
+		{
+			storedParent = CreateParent(1);
+			using (var s = OpenSession())
+			{
+				using (var tx = s.BeginTransaction())
+				{
+					s.Save(storedParent);
+					tx.Commit();
+				}
+			}
+		}
+
+		protected override void OnTearDown()
+		{
+			using (var s = sessions.OpenSession())
+			{
+				using (var tx = s.BeginTransaction())
+				{
+					s.Delete("from " + storedParent.GetType());
+					tx.Commit();
+				}
+			}
+		}
+
+		protected override void Configure(Cfg.Configuration configuration)
+		{
+			configuration.SetProperty(Cfg.Environment.GenerateStatistics, "true");
+		}
+
+		protected override IList Mappings => new[]{"NHSpecificTest." + BugNumber + ".Model.Mappings.hbm.xml"};
+	}
+}
--- a/src/NHibernate/Type/CollectionType.cs
+++ b/src/NHibernate/Type/CollectionType.cs
@@ -9,6 +9,7 @@ using NHibernate.Proxy;
 using NHibernate.SqlTypes;
 using NHibernate.Util;
 using System.Collections.Generic;
+using System.Linq;
 using NHibernate.Impl;
 
 namespace NHibernate.Type
@@ -359,36 +360,54 @@ namespace NHibernate.Type
 			return result;
 		}
 
-		public virtual object ReplaceElements(object original, object target, object owner, IDictionary copyCache,
-											  ISessionImplementor session)
+		public virtual object ReplaceElements(object original, object target, object owner, IDictionary copyCache, ISessionImplementor session)
 		{
-			// TODO: does not work for EntityMode.DOM4J yet!
-			object result = target;
-			Clear(result);
+			var elemType = GetElementType(session.Factory);
+			var targetPc = target as IPersistentCollection;
+			var originalPc = original as IPersistentCollection;
+			var iterOriginal = (IEnumerable)original;
+			var clearTargetsDirtyFlag = ShouldTargetsDirtyFlagBeCleared(targetPc, originalPc, iterOriginal);
 
 			// copy elements into newly empty target collection
-			IType elemType = GetElementType(session.Factory);
-			IEnumerable iter = (IEnumerable)original;
-			foreach (object obj in iter)
+			Clear(target);
+			foreach (var obj in iterOriginal)
 			{
-				Add(result, elemType.Replace(obj, null, session, owner, copyCache));
+				Add(target, elemType.Replace(obj, null, session, owner, copyCache));
 			}
 
-			// if the original is a PersistentCollection, and that original
-			// was not flagged as dirty, then reset the target's dirty flag
-			// here after the copy operation.
-			// One thing to be careful of here is a "bare" original collection
-			// in which case we should never ever ever reset the dirty flag
-			// on the target because we simply do not know...
-			IPersistentCollection originalPc = original as IPersistentCollection;
-			IPersistentCollection resultPc = result as IPersistentCollection;
-			if(originalPc != null && resultPc!=null)
+			if(clearTargetsDirtyFlag)
+			{
+				targetPc.ClearDirty();
+			}
+
+			return target;
+		}
+
+		internal bool ShouldTargetsDirtyFlagBeCleared(IPersistentCollection targetPc, IPersistentCollection originalPc, IEnumerable original)
+		{
+			if (targetPc == null)
+				return false;
+
+			if (originalPc == null)
+			{
+				if (!targetPc.IsDirty && AreCollectionElementsEqual(original, (IEnumerable)targetPc))
+				{
+					return true;
+				}
+			}
+			else
 			{
 				if (!originalPc.IsDirty)
-					resultPc.ClearDirty();
+				{
+					return true;
+				}
 			}
+			return false;
+		}
 
-			return result;
+		protected virtual bool AreCollectionElementsEqual(IEnumerable original, IEnumerable target)
+		{ 
+			return original.Cast<object>().SequenceEqual(target.Cast<object>());
 		}
 
 		public IType GetElementType(ISessionFactoryImplementor factory)
--- a/src/NHibernate/Type/GenericBagType.cs
+++ b/src/NHibernate/Type/GenericBagType.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using NHibernate.Collection;
 using NHibernate.Collection.Generic;
 using NHibernate.Engine;
@@ -69,6 +71,15 @@ namespace NHibernate.Type
 		public override object Instantiate(int anticipatedSize)
 		{
 			return anticipatedSize <= 0 ? new List<T>() : new List<T>(anticipatedSize + 1);
+		}
+
+		protected override bool AreCollectionElementsEqual(IEnumerable original, IEnumerable target)
+		{
+			var first = (IEnumerable<T>)original;
+			var second = (IEnumerable<T>) target;
+
+			return first.Count() == second.Count() && 
+				first.All(element => first.Count(x => x.Equals(element)) == second.Count(x => x.Equals(element)));
 		}
 	}
 }
--- a/src/NHibernate/Type/GenericIdentifierBagType.cs
+++ b/src/NHibernate/Type/GenericIdentifierBagType.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using NHibernate.Collection;
 using NHibernate.Collection.Generic;
 using NHibernate.Engine;
@@ -81,6 +83,74 @@ namespace NHibernate.Type
 		protected override void Add(object collection, object element)
 		{
 			((IList<T>)collection).Add((T)element);
+		}
+
+		protected override bool AreCollectionElementsEqual(IEnumerable original, IEnumerable target)
+		{
+			var first = (IList<T>)original;
+			var second = (IList<T>)target;
+
+			return first.Count == second.Count &&
+			       first.All(element => first.Count(x => x.Equals(element)) == second.Count(x => x.Equals(element)));
+		}
+
+		public override object ReplaceElements(
+			object original,
+			object target,
+			object owner,
+			IDictionary copyCache,
+			ISessionImplementor session)
+		{
+			var elemType = GetElementType(session.Factory);
+			var targetPc = target as PersistentIdentifierBag<T>;
+			var originalPc = original as IPersistentCollection;
+			var iterOriginal = (IList<T>)original;
+			var clearTargetsDirtyFlag = false;
+			var clearTarget = true;
+
+			if (targetPc != null)
+			{
+				if (originalPc == null)
+				{
+					if (!targetPc.IsDirty && AreCollectionElementsEqual(iterOriginal, targetPc))
+					{
+						clearTargetsDirtyFlag = true;
+						clearTarget = false;
+					}
+				}
+				else
+				{
+					if (!originalPc.IsDirty)
+					{
+						clearTargetsDirtyFlag = true;
+					}
+				}
+			}
+
+			if (clearTarget)
+			{
+				Clear(target);
+				foreach (var obj in iterOriginal)
+				{
+					Add(target, elemType.Replace(obj, null, session, owner, copyCache));
+				}
+			}
+			else
+			{
+				for (var i = 0; i < targetPc.Count; i++)
+				{
+					var currTarget = targetPc[i];
+					var orgToUse = iterOriginal.First(x => currTarget.Equals(x));
+					targetPc[i] = (T)elemType.Replace(orgToUse, null, session, owner, copyCache);
+				}
+			}
+
+			if (clearTargetsDirtyFlag)
+			{
+				targetPc.ClearDirty();
+			}
+
+			return target;
 		}
 	}
 }
--- a/src/NHibernate/Type/GenericMapType.cs
+++ b/src/NHibernate/Type/GenericMapType.cs
@@ -78,25 +78,26 @@ namespace NHibernate.Type
 
 		public override object ReplaceElements(object original, object target, object owner, IDictionary copyCache, ISessionImplementor session)
 		{
-			ICollectionPersister cp = session.Factory.GetCollectionPersister(Role);
+			var cp = session.Factory.GetCollectionPersister(Role);
 
-			IDictionary<TKey, TValue> result = (IDictionary<TKey, TValue>)target;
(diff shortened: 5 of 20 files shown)


RogerKratz:
Contains tests + fix

src/NHibernate/Type/CollectionType.cs
```diff
@@ -391,14 +392,36 @@ public virtual object InstantiateResult(object original)
...
-				if (!originalPc.IsDirty)
+				if (elementsEqual || (originalPc != null && !originalPc.IsDirty))
 					resultPc.ClearDirty();
 			}
 
-			return result;
+			return target;
+		}
+
+		private static bool AreCollectionElementsEqual(IEnumerable originalEnumerator, IEnumerator targetEnumerator, IType elementType, ISessionImplementor session)
```
hazzik: Hey, @RogerKratz

RogerKratz: @hazzik Thanks for your review.
Sure, I'll update PR soon.
(on the other hand makes current PR it work in >95% of the cases, it's only a problem if you on a detached set remove an element and put it back again...)

RogerKratz:
I've updated PR with test + impl for special set case. 
Also, I've simplified collectiontype's impl a bit. 
Commits are squashed.

hazzik:
@RogerKratz tests are failing for SqlServer Compact. It's because Child has only Id property. Can you please add at least one other property, "Name", for eg.?

RogerKratz:
@hazzik Ok, didn't know that. I've updated the PR.

RogerKratz:
@hazzik The "changes requested" mentioned in this PR are fixed, right?

src/NHibernate/Type/GenericSetType.cs
```diff
@@ -76,5 +77,20 @@ public override object Instantiate(int anticipatedSize)
 		{
 			return new HashSet<T>();
 		}
+
+		protected override bool AreCollectionElementsEqual(IEnumerable original, IEnumerable target)
```
fredericDelaporte: [`ISet.SetEquals`](https://msdn.microsoft.com/en-us/library/dd412096(v=vs.110).aspx) does not require its argument to be a set. It would be slightly more efficient to rebuild target as a list instead of as a set.

(There is a `SetEquals` on `CollectionHelper` from `NHibernate.Util` namespace too. A bit redundant with .Net `ISet.SetEquals`, but NHibernate is quite older than sets in .Net.)

src/NHibernate/Type/CollectionType.cs
```diff
@@ -391,14 +392,36 @@ public virtual object InstantiateResult(object original)
 			// in which case we should never ever ever reset the dirty flag
 			// on the target because we simply do not know...
 			IPersistentCollection originalPc = original as IPersistentCollection;
-			IPersistentCollection resultPc = result as IPersistentCollection;
-			if(originalPc != null && resultPc!=null)
+			IPersistentCollection resultPc = target as IPersistentCollection;
+			if(resultPc!=null)
 			{
-				if (!originalPc.IsDirty)
+				if (elementsEqual || (originalPc != null && !originalPc.IsDirty))
```
fredericDelaporte: I do not see why this would be valid. Why original equality with target warrants target to be non-dirty, especially if original was dirty?  
Reading [NH-3932](https://nhibernate.jira.com/browse/NH-3932), this PR is about avoiding unneeded updates when the original is not a persistent collection. This is precisely the case the comment just above those lines writes we should never ever reset the dirty flag, cause we do not know... If this comment is just plain wrong, it should be amended. But I do not believe it to be wrong. It looks to me this change will optimize some cases will breaking other cases.

At the very least, condition for reseting the dirty flag should be hardened in my opinion. Maybe something like:

- check before clearing target if it was already flagged dirty (save a `targetDirty` variable, value irrelevant if target is not a persistent collection),
- if original is a `IPersistentCollection`:
  - if original is not dirty, reset dirty.
- **else** (original not a `IPersistentCollection`) if elements were equals and target was not dirty, reset dirty.

But even this, I am not sure it would be valid for all cases.

RogerKratz: You are correct I am sure. To be honest, I don't know these parts of NH that well. I did a fix that solved my test cases and saw that all tests were green.

Can you "translate" your doubts into a failing test? Or can you point me in the right direction here? Seems there are some tests missing here...

src/NHibernate/Type/CollectionType.cs
```diff
@@ -391,14 +392,36 @@ public virtual object InstantiateResult(object original)
...
 					resultPc.ClearDirty();
 			}
 
-			return result;
+			return target;
+		}
+
+		protected virtual bool AreCollectionElementsEqual(IEnumerable original, IEnumerable target)
```
RogerKratz: Ok. I'll be glad to change it if necessary.

src/NHibernate/Type/GenericSetType.cs
RogerKratz: Ok. I'll be glad to change it if necessary.

src/NHibernate/Type/CollectionType.cs
fredericDelaporte: Well, maybe something like:

- Load an entity with a collection, change some elements in the collection.
- Instantiate a detached replica of the entity, with its collection filled with a copy of the modified one.
- Merge the detached entity into the session.

With your change, I think the entity would then have its collection flagged as non dirty, while it should be dirty.

src/NHibernate/Type/GenericSetType.cs
fredericDelaporte: I am just a contributor as you, I will not decide if this is necessary. Those comments are just my thoughts.

RogerKratz:
@fredericDelaporte 
About your suggested test case...
Do I understand you correct that you were looking for something like...
1. Load entity
2. Modify entity's collection elements
3. Make a replica/clone of 2
4. Merge 3 into session where 2 exists
...?

If so, why do you think the session should become dirty?

fredericDelaporte:
2. Modify without flushing. So the entity collection is dirty. It should stay dirty after the merge.

RogerKratz:
Ok, I see what you mean (I think ;)).

....however, I cannot get the test to fail. I've "translated" what you suggested to this...


```
[Test]
public void Foo()
{
	using (var s = OpenSession())
	{
		using (s.BeginTransaction())
		{
			var loaded = s.Get<SetParent>(((SetParent) StoredParent).Id);
			loaded.Children.First().Name = "modified";
			Assert.IsTrue(s.IsDirty());

			var loadedClone = loaded.Clone();
			s.Merge(loadedClone);
			Assert.IsTrue(s.IsDirty());
		}
	}
}
```
...but test is green (SetParent.Clone makes a copy of the collection).

RogerKratz:
....I've changed the AreCollectionEquals impl as you suggested.

fredericDelaporte:
Thanks for the test,
Some changes are required.
That is the collection which should be changed, not an element of the collection. Collections have their own persister, and their own dirtiness. You should add or remove a child into it.
Then the test should assert collection dirtiness, not the whole session dirtiness since the children would probably be dirty too. (Using a many-to-many collection would allow to dodge this.)

RogerKratz:
Thanks. I change the test to clear the collection in "2" then it fails as you expected.
I have a look to see if I can make it pass.

RogerKratz:
Just pushed test + fix for "your" issue.

RogerKratz:
@fredericDelaporte 
I just pushed a small perf fix, not checking "elemets equality" if not necessary.
Let me know if you want me to do something more.

src/NHibernate/Type/CollectionType.cs
```diff
@@ -369,19 +370,36 @@ public virtual object InstantiateResult(object original)
...
+					}
+				}
+			}
 
+			Clear(target);
 			// copy elements into newly empty target collection
-			IType elemType = GetElementType(session.Factory);
-			IEnumerable iter = (IEnumerable)original;
-			foreach (object obj in iter)
+			var iter = (IEnumerable)original;
```
fredericDelaporte: Please move up this line for avoiding casting it too when calling `AreCollectionElementsEqual`.

src/NHibernate/Type/CollectionType.cs
```diff
@@ -390,15 +408,17 @@ public virtual object InstantiateResult(object original)
...
+				resultPc.ClearDirty();
 			}
 
-			return result;
+			return target;
+		}
+
+		protected virtual bool AreCollectionElementsEqual(IEnumerable original, IEnumerable target)
+		{ 
+			return CollectionHelper.CollectionEquals<object>(original.Cast<object>().ToArray(), target.Cast<object>().ToArray());
```
fredericDelaporte: I was not anticipating this `Cast<object>` with array conversion, which results in a double enumeration then followed by CollectionEquals own enumeration.
Please use `Enumerable.SequenceEqual` instead of `CollectionHelper.CollectionEquals`. This would avoid the `ToArray`. (And at that point, neither original nor target are supposed to be null, `SequenceEqual` can be safely used here without rechecking null.)

src/NHibernate/Type/CollectionType.cs
fredericDelaporte: This method ends up inherited by `GenericMapType`, `GenericBagType`, `GenericIdentifierBagType` too. As for the `GenericSetType`, they are not sensitive to order (if they contains the same elements, they should be considered equal even if the order is not the same).

Maps requires checking they have the same elements count, that each key of one of them are found in the other one with the same value associated. It should be doable as an O(n) algorithm.

Bags allow duplicates. Checking two bags are equal implies checking they contain the same set of elements whatever their order, and counting how many time each element appears in each bags and checking the counts are the same. O(n) complexity is probably still achievable by building two map of their elements to their duplication count and then comparing the two maps.

@hazzik, too much added complexity?

Should we rename this method `AreCollectionsUndoubtedlyEqual` instead and just return false for those other types?

src/NHibernate/Type/CollectionType.cs
```diff
@@ -369,19 +370,36 @@ public virtual object InstantiateResult(object original)
...
+			var resultPc = target as IPersistentCollection;
+			var originalPc = original as IPersistentCollection;
+			var clearTargetsDirtyFlag = false;
+			if (resultPc != null)
+			{
+				if (originalPc == null)
+				{
+					if (!resultPc.IsDirty && AreCollectionElementsEqual((IEnumerable)original, (IEnumerable)target))
+					{
+						clearTargetsDirtyFlag = true;
```
RogerKratz: Ok. Will be pushed soon.
(
IMHO, I'm not really a big fan of (code) comments at all. Red tests are the only "comments" that work I believe... But in this case a red test cannot be written for this so... I sort of agree ;)
)

src/NHibernate/Type/CollectionType.cs
RogerKratz: Sure.

src/NHibernate/Type/CollectionType.cs
```diff
@@ -390,15 +408,17 @@ public virtual object InstantiateResult(object original)
 			// One thing to be careful of here is a "bare" original collection
 			// in which case we should never ever ever reset the dirty flag
 			// on the target because we simply do not know...
```
RogerKratz: Will delete it.

src/NHibernate/Type/CollectionType.cs
RogerKratz: Ok

RogerKratz:
Thanks. Updated CollectionType following your suggestions.

RogerKratz:
@fredericDelaporte 
About your comment about different collection types... (cannot comment that comment for some reason so I write it down here). The default impl checks for equal elements in same order. This is overriden for set collections. 
Now, as you mention, other collection types like bags and maps could be impl "smarter" because they're not ordered. But as I see it, it's nothing that needs to be impl now-now. The default impl will, in worst case for some collection types, lead to unnecessary updates but no worse than before this fix.

About your other comments and suggestions... They're impl in this PR now.

fredericDelaporte:
Yes I have seen your changes, thanks. About the other collection types having a different equal semantic, since that was Hazzik who had seen that first for the set case and mandated a change, I wait for his point of view.

src/NHibernate/Type/GenericSetType.cs
```diff
@@ -76,5 +78,15 @@ public override object Instantiate(int anticipatedSize)
...
+		{
+			var orgSet = new HashSet<object>();
```
hazzik: This can be just `return ((ISet<T>) original).SetEquals((IEnumerable<T>) target);`

src/NHibernate/Type/GenericSetType.cs
```diff
@@ -76,5 +78,15 @@ public override object Instantiate(int anticipatedSize)
 		{
 			return new HashSet<T>();
 		}
+
+		protected override bool AreCollectionElementsEqual(IEnumerable original, IEnumerable target)
```
hazzik: This can be just `return ((ISet<T>) original).SetEquals((IEnumerable<T>) target);`

src/NHibernate/Type/CollectionType.cs
hazzik: >@hazzik, too much added complexity?

No. We either do it properly for all collection types or not do it. Why do we favor Sets over other collection types?

hazzik:
Please implement correct semantics for _all_ collection types. (The idea of how can it be done lays in `IPersistentCollection.EqualsSnapshot` methods

src/NHibernate/Type/GenericSetType.cs
fredericDelaporte: I have thought about it, but I have seen many entity classes with set property loosely typed `ICollection` and initialized with something else than a set (a `List<>`by example) and then added to the session: it works. Changing the code above could cause it to no more work.

src/NHibernate.Test/NHSpecificTest/NH3932/Fixture.cs
```diff
@@ -0,0 +1,169 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3932
+{
+	public class BagFixture : Fixture
```
hazzik: `[TestFixture]` is missing

src/NHibernate.Test/NHSpecificTest/NH3932/Fixture.cs
```diff
@@ -0,0 +1,169 @@
...
+	{
+		protected override IParent CreateParent()
+		{
+			return new BagParent {Children = new List<Child> {new Child { Name="child1" } }};
+		}
+	}
+
+	public class SetFixture : Fixture
```
hazzik: `[TestFixture]` is missing

src/NHibernate/Type/GenericSetType.cs
hazzik: >Changing the code above could cause it to no more work.
I doubt

RogerKratz:
TestFixture is optional in nunit (since 2.5 or so)



Sent from my Samsung Galaxy smartphone.


-------- Original message --------

src/NHibernate/Type/GenericSetType.cs
hazzik: Ok, I've checked and you are right.

hazzik: So then:
```
var set = original as ISet<T> ?? new HashSet<T>((IEnumerable<T>)original);
return set.SetEquals((IEnumerable<T>) target);
```

RogerKratz:
@hazzik @fredericDelaporte 
Now impl for bags, lists, maps and sets. Also cleaned up tests a bit. I've done what was needed now I believe - if not, please tell what's missing.

Was thinking about default impl of AreCollectionElementsEqual in CollectionType... Maybe this should be made abstract instead? Or is users suppose to be able to derive from this type (=will be a breaking change)?

src/NHibernate/Type/CollectionType.cs
```diff
@@ -359,36 +360,62 @@ public virtual object InstantiateResult(object original)
...
 
-		public virtual object ReplaceElements(object original, object target, object owner, IDictionary copyCache,
-											  ISessionImplementor session)
+		public virtual object ReplaceElements(object original, object target, object owner, IDictionary copyCache, ISessionImplementor session)
 		{
-			// TODO: does not work for EntityMode.DOM4J yet!
-			object result = target;
-			Clear(result);
+			var elemType = GetElementType(session.Factory);
+			var resultPc = target as IPersistentCollection;
```
fredericDelaporte: It would be nice to rename it `targetPc` for naming consistency, especially in the `ClearTargetsDirtyFlag` method.

src/NHibernate/Type/GenericBagType.cs
```diff
@@ -70,5 +72,12 @@ public override object Instantiate(int anticipatedSize)
 		{
 			return anticipatedSize <= 0 ? new List<T>() : new List<T>(anticipatedSize + 1);
 		}
+
+		protected override bool AreCollectionElementsEqual(IEnumerable original, IEnumerable target)
+		{
+			var first = (IEnumerable<T>)original;
+			var second = (IEnumerable<T>) target;
+			return first.All(element => first.Count(x => x.Equals(element)) == second.Count(x => x.Equals(element)));
```
fredericDelaporte: Testing general count equality is required, otherwise a bad `true` will be yielded if `target` is "wider" than `original` by a few elements not present in `original`. So it should be:

```C#
return first.Count() == second.Count() &&
    first.All(element => first.Count(x => x.Equals(element)) == second.Count(x => x.Equals(element)));
```

src/NHibernate/Type/GenericMapType.cs
```diff
@@ -116,5 +117,12 @@ public override object IndexOf(object collection, object element)
 				.Select(pair => pair.Key)
 				.FirstOrDefault();
 		}
+
+		protected override bool AreCollectionElementsEqual(IEnumerable original, IEnumerable target)
+		{
+			var first = (IDictionary<TKey, TValue>)original;
+			var second = (IDictionary<TKey, TValue>)target;
+			return first.Count == second.Count && first.All(keyValue => second.Contains(keyValue));
```
fredericDelaporte: Equality of keys is not enough, each key must map the same value in both too.

src/NHibernate/Type/CollectionType.cs
RogerKratz: Agreed. Fixed

src/NHibernate/Type/GenericMapType.cs
RogerKratz: Will look into this

src/NHibernate/Type/GenericBagType.cs
RogerKratz: Yes, will make a test + fix for this.

fredericDelaporte:
The MySql failure is due to a recently added test in #580, which seems flaky to me. I do not expect this NH-3932 PR to have anything to do about it.

src/NHibernate/Type/GenericMapType.cs
RogerKratz: Hm. In what cases do you mean that is needed? Maybe I'm too tired now and miss something obvious but... 
I mean KeyValuePair is a struct and will check its data for equality by default?

fredericDelaporte: Your are right, that is me who have miss read, sorry. I was believing you were testing only keys with `ContainsKey`.
So this is O(n²) indeed. Well, as the bag `EqualsSnapshot` implementation is, so why not.

RogerKratz:
@hazzik @fredericDelaporte 
Thanks for the input guys! I just pushed a bunch of small commits that covers what we discussed above I think.

...except that I ignored one test for idbags. First time I really digged down to the internals of that collection type so I might have looked in the wrong place but... The problem is that when the collection is cleared, so is its "_identifiers". When collection elements is added in ReplaceElements using the Add method, the _identifiers aren't put back. This will lead to an unnessecery update in one test due to the impl EqualsSnapshot that uses the "_identifiers" field.

If this is ...
* ...another bug/optimization than what is supposed to be covered in this PR/jira ticket...
* ...should be fixed by making sure _identifers somehow is put back to the state before it was cleaned...
* ...should be fixed by making sure that Add method also adds to _identifiers...
* ...should be fixed by some other way...
* ...is exptected behaviour in this specific test...
...I don't know. Don't really know the exact sematics of "idbag" I'm afraid...

Do you have any input on this?

fredericDelaporte:
Thanks for all the work,

About the idbag, I was not sure my proposal was completely right. Your tests show it is not.

For preventing this update with an idbag, the `ReplaceElements` would have to go quite beyond its current implementation for the special case "*`original` is not a `PersistentCollection` but is equal to `target`*". For that case (and only for it), it would have to: 1. cease clearing `target` ; 2. for each element in `target`, find the first one equal to it in `original`, and replace the target element. (Or to be even stricter, find the *n*th equal element, where *n* is the number of times the element has already been seen in `target` enumeration.) To replace the `target` element, `target` would have to be cast to `PersistentIdentifierBag<T>` and the element replaced with `target[i] = ...` syntax, with `i` being the count of elements already replaced.

This is valid if, as I believe, the identifiers are here only for optimizing bags manipulations (avoiding the "delete all and reinsert all" default behavior on bag element removal). And so those identifiers have no semantic meaning.  
Otherwise the "*`original` is not a `PersistentCollection`*" case should cause `ClearTargetsDirtyFlag` to return `false` without comparing the collections (so it needs to be overridden for idbag).  
This means a "clone" of an idbag will always trigger an update if merged back. A "clone" of an idbag has no means to know about the internal identifiers associated with each elements.

src/NHibernate/Type/CollectionType.cs
```diff
@@ -359,36 +360,54 @@ public virtual object InstantiateResult(object original)
...
-			if(originalPc != null && resultPc!=null)
+			if(clearTargetsDirtyFlag)
+			{
+				targetPc.ClearDirty();
+			}
+
+			return target;
+		}
+
+		protected bool ClearTargetsDirtyFlag(object target, IPersistentCollection targetPc, IPersistentCollection originalPc, IEnumerable iter)
```
fredericDelaporte: I understand the need for this method, but I dislike its name and semantic. It sounds too internal for being exposed to users implementing their own collection type (normally by deriving from `CustomCollectionType`, which derives from `CollectionType`).

I think it should be `internal` instead of `protected`.

RogerKratz: Agreed. Renamed it and changed to internal.

RogerKratz:
I just pushed a fix for the ignored test.

@fredericDelaporte Don't know if I understood your suggestion (thanks!), but I simply added a new flag in GenericIdentifierBagType.ReplaceElements saying if target should be cleared and then elements added or elements instead should be replaced.
A lot if if:s/branches in that method but... All green now at least ;).

src/NHibernate/Type/GenericIdentifierBagType.cs
```diff
@@ -82,5 +84,71 @@ protected override void Add(object collection, object element)
...
+			var originalPc = original as IPersistentCollection;
+			var iter = (IList<T>)original;
+			var clearTargetsDirtyFlag = false;
+			var clearTarget = true;
+
+			if (targetPc != null)
+			{
+				if (originalPc == null)
+				{
+					if (!targetPc.IsDirty && AreCollectionElementsEqual(iter, (IEnumerable)target))
```
fredericDelaporte: Use `targetPc` without cast here instead of `target`.

src/NHibernate/Type/GenericIdentifierBagType.cs
```diff
@@ -82,5 +84,71 @@ protected override void Add(object collection, object element)
...
+				foreach (var obj in iter)
+				{
+					Add(target, elemType.Replace(obj, null, session, owner, copyCache));
+				}
+			}
+			else
+			{
+				for (var i = 0; i < targetPc.Count; i++)
+				{
+					targetPc[i] = (T)elemType.Replace(iter[i], null, session, owner, copyCache);
```
fredericDelaporte: `iter` (to be renamed `originalIter` maybe) can be equal without having the same ordering of elements (order is not significant in bags). `iter[i]` can be another element than `targetPc[i]`. We need to search in `iter` the element corresponding to `targetPc[i]`. Otherwise `targetPc` may swap its internal identifiers between elements, which can result in deleting the wrong element in db if we remove one element from it later on.

RogerKratz: Well spotted - thanks! There was a testcase missing. Just added it - push fix soon.

RogerKratz:
Pushed a fix for idbags and some code clean up as suggested.

src/NHibernate/Type/CollectionType.cs
```diff
@@ -359,36 +360,54 @@ public virtual object InstantiateResult(object original)
...
-											  ISessionImplementor session)
+		public virtual object ReplaceElements(object original, object target, object owner, IDictionary copyCache, ISessionImplementor session)
 		{
-			// TODO: does not work for EntityMode.DOM4J yet!
-			object result = target;
-			Clear(result);
+			var elemType = GetElementType(session.Factory);
+			var targetPc = target as IPersistentCollection;
+			var originalPc = original as IPersistentCollection;
+			var iter = (IEnumerable)original;
```
fredericDelaporte: Renaming this one one too would be better.

src/NHibernate/Type/GenericMapType.cs
```diff
@@ -80,23 +80,24 @@ public override object ReplaceElements(object original, object target, object ow
 		{
 			ICollectionPersister cp = session.Factory.GetCollectionPersister(Role);
 
-			IDictionary<TKey, TValue> result = (IDictionary<TKey, TValue>)target;
+			var targetPc = target as IPersistentCollection;
+			var originalPc = original as IPersistentCollection;
+			var iter = (IDictionary<TKey, TValue>)original;
```
fredericDelaporte: Renaming this one one too would be better.

src/NHibernate.Test/NHSpecificTest/NH3932/Model.cs
```diff
@@ -0,0 +1,268 @@
...
+{
+	public interface IParent
+	{
+		IParent Clone();
+		void ReverseChildren();
+		void ClearChildren();
+		void RemoveLastChild();
+	}
+
+	public class MapParent : IParent
```
hazzik: True, version shall not be included into the Equals

src/NHibernate.Test/NHSpecificTest/NH3932/Fixture.cs
```diff
@@ -0,0 +1,205 @@
...
+			var parentClone = storedParent.Clone();
+			using (var s = OpenSession())
+			{
+				s.SessionFactory.Statistics.Clear();
+				using (var tx = s.BeginTransaction())
+				{
+					s.Merge(parentClone);
+					tx.Commit();
+				}
+				Assert.That(s.IsDirty(), Is.False);
```
RogerKratz: Yes, my fault! Was added as an "extra" test next to the one counting entityupdatecount.... So this one was probably never red... :(.

Fixed.

src/NHibernate.Test/NHSpecificTest/NH3932/Model.cs
RogerKratz: fixed

RogerKratz:
Fixed the issues in test + renaming as suggested.

src/NHibernate.Test/NHSpecificTest/NH3932/Model.cs
```diff
@@ -252,7 +275,7 @@ public override bool Equals(object obj)
 
 		public override int GetHashCode()
 		{
-			return 1; //for simlicity - care about equals impl
+			return Id;
```
fredericDelaporte: With this change, we now have following expression on the parent set, after the `ISession.Save`, yielding `false`:
```C#
((SetParent)storedParent).Children.Contains(System.Linq.Enumerable.First(((SetParent)storedParent).Children))
```
Which is wrong, of course. This happens because the child has been added in the set with `0` as hashcode but now yields `1` as hashcode, causing the set to no more knows it already contains it.

It does not trigger a bug in current test cases because the way `GenericSetType.AreCollectionElementsEqual` is currently coded, the test is done on the cloned set, which has stored the children with its latest hashcode. Inverting on which set the `SetEquals` call is done triggers the bug.

The usual way to avoid this trouble is to compute the hashcode only once per instance, storing it and then always yielding the first computed value. (A `Lazy<int>` behavior.) This does not fix all potential troubles since querying the child from a new session would result in obtaining a clone child, being equal but with a different hashcode. In the case of your test, this causes the merging of the "clean clone" (which only clone the parent, not the children) to fail due to this.

The safer in the case of your tests would be to reload the to be stored parent entity from a new session before storing it in the fixture. Or use the `assigned` id generator and assign child id before adding it to a set.

RogerKratz: First, thanks for your careful and quick reviews!

About the gethashcode impl in test model... IMHO, I think the test model should meet the requirements of the tests. They don't have to work for scenarios not (yet) present in the test.
(On the other hand they should not make tests false positive and they should also, to some extend, show "good" use case examples - which is more what you, correctly, pointing out).

I change it to simplest impl to always give a correct result (=hard coded value). Good enough?

RogerKratz:
Pushed a "fix"/potential false positives for gethashcode impl.

src/NHibernate.Test/NHSpecificTest/NH3932/Model.cs
fredericDelaporte: Good enough for me.

hazzik:
:tada:

hazzik:
Squashed.

hazzik:
Merged. Thank you very much for the hard work and patience.

PR #546, opened 2017-01-09
NH-3934 Add methods to QueryOver WhereNot and AndNot with ICriterion …

KrzysztofLapinski:
QueryOver has methods Where(ICriterion) and And(ICriterion) but hasnt WhereNot(ICriterion) and AndNot(ICriterion). It's usefull for example when work with subqueries


hazzik:
Thanks!

PR #547, opened 2017-01-09
Regression caused by NH-3904

gliljas:
Just a suggestion. Please advise.


hazzik:
Some options here:

1) fast & easy solution: rollback this heuristic to get custom type and force user to use `.MappedAs()` extension method (NH-2401/#421, also in 4.1) . 
2) properly implement `GuessType` (see `AbstractQueryImpl.GuessType` to get the idea). But this solution will still have some flaws as sometimes the wrong type will be picked anyway.
3) same as above, but also analyse surrounding expression to get the idea about constant expression true type (similar to #480)

My preferred option is 1. 
/cc @oskarb?

gliljas:
Option 1 makes sense, since then there will only be an explicit way to use user type constants, and not also a brittle, implicit way.

Option 2 was what I wanted to do, but for an urgent fix, it seemed lika bad idea.

Option 3 is the best way, IMO, but also potentially tricky.

hazzik:
Found that `MappedAs` actually does not work...

PR #554, opened 2017-02-19
NH-3948 removed duplicated CheckAndUpdateSessionStatus()

gioski:
call in CreateFilter method inside SessionImpl class.

https://nhibernate.jira.com/browse/NH-3948


hazzik:
thanks

PR #560, opened 2017-02-28
NH-3953 - Added take clause to ProcessSingle, similar to ProcessFirst

nkreipke:
Probably optimizes the records returned by the database.

### DIFF
--- a/src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessSingle.cs
+++ b/src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessSingle.cs
@@ -12,6 +12,8 @@ namespace NHibernate.Linq.Visitors.ResultOperatorProcessors
                                   : ReflectionHelper.GetMethodDefinition(() => Queryable.Single<object>(null));
 
             AddClientSideEval(firstMethod, queryModelVisitor, tree);
+
+            tree.AddTakeClause(tree.TreeBuilder.Constant(2));
         }
     }
 }


fredericDelaporte:
Well, adding `top 2` may optimize a failure case for `Single`: more than one element, and for this to optimize, even more than two.
The normal, not failing case will not be optimized by this change: only one element, or none in case of `OrDefault`.
Still an optimization, but for an edge case I think, contrary to the almost same code for `First`.

Anyway, I think you should probably add an improvement issue in [NHibernate Jira](https://nhibernate.jira.com) too for a better change tracking. (I am not from NHibernate team though, this is just my opinion.)

hazzik:
Hi @nkreipke. Please add a JIRA issue.

nkreipke:
Added a [JIRA issue NH-3953](https://nhibernate.jira.com/browse/NH-3953).

fredericDelaporte:
I now think this may be a breaking change for fetching cases with `Single`. In case a `FetchMany` is done, it will, with this PR, only be able of loading the two first related entities at most.

Yes, `FetchMany` is already broken with `First`, but is it a reason for having it broken with `Single` too?

As said in a previous comment, this change is only optimizing some failure cases anyway. And now it looks to me it creates a bug for passing cases.

(While the `take 1` for `First` optimize passing cases too, and removing it for supporting `FetchMany` may cause passing cases to have unacceptable performances.)

Maybe I am wrong: I have not written test cases for checking that `FetchMany` currently works as intended with three or more children to load and `Single`, and do no more works with this PR. Such a test case should probably be added to this PR.

hazzik:
good points @fredericDelaporte 

Actually, I think, that we need to disallow (eg throw an exception) users to combine eager fetching with paging (there are several JIRAs have been closed as "Not an issue" for criteria api, and some still open for Linq).

fredericDelaporte:
Disallowing `FetchMany` with `Skip` or `Take` will make sens. But `Fetch` should probably remain allowed, when not used on collections (it appears it works on collection too, apart it does not allow proper use of `ThenFetch` or `ThenFetchMany` if `Fetch` is used on a collection).

Supporting `FetchMany` with `Skip` or `Take` does not look impossible to me, but this would require supporting sub-queries in `from` statement. Probably not an easy task.

PR #561, opened 2017-03-03
NH-3954 - Dynamic proxy cache may yield a wrong proxy

fredericDelaporte:
Fixes [NH-3954](https://nhibernate.jira.com/browse/NH-3954) - Dynamic proxy cache may yield a wrong proxy.

It relies on a dictionary cache, using a key which was implementing equals as hashcode equality.  
But hascode are not unique, and two different objects, even of the same class with different properties/members values, may have the same hashcode. This is documented on [msdn](https://msdn.microsoft.com/en-us/library/zdee4b3y(v=vs.110).aspx).

Quite unlikely, but it happened to me this night. While testing some other changes, `StatelessSessionFetchingTest.DynamicFetch` was unexpectedly failing. And this was because some other proxy got cached with the same hashcode as the proxy required by this test.

### DIFF
--- a/src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
+++ b/src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
@@ -11,52 +11,57 @@ using System.Collections.Generic;
 
 namespace NHibernate.Proxy.DynamicProxy
 {
-	[Serializable]
-	public class ProxyCacheEntry
+	public class ProxyCacheEntry : IEquatable<ProxyCacheEntry>
 	{
-		private readonly int hashCode;
+		private readonly int _hashCode;
 
 		public ProxyCacheEntry(System.Type baseType, System.Type[] interfaces)
 		{
 			if (baseType == null)
-			{
-				throw new ArgumentNullException("baseType");
-			}
+				throw new ArgumentNullException(nameof(baseType));
 			BaseType = baseType;
-			Interfaces = interfaces ?? new System.Type[0];
+			_uniqueInterfaces = new HashSet<System.Type>(interfaces ?? new System.Type[0]);
 
-			if (Interfaces.Length == 0)
+			if (_uniqueInterfaces.Count == 0)
 			{
-				hashCode = baseType.GetHashCode();
+				_hashCode = baseType.GetHashCode();
 				return;
 			}
-
-			// duplicated type exclusion
-			var set = new HashSet<System.Type>(Interfaces) { baseType };
-			hashCode = 59;
-			foreach (System.Type type in set)
+			
+			var allTypes = new List<System.Type>(_uniqueInterfaces) { baseType };
+			_hashCode = 59;
+			foreach (System.Type type in allTypes)
 			{
-				hashCode ^= type.GetHashCode();
+				// This simple implementation is nonsensitive to list order. If changing it for a sensitive one,
+				// take care of ordering the list.
+				_hashCode ^= type.GetHashCode();
 			}
 		}
 
-		public System.Type BaseType { get; private set; }
-		public System.Type[] Interfaces { get; private set; }
+		public System.Type BaseType { get; }
+		public IReadOnlyCollection<System.Type> Interfaces { get { return _uniqueInterfaces; } }
+		
+		private HashSet<System.Type> _uniqueInterfaces;
 
 		public override bool Equals(object obj)
 		{
 			var that = obj as ProxyCacheEntry;
-			if (ReferenceEquals(null, that))
+			return Equals(that);
+		}
+
+		public bool Equals(ProxyCacheEntry other)
+		{
+			if (ReferenceEquals(null, other) ||
+				// hashcode inequality allows an early exit, but their equality is not enough for guaranteeing instances equality.
+				_hashCode != other._hashCode ||
+				BaseType != other.BaseType)
 			{
 				return false;
 			}
 
-			return hashCode == that.GetHashCode();
+			return _uniqueInterfaces.SetEquals(other.Interfaces);
 		}
 
-		public override int GetHashCode()
-		{
-			return hashCode;
-		}
+		public override int GetHashCode() => _hashCode;
 	}
 }
--- a/src/NHibernate.Test/NHSpecificTest/NH3954/Entity.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3954/Entity.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace NHibernate.Test.NHSpecificTest.NH3954
+{
+	class Entity1
+	{
+	}
+	class Entity1FakeProxy
+	{
+	}
+	class Entity1FakeProxy2
+	{
+	}
+
+	class Entity2
+	{
+	}
+	class Entity2FakeProxy
+	{
+	}
+	class Entity2FakeProxy2
+	{
+	}
+
+	class Entity3
+	{
+	}
+	class Entity3FakeProxy
+	{
+	}
+	class Entity3FakeProxy2
+	{
+	}
+
+	class Entity4
+	{
+	}
+	class Entity4FakeProxy
+	{
+	}
+	class Entity4FakeProxy2
+	{
+	}
+
+	class Entity5
+	{
+	}
+	class Entity5FakeProxy
+	{
+	}
+	class Entity5FakeProxy2
+	{
+	}
+}
--- a/src/NHibernate.Test/NHSpecificTest/NH3954/EqualsFixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3954/EqualsFixture.cs
@@ -1,0 +1,79 @@
+﻿using System;
+using System.Reflection;
+using NHibernate.Proxy;
+using NHibernate.Proxy.DynamicProxy;
+using NHibernate.Type;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3954
+{
+	[TestFixture]
+	public class EqualsFixture
+	{
+		private static readonly FieldInfo HashCodeField =
+			typeof(ProxyCacheEntry).GetField("_hashCode", BindingFlags.Instance | BindingFlags.NonPublic);
+
+		/// <summary>
+		/// Allows to simulate a hashcode collision. Issue would be unpractical to test otherwise.
+		/// Hashcode collision must be supported for avoiding unexpected and hard to reproduce failures.
+		/// </summary>
+		private void TweakEntry(ProxyCacheEntry entryToTweak, int hashcode)
+		{
+			// Though hashCode is a readonly field, this works at the time of this writing. If it starts breaking and cannot be fixed,
+			// ignore those tests or throw them away.
+			HashCodeField.SetValue(entryToTweak, hashcode);
+		}
+
+		// Non reg test case
+		[Test]
+		public void TypeEquality()
+		{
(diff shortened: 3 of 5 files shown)


src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
```diff
@@ -43,15 +45,35 @@ public ProxyCacheEntry(System.Type baseType, System.Type[] interfaces)
...
 			{
 				return false;
 			}
 
-			return hashCode == that.GetHashCode();
+			// In case the instance was de-serialized, the _uniqueInterfaces will be null. Re-fetch its _uniqueInterfaces if needed.
+			if (_uniqueInterfaces == null)
+				_uniqueInterfaces = new HashSet<System.Type>(Interfaces);
+			if (that._uniqueInterfaces == null)
+				that._uniqueInterfaces = new HashSet<System.Type>(Interfaces);
```
hazzik: this shall be 

```
that._uniqueInterfaces = new HashSet<System.Type>(that.Interfaces);
```
(note `that.` in front of `Interfaces`)

fredericDelaporte: Damn, I should not code that late...

fredericDelaporte: Fixed, squashed, ... I think this `Serializable` attribute is indeed unneeded for current usage of that class. Removing it would allow a simpler implementation. Maybe I have been too conservative here.

fredericDelaporte:
I forget to mention in my PR message: on [Jira side in a comment](https://nhibernate.jira.com/browse/NH-3954?focusedCommentId=49200&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-49200), I have listed a bunch of other classes having the same defect in NHibernate. This PR does not fix them.

fredericDelaporte:
The TeamCity failure occured for MySql build, new test failure in `ExpressionSessionLeakTest.SessionGetsCollected`.

It appears that it has not fail on the fisrt build for this PR, executed on code before the fix I have done in response to Hazzik's review.

Due to the last change being very small, I doubt this failure has something to do with the change. Is it a known test case for unexpected failures? 
(Furthermore, debugging the test and setting breakpoint in `ProxyCache`, it appears this test  does not use this feature.)

fredericDelaporte:
I have made changes to tests: added direct tests on `Equals`, set cache tests as only run explicitly since a bit unpredictable, and simplified test implementation. (Tests locally re-tested both with original code and fixed code.)

fredericDelaporte:
Updated: was re-implementing set equality, used .Net built-in method instead. And re-based.

fredericDelaporte:
Rebased for resolving a conflict on test project.

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
```diff
@@ -43,15 +44,26 @@ public ProxyCacheEntry(System.Type baseType, System.Type[] interfaces)
...
 			}
 
-			return hashCode == that.GetHashCode();
+			// In case the instance was de-serialized, the _uniqueInterfaces will be null. Re-fetch its _uniqueInterfaces if needed.
+			if (_uniqueInterfaces == null)
+				_uniqueInterfaces = new HashSet<System.Type>(Interfaces);
+			if (that._uniqueInterfaces == null)
+				that._uniqueInterfaces = new HashSet<System.Type>(that.Interfaces);
+
+			return _uniqueInterfaces.SetEquals(that._uniqueInterfaces);
```
hazzik: `SetEquals` does not require the argument to be `ISet<>`, so the second if (`if (that._uniqueInterfaces == null)`) is not required

hazzik:
There is a possibility that two instances of ProxyCacheEntry will be semantically equal (have the same base type and interfaces) but falsely declared unequal due the fact that set is not ordered.

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
```diff
@@ -43,15 +44,26 @@ public ProxyCacheEntry(System.Type baseType, System.Type[] interfaces)
 		public System.Type BaseType { get; private set; }
 		public System.Type[] Interfaces { get; private set; }
```
hazzik: As we are allowed to break everything in NH5, Please make this `IReadOnlyCollection<System.Type>` without getter.

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
```diff
@@ -43,15 +44,26 @@ public ProxyCacheEntry(System.Type baseType, System.Type[] interfaces)
 		public System.Type BaseType { get; private set; }
```
hazzik: Please remove getter as per below.

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
```diff
@@ -24,6 +24,7 @@ public ProxyCacheEntry(System.Type baseType, System.Type[] interfaces)
 			}
 			BaseType = baseType;
 			Interfaces = interfaces ?? new System.Type[0];
+			_uniqueInterfaces = new HashSet<System.Type>(Interfaces);
```
hazzik: I think, what we need is presort `Interfaces` (eg. `interfaces.Distinct().OrderBy(x=>x.FullName).ToArray()`).

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
```diff
@@ -43,15 +44,26 @@ public ProxyCacheEntry(System.Type baseType, System.Type[] interfaces)
...
 
+		[NonSerialized]
+		private HashSet<System.Type> _uniqueInterfaces;
+
 		public override bool Equals(object obj)
 		{
 			var that = obj as ProxyCacheEntry;
-			if (ReferenceEquals(null, that))
+			if (ReferenceEquals(null, that) ||
+				hashCode != that.hashCode ||
```
hazzik: I think `hashCode` check is redundant here, but it's ok to leave it as is.

fredericDelaporte: I left it as an easy early "exit case". It will most of time speed up non equal cases, while being a very minor overhead for equal cases. I should add a comment for warning the hashcode is only good for early false yielding, not for telling it is equal.

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
fredericDelaporte: [`SetEquals`](https://msdn.microsoft.com/en-us/library/dd412096(v=vs.110).aspx) is not sensible to order.

> This method ignores the order of elements and any duplicate elements in other.

So I do not expect any order mismatch to cause an issue.

hazzik: Order mismatch can (?) happen in the calculation of `hashCode`

hazzik: Ok,  no it can not, as hashCode is calculated using just xor (which is strange as there will be lots of collisions)

fredericDelaporte:
> Could you please check the deserialization case?

About this serialization, I was wondering:

> I think this Serializable attribute is indeed unneeded for current usage of that class. Removing it would allow a simpler implementation. Maybe I have been too conservative here.

Is there a known reason for having this class serializable? It may be something inherited from the LinFu library but maybe useless for NHibernate.

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
fredericDelaporte: But there is the hashcode computation too. I am wrong thinking xor is permutable? (Will yield same result whatever the evaluation order.) I am no mure sure about that. But I can not find an exemple for which it would be false.

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
hazzik: Fixed. I was talking about setter, of cource

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
fredericDelaporte: There is a second HashSet used on the one I have introduced, which is now useless and could be replaced by a simple List. I will change that too.

fredericDelaporte: > Ok, no it can not, as hashCode is calculated using just xor (which is strange as there will be lots of collisions)

When used to combine hashcode from a set (so different objects), it should not be that bad. It is not like `&` or `|` which ends up zeroing or "one-ing" all bits. It seems it is even a [common practice in Java](http://stackoverflow.com/a/19139500/1178314).

Now if we want to change it for a more classical approach like multiplying by a prime then adding next value while overflowing, we will need to take care of the order.

fredericDelaporte:
A more convoluted test case is added, with many interfaces, different order, duplicates, resolving to same set.

Code changed as suggested (with a bit more c#6/7 syntax, and more comments).

Additionally, I have tested the 5 thousands tests without this `[Serializable]` on `ProxyEntryCache`, no failure. (I have not committed this removal.)

fredericDelaporte:
Hum, team city build failed. Maybe using C#7 features is too early. I removed the ["throw expression"](https://stackoverflow.com/documentation/c%23/1936/c-sharp-7-0-features/20215/throw-expressions#t=201703200103502320517).

hazzik:
Yep, no C#7 until #569

fredericDelaporte:
Rebased (& commit reworded).

fredericDelaporte:
Rebased and done.

src/NHibernate/Proxy/DynamicProxy/ProxyCacheEntry.cs
```diff
@@ -11,52 +11,57 @@
...
-			{
-				throw new ArgumentNullException("baseType");
-			}
+				throw new ArgumentNullException(nameof(baseType));
 			BaseType = baseType;
-			Interfaces = interfaces ?? new System.Type[0];
+			_uniqueInterfaces = new HashSet<System.Type>(interfaces ?? new System.Type[0]);
 
-			if (Interfaces.Length == 0)
+			if (Interfaces.Count == 0)
```
hazzik: Please change this to `_uniqueInterfaces.Count == 0`

fredericDelaporte: Done.

fredericDelaporte:
Well, I just discover following test: `Legacy.MasterDetailTests.Serialize`. It does serialize a whole `ISession`, deserialize it and goes on using it... So if we want to maintain this possibility, cache entries should remain serializable. But I would still keep a simplified implementation of it, the hashtable can be serialized. I would just add back that [Serializable] attribute.

hazzik:
Does it fail?

fredericDelaporte:
No, but very likely because it does not try to serialize a session having cached some dynamic proxy.

fredericDelaporte:
I will try to cause this serialization test to have a cached proxy.

hazzik:
The ProxyFactory and its dependencies do not get serialized. It's all fine here.

fredericDelaporte:
Ok, great, thanks.

PR #562, opened 2017-03-03
NH-3956 - native SQL query plan cache fix.

fredericDelaporte:
Fix and test cases for [NH-3956](https://nhibernate.jira.com/browse/NH-3956): Native SQL query plan may get wrong plan.

Equals was relying on hashcode equality.

Code taken back prior to that "bad optimization", and compared to hibernate [code base](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/engine/query/spi/sql/NativeSQLQuerySpecification.java) (main difference: NHibernate shortcuts on hashcode inequality, not Hibernate).

I have spotted some other discrepancies by the way, but left them untouched: `Equals` implementations for `INativeSQLQueryReturn` classes do only test their `alias` properties on NHibernate side, while Hibernate test all properties.

Test cases only test `Equality`, they do not showcase query plan cache failures which may result from those `Equality` failures, contrary to those of NH-3954 for the proxy cache. But there is no reason the query plan cache would not have been affected by those `Equals` failures.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3956/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3956/Fixture.cs
@@ -1,0 +1,221 @@
+﻿using System.Reflection;
+using NHibernate.Engine.Query.Sql;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3956
+{
+	[TestFixture]
+	public class Fixture
+	{
+		private static readonly FieldInfo HashCodeField =
+			typeof(NativeSQLQuerySpecification).GetField("hashCode", BindingFlags.Instance | BindingFlags.NonPublic);
+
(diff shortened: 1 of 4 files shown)


fredericDelaporte:
I have re-based only for amending the commit message. It looks like a general failure on Team City side was affecting most tests: the builds triggered by this rebase now looks to be passing. (Already 3 good, while previous builds had them all failed but one.)

fredericDelaporte:
MySql failure is a timeout. To me, it looks like a test infrastructure frequent failure rather than a regression. The same trouble with the same test happened for [PR](https://github.com/nhibernate/nhibernate-core/pull/564), though it does not changes the same things at all.

hazzik:
@fredericDelaporte can you please rebase on top of master?

PR #564, opened 2017-03-04
NH-3957 - fix 2nd level query cache bug and test cases

fredericDelaporte:
Test cases and fix for [NH-3957](https://nhibernate.jira.com/browse/NH-3957): `Equals` overrides vulnerable to hashcode collisions have been added in a bunch of result transformers classes.  
This causes the second level query cache to be unreliable: we may get a wrong cached query result from it. Granted, due to the way those hashcode are calculated, is it very unlikely, but still possible. This cause a dire situation where an application thoroughly tested may starts failing till restarted, without much possibilities of reproducing the issue afterward.

Test cases only test `Equality`, they do not showcase second level query cache failures which may result from those `Equality` failures, contrary to those of NH-3954 for the proxy cache. But there is no reason the second level query cache would not have been affected by those `Equals` failures.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3957/ResultTransformerEqualityFixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3957/ResultTransformerEqualityFixture.cs
@@ -1,0 +1,136 @@
+﻿using System.Collections.Generic;
+using System.Reflection;
+using NHibernate.Transform;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3957
+{
+	[TestFixture]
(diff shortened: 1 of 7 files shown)


fredericDelaporte:
MySql failure is a timeout. To me, it looks like a test infrastructure frequent failure rather than a regression. The same trouble with the same test happened for [PR](https://github.com/nhibernate/nhibernate-core/pull/562), though it does not changes the same things at all.

hazzik:
@fredericDelaporte the JIRA task wasn't closed. The automatic transitions happen only to Bugs.

fredericDelaporte:
Ok, I was not knowing it. I had already seen it but was just believing there was sometime some delay...

PR #567, opened 2017-03-07
NH-2444: Adding some documentation for Linq querying API.

fredericDelaporte:
[NH-2444](https://nhibernate.jira.com/browse/NH-2444) ask for documenting the Linq provider. Here is a minimal documentation for it.  
There is no "unsupported feature" section as asked in a comment: this may change too frequently in my opinion. (And I do not have an exhaustive knowledge of it.)  
There is no "extensions" section too: I feel it a bit hard to properly document.

Any feedback welcome!

### DIFF
--- a/doc/reference/master.xml
+++ b/doc/reference/master.xml
@@ -17,6 +17,7 @@
 <!ENTITY query-hql              SYSTEM "modules/query_hql.xml">
 <!ENTITY query-criteria         SYSTEM "modules/query_criteria.xml">
 <!ENTITY query-queryover        SYSTEM "modules/query_queryover.xml">
+<!ENTITY query-linq             SYSTEM "modules/query_linq.xml">
 <!ENTITY query-sql              SYSTEM "modules/query_sql.xml">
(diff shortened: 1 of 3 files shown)


fredericDelaporte:
I think I should add a "supported data types methods and properties" section for documenting things like support of `DateTime.Year`, `string.Length`, and whatever else I find in the `DefaultLinqToHqlGeneratorsRegistry`.  
I may move there a part of the "miscellaneous" section, which may then disappear (I do not like such section).
And by the way I could give some starting point for adding custom generators, but without much more explanations than "go see the sources for examples", especially about expression conversion to HQL trees.

fredericDelaporte:
"Supported methods and members" section committed. "Miscellaneous" is shorter but still there, and I have not added explanations for adding custom generator for now.

fredericDelaporte:
Minimal explanations for a custom generator added. Heavily example based, and far from covering all cases, it recommends checking NHibernate sources for more. (Should I include a direct link on Github?)
There is no "misc" section anymore. 

I do not think I will made more changes, unless I have to take into account some feedback.

PR #568, opened 2017-03-16
NH-3944 - Upgrade Remotion.Linq to v2.1

hazzik:
* [x] Fix package build (release).


fredericDelaporte:
Great job, looks like it was a bit a heavy task. Probably better to merge it soon for allowing pending Linq related PR to get re-based and adapted to those incoming changes.

fredericDelaporte:
Is there any clean up you have thought of I may do?

hazzik:
I actually do not like all these new `INh...` stuff (`INhQueryModelVisitor`, `NhHavingClause`, etc). Now it looks like we are using relinq improperly.

fredericDelaporte:
I have tried to understand what is going on.

Currently, I think Relinq provide a Linq model meant to represent the Linq query, and a visitor logic to transform this model to the target (HQL tree for NHibernate).

The model supports being modified during visits. But NHibernate not only modifies it but also extends its semantic with clauses unknown from Linq (`having`, `with`, the `join` accepting `with` instead of `where`).

This does not look expected by Relinq designers. And doing that is harder with v2 since it has sealed a bunch of classes NHibernate was deriving from for adding and handling those additional clauses.

If we want to cease tinkering the model with additional non Linq semantic, I think we need to remove rewriters like `MergeAggregatingResultsRewriter` and handle what they were doing during the final model visit.

Maybe it is not even feasible. But it looks to me we would have easier trying to do that with current master implementation, where those additional semantics were not needing a lot of supporting code, rather than with this current migration to Relinq v2.

(And maybe there are some other issues I have not spotted currently.)

hazzik:
Agreed

hazzik:
@fredericDelaporte I'm going to update this PR with Expression Trees v2 Spec. At this stage, I would prefer to leave the code as close to the original as possible.

fredericDelaporte:
Ok, closing #594 and #599. I will add a Jira for `DbOnly`. Making it a purpose of `LinqExtensionMethod` requires to have an additional case for specifically handling `MappedAs`, hence my choice to have it separated, but why not.

fredericDelaporte:
`DbOnly` is usable on other user extensions not being a simple sql function call too. Providing it only through `LinqExtensionMethod` would be more limiting than having it as standalone.

We may keep it while adding an option on `LinqExtensionMethod` for enabling it to control that itself.

src/NHibernate/Linq/Visitors/SelectClauseNominator.cs
```diff
@@ -121,11 +121,11 @@ private bool IsRegisteredFunction(Expression expression)
...
-			else if (expression.NodeType == (ExpressionType)NhExpressionType.Sum ||
-						expression.NodeType == (ExpressionType)NhExpressionType.Count ||
-						expression.NodeType == (ExpressionType)NhExpressionType.Average ||
-						expression.NodeType == (ExpressionType)NhExpressionType.Max ||
-						expression.NodeType == (ExpressionType)NhExpressionType.Min)
+			else if (expression is NhSumExpression ||
+			         expression is NhCountExpression ||
+			         expression is NhAverageExpression ||
+			         expression is NhMaxExpression ||
+			         expression is NhMinExpression)
```
hazzik: Cast should be as cheap as the equal. 

The alignment were discussed somewhere with @oskarb. The consensus was that we use tabs for indents and spaces (after tabs) to align. This way it looks the same everywhere despite the size of a tab.

src/NHibernate/Linq/Clauses/NhHavingClause.cs
```diff
@@ -1,19 +1,91 @@
...
+		/// <param name="queryModel">The query model in whose context this clause is visited.</param>
+		/// <param name="index">
+		///     The index of this clause in the <paramref name="queryModel" />'s
+		///     <see cref="P:Remotion.Linq.QueryModel.BodyClauses" /> collection.
+		/// </param>
+		public void Accept(IQueryModelVisitor visitor, QueryModel queryModel, int index)
 		{
+			if (visitor == null) throw new ArgumentNullException(nameof(visitor));
+			if (queryModel == null) throw new ArgumentNullException(nameof(queryModel));
+			((INhQueryModelVisitor) visitor).VisitNhHavingClause(this, queryModel, index);
```
hazzik: Check this out: https://www.re-motion.org/jira/browse/RMLNQ-86. In short they added IBodyClause and IFromClause because of us.

src/NHibernate/Linq/EagerFetchingExtensionMethods.cs
```diff
@@ -15,7 +15,7 @@ public static class EagerFetchingExtensionMethods
             if (query == null) throw new ArgumentNullException(nameof(query));
             if (relatedObjectSelector == null) throw new ArgumentNullException(nameof(relatedObjectSelector));
 
-            var methodInfo = ((MethodInfo)MethodBase.GetCurrentMethod()).MakeGenericMethod(typeof(TOriginating), typeof(TRelated));
+	        var methodInfo = ((MethodInfo)MethodBase.GetCurrentMethod()).MakeGenericMethod(typeof(TOriginating), typeof(TRelated));
```
hazzik: Probably rebase glitch.

src/NHibernate/Linq/Clauses/NhJoinClause.cs
```diff
@@ -3,60 +3,171 @@
...
+		///     The index of this clause in the <paramref name="queryModel" />'s
+		///     <see cref="P:Remotion.Linq.QueryModel.BodyClauses" /> collection.
+		/// </param>
+		public void Accept(IQueryModelVisitor visitor, QueryModel queryModel, int index)
+		{
+			if (visitor == null) throw new ArgumentNullException(nameof(visitor));
+			if (queryModel == null) throw new ArgumentNullException(nameof(queryModel));
+			var nhVisitor = visitor as INhQueryModelVisitor;
+			if (nhVisitor == null)
+				throw new ArgumentException("Expect visitor to implement INhQueryModelVisitor", nameof(visitor));
```
hazzik: I'll align all these usages with the same approach

src/NHibernate/Linq/Visitors/SelectClauseNominator.cs
fredericDelaporte: The default VS tooling does not support this well. As soon as we have written some code (or specifically, pasted some java code) and want to auto-fix the formatting, it switches those spaces alignements to tabs in the whole file.

Maybe I have mis-configured the tooling, but then I am unable to find the right option for preserving such formatting. (And I have tried with R# too.)

hazzik:
Cleaned up, rebased, and squashed.

ngbrown:
Anything holding this back from being merged?

hazzik:
I would like to wait until the release of "fetching"

ngbrown:
@hazzik [Remotion.Linq.EagerFetching v2.0.1](https://www.nuget.org/packages/Remotion.Linq.EagerFetching/2.0.1) should be fine, because there's not actually any code changes to v2.1/2.2 just compiling for .NETStandard.

hazzik:
I don't like binding redirects

fredericDelaporte:
Even with the [beta 2.1](https://www.nuget.org/packages/Remotion.Linq.EagerFetching/2.1.0-beta-001) those binding redirects stays required, although it directly depends on the ReLinq 2.1 we use.

hazzik:
>Even with the beta 2.1 those binding redirects stays required,  why is that?

fredericDelaporte:
I do not know. I have just upgraded to the beta lib, seen the bindings were still there, removed them, try testing, failure trying to load Relinq 2.0 instead of 2.1. I have not taken yet the time to enable binding failure trace and check what was going on.

hazzik:
Ok, there is a broken reference in the beta.

hazzik:
https://www.re-motion.org/jira/browse/RMLNQFETCH-21

hazzik:
Upgraded to re-linq 2.1.2, re-linq fetch 2.1.0. Rebased & squashed. 

@fredericDelaporte please review.

src/NHibernate/Linq/Clauses/NhJoinClause.cs
```diff
@@ -3,60 +3,166 @@
...
+		/// </summary>
+		/// <remarks>
+		///     Item names are inferred when a query expression is parsed, and they usually correspond to the variable names
+		///     present in that expression.
+		///     However, note that names are not necessarily unique within a <see cref="T:Remotion.Linq.QueryModel" />. Use names
+		///     only for readability and debugging, not for
+		///     uniquely identifying <see cref="T:Remotion.Linq.Clauses.IQuerySource" /> objects. To match an
+		///     <see cref="T:Remotion.Linq.Clauses.IQuerySource" /> with its references, use the
+		///     <see cref="P:Remotion.Linq.Clauses.Expressions.QuerySourceReferenceExpression.ReferencedQuerySource" /> property
+		///     rather than the <see cref="P:Remotion.Linq.Clauses.NhFromClauseBase.ItemName" />.
```
fredericDelaporte: Invalid `cref`. But not even sure the whole remarks section is useful here.

src/NHibernate/Linq/Clauses/NhJoinClause.cs
```diff
@@ -3,60 +3,166 @@
...
+				if (string.IsNullOrEmpty(value)) throw new ArgumentException("Value cannot be null or empty.", nameof(value));
+				_itemName = value;
+			}
+		}
+
+		/// <summary>
+		///     Gets or sets the type of the items generated by this from clause.
+		/// </summary>
+		/// <note type="warning">
+		///     Changing the <see cref="P:Remotion.Linq.Clauses.NhFromClauseBase.ItemType" /> of a
```
fredericDelaporte: Invalid `cref`.

src/NHibernate/Linq/LinqExtensionMethods.cs
```diff
@@ -112,7 +112,7 @@ public static IEnumerable<TSource> ToFuture<TSource>(this IQueryable<TSource> so
 		/// <exception cref="T:System.ArgumentNullException"><paramref name="source" /> is <see langword="null"/>.</exception>
 		/// <exception cref="T:System.NotSupportedException"><paramref name="source" /> <see cref="IQueryable.Provider"/> is not a <see cref="INhQueryProvider"/>.</exception>
 		public static IFutureValue<TSource> ToFutureValue<TSource>(this IQueryable<TSource> source)
-		{
+			{
```
fredericDelaporte: Rebase glitch?

src/NHibernate/Linq/Visitors/QuerySourceLocator.cs
```diff
@@ -37,16 +37,30 @@ public override void VisitAdditionalFromClause(AdditionalFromClause fromClause,
...
-            }
-        }
-    }
-}
\ No newline at end of file
+		public override void VisitNhJoinClause(NhJoinClause joinClause, QueryModel queryModel, int index)
+		{
+			if (_type.IsAssignableFrom(joinClause.ItemType))
+			{
+				if (_querySource == null)
```
fredericDelaporte: Looks as an additional change, for which I do not get the rational. Is it fixing an untested case?

src/NHibernate/Linq/Visitors/WhereJoinDetector.cs
```diff
@@ -92,7 +93,23 @@ public void Transform(WhereClause whereClause)
...
+			var values = _values.Pop();
+
+			foreach (var memberExpression in values.MemberExpressions)
+			{
+				// If outer join can never produce true, we can safely inner join.
+				if (!values.GetValues(memberExpression).Contains(true))
+				{
+					_joiner.MakeInnerIfJoined(memberExpression);
+				}
+			}
```
fredericDelaporte: From [L100](https://github.com/nhibernate/nhibernate-core/pull/568/files#diff-4106d8fef553f297317a11b60b40b815R100) till here, could be refactored in a method with code from [L84](https://github.com/nhibernate/nhibernate-core/pull/568/files#diff-4106d8fef553f297317a11b60b40b815R84) to [L93](https://github.com/nhibernate/nhibernate-core/pull/568/files#diff-4106d8fef553f297317a11b60b40b815R93).

src/NHibernate/Linq/Visitors/QuerySourceLocator.cs
hazzik: It was here, see L30

fredericDelaporte: Ok, got confused by reading `VisitMainFromClause` instead of `VisitAdditionalFromClause`, sorry.

src/NHibernate/Linq/LinqExtensionMethods.cs
hazzik: Must be

src/NHibernate/Linq/Visitors/WhereJoinDetector.cs
hazzik: I'll do it later

fredericDelaporte:
I haven't seen you had made changes and so have not reviewed it again for having the merge "green". I you rather have an accepted review, feel free to "ping" me after changes or tell it is good enough to go without more changes.

PR #569, opened 2017-03-17
NH-3962 - Build with MSBuild Tools 2017

hazzik:


fredericDelaporte:
Somewhat related: there is an old MSBuild version in use for the documentation build. I had to change it locally for getting the `nhibernate-core\doc>nant doc` build succeeding at building the api doc using SHFB from GitHub instead of from VSHIK (which seems discontinued).

Maybe should this be updated too.

I have pushed the changes I have used so far locally [here](https://github.com/fredericDelaporte/nhibernate-core/commit/fe30159742e3b667cf040559ee70dcb99e3d4921), if you wish to consult them.

hazzik:
@fredericDelaporte thanks

fredericDelaporte:
Now that I code with VS2017 since some days, here is my opinion on that move for MSBuild tools 2017: I am for it.  
Since VS Community edition is available for free for open source projects and private individuals, I do not think we need to wait for newer VS version to gain a large audience before switching to them. It will not cost anything else than a couple download and install minutes for people not having it and wanting to contribute to NHibernate. Maybe should we add a notice in the two contributing files though.

Of course we need to avoid switching to a newer VS when it has some bad flaw. About VS 2017, I have encountered this [one](https://stackoverflow.com/q/43035875/1178314), with which I can cope and I do not consider it that bad.

hazzik:
Thanks. I'll merge it later

PR #574, opened 2017-03-19
NH-3964 - Refactor reflection patterns

fredericDelaporte:
[NH-3964](https://nhibernate.jira.com/browse/NH-3964) - This is a sort of follow up to NH-3952: remove usage of `EnumerableHelper`.
They were replaced by `ReflectionHelper`, and additionally cached whenever possible and useful.
This improvement aims at similarly caching other `MethodInfo`/`PropertyInfo` reflected through older `ReflectionHelper` usages (impacts `Linq` namespace), or any other means (most other NHibernate namespaces). 
If adequate, those other means of reflecting should be replaced by `ReflectionHelper`.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3964/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3964/Fixture.cs
@@ -1,0 +1,18 @@
+﻿using System.Collections;
+using NHibernate.Util;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3964
+{
+	[TestFixture]
+	public class Fixture
+	{
+		[Test(Description = "Test for removal of a workaround for an old Fx bug (<v4)")]
+		public void AddingNullToNonGenericListShouldNotThrow()
+		{
+			var a1 = new ArrayList { null };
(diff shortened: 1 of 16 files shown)


fredericDelaporte:
Rebased for resolving a conflict on test project.

fredericDelaporte:
By the way, I may refactor further: put the methods Hazzik was suggesting [here](https://github.com/nhibernate/nhibernate-core/pull/557#issuecomment-282392488) in `Util.ReflectHelper`:

>         public static MethodInfo GetMethod<T, TResult>(Func<T, TResult> func, T arg1)
>         {
>             return func.Method;
>         }

Then obsolete `Linq.ReflectionHelper` (> 200 calls to move to `ReflectHelper`).

hazzik:
@fredericDelaporte I was trying to hijack the PR and added refactoring to use `func.Method`, but it did no work well. I think for now the PR can be merged as is.

hazzik:
If you are going to do more work in this field, please use the same JIRA number.

fredericDelaporte:
Ok. I think `Linq.ReflectionHelper` should be consolidated into `Util.ReflectHelper`, and doing it with your methods would be better. I will probably come back later on that subject.

PR #575, opened 2017-03-19
NH-3247 - Char value gets 'cached' - fix and test cases

fredericDelaporte:
[NH-3247](https://nhibernate.jira.com/browse/NH-3247) - Char value gets 'cached' - fix and test cases.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3247/Entity.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3247/Entity.cs
@@ -1,0 +1,11 @@
+﻿using System;
+
+namespace NHibernate.Test.NHSpecificTest.NH3247
+{
+	class Entity
+	{
+		public virtual Guid Id { get; set; }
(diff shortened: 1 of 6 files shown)


fredericDelaporte:
Unfortunatly my test case is not supported by Oracle managed32 driver it seems.

>     Fixture.CharParameterValueShouldNotBeCached
>     ======= Failed test run #1 ==========
>     OneTimeSetUp: NHibernate.HibernateException : ORA-00904: : invalid identifier
>       ----> Oracle.ManagedDataAccess.Client.OracleException : ORA-00904: : invalid identifier

I do not have any Oracle knowledge. I may install some free edition if any to try check what happen and whether this test should simply be ignored for Oracle, or need some fixing. (Or is there some place in teamcity to gather more information on the failure?)

fredericDelaporte:
Forget it, the error code is sufficiently documented, the trouble is I have named a property `Initial` which is a reserved word for Oracle. Gonna escape it.

PR #576, opened 2017-03-20
Fixes NH-3965: Method 'SomeClass.Add_callback' does not have a method body. (on mono)

pruiz:
This is a (really) simple pull-req for https://nhibernate.jira.com/browse/NH-3965.

No unit-test, as this is only failing on mono.

### DIFF
--- a/src/NHibernate/Proxy/DynamicProxy/DefaultMethodEmitter.cs
+++ b/src/NHibernate/Proxy/DynamicProxy/DefaultMethodEmitter.cs
@@ -176,11 +176,16 @@ namespace NHibernate.Proxy.DynamicProxy
 				// arguments from the generated method. Using the open method definition
 				// directly works on .Net 2.0, which might be FW bug, but fails on 4.0.
 				//
+				// Also, on mono, we should use open generics type defimition too, as 
+				// stated by NH-3965.
+				//
 				// Equivalent pseudo-code:
 				// MethodInfo mi = methodof(TheClass.TheMethod<>)
 				// versus
 				// MethodInfo mi = methodof(TheClass.TheMethod<T0>)
-				var instantiatedMethod = method.MakeGenericMethod(generatedMethod.GetGenericArguments());
+				var instantiatedMethod = System.Type.GetType("Mono.Runtime") != null ?
+					method.GetGenericMethodDefinition()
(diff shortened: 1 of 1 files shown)


pruiz:
It looks like I will need to find a solution valid on both environments..

pruiz:
Ok, it looks like mono does still require an open-generics definition in order to work, so I am making a choice based on framework in order to create the proxy as open or closed.

src/NHibernate/Proxy/DynamicProxy/DefaultMethodEmitter.cs
```diff
@@ -176,11 +176,16 @@ private static void PushTargetMethodInfo(ILGenerator IL, MethodBuilder generated
 				// arguments from the generated method. Using the open method definition
 				// directly works on .Net 2.0, which might be FW bug, but fails on 4.0.
 				//
+				// Also, on mono, we should use open generics type defimition too, as
```
fredericDelaporte: Typo on definition word.

src/NHibernate/Proxy/DynamicProxy/DefaultMethodEmitter.cs
```diff
@@ -176,11 +176,16 @@ private static void PushTargetMethodInfo(ILGenerator IL, MethodBuilder generated
...
 				//
+				// Also, on mono, we should use open generics type defimition too, as 
+				// stated by NH-3965.
+				//
 				// Equivalent pseudo-code:
 				// MethodInfo mi = methodof(TheClass.TheMethod<>)
 				// versus
 				// MethodInfo mi = methodof(TheClass.TheMethod<T0>)
-				var instantiatedMethod = method.MakeGenericMethod(generatedMethod.GetGenericArguments());
+				var instantiatedMethod = System.Type.GetType("Mono.Runtime") != null ?
```
fredericDelaporte: Could this GetType be cached in a static variable? Or even the boolean telling we are under mono.

pruiz: Sure, I'll rework it this afternoon.

hazzik:
@pruiz can you please provide a test case?

hazzik:
As I understand this is no longer required.

PR #579, opened 2017-03-23
NH-3722 - Remove entity mode switching capability

hazzik:
* Ported from HHH-6330 https://github.com/hibernate/hibernate-orm/commit/4a4f636caf9ecc62fe0d230f422ad3eab2517db0
* ISession.GetSession is renamed to ISession.GetChildSession


src/NHibernate/Cfg/SettingsFactory.cs
```diff
@@ -268,11 +268,6 @@ public Settings BuildSettings(IDictionary<string, string> properties)
 			log.Info("Default flush mode: " + defaultFlushMode);
 			settings.DefaultFlushMode = defaultFlushMode;
 
-			EntityMode defaultEntityMode =
-				EntityModeHelper.Parse(PropertiesHelper.GetString(Environment.DefaultEntityMode, properties, "poco"));
-			log.Info("Default entity-mode: " + defaultEntityMode);
```
hazzik: Here, it's better to warn that default entity mode is deprecated rather then silently remove it.

src/NHibernate/ISession.cs
```diff
@@ -932,14 +929,12 @@ public interface ISession : IDisposable
...
-		/// </summary>
-		/// <param name="entityMode">The entity mode to use for the new session.</param>
+		///  <summary>
+		///  Starts a new Session. This secondary Session inherits the connection, transaction,
+		///  and other context information from the primary Session. It doesn't need to be flushed
+		///  or closed by the developer.
+		///  </summary>
 		/// <returns>The new session</returns>
-		ISession GetSession(EntityMode entityMode);
+		ISession GetChildSession();
```
hazzik: Need to add this method to 4.1.x and deprecate old one.

src/NHibernate/Persister/PersisterFactory.cs
```diff
@@ -127,18 +115,12 @@ public static class PersisterFactory
...
 		{
 			ConstructorInfo pc;
-			var use4Parameters = false;
 			try
 			{
 				pc = persisterClass.GetConstructor(CollectionPersisterConstructorArgs);
-				if (pc == null)
-				{
-					use4Parameters = true;
-					pc = persisterClass.GetConstructor(CollectionPersisterConstructor2Args);
```
fredericDelaporte: Could this removal be a breaking change? (Is it possible for NHibernate users to supply custom persister to NHibernate?)

src/NHibernate/Type/MetaType.cs
```diff
@@ -87,11 +86,6 @@ public override bool IsDirty(object old, object current, bool[] checkable, ISess
 			return checkable[0] && IsDirty(old, current, session);
 		}
 
-		public override object FromXMLNode(XmlNode xml, IMapping factory)
-		{
-			return FromXMLString(xml.Value, factory);
-		}
-
 		public object FromXMLString(string xml, IMapping factory)
```
fredericDelaporte: `FromXMLString` looks obsolete, no references found to it. A bit same for `ToXMLString`, only serves for `ToLoggableString` which could directly take `ToXmlString` implementation instead.

src/NHibernate/Type/NullableType.cs
```diff
@@ -343,16 +341,6 @@ public override bool[] ToColumnNullness(object value, IMapping mapping)
...
-		{
-			return FromXMLString(xml.InnerText, factory);
-		}
-
-		public override void SetToXMLNode(XmlNode xml, object value, ISessionFactoryImplementor factory)
-		{
-			xml.InnerText = ToXMLString(value, factory);
-		}
-
 		public string ToXMLString(object value, ISessionFactoryImplementor pc)
```
fredericDelaporte: Same as corresponding methods of `MetaType`: looks obsolete.

hazzik:
>Are we removing the dynamic-map mode altogether? I do not see how we are supposed to activate it since default_entity_mode parameter is removed. On Hibernate side, it looks like they are still supporting it.

The entity mode is supported only on a entity/component level, eg. entity/component has to be defined as dynamic-map, to be able to use it. Please see `DynamicClassFixture`.

Default entity mode was passed only to a SessionImpl constructor, so it's not needed and do not have any effect now.

The drivers for the changes are ComponentMetamodel and EntityMetamodel: replace EntityModeToTuplizerMapping with static ITuplizer and remove all unused EntityModes

>src/UpgradeLog2.htm: that file is probably an unexpected add.
Sure.

src/NHibernate/Persister/PersisterFactory.cs
hazzik: It is, it needs to be deprecated through additional release of 4.x version

hazzik:
>nhibernate-mapping.xsd: still references xml entity mode (line 1435).

I'm not sure how shall we handle xsd schema changes.

hazzik:
>And there is a number of embed-xml attributes which should be removed if they were for the removed embedded xml. (I do not find any reference to those attributes, even in master.)

Because the EntityMode.Xml was never actually implemented

hazzik:
Code issues addressed. I'll update documentation later

fredericDelaporte:
> I'm not sure how shall we handle xsd schema changes.

Me too, but it looks to me your 2 commits have done it right. I guess changing the xsd namespace version would be a hassle for migration, requiring everyone to change it in their existing mappings.

Great work by the way, code base simplification is always welcome.

fredericDelaporte:
On the subject of methods to obsolete, I wonder if this is not a bit late to obsolete them in 4.x for suppressing them completely in 5.
Many users may step over last 4.x versions for upgrading directly to 5.
Maybe should we put them back in 5 but marked obsolete, and remove them only in 6.

hazzik:
I think we can do it for `ISession.GetSession(EntityMode)`. For other methods this does not really make a lot of sense. The thing is, that we are now (in 5.0) in an unique position to reshuffle the API as we need to be able to add new features in the future.

PR #580, opened 2017-03-24
NH-2520 - Add TimestampUtc type

hazzik:
Replaces #227


fredericDelaporte:
With current #545 failure, I realize this PR contains a potentially flaky test: it creates a timestamp versioned entity and re-read it from db, then test its version is less than one second off current utc time.

But sometimes it appears TeamCity build servers performances are quite sluggish, so one second is a bit too short. 30 seconds will be more safe. And since I guess the aim was to detect a local time timestamp instead of an utc one, it could be anything up to a couple of minutes. (Currently, even near but less than one hour : there are zones off by a quarter or an half hour, but not around +0.)

PR #581, opened 2017-03-24
NH-2520 - update documentation for including new TimestampUtc type.

fredericDelaporte:
NH-2520 follow up for updating documentation in order to include new `TimestampUtc` type.

I find its name a bit unfortunate, in regard to already existing `UtcDateTime`. It should have been named `UtcTimestamp`, or we should obsolete `UtcDateTime` and replace it with a `DateTimeUtc`.

Let me know if agreeing with my viewpoint, I would then add some commit to this PR for renaming this new type.

Otherwise I will probably merge this tomorrow.

Done some cleanup on documentation. Non white-space changes range from line 2603 to 2790.

### DIFF
--- a/doc/reference/modules/basic_mapping.xml
+++ b/doc/reference/modules/basic_mapping.xml
@@ -52,14 +52,12 @@
 </hibernate-mapping>]]></programlisting>
 
         <para>
-             We will now discuss the content of the mapping document. We will only describe the 
-             document elements and attributes that are used by NHibernate at runtime. The mapping 
-             document also contains some extra optional attributes and elements that affect the 
+             We will now discuss the content of the mapping document. We will only describe the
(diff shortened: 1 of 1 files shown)


fredericDelaporte:
`TimestampUtc` tests fail with Firebird, . Error code `-104`, failing in merged [PR](https://github.com/nhibernate/nhibernate-core/pull/580) too. This error code is insufficient for knowing what is the issue, it has more than 30 error cases (see this [doc](https://firebirdsql.org/file/documentation/reference_manuals/reference_material/Firebird-2.1-ErrorCodes.pdf), not found a newer version).
I guess we need to debug it locally if we want to have them pass with Firebird.

hazzik:
The tests has been fixed

PR #582, opened 2017-03-26
NH-3931 - Invalid order of child inserts

fredericDelaporte:
Test cases and fix for [NH-3931](https://nhibernate.jira.com/browse/NH-3931).

Test cases ported from HHH-9864, HHH-11216 and HHH-11585, relaxed in terms of batching expectations, and completed with some missing cases.

Fix not ported but elaborated from current NHibernate implementation.

This PR will hopefully replace the [previous one](https://github.com/nhibernate/nhibernate-core/pull/577), which was attempting a new sorting algorithm. This new algorithm was better at batching some cases, but was failing on many previously supported cases. And it was having an O(n^3) complexity over batched entities classes count, where current algorithm, even fixed, has an O(n) complexity over batched entities count.

### DIFF
--- a/src/NHibernate/Engine/ActionQueue.cs
+++ b/src/NHibernate/Engine/ActionQueue.cs
@@ -6,6 +6,7 @@ using System.Text;
 
 using NHibernate.Action;
 using NHibernate.Cache;
+using NHibernate.Type;
 
 namespace NHibernate.Engine
 {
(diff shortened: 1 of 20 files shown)


src/NHibernate/Engine/ActionQueue.cs
```diff
@@ -565,36 +579,47 @@ private int GetBatchNumber(EntityInsertAction action, string entityName)
...
 				return batchNumber;
 			}
 
-			private bool IsProcessedAfterAllAssociatedEntities(EntityInsertAction action, int latestBatchNumberForType)
+			private bool RequireNewBatch(EntityInsertAction action, int latestBatchNumberForType)
 			{
+				// This method assumes the original action list is already sorted in order to respect dependencies.
 				var propertyValues = action.State;
-				var propertyTypes = action.Persister.ClassMetadata.PropertyTypes;
+				var propertyTypes = action.Persister.EntityMetamodel?.PropertyTypes;
```
fredericDelaporte: `ClassMetadata` current NHibernate implementation is reaching the underlying `EntityMetamodel.PropertyTypes`. As `ClassMetadata` seems "more optional" than `EntityMetamodel`, better use `EntityMetamodel` directly. (`CustomPersister` on NHibernate side does not provide any of them, but on Hibernate side, it does provide `EntityMetamodel` while still not providing `ClassMetadata`.)

hazzik:
@fredericDelaporte next time, if you want to preserve multiple commits on a PR, please make a merge commit.

So:
1. If PR has a single commit - rebase on top of `master` (if it is already on top of master, just apply it without merge commit).
3. If PR has multiple commits, and you want to preserve them - make a merge commit.
2. If PR has multiple commits, but you do not want them, then squash PR to `master`.

fredericDelaporte:
Ok, understood.

PR #584, opened 2017-03-26
NH-3850 - Polymorphic Linq query results aggregators failures

fredericDelaporte:
Fix and test cases for [NH-3850](https://nhibernate.jira.com/browse/NH-3850).

Currently, LINQ queries such as `query.Count()`, `query.Any()` do not support polymorphic queries: they take into account only the results of the first entity class matching the query and ignores the other classes results.

This fix implements polymorphic support for `Aggregate`, `Any`, `Count`, `LongCount`, `Max`, `Min` and `Sum`.  
`Average` is not fixed: supporting it requires to select additional columns in the query (`Sum` and `Count`) and compute the average on client side. I do not think it is worth the hassle.

This fix introduces a possible breaking change: `Sum` on nullables will no more yield `null` when there is no rows to compute from, but will yield `0`. This change is due to usage of `Enumerable.Sum` for aggregating polymorphic results, and this is its documented behavior. This happens even on non polymorphic queries: when adding the `Enumerable.Sum` post result transformer, we do not already knows if the query will be polymorphic, so we add it to non polymorphic queries too. (Determining if the query is polymorphic would require some computation currently done in query plan computation.)  
Avoiding this change should be doable by building a more elaborated expression as post result transformer. But this changes looks to me very minor and not contradictory with Linq contracts. It causes the loss of the ability to tell from the result there was nothing to `Sum`, but that does not look to me as being part of `Sum` contract. So I have not implemented a workaround for avoiding this change.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3850/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3850/Fixture.cs
@@ -1,0 +1,1255 @@
+﻿using System;
+using System.Data;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using NHibernate.Dialect;
+using NHibernate.Driver;
+using NHibernate.Linq;
+using NHibernate.Util;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3850
+{
+	[TestFixture]
+	public class Fixture : BugTestCase
+	{
+		private const string _searchName1 = "name";
+		private const string _searchName2 = "name2";
+		private const int _totalEntityCount = 10;
+		private readonly DateTime _testDate = DateTime.Now;
+		private readonly DateTimeOffset _testDateWithOffset = DateTimeOffset.Now;
+		
+		protected override bool AppliesTo(Dialect.Dialect dialect)
+		{
+			var typeNames = (TypeNames)typeof(Dialect.Dialect).GetField("_typeNames", ReflectHelper.AnyVisibilityInstance).GetValue(Dialect);
+			try
+			{
+				typeNames.Get(DbType.DateTimeOffset);
+			}
+			catch (ArgumentException)
+			{
+				return false;
+			}
+
+			return true;
+		}
+
+		protected override bool AppliesTo(Engine.ISessionFactoryImplementor factory)
+		{
+			// Cannot handle DbType.DateTimeOffset via ODBC.
+			return !(factory.ConnectionProvider.Driver is OdbcDriver);
+		}
+
+		protected override void OnSetUp()
+		{
+			base.OnSetUp();
+			using (var session = OpenSession())
+			{
+				var dateTime1 = _testDate.AddDays(-1);
+				var dateTime2 = _testDate.AddDays(1);
+				var dateTimeOffset1 = _testDateWithOffset.AddDays(-1);
+				var dateTimeOffset2 = _testDateWithOffset.AddDays(1);
+				Action<DomainClassBase> init1 = dc =>
+				{
+					dc.Id = 1;
+					dc.Name = _searchName1;
+					dc.Integer = 1;
+					dc.Long = 1;
+					dc.Decimal = 1;
+					dc.Double = 1;
+					dc.DateTime = dateTime1;
+					dc.DateTimeOffset = dateTimeOffset1;
+					dc.NonNullableDecimal = 1;
+				};
+				Action<DomainClassBase> init2 = dc =>
+				{
+					dc.Id = 2;
+					dc.Name = _searchName2;
+					dc.Integer = 2;
+					dc.Long = 2;
+					dc.Decimal = 2;
+					dc.Double = 2;
+					dc.DateTime = dateTime2;
+					dc.DateTimeOffset = dateTimeOffset2;
+					dc.NonNullableDecimal = 2;
+				};
+
+				DomainClassBase entity = new DomainClassBExtendedByA();
+				init1(entity);
+				session.Save(entity);
+				entity = new DomainClassBExtendedByA();
+				init2(entity);
+				session.Save(entity);
+
+				entity = new DomainClassCExtendedByD();
+				init1(entity);
+				session.Save(entity);
+				entity = new DomainClassCExtendedByD();
+				init2(entity);
+				session.Save(entity);
+
+				entity = new DomainClassE();
+				init1(entity);
+				session.Save(entity);
+				entity = new DomainClassE();
+				init2(entity);
+				session.Save(entity);
+
+				entity = new DomainClassGExtendedByH();
+				init1(entity);
+				session.Save(entity);
+				entity = new DomainClassGExtendedByH();
+				init2(entity);
+				session.Save(entity);
+				entity = new DomainClassHExtendingG
+				{
+					Id = 3,
+					Name = _searchName1,
+					Integer = 3,
+					Long = 3,
+					Decimal = 3,
+					Double = 3,
+					DateTime = dateTime1,
+					DateTimeOffset = dateTimeOffset1,
+					NonNullableDecimal = 3
+				};
+				session.Save(entity);
+				entity = new DomainClassHExtendingG
+				{
+					Id = 4,
+					Name = _searchName2,
+					Integer = 4,
+					Long = 4,
+					Decimal = 4,
+					Double = 4,
+					DateTime = dateTime2,
+					DateTimeOffset = dateTimeOffset2,
+					NonNullableDecimal = 4
+				};
+				session.Save(entity);
+
+				session.Flush();
+			}
+		}
+
+		protected override void OnTearDown()
+		{
+			base.OnTearDown();
+			using (var session = OpenSession())
+			{
+				var hql = "from System.Object";
+				session.Delete(hql);
+				session.Flush();
+			}
+		}
+
+		// Failing case till NH-3850 is fixed
+		[Test]
+		public void AggregateGBase()
+		{
+			using (var session = OpenSession())
+			{
+				// This case should work because the aggregate is insensitive to ordering.
+				var result = session.Query<DomainClassGExtendedByH>()
+					.OrderBy(dc => dc.Id)
+					.Select(dc => dc.Id)
+					.Aggregate((p, n) => p + n);
+				Assert.AreEqual(10, result);
+			}
+		}
+
+		// Failing case due to lack of polymorphic results ordering.
+		[Test, Ignore("Polymorphic results sets are unioned without reordering, whatever the API")]
+		public void AggregateGBaseOrderingMismatch()
+		{
+			using (var session = OpenSession())
+			{
+				// This case cannot work because the aggregate is sensitive to ordering, and NHibernate currently always order polymorphic queries by class names,
+				// then only honors query ordering as secondary order criteria.
+				var result = session.Query<DomainClassGExtendedByH>()
+					.OrderByDescending(dc => dc.Id)
+					.Select(dc => dc.Id.ToString())
+					.Aggregate((p, n) => p + "," + n);
+				// Currently yields "2,1,4,3" instead.
+				Assert.AreEqual("4,3,2,1", result);
+			}
+		}
+
+		// Non-reg case
+		[Test]
+		public void AggregateMutableSeedGBase()
+		{
+			using (var session = OpenSession())
+			{
+				// This case works because the ordering accidentally matches with classes ordering.
+				// (And moreover, with current dataset, selected values are same whatever the classes.)
+				var result = session.Query<DomainClassGExtendedByH>()
+					.OrderBy(dc => dc.Id)
+					.Aggregate(new StringBuilder(), (s, dc) => s.Append(dc.Name).Append(","));
+				Assert.AreEqual(_searchName1 + "," + _searchName2 + "," + _searchName1 + "," + _searchName2 + ",", result.ToString());
+			}
+		}
+
+		// Failing case till NH-3850 is fixed
+		[Test]
+		public void AggregateSeedGBase()
+		{
+			using (var session = OpenSession())
+			{
+				// This case should work because the aggregate is insensitive to ordering.
+				var result = session.Query<DomainClassGExtendedByH>()
+					.OrderBy(dc => dc.Id)
+					.Aggregate(5, (s, dc) => s + dc.Id);
+				Assert.AreEqual(15, result);
+			}
+		}
+
+		// Failing case till NH-3850 is fixed
+		[Test]
+		public void AllBBaseWithName()
+		{
+			using (var session = OpenSession())
+			{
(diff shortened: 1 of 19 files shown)


fredericDelaporte:
@gliljas, @hazzik, does the change on `Sum` behavior looks ok for you?

> This fix introduces a possible breaking change: `Sum` on nullables will no more yield `null` when there is no rows to compute from, but will yield `0`. This change is due to usage of `Enumerable.Sum` for aggregating polymorphic results, and this is its documented behavior. This happens even on non polymorphic queries: when adding the `Enumerable.Sum` post result transformer, we do not already knows if the query will be polymorphic, so we add it to non polymorphic queries too. (Determining if the query is polymorphic would require some computation currently done in query plan computation.)  
> Avoiding this change should be doable by building a more elaborated expression as post result transformer. But this changes looks to me very minor and not contradictory with Linq contracts. It causes the loss of the ability to tell from the result there was nothing to `Sum`, but that does not look to me as being part of `Sum` contract. So I have not implemented a workaround for avoiding this change.

gliljas:
I'm not sure. It satisfies the LINQ contract (which is probably paramount) but sacrifices SQL compatibility. Not a big problem, but it becomes a bit unpredictable if this only happens on results.

customersQueryable.Where(x=>x.Orders.Sum(y=>y.Amount)>0)

will not return customers with no orders. In a LINQ to Objects query, it would.

I've handled this is my current project by introducing an expression visitor which rewrites all such queries, adding a Coalesce to the expression tree. It adds query cost, so it's not used everywhere.

fredericDelaporte:
Good point. Well I will have a look to check if that is that hard to preserve this behavior.

fredericDelaporte:
Ok, done. At first it was looking easy, but I had discovered another polymorphic case still not supported : aggregation of non-nullable values when some concrete class have no rows when other was. I have fixed it.

Now there is still a breaking change: when aggregating non nullable values while there is no rows for all concrete classes, it yields an exception, as expected. But previously, it was a GenericADOException having an ArgumentNullException as inner exception. Now it is an InvalidOperationException "Value cannot be null".

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -128,16 +133,148 @@ static QueryModelVisitor()
...
+				// NH-3850: handle polymorphic scalar results aggregation
+				switch ((NhExpressionType)Model.SelectClause.Selector.NodeType)
+				{
+					case NhExpressionType.Average:
+						// Polymorphic case complex to handle and not implemented. (HQL query must be reshaped for adding
+						// additional data to allow a meaningful overall average computation.)
+						// Leaving it untouched for allowing non polymorphic cases to work.
+						break;
+					case NhExpressionType.Max:
+						AddPostExecuteTransformerForResultAggregate(ReflectionCache.QueryableMethods.MaxDefinition);
```
hazzik: Why do you use `IQueryable<>` for aggregate methods? It does not make sense for me.

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -128,16 +133,148 @@ static QueryModelVisitor()
...
+
+		private void AddPostExecuteTransformerForResultAggregate(MethodInfo aggregateMethodTemplate)
+		{
+			var elementType = Model.ResultTypeOverride;
+
+			LambdaExpression aggregateLambda;
+
+			if (elementType.IsNullableOrReference())
+			{
+				var inputListType = typeof(IQueryable<>).MakeGenericType(elementType);
```
hazzik: Again, why `IQueryable<>`?

src/NHibernate.Test/NHSpecificTest/NH3850/Fixture.cs
```diff
@@ -0,0 +1,1314 @@
...
+
+		// Non-reg case
+		[Test]
+		public void SumE()
+		{
+			Sum<DomainClassE>(3);
+		}
+
+		// Non-reg case
+		[Test, Ignore("Candidate potential breaking changes: Sum would now yield 0 instead of null on empty result sets.")]
```
hazzik: When I removed `Ignore` test did pass.

hazzik: It's because the query isn't polymorphic

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
fredericDelaporte: The `IntermediateHqlTree._postExecuteTransformers` is a list of `LambdaExpression` which get merged together then compiled and called on `IQueryable<T>`. As if they should be chainable, as if we it were useful and possible to add many of them, though the usages I have currently seen for them were all incompatible with chaining (they all yield another type than their input).

Yes that does not make sens to me either, that overly complicates things for defining the post execute transform (messing with expressions for runtime execution instead of just supplying directly an executable delegate), but there are no comments as for why such a design decision, and I have not decided to redesign this.

Redesign those transformers should imply considering the design of the list transformer, which works the same way and is way more used. But maybe this one has some cases where it could be translated to DB, though I doubt.

src/NHibernate.Test/NHSpecificTest/NH3850/Fixture.cs
fredericDelaporte: Rebase & conflict resolution miss, I had removed it prior to rebase. Going to remove it again.

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
hazzik: The `IEnumerable<>` just works here.

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
fredericDelaporte: With my previous answer in mind, it would work with the `IEnumerable` version too for current usage of the post execute transformer.  I have made that choice for aligning with current input type of post execute transformers. True, there is an `AsQueryable` call just for supplying that input type to it. I really do not understand what were the design rationals behind the way this post execute transformer currently works. It looks to me as lot of overhead for no added value. (Enable  "merge" of execute transformer, but none of existing one seems mergeable to me. Maybe some serialization needs again. All that ends up in the query plan, which is serializable.)

With the current state of usage of this post execute transformer, it should works with `IEnumerable` methods too.

hazzik: It requires `.AsQueryable()`/`IQueryable<>` for some cases where the method is from the original query expression (First, FirstOrDefault, Single, SingleOrDefault).

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -128,16 +133,148 @@ static QueryModelVisitor()
...
+			// will check and adjust it if it is long instead of int (GetAggregateMethodCall does that).
+			var aggregateCall = GetAggregateMethodCall(ReflectionCache.QueryableMethods.SumOnInt, inputListType, concreteQueryElementType, inputList);
+
+			var allMethod = ReflectionCache.EnumerableMethods.AllDefinition.MakeGenericMethod(concreteQueryElementType);
+			var element = Expression.Parameter(concreteQueryElementType, "element");
+			var noSumCondition = Expression.Equal(element, Expression.Constant(null));
+			var allCall = Expression.Call(allMethod, inputList, Expression.Lambda(noSumCondition, element));
+
+			Expression conditionalAggregateCall = Expression.Condition(allCall,
+				Expression.Convert(Expression.Constant(null), concreteQueryElementType),
```
hazzik: Here we do not need cast. `Expression.Constant` has an overload which accepts a type.

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
hazzik: >I really do not understand what were the design rationale behind the way this post execute transformer currently works

I think it's just a copy of ListTransformer implementation, which indeed chainable.

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
fredericDelaporte: Changing it on the above constant too, better to compare both side having the same type.

fredericDelaporte:
Changes done. Added some more comments in code too.

src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessAggregate.cs
```diff
@@ -19,20 +19,22 @@ public void Process(AggregateResultOperator resultOperator, QueryModelVisitor qu
...
 
-			var inputList = Expression.Parameter(typeof(IEnumerable<>).MakeGenericType(typeof(object)), "inputList");
-
-			var castToItem = ReflectionCache.EnumerableMethods.CastDefinition.MakeGenericMethod(new[] { inputType });
-			var castToItemExpr = Expression.Call(castToItem, inputList);
-
+			// NH-3850: changed from list transformer (working on IEnumerable<object>) to post execute
+			// transformer (working on IEnumerable<inputType>) for globally aggregating polymorphic results
+			// instead of aggregating results for each class separately and yielding only the first.
+			// If the aggregation relies on ordering, final result will still be wrong due to
```
hazzik: >If the aggregation relies on ordering, final result will still be wrong

Can you demonstrate this with a test?

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -128,16 +133,150 @@ static QueryModelVisitor()
...
+				// On non nullable, we have to preserve the behavior "nothing to aggregate => exception",
+				// while supporting the polymorphic case "nothing to aggregate for some classes but something to aggregate
+				// for others" => aggregate of others.
+				// For this, overriding result type of concrete queries as nullable, aggregating concrete results,
+				// casting back to non nullable (and failing if null).
+				// This still causes a change: the expected failure case will yield an InvalidOperationException instead of
+				// a GenericADOException with an ArgumentNullException inner exception.
+
+				var nullableElementType = typeof(Nullable<>).MakeGenericType(new[] { elementType });
+				var nullableInputListType = typeof(IQueryable<>).MakeGenericType(nullableElementType);
```
hazzik: You need to use `IEnumerable<>` here (and everywhere else).

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -128,16 +133,150 @@ static QueryModelVisitor()
...
+			{
+				// On non nullable, we have to preserve the behavior "nothing to aggregate => exception",
+				// while supporting the polymorphic case "nothing to aggregate for some classes but something to aggregate
+				// for others" => aggregate of others.
+				// For this, overriding result type of concrete queries as nullable, aggregating concrete results,
+				// casting back to non nullable (and failing if null).
+				// This still causes a change: the expected failure case will yield an InvalidOperationException instead of
+				// a GenericADOException with an ArgumentNullException inner exception.
+
+				var nullableElementType = typeof(Nullable<>).MakeGenericType(new[] { elementType });
```
hazzik: Please do not use array creation in `MakeGenericType`

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
fredericDelaporte: Reviewing again some other parts of the fix, it appears to me I was not consistent with that. This fix has spanned a long period, I have ended up not remembering all of it.

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
fredericDelaporte: Sorry for having not spotted the `params` modifier. Added to the new `ReflectionHelper.GetMethodOverload` by the way.

fredericDelaporte:
Newer changes done. And removed some unneeded private setter, internal => private.

src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessAggregate.cs
fredericDelaporte: Test added. By the way, hardened the non mutable seed test for it to be no more sensible to ordering. I prefer avoid relying on current polymorphic classes ordering whenever possible for tests, since it does look undocumented.

fredericDelaporte:
Some additional considerations about those post result transformer: I have contemplated adding a `PolymorphicResultTransformer` into the mix, which would be call before the post result transformer, only if the query yielded many polymorphic results.

This would avoid running this transformer on non polymorphic cases. But for handling polymorphic cases requiring a type override, this would require an additional `PolymorphicTypeOverride` for applying it if the query is polymorphic, while keeping the normal type otherwise. (And the aggregate case would require even more tweaking, such as an option for discarding the list transformer if the query is polymorphic, ...)

Moreover, the main overhead is probably currently more in constructing this transformer than in using it.

Now if we wanted to fix the ordering issue of polymorphic issue, we could try to extract the order by from the Linq query for reapplying it on union-ed polymorphic results. But then, re-ordering non polymorphic results would be a significant overhead without any value. There a `PolymorphicResultTransformer` for re-ordering only polymorphic cases could be useful.

hazzik:
After all these efforts I'm thinking about banning all implicit polymorphism with LINQ.

Otherwise the cases needs to be supported as well:
- ordering
- paging

fredericDelaporte:
While ordering alone looks achievable, though already bad for performances (in-memory ordering required), paging support with ordering would be quite a monster for performances, especially for reaching far away pages. Depending on required order, the polymorphic page could have to pick elements ranging from row 0 up to "end of required page row number" in each concrete table. This would cause paging to be pointless.

Polymorphism with other API are not in a better state, they do not support paging or ordering better. Indeed, polymorphic aggregates with HQL yields as many rows as concrete classes. So taking the first achieve the same result than current Linq. Trying `UniqueResult` blows away.

Concrete results set are just union-ed without ordering with HQL too. Linq an HQL run the same code for that. There is some code for trying a paging accounting polymorphism, but without handling the ordering issue, and as far as I remember, quite flawed (was not yielding pages of expected size I believe).

(I have not checked the criteria/queryover cases.)

So well, why not disabling it for all API if we decide to disable it for Linq?
Apart for test teardown (`delete * from System.Object`), are there many actual usages of implicit polymorphism? (In the apps I had to work with, it never served me.)

fredericDelaporte:
Rebased for resolving conflicts with reflection refactoring.

fredericDelaporte:
Fix some bad white-spaces I had inadvertently introduced, fix some comments and adjusted some reflection cache parameter specification syntax (using `default` instead of cast or instanciation of structs).

PR #585, opened 2017-03-27
NH-3970 - avoid losing original failure in case of additional failure

fredericDelaporte:
Test cases (explicit execution), and fix for [NH-3970](https://nhibernate.jira.com/browse/NH-3970) - TestCase base class: avoid hiding test failure on tear-down.

There is many things I would like to fix in the tests, but this one is really annoying. When a test fails but we cannot have its root cause (at least in NUnit gui), it is quite impractical.

### DIFF
--- a/src/NHibernate.Test/TestCase.cs
+++ b/src/NHibernate.Test/TestCase.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections;
 using System.Data;
-using System.Data.Common;
 using System.Reflection;
 using log4net;
 using log4net.Config;
@@ -13,6 +12,10 @@ using NHibernate.Tool.hbm2ddl;
 using NHibernate.Type;
 using NUnit.Framework;
 using NHibernate.Hql.Ast.ANTLR;
+using System.Collections.Concurrent;
+using NUnit.Framework.Interfaces;
+using System.Text;
+using static NUnit.Framework.TestContext;
 
 namespace NHibernate.Test
 {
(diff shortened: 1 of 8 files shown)


src/NHibernate.Test/TestCase.cs
```diff
@@ -154,30 +158,81 @@ protected virtual void OnTearDown()
...
+		{
+			var allClosed = true;
+			foreach (var session in _openedSessions)
 			{
-				log.Error("Test case didn't close a session, closing");
-				lastOpenedSession.Close();
-				return false;
+				if (session.IsOpen)
+				{
+					if (allClosed)
```
hazzik: Maybe report each closing session separately? 

```
log.Error("Test case didn't close a session " + session.SessionId + " , closing");
```

fredericDelaporte: And may help finding the session "at fault".

src/NHibernate.Test/TestCase.cs
```diff
@@ -154,30 +158,78 @@ protected virtual void OnTearDown()
...
-			if (lastOpenedSession != null && lastOpenedSession.IsOpen)
+			var allClosed = true;
+			foreach (var session in _openedSessions)
 			{
-				log.Error("Test case didn't close a session, closing");
-				lastOpenedSession.Close();
-				return false;
+				if (session.IsOpen)
+				{
+					log.Error($"Test case didn't close session {(session as ISessionImplementor)?.SessionId}, closing");
```
hazzik: It perhaps should be `session.GetSessionImplementor().SessionId`

fredericDelaporte: Huh, I was not expecting the internal contract to be available from public interface!

PR #586, opened 2017-03-27
NH-3964 - obsolete Linq.ReflectionHelper in favor of Util.ReflectHelper.

fredericDelaporte:
Follow up of [NH-3964](https://nhibernate.jira.com/browse/NH-3964) and its first PR #574 .

Methods are just moved to ReflectHelper, without changing them.

Trying to use

>    public static MethodInfo GetMethod<T, TResult>(Func<T, TResult> func, T arg1)
>    {
>        return func.Method;
>    }

is requiring as many overloads as function arguments counts

    public static MethodInfo GetMethod<T1, T2, TResult>(Func<T1, T2, TResult> func, T1 arg1, T2 arg2)
    {
        return func.Method;
    }

Then type inference is sometime inferior:

    return GetMethod(Enumerable.GroupBy, (IEnumerable<int>)null, (Func<int, decimal>)null);

Instead of 

    return GetMethod(() => Enumerable.GroupBy(null, (Func<int, decimal>)null));

And handling instance methods are requiring a `Func<Func<>>` with a non-null instance:

	public static MethodInfo GetMethod<TSource, T1, TResult>(Func<TSource, Func<T1, TResult>> func, TSource source, T1 arg1)
	{
		return func(source).Method;
	}

    GetMethod<string, int, string>(s => s.Substring, "", 10);

So well, even though some "brute force" runtime performance comparison show it is "better", I do not think it is worth all the required changes and additional code for using them.

### DIFF
--- a/src/NHibernate/Util/ReflectHelper.cs
+++ b/src/NHibernate/Util/ReflectHelper.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Reflection;
 using System.Text;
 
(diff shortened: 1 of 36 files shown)


hazzik:
* We need about 4 overloads for GetMethodInfo & 4 overloads for GetMethodDefinition, it's true.
* There are only few cases where we do not have an instance upfront. For string we can use `string.Empty` and for structs `default(T)` (where `T` is a struct). Where we need a real instance we can use the old method.
* `(IEnumerable<int>)null` looks clearer as `default(IEnumerable<int>)`

fredericDelaporte:
Snapshot dependency failure for failed build: looks like a Team City infrastructure failure, doesn't it? I should probably trigger a later rebuild with a rebase.

src/NHibernate/Util/ReflectHelper.cs
```diff
@@ -565,7 +651,7 @@ internal static object GetConstantValue(string qualifiedName, ISessionFactoryImp
 			return null;
 		}
 
-		[Obsolete("Please use Linq.ReflectionHelper instead")]
+		[Obsolete("Please use Linq.ReflectHelper instead")]
```
fredericDelaporte: Well, search&replace does not make it for this obsolete reason. I would fix it tomorrow for triggering another build by the way.

fredericDelaporte:
Amended for updating a doc example referencing `ReflectionHelper`, and rebased.

fredericDelaporte:
Found two additional method reflection which are cacheable and were not. Now cached. And rebased.

PR #590, opened 2017-04-04
NH-3977 - Thread safety weaknesses of MapBasedSessionContext

fredericDelaporte:
[NH-3977](https://nhibernate.jira.com/browse/NH-3977) - Thread safety weaknesses of MapBasedSessionContext, test case and fix.

Minor improvement for avoiding a theoretical thread safety issue.

### DIFF
--- a/src/NHibernate/Context/MapBasedSessionContext.cs
+++ b/src/NHibernate/Context/MapBasedSessionContext.cs
@@ -1,5 +1,5 @@
 using System.Collections;
-
+using System.Collections.Concurrent;
 using NHibernate.Engine;
 
 namespace NHibernate.Context
@@ -7,6 +7,9 @@ namespace NHibernate.Context
 	public abstract class MapBasedSessionContext : CurrentSessionContext
 	{
 		private readonly ISessionFactoryImplementor _factory;
+
+		// Must be static, different instances of MapBasedSessionContext may have to yield the same map.
+		private static readonly object _locker = new object();
 
 		protected MapBasedSessionContext(ISessionFactoryImplementor factory)
 		{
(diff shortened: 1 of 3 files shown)


src/NHibernate/Context/MapBasedSessionContext.cs
```diff
@@ -20,30 +23,35 @@ protected override ISession Session
...
-				IDictionary map = GetMap();
-				if (map == null)
-				{
-					return null;
-				}
-				else
-				{
-					return map[_factory] as ISession;
-				}
+				return GetConcreteMap()?.GetOrAdd(_factory, default(ISession));
```
fredericDelaporte: A `GetOrDefault` would have been enough here, but as it does not exists, `GetOrAdd` is the way to go. This potential add will not harm: having no entry or having `null` is equivalent. And its semantics ensures it will not override a concurrent item set from the setter.

`default(ISession)` required instead of `null` due to overload ambiguity.

src/NHibernate/Context/MapBasedSessionContext.cs
```diff
@@ -20,30 +23,35 @@ protected override ISession Session
...
+		private ConcurrentDictionary<ISessionFactoryImplementor, ISession> GetConcreteMap()
+		{
+			return (ConcurrentDictionary<ISessionFactoryImplementor, ISession>)GetMap();
+		}
+
 		/// <summary>
-		/// Get the dictionary mapping session factory to its current session.
+		/// Get the dictionary mapping session factory to its current session. Yield <c>null</c> if none have been set.
 		/// </summary>
 		protected abstract IDictionary GetMap();
```
fredericDelaporte: I could have changed the returned type for the concrete class type in order to avoid the above cast, but this would be a breaking change for external implementers, and this would expose a technical detail.

src/NHibernate/Context/MapBasedSessionContext.cs
hazzik: Shouldn't it be TryGetValue?

fredericDelaporte: More correct as of what this getter has to do, you are right.
A bit more verbose:

```C#
ISession value = null;
GetConcreteMap()?.TryGetValue(_factory, out value);
// We want null if no value was there, no need to explicitly handle false outcome of TryGetValue.
return value;
```

Or:

```C#
var map = GetConcreteMap();
if (map == null)
	return null;
map.TryGetValue(_factory, out var value);
// We want null if no value was there, no need to explicitly handle false outcome of TryGetValue.
return value;
```

fredericDelaporte:
Completed a comment about why Lazy<T> was not usable, rebased and squashed.

PR #591, opened 2017-04-05
NH-3975 - Synchronize some features dialect support properties

fredericDelaporte:
[NH-3975](https://nhibernate.jira.com/browse/NH-3975) - Synchronize some features dialect support properties (Informational metadata).

A bunch of "Informational metadata" properties are defined on the base dialect class, and allow tests to ignore dialects which do not support the tested feature. Unfortunately, some of them are not up-to-date and causes such tests to fail instead of ignoring the dialect.

This PR aims at updating those properties according to their current valu in Hibernate Java dialects implementation.

This has been done in a very conservative manner: each such property have been checked for not having any usage inside NHibernate (being purely informational), otherwise I have left them untouched. Some NHibernate dialect are overriding some of those properties while their Java counterpart do not, I have kept NHibernate overrides.

It would be probably nice to do a complete dialect sync, but this would have heavy transverse impacts. By example, the `SupportsTemporaryTable` seems gone in Java, while in NHibernate it enables updates and deletes on classes having joined table (subclasses notably). This is why I have limited the scope of this Jira to purely informational properties.

### DIFF
--- a/src/NHibernate.Test/SqlTest/Custom/MySQL/MySQLTest.cs
+++ b/src/NHibernate.Test/SqlTest/Custom/MySQL/MySQLTest.cs
@@ -14,7 +14,7 @@ namespace NHibernate.Test.SqlTest.Custom.MySQL
 
 		protected override bool AppliesTo(Dialect.Dialect dialect)
 		{
-			return dialect is MySQL5Dialect || dialect is MySQLDialect;
+			return dialect is MySQLDialect;
 		}
 	}
 }
--- a/src/NHibernate/Dialect/DB2Dialect.cs
+++ b/src/NHibernate/Dialect/DB2Dialect.cs
@@ -273,30 +273,15 @@ namespace NHibernate.Dialect
 
 		#region Overridden informational metadata
 
-		public override bool SupportsEmptyInList
-		{
-			get { return false; }
-		}
+		public override bool SupportsEmptyInList => false;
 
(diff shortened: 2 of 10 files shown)


src/NHibernate.Test/SqlTest/Custom/MySQL/MySQLTest.cs
```diff
@@ -14,7 +14,7 @@ protected override IList Mappings
 
 		protected override bool AppliesTo(Dialect.Dialect dialect)
 		{
-			return dialect is MySQL5Dialect || dialect is MySQLDialect;
+			return dialect is MySQLDialect;
```
fredericDelaporte: MySQL5Dialect inherits MySQLDialect, so I have just simplified this.

src/NHibernate/Dialect/DB2Dialect.cs
```diff
@@ -270,5 +270,34 @@ public override string ForUpdateString
 		{
 			get { return " for read only with rs"; }
 		}
+
+		#region Overridden informational metadata
+
+		public override bool SupportsEmptyInList
+		{
+			get { return false; }
```
hazzik: Can these one-line properties be written with expression syntax? I know that it'll make an inconsistency, but I'm ok with it.

fredericDelaporte: Agreed and done.

PR #594, opened 2017-04-10
NH-3944 - preliminary work, Nh clauses eliminated.

fredericDelaporte:
As stated in @hazzik's PR in a [comment](https://github.com/nhibernate/nhibernate-core/pull/568#issuecomment-292796674), current implementation of Linq provider seems to diverge from Relinq expected use cases.

The `NhHavingClause`, `NhWithClause` and `NhJoinClause` added to the Relinq model does not match Linq semantic. Moreover, Relinq v2 make it quite harder to add such clauses.

This PR removes these clauses. Contrary to the linked comment expected resolution, I have not moved rewriters using those clauses to the last step. I have instead enriched the `VisitorParameters` class for allowing tracking Linq clauses requiring a different Hql translation than the default.

This has far less impacts.

This may have a weakness, as stated in a newly added comment in `PagingRewriter`: if a `Clone` operation is done on the model after `VisitorParameters` has been fetched, we may no more be able to match those clauses requiring a different Hql translation. Currently, only `PagingRewriter` does such a `Clone`, and in a case not supporting the above clauses, so this does not cause any issue.


src/NHibernate/Linq/GroupBy/PagingRewriter.cs
```diff
@@ -46,9 +45,13 @@ private static void FlattenSubQuery(SubQueryExpression subQueryExpression, Query
 			{
 				var cro = new ContainsResultOperator(new QuerySourceReferenceExpression(subQueryMainFromClause));
 
+				// Cloning may cause having/join/with clauses listed in VisitorParameters to no more be matched.
+				// Not a problem for now, because those clauses imply a projection, which is not supported
+				// by the "new WhereClause(new SubQueryExpression(newSubQueryModel))" below. See
+				// SingleKeyGroupAndCountWithHavingClausePagingAndOuterWhere test by example.
```
fredericDelaporte: This assertion was tested by adding following test to `NHibernate.Test.Linq.ByMethod.GroupByHavingTests`:

```C#
[Test, Explicit("Demonstrate an unsupported case for PagingRewriter")]
public void SingleKeyGroupAndCountWithHavingClausePagingAndOuterWhere()
{
	var orderCounts = db.Orders
		.GroupBy(o => o.Customer.CompanyName)
		.Where(g => g.Count() > 10)
		.Select(g => new { CompanyName = g.Key, OrderCount = g.Count() })
		.OrderBy(oc => oc.CompanyName)
		.Skip(5)
		.Take(10)
		.Where(oc => oc.CompanyName.Contains("F"))
		.ToList();

	Assert.That(orderCounts, Has.Count.EqualTo(3));
	var frankRow = orderCounts.Single(row => row.CompanyName == "Frankenversand");
	Assert.That(frankRow.OrderCount, Is.EqualTo(15));
}
```
The `WhereClause` line yields an `ArgumentException` with message

```
The items of the input sequence of type '<>f__AnonymousType6`2[System.String,System.Int32]' are not
compatible with the item expression of type
'System.Linq.IGrouping`2[System.String,NHibernate.DomainModel.Northwind.Entities.Order]'.
```

Which lets me think this flattening process does not support projection (the `Select(g =>`), which are needed for introducing having or left join clauses.

I have included that test case in this PR just now.

src/NHibernate/Linq/NestedSelects/NestedSelectRewriter.cs
```diff
@@ -59,98 +59,104 @@ public static void ReWrite(QueryModel queryModel, ISessionFactory sessionFactory
...
+				subQueryMainFromClause.ItemName,
+				subQueryMainFromClause.ItemType,
+				subQueryMainFromClause.FromExpression);
 
-			queryModel.BodyClauses.Add(@join);
+			parameters.AddLeftJoin(join, restrictions);
+			queryModel.BodyClauses.Add(join);
+			foreach (var withClause in restrictions)
+			{
+				queryModel.BodyClauses.Add(withClause);
```
fredericDelaporte: Adding them to the query model allows them to get transformed by any later `TransformExpressions`. This has no other usage currently.

They will not be added as HQL where because the query model visitor now check with visitor parameters each where clause for not being known as a with clause.

src/NHibernate/Linq/Visitors/LeftJoinRewriter.cs
```diff
@@ -28,26 +34,21 @@ public override void VisitAdditionalFromClause(AdditionalFromClause fromClause,
...
-			var restrictions = subQueryModel.BodyClauses
-				.OfType<WhereClause>()
-				.Select(w => new NhWithClause(w.Predicate));
-
-			var join = new NhJoinClause(mainFromClause.ItemName,
-										mainFromClause.ItemType,
-										mainFromClause.FromExpression,
-										restrictions);
+			var join = new AdditionalFromClause(mainFromClause.ItemName, mainFromClause.ItemType, mainFromClause.FromExpression);
+			var restrictions = subQueryModel.BodyClauses.OfType<WhereClause>().ToList();
```
fredericDelaporte: No cloning here. Reusing directly those from the flattened sub-query.

src/NHibernate/Linq/NestedSelects/NestedSelectRewriter.cs
```diff
@@ -59,98 +59,104 @@ public static void ReWrite(QueryModel queryModel, ISessionFactory sessionFactory
...
 
-			var restrictions = subQueryModel.BodyClauses
-											.OfType<WhereClause>()
-											.Select(w => new NhWithClause(w.Predicate));
-
-			var join = new NhJoinClause(subQueryMainFromClause.ItemName,
-										subQueryMainFromClause.ItemType,
-										subQueryMainFromClause.FromExpression,
-										restrictions);
+			var restrictions = subQueryModel.BodyClauses.OfType<WhereClause>().ToList();
```
fredericDelaporte: No cloning here. Reusing directly those from the flattened sub-query.

src/NHibernate/Linq/Visitors/LeftJoinRewriter.cs
```diff
@@ -28,26 +34,21 @@ public override void VisitAdditionalFromClause(AdditionalFromClause fromClause,
...
 			var innerSelectorMapping = new QuerySourceMapping();
 			innerSelectorMapping.AddMapping(fromClause, subQueryModel.SelectClause.Selector);
 
 			queryModel.TransformExpressions(ex => ReferenceReplacingExpressionTreeVisitor.ReplaceClauseReferences(ex, innerSelectorMapping, false));
 
 			queryModel.BodyClauses.RemoveAt(index);
-			queryModel.BodyClauses.Insert(index, @join);
-			InsertBodyClauses(subQueryModel.BodyClauses.Where(b => !(b is WhereClause)), queryModel, index + 1);
+			queryModel.BodyClauses.Insert(index, join);
+			InsertBodyClauses(subQueryModel.BodyClauses, queryModel, index + 1);
```
fredericDelaporte: No more filtering of where clauses: we need them in the model for ensuring expressions transformations are applied to them too.
If we were cloning them for building the restrictions list, then we would have to filter them out here then insert their clones.

They will not be added as HQL where because the query model visitor now check with visitor parameters each where clause for not being known as a with clause.

src/NHibernate/Linq/GroupBy/AggregatingGroupByRewriter.cs
```diff
@@ -81,11 +77,9 @@ private static void FlattenSubQuery(QueryModel queryModel, QueryModel subQueryMo
 				clause.TransformExpressions(s => GroupBySelectClauseRewriter.ReWrite(s, groupBy, subQueryModel));
 
 				//all outer where clauses actually are having clauses
-				var whereClause = clause as WhereClause;
-				if (whereClause != null)
+				if (clause is WhereClause whereClause)
 				{
-					queryModel.BodyClauses.RemoveAt(i);
-					queryModel.BodyClauses.Insert(i, new NhHavingClause(whereClause.Predicate));
+					parameters.AddHavingClause(whereClause);
```
fredericDelaporte: Juste letting them in the model for ensuring any later expression transformation get applied to them.

They will not be added as HQL where because the query model visitor now check each where clause with the visitor parameters. If they are known as having clauses, they will be added as HQL having clauses instead of  where.

src/NHibernate/Linq/Visitors/LeftJoinRewriter.cs
```diff
@@ -57,14 +58,14 @@ private static void InsertBodyClauses(IEnumerable<IBodyClause> bodyClauses, Quer
 			foreach (var bodyClause in bodyClauses)
 			{
 				destinationQueryModel.BodyClauses.Insert(destinationIndex, bodyClause);
-				++destinationIndex;
+				destinationIndex++;
```
fredericDelaporte: Readability. I have seen some debate about performances of the former vs the later. Not only I do know believe it to be relevant for .Net (as Alexander it seems), but also I consider that such tiny optimization should by default be discarded in favor of readability. (Only in very specific cases where it can actually have a human measurable impact would I left such a thing.)

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -162,56 +156,39 @@ public override void VisitMainFromClause(MainFromClause fromClause, QueryModel q
...
 		{
-			var querySourceName = VisitorParameters.QuerySourceNamer.GetName(fromClause);
-
-			var joinClause = fromClause as NhJoinClause;
-			if (joinClause != null)
-			{
-				VisitNhJoinClause(querySourceName, joinClause);
-			}
-			else if (fromClause.FromExpression is MemberExpression)
+			if (fromClause.FromExpression is MemberExpression)
```
fredericDelaporte: Here, I have to admit I have left the test cases check for me that apparently, `NhJoinClause` were never used with something not castable to `MemberExpression`. Otherwise, some cases previously handled with `NhJoinClause` but not using a `MemberExpression` will no more work.

It looks to me reasonable to do such an assumption since HQL permit joins only through member references as far as I know.

fredericDelaporte:

fredericDelaporte:
Choosing #568 instead.

PR #595, opened 2017-04-10
NH-3981 - maps equality fix, collection helper cleanup.

fredericDelaporte:
[NH-3981](https://nhibernate.jira.com/browse/NH-3981) - maps equality fix, collection helper cleanup.
It is a kind of follow up to #545. For cleanup I would have ask to use the `CollectionHelper.DictionaryEquals` method, but I discovered it was flawed.

And I would have asked to put a bag equality method there for avoiding duplicating between bag and idbag. For completeness, I have already added it in this class.

As the name `CollectionEquals` was ambiguous, I have renamed it and re-added as obsolete. A bunch of dead methods are now marked obsolete too.

In this same class, a usage of `StringComparer.InvariantCultureIgnoreCase` instead of `StringComparer.OrdinalIgnoreCase` for avoiding the [Turkish i trouble](http://haacked.com/archive/2012/07/05/turkish-i-problem-and-why-you-should-care.aspx/) as stated in the xml comment has been seen and fixed by the way. (Invariant is English culture, which does not expose to the trouble, but in that case we really just want to go ordinal.)

### DIFF
--- a/src/NHibernate/Util/CollectionHelper.cs
+++ b/src/NHibernate/Util/CollectionHelper.cs
@@ -1,7 +1,7 @@
 using System;
 using System.Collections;
-using Iesi.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace NHibernate.Util
 {
@@ -132,6 +132,7 @@ namespace NHibernate.Util
 			}
 		}
 
+		// To be removed in v6.0
 		[Serializable]
 		private class EmptyListClass : IList
 		{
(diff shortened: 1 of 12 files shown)


hazzik:
Can we either make these new methods internal, or, add overloads with `IEqualityComparer<>`(s)

src/NHibernate/Util/CollectionHelper.cs
```diff
@@ -633,7 +604,30 @@ public static bool SetEquals<T>(ISet<T> a, ISet<T> b)
...
+		/// <param name="c1">The first collection.</param>
+		/// <param name="c2">The second collection.</param>
+		/// <returns><c>true</c> if collection are equals, <c>false</c> otherwise.</returns>
+		public static bool BagEquals<T>(ICollection<T> c1, ICollection<T> c2)
+		{
+			var result = FastCheckEquality(c1, c2);
+			if (result.HasValue)
+				return result.Value;
+			var secondCardinalities = c2.GroupBy(e => e).ToDictionary(g => g.Key, g => g.Count());
+			return c1.GroupBy(e => e).All(g => secondCardinalities.TryGetValue(g.Key, out var secondCount) && g.Count() == secondCount);
```
hazzik: I believe following is more performant:
```csharp
var l2 = c2.ToLookup(e => e);
return c1.ToLookup(e => e).All(g => g.Count() == l2[g.Key].Count());
```

fredericDelaporte: Done. And used in bag and idbag method `AreCollectionElementsEqual` from NH-3932.

fredericDelaporte:
Overloads added. Argument types now wider: `IEnumerable` instead of `ICollection`. It simplifies usage for bag and idbag by example. And switching `Count` for Linq `Count()` should not cost must, Linq inspect the underlying type for using its length or count property if any.

fredericDelaporte:
Relaunched the failed build without changes, now it has passed.

PR #596, opened 2017-04-11
NH-3932 - follow up for replacing an O(n²) algorithm by an O(n) one.

fredericDelaporte:
[NH-3932](https://nhibernate.jira.com/browse/NH-3932) - merge firing unneeded updates, follow up for replacing an O(n²) algorithm by an O(n) one.

### DIFF
--- a/src/NHibernate/Type/GenericIdentifierBagType.cs
+++ b/src/NHibernate/Type/GenericIdentifierBagType.cs
@@ -137,10 +137,11 @@ namespace NHibernate.Type
 			}
 			else
 			{
+				var originalLookup = iterOriginal.ToLookup(e => e);
 				for (var i = 0; i < targetPc.Count; i++)
 				{
 					var currTarget = targetPc[i];
-					var orgToUse = iterOriginal.First(x => currTarget.Equals(x));
+					var orgToUse = originalLookup[currTarget].First();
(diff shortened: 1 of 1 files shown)


src/NHibernate/Type/GenericIdentifierBagType.cs
```diff
@@ -137,10 +137,11 @@ protected override bool AreCollectionElementsEqual(IEnumerable original, IEnumer
 			}
 			else
 			{
+				var originalLookup = iterOriginal.ToLookup(e => e);
 				for (var i = 0; i < targetPc.Count; i++)
 				{
 					var currTarget = targetPc[i];
-					var orgToUse = iterOriginal.First(x => currTarget.Equals(x));
+					var orgToUse = originalLookup[currTarget].First();
```
fredericDelaporte: Moreover, this has the advantage of using the default equality comparer for the type, instead of using the base object `Equals`.

I have thought about no taking first but building a map to an original equals elements stack, in order to ensure each original element gets its chance at replacing target element with its own state, since equals original elements may have different states.

Prior to NH-3932, merge would have done that. But well, since bags are un-ordered, which element is taken last is unspecified (original may not be a NHibernate collection, so we cannot tell it will be our internal list ordering). So the resulting state in case equals elements were merge with different states is anyway unpredictable. So I concluded that trying to preserve that behavior was not worth it.

PR #597, opened 2017-04-12
NH-3693 - AliasToBeanResultTransformer fails under Firebird

fredericDelaporte:
[NH-3693](https://nhibernate.jira.com/browse/NH-3693) - AliasToBeanResultTransformer fails under Firebird, fix and test cases

This is based on #330, rebased, squashed in one commit for original work of @amroel, another one for the "done at the same time" fix and test cases for [NH-1904](https://nhibernate.jira.com/browse/NH-1904) (trouble with struct having properties with same name but different case), and a last one for my proposed implementation of setter resolver (with test case for the ambiguous match case).


src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
```diff
@@ -0,0 +1,197 @@
...
+		/// <param name="resultObj">The object on which to set the property or field. It must be of the type for which
+		/// this instance has been built.</param>
+		/// <exception cref="PropertyNotFoundException">Thrown if no matching property or field can be found.</exception>
+		/// <exception cref="AmbiguousMatchException">Thrown if many matching properties or fields are found, having the
+		/// same visibility and inheritance depth.</exception>
+		public void SetProperty(string alias, object value, object resultObj)
+		{
+			if (_propertiesByNameCaseSensitive.TryGetValue(alias, out var property))
+			{
+				checkMember(property);
```
fredericDelaporte: On case sensitive case, we should never have an ambiguous case, but better be on the safe side.

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
```diff
@@ -0,0 +1,197 @@
...
+			while (currentType != null && currentType != typeof(object))
+			{
+				fields.AddRange(
+					currentType
+						.GetFields(bindingFlags)
+						.Select(f => new RankedMember<FieldInfo> { Member = f, VisibilityRank = getFieldVisibilityRank(f), HierarchyRank = rank }));
+				properties.AddRange(
+					currentType
+						.GetProperties(bindingFlags)
+						.Where(p => p.CanWrite)
```
hazzik: Need to check [`p.GetIndexParameters().Length ==0`](https://msdn.microsoft.com/en-us/library/system.reflection.propertyinfo.getindexparameters.aspx)

fredericDelaporte: Yes. I have done some cleanup by the way.

src/NHibernate.Test/TransformTests/AliasToBeanResultTransformerFixture.cs
```diff
@@ -73,7 +117,100 @@ public void WorkWithOutPublicParameterLessCtor()
...
+				Setup();
+
+				using (ISession s = OpenSession())
+				{
+					var transformer = Transformers.AliasToBean<PublicInheritedPropertiesSimpleDTO>();
+					IList<PublicInheritedPropertiesSimpleDTO> l = s.CreateSQLQuery("select * from Simple")
+						.SetResultTransformer(transformer)
+						.List<PublicInheritedPropertiesSimpleDTO>();
+					Assert.That(l.Count, Is.EqualTo(2));
+					Assert.That(l, Has.All.Not.Null);
```
hazzik: Can you also check that all properties get the desired values?

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
```diff
@@ -0,0 +1,196 @@
...
+
+			void checkMember<T>(NamedMember<T> member) where T : MemberInfo
+			{
+				if (member.Member != null)
+					return;
+
+				if (member.AmbiguousMembers == null || member.AmbiguousMembers.Length < 2)
+					throw new InvalidOperationException($"{nameof(NamedMember<T>.Member)} missing and {nameof(NamedMember<T>.AmbiguousMembers)} invalid.");
+
+				throw new AmbiguousMatchException(
```
hazzik: I think we need tests for these exceptional cases

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
```diff
@@ -0,0 +1,196 @@
...
+				return 5;
+			}
+
+			Dictionary<string, NamedMember<T>> getMapByName<T>(IEnumerable<RankedMember<T>> members, StringComparer comparer) where T : MemberInfo
+			{
+				return members
+					.GroupBy(m => m.Member.Name,
+						(k, g) =>
+							new NamedMember<T>(k,
+								g
```
hazzik: Need to have a test for this behaviour

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
fredericDelaporte: What about the `ToPropertiesInsensitivelyDuplicated_WithoutAnyProjections` test I have added with the third commit?

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
fredericDelaporte: Ok, pushed, tests heavily rewritten.

fredericDelaporte:
Attempt to fix the new cases for the 3 failing DB. Guessing they yield full lower case aliases instead of uppercase as I was believing.
And added some more xml doc comments.

hazzik:
@fredericDelaporte different tests fail on different DBs

hazzik:
It seems that Oracle and Firebird return upper case, and PostgreSQL - lowercase

fredericDelaporte:
Yes I have seen there are troubles, I have already made some attempts but it looks like I will have to debug them with local installs.

fredericDelaporte:
Two unrelated files were added in my previous commit amendment, I have removed them.

Unless told otherwise, I intend to merge this PR in the days to come then close #330.

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
```diff
@@ -0,0 +1,196 @@
...
+
+namespace NHibernate.Transform
+{
+	/// <summary>
+	/// Resolves setter for an alias with a heuristic: search among properties then fields for matching name and case, then,
+	/// if no matching property or field was found, retry with a case insensitive match. For members having the same name, it
+	/// sorts them by inheritance depth then by visibility from public to private, and takes those ranking first.
+	/// </summary>
+	[Serializable]
+	public class QueryAliasToObjectPropertySetter
```
hazzik: Can this please be internal?

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
```diff
@@ -0,0 +1,196 @@
...
+			}
+
+			var fieldsByNameCaseSensitive = getMapByName(fields, StringComparer.Ordinal);
+			var fieldsByNameCaseInsensitive = getMapByName(fields, StringComparer.OrdinalIgnoreCase);
+			var propertiesByNameCaseSensitive = getMapByName(properties, StringComparer.Ordinal);
+			var propertiesByNameCaseInsensitive = getMapByName(properties, StringComparer.OrdinalIgnoreCase);
+
+			return new QueryAliasToObjectPropertySetter(fieldsByNameCaseSensitive, fieldsByNameCaseInsensitive, propertiesByNameCaseSensitive, propertiesByNameCaseInsensitive);
+
+			int getFieldVisibilityRank(FieldInfo field)
```
hazzik: ReSharper suggests that these shall be named as usual method.

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
```diff
@@ -0,0 +1,196 @@
...
+		/// <returns>A <c>QueryAliasToObjectPropertySetter</c>.</returns>
+		public static QueryAliasToObjectPropertySetter MakeFor(System.Type objType)
+		{
+			var bindingFlags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly;
+			var fields = new List<RankedMember<FieldInfo>>();
+			var properties = new List<RankedMember<PropertyInfo>>();
+			var currentType = objType;
+			var rank = 1;
+			// For grasping private members, we need to manually walk the hierarchy.
+			while (currentType != null && currentType != typeof(object))
```
hazzik: Can you move the while loop to a separate method?

hazzik:
To be honest I'm not a big fun of local functions, and would like to avoid them until completely necessary.

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
```diff
@@ -0,0 +1,196 @@
...
+using System.Reflection;
+
+namespace NHibernate.Transform
+{
+	/// <summary>
+	/// Resolves setter for an alias with a heuristic: search among properties then fields for matching name and case, then,
+	/// if no matching property or field was found, retry with a case insensitive match. For members having the same name, it
+	/// sorts them by inheritance depth then by visibility from public to private, and takes those ranking first.
+	/// </summary>
+	[Serializable]
```
hazzik: If this class is Serializable we need to implement `IDeserializationCallback` because we have `IDictionary<,>` fields. Or we need to exclude these fields non serialized and add a deserialization constructor for a class to repopulate these fields

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
hazzik: Or, actually inlined (not just make a nested class, but actually inlined) into the `AliasToBeanTransformer`

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
fredericDelaporte: PascalCase? Microsoft seems to mandate camelCase instead, for exhibiting their local nature.

fredericDelaporte:
> To be honest I'm not a big fun of local functions, and would like to avoid them until necessary.

I find them handy for avoiding bloating a class with helper functions useful only for a single method. But well, we have cope without them for many years and can go on without them.

src/NHibernate/Transform/QueryAliasToObjectPropertySetter.cs
fredericDelaporte: Maybe better simply store the concrete implementation which is serializable.

fredericDelaporte:
Bunch of changes done.

src/NHibernate/Transform/AliasToBeanResultTransformer.cs
```diff
@@ -29,37 +31,46 @@ namespace NHibernate.Transform
 	/// sorts them by inheritance depth then by visibility from public to private, and takes those ranking first.
 	/// </remarks>
 	[Serializable]
-	public class AliasToBeanResultTransformer : AliasedTupleSubsetResultTransformer
+	public class AliasToBeanResultTransformer : AliasedTupleSubsetResultTransformer, IEquatable<AliasToBeanResultTransformer>
```
hazzik: [Still need](http://stackoverflow.com/a/25583618/259946) `IDeserializationCallback`

fredericDelaporte: I will add a serialization/deserialization test. Your link on SO seems unrelated: the trouble was attempting to use the map during the deserialization process. The OP states that the map was correclty filled after, but he needed it during.

fredericDelaporte:
Serialization/deserialization test added, no troubles.

Now if we consider serialization should be spatially optimized (consumed memory) rather than temporally (execution time), flagging dictionaries as non serializable and reconstructing them after deserialization would be better. (I do not tend to bother with those considerations when I do not actually understand the serialization use case for the application: hard to know which optimization to favor in such case.)

fredericDelaporte:
Rebased and last three commits squashed in one.

fredericDelaporte:
Thanks for the reviews.

PR #598, opened 2017-04-13
NH-3984 - MsSql - fixed reading milliseconds from datetime2 column via CreateSQLQuery()

Ralf1108:
"GetFieldType()" returns for a "datetime2" column to  "Datetime" which resolves to the "DateTimeType" from NHibernate which cuts the milliseconds.

If we use "GetDataTypeName()" for getting column type we get "datetime2" which then resolves correctly to "DateTime2Type".

So I think using "GetFieldType()" should only be used as a fallback if resolving via "GetDataTypeName()" didn't return a result.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/EntityWithDatetime2/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/EntityWithDatetime2/Fixture.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using System.Linq;
+using NHibernate.Linq;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.EntityWithDatetime2
+{
+    [TestFixture]
+    public class Fixture : BugTestCase
+    {
+        protected override void OnTearDown()
+        {
+            using (var s = OpenSession())
+            using (var tx = s.BeginTransaction())
+            {
+                s.CreateSQLQuery("delete from LogEntry").ExecuteUpdate();
+                tx.Commit();
+            }
+        }
+
+        [Test]
+        public void ShouldBePossibleToReadMillisecondsFromDatetime2ViaCreateSQLQuery()
+        {
+            var now = DateTime.UtcNow;
+
+            using (var session = OpenSession())
+            using (var transaction = session.BeginTransaction())
+            {
+                var entry = new LogEntry {Text = "Test", CreatedAt = now};
+                session.Save(entry);
+
+                session.Flush();
+                transaction.Commit();
+            }
+
+            using (var session = OpenSession())
+            {
+                var datetimeViaQuery = session
+                    .Query<LogEntry>()
+                    .Select(x => x.CreatedAt)
+                    .Single();
+
+                Assert.That(datetimeViaQuery.Millisecond, Is.EqualTo(now.Millisecond));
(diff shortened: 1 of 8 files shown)


fredericDelaporte:
Hello,

Your changes add some white-spaces inconsistencies in the formatting: NHibernate sources use tabs, not spaces. We also need a [Jira](https://nhibernate.jira.com) ticket for tracking changes. Please check [contributing guidelines](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md) for more details.

Thanks.

Ralf1108:
here is the Jira ticket
https://nhibernate.jira.com/browse/NH-3984

Ralf1108:
TeamCity build fails indicate that for other database dialect "datetime2 is not supported which makes sense.

So the unittest for the bug has to be moved to SqlServer test suite only

Ralf1108:
hmm.. where should the database specific tests go?

Ralf1108:
hope that s how to run test only for MsSql2008

fredericDelaporte:
Yes, this is the right way.

hazzik:
It still needs to use the right indents (tabs)

Since the beginning of time Hibernate is using the following implementation:
```
		public Type getHibernateType(int columnPos) throws SQLException {
				int columnType = resultSetMetaData.getColumnType( columnPos );
				int scale = resultSetMetaData.getScale( columnPos );
				int precision = resultSetMetaData.getPrecision( columnPos );
				return TypeFactory.heuristicType(
						factory.getDialect().getHibernateTypeName(
								columnType,
								precision,
								precision,
								scale
						)
				);
			}
```

Now it looks like this: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/loader/custom/JdbcResultMetadata.java#L66

Ralf1108:
hope I got indendation now correct.

@hazzik your remark means that my fix is wrong or would you prefer to implement it in another way. Hibernate solution looks like the active database dialect has a chance to influence the result type.

fredericDelaporte:
LogEntry and Fixture are still using spaces instead of tabs.

Ralf1108:
didn't thought about the test files, sry :-(

hazzik:
>@hazzik your remark means that my fix is wrong or would you prefer to implement it in another way. 

@Ralf1108 I agree that the bug shall be fixed, but the thing is that your fix just accidentally works in this case, and does not work in other cases.

Ralf1108:
As NHibernate is rather mature and there are no other bug reports for such cases maybe those other hypothetical cases dont exist yet :-)

So for now stay with the fix until there is better one? (principles: YAGNI or KISS)

fredericDelaporte:
The better could already be there: the Hibernate change should be analyzed, and ported to NHibernate if it looks better.

fredericDelaporte:
We are now heading toward some significant changes in date time types, see [NH-3919](https://nhibernate.jira.com/browse/NH-3919). This should cause this issue to become obsolete.

hazzik:
@fredericDelaporte can you merge tests from here to #703 ?

fredericDelaporte:
Done.

hazzik:
Replaced by #703

PR #599, opened 2017-04-15
NH-3944 - Relinq v2 migration

fredericDelaporte:
Based on #594, here is the corresponding migration to Relinq v2.

It differs on some more points with #568, I am going to comment them with a review.


src/NHibernate/Linq/Visitors/NhExpressionVisitor.cs
```diff
@@ -0,0 +1,70 @@
...
+				return null;
+			}
+
+			// Keep this variable for easy examination during debug.
+			var expr = base.Visit(expression);
+			return expr;
+		}
+
+		public virtual Expression VisitNhStar(NhStarExpression expression)
+			=> VisitExtension(expression);
```
fredericDelaporte: With [Expression Trees v2 Spec](http://www.codeplex.com/Download?ProjectName=dlr&DownloadId=246540), those methods are now called by `Expression.Accept`. The old default implementation (calling `Accept` on the expression) would cause an infinite calls loop. Instead, the default implementation should be now to call `VisitExtension`.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using System.Linq.Expressions;
+using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Parsing;
+using Remotion.Linq.Parsing.ExpressionVisitors;
+using Remotion.Linq.Parsing.ExpressionVisitors.TreeEvaluation;
+
+namespace NHibernate.Linq.Visitors
+{
+	internal class NhPartialEvaluatingExpressionVisitor : RelinqExpressionVisitor
```
fredericDelaporte: Does no more implements `IPartialEvaluationExceptionExpressionVisitor`. It was needed for supporting failed attempts at partially evaluating expressions on methods to be translated by the Linq provider. With the `NhEvaluatableExpressionFilter`, we can now avoid those failures by marking methods which should not be evaluated.

src/NHibernate/Linq/Visitors/WhereJoinDetector.cs
```diff
@@ -119,174 +119,164 @@ public override Expression VisitExpression(Expression expression)
...
-			{
-				HandleBinaryOperation((a, b) => a.GreaterThan(b));
-			}
-			else if (expression.NodeType == ExpressionType.LessThan)
-			{
-				HandleBinaryOperation((a, b) => a.LessThan(b));
-			}
-			else if (expression.NodeType == ExpressionType.LessThanOrEqual)
-			{
-				HandleBinaryOperation((a, b) => a.LessThanOrEqual(b));
```
fredericDelaporte: Huge pile of `else if` instead of a switch case. Not really related to the migration, but well, with the amount of other changes on those classes, I have decide to refactor such things by the way.

fredericDelaporte:
For reference, here is the [page linking to ET v2 reference documentation](http://dlr.codeplex.com/wikipage?title=Docs%20and%20specs&referringTitle=Documentation).

fredericDelaporte:
Now I have done this migration myself, I am no more convinced NHibernate implementation was really deviating from "Relinq good practices" as I have stated [here](https://github.com/nhibernate/nhibernate-core/pull/568#issuecomment-292796674).

It seems to me extending `Extension` with our own is still supported by Relinq, and indeed this PR migration still does that for a bunch of clauses like aggregates. But Relinq does no more allow extending further its own extensions, and I find it a bit unfortunate, even maybe a bit short-sighted. That forces us either to duplicate similar behavior for NHibernate left joins or joins with restrictions, or to "externalize" some data as I have done for handling our additional semantics.

Have a look at Relinq `NullEvaluatableExpressionFilter`: it has been created internal, forcing anyone willing to use directly a `PartialEvaluatingExpressionVisitor` without any filter to re-implement it... I understand internal detail should stay internal, but here I feel like it is going too far and just hinders users pointlessly.

hazzik:
So, what implementation shall we choose?

fredericDelaporte:
Well, this one match a bit closer ET v2 spec guidelines, but this is not related to the having/with/join clauses removal. We could do the same with #568 (and probably should if you want to choose it).

The removal of having/with/join clauses seems to simplify a bit the implementation of Relinq migration, but I find the result to be maybe less robust. (Due to having to lookup expressions in an additional data map for finding out they are a bit more than what they are by themselves.)

I think I need to get a closer look to #568 to see if we can simplify it or if we would have better to use #599 instead.

fredericDelaporte:
I have looked again to all that, and I have still mixed feelings about both PR.

Extending expressions is well supported by Relinq, but extending clauses is a bit less. With Relinq v2 migration, as soon as NH own clauses from #568 are added in the model, only visitors aware of them could be used. Currently that does not cause issues, but maybe later could it turn out problematic.

On the other hand, #599 flagging clauses as having additional meaning and matching them later will breaks in case of cloning operations.

I think #599 relative simplicity will allow it to better follow future Relinq changes, compared to #568. So I would vote for choosing #599 rather than #568.

fredericDelaporte:

fredericDelaporte:
Choosing #568 instead.

PR #600, opened 2017-04-16
NH-3985 - Bugfix for subsequent child sessions being returned in disposed state

alobakov:
[NH-3985](https://nhibernate.jira.com/browse/NH-3985) - Test case and fix for subsequent child sessions being returned in disposed state

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3985/Entity.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3985/Entity.cs
@@ -1,0 +1,10 @@
+﻿using System;
+
+namespace NHibernate.Test.NHSpecificTest.NH3985
+{
+	public partial class Process
+	{
+		public virtual Guid ProcessID { get; set; }
+		public virtual string Name { get; set; }
+	}
+}
--- a/src/NHibernate.Test/NHSpecificTest/NH3985/Mappings.hbm.xml
+++ b/src/NHibernate.Test/NHSpecificTest/NH3985/Mappings.hbm.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<hibernate-mapping xmlns="urn:nhibernate-mapping-2.2" assembly="NHibernate.Test" namespace="NHibernate.Test.NHSpecificTest.NH3985">
+	<class name="Process">
+		<id name="ProcessID" generator="guid.comb" />
+		<property name="Name" />
+	</class>
+</hibernate-mapping>
--- a/src/NHibernate.Test/NHibernate.Test.csproj
+++ b/src/NHibernate.Test/NHibernate.Test.csproj
@@ -743,6 +743,8 @@
     <Compile Include="NHSpecificTest\EntityWithUserTypeCanHaveLinqGenerators\IExample.cs" />
     <Compile Include="Insertordering\NH3931Entities.cs" />
     <Compile Include="NHSpecificTest\NH1904\StructFixture.cs" />
+    <Compile Include="NHSpecificTest\NH3985\Entity.cs" />
+    <Compile Include="NHSpecificTest\NH3985\Fixture.cs" />
     <Compile Include="NHSpecificTest\NH3247\Entity.cs" />
     <Compile Include="NHSpecificTest\NH3247\Fixture.cs" />
     <Compile Include="NHSpecificTest\NH3386\Entity.cs" />
@@ -2668,6 +2670,7 @@
     <EmbeddedResource Include="NHSpecificTest\NH2195\Mappings.hbm.xml" />
     <EmbeddedResource Include="NHSpecificTest\NH3187\Mappings.hbm.xml" />
     <EmbeddedResource Include="NHSpecificTest\NH3932\Model\Mappings.hbm.xml" />
+    <EmbeddedResource Include="NHSpecificTest\NH3985\Mappings.hbm.xml" />
     <EmbeddedResource Include="TypedManyToOne\Customer.hbm.xml" />
   </ItemGroup>
   <ItemGroup>
--- a/src/NHibernate/Impl/SessionImpl.cs
+++ b/src/NHibernate/Impl/SessionImpl.cs
@@ -73,7 +73,7 @@ namespace NHibernate.Impl
 		private readonly StatefulPersistenceContext persistenceContext;
 
 		[NonSerialized]
-		private readonly ISession rootSession;
+		private readonly SessionImpl rootSession;
 
 		[NonSerialized]
 		ISession _childSession;
@@ -381,6 +381,7 @@ namespace NHibernate.Impl
 				{
 					SetClosed();
 					Cleanup();
+					if (rootSession != null) rootSession._childSession = null;
 				}
 			}
 		}
@@ -1681,6 +1682,7 @@ namespace NHibernate.Impl
 				// free unmanaged resources here
 
 				IsAlreadyDisposed = true;
+
 				// nothing for Finalizer to do - so tell the GC to ignore it
 				GC.SuppressFinalize(this);
 			}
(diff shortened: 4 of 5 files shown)


src/NHibernate.Test/NHSpecificTest/NH3985/Mappings.hbm.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<hibernate-mapping xmlns="urn:nhibernate-mapping-2.2" assembly="NHibernate.Test" namespace="NHibernate.Test.NHSpecificTest.NH3985">
+	<class name="Process" table="Processes">
+		<id name="ProcessID" type="Int32">
+			<column name="ProcessID" not-null="true" sql-type="INT" />
+			<generator class="identity" />
+		</id>
+		<property name="Name" type="String">
+			<column name="Name" not-null="true" length="50" sql-type="NVARCHAR(50)" />
```
fredericDelaporte: Avoid specifying db specific types. Better not specify any type at all, NHibernate infers them.

src/NHibernate.Test/NHSpecificTest/NH3985/Entity.cs
```diff
@@ -0,0 +1,45 @@
...
+			if (Object.Equals(this.ProcessID, default(long)))
+				return base.GetHashCode();
+
+			hashCode = (hashCode * 7) + ProcessID.GetHashCode();
+			return hashCode;
+		}
+
+		#endregion
+
+		public virtual int ProcessID { get; set; }
```
fredericDelaporte: Oracle will use decimal instead for identities/sequences, so Oracle build may failed due to this.

src/NHibernate.Test/NHSpecificTest/NH3985/Mappings.hbm.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<hibernate-mapping xmlns="urn:nhibernate-mapping-2.2" assembly="NHibernate.Test" namespace="NHibernate.Test.NHSpecificTest.NH3985">
+	<class name="Process" table="Processes">
+		<id name="ProcessID" type="Int32">
+			<column name="ProcessID" not-null="true" sql-type="INT" />
+			<generator class="identity" />
```
fredericDelaporte: Identity with Oracle use decimal. Using `native` as generator would be safer by the way, current NHibernate Oracle dialect does not implement identity though Oracle supports it. I believe it does fallback to sequence nether-less, so will work like `native`, but at least for `native` that is its documented behavior.

You may also consider using an app side generator like `guid.comb`, that helps avoid db specific issues.

alobakov:
Doh... Man, thanks for all the reminders. That was a total brain fart on my part. :))
- Will fix.

alobakov:
Ok, fixed that. - Let me know if anything else doesn't look right.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -2221,7 +2221,10 @@ public ISession GetChildSession()
 
 				CheckAndUpdateSessionStatus();
 
-				if (_childSession == null)
+				var childImpl = _childSession as SessionImpl;
+
+				// if child session never existed or has already been disposed, create new
+				if (childImpl == null || childImpl.IsAlreadyDisposed)
```
hazzik: What if instead of this check we remove the reference on disposal?

alobakov: That's what I'm doing in the implementation of NH-3986 with a collection of child sessions, but that will come in as a separate pull request. If both end up included in the same release, this becomes moot point.

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -3248,6 +3250,9 @@
     <EmbeddedResource Include="NHSpecificTest\NH1291AnonExample\Mappings.hbm.xml" />
   </ItemGroup>
   <ItemGroup>
+    <EmbeddedResource Include="NHSpecificTest\NH3985\Mappings.hbm.xml">
+      <SubType>Designer</SubType>
```
alobakov: Yes, I didn't like that either, but Visual Studio seems to keep adding it. Not sure exactly what it means and why it gets added for some mapping files while not others.

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -3248,6 +3244,7 @@
     <EmbeddedResource Include="NHSpecificTest\NH1291AnonExample\Mappings.hbm.xml" />
   </ItemGroup>
   <ItemGroup>
+    <EmbeddedResource Include="NHSpecificTest\NH3985\Mappings.hbm.xml" />
```
alobakov: Removed this and other SubType elements, however it seems to have been an ongoing issue with Visual Studio since VS2005, so there's no clear way to eradicate them once and for all. They appear to be completely harmless though.

src/NHibernate.Test/NHibernate.Test.csproj
alobakov: Removed

src/NHibernate.Test/NHibernate.Test.csproj
hazzik: Thanks. 

These tags become harmful when you try to merge branches as they provide false anchors for the merge tools.

src/NHibernate/Impl/SessionImpl.cs
hazzik: There are 2 points to implement the correct logic here:

1. It can be back-ported (why? because this is actually a bug) to previous versions without a lot of work required

2. I'm not yet sure that the improvement will make it to 5.0

hazzik:
I modified your PR to remove child session from parent as soon as it gets disposed.

hazzik:
@alobakov next time please make sure that PR is made from a branch, not from master.

alobakov:
Ok will do. 
It did have its own branch in my fork, though there were no specific instructions in CONTRIBUTING.md

fredericDelaporte:
The [Commit your Test Case](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md#commit-your-test-case) section tells it. Maybe a bit late in this guide. Not sure it mandates a change.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -2604,5 +2608,10 @@ public override IEntityPersister GetEntityPersister(string entityName, object ob
 				}
 			}
 		}
+
+		void RemoveChildSession(ISession session)
+		{
+			Interlocked.CompareExchange(ref _childSession, null, session);
```
fredericDelaporte: Session is not thread safe, so we should not care about thread safety here.

I know there is a case where `Dispose` can be called from another thread: when a transaction in which the session was enlisted ends. But this case is buggy and doomed anyway: the child-session is already awaiting disposal (`TransactionContext.ShouldCloseSessionOnDistributedTransactionCompleted` being `true`) and should already be considered unusable for any other purpose than ending that transaction. So having it still returnable by its parent session is a trouble.

Being thread safe here will not help avoid troubles. Either we should ensure the child session is immediately removed from its parent on child dispose/close call, even when its disposal is postponed due to an ongoing transaction (`TransactionContext` not null), or we should cease postponing disposal when there is an ongoing transaction, as Hibernate do with `JdbcContext`. In both cases, we would no more be exposed to the transaction completion occurring on a distinct thread trouble.

fredericDelaporte: Just seeing know these [commit comments](https://github.com/nhibernate/nhibernate-core/commit/390011bca21bf90bb6912d62d97c229de3ec26cf#commitcomment-21874582) on the same subject.

In do not really get how another child-session could be there with current implementation. But well, if it is not about thread-safety but only for not removing another one, I would rather implement it without any threading helper, with a if or iif.

hazzik:
Changed as per review comments

hazzik:
Thanks everyone. Squashed and applied to master now.

PR #601, opened 2017-04-16
NH-3987 - Re-implement NhQueryable options

fredericDelaporte:
[NH-3987](https://nhibernate.jira.com/browse/NH-3987) Re-implement NhQueryable options.

At least that way, adding more options should be way less intrusive.

Instead of having one extension method per option, we now would have a single `SetOptions` extension method, taking a delegate for setting options with a `IQueryableOptions` object.

    query = query.SetOptions(o => o.SetTimeout(10).SetCacheable(true));

I could have directly exposed the `IQuery` instead but well, it would have exposed undesirable features in a Linq context I think.


hazzik:
So, what implementation shall we choose?

fredericDelaporte:
I do not propose other alternative, I think this one is still way better than previous, although it is still a node in the expression tree.

Now using a delegate as a quey parameter may cause some new 2nd level cache miss, and I think I can transparently replace it with the option object.

hazzik:
Ah, sorry, I've meant to ask this question on #599

fredericDelaporte:
The `SetOptions` parameter does not end up in cache parameters key, so it does not cause an issue. and keeping the delegate will allow more code simplification if we mutualise later the options interface on all query API.

fredericDelaporte:
Updated the reference documentation too. And fixed the `SetOptions` xml comment.

fredericDelaporte:

hazzik:
I was thinking. Do we need to have the separate method of options? Why not include it into `.Query(options)`?

fredericDelaporte:
Well I have tried at first something similar. But there is a roadblock: each Linq call creates a new `NhQueryable`, which only gets the expression and the provider. I have not found better ways for transmitting the options than keeping them in the expression.

Then, only allowing setting the options when obtaining the queryable from the session may be limiting: I frequently build different queries from the same queryable base on which I have already applied some Linq transformation, and those resulting different queries may not have the same options requirements.

fredericDelaporte:
Thanks Alexander,
I will merge that after some more PR on fixing tests.

PR #602, opened 2017-04-17
NH-3988 - Replace ApplicationException with Exception

hazzik:


src/NHibernate/Logging.cs
```diff
@@ -63,15 +63,15 @@ private static ILoggerFactory GetLoggerFactory(string nhibernateLoggerClass)
 			}
 			catch (MissingMethodException ex)
 			{
-				throw new ApplicationException("Public constructor was not found for " + loggerFactoryType, ex);
+				throw new Exception("Public constructor was not found for " + loggerFactoryType, ex);
```
fredericDelaporte: I would rather throw an ArgumentException here (or InvalidOperationException). We should not throw the base Exception class.

src/NHibernate/Logging.cs
```diff
@@ -63,15 +63,15 @@ private static ILoggerFactory GetLoggerFactory(string nhibernateLoggerClass)
...
 			}
 			catch (InvalidCastException ex)
 			{
-				throw new ApplicationException(loggerFactoryType + "Type does not implement " + typeof (ILoggerFactory), ex);
+				throw new Exception(loggerFactoryType + "Type does not implement " + typeof (ILoggerFactory), ex);
```
fredericDelaporte: Same as previous.

src/NHibernate/Logging.cs
```diff
@@ -63,15 +63,15 @@ private static ILoggerFactory GetLoggerFactory(string nhibernateLoggerClass)
...
 			}
 			catch (Exception ex)
 			{
-				throw new ApplicationException("Unable to instantiate: " + loggerFactoryType, ex);
+				throw new Exception("Unable to instantiate: " + loggerFactoryType, ex);
```
fredericDelaporte: Same as previous.

hazzik:
Replaced bare `Exception` with `InstantiationException`

PR #604, opened 2017-04-21
NH-3580 - Remove explicit dependency on Iesi.Collections

hazzik:
[NH-3580](https://nhibernate.jira.com/browse/NH-3580)

### DIFF
--- a/src/NHibernate/Type/GenericOrderedSetType.cs
+++ b/src/NHibernate/Type/GenericOrderedSetType.cs
@@ -12,7 +12,7 @@ namespace NHibernate.Type
 	public class GenericOrderedSetType<T> : GenericSetType<T>
 	{
 		// ReSharper disable once StaticMemberInGenericType
-		static ConstructorInfo _setConstructor;
+		static readonly Lazy<ConstructorInfo> SetConstructor = new Lazy<ConstructorInfo>(GetSetConstructor);
 
 		/// <summary>
 		/// Initializes a new instance of a <see cref="GenericOrderedSetType{T}"/> class for
@@ -25,24 +25,6 @@ namespace NHibernate.Type
 		public GenericOrderedSetType(string role, string propertyRef)
 			: base(role, propertyRef)
 		{
-		}
-
-		static ConstructorInfo SetConstructor
-		{
-			get
-			{
-				if (_setConstructor == null)
-				{
-					lock (typeof(GenericOrderedSetType<>))
-					{
-						if (_setConstructor == null)
-						{
-							_setConstructor = GetSetConstructor();
-						}
-					}
-				}
-				return _setConstructor;
-			}
 		}
 
 		static ConstructorInfo GetSetConstructor()
@@ -61,7 +43,7 @@ Please make sure that you have referenced the Iesi.Collections package.");
 
 		public override object Instantiate(int anticipatedSize)
 		{
-			return SetConstructor.Invoke(null);
+			return SetConstructor.Value.Invoke(null);
 		}
 	}
 }
--- a/src/NHibernate/Type/GenericOrderedSetType.cs
+++ b/src/NHibernate/Type/GenericOrderedSetType.cs
@@ -1,6 +1,6 @@
 using System;
 using System.Collections.Generic;
-using Iesi.Collections.Generic;
+using System.Reflection;
 
 namespace NHibernate.Type
 {
@@ -11,6 +11,9 @@ namespace NHibernate.Type
 	[Serializable]
 	public class GenericOrderedSetType<T> : GenericSetType<T>
 	{
+		// ReSharper disable once StaticMemberInGenericType
+		static ConstructorInfo _setConstructor;
+
 		/// <summary>
 		/// Initializes a new instance of a <see cref="GenericOrderedSetType{T}"/> class for
 		/// a specific role.
@@ -24,9 +27,41 @@ namespace NHibernate.Type
 		{
 		}
 
+		static ConstructorInfo SetConstructor
+		{
+			get
+			{
+				if (_setConstructor == null)
+				{
+					lock (typeof(GenericOrderedSetType<>))
+					{
+						if (_setConstructor == null)
+						{
+							_setConstructor = GetSetConstructor();
+						}
+					}
+				}
+				return _setConstructor;
+			}
+		}
+
+		static ConstructorInfo GetSetConstructor()
+		{
+			var type = System.Type.GetType("Iesi.Collections.Generic.LinkedHashSet`1, Iesi.Collections", false);
+			if (type == null)
+				throw new TypeLoadException(
+					@"Could not load Iesi.Collections.Generic.LinkedHashSet`1
+
+Please make sure that you have referenced the Iesi.Collections package.");
+
+			return type
+				.MakeGenericType(typeof(T))
+				.GetConstructor(System.Type.EmptyTypes);
+		}
+
 		public override object Instantiate(int anticipatedSize)
 		{
-			return new LinkedHashSet<T>();
+			return SetConstructor.Invoke(null);
 		}
 	}
 }
--- a/default.build
+++ b/default.build
@@ -188,7 +188,6 @@
 		</copy>
 		<copy todir="${bin-pack.required}">
 			<fileset basedir="${bin.dir}">
-				<include name="Iesi.Collections.???" />
 				<include name="NHibernate.???" />
 			</fileset>
 		</copy>
--- a/src/NHibernate.Test/NHSpecificTest/NH3221/Person.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3221/Person.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using Iesi.Collections.Generic;
 
 namespace NHibernate.Test.NHSpecificTest.NH3221
 {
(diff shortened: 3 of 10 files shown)


src/NHibernate/Type/GenericOrderedSetType.cs
```diff
@@ -24,9 +27,41 @@ public GenericOrderedSetType(string role, string propertyRef)
 		{
 		}
 
+		static ConstructorInfo SetConstructor
+		{
+			get
+			{
+				if (_setConstructor == null)
+				{
+					lock (typeof(GenericOrderedSetType<>))
```
fredericDelaporte: Lock should be put on privately held fields for avoiding deadlock in case some code elsewhere put a lock on the same public thing. (Msdn recommended practice, with some discussion about it [here](http://stackoverflow.com/q/1603008/1178314).) Illustrated in this [PR](https://github.com/nhibernate/nhibernate-core/pull/590/files) by the way.

fredericDelaporte: By the way, why not simply using a Lazy<ConstructorInfo> here?

hazzik:
Are there any users who do not add `Iesi.Collections`? No:) because it was a mandatory dependency, and people are using old NuGet's packages.config where the all nested dependency are listed.

hazzik:
Another possible solution to this issue is to inline the `LinkedHashSet<>` class into NHibernate (with all necessary type transfer attributes added). But I think it would cause more harm than the current PR, because it will make some people angry as they would need to reference the NHibernate into the domain model, which some adepts of persistence ignorance hate.

src/NHibernate/Type/GenericOrderedSetType.cs
hazzik: I don't think that anyone would make a lock on this type. Btw lock on a type is how `[MethodImpl(Synchronized)]` attribute is implemented. 

The reason for this field being lazy initialized because we cannot throw expetions from the static constructor - this will make the class unusable.

fredericDelaporte: And this `MethodImpl(Synchronized)` is not recommended on public types too by [Msdn](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.methodimploptions(v=vs.110).aspx). 

> Locking on the instance or on the type, as with the Synchronized flag, is not recommended for public types, because code other than your own can take locks on public types and instances. This might cause deadlocks or other synchronization problems.

Of course it is unlikely that other code do that. But that the path toward unrecoverable freezes. And still, why not using the `Lazy<T>` class for lazy initialisation where possible?

hazzik:
I'll redo this with Lazy

fredericDelaporte:
> Are there any users who do not add Iesi.Collections? No:) because it was a mandatory dependency, and people are using old NuGet's packages.config where the all nested dependency are listed.

Of course for people upgrading, but this will not be future proof, and will surely lead to undue bug reports, while other may think that is a bug and move out without even reporting.

This move looks to me as useless self-inflicted pain.

hazzik:
There are few reasons for this change:

1. There is only one class used from the Iesi.Collections, so it is not really justified to have this dependency.
1a. I do not want to force people to have a hard dependency on NHibernate, if they want to do persistence ignorance

2. We do not have resource to support the library, and removing the hard dependency will allow us to migrate to .NET Core 2.0

3. The nature of dependency is different to Antlr3 and ReLinq, they are frameworks, but Iesi is a library and in theory it have to be replaceable.

fredericDelaporte:
I am still not convinced. It is not a subject worth a "veto" anyway (not that I believe to have right to veto anything), so go on as you think is best.

The rest of this comment is why I am not convinced by the above three points. If you wish, do not bother reading it. We could simply acknowledge we disagree. Then just do what you think should be done.

1. Include that class code directly in NHibernate then, as this was done some time ago with proxy factories. I do not know the reasons for the move with proxies, but I guess ceasing to have them as an external custom reference had also removed a bunch of issues depending on which proxy providers were chosen (or mis-configured/mis-referenced).  
1a. I do not get it either. My projects do not need to have a direct reference to Iesi although I am using the ordered sets in mappings. My domain classes just reference `ISet<>` interface, and have no knowledge of Iesi. When instantiated by NHibernate, they end up having it injected by NHibernate, and it just works. (Provided the Iesi assembly is present along the project assembly. If it is missing on the file system, trying to load an ordered set collection from an entity fails of course. As it would do with this PR.) Yes, when I create a transient instance, I do not initialize them with Iesi, and the order is not guaranteed at that time. But it never happened to me to have an application requiring set ordering when creating transient instances.  
Now, if by "persistence ignorance" you mean entity instances which does not have any internal composition with some NHibernate related types, well, we have a lot of work to do, since currently a persistent instance use NHibernate persistent collections. Removing the explicit Iesi dependency from NHibernate will not change that.  
So probably we should document that this functionality uses a non standard concrete set implementation. (And if this PR is merged, it really should be documented in mapping reference section that an explicit reference to Iesi is needed for using the `order-by` feature on `set`, instead of letting users discover it fails otherwise.)
2. If handling this class is such an issue (even if directly integrated in NHibernate sources), we should not even make the effort of loading it dynamically, but replace the feature by a new "NotImplemented" or "NotSupported" exception, like ordered map. (But I expect we will have some bad feedback from this move after release.)  
.Net Portability Analyzer find Iesi to be 100% compatible with .Net Core 2.0 (and .Net Standard 2.0; for lower versions, serializable need to be removed), so the only issue seems to be handling the dual release if we keep it as a separate assembly.
3. Why not giving the opportunity to users to tell NHibernate to use some other dictionaries/lists implementations too, if we go that way?

src/NHibernate/Type/GenericOrderedSetType.cs
```diff
@@ -11,6 +11,9 @@ namespace NHibernate.Type
 	[Serializable]
 	public class GenericOrderedSetType<T> : GenericSetType<T>
 	{
+		// ReSharper disable once StaticMemberInGenericType
```
fredericDelaporte: It tend to favor disabling globally such R# rules rather than putting R# specific comment in code. (Indeed I have ended up disabling so many rules that were too noisy at my previous work, that I have ended disabling wholly R#, and not missing it much.)

src/NHibernate/Type/GenericOrderedSetType.cs
```diff
@@ -24,9 +27,23 @@ public GenericOrderedSetType(string role, string propertyRef)
...
 		}
 
+		static ConstructorInfo GetSetConstructor()
+		{
+			var type = System.Type.GetType("Iesi.Collections.Generic.LinkedHashSet`1, Iesi.Collections", false);
+			if (type == null)
+				throw new TypeLoadException(
+					@"Could not load Iesi.Collections.Generic.LinkedHashSet`1
+
+Please make sure that you have referenced the Iesi.Collections package.");
```
fredericDelaporte: For having it a bit more explicit, and letting users guess they are not more NHibernate features which may blow if not referencing Iesi:
`Please make sure that you have referenced the Iesi.Collections package, or do not use the order by feature on set mapping.`

src/NHibernate/Type/GenericOrderedSetType.cs
hazzik: I think this warning is useful, but in this case, I want to have exact this side effect - having a distinct static field per closed generic type.

src/NHibernate/Type/GenericOrderedSetType.cs
hazzik: I'm planning to add a release note about this implicit dependency in a "Possible breaking change" part. I'll try to reword the exception to provide more clarity.

hazzik:
I'm not by any means trying to argue:)

1. There were some problems with the proxy factories: firstly, they are again frameworks, but such a framework which is used by people for the sort of their different needs (Dependency Injection, for ex). So, NH team had to maintain the 3-4 providers to make sure that we use the compatible and fresh, and etc. Then there is a new version which is incompatible, and people want to upgrade (or do not want to upgrade instead). At the end, you have gazillions versions to support. Supporting "bridging" packages _is hard_. So this was the same matter of a least effort.

   1a. I wouldn't agree that having the different behavior depending on the state of an entity is a good idea. So I would say if someone wants to use the OrderedSet feature they need to use it properly.

2. We load a lot of things dynamically, so one more or one less, who cares? 
    Even a PortableAnalyzer shows that Iesi is fully convertible to .NET Standard/Core, we still need to do the work and change _all_ the build tooling: make sure that we build with proper MSBuild version, change the way we package NuGet packages, how we run tests, etc.
   Throwing NotSupported would be misleading because this feature is in fact supported.
   Throwing any other type of the exceptions would require us (or someone else) to provide the compatibility package where it is supported/implemented. It will require greater effort than before.

3. Yeah, the CollectionFactory is already fully replaceable as a whole.

---

So, there are a few possible ways forward:

1. Completely remove "order-by" on sets (#245)
2. OR,
    1. Leave everything as is and overcome a pain of upgrading Iesi to .NET Core
    2. Dynamically load Iesi
        1. Make ordered-set configurable
    3. Throw exception and maintain the collection factory package.

ngbrown:
I'm not a fan of adding more reflection when we have NuGet's dependency management.  Reflection makes it harder to reason about what is needed for an application before run-time.

hazzik:
Ok, closing

fredericDelaporte:
Thanks!

PR #605, opened 2017-04-25
NH-3990 - Upgrade to VS2017 project structure

hazzik:


hazzik:
Doesn't seem to be buildable on a build agent without installing full VS

fredericDelaporte:
The error in TeamCity does not let me think it needs a full VS. That looks like Antlr3 build is trying to use an old MsBuild dependency, and can no more find it when launched by the newest tools.

`C:\Windows\system32\config\systemprofile\.nuget\packages\antlr3\3.5.1\build\Antlr3.targets(126,5): error MSB4062: The "Antlr3.Build.Tasks.AntlrClassGenerationTask" task could not be loaded from the assembly C:\Windows\system32\config\systemprofile\.nuget\packages\antlr3\3.5.1\build\.\AntlrBuildTask.dll. Could not load file or assembly 'Microsoft.Build.Utilities.v3.5, Version=3.5.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.`

It seems to fail because it has a dependency on the old `Microsoft.Build.Utilities.v3.5`. I do not think having a full VS would avoid that failure. Enabling .NET Framework 3.5 feature on windows is supposed to provide it. The build agent seems to have it. So it looks more like the new build command is unable to resolve all the required dependencies.

hazzik:
There are a few issues:

1. BuildTools does not know how to build the Sdk projects and so there is "Microsoft.NET.Sdk is missing" error (https://github.com/Microsoft/msbuild/issues/1697)
2. If you specify MSBuildSDKsPath there is an error about VB projects saying that `Restore` task does not exist. (This way it does not build on my machine either)
3. If you put `dotnet msbuild %*` as suggested you get this Antlr3 error (which I get on my machine as well)

There is another workaround - to copy the Sdk folder from VS to BuildTools, but I don't want to do it.

So yes, it's not buildable in the current state without full VS

fredericDelaporte:
I have found [this](https://github.com/Microsoft/msbuild/issues/1333) which looks related. If this is the root of our failure, we then have either to wait for the dotnet sdk to get fixed, or upgrade all build tasks for using newer msbuild, or avoid having parts to build with older msbuild.

I am not sure but I believe the only option we can do ourselves is the last, by reverting 9292e6e88788e7a496f13592f31d60aac0a1d26b by example.

hazzik:
>I am not sure but I believe the only option we can do ourselves is the last, by reverting 9292e6e by example.

Too much effort:)

hazzik:
I've filled the bug: https://github.com/antlr/antlrcs/issues/71

hazzik:
>I am not sure but I believe the only option we can do ourselves is the last, by reverting 9292e6e by example.

I performed some investigation, it seems now it would be the best option.

src/NHibernate.TestDatabaseSetup/NHibernate.TestDatabaseSetup.csproj
```diff
@@ -1,112 +1,38 @@
...
+    <ProjectReference Include="..\NHibernate.Test\NHibernate.Test.csproj" />
+  </ItemGroup>
+
   <ItemGroup>
     <Reference Include="FirebirdSql.Data.FirebirdClient">
       <HintPath>..\..\lib\teamcity\firebird\FirebirdSql.Data.FirebirdClient.dll</HintPath>
     </Reference>
+    <Reference Include="Mono.Security">
+      <HintPath>..\..\lib\teamcity\postgresql\Mono.Security.dll</HintPath>
+    </Reference>
```
hazzik: Yes, probably a rebate glitch

hazzik:
Yes, it requires either a clean checkout or purge of `obj` folders. Also it requires SDK2.0 Preview 2 and, possible, VS2017.3 Preview 6 (check without the VS first).

hazzik:
>NHibernate.TestDatabaseSetup could not be loaded.

Probably because of Mono.Security

fredericDelaporte:
E, F, G & I builds fixed by the Core preview. (I have still the non preview VS2017.)

The Nuget difficulties for TestDatabaseSetup remains. Trying to build anything while Nuget struggles with TestDatabasSetup just fails, whatever the project. So it is a bit annoying, but well, not blocking either.
No, Mono.Security does not seem implied into this. I have tested that by removing it, that trouble is still there.

D still fails, if projects bin directories where cleaned before rebuilding with the new project format. (Otherwise it uses obsolete binaries from compilation with older project format.)

2.0 release is slated for Q3, so we will have to wait up to two months for the official release. Do you want to merge a dependency on a preview in master, or to wait for it to be released?

The new project format seems detrimental for VS load time of the solution. But without working with it and many branches, I cannot say if it will really be an annoyance. (And I still have R# enabled, which has also quite an impact on solution loading time.)

hazzik:
>D still fails

I did not realize that it requires to have binaries built by VS to run tests. Weird. 

>Do you want to merge a dependency on a preview in master, or to wait for it to be released?

I want to merge, don't mind being dependent on preview. It's just an dev dependency, so does not matter.

hazzik:
I've fixed "D" menu to the state as it was before this PR.

It requires to have tools packages restored & binaries being built by VS. I think we shall tackle this in a separate PR.

fredericDelaporte:
> I did not realize that it requires to have binaries built by VS to run tests. Weird.

I was aware of this, and am using it for globally testing the last build I have done in VS. I was thinking this was made this way on purpose.

> I want to merge, don't mind being dependent on preview. It's just an dev dependency, so does not matter.

Ok; Contributing.md should probably be updated for mentioning the build dependency on Core 2.0 (Preview or not indeed).

NHibernate.nunit
```diff
@@ -1,13 +1,13 @@
 ﻿<NUnitProject>
   <Settings activeconfig="Debug" processModel="Multiple" />
   <Config name="Debug" binpathtype="Auto" runtimeFramework="net-4.0">
-    <assembly path="src\NHibernate.TestDatabaseSetup\bin\Debug\NHibernate.TestDatabaseSetup.dll" />
-    <assembly path="src\NHibernate.Test\bin\Debug-2.0\NHibernate.Test.dll" />
-    <assembly path="src\NHibernate.Test.VisualBasic\bin\Debug\NHibernate.Test.VisualBasic.dll" />
+    <assembly path="src\NHibernate.TestDatabaseSetup\bin\Debug\net461\NHibernate.TestDatabaseSetup.dll" />
+    <assembly path="src\NHibernate.Test\bin\Debug\net461\NHibernate.Test.dll" />
+    <assembly path="src\NHibernate.Test.VisualBasic\bin\Debug\net461\NHibernate.Test.VisualBasic.dll" />
```
fredericDelaporte: This one has not moved. It is still in `NHibernate.Test.VisualBasic\bin\Debug\NHibernate.Test.VisualBasic.dll`.

NHibernate.nunit
```diff
@@ -1,13 +1,13 @@
...
   </Config>
   <Config name="Release" binpathtype="Auto" runtimeFramework="net-4.0">
-    <assembly path="src\NHibernate.TestDatabaseSetup\bin\Release\NHibernate.TestDatabaseSetup.dll" />
-    <assembly path="src\NHibernate.Test\bin\Release-2.0\NHibernate.Test.dll" />
-    <assembly path="src\NHibernate.Test.VisualBasic\bin\Release\NHibernate.Test.VisualBasic.dll" />
+    <assembly path="src\NHibernate.TestDatabaseSetup\bin\Release\net461\NHibernate.TestDatabaseSetup.dll" />
+    <assembly path="src\NHibernate.Test\bin\Release\net461\NHibernate.Test.dll" />
+    <assembly path="src\NHibernate.Test.VisualBasic\bin\Release\net461\NHibernate.Test.VisualBasic.dll" />
```
fredericDelaporte: Same as above.

NHibernate.nunit
hazzik: True. VB has not been migrated to VS2017 format as it is supported only in VS2017.3

hazzik: Done

fredericDelaporte: I was not knowing it would be supported fo VB soon. I was not finding any info about vb support case, so was not expecting it at all.

hazzik: It is, I saw somewhere on SO, but could not find the link now.

src/NHibernate.Tool.HbmXsd/NHibernate.Tool.HbmXsd.csproj
```diff
@@ -1,95 +1,18 @@
...
-    <UpdateRequired>false</UpdateRequired>
-    <MapFileExtensions>true</MapFileExtensions>
-    <ApplicationRevision>0</ApplicationRevision>
-    <ApplicationVersion>1.0.0.%2a</ApplicationVersion>
-    <IsWebBootstrapper>false</IsWebBootstrapper>
-    <UseApplicationTrust>false</UseApplicationTrust>
-    <BootstrapperEnabled>true</BootstrapperEnabled>
-    <TargetFrameworkProfile>
-    </TargetFrameworkProfile>
+    <TargetFramework>net461</TargetFramework>
```
ngbrown: To output an `.EXE`, the following need to be added:

```xml
    <OutputType>Exe</OutputType>
    <StartupObject>NHibernate.Tool.HbmXsd.Program</StartupObject>
```

ngbrown: Created pull request #672

PR #607, opened 2017-04-26
NH-3684 - Remove obsolete <Time>Part extension methods of QueryOver

hazzik:


fredericDelaporte:
Well, I have put in my personal todo list:

- drop old obsolete (comme YearPart, obsolète depuis la 4.0)

So of course I am for it!

I was thinking about doing that after sorting out the transaction handling troubles we have.

 I was considering checking all obsolete in code, listing them for deletion in 5.0 if there were already obsolete in 4.0, and adding a comment on others for knowing when to delete them.

About that comment, I have started putting comments like `// to removed for 6.0` on some obsolete I have introduced, but now that does not appear to me as future proof: better tell in the comment in which version it was set as obsolete, to let later dev decide when to drop them.

hazzik:
We can use something like [ObsoleteEX](https://github.com/Fody/Obsolete) from Fody.

PR #609, opened 2017-04-27
Fix mapping of intermediate inherited classes

### LINKED ISSUE #1358: NH-3992 - Intermediate inherited classes are not mapped correctly
**bpogson** created an [issue](https://nhibernate.jira.com/browse/NH-3992) — <time datetime="2017-04-27T03:41:56.385Z" title="2017-04-27T03:41:56.385Z">27th April 2017, 3:41:56</time>:
> If two mapped classes A and B are in a class hierarchy where B indirectly inherits from A and mappings are declared for A and B, but not for the classes between A and B, NHibernate doesn't find any members declared in the classes between A and B that are mapped in B.
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2017-04-27T03:43:44.503Z" title="2017-04-27T03:43:44.503Z">27th April 2017, 3:43:44</time>:
> <~bpogson> can you please provide a test case? Better if it is in a form of Pull Request on Github
> 
---
**bpogson** added a comment — <time datetime="2017-04-27T04:20:56.929Z" title="2017-04-27T04:20:56.929Z">27th April 2017, 4:20:56</time>:
> <~hazzik>, I have written a unit test and Bill has created a fix for this, both are submitted under https://github.com/nhibernate/nhibernate-core/pull/609
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2017-04-27T04:28:15.112Z" title="2017-04-27T04:28:15.112Z">27th April 2017, 4:28:15</time>:
> Cool, thanks. It seems that was too fast in that case
>

ByronPogson:
If two mapped classes A and B are in a class hierarchy where B indirectly inherits from A and mappings are declared for A and B, but not for the classes between A and B, NHibernate doesn't find any members declared in the classes between A and B that are mapped in B.

Fixes #1358

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3992/FixtureByCode.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3992/FixtureByCode.cs
@@ -1,0 +1,413 @@
+﻿using NHibernate.Mapping.ByCode;
+using NUnit.Framework;
+using System.Linq;
+
+namespace NHibernate.Test.NHSpecificTest.NH3992
+{
+	/// <summary>
+	/// Fixture using 'by code' mappings
+	/// </summary>
+	public class ByCodeFixture
+	{
+		[Test]
+		public void Test_MappedSubclass_AllMapsAtTopLevel()
+		{
+			var mapper = new ModelMapper();
+			mapper.AddMapping(typeof(BaseEntityMapping));
+			mapper.AddMapping(typeof(MappedEntitySubclassMapping));
+
+			var mapping = mapper.CompileMappingForAllExplicitlyAddedEntities();
+
+			// Check that the subclass was mapped
+			var baseMapping = mapping.RootClasses.SingleOrDefault(x => x.Name == "BaseEntity");
+			Assert.IsNotNull(baseMapping, "Mapping did not return mapping for BaseEntity");
+			var targetMapping = mapping.SubClasses.SingleOrDefault(x => x.Name == "MappedEntity");
+			Assert.IsNotNull(targetMapping, "Mapping did not return mapping for subclass MappedEntity");
+			// Check that all three fields are mapped
+			var baseProperty = baseMapping.Properties.SingleOrDefault(p => p.Name == "BaseField");
+			Assert.IsNotNull(baseProperty, "Base class mapping did not map base class property");
+			var extendedProperty = targetMapping.Properties.SingleOrDefault(p => p.Name == "ExtendedField");
+			Assert.IsNotNull(extendedProperty, "Sub class mapping did not map extended (class not mapped) class property");
+			var topLevelProperty = targetMapping.Properties.SingleOrDefault(p => p.Name == "TopLevelField");
+			Assert.IsNotNull(topLevelProperty, "Sub class mapping did not map base class property");
+
+		}
+
+		[Test]
+		public void Test_MappedJoinedSubclass_AllMapsAtTopLevel()
+		{
+			var mapper = new ModelMapper();
+			mapper.AddMapping(typeof(BaseEntityMapping));
+			mapper.AddMapping(typeof(MappedEntityJoinedSubclassMapping));
+
+			var mapping = mapper.CompileMappingForAllExplicitlyAddedEntities();
+
+			// Check that the subclass was mapped
+			var baseMapping = mapping.RootClasses.SingleOrDefault(x => x.Name == "BaseEntity");
+			Assert.IsNotNull(baseMapping, "Mapping did not return mapping for BaseEntity");
+			var targetMapping = mapping.JoinedSubclasses.SingleOrDefault(x => x.Name == "MappedEntity");
+			Assert.IsNotNull(targetMapping, "Mapping did not return mapping for subclass MappedEntity");
+			// Check that all three fields are mapped
+			var baseProperty = baseMapping.Properties.SingleOrDefault(p => p.Name == "BaseField");
+			Assert.IsNotNull(baseProperty, "Base class mapping did not map base class property");
+			var extendedProperty = targetMapping.Properties.SingleOrDefault(p => p.Name == "ExtendedField");
+			Assert.IsNotNull(extendedProperty, "Sub class mapping did not map extended (class not mapped) class property");
+			var topLevelProperty = targetMapping.Properties.SingleOrDefault(p => p.Name == "TopLevelField");
+			Assert.IsNotNull(topLevelProperty, "Sub class mapping did not map base class property");
+
+		}
+
+		[Test]
+		public void Test_MappedUnionSubclass_AllMapsAtTopLevel()
+		{
+			var mapper = new ModelMapper();
+			mapper.AddMapping(typeof(BaseEntityMapping));
+			mapper.AddMapping(typeof(MappedEntityUnionSubclassMapping));
+
+			var mapping = mapper.CompileMappingForAllExplicitlyAddedEntities();
+
+			// Check that the subclass was mapped
+			var baseMapping = mapping.RootClasses.SingleOrDefault(x => x.Name == "BaseEntity");
+			Assert.IsNotNull(baseMapping, "Mapping did not return mapping for BaseEntity");
+			var targetMapping = mapping.UnionSubclasses.SingleOrDefault(x => x.Name == "MappedEntity");
+			Assert.IsNotNull(targetMapping, "Mapping did not return mapping for subclass MappedEntity");
+			// Check that all three fields are mapped
+			var baseProperty = baseMapping.Properties.SingleOrDefault(p => p.Name == "BaseField");
+			Assert.IsNotNull(baseProperty, "Base class mapping did not map base class property");
+			var extendedProperty = targetMapping.Properties.SingleOrDefault(p => p.Name == "ExtendedField");
+			Assert.IsNotNull(extendedProperty, "Sub class mapping did not map extended (class not mapped) class property");
+			var topLevelProperty = targetMapping.Properties.SingleOrDefault(p => p.Name == "TopLevelField");
+			Assert.IsNotNull(topLevelProperty, "Sub class mapping did not map base class property");
+
+		}
+
+		[Test]
+		public void Test_MappedSubclassInterface_AllMapsAtTopLevel()
+		{
+			var mapper = new ModelMapper();
+			mapper.AddMapping(typeof(BaseInterfaceMapping));
+			mapper.AddMapping(typeof(MappedEntityFromInterfaceSubclassMapping));
+
+			var mapping = mapper.CompileMappingForAllExplicitlyAddedEntities();
+
+			// Check that the subclass was mapped
+			var baseMapping = mapping.RootClasses.SingleOrDefault(x => x.Name == "IBaseInterface");
+			Assert.IsNotNull(baseMapping, "Mapping did not return mapping for IBaseInterface");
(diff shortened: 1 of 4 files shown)


src/NHibernate/Mapping/ByCode/IModelInspector.cs
```diff
@@ -7,7 +7,7 @@ public interface IModelInspector
 	{
 		bool IsRootEntity(System.Type type);
 		bool IsComponent(System.Type type);
-		bool IsEntity(System.Type type);
+		bool IsEntity(System.Type type, bool explicitlyDeclared = false);
```
gliljas: Since the interface is changed, an IsDeclaredEntity method would make more sense, or is there something I'm not seeing?

src/NHibernate.Test/NHSpecificTest/NH3992/FixtureByCode.cs
```diff
@@ -0,0 +1,71 @@
...
+			{
+				session.Delete("from System.Object");
+
+				session.Flush();
+				transaction.Commit();
+			}
+		}
+
+		[Test]
+		public void Test_InhertiedClasses_AllMapsAtTopLevel()
```
gliljas: The test should test the generated mapping, not the outcome of the mapping.

fredericDelaporte:
Thanks for your contribution,

May you please reword your commits for including the issue number?

Something like:

- NH-3992 - Intermediate inherited classes test cases
- NH-3992 - Intermediate inherited classes fix

This allows the commits to be associated in Jira, as explained in [contributing guidelines](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md).

This PR introduces default values for parameters. It does not look to me as an accepted practice in NHibernate, since I can not find any of them (excepted in test project, where I have put some recently myself). And well, following @gliljas recommendation would avoid it.

src/NHibernate/Mapping/ByCode/IModelInspector.cs
ByronPogson: Given that default parameters aren't currently used in the code base this makes sense to expose two different methods, will update

src/NHibernate.Test/NHSpecificTest/NH3992/FixtureByCode.cs
ByronPogson: Seems obvious now that someone points it out! Will update and re-push

ByronPogson:
Thanks for the feedback guys, I've made the changes are requested

fredericDelaporte:
Thanks,

I wonder whether `IsMappedClass` would be a more explicit name than `IsDeclaredEntity` or not. Note that I am not used to mapping by code, I personally only use hbm currently. That is how I understand that new method from my hbm mapping perspective.

ByronPogson:
The original existing method that was being called was IsEntity, so I think IsDeclaredEntity was suggested because it fits with that . Happy to change if others agree, I was just going with the suggestion from @gliljas

src/NHibernate/Mapping/ByCode/ModelMapper.cs
```diff
@@ -736,7 +736,7 @@ private System.Type GetEntityBaseType(System.Type type)
 			while (analyzingType != null && analyzingType != typeof (object))
 			{
 				analyzingType = analyzingType.BaseType;
-				if (modelInspector.IsEntity(analyzingType))
+				if (modelInspector.IsDeclaredEntity(analyzingType))
 				{
 					return analyzingType;
 				}
```
hazzik: Next line shall be also changed to IsDeclaredEntity

hazzik: (with tests)

ByronPogson: Of course :) Just trying to work out a test that will make sure that line is hit

src/NHibernate/Mapping/ByCode/SimpleModelInspector.cs
```diff
@@ -31,7 +31,13 @@ public override bool IsRootEntity(System.Type entityType)
...
 			public bool IsEntity(System.Type type)
 			{
-				return RootEntities.Contains(type) || type.GetBaseTypes().Any(t => RootEntities.Contains(t)) || HasDelayedEntityRegistration(type);
+				return RootEntities.Contains(type) || type.GetBaseTypes().Any(t => RootEntities.Contains(t))
+				       || HasDelayedEntityRegistration(type);
+			}
+
+			public bool IsDeclaredEntity(System.Type type)
+			{
+				return RootEntities.Contains(type) || HasDelayedEntityRegistration(type);
```
hazzik: I doubt that HasDelayedEntityRegistration is required here. Could you please check with test that this is required?

If this is not required, then this method is equivalent to `IsRootEntity`

hazzik:
I've replaced `IsDeclaredEntity` with `IsRootEntity` and your tests still work

src/NHibernate.Test/NHSpecificTest/NH3992/FixtureByCode.cs
```diff
@@ -0,0 +1,92 @@
...
+using NHibernate.Mapping.ByCode;
+using NUnit.Framework;
+using System.Linq;
+
+namespace NHibernate.Test.NHSpecificTest.NH3992
+{
+	/// <summary>
+	/// Fixture using 'by code' mappings
+	/// </summary>
+	public class ByCodeFixture : TestCaseMappingByCode
```
hazzik: Do not need to inherit TestCaseMappingByCode here.

src/NHibernate.Test/NHSpecificTest/NH3992/FixtureByCode.cs
```diff
@@ -0,0 +1,92 @@
...
+	{
+		protected override HbmMapping GetMappings()
```
hazzik: Not needed as per comment above

src/NHibernate/Mapping/ByCode/ModelMapper.cs
hazzik: I think it would be triggered with a subclass of a subclass 

```
GrandSubclass -> Subclass -> Unmapped -> RootEntity
```

hazzik: You need to have an interface as your `BaseEntity`

src/NHibernate/Mapping/ByCode/SimpleModelInspector.cs
hazzik: I think it would be required with a subclass of a subclass 

```
GrandSubclass -> Subclass -> Unmapped -> RootEntity
```

ByronPogson:
Sorry for the delay. I have now made the changes above in regards to a) unit testing around interface inheritor and b) using the existing IsRootEntity method instead

ByronPogson:
@hazzik I think this is occurring because the test is setting up it's classes incorrectly. The Test has the following classes:
abstract Animal
abstract Reptile: Animal
Lizard: Reptile
abstract Mammal: Animal
Dog: Mammal
Cat: Mammal

These are then _all_ passed to the ConventionModelMapper.CompileMappingsFor() which throws the following exception:
System.NotSupportedException : The class NHibernate.Test.NHSpecificTest.NH2691.Reptile can't be instatiated and does not have mapped subclasses; 
possible solutions:
- don't map the abstract class
- map its subclasses.

This was probably fine when the intermediate classes were ignored but now they're not this is causing problems with this specific test. I can't find any documentation around how ConventionModelMapper should be used and what the intended usage is. All other tests using this are still passing. Removing the abstract intermediates (as suggested by the error) resolves the issue for this test. Is that sufficient or is there some documentation around which classes should be passed to this mapper?

hazzik:
I think it's a valid use case. The code generated is like following:

<?xml version="1.0" encoding="utf-8"?>
<hibernate-mapping xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" namespace="NHibernate.Test.NHSpecificTest.NH2691" assembly="NHibernate.Test, Version=5.0.0.0, Culture=neutral, PublicKeyToken=null" xmlns="urn:nhibernate-mapping-2.2">
  <class name="Animal" abstract="true">
    <id name="Id" type="Int32" />
    <discriminator />
    <property name="Description" />
    <property name="Sequence" />
  </class>
  <subclass name="Reptile" extends="Animal">
    <property name="BodyTemperature" />
  </subclass>
  <subclass name="Mammal" extends="Animal">
    <property name="Pregnant" />
    <property name="BirthDate" />
  </subclass>
  <subclass name="Lizard" extends="Animal">
    <property name="BodyTemperature" />
  </subclass>
  <subclass name="Dog" extends="Animal">
    <property name="Pregnant" />
    <property name="BirthDate" />
  </subclass>
  <subclass name="Cat" extends="Animal">
    <property name="Pregnant" />
    <property name="BirthDate" />
  </subclass>
</hibernate-mapping>

hazzik:
> I doubt that HasDelayedEntityRegistration is required here. Could you please check with test that this is required?
> If this is not required, then this method is equivalent to IsRootEntity
> I think it would be required with a subclass of a subclass
> 
> ```
> GrandSubclass -> Subclass -> Unmapped -> RootEntity
> ```
It seems that the tests shows that this is required

ByronPogson:
I've added in a further test to ensure that the mapping where intermediate classes are abstract and reverted to the previous method including the HasDelayedEntityRegistration

hazzik:
I saw that you pushed this commit, so my last to comments affect it. I mean
that after this commit it probably still possible to get incorrect mappings
built

ByronPogson:
The comment that HasDelayedEntityRegistration is required? I've put that check back in. What other circumstances need to be tested?

hazzik:
Oh. My two comments was not send through. :(

hazzik:
I think it will still fail in some longer inheritance chains:

Mapped -> Unmapped -> Mapped -> Root
Mapped -> Mapped -> Unmapped -> Root (this most likely works)
Mapped -> Unmapped -> Mapped -> Unmapped -> Root

For these longer chains, could you please add tests to ensure that children "extend" correct parents?

ByronPogson:
Those chains you suggested don't work in the current code when calling the ConventionModelMapper with only the mapped classes. The unmapped classes don't get set as the "extend" and their properties aren't mapped. Looking through the code this is because the ConventionModelMapper uses the SimpleModelInspector and the IsDeclaredEntity method that is implemented (SimpleModelInspector:706) is identical to the IsEntity method. I suspect that either this method should return the declaredRelult parameter immediately or a isDeclaredEntity func should be added on line 174 with the default being set to return declared which could then be overridden as required.

This along however doesn't solve the issue for the ConventionModelMapper as presently no classes are registered as explicitly declared so none of them get returned as declared when the IModelExplicitDeclarationsHolder is declared model is examined. One solution I have is to append events in the ConventionModelMapper for each of the after class mapping events to add these into the declarations holder but this feels wrong as it's only addressing this issue and doesn't result in anything else being populated which seems outside the scope of this particular issue. Can anyone suggest a better solution to this?

ByronPogson:
We're keen to get this included in 5.1, can anyone provide feedback on the above so we can finish this one off?

ByronPogson:
@hazzik can you please provide guidance on this?

@billpoole-mi will be taking this one over

hazzik:
>It took me some time to figure it out.

It took me almost two years to figure out what the fix should look like 😅.

hazzik:
Squashed commits to reduce noise. And force pushed. No changes to last review by @fredericDelaporte

PR #614, opened 2017-04-28
NH-3723 - Some tests are failing when log level set to DEBUG

hazzik:


fredericDelaporte:
NH-2297 test fails because in debug log level, the mapping loading process try to log NHibernate type of properties, which throws due to the test (which is expected, as an `InvalidOperationException`). But since we are in mapping loading, the exception get encapsulated into a `MappingException`, and then the test fails to recognize it.

The additional log call done in mappings loading is located [here](https://github.com/hazzik/nhibernate-core/blob/master/src/NHibernate/Cfg/XmlHbmBinding/PropertiesBinder.cs#L186) with its code [here](https://github.com/hazzik/nhibernate-core/blob/master/src/NHibernate/Cfg/XmlHbmBinding/MappingLogExtensions.cs).

Swallowing exceptions in the log code could avoid this. I would rather extend the test condition for accepting exceptions having a corresponding inner exception too.

The same trouble happen for NH-712.

hazzik:
The problem that error happens as soon as you add mapping, and not when the
tests do expect failure

fredericDelaporte:
Right, the error happens too early in debug log level for the tests, on top of changing the exception. Then either we swallow exceptions in the additional logging occurring at debug log level, or we check the log level in the fixture setup and tests code for not adding the mappings and ignoring the test when the level is debug.

hazzik:
I prefer to swallow the exception at debug level (and show type as "unable to load") and then throw the error where the test expect it to be thrown

fredericDelaporte:
Ok, agreed.

src/NHibernate.Test/App.config
```diff
@@ -78,8 +78,8 @@
 		</appender>
 
 		<root>
-			<priority value="WARN" />
-			<appender-ref ref="console" />
+			<priority value="DEBUG" />
+			<!--<appender-ref ref="console" />-->
```
fredericDelaporte: I suppose those changes will be reverted before merging.

src/NHibernate/Cfg/XmlHbmBinding/MappingLogExtensions.cs
```diff
@@ -1,22 +1,39 @@
...
 		}
+
+		static string SafeGetPropertyTypeName(Property property)
+		{
+			try
+			{
+				//property.Type property can trigger a type load
+				if (property.Type != null) return property.Type.Name;
+			}
+			catch (MappingException)
```
fredericDelaporte: At that point it is not yet converted to a `MappingException`. This occurs later (higher in the stack). So here, for our two current failing test cases, it can be an `InvalidOperationException` or an `InstanciationException`.

The more "future proof" here would be to swallow any `Exception`.

I wonder whether we should additionally log that exception or not. Since this occurs when generating debug log only, I would tend to log it as debug too, if we log it at all.

src/NHibernate.Test/App.config
hazzik: Yes

src/NHibernate/Cfg/XmlHbmBinding/MappingLogExtensions.cs
fredericDelaporte: Ok, fixed now.

src/NHibernate/Cfg/XmlHbmBinding/MappingLogExtensions.cs
```diff
@@ -1,22 +1,39 @@
...
 		}
+
+		static string SafeGetPropertyTypeName(Property property)
+		{
+			try
+			{
+				//property.Type property can trigger a type load
+				if (property.Type != null) return property.Type.Name;
+			}
+			catch (Exception)
```
fredericDelaporte: `using System;` missing, if you have not already looked why it is failing.

fredericDelaporte:
To be squashed, of course.

src/NHibernate/Cfg/XmlHbmBinding/MappingLogExtensions.cs
hazzik: Thanks.

hazzik:
Cleaned up version has been applied to master.

PR #616, opened 2017-04-30
NH-3998 - Alias of a column should be calculated based on a CanonicalName rather than raw name

hazzik:
Expected to fix 125 tests on SqlServer CE

### DIFF
--- a/src/NHibernate/Mapping/Column.cs
+++ b/src/NHibernate/Mapping/Column.cs
@@ -127,17 +127,18 @@ namespace NHibernate.Mapping
 
 		private string GetAlias(int maxAliasLength)
 		{
-			string alias = _name;
+			var name = CanonicalName;
+			string alias = name;
 			string suffix = UniqueInteger.ToString() + StringHelper.Underscore;
 
-			int lastLetter = StringHelper.LastIndexOfLetter(_name);
+			int lastLetter = StringHelper.LastIndexOfLetter(name);
 			if (lastLetter == -1)
 			{
 				alias = "column";
 			}
-			else if (lastLetter < _name.Length - 1)
+			else if (lastLetter < name.Length - 1)
 			{
-				alias = _name.Substring(0, lastLetter + 1);
+				alias = name.Substring(0, lastLetter + 1);
 			}
 
 			// Updated logic ported from Hibernate's fix for HHH-8073.
@@ -147,9 +148,9 @@ namespace NHibernate.Mapping
 			// reason, the checks for "_quoted" and "rowid" looks redundant. If you remove
 			// those checks, then the double checks for total length can be reduced to one.
 			//    But I will leave it like this for now to make it look similar. /Oskar 2016-08-20
-			bool useRawName = _name.Length + suffix.Length <= maxAliasLength &&
+			bool useRawName = name.Length + suffix.Length <= maxAliasLength &&
 			                  !_quoted &&
-			                  !StringHelper.EqualsCaseInsensitive(_name, "rowid");
+			                  !StringHelper.EqualsCaseInsensitive(name, "rowid");
 			if (!useRawName)
 			{
 				if (suffix.Length >= maxAliasLength)


fredericDelaporte:
This could be a breaking changes for databases supporting column differing only by cases. I am checking a test case locally with a Sql Server db set to cs collation (that is all it takes to get a case sensitive db with Sql Server).

fredericDelaporte:
Now I confirm it is a breaking change. I can push the test to your branch if you wish. (Probably as a WIP, it is far from being db agnostic currently.)

hazzik:
Go for it

hazzik:
I think we should set up at least one DB with case sensitive column names on Build pipeline.

fredericDelaporte:
Test NH-3998 added for non-reg. Will be ignored if db is not "detected" as column sensitive. Does manually creates the schema, as NHibernate does not currently support creating columns differing only by names (does only create one).

For testing a case sensitive db, we have some work ahead: hundreds of tests have some casing inconsistencies in the mappings, causing them to fail on case sensitive db. Some tests assume ci string comparison too, and since with SQL Server, changing sensitivity of columns is done through collation, those tests fail too.

I further expect some issues when tempdb is involved: its collation depends of the server with SQL Server, so if server collation and database collation does not match, it may create a bunch of additional issues.

hazzik:
Ok, then we need to fix a deeper problem.

NHibernate.Test.Component.Basic.ComponentTest (and other) fails because of semi-"invalid" mapping:

```
<hibernate-mapping 
	xmlns="urn:nhibernate-mapping-2.2" 
	assembly="NHibernate.Test" 
	namespace="NHibernate.Test.Component.Basic">

	<class name="User" table="T_USER">
		<id name="UserName"/>
		<timestamp name="LastModified"/>
		<property name="Password" not-null="true" optimistic-lock="false"/>
		<component name="Person">
			<property name="Name" update="false" not-null="true"/>
			<property name="Dob" update="false" not-null="true"/>
			<property name="Address"/> <!-- here column assumed to be "Address" (with capital A) -->
			<property name="PreviousAddress" insert="false"/>
			<property name="Yob" formula="year(dob)"/>
			<property name="CurrentAddress"
				column="address" <!-- here column as it is stated -->
				insert="false"
				update="false"/>
		</component>
	</class>
</hibernate-mapping>
```

fredericDelaporte:
Yes, and this only one among many cases. (Hundreds of tests concerned, but not as many inconsistencies to fix: many tests fail due to the same mapping.)

For some tests, we will have to add at least ways to detect column case sensitivity and string comparison sensitivity. For a db like SQL Server which can be configured about case sensitivity, either we would have to issue some test query as do the case I have added, or we will need to add some configuration parameter for letting user tell when they use a case sensitive db on columns.

And if we want to support all cases, things get worst: MySql by example is case sensitive on tables names when running under linux, but case insensitive when running under windows. And under Mac, it depends on the used file system... (Indeed that is the case for all OSes.) But about columns name, it is always case insensitive.

hazzik:
It seems that the rule as following: we can have many collumns with the same name in different cases as long as they are quoted. Unquoted columns will be considered the same column. So, your test need to be adjusted to include quotes around column names.

fredericDelaporte:
Checking that with and without this pr change, will amend the commit.

hazzik:
So, quoting lowercase address column makes the test pass:

```
    SELECT
        user0_.UserName as UserNa1_0_0_,
        user0_.LastModified as LastMo2_0_0_,
        user0_.Password as Passwo3_0_0_,
        user0_.Name as Name4_0_0_,
        user0_.Dob as Dob5_0_0_,
        user0_.Address as Addres6_0_0_,
        user0_.PreviousAddress as Previo7_0_0_,
        user0_."address" as addres7_0_0_,
        datepart(year,
        user0_.dob) as formula0_0_ 
    FROM
        T_USER user0_ 
    WHERE
        user0_.UserName=@p0;
    @p0 = 'gavin' [Type: String (0:0:0)]
```

fredericDelaporte:
On a case sensitive db, but it should then cause it to fail on a case insensitive one, shouldn't it?

hazzik:
Not sure that I understood your last comment. 

It does not fail on case insensitive DB. 

I'll change how aliases are generated for columns. They should be built for "canonical" column names.

hazzik:
I reimplemented the changes. Now the aliases are consistent with the logic of column equality.

fredericDelaporte:
> Not sure that I understood your last comment.

Changing the mapping to:

```
<hibernate-mapping 
	xmlns="urn:nhibernate-mapping-2.2" 
	assembly="NHibernate.Test" 
	namespace="NHibernate.Test.Component.Basic">

	<class name="User" table="T_USER">
		<id name="UserName"/>
		<timestamp name="LastModified"/>
		<property name="Password" not-null="true" optimistic-lock="false"/>
		<component name="Person">
			<property name="Name" update="false" not-null="true"/>
			<property name="Dob" update="false" not-null="true"/>
			<property name="Address" column="`Address`"/>  <!-- quoted -->
			<property name="PreviousAddress" insert="false"/>
			<property name="Yob" formula="year(Dob)"/> <!-- case fixed -->
			<property name="CurrentAddress"
				column="`address`" <!-- quoted -->
				insert="false"
				update="false"/>
		</component>
	</class>
</hibernate-mapping>
```

Causes the corresponding tests to start failing on case insensitive databases. "Column names in each table must be unique."

Quoting my test causes it to start working with your change: columns get different alias suffixes numbers. Weird behavior. I guess that the point your last commit adresses.

hazzik:
>I guess that is the point your last commit addresses.

Now it's my only commit:) 

It's all because of how columns are deem to be equal

fredericDelaporte:
Ok, changes on SelectFragment reverted, and this starts fixing another trouble causing, I guess, a bad alias computing for SQL Server CE.

fredericDelaporte:
Ok, seen that:

		public string CanonicalName
		{
			get { return _quoted ? _name : _name.ToLowerInvariant(); }
		}

So now I better understand your change for having more consistent behavior when generating alias names.

Maybe worth adding a note in [SQL Quoted Identifier](http://nhibernate.info/doc/nhibernate-reference/mapping.html#mapping-quotedidentifiers).

hazzik:
Maybe... but I'm going to sleep now:)

PR #617, opened 2017-04-30
NH-3999 - quoted identifier effect on column name case sensitivity

fredericDelaporte:
[NH-3999](https://nhibernate.jira.com/browse/NH-3999) - documenting quoted identifier effect on column name case sensitivity.

### DIFF
--- a/doc/reference/modules/basic_mapping.xml
+++ b/doc/reference/modules/basic_mapping.xml
@@ -3114,6 +3114,11 @@
     ...
 </class>]]></programlisting>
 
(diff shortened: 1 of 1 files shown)


fredericDelaporte:
This change is more a sort of warning about quoted identifiers changing some behavior of NHibernate about case sensitivity.

Giving more details would render this section a bit cumbersome I think. (Like explaining it has some importance for schema generation, but may still work with inconsistent naming when the db is case insensitive, and whatever...)

PR #618, opened 2017-04-30
Fixing a test db setup bug in build menu, adding files for testing sq…

fredericDelaporte:
…l server ce.

### DIFF
--- a/ShowBuildMenu.bat
+++ b/ShowBuildMenu.bat
@@ -130,6 +130,7 @@ set LIB_FILES2=
 goto test-setup-generic
 
 :test-setup-generic
+set CFGNAME=
 set /p CFGNAME=Enter a name for your test configuration or press enter to use default name: 
 if /I "%CFGNAME%"=="" set CFGNAME=%CONFIG_NAME%-%PLATFORM%
 mkdir "%AVAILABLE_CONFIGURATIONS%\%CFGNAME%"
--- a/lib/teamcity/sqlServerCe/sqlServerCe_installation.txt
+++ b/lib/teamcity/sqlServerCe/sqlServerCe_installation.txt
@@ -1,0 +1,15 @@
+
+Installation steps for local testing:
+
+1.  Create a project referencing System.Data.SqlServerCe.dll
+2.  Put following code in it, changing database file name and path as appropriate:
+	using (var en = new SqlCeEngine("DataSource=\"C:\\SqlCeData\\NHibernate.sdf\""))
+	{
+		en.CreateDatabase();
+	}
+3.  Run it.
+4.  Ensure your test configuration file references your newly created SQL Server CE database file. By example:
+	<property name="connection.connection_string">
+		Data Source=C:\SqlCeData\NHibernate.sdf
+	</property>
+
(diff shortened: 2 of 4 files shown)


ShowBuildMenu.bat
```diff
@@ -130,6 +130,7 @@ set LIB_FILES2=
 goto test-setup-generic
 
 :test-setup-generic
+set CFGNAME=
 set /p CFGNAME=Enter a name for your test configuration or press enter to use default name:
```
fredericDelaporte: The first run is able of using default name if user does not input anything. But subsequent runs re-use the last set name, because `CFGNAME` remains defined in the shell till we close and relaunch it.

So I have added a `SET` for ensuring `CFGNAME` is empty before prompting user.

src/NHibernate.Config.Templates/SqlServerCe32.cfg.xml
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- 
+This template was written to work with NHibernate.Test.
+Copy the template to your NHibernate.Test project folder and rename it in hibernate.cfg.xml and change it 
+for your own use before compile tests in VisualStudio.
+-->
+<!-- This is the System.Data.dll provider for SQL Server -->
+<hibernate-configuration  xmlns="urn:nhibernate-configuration-2.2" >
+	<session-factory name="NHibernate.Test">
+		<property name="command_timeout">0</property>
```
fredericDelaporte: Taken from TeamCity build config, I have not tested it was mandatory.

lib/teamcity/sqlServerCe/sqlServerCe_installation.txt
```diff
@@ -0,0 +1,15 @@
+
+Installation steps for local testing:
+
+1.  Create a project referencing System.Data.SqlServerCe.dll
+2.  Put following code in it, changing database file name and path as appropriate:
+	using (var en = new SqlCeEngine("DataSource=\"C:\\SqlCeData\\NHibernate.sdf\""))
+	{
+		en.CreateDatabase();
+	}
+3.  Run it.
```
fredericDelaporte: Another way would be to add some special handling in the NHibernate.Test project depending on the dialect, but that would require adding the reference to sql server ce in it.

fredericDelaporte:
Since all those changes only impact local testing, I think a Jira is not needed.

src/NHibernate.Config.Templates/SqlServerCe32.cfg.xml
hazzik: It is. SqlServer CE does not support any timeout other than 0

lib/teamcity/sqlServerCe/sqlServerCe_installation.txt
hazzik: This is what TestDatabaseSetup does. But, it does not work properly for file-based DBs if you run the the tests from/with VS/R#.

fredericDelaporte: Indeed, I have taken it from there. So, another reason to not try putting that in the test project.

PR #619, opened 2017-04-30
NH-3997 - identity and Sql Server Ce: fix

fredericDelaporte:
[NH-3997](https://nhibernate.jira.com/browse/NH-3997) - identity and Sql Server Ce: fix releasing of connection between insert and select of identity.

This is a proposition of another fix to the Sql Server Ce trouble with identity.

Sql Server Ce supports identity, replacing it with HiLo because it was failing when used without a transaction is a breaking change forcing those using it (with transaction) to change their schema, which could be a serious hassle.

Studying what causes the failure without transaction, I think NHibernate has a more general bug concerning most db but probably mitigated by connection pooling: when the connection release mode is `AfterTransaction` and there is no transaction, the connection is released between the actual insert and the identity select. This causes the identity select command to potentially run with another connection. In such case, it cannot retrieve inserted identity.

The release of connection between actual insert and retrieval of identity is triggered by the closing of insert command. So as a "quick" workaround, I have delayed this closure to after identity retrieval.

I know Java Hibernate has made heavy changes on its identity handling and on its connection releasing, and they very likely handle that more "gracefully". A better fix would probably to sync with Hibernate, but it looks heavy.

If agreeing on this fix, I will reword NH-3997 in Jira.

### DIFF
--- a/src/NHibernate/Dialect/MsSqlCeDialect.cs
+++ b/src/NHibernate/Dialect/MsSqlCeDialect.cs
@@ -125,8 +125,6 @@ namespace NHibernate.Dialect
 			get { return false; }
 		}
 
-		public override System.Type NativeIdentifierGeneratorClass => typeof(TableHiLoGenerator);
-
 		public override bool SupportsCircularCascadeDeleteConstraints => false;
 
 		public override IDataBaseSchema GetDataBaseSchema(DbConnection connection)
--- a/src/NHibernate/Id/Insert/AbstractSelectingDelegate.cs
+++ b/src/NHibernate/Id/Insert/AbstractSelectingDelegate.cs
@@ -28,55 +28,59 @@ namespace NHibernate.Id.Insert
 
 		public object PerformInsert(SqlCommandInfo insertSQL, ISessionImplementor session, IBinder binder)
 		{
+			DbCommand insert = null;
 			try
 			{
-				// prepare and execute the insert
-				var insert = session.Batcher.PrepareCommand(insertSQL.CommandType, insertSQL.Text, insertSQL.ParameterTypes);
 				try
 				{
+					// prepare and execute the insert
+					insert = session.Batcher.PrepareCommand(insertSQL.CommandType, insertSQL.Text, insertSQL.ParameterTypes);
 					binder.BindValues(insert);
 					session.Batcher.ExecuteNonQuery(insert);
 				}
-				finally
+				catch (DbException sqle)
 				{
-					session.Batcher.CloseCommand(insert, null);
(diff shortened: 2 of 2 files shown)


hazzik:
I did not look into the patch yet. But. Identity is still supported: i
changed only "native" generator (the one, I believe, shall no one in
reality use) to be hilo.

fredericDelaporte:
You sleep that few hours? (Or maybe should you not let your mails be notified on your phone if sleeping with it.)

But yes, you are right, your change can still be breaking but is easy to fix, by changing native for identity.

I think we still have a more general issue to fix, as written in this PR, and which by the way fix SQL Server Ce tests, thus rendering the native generator change unneeded. I am waiting for builds to verify my change does not have unexpected consequences.

hazzik:
This issue have own JIRA (or 2) in fact.

hazzik:
> You sleep that few hours?

~640 kb~ 6 hours shall be enough for everyone

fredericDelaporte:
I have not felt the time passing then.

So, we have [NH-926](https://nhibernate.jira.com/browse/NH-926), [NH-1917](https://nhibernate.jira.com/browse/NH-1917) and [NH-2145](https://nhibernate.jira.com/browse/NH-2145) which looks related to it.

NH-2145 seems to have been solved by retrieving the identity with the insert (looking at the associated commit). That should be the preferred way for any dialect supporting it. I do not know if it is done.

hazzik:
You have 237 new tests failing because SqlServer CE does not support insert empty columns (NH-3761). So, I would really prefer the "native" generator to be "hilo" while still properly fixing NH-926/NH-2145. NH-926 thou once was closed with no explanation. 

NH-1917 seems to be unrelated.

fredericDelaporte:
Agreed. I need to change that whole PR and commit message for matching those other issues.

fredericDelaporte:
NH-926 was closed because fixed by commit 3f304edaa1f2ee6ff36fb1eaf4296f5aff120ccb. But some later port of Hibernate changes lost this fix. (27166242d2679fafde3f77b5da0767f211f392e3 & 3ce5929a9730ca67af1c9def73c32a7eeb19f37f)

Lack of covering test I guess.

fredericDelaporte:
Will reopen another one.

hazzik:
>NH-926 was closed because fixed by commit 3f304ed.

After the "fix" Sergej Koščejev wrote: "Looks like it wasn't fixed.", and reopened the issue. In a few minutes ayende explained why it wasn't fixed (or wasn't; I really don't understand what did he say there)

And then 4 days later the issue was closed. 

---

Maybe you are right that the changes had lost in porting effort.

PR #620, opened 2017-04-30
NH-2145 - identity: releasing of connection between insert and identity retrieval

fredericDelaporte:
[NH-2145](https://nhibernate.jira.com/browse/NH-2145) - [NH-926](https://nhibernate.jira.com/browse/NH-926) (re) - [NH-3600](https://nhibernate.jira.com/browse/NH-3600) - identity: test cases for releasing of connection between insert and identity retrieval.

Will wait the test failures for checking those new tests seems to do their job. (`NonTransacted` should failed on server supporting identity without supporting select with insert, and pooling version should failed for same servers if they allow specifying pooling. Unfortunately, it does not look like we have such servers excepted SQL Server CE (which does not support (un)setting pooling).)

### DIFF
--- a/src/NHibernate/AdoNet/ConnectionManager.cs
+++ b/src/NHibernate/AdoNet/ConnectionManager.cs
@@ -3,7 +3,6 @@ using System.Data;
 using System.Data.Common;
 using System.Runtime.Serialization;
 using System.Security;
-using System.Security.Permissions;
 
 using NHibernate.Engine;
 
@@ -13,7 +12,7 @@ namespace NHibernate.AdoNet
 	/// Manages the database connection and transaction for an <see cref="ISession" />.
 	/// </summary>
 	/// <remarks>
-	/// This class corresponds to ConnectionManager and JDBCContext in Hibernate,
+	/// This class corresponds to LogicalConnectionImplementor and JdbcCoordinator in Hibernate,
 	/// combined.
 	/// </remarks>
 	[Serializable]
@@ -44,7 +43,7 @@ namespace NHibernate.AdoNet
 		private readonly IInterceptor interceptor;
 
 		[NonSerialized]
-		private bool isFlushing;
+		private bool _releasesEnabled = true;
 
 		private bool flushingFromDtcTransaction;
 
@@ -227,9 +226,9 @@ namespace NHibernate.AdoNet
 		{
 			if (IsAggressiveRelease)
 			{
-				if (isFlushing)
+				if (!_releasesEnabled)
 				{
-					log.Debug("skipping aggressive-release due to flush cycle");
+					log.Debug("skipping aggressive-release due to manual disabling");
 				}
 				else if (batcher.HasOpenResources)
 				{
@@ -259,16 +258,31 @@ namespace NHibernate.AdoNet
 			}
 		}
 
+		[NonSerialized]
+		private int _flushDepth;
+
 		public void FlushBeginning()
 		{
-			log.Debug("registering flush begin");
-			isFlushing = true;
+			if (_flushDepth == 0)
+			{
+				log.Debug("registering flush begin");
+				_releasesEnabled = false;
+			}
+			_flushDepth++;
 		}
 
 		public void FlushEnding()
 		{
-			log.Debug("registering flush end");
-			isFlushing = false;
+			_flushDepth--;
+			if (_flushDepth < 0)
+			{
+				throw new HibernateException("Mismatched flush handling");
+			}
+			if (_flushDepth == 0)
+			{
+				_releasesEnabled = true;
+				log.Debug("registering flush end");
+			}
 			AfterStatement();
 		}
 
--- a/src/NHibernate/Id/Insert/AbstractSelectingDelegate.cs
+++ b/src/NHibernate/Id/Insert/AbstractSelectingDelegate.cs
@@ -26,57 +26,69 @@ namespace NHibernate.Id.Insert
 
 		public abstract IdentifierGeneratingInsert PrepareIdentifierGeneratingInsert();
 
-		public object PerformInsert(SqlCommandInfo insertSQL, ISessionImplementor session, IBinder binder)
+		public object PerformInsert(SqlCommandInfo insertSql, ISessionImplementor session, IBinder binder)
 		{
+			// NH-2145: Prevent connection releases between insert and select when we cannot perform
+			// them as a single statement. Retrieving id most of the time relies on using the same connection.
+			session.ConnectionManager.FlushBeginning();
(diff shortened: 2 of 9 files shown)


hazzik:
Fails as expected. There are tons of other tests failing with the same error on SqlServer CE.

fredericDelaporte:
Only 4 less failing (comparing [this](https://teamcity.jetbrains.com/viewLog.html?buildId=1058161&buildTypeId=bt1175) and [this one](https://teamcity.jetbrains.com/viewLog.html?buildId=1058052&buildTypeId=bt1175) from #616).

#391 would remove more failures by the way, as I have completed the prerequisites of BulkManipulation fixture while fixing LinqBulkManipulation added in it.

fredericDelaporte:
Otherwise looks good to me, I intend merging it soon.

fredericDelaporte:
From #619 

>The release of connection between actual insert and retrieval of identity is triggered by the closing of insert command. So as a "quick" workaround, I have delayed this closure to after identity retrieval.
>
> I know Java Hibernate has made heavy changes on its identity handling and on its connection releasing, and they very likely handle that more "gracefully". A better fix would probably to sync with Hibernate, but it looks heavy.

Yes it would be better to port Hibernate changes and check they fix the issue. But it is still hard right now for me to estimate how long it would take me. This PR is here for fixing this issue even if the port does not get done for 5.0.

fredericDelaporte:
While looking at what to do next for realigning transaction management on Hibernate, I find nothing in Hibernate code for avoiding this NH-2145 bug. It looks like Hibernate `JdbcCoordinator` may release the connection between insert and select as do NHibernate `ConnectionManager`.

But in Hibernate case, it is way more unlikely to happen: its `afterExecuteStatement` releases connection only with `ConnectionReleaseMode.AfterStatement`, while NHibernate does it with `ConnectionReleaseMode.AfterTransaction` if there is no ongoing transaction.

So this bug needs its own fix, porting Hibernate code will not be enough. 

To do a cleaner fix, I intend to port current Hibernate `JdbcCoordinator.flushBegining/flushEnding` implementation which uses a depth counter instead of only a boolean as `ConnectionManager.FlushBeginning` does, then use it for surrounding the insert/select. After all, this insert/select is a partial flush.

src/NHibernate/AdoNet/ConnectionManager.cs
```diff
@@ -13,7 +12,7 @@ namespace NHibernate.AdoNet
 	/// Manages the database connection and transaction for an <see cref="ISession" />.
 	/// </summary>
 	/// <remarks>
-	/// This class corresponds to ConnectionManager and JDBCContext in Hibernate,
+	/// This class corresponds to LogicalConnectionImplementor and JdbcCoordinator in Hibernate,
```
fredericDelaporte: Synchronizing with current names on Hibernate side.

src/NHibernate/AdoNet/ConnectionManager.cs
```diff
@@ -44,7 +43,7 @@ public interface Callback
 		private readonly IInterceptor interceptor;
 
 		[NonSerialized]
-		private bool isFlushing;
+		private bool _releasesEnabled = true;
```
fredericDelaporte: On Hibernate side, two additional public methods are there: `enableReleases` and `disableReleases`. They seem unused and I have not port them. They weaken the class in my opinion, since calling them set directly `releasesEnabled` without taking care of on-going flushes.

src/NHibernate/AdoNet/ConnectionManager.cs
```diff
@@ -259,16 +258,31 @@ private void AggressiveRelease()
...
+		[NonSerialized]
+		private int _flushDepth;
+
 		public void FlushBeginning()
 		{
-			log.Debug("registering flush begin");
-			isFlushing = true;
+			if (_flushDepth == 0)
+			{
+				log.Debug("registering flush begin");
```
fredericDelaporte: No log on Hibernate side for this. For minimizing logs, I issue it only when the releases flag is toggled. If we would like to use it for tracking begin/end mismatches, it then needs to be moved.

src/NHibernate/Id/Insert/AbstractSelectingDelegate.cs
```diff
@@ -26,57 +26,67 @@ protected internal AbstractSelectingDelegate(IPostInsertIdentityPersister persis
...
+					{
+						session.Batcher.CloseCommand(insert, null);
+					}
 				}
-				finally
+				catch (DbException sqle)
 				{
-					session.Batcher.CloseCommand(insert, null);
+					throw ADOExceptionHelper.Convert(session.Factory.SQLExceptionConverter, sqle,
+					                                 "could not insert: " + persister.GetInfoString(), insertSql.Text);
```
fredericDelaporte: Yes, I have forced myself putting back those spaces in place after they have been changed to tabs.

src/NHibernate/Id/Insert/AbstractSelectingDelegate.cs
```diff
@@ -26,57 +26,67 @@ protected internal AbstractSelectingDelegate(IPostInsertIdentityPersister persis
 
 		public abstract IdentifierGeneratingInsert PrepareIdentifierGeneratingInsert();
 
-		public object PerformInsert(SqlCommandInfo insertSQL, ISessionImplementor session, IBinder binder)
+		public object PerformInsert(SqlCommandInfo insertSql, ISessionImplementor session, IBinder binder)
 		{
+			session.ConnectionManager.FlushBeginning();
```
fredericDelaporte: Significant changes are the addition of lines 31 to 33 and 87 to 89. Other changes are indentation or renaming.

hazzik:
>But in Hibernate case, it is way more unlikely to happen: its afterExecuteStatement releases connection only with ConnectionReleaseMode.AfterStatement, while NHibernate does it with ConnectionReleaseMode.AfterTransaction if there is no ongoing transaction.

This is what I meant by 

> it seems that we need to fix (implement) "AfterStatement" connection release mode instead.

fredericDelaporte:
Added a comment about the call to `FlushBeginning` in id generation.

fredericDelaporte:
About:
> What does NHibernate currently as AfterTransaction mode looks sensible by the way: when there are no transaction, every statement is indeed a short transaction.

I have seen today it is [documented](http://nhibernate.info/doc/nhibernate-reference/transactions.html#transactions-connection-release):
>Note that with `ConnectionReleaseMode.AfterTransaction`, if a session is considered to be in auto-commit mode (i.e. no transaction was started) connections will be released after every operation.

PR #621, opened 2017-05-02
NH-4004 - Restrict tests running on SqlServer CE

hazzik:
[NH-4004](https://nhibernate.jira.com/browse/NH-4004)


src/NHibernate.Test/ReadOnly/TextHolder.cs
```diff
@@ -10,7 +10,7 @@ public class TextHolder
 		/// </summary>
 		public static bool SupportedForDialect(Dialect.Dialect dialect)
 		{
-			return !(dialect is FirebirdDialect || dialect is Oracle8iDialect);
+			return !(dialect is FirebirdDialect || dialect is Oracle8iDialect || dialect is MsSqlCeDialect);
```
fredericDelaporte: Here too, I would rather have a `SupportsTextBlobType` feature. This blacklisting may miss a bunch of supported db we are currently not testing. And when those `text`/ `ntext` types will start being dropped from newer version of SQL Server, the blacklisting will start to be cumbersome. (Granted, this looks like it may never happen, `ntext` and `text` are in the ["will be removed in an undetermined future version"](https://docs.microsoft.com/en-us/sql/database-engine/deprecated-database-engine-features-in-sql-server-2016#features-not-supported-in-a-future-version-of-sql-server) list since SQL 2005 till SQL 2016. Still waiting to see them in the "will be removed in next version list".)

By the way, character types looks like being another muddy place a bit alike dates types^^. (Even worse indeed.)

fredericDelaporte:
I think it is a very welcome change overall.

src/NHibernate.Test/NHSpecificTest/NH1521/Fixture.cs
```diff
@@ -11,7 +11,8 @@ public class Fixture
 	{
 		private static void CheckDialect(Configuration configuration)
 		{
-			if (!configuration.Properties[Environment.Dialect].Contains("MsSql"))
+			var dialect = configuration.Properties[Environment.Dialect];
+			if (!dialect.Contains("MsSql") || dialect.Contains("MsSqlCe"))
```
hazzik: `Dialect` is not available here. This is a dialect specific test (I think it's MsSql2005/8Dialect), but unfortunately MsSqlCe dialects starts from the same string :)

fredericDelaporte: Granted I have not checked it was inheriting from `TestCase`. But well, grabbing the dialect object looks simple: `NHibernate.Dialect.Dialect.GetDialect(configuration.Properties)`.

src/NHibernate.Test/NHSpecificTest/NH1635/Fixture.cs
```diff
@@ -7,6 +8,11 @@ namespace NHibernate.Test.NHSpecificTest.NH1635
 	[TestFixture]
 	public class Fixture : BugTestCase
 	{
+		protected override bool AppliesTo(Dialect.Dialect dialect)
+		{
+			return !(dialect is MsSqlCeDialect);
```
hazzik: I probably did not add this property yet when I was making change to this file

src/NHibernate.Test/Hql/Ast/BulkManipulation.cs
```diff
@@ -1,28 +1,21 @@
...
 using NHibernate.Hql.Ast.ANTLR;
-using NHibernate.Id;
-using NHibernate.Persister.Entity;
+using NHibernate.Linq;
 using NUnit.Framework;
 
 namespace NHibernate.Test.Hql.Ast
 {
 	[TestFixture]
 	public class BulkManipulation : BaseFixture
```
fredericDelaporte: Heavily rewritten, as it was already done in #391: proper setup&teardown allowing to handle the "no temp table" case in which MsSqlCe falls.

src/NHibernate/Dialect/MsSqlCeDialect.cs
```diff
@@ -339,6 +339,9 @@ public override long TimestampResolutionInTicks
 		/// </summary>
 		public override bool SupportsPoolingParameter => false;
 
+		/// <inheritdoc/>
+		public override bool SupportsScalarSubSelects => false;
```
fredericDelaporte: Ms SQL CE supports only sub-queries defining a set, like with in `where exists(sub-query)`, `join (sub-query)`, `from (sub-query)`, `in (sub-query)`. Sub-queries to be handled as a scalar value (single column & single row) are not supported. So things like `select (sub-query) as colName`, `where (sub-query) = someValue`, `order by (sub-query)`, ..., are not supported.

src/NHibernate.Test/Hql/Ast/HqlFixture.cs
```diff
@@ -74,188 +74,95 @@ public void OrderByPropertiesImplicitlySpecifiedInTheSelect()
 			{
 				s.CreateQuery("select distinct z from Animal a join a.zoo as z order by z.name").List();
 			}
-		} 
+		}
 
 		[Test]
 		public void CaseClauseInSelect()
```
fredericDelaporte: HqlFixture rewritten a bit like bul manipulation fixture, for the same reasons.

src/NHibernate.Test/Hql/Ast/WithClauseFixture.cs
```diff
@@ -7,115 +6,88 @@ namespace NHibernate.Test.Hql.Ast
 	[TestFixture]
 	public class WithClauseFixture : BaseFixture
```
fredericDelaporte: Rewritten a bit like bul manipulation fixture, for the same reasons.

src/NHibernate.Test/Legacy/SQLLoaderTest.cs
```diff
@@ -333,8 +333,8 @@ public void EscapedODBC()
 		[Test]
 		public void DoubleAliasing()
 		{
-			if (Dialect is MySQLDialect) return;
-			if (Dialect is FirebirdDialect) return; // See comment below
```
fredericDelaporte: Appears to work for them, probably since some time.

src/NHibernate.Test/NHSpecificTest/NH1812/Fixture.cs
```diff
@@ -1,20 +1,26 @@
 using System;
-using System.Collections.Generic;
 using NUnit.Framework;
 
 namespace NHibernate.Test.NHSpecificTest.NH1812
 {
-	public class AstBugBase : BugTestCase
+	public class AstBug : BugTestCase
```
fredericDelaporte: This test was originally having:

-  `AstBugBase` class without a `Fixture` attribute, not meant to be run (should have been further marked `abstract` for guaranteeing this).
- `AstBug`, just deriving from `AstBugBase` with a `Fixture` attribute, meant to demonstrate the bug.
- `ItWorksWithClassicParser`, deriving from `AstBugBase` with a `Fixture` attribute and alter the configuration for using the "classic parser". Meant to demonstrate this was a regression.

The later has been removed, rendering the inheritance scheme moot. But that was not cleaned. And NUnit changes caused the "base" class to be run too.

I have cleaned up that.

src/NHibernate/Dialect/MsSqlCeDialect.cs
```diff
@@ -187,10 +188,15 @@ protected virtual void RegisterFunctions()
...
 			RegisterFunction("trim", new AnsiTrimEmulationFunction());
+			RegisterFunction("iif", new SQLFunctionTemplate(null, "case when ?1 then ?2 else ?3 end"));
 
 			RegisterFunction("concat", new VarArgsSQLFunction(NHibernateUtil.String, "(", "+", ")"));
+			RegisterFunction("mod", new SQLFunctionTemplate(NHibernateUtil.Int32, "((?1) % (?2))"));
 
 			RegisterFunction("round", new StandardSQLFunction("round"));
+
+			RegisterFunction("bit_length", new SQLFunctionTemplate(NHibernateUtil.Int32, "datalength(?1) * 8"));
+			RegisterFunction("extract", new SQLFunctionTemplate(NHibernateUtil.Int32, "datepart(?1, ?3)"));
```
fredericDelaporte: All these added function were already tested (and previously failing) in `HQLFunctions` fixture.

fredericDelaporte:
Since I have made notable additions on this PR, I would like to have your feedback @hazzik , rather than self approving then merging those additions.

And maybe doing a rebase before would be better.

hazzik:
Can we have BulkManipulation  refactoring as a separate request? It's not clear what's going on there.

fredericDelaporte:
I will not be clearer with a separate request. It will just cause more overhead to tackle those pile of badly written tests. First it will force to wholly disable them for SqlServer Ce if we aiming at eliminating tests failure with this current PR. Then it will cause re-doing the work in a separate request for allowing testing DML operations for SqlServer Ce too.

hazzik:
So, refactoring PR first. Then this PR. I think this way it would work.

fredericDelaporte:
Sorry, I do not see enough added value in proceeding that way. All NHibernate subjects here takes way to much time to tackle due in part to what I see as overdue nitpicking. No wonder NHibernate project is unable to foster collaboration and is so much lacking active contributors.

hazzik:
https://en.wikipedia.org/wiki/Correlation_does_not_imply_causation

fredericDelaporte:
Everything can cause errors, we are just humans. But that is not a trouble, errors are recoverable.

For me there is a trouble with the pace at which we can contribute to NHibernate. I find it sluggish. Error rate is maybe very low, but at an excessive contributing time cost in my opinion. And as showcased by the VS2017 csproj format migration, there are still errors. But well, that can be fixed (and already is for this example).

At some point trying hard to have a very low error rate can end up costing way more time than letting more errors happen then fixing them.

About the refactoring itself in a separated PR, I would have agreed if it was about a feature. But here this is just fixing tests, and it has its own commit. A PR for refactoring one test fixture as a prerequisite to actually fixing a trouble, instead of doing it all in one PR, is overkill in my opinion.

The change is not just about `try catch/using`, these were just a "bonus". With the way you seems to consider such changes, I should not have done them.
The main point is the isolation of data setup and tear-down in their NUnit ad-hoc methods. Instead of letting them embedded in each test, causing tear-down to be skipped in case of test failure. And by the way wholly consolidating those setup and tear-down logic in one single place, the previous `TestData` class logic covering only part of it. Without that the special tear-down logic required for Sql Server Ce would have required to be duplicated in many tests. 
Is JUnit without counterparts for these methods too?

Another trouble with those tests, which I have not tackled, is their tendency to test many different things in each test, causing them to get some "if supported then additionally test that" logic, which looks bad. Better have many more tests whit some of them fully ignored because fully unsupported, rather than having fewer and almost all successful (rather than ignored) tests, but having ignored some parts of them  without emitting much clues about that from the test framework perspective.
That is why I wrote those tests are badly written, and it does not look to me those points are due to Java constraints.

hazzik:
>and it has its own commit.

I don't see this commit. I would be OK with the separate commit for the refactoring and a separate commit for disabling irrelevant tests.

>But here this is just fixing tests

Tests are crucial to software, they are not "just tests". I've seen many times how such "refactorings" made test incorrect and because these errors were accompanied by feature changes they went unnoticed for a long time.  

What I'm trying to say: bigger commits require longer time to review. Changes which Github decides to collapse because they are too big to be displayed require the infinite amount of time.

Didn't comment anything else because it's almost bed time and I don't want to spend the night arguing.

hazzik:
I checked some of the changes of BulkManipulation and I completely disagree with them. A couple of tests have completely changed semantic now. Also, for some of the tests, the test routine has moved to the setup, which made these tests green but pointless. From the TC log, I can see that almost only "negative" tests (the tests checking that incorrect usage will throw) are run for Sql CE. 

Now I think that this is our best option:

> to wholly disable them for SqlServer Ce

fredericDelaporte:
> A couple of tests have completely changed semantic now. Also, for some of the tests, the test routine has moved to the setup, which made these tests green but pointless.

These are serious statements. I do not believe to have made such mistakes. Please back such statements with pointers to at least some sample of these tests you consider affected by such mistakes.

Many tests are disabled for sql server ce, specifically all the ones relying on multiple tables operation (join mapping). But other more usual cases (class mapped to a single table) do work and are tested.

Anyway, if you rather "fix" that by disabling dml tests wholly for sql server ce, drop my commit and go ahead.

hazzik:
InsertWithGeneratedId, InsertWithGeneratedVersionAndId, InsertWithGeneratedTimestampVersion and probably some other

fredericDelaporte:
Ok, so that is the filter for avoiding the old Firebird "instable cursor" bug that you consider as a semantic change I guess. Not untrue, but the fact that the query was filtered or not was not IMO the subject of the test, and was an external issue, not a one NHibernate could fix. Anyway this Firebird workaround could be no more needed with the upgrade to 3.0 done some PR ago, so it may be rewritten without it. This workaround was coming from the changes already done in strongly typed DML PR (as of most of the other changes in this file).

Anyway, enough with that subject, I probably going to drop that commit. (So force update will be required, so will rebase by the way, but due to having limited connectivity currently I am a bit stuck with VS 2017.2 for now and will have to wait.)
I have checked gain which tests remain for SQL Server Ce, and there is at least one for each operation (insert/update/delete) actually performing the operation, not simply testing an error case. So just disabling them for SQL Server CE is a bit unfortunate, unless we already consider sun-setting its support (as I was suggested some time ago).
(Granted, this is a bit obfuscated by the fact that most entities in those tests are not simple ones, so few tests actually check the basic and probably most usual use cases.)

fredericDelaporte:
> Do you want them to be reverted too?

Now done anyway.

hazzik:
Can we focus on AsyncGenerator and Linq DML instead?

fredericDelaporte:
No feed back on Linq DML, nothing more te be done on my side without any responses from others.
Asyncgenerator: well that is precisely for this one that I am doing all those tests cleanup: that should remove some of those async failing tests due to their sync counterpart failing too.

hazzik:
Thanks. I've recombined the commits into 4.

PR #623, opened 2017-05-04
[WIP] - NH-2088 - Preliminary work required to generate async code

hazzik:
It does not compile because the code expects some methods to be already generated.

### DIFF
--- a/src/NHibernate/Event/Default/DefaultLockEventListener.cs
+++ b/src/NHibernate/Event/Default/DefaultLockEventListener.cs
@@ -1,6 +1,7 @@
 using System;
 using NHibernate.Engine;
 using NHibernate.Persister.Entity;
+using NHibernate.Proxy;
 
 namespace NHibernate.Event.Default
 {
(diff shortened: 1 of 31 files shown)


src/NHibernate/Event/Default/DefaultLockEventListener.cs
```diff
@@ -42,7 +43,7 @@ public virtual void OnLock(LockEvent @event)
 			{
 				IEntityPersister persister = source.GetEntityPersister(@event.EntityName, entity);
 				object id = persister.GetIdentifier(entity);
-				if (!ForeignKeys.IsNotTransient(@event.EntityName, entity, false, source))
+				if (ForeignKeys.IsTransientFast(@event.EntityName, entity, source).GetValueOrDefault())
```
hazzik: `IsNotTransient` does additional check on `IsProxy` and `IsEntryFor`. Both of these checks are performed above at lines L31 and L41-L42.

hazzik:
Replaced by #677

PR #624, opened 2017-05-04
NH-4003 - porting ISessionCreationOptions

fredericDelaporte:
[NH-4003](https://nhibernate.jira.com/browse/NH-4003) - Refactor session constructor

I consider this as a prerequisite for porting the current transaction handling from Hibernate. This handling especially relies on the `ISharedSessionBuilder`, introduced by this PR.

I have dropped the obsolete parameter `InterceptorsBeforeTransactionCompletionIgnoreExceptions` in the process.

The `GetChildSession` introduced in 5.0 by the removal of entity mode switching is removed in favor of Hibernate `ISession.WithSessionOptions()` which yield a builder of session, allowing to build new sessions taking a part of their stuff (like the connection manager currently) from the original session. These new sessions are not closed or flushed from the original session.

I have reintroduced as obsolete the old `ISession.GetSession()`, for hinting at using `ISession.WithSessionOptions()`.

All that by the way obsoletes [NH-3985](https://nhibernate.jira.com/browse/NH-3985) (already resolved, #600, about disposed child session being returned) and [NH-3986](https://nhibernate.jira.com/browse/NH-3986) (Enable multiple child sessions per root session: the new semantic allows this, though there are no more actually children, the "parent" has no knowledge of them currently).

Some stuff has moved around, Hibernate having put many more things in the AbstractSessionImpl.

### DIFF
--- a/src/NHibernate/Cfg/MappingSchema/Hbm.generated.cs
+++ b/src/NHibernate/Cfg/MappingSchema/Hbm.generated.cs
@@ -4294,14 +4294,18 @@ namespace NHibernate.Cfg.MappingSchema {
         /// <remarks/>
         [System.Xml.Serialization.XmlEnumAttribute("auto")]
         Auto,
-        
+
         /// <remarks/>
-        [System.Xml.Serialization.XmlEnumAttribute("never")]
-        Never,
-        
+        [System.Xml.Serialization.XmlEnumAttribute("manual")]
+        Manual,
+
         /// <remarks/>
         [System.Xml.Serialization.XmlEnumAttribute("always")]
         Always,
+
+        /// <remarks/>
+        [System.Xml.Serialization.XmlEnumAttribute("never")]
+        Never,
     }
     
     /// <remarks/>
--- a/src/NHibernate/Impl/SessionFactoryImpl.cs
+++ b/src/NHibernate/Impl/SessionFactoryImpl.cs
@@ -66,11 +66,11 @@ namespace NHibernate.Impl
 	/// , but also highly concurrent.  Synchronization must be used extremely sparingly.
 	/// </para>
 	/// </remarks>
-	/// <seealso cref="NHibernate.Connection.IConnectionProvider"/>
-	/// <seealso cref="NHibernate.ISession"/>
-	/// <seealso cref="NHibernate.Hql.IQueryTranslator"/>
-	/// <seealso cref="NHibernate.Persister.Entity.IEntityPersister"/>
-	/// <seealso cref="NHibernate.Persister.Collection.ICollectionPersister"/>
+	/// <seealso cref="IConnectionProvider"/>
+	/// <seealso cref="ISession"/>
+	/// <seealso cref="IQueryTranslator"/>
+	/// <seealso cref="IEntityPersister"/>
+	/// <seealso cref="ICollectionPersister"/>
 	[Serializable]
 	public sealed class SessionFactoryImpl : ISessionFactoryImplementor, IObjectReference
 	{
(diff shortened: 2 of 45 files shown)


src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1270,5 +1267,144 @@ public string Uuid
 		}
 
 		#endregion
+
+		// NH specific: implementing return type covariance with interface is a mess in .Net.
+		internal class SessionBuilderImpl : SessionBuilderImpl<ISessionBuilder>, ISessionBuilder
+		{
+			public SessionBuilderImpl(SessionFactoryImpl sessionFactory) : base(sessionFactory)
+			{
+				SetSelf(this);
```
fredericDelaporte: Not found another a better way... See below its usage.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1270,5 +1267,144 @@ public string Uuid
...
+				}
+
+				return session;
+			}
+
+			public virtual T Interceptor(IInterceptor interceptor)
+			{
+				// NH different implementation: Hibernate accepts null.
+				_interceptor = interceptor ?? throw new ArgumentNullException(nameof(interceptor));
+				return _this;
```
fredericDelaporte: Here lies the trouble. The compiler refuses the cast `(T)this`, although `T : ISessionBuilder<T>` and `SessionBuilderImpl<T> : ISessionBuilder<T>`. So I have used that `_this` hack.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1270,5 +1267,144 @@ public string Uuid
...
+					_interceptor.SetSession(session);
+				}
+
+				return session;
+			}
+
+			public virtual T Interceptor(IInterceptor interceptor)
+			{
+				// NH different implementation: Hibernate accepts null.
+				_interceptor = interceptor ?? throw new ArgumentNullException(nameof(interceptor));
```
fredericDelaporte: Makes happy some tests. Not sure we should keep this. Allowing setting null would allow for reverting a previously set session interceptor and fall-backing again on factory interceptor. Now, since builder are cheap to build, that does not look important.

Note that on Hibernate side, the `NoInterceptor` causes fallback to factory interceptor, instead of configuring a "no-op" interceptor. That does not sound good to me, the naming does not match well.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1270,5 +1267,144 @@ public string Uuid
...
+				return _this;
+			}
+
+			public virtual T NoInterceptor()
+			{
+				_interceptor = EmptyInterceptor.Instance;
```
fredericDelaporte: Same on Hibernate side, the difference with NHibernate is in `GetInterceptor()` above.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -40,12 +40,7 @@ public sealed class SessionImpl : AbstractSessionImpl, IEventSource, ISerializab
 	{
 		private static readonly IInternalLogger log = LoggerProvider.LoggerFor(typeof(SessionImpl));
 
-		private readonly long timestamp;
-
 		private CacheMode cacheMode = CacheMode.Normal;
-		private FlushMode flushMode = FlushMode.Auto;
-
-		private readonly IInterceptor interceptor;
```
fredericDelaporte: Fields moved to base class.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -73,19 +68,11 @@ public sealed class SessionImpl : AbstractSessionImpl, IEventSource, ISerializab
 		private readonly StatefulPersistenceContext persistenceContext;
 
 		[NonSerialized]
-		private readonly SessionImpl rootSession;
-
-		[NonSerialized]
-		ISession _childSession;
-
-		[NonSerialized]
-		private readonly bool flushBeforeCompletionEnabled;
```
fredericDelaporte: Fields removed.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -73,19 +68,11 @@ public sealed class SessionImpl : AbstractSessionImpl, IEventSource, ISerializab
...
-		[NonSerialized]
 		private readonly bool autoCloseSessionEnabled;
 		[NonSerialized]
-		private readonly bool ignoreExceptionBeforeTransactionCompletion;
-		[NonSerialized]
 		private readonly ConnectionReleaseMode connectionReleaseMode;
+		[NonSerialized]
+		private readonly bool _transactionCoordinatorShared;
```
fredericDelaporte: NH addition, I expect this to be no more needed once transaction handling will be ported.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -140,15 +126,19 @@ void ISerializable.GetObjectData(SerializationInfo info, StreamingContext contex
 			{
 				throw new InvalidOperationException("Cannot serialize a Session while connected");
 			}
+			if (_transactionCoordinatorShared)
+			{
+				throw new InvalidOperationException("Cannot serialize a Session sharing its transaction coordinator");
```
fredericDelaporte: After deserialization, the shared thing (currently connection manager) would be no more shared.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1031,9 +946,9 @@ public override string BestGuessEntityName(object entity)
 				}
 				if (FieldInterceptionHelper.IsInstrumented(entity))
 				{
-					// NH: support of field-interceptor-proxy
-					IFieldInterceptor interceptor = FieldInterceptionHelper.ExtractFieldInterceptor(entity);
-					return interceptor.EntityName;
+					// NH: support of field-Interceptor-proxy
+					IFieldInterceptor Interceptor = FieldInterceptionHelper.ExtractFieldInterceptor(entity);
+					return Interceptor.EntityName;
```
fredericDelaporte: Bad replace all, fixed locally, not yet pushed. (Fixed in comments/string too.)

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -2206,36 +2124,21 @@ public ISession SetBatchSize(int batchSize)
...
+		[Obsolete("Please use SessionWithOptions instead.")]
+		public ISession GetSession(EntityMode entityMode)
 		{
-			get { return interceptor; }
+			return SessionWithOptions()
+				.Connection()
+				.ConnectionReleaseMode()
+				.FlushMode()
+				.Interceptor()
+				.OpenSession();
```
fredericDelaporte: Have to do all that with the new method if you want to be as close as possible... But we have no covering tests for that indeed: in the tests, I have just call `Connection()`, and that is enough for all of them.

src/NHibernate/Cfg/MappingSchema/Hbm.generated.cs
```diff
@@ -4294,14 +4294,18 @@ public enum HbmFlushMode {
...
+        [System.Xml.Serialization.XmlEnumAttribute("manual")]
+        Manual,
+
         /// <remarks/>
         [System.Xml.Serialization.XmlEnumAttribute("always")]
         Always,
+
+        /// <remarks/>
+        [System.Xml.Serialization.XmlEnumAttribute("never")]
+        Never,
```
hazzik: Shall be `Never = Manual,` (as in another place)

hazzik:
It feels like some of the methods need to be CSharpified (eg. replaced with properties).

hazzik:
LGTM

src/NHibernate/Cfg/MappingSchema/Hbm.generated.cs
fredericDelaporte: Semantically cleaner, but not sure the generator from xsd would do that, it has no mean to know it should (or I do not know how to tell it in xsd).

fredericDelaporte:
> It feels like some of the methods need to be CSharpified (eg. replaced with properties).

At first I wanted to do that for `ISessionCreationOptions`, but that was then conflicting with `ISessionBuilder`. Types implementing the former implements the later too. I would then have to either do some additional renaming on ISessionBuilder (which uses a property like naming for methods) and then deviate from Hibernate public API (which I would rather avoid), or implement explicitly `ISessionCreationOptions` or artificially find some different names for its members (less a trouble than for `ISessionBuilder`, most users are not supposed to ever see the `ISessionCreationOptions` interface).

So I have just decided that converting them to properties was not really worth the trouble in that case. Written as comment in its implementing class:

	// NH note: it is tempting to convert them to properties, but then their names conflict
	// with SessionBuilder interface.

I will add that comment on `ISessionCreationOptions` directly.

About `WithOptions()`, since it returns a new object at each call, it does not look as a good fit for property semantic.

Would you rather handle `ISessionCreationOptions` case differently? Since some of its members naming does already look a bit weird, why not additionally rename the others in order to avoid the conflict with `ISessionBuilder`, if you wish.

Have you seen other missed case of property conversion?

fredericDelaporte:
(Pushed the fixes I had announced in review.)

fredericDelaporte:
Found a bug, will fix and add test cases.

fredericDelaporte:
Bugs fixed (a `SetSelf` call and a reset of user supplied connection were missing for shared case), test cases added.

PR #626, opened 2017-05-17
NH-4008 - Split database providers into own projects.

ngbrown:
Implements [NH-4008](https://nhibernate.jira.com/projects/NH/issues/NH-4008).

This is a breaking change, but I think is a good direction for the next major version--that is NHibernate 5.0.

It separates all the implementations of `DriverBase` into separate projects to allow those individual projects to pull in NuGet dependencies.  This means that `connection.driver_class` in configuration files will probably have to be assembly qualified now.

I also added `NHibernate.Cfg.Loquacious` helpers to enable quick configuration when using configure-by-code, much like how middleware is added to OWIN in Asp.Net projects.

The update matrix looks like this:

| Previous `connection.driver_class` |  New `connection.driver_class` | Configure-by-code helper | Also pull NuGet package into top level project |
| --- | --- | --- | --- |
| `NHibernate.Driver.FirebirdClientDriver` | `NHibernate.Driver.FirebirdClientDriver, NHibernate.Driver.FirebirdClient` | `cfg.Connected.ByFirebirdClientDriver()` | N/A |
| `NHibernate.Driver.SqlClientDriver` | `NHibernate.Driver.SqlClientDriver, NHibernate.Driver.SqlClient` | `cfg.Connected.BySqlClientDriver()` | N/A |
| `NHibernate.Driver.Sql2008ClientDriver` | `NHibernate.Driver.Sql2008ClientDriver, NHibernate.Driver.SqlClient` | `cfg.Connected.BySql2008ClientDriver()` | N/A |
| `NHibernate.Driver.MySqlDataDriver` | `NHibernate.Driver.MySqlDataDriver, NHibernate.Driver.MySql` | `cfg.Connected.ByMySqlDataDriver()` | N/A |
| `NHibernate.Driver.NpgsqlDriver` | `NHibernate.Driver.NpgsqlDriver, NHibernate.Driver.Npgsql` | `cfg.Connected.ByNpgsqlDriver()` | N/A |
| `NHibernate.Driver.SQLite20Driver` | `NHibernate.Driver.SQLiteDriver, NHibernate.Driver.SQLite` | `cfg.Connected.BySQLiteDriver()` | [System.Data.SQLite.Core](https://www.nuget.org/packages/System.Data.SQLite.Core/) |
| `NHibernate.Driver.SqlServerCeDriver` | `NHibernate.Driver.SqlServerCompactDriver, NHibernate.Driver.SqlServer.Compact` | `cfg.Connected.BySqlServerCompactDriver()` | [Microsoft.SqlServer.Compact](https://www.nuget.org/packages/Microsoft.SqlServer.Compact/) |
| `NHibernate.Driver.OracleManagedDataClientDriver` | `NHibernate.Driver.OracleManagedDataAccessDriver, NHibernate.Driver.Oracle.ManagedDataAccessDriver` | `cfg.Connected.ByOracleManagedDataAccessDriver()` | N/A |
| `NHibernate.Driver.OdbcDriver` | `NHibernate.Driver.OdbcDriver, NHibernate.Driver.Odbc` | `cfg.Connected.ByOdbcDriver()` | N/A |
| `NHibernate.Driver.OleDbDriver` | `NHibernate.Driver.OleDbDriver, NHibernate.Driver.OleDb` | `cfg.Connected.ByOleDbDriver()` | N/A |

All the legacy reflection based packages moved into a shared `NHibernate.Driver.AdoNet` package, so the assembly name will need to be added to the type in the config file.  Over time, as more of the underlying official Ado.Net drivers become available on NuGet, they can be moved to their own package.

| Previous `connection.driver_class` |  New `connection.driver_class` | Configure-by-code helper |
| --- | --- | --- |
| `NHibernate.Driver.OracleClientDriver` | NHibernate.Driver.OracleClientDriver, NHibernate.Driver.AdoNet | `cfg.Connected.ByReflectionDriver<OracleClientDriver>()` <br/> or <br/> `cfg.Connected.By<OracleClientDriver>()` |

There are still 14 drivers that don't have official Ado.Net drivers on NuGet, so they will have to remain as reflection based.


hazzik:
@ngbrown great work, as usual.

hazzik:
Shouldn't dialects be in separate packages?

ngbrown:
@hazzik I started that direction, but the only actual dependency was in the xmldoc comments, so that's why I didn't move them at first.  I can circle around again and see if it makes sense moving them.

Then I can expand the use of configure-by-code to something like `cfg.UsingNpgsql()` to setup the driver and dialects at the same time.

ngbrown:
I also really wanted to rename `NHibernate.Driver.SqlClient` to `NHibernate.Driver.SqlServer` to match how Entity Framework does it and the `NHibernate.SqlServer.Compact`.  What do you think? I've made enough other changes, so I might as well make that one?

ngbrown:
@hazzik I remember why I didn't pull the dialects into the per-database projects... A major reason was the 5 Oracle dialects have 4 different drivers, and everything mixes and matches...  Only one of which is officially available on NuGet; the  [Oracle.ManagedDataAccess](https://www.nuget.org/packages/Oracle.ManagedDataAccess/) one.

Any ideas?

ngbrown:
@hazzik @fredericDelaporte I noticed this isn't marked as going into [v5.0.0](https://github.com/nhibernate/nhibernate-core/milestone/1).  I think this is a worthy breaking change and would like to see it merged.

The last thing to do is update the NAnt build here to create NuGet packages per driver.

If the Support for .NET Core 2.0 pull request (#633) is going to follow quickly it could be deferred to that. Everything done here for NuGet package is thrown away in that pull request, since msbuild is used in #633 and NAnt is used here.

ngbrown:
I've updated the NAnt build to build and publish the NuGet packages.  Importing secondary database drivers works even better that I originally hoped, even importing the native drivers for SQLite and SqlServer.Compact.

I did rename `NHibernate.Driver.SqlClient` to `NHibernate.Driver.SqlServer` to be consistent with SqlServer.Compact and Entity Framework and other ORM's naming scheme.

Once the build is finished, the nuget packages should be published on TeamCity's feed.  Can someone try it out?

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -35,6 +35,8 @@
     <StartProgram>$(MSBuildProjectDirectory)\..\..\Tools\nunit\nunit-x86.exe</StartProgram>
     <StartArguments>NHibernate.Test.dll</StartArguments>
     <TargetFrameworkProfile />
+    <NuGetPackageImportStamp>
+    </NuGetPackageImportStamp>
```
fredericDelaporte: Seems to be an old thing from VS2013.

src/NHibernate/SqlCommand/Parser/MsSqlSelectParser.cs
```diff
@@ -1,11 +1,10 @@
...
 using System.Collections.Generic;
+using System.Linq;
 
 namespace NHibernate.SqlCommand.Parser
 {
-	using System.Linq;
-
-	/// <summary>
+    /// <summary>
 	/// Represents SQL Server SELECT query parser, primarily intended to support generation of
```
fredericDelaporte: Whitespace miss.

fredericDelaporte:
About the dialects, due to the Oracle (and Sybase, and potentially Firebird) situation, having many drivers usable for the same database, so many usable drivers for the same dialect, extracting dialect from the main NHibernate project would require to put them in their own dialect projects, letting then the user choosing which drivers and which dialects he wants. (Maybe could we then put a dependency from the driver project to the dialects project, but with the Sybase case, it is not even sure we can.)

I do not think it is worth it.

fredericDelaporte:
The firebird dialect still have a method, `ClearPool`, which uses reflection while it does no more need to, since it now have a direct reference on the data provider.

ShowBuildMenu.bat
```diff
@@ -15,7 +15,7 @@ echo.
 echo --- TESTING ---
 echo B. (Step 1) Set up a new test configuration for a particular database.
 echo C. (Step 2) Activate a test configuration.
-echo D. (Step 3) Run tests using active configuration.
+echo D. (Step 3) Run tests using active configuration (Build in Visual Studio).
```
ngbrown: No, I mean to run this test step, you have to build in Visual Studio first.  Performing the build from the menu outputs to a different location.

ShowBuildMenu.bat
```diff
@@ -165,7 +165,9 @@ echo Configuration activated.
 goto main-menu
 
 :test-run
-start "nunit3-console" cmd /K %NUNIT% --x86 --agents=1 --process=separate NHibernate.nunit
+SET NUNITPLATFORM=
+IF /I "%PLATFORM%" NEQ "x64" set NUNITPLATFORM=--x86
+start "nunit3-console" cmd /K %NUNIT% %NUNITPLATFORM% --agents=1 --process=separate NHibernate.nunit
```
ngbrown: I was just trying to get it to work correctly from a mistake I had made in NH-3900 (update to NUnit 3).  There were 64 bit configurations available (SQLite and SqlServer.Compact) that weren't running correctly.

ShowBuildMenu.bat
fredericDelaporte: Precisely, I would use "built" for meaning this. It should have already been built in Visual Studio.

fredericDelaporte:
I have tested with a dummy mvc project, referencing the resulting Nuget packages (from a local repository) for NHibernate and SqlServer. It works.

ShowBuildMenu.bat
ngbrown: Reworded to "Needs built in Visual Studio".

src/NHibernate.Test/NHibernate.Test.csproj
ngbrown: Removed, but came from VS 2017...

src/NHibernate/SqlCommand/Parser/MsSqlSelectParser.cs
ngbrown: Fixed.

ngbrown:
@fredericDelaporte using the NuGet packages really shines when it can pull database specific drivers (such as `Npgsql` or `System.Data.SQLite.Core` down along with the respective NHibernate.Driver package.  I noticed the NuGet builds aren't really versioned in TeamCity (allways `5.0.0-Alpha1`), so dependencies between the preview releases may get messed up sometimes.

fredericDelaporte:
We do not actually publish them, so it should not cause issues.

Of course for testing with local repositories, this could cause issues. But since only contributors will likely do that, they should be able to fix it.

By the way I have not checked how was handled that versioning. I have not yet been involved in Nuget publishing. Alexander knows that subject way better than me.

build-common/teamcity-hibernate.cfg.xml
```diff
@@ -15,7 +15,7 @@
 			<property name="format_sql">true</property>
 
 			<!-- This is the System.Data.dll provider for MSSQL Server -->
-			<property name="connection.driver_class">NHibernate.Driver.SqlClientDriver</property>
+			<property name="connection.driver_class">NHibernate.Driver.SqlClientDriver, NHibernate.Driver.SqlClient</property>
```
ngbrown: Yes. Fixed.  Must not be used for TeamCity builds though...

fredericDelaporte: Indeed, that is overridden by default values from nhibernate-properties.xml or specific values defined for the build.

hazzik:
>I noticed this isn't marked as going into v5.0.0. I think this is a worthy breaking change and would like to see it merged.

I don't think that there is a rush to implement this for a 5.0. In fact, this is just nice-to-have feature, which can be implemented [without breaking things](https://github.com/nhibernate/NHibernate.Oracle).

In the current state, it would be useful to provide "aliases"/redirects for dialects and drivers. Eg, user can still specify "NHibernate.Driver.SqlClientDriver" but will get an warning that they need to update a configuration if the correct package is present, and error if package is not found.

ngbrown:
> In the current state, it would be useful to provide "aliases"/redirects for dialects and drivers. Eg, user can still specify "NHibernate.Driver.SqlClientDriver" but will get an warning that they need to update a configuration if the correct package is present, and error if package is not found.

I think this would have to happen at the configuration level, because C# doesn't have weak linking as far as I know...  Duplicate class names in the same namespace across assemblies may result in ambiguous name exceptions.

If there was a way to make it a warning (which may not be noticed), then it could be introduced in 4.x and then break in 5.0.

Historically NHibernate has gone through major versions fairly slow, so with .NET Core 2.0 arriving by November, when would v6 be?  I think the timing for getting this into v5 is better.  .NET Core doesn't have to be supported in v5.0, but it could then be supported in v5.x without any breaking changes for .NET Framework users.

If the thinking is to port all the .NET Core changes back to the current driver layout, I would recommend against that. I certainly don't want to do anymore manual linking, and supporting .NET Core is even more dependent on NuGet for bringing the right combination of dependency versions in for the myriad of framework and version supported by each individual package.

hazzik:
>I think this would have to happen at the configuration level, because C# doesn't have weak linking as far as I know... Duplicate class names in the same namespace across assemblies may result in ambiguous name exceptions.

image

There are few options:

a. Add a code which will manually translate the old type to a new type somewhere in Dialect#GetDialect or ReflectHelper#GetClassForName. Warning to be generated there as well. 

b. There is a type forwarding feature, but I'm not sure how it'll work if the forwarded assembly is not found, as I rarely tried it.

>Historically NHibernate has gone through major versions fairly slow, so with .NET Core 2.0 arriving by November, when would v6 be? I think the timing for getting this into v5 is better.

Think it this way: more features we are willing to add to a version, less often we will be able to release. So, it's definitely would be better to drop this feature out of radar. 


>If the thinking is to port all the .NET Core changes back to the current driver layout, I would recommend against that. I certainly don't want to do anymore manual linking, and supporting .NET Core is even more dependent on NuGet for bringing the right combination of dependency versions in for the myriad of framework and version supported by each individual package.

I did not get this point. In my view everything should just work for the ADO.NET providers which do support the .NET core and have the same name as it's "grown-up" Full .NET FX counterparts.

ngbrown:
I can setup some warning or redirection, but I think it will still be a breaking change.

The timing of v6 is my concern.  If v5 is coming out soonish, and v6 by the end of the year, then I can be fine with that.

> I did not get this point. In my view everything should just work for the ADO.NET providers which do support the .NET core and have the same name as it's "grown-up" Full .NET FX counterparts.

Tracking down all the correct versions can be a pain. With NuGet packages that contain the correct dependencies, the developer can just pick the relevant NHibernate.Driver.WhateverDB package, and then adding some configuration-by-code to tie it all together.

hazzik:
>Tracking down all the correct versions can be a pain. 

Does this assume that we need to release intermediate packages with the every new release of relevant ado.net provider?

fredericDelaporte:
> We can still do this, but without breaking changes, just by providing relevant NHibernate.Driver.WhateverDB packages in addition to built-in providers.

Meaning we would keep all the reflection based drivers while adding the possibility to use Nuget's ones? Sound a good way of not breaking. Maybe should we add in those reflection based drivers constructor some log4net warning for hinting at migrating to Nuget based ones.

ngbrown:
Since the driver packages re-use the namespace and class names of the original reflection-based drivers, adding back all the original reflection classes gets ugly fast with needing to use alias and `extern` to disambiguate what assembly you are trying to refer too.

If an intermediate step is really needed, how about a "meta-package"?  The now driverless NHibernate project would have a `PackageId` of `NHibernate.Core`, and a new meta-package of `NHibernate` would pull in every Driver package.  The `hibernate-configuration` `connection.driver_class` parsing would point out to the correct assemblies with a stern error message in the log.

I would name this meta-package `NHibernate.All`, but by not changing the name, there is the potential that just updating the package version from 4.x to the new one could actually work.

I think with all the packages pulled in, and the logged errors, and a changed description in NuGet.org, people would get the hint to include only the one Driver package, and not the overall package.

ngbrown:
I've recreated this pull request on top of v5.0-GA.  I've paired back the scope and am focusing just on pulling out the drivers that have external, maintained (that is official or semi-official), NuGet packages to depend on.

All drivers released in v5.0 will continue working as before, but if they have an external `NHibernate.Driver.XXX` package available, they will be marked as `[Obsolete]` and have logged message indicating as such (for when pulled in through .xml config)

The major idea is that the .NetStandard 2.0 targeted version of NHibernate will have no drivers included in the main package.  That targeted version of NHibernate would support only drivers that directly depend on an external NuGet packaged version of the ADO.Net driver.  Even as the same-version .NET 4.61 targeted version of NHibernate keeps the original set of reflection based drivers.

.NET Core 2.0 also doesn't support ODBC or OleDb drivers, so they would also not be available.

ngbrown:
@hazzik This is now done in a completely non-breaking way.  Can it be slated for v5.1?

annymsMthd:
+1 This makes a lot of sense. I've been following the .net standard work that you have been doing and I'm looking at sending in a pr for FluentNHibernate. This driver breakup is pretty essential. Some of the driver's just aren't supported on a .net standard.

x2764tech:
@annymsMthd How far along are you with FluentNHibernate port? There's some discussion at https://github.com/jagregory/fluent-nhibernate/issues/361

fredericDelaporte:
Maybe NHibernate driver needs to target the lowest driver version possible, and let user choose to use a newer one if he wishes.

Oracle versions greater than 12.1.2400 fail [NH-1171](https://nhibernate.jira.com/browse/NH-1171) tests, it seems it has its own bug with comments in SQL. And it is a known bug since quite long as shown [here](https://stackoverflow.com/a/7542670/1178314), but apparently with some versions suffering less of it.

Npgsql latest (3.2.5) does no more support cases where NHibernate closes the connection from the second phase of distributed transaction. (A thing that our current doc advise against anyway, urging the user to flush himself its sessions at least when he closes them inside the transaction scope, and to disable the feature delaying connection close to transaction scope completion.)

ngbrown:
> Maybe NHibernate drivers needs to target the lowest data provider version possible, and let user choose to use a newer one if he wishes.

So that was mostly my original approach, but with these new commits, I brought it up to the level of your pull request because I didn't want to be the one breaking the PostgreSQL build for everyone.

Maybe the biggest issue though, is how the `.csproj` new`<PackageReference \>` usage has changed things.  With the old projects, when importing a NuGet package, it also imported all child dependencies.  This made it very simple to update or not update any packages as you so wished.  With the new project system, the dependencies multiple levels down are not directly referenced by the top level project, and the need has to be fairly great to directly reference them and update the version.  Because of that, I think the approach should be to reference a reasonably recent, highly-compatible version, and set that as the minimum.

> Oracle versions greater than 12.1.2400 ...

Saw you changed yours, and so I changed mine back.

> Npgsql latest (3.2.5) does no more support...

Targeting just Npgsql 3.2.4.1 now, I originally was targeting 3.2.2, and for the .net core port, that wasn't so I had split the dependencies and gone with the most recent for .net core.

fredericDelaporte:
Be it for Npgsql or Oracle, if we need to upgrade for some reason, I do not think the failures they introduce should prevent us of upgrading. We should instead adjust the tests to ignore them. For Npgsql, because what fail with 3.2.5 is anyway a thing we now advise against using (not disabling `transaction.use_connection_on_system_prepare` when disposing sessions inside scopes: implies identifying tests doing that and adding a condition for ignoring them), for Oracle, because it clearly seems to be an external issue.

I do not think the package reference change is really an issue, as long as a project like NHibernate.Test is still able of overriding a lower version with a higher one if it wishes to.

src/NHibernate/Driver/NpgsqlDriver.cs
```diff
@@ -24,8 +25,23 @@ namespace NHibernate.Driver
 	/// <a href="http://pgfoundry.org/projects/npgsql">http://pgfoundry.org/projects/npgsql</a>. 
 	/// </p>
 	/// </remarks>
+#if DRIVER_PACKAGE
+	public class NpgsqlDriver : DriverBase
```
fredericDelaporte: Could be renamed `PostgreSQLDriver` or `PostgresDriver`, if we accept abstracting away the chosen data provider.

src/NHibernate/Driver/OracleManagedDataClientDriver.cs
```diff
@@ -1,8 +1,14 @@
...
 namespace NHibernate.Driver
 {
 	/// <summary>
 	/// A NHibernate Driver for using the Oracle.ManagedDataAccess DataProvider
 	/// </summary>
+#if !DRIVER_PACKAGE
+	[Obsolete("Use NHibernate.Driver.Oracle.Managed NuGet package and OracleManagedDataClientDriver."
+	          + "  There are also Loquacious configuration points: .Connection.ByOracleManagedDataClientDriver() and .DataBaseIntegration(x => x.OracleManagedDataClientDriver()).")]
+#endif
 	public class OracleManagedDataClientDriver : OracleDataClientDriverBase
```
fredericDelaporte: Could be renamed `OracleManagedDriver`. No added value here with `DataClient` in my opinion.

It seems ceasing to use reflection has not be done for avoiding duplicating the code, due to the complexity of keeping old drivers with their common reflection base with a new driver having a non reflection base. Maybe can it still overrides `CreateConnection` and `CreateCommand` in a `if DRIVER_PACKAGE` for not actually using the reflection. (And maybe `InitializeParameter` and `OnBeforePrepare` too, but at that point it would no mor need to inherit the common base.)

src/NHibernate/Driver/SQLite20Driver.cs
```diff
@@ -19,74 +20,85 @@ namespace NHibernate.Driver
...
+#if DRIVER_PACKAGE
+	public class SQLiteDriver : DriverBase
+#else
+	[Obsolete("Use NHibernate.Driver.SQlLite NuGet package and SQLiteDriver."
+		+ "  There are also Loquacious configuration points: .Connection.BySQLiteDriver() and .DataBaseIntegration(x => x.SQLiteDriver()).")]
 	public class SQLite20Driver : ReflectionBasedDriver
+#endif
 	{
+#if DRIVER_PACKAGE
+		public SQLiteDriver()
```
fredericDelaporte: If we consider our naming should hint at the chosen data provider, it would then need to be renamed `SystemDataSQLiteDriver`. But personally I am more for a neutral naming like `SQLiteDriver` chosen here.

src/NHibernate/Driver/SqlServerCeDriver.cs
```diff
@@ -127,13 +146,14 @@ private void AdjustDbParamTypeForLargeObjects(DbParameter dbParam, SqlType sqlTy
...
 			}
+#endif
 		}
 
 		public override bool SupportsNullEnlistment => false;
 
 		/// <summary>
 		/// <see langword="false"/>. Enlistment is completely disabled when auto-enlistment is disabled.
-		/// <see cref="DbConnection.EnlistTransaction(System.Transactions.Transaction)"/> does nothing in
+		/// <code>DbConnection.EnlistTransaction(System.Transactions.Transaction)</code> does nothing in
```
fredericDelaporte: I think this should be reverted. `code` is for block code and what was the harm of `see`?

ngbrown:
Hi @fredericDelaporte I'll address everything here in one comment:

> The doc/reference needs some updates, especially the driver setting documentation.

I would be inclined to do the documentation as a second pull request once I was sure this was going in by being approved and merged.

> I would rather avoid the external alias trick.

The end developer generally wouldn't have to deal with this because they would use either the assembly qualified name in XML or extension methods for fluent configuration.  Renaming is an option, but it isn't strictly necessary, and I didn't want to incur too much confusion.

The alias won't be needed long if the old drivers are removed in NHibernate 6 and live exclusively in their separate assemblies.

> New projects should have `<TreatWarningsAsErrors>True</TreatWarningsAsErrors>`...

Fixed.

> I would rather write all that with a switch

You are referring to the new C# 7 pattern matching.  It is clearer, and did catch an error that I had.

> Dialect must be non-static. It is configured, and the configuration is not static and can change.
> Needs to be reverted since GetFbTypeForParam can not be static.

Ok.

> why removing the `<see cref="System.Transactions.Transaction"/>`?
> `code` is for block code and what was the harm of `see`?

The driver packages don't need to reference`System.Transactions` for anything but these comments to avoid warning CS1580 and CS1574. Instead of referencing that additional assembly, I just changed them to pure comments instead of adding the additional reference.

I changed the `<code>` to `<c>`.

> Could be renamed MySqlDriver, if we accept to not use a naming hinting at the chosen data provider.

There is an alternate ADO.Net drivers for MySQL that already has an issue open to be added to NHibernate (#1375) so the driver name should be for MySQL should indicate which one it supports.

> It seems ceasing to use reflection has not be done for avoiding duplicating the code, due to the complexity of keeping old drivers with their common reflection base with a new driver having a non reflection base.

Yes, since the merge of all the Oracle drivers happened, I didn't want to undo all of that and then have to maintain a separate copy of the special logic.  I will add an an override of `CreateConnection` and `CreateCommand` though.

> If we consider our naming should hint at the chosen data provider, it would then need to be renamed SystemDataSQLiteDriver. But personally I am more for a neutral naming like SQLiteDriver chosen here.

There's actually a Microsoft supported driver for SQLite that works on .NET Core that I've tried to get working with NHibernate, but it has limitations on nested transactions.  If it worked out, I was planning on having the assembly named `NHibernate.Driver.SQLite.Microsoft`, since the one currently being used in NHibernate is the "official" SQLite driver.

fredericDelaporte:
Hi,

> > The doc/reference needs some updates, especially the driver setting documentation.
> 
> I would be inclined to do the documentation as a second pull request once I was sure this was going in by being approved and merged.

I am in favor of doing that within the PR causing the change to be documented. Documentation is too often forgotten. Of course postponing it till the PR changes are settled is a valid option, but better add a last commit with the documentation before merging.

> > I would rather avoid the external alias trick.
> 
> The end developer generally wouldn't have to deal with this because they would use either the assembly qualified name in XML or extension methods for fluent configuration.  Renaming is an option, but it isn't strictly necessary, and I didn't want to incur too much confusion.
> 
> The alias won't be needed long if the old drivers are removed in NHibernate 6 and live exclusively in their separate assemblies.

Those using the loquacious `IConnectionConfiguration.By<TDriver>` may have a hard time finding out how to disambiguate the class to use when wanting to fix the obsolete warning they will get if referencing an "old" reflection base driver. That would mean at least more documentation (like a big detailed obsolete message for class with name conflicts) for guiding them, hinting to alternate way for not referencing the class directly and otherwise hinting at the `extern alias` trick.

Changing more names looks less confusing to me than having the class conflict.

> > Could be renamed MySqlDriver, if we accept to not use a naming hinting at the chosen data provider.
>
> There is an alternate ADO.Net drivers for MySQL that already has an issue open to be added to NHibernate (#1375) so the driver name should be for MySQL should indicate which one it supports.

Same for Firebird, SQLite (as you write later) and probably some others who have an "official" data provider and some alternates. I think it is reasonable to not hint at the chosen data provider when we use the official one, and hint to it when we use an alternate one. Or otherwise, for having a really consistent naming for all the drivers we manage, name all of them with their specific data provider name. 

Currently we use the "official" data provider (or mainstream) with a mix of naming: sometime using the data provider name, other times just the db name. Of course Oracle is a special case, since it has many official data provider...

ngbrown:
The `Obsolete` warning does say: "There are also Loquacious configuration points: .Connection.ByNpgsqlDriver() and .DataBaseIntegration(x => x.NpgsqlDriver())."

ngbrown:
Since it seems like naming is hanging this up, while I am indifferent to changing all the names, here is something that would work:

| Previous `connection.driver_class` |  New `connection.driver_class` | New Configure-by-code helper |
| --- | --- | --- |
| `NHibernate.Driver.FirebirdClientDriver` | `NHibernate.Driver.FirebirdClientDriver, NHibernate.Driver.Firebird` | `cfg.Connected.ByFirebirdClientDriver()` |
| `NHibernate.Driver.SqlClientDriver` | `NHibernate.Driver.SqlServerDriver, NHibernate.Driver.SqlServer` | `cfg.Connected.BySqlServerDriver()` |
| `NHibernate.Driver.Sql2008ClientDriver` | `NHibernate.Driver.SqlServer2008Driver, NHibernate.Driver.SqlServer` | `cfg.Connected.BySqlServer2008Driver()` |
| `NHibernate.Driver.MySqlDataDriver` | `NHibernate.Driver.MySqlDataDriver, NHibernate.Driver.MySql` | `cfg.Connected.ByMySqlDataDriver()` |
| `NHibernate.Driver.NpgsqlDriver` | `NHibernate.Driver.NpgsqlDriver, NHibernate.Driver.Npgsql` | `cfg.Connected.ByNpgsqlDriver()` |
| `NHibernate.Driver.SQLite20Driver` | `NHibernate.Driver.SQLiteDriver, NHibernate.Driver.SQLite` | `cfg.Connected.BySQLiteDriver()` |
| `NHibernate.Driver.SqlServerCeDriver` | `NHibernate.Driver.SqlServerCompactDriver, NHibernate.Driver.SqlServer.Compact` | `cfg.Connected.BySqlServerCompactDriver()` |
| `NHibernate.Driver.OracleManagedDataClientDriver` | `NHibernate.Driver.OracleManagedDataClientDriver, NHibernate.Driver.Oracle.Managed` | `cfg.Connected.ByOracleManagedDataClientDriver()` |

ngbrown:
*Sorry if you got an email for an earlier version of this, I hadn't finished when I accidentally hit submit...*

Since it seems like naming is hanging this up, while I am indifferent to changing all the names, here is something that would work.  Primary official drivers available on NuGet are named based on the database they support.  Additional drivers would be named as derivatives of that.

| Previous `connection.driver_class` |  New `connection.driver_class` |
| --- | --- | --- |
| `NHibernate.Driver.FirebirdClientDriver` | `NHibernate.Driver.FirebirdDriver, NHibernate.Driver.Firebird` |
| `NHibernate.Driver.SqlClientDriver` | `NHibernate.Driver.SqlServerDriver, NHibernate.Driver.SqlServer` |
| `NHibernate.Driver.Sql2008ClientDriver` | `NHibernate.Driver.SqlServer2008Driver, NHibernate.Driver.SqlServer` |
| `NHibernate.Driver.MySqlDataDriver` | `NHibernate.Driver.MySqlDriver, NHibernate.Driver.MySql` |
| `NHibernate.Driver.NpgsqlDriver` | `NHibernate.Driver.PostgreDriver, NHibernate.Driver.Postgre` |
| `NHibernate.Driver.SQLite20Driver` | `NHibernate.Driver.SQLiteDriver, NHibernate.Driver.SQLite` |
| `NHibernate.Driver.SqlServerCeDriver` | `NHibernate.Driver.SqlServerCompactDriver, NHibernate.Driver.SqlServer.Compact` |
| `NHibernate.Driver.OracleManagedDataClientDriver` | `NHibernate.Driver.OracleManagedDriver, NHibernate.Driver.Oracle.Managed` |

Alternative ADO.Net drivers that could be supported in the future could named like this:

* `NHibernate.Driver.MySqlConnectorDriver, NHibernate.Driver.MySql.MySqlConnector` (#1375)
* `NHibernate.Driver.SQLiteMicrosoftDriver, NHibernate.Driver.SQLite.Microsoft` [*](https://www.nuget.org/packages/Microsoft.Data.Sqlite/)
* `NHibernate.Driver.PostgreDotConnectDriver, NHibernate.Driver.Postgre.DotConnect`
* etc...

I do think that any alternative driver projects and NuGet package generation should live in the main NHibernate repository.  These drivers are fairly thin and easy to maintain. This will enable catching any compile-time breakage.  The drivers NuGet packages could end up becoming the primary dependency for NHibernate of client applications, so having them all in this repository will enable simultaneous publishing.

If these names are fine, I'll rename the new projects and drivers.  I'll also base the documentation on these new names.

fredericDelaporte:
Small detail: it should be `Potsgres` or `PostgreSQL`, but not `Postgre`. See [ProjectName](https://wiki.postgresql.org/wiki/ProjectName).

While I favor the naming change your last comment proposes, I do not consider the "extern alias" trick as a no-go, and can accept this PR in its current state. But I would like to have @hazzik or @gliljas point of view on this subject before.

If you do not know what I am referring to, here an abstract: the new driver assembly does in some cases have a driver name identical to the old one in NHibernate assembly, within the same namespace. This may cause ambiguous references, which will need to be either avoided (using configuration instead of direct reference by example, or some helper added by this PR) or fixed with an `extern alias` (like [here](https://github.com/nhibernate/nhibernate-core/pull/626/files#diff-a8c868957e22547252445791977d6ca4R11) (wait the load for having it scrolled appropriately), with a definition in [csproj](https://github.com/nhibernate/nhibernate-core/pull/626/files#diff-baeb45e98a980e825d2b6e33ebcf3016R47)). The discussion about it is somewhat complete if reading it from this [comment](https://github.com/nhibernate/nhibernate-core/pull/626#issuecomment-345503543) and the ones following it.

src/NHibernate/Dialect/MsSql2000Dialect.cs
```diff
@@ -42,6 +41,18 @@ namespace NHibernate.Dialect
...
 	{
+		public const int MaxSizeForAnsiClob = 2147483647; // int.MaxValue
+		public const int MaxSizeForClob = 1073741823; // int.MaxValue / 2
+		public const int MaxSizeForBlob = 2147483647; // int.MaxValue
+
+		public const int MaxSizeForLengthLimitedAnsiString = 8000;
+		public const int MaxSizeForLengthLimitedString = 4000;
+		public const int MaxSizeForLengthLimitedBinary = 8000;
+		public const byte MaxPrecision = 28;
+		public const byte MaxScale = 5;
```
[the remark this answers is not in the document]
fredericDelaporte: > If #708 gets merged before, `MaxScale` and `MaxPrecision` would have to be removed from here, and just left obsolete in the driver, as "having no more usages".

#708 is now merged

src/NHibernate/Connection/ConnectionProvider.cs
```diff
@@ -112,9 +112,49 @@ protected virtual void ConfigureDriver(IDictionary<string, string> settings)
...
+						LogObsoleteDriver(nameof(SqlServerCeDriver), "SqlServer.Compact", "SqlServerCompactDriver");
+						break;
+				}
+#pragma warning restore 618
 			}
 		}
 
+		private static void LogObsoleteDriver(string obsoleteDriverName, string nugetName, string newDriverName)
+		{
+			log.Error(
```
fredericDelaporte: An obsolete is usually a warning, not an error. I think it should change to `Warn`.

fredericDelaporte:
Updated update matrix:

| Previous `connection.driver_class` |  New `connection.driver_class` | New Configure-by-code helper |
| --- | --- | --- |
| `NHibernate.Driver.FirebirdClientDriver` | `NHibernate.Driver.FirebirdDriver, NHibernate.Driver.Firebird` | `cfg.Connected.ByFirebirdDriver()` |
| `NHibernate.Driver.SqlClientDriver` | `NHibernate.Driver.SqlServer2000Driver, NHibernate.Driver.SqlServer` | `cfg.Connected.BySqlServer2000Driver()` |
| `NHibernate.Driver.Sql2008ClientDriver` | `NHibernate.Driver.SqlServer2008Driver, NHibernate.Driver.SqlServer` | `cfg.Connected.BySqlServer2008Driver()` |
| `NHibernate.Driver.MySqlDataDriver` | `NHibernate.Driver.MySqlDriver, NHibernate.Driver.MySql` | `cfg.Connected.ByMySqlDriver()` |
| `NHibernate.Driver.NpgsqlDriver` | `NHibernate.Driver.PostgreSqlDriver, NHibernate.Driver.PostgreSql` | `cfg.Connected.ByPostgreSqlDriver()` |
| `NHibernate.Driver.SQLite20Driver` | `NHibernate.Driver.SQLiteDriver, NHibernate.Driver.SQLite` | `cfg.Connected.BySQLiteDriver()` |
| `NHibernate.Driver.SqlServerCeDriver` | `NHibernate.Driver.SqlServerCompactDriver, NHibernate.Driver.SqlServer.Compact` | `cfg.Connected.BySqlServerCompactDriver()` |
| `NHibernate.Driver.OracleManagedDataClientDriver` | `NHibernate.Driver.OracleManagedDriver, NHibernate.Driver.Oracle.Managed` | `cfg.Connected.ByOracleManagedDriver()` |

hazzik:
Could someone explain, why this change is REQUIRED for .NET Core 2.0?

hazzik:
Could someone explain, why this change is REQUIRED for .NET Core 2.0? No one stops anyone from creating own driver packages now. We even have several: [Oracle](https://github.com/nhibernate/NHibernate.Oracle), [Jet (MS Access)](https://github.com/nhibernate/NHibernate.JetDriver). I believe we just overcomplicating the design and trying to solve mistical unicorn case instead of solving the real problem which is bring .NET Core 2.

My belief that this changes are not required to be able to run .NET Core 2.0

ngbrown:
The Microsoft SQL Server client isn't included in the primary set of dependencies in .NET Core like it is for .NET Framework.

Without this change the SQL Server ADO.NET dependencies would be required for every database user whether or not they needed it.  Alternatively it could be loaded by reflection (but only in .NET Core, which is also confusing).

As far as over-complicating the design, why have most ORMs that came after NHibernate gone down this path?  NHibernate's current design reflects a time before a mature .NET dependency management system was in place (NuGet).  The world has changed since then.

hazzik:
>Alternatively it could be loaded by reflection (but only in .NET Core, which is also confusing).

I'll vote for this option.

annymsMthd:
@ngbrown lol you should just fork nhibernate and be done with it. We are using your changes in production after updating a 5 year old app and everything is running great. It is really sad how slow things are moving here. Well over 6 months and there seems to be a debate going on about packaging multiple database driver into one nuget.. I don't want some postgres driver hanging around when I don't need it taking up memory. Why is this even a debate? In the words on Monty Python.. Get On With It!

gliljas:
@annymsMthd I wouldn't say it's slow. @hazzik  and @fredericDelaporte are extremely active. Maybe this issue has lingered, but it's sometimes hard to prioritize.

gliljas:
While moving SQL Server out adds a bit of inconvenience, it's certainly the cleanest approach.

fredericDelaporte:
>>Alternatively it could be loaded by reflection (but only in .NET Core...).
>
>I'll vote for this option.

This is certainly the more lightweight option in term of code impact. But the current state of this PR (with my proposed changes in ngbrown/nhibernate-core#5) looks quite good to me and allowing a cleaner way of pulling in the required drivers. It has my preference.

> I don't want some postgres driver hanging around when I don't need it taking up memory.

I do not understand this. The current "reflection based" drivers do not pull in any dependency like Npgsql by itself. If your app does not pull it, it will not be there. Only a few lines of code for handling dynamically loading it when used (and failing if it is missing) will be there.

ngbrown:
The pull request now contains @fredericDelaporte changes from ngbrown/nhibernate-core#5

fredericDelaporte:
My `ObsoleteDriverFixture` tests do not exclude the Odbc driver case, which is not obsoleted, causing the failure of ODBC build.
I propose to adjust its `AppliesTo` this way:
```c#
{
	switch (dialect)
	{
		case FirebirdDialect _:
		case MsSql2000Dialect _:
		case MsSqlCeDialect _:
		case MySQLDialect _:
		case PostgreSQLDialect _:
		case SQLiteDialect _:
			return !ReflectHelper.ClassForName(cfg.GetProperty(Cfg.Environment.ConnectionDriver)).IsOdbcDriver();
		default:
			return false;
	}
}
```

I can push a fixup commit (with async regen) if you wish.

src/NHibernate/Driver/SqlClientDriver.cs
```diff
@@ -12,23 +13,49 @@ namespace NHibernate.Driver
 	/// <summary>
 	/// A NHibernate Driver for using the SqlClient DataProvider
 	/// </summary>
+#if DRIVER_PACKAGE
+	public class SqlServer2000Driver : DriverBase, IEmbeddedBatcherFactoryProvider
+#else
+	[Obsolete("Use NHibernate.Driver.SqlServer NuGet package and SqlServer2000Driver."
+		+ "  There are also Loquacious configuration points: .Connection.BySqlServer2000Driver() and .DataBaseIntegration(x => x.SqlServer2000Driver()).")]
 	public class SqlClientDriver : DriverBase, IEmbeddedBatcherFactoryProvider
```
fredericDelaporte: For this PR to be a complete preparation to .NetStandard, shouldn't this driver be switched to reflection based when compiled inside NHibernate.csproj?

src/NHibernate/Driver/Sql2008ClientDriver.cs
```diff
@@ -5,7 +5,13 @@
```
fredericDelaporte: For this PR to be a complete preparation to .NetStandard, shouldn't this driver be switched to reflection based when compiled inside NHibernate.csproj, and so cease using directly `System.Data.SqlClient`?

fredericDelaporte:
@hazzik, after having review it fully one more time, I do not see anymore any possible breaking changes in this PR for NHibernate users.
So I remove this label.

src/NHibernate/Driver/SqlClientDriver.cs
ngbrown: My intention was for .NetStandard to not have these reflection-based drivers at all.  If an application is migrating to .NetStandard, it isn't break-free, so no reason to introduce an already obsolete driver when it could be migrated ahead of time or at the same time.

fredericDelaporte: So the plan is to have those files excluded in csproj when target is .NetStandard? Ok.

doc/reference/modules/configuration.xml
```diff
@@ -1219,65 +1228,91 @@ in the parameter binding.</programlisting>
                         <row>
                             <entry>Microsoft SQL Server 7</entry>
                             <entry><literal>NHibernate.Dialect.MsSql7Dialect</literal></entry>
-                            <entry></entry>
+                            <entry>Use same default driver than <literal>NHibernate.Dialect.MsSql2000Dialect</literal>.</entry>
```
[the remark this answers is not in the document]
ngbrown: I'm not sure what you mean here.  Is "article" an xml element for documents?  Or did you mean you should use a different word instead of "than"?  This would read better:

"Use the same default driver **as** `<literal>NHibernate.Dialect.MsSql2000Dialect</literal>`."

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -35,35 +35,22 @@
...
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="FirebirdSql.Data.FirebirdClient" Version="5.11.0" />
     <PackageReference Include="log4net" Version="2.0.8" />
     <PackageReference Include="NSubstitute" Version="3.0.1" />
-    <PackageReference Include="Microsoft.SqlServer.Compact" Version="4.0.8876.1" />
-    <PackageReference Include="MySql.Data" Version="6.9.10" />
-    <PackageReference Include="Npgsql" Version="3.2.4.1" />
     <PackageReference Include="nunit" Version="3.7.1" />
-    <PackageReference Include="Oracle.ManagedDataAccess" Version="12.1.2400" />
```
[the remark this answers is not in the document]
ngbrown: The nunit line moved, that was the problem here.  It's rebased now.

doc/reference/modules/configuration.xml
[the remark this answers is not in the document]
fredericDelaporte: The current sentence is `Use same default driver`, without `the` (which is what I have called an "article", in French "déterminant"). I think it should be `Use the same default driver`. About `than`, yes, better put `as` instead.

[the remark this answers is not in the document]
ngbrown: Updated text and also added links to where the NuGet packages would be on nuget.org.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2393,5 +2394,12 @@ public virtual ISQLExceptionConverter BuildSQLExceptionConverter()
 			// may override to return whatever is most appropriate for that vendor.
 			return new SQLStateConverter(ViolatedConstraintNameExtracter);
 		}
+
+		internal static string GetDriverName<TFallback>(string preferredName) where TFallback : IDriver
+		{
+			return ReflectHelper.ClassForFullNameOrNull(preferredName) == null
+				? typeof(TFallback).FullName
+				: preferredName;
```
fredericDelaporte: Checking this again, I realize it generates a lot of exceptions when it needs to fallback, because it is called at each dialect creation, and dialects are created numerous times even for a single test. Even when it does not need to fallback, being called frequently is not a good thing since it performs reflection.

So I propose to cache the name checking:

```c#
private static readonly ConcurrentDictionary<string, System.Type> DriverClasses = new ConcurrentDictionary<string, System.Type>();

internal static string GetDriverName<TFallback>(string preferredName) where TFallback : IDriver
{
	return DriverClasses.GetOrAdd(preferredName, ReflectHelper.ClassForFullNameOrNull) == null
		? typeof(TFallback).FullName
		: preferredName;
}
```

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2397,9 +2399,9 @@ public virtual ISQLExceptionConverter BuildSQLExceptionConverter()
 
 		internal static string GetDriverName<TFallback>(string preferredName) where TFallback : IDriver
 		{
-			return ReflectHelper.ClassForFullNameOrNull(preferredName) == null
-				? typeof(TFallback).FullName
-				: preferredName;
+			return GetDriverNameMemoize.GetOrAdd(
+				preferredName,
+				s => ReflectHelper.ClassForFullNameOrNull(s) == null ? typeof(TFallback).FullName : s);
```
fredericDelaporte: Strictly speaking, incorrect. Currently, `TFallback` is always the same for a given `preferredName`, but nothing guarantees this. This is why my proposed code was not caching the function result but only the preferred type resolution.

fredericDelaporte:
@hazzik, you were seemingly not agreed with the changes here, that I have now validated, and that seems to be considered good at least on their principles by @gliljas ((see here)[https://github.com/nhibernate/nhibernate-core/pull/626#issuecomment-354667564]). Are you still against merging this?

hazzik:
>Are you still against merging this?

@fredericDelaporte preliminarily yes. I'll take a closer look later.

ngbrown:
I did rebase these commits to head again.  I am still advocating for the driver split to be part of v6.1.

I do think the drivers should still live in the same repository as NHibernate, instead of being separately managed and released.

fredericDelaporte:
> I would still like the database provider nuget packages to be in v5.1. Because the .net core stuff went ahead I am going rebase the work.

I am agree with this, but it looks like the 5.1 may not wait much longer.

ngbrown:
I split off some pre-work in #1610 that can be merged quickly.

For this pull request I needed to put some more effort into the tests, and was hoping the effort wouldn't be wasted.

ngbrown:
This pull request is now rebased.

The last commit, 'In psake.ps1, build and test in one step.' is needed because not every project can (or should) target `netcoreapp2.0`.

src/NHibernate.Driver.Oracle.Managed/NHibernate.Driver.Oracle.Managed.csproj
```diff
@@ -0,0 +1,47 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="../../build-common/NHibernate.props" />
+
+  <PropertyGroup>
+    <Description>Driver for Oracle databases to be used with NHibernate 5.</Description>
+    <PackageTags>NHibernate; Driver; ODP.Net; Oracle; ADO.Net</PackageTags>
+
+    <TargetFramework>net461</TargetFramework>
```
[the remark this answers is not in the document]
ngbrown: Since the beta version of `Oracle.ManagedDataAccess` with .NET Core support isn't on NuGet yet, it shouldn't be added yet, but once it is, then yes, that would be an option.

I addressed the rest of the feedback as well.

ngbrown:
Can this be in 5.1 then?

ngbrown:
Can this be in 5.2 then?

fredericDelaporte:
With the requests on #1635 and #1638, maybe should we go some steps more far, splitting them in a new sub repository like nhibernate/providers. This would also allow to publish new providers without having to wait for a nhibernate-core release.

I may work on this later.

ngbrown:
There are several reasons to keep all drivers in this main repository:

- The drivers will still be needed for the various unit tests of each of the databases.
- This repository already has the tooling and scripts for building, testing, and releasing any number of NuGet packages.
- The implementation of the separate NuGet packages needs to stay synchronized with the reflection based driver until they are removed (NHibernate 6.0, if this pull request get in on a 5.x release).
- There's not a lot of code in these drivers, so they aren't going to need a lot of updates.
- When adding new drivers, back-versioning them to the last NHibernate release when adding them to NuGet shouldn't be a problem.
- The driver's don't have to stay up to date with each ADO.Net release because the NuGet reference to the 3rd party ADO.Net driver is only a minimum version; an end application can force an update to a newer version by adding a direct reference to the ADO.Net driver.

maulik-modi:
@ngbrown and @fredericDelaporte , This is excellent feature so individual project can be upgraded without affecting other drivers. Moreover, drivers such as System.DAta.OracleClient could be removed in future. 

We are eagerly awaiting this feature. What is holding us here?

PR #628, opened 2017-05-20
NH-4013 - SQL-Server batcher failure after CloseCommands.

fredericDelaporte:
[NH-4013](https://nhibernate.jira.com/browse/NH-4013) - `SqlClientBatchingBatcher` CloseCommands contract violated

`CloseCommands` causes `SqlClientBatchingBatcher` and `MySqlClientBatchingBatcher` to be unusable although its xml comment states:

> ```C#
> /// Use this method instead of <c>Dispose</c> if the <see cref="IBatcher"/>
> /// can be used again.
> ```

Test case and fix.

### DIFF
--- a/src/NHibernate/AdoNet/MySqlClientBatchingBatcher.cs
+++ b/src/NHibernate/AdoNet/MySqlClientBatchingBatcher.cs
@@ -74,7 +74,6 @@ namespace NHibernate.AdoNet
 			if (Factory.Settings.SqlStatementLogger.IsDebugEnabled)
 			{
 				Factory.Settings.SqlStatementLogger.LogBatchCommand(currentBatchCommandsLog.ToString());
-				currentBatchCommandsLog = new StringBuilder().AppendLine("Batch commands:");
 			}
 
 			int rowsAffected;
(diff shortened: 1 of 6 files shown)


src/NHibernate/AdoNet/MySqlClientBatchingBatcher.cs
```diff
@@ -74,7 +74,6 @@ protected override void DoExecuteBatch(DbCommand ps)
 			if (Factory.Settings.SqlStatementLogger.IsDebugEnabled)
 			{
 				Factory.Settings.SqlStatementLogger.LogBatchCommand(currentBatchCommandsLog.ToString());
-				currentBatchCommandsLog = new StringBuilder().AppendLine("Batch commands:");
```
fredericDelaporte: Moved to `ClearCuurentBatch`.

hazzik:
I think https://github.com/nhibernate/nhibernate-core/pull/513 has more correct implementation

fredericDelaporte:
#513 looks invalid as an issue to me, since we are not supposed to reuse the session (and hence its batcher) after a failure. So for me there is no point trying to fix such case. Unless we intend someday to enable session recovering from failures.

Here I fix a "no exception" case, where calling `CloseCommands` causes the batcher to be no more usable while it should still be. This case may happen in almost "normal use" cases, when an user open a reader then commit a transaction without having closed the reader, while not using `OnClose` connection release mode.

fredericDelaporte:
And now that is #513 which is "behind", since it will still leave the batcher unusable in case of preparation failure (network error or whatever). I think that is a bit non-sens to change code for an explicitly unsupported use case like re-using the session after an exception. The batcher will still be unusable if the failure occurs outside of `DoExecuteBatch` by example.

fredericDelaporte:
Rebased for forcing rebuild, failure on teamcity side (xml report truncated).

PR #627, opened 2017-05-18
NH-4011 - transaction scope failures

fredericDelaporte:
~~[NH-2176](https://nhibernate.jira.com/browse/NH-2176)~~ [NH-4011](https://nhibernate.jira.com/browse/NH-4011) - That is gone to the point of fixing it as much as possible.

Since the branch used in #410 is public and in the official NHibernate repository, rebasing  it is debatable.

So for going on with my tests, I have switched on a branch in my own repository.

Rebased, custom factory dropped, slightly squashed.

DTC failure tests cleaned up and completed.

PostgreSQL reactivated for DTC in order to check what is going on with it.

PR #630, opened 2017-05-24
NH-4015 - Update Npgsql driver and enable DTC for it

fredericDelaporte:
[NH-4015](https://nhibernate.jira.com/browse/NH-4015) - Update Npgsql driver and enable DTC for it

I expect this to fail due to DTC not yet enabled on teamcity server side.

I have seen the link to open a rdp toward it, but I have not found a valid login for connecting and adjusting that.


fredericDelaporte:
Got DTC errors as expected, we need to enable it on TeamCity build agent, but I do not knows the credentials for this.

Got some other tests errors due to a breaking change in Npgsql, rebased on #631 for solving them.

fredericDelaporte:
With a `Thread.Sleep(500)` after committing distributed transaction, flaky tests for Oracle and even SQLite looks no more flaky... Quite ugly. For me that is a deficiency in those databases data provider implementations, causing them to have not always actually finished the commit work when the scope disposal is left.

As discussed [here](https://github.com/npgsql/npgsql/issues/1571#issuecomment-303756783), I expect PostegreSql to have similar issues.

fredericDelaporte:
PostgreSql re-run had one DTC test failing, apparently due to previously done insert not already taken into account. It looks like it has the "latency" issues I am suspecting [here](https://github.com/nhibernate/nhibernate-core/pull/627#issuecomment-304329241).

Going to rebase on NH-4017 anyway.

lib/teamcity/postgresql/License_System.Threading.Tasks.Extensions.txt
```diff
@@ -0,0 +1,128 @@
+
+MICROSOFT SOFTWARE LICENSE TERMS
+
+
+MICROSOFT .NET LIBRARY
```
fredericDelaporte: We need that dependency for testing Npgsql v3. So I have put it there, but this causes NHibernate to redistribute it and so being subject to comply with that licence.

Maybe we should instead add the NuGet dependency in the test project, it would avoid having to "redistribute" that only for tests.

src/NHibernate.Test/NHSpecificTest/DtcFailures/DtcFailuresFixture.cs
```diff
@@ -276,6 +276,9 @@ public void CanDeleteItemInDtc()
 				}
 			}
 
+			// Dodging "latency" due to db still haven't actually committed a distributed tx after scope disposal.
+			Thread.Sleep(100);
```
fredericDelaporte: I intend to put a better "dodge" logic in #627. It will apply only to database drivers showcasing the "latency" trouble.

lib/teamcity/postgresql/License_System.Threading.Tasks.Extensions.txt
hazzik: I vote for NuGet

fredericDelaporte: Done.

fredericDelaporte:
Any concern against merging this?

src/NHibernate.Test/TestDialects/PostgreSQL83TestDialect.cs
```diff
@@ -21,13 +16,5 @@ public override bool SupportsNullCharactersInUtfStrings
         {
             get { return false; }
         }
-
-        /// <summary>
-        /// Npgsql's DTC code seems to be somewhat broken as of 2.0.11.
-        /// </summary>
-        public override bool SupportsDistributedTransactions
```
hazzik: Need to check when DTC become available in Postgres.

fredericDelaporte: But, it is available since long. Tests succeed. It was failing previously not because Postgres was not supporting it, but because we were not at all enlisting Postgres connection into scopes: enlist is `false` by default with npgsql. This PR switch it to `true`.

hazzik: Ah, sorry, I did not see that this is a dialect's test-double.

hazzik: Looks good then

hazzik:
Merged

PR #632, opened 2017-05-27
NH-4019 - Pass assembly into log4net functions

ngbrown:
[NH-4019](https://nhibernate.jira.com/browse/NH-4019) - Pass assembly into log4net functions

This will be needed for the .netstandard version because `log4net` does not have access to Assembly.GetCallingAssembly() in .netstandard version.

This pull request is minimal changes and no `#if` cases.

### DIFF
--- a/src/NHibernate/Logging.cs
+++ b/src/NHibernate/Logging.cs
@@ -3,6 +3,7 @@ using System.Configuration;
 using System.IO;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using NHibernate.Util;
 
 namespace NHibernate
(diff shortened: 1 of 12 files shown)


src/NHibernate/Logging.cs
```diff
@@ -222,16 +223,16 @@ public void WarnFormat(string format, params object[] args)
...
 		static Log4NetLoggerFactory()
 		{
-			GetLoggerByNameDelegate = GetGetLoggerMethodCall<string>();
+			GetLoggerByNameDelegate = GetGetLoggerByNameMethodCall();
 			GetLoggerByTypeDelegate = GetGetLoggerMethodCall<System.Type>();
 		}
 		public IInternalLogger LoggerFor(string keyName)
 		{
-			return new Log4NetLogger(GetLoggerByNameDelegate(keyName));
+			return new Log4NetLogger(GetLoggerByNameDelegate(typeof(Log4NetLoggerFactory).Assembly, keyName));
```
fredericDelaporte: Don't we need an overload for specifying another `Assembly` than NHibernate.dll? They are two calls in NHibernate.Test.dll which were likely resolving to the test assembly instead.

ngbrown: No, because the delegate is running in the context of the `Log4NetLoggerFactory` class, so when `Assembly.GetCallingAssembly()` was called in `log4net`, it only goes one call-level up--back to `NHibernate`.

fredericDelaporte:
May you amend the commit message for it to be associated with NH-4019 please?

ngbrown:
@fredericDelaporte done.

fredericDelaporte:
Thanks.

fredericDelaporte:
"Side effect": we will have to take care that any new logger call pass the assembly in, till we have something in the build starting failing otherwise. Or we will have to "redo" that later.

PR #633, opened 2017-05-28
NH-3807 - Support for .NET Core 2.0

ngbrown:
This is ~[NH-3807](https://nhibernate.jira.com/browse/NH-3807)~ #954 - A Full .NetStandard 2.0 and .Net Core App 2.0 conversion.  It is currently a work in progress, but is working well enough to be reviewed and receive feedback.

It is my opinion that if it doesn't run on Linux/macOS, then it's not really compatible with .NetStandard.   That means the default fallback of importing .Net Framework assemblies needs disabled in each `.csproj` targeting .NetStandard or .NetCoreApp:

    <DisableImplicitPackageTargetFallback>true</DisableImplicitPackageTargetFallback>

This version is different than @hazzik's [NH-3807 branch](https://github.com/hazzik/nhibernate-core/commits/NH-3807) in than it aims to fully delineate what parts are available on .Net Core, and what parts are going to require the full .NET Framework.  This version does that by splitting the functionality up among the different database drivers.  A .Net Framework 4.6.1 output is still cross-compiled and included in the NuGet packages for full traditional functionality, and to avoid the extra NuGet imports that happen with .netstandard only packages.

What's working:
- [x] Using [.NET Core SDK 2.0.0](https://github.com/dotnet/core/blob/master/release-notes/download-archives/2.0.0-download.md)
- [x] Builds with `dotnet build` on both Windows and Linux.
- [x] Tests with `dotnet test` on both Windows and Linux.
- [x] Runs on Windows against SQL Server.  The .NET Framework version still passes all tests.
- [x] Runs on Linux (Ubuntu 14.04) against SQL Server.
- [x] NuGet packages automatically built.

What's left:
- [x] Need NuGet feed for new version of `Remotion.Linq.EagerFetching`.  ~Currently using checked-in copy of attachment available [here](https://www.re-motion.org/jira/projects/RMLNQFETCH/issues/RMLNQFETCH-19).~
- [x] Resolve ~311~ 0 failed tests on Windows when running as a .NetCoreApp.
- [x] Resolve ~312~ 0 failed tests on Linux.
- ~Build `SharedAsemblyInfo.cs` on Linux.~ _Using assembly versions didn't set the NuGet package version correctly._
- [x] Set versions in project files during build.
- [ ] Cross-platform build.
- [x] Depends on ~#568 - NH-3944~ - Upgrade Remotion.Linq to v2.1.
- [ ] Depends on #626 - NH-4008 - Split database providers into own projects.
- [x] Depends on ~#654 - NH-4043~ - Complete keyword registration in dialects.
- [x] Depends on ~nhibernate/iesi.collections#3~
- ~Integrate with Ant build system.~ _Planning on using fully native msbuild instead._
- [x] All projects switched over.
  - [x] Switch `NHibernate.Test.VisualBasic` to .NetCoreApp ~(not supported in the preview 1 CLI?)~
  - [x] Switch `NHibernate.TestDatabaseSetup`  to .NetCoreApp
  - ~Switch `NHibernate.Tool.HbmXsd` to .NetCoreApp~ _HbmXsd uses `XmlCodeExporter`, which is not available in .NetCoreApp_
  - [x] Switch `NHibernate.Example.Web` to ASP.Net Core 2.0.  Started with #675 - NH-4073.
  - [x] Switch back to normal solution file.
- [x] Use `dotnet build` generated NuGet packages?
- [x] Decide if this should avoid a rollback of commit 5c448337500970489009feb2c2e87a59ecdd4b31.  That means ~antlr/antlrcs#71~ will need to be resolved.  What if it still won't run on Linux?  Do the ANTLR definition files change all that often?
- [x] Test on Linux with these databases:
  - [x] Microsoft SQL Server - _The [RC1 release for Linux](https://hub.docker.com/r/microsoft/mssql-server-linux/) is unstable but when the server survives, all tests pass_.
  - [x] Firebird - works well
  - [ ] MySql - Passes all but 58 tests.  Fails some tests with the .netstandard driver (on Windows too) with `There is already an open DataReader associated with this Connection which must be closed first.`
  - [x] PostgreSQL - fastest at completing the tests.
  - ~[Microsoft's SQLite](https://www.nuget.org/packages/Microsoft.Data.Sqlite/)~ _only allows one transaction, so errors when trying to get HiLo id values.  Allowing multiple connections results in deadlocks._
- [x] Output test assemblies for both .NET Framework 4.6.1 and .NetCoreApp 2.0 at the same time.
- [x] Rebase to released v5.0 to include Async functionality.

Some small non-breaking changes can be merged ahead such as ~#632~, ~#634~, and ~#695~.

What is different missing on .NET Core vs. full .NET Framework?

* .NET Core is unable to participate in distributed or multi-phase enlisted transactions.  This means that `new TransactionScope()` may not work like you expect.  You will get a run-time exception.
* ~Serialization of the configuration or session isn't going to work. You will get a run-time exception.~ (Compatible serialization pending in #1425)
* Certain types aren't available to use, such as `Image` and `Bitmap`. You will get a compile-time error.
* Not every database driver will be available from .NET Core.  Depending on how you configure NHibernate, you may get a compile-time or a run-time exception.
* Batching is not available in any of the drivers since it relied on reflection to internals that changed.
* MySQL with .netstandard does not support memory transport, even in Windows, so remove `Protocol=memory;` from the connection string and add the Port number.

Other tasks?  Thoughts?

**To test:**
These are being built in Appveyor and CircleCI.  The built NuGet packeges are available on this [Appveyor NuGet feed](https://ci.appveyor.com/nuget/nhibernate-core-6cm11v2xdhud).  The versions look like this: 5.0.0-ci-00138-dbg-NH-3807.


src/NHibernate/NHibernate.csproj
```diff
@@ -1,1819 +1,41 @@
...
-    <Compile Include="Logging.cs" />
-    <Compile Include="Mapping\ISqlCustomizable.cs" />
-    <Compile Include="Param\AbstractExplicitParameterSpecification.cs" />
-    <Compile Include="Param\AggregatedIndexCollectionSelectorParameterSpecifications.cs" />
-    <Compile Include="Param\CollectionFilterKeyParameterSpecification.cs" />
-    <Compile Include="Param\DynamicFilterParameterSpecification.cs" />
-    <Compile Include="Param\IExplicitParameterSpecification.cs" />
-    <Compile Include="Param\IParameterSpecification.cs" />
-    <Compile Include="Param\NamedParameterSpecification.cs" />
-    <Compile Include="Para
```
hazzik: There will be no "System.Reflection.Emit" for .NetStandard 2.0, only for .NetCoreApp 2.0

hazzik: So, I'm not sure if this is a legitimate use of this package.

shallowtempo:
I pulled down this PR and ran the tests against the PostgreSQL83Dialect and .NET Standard 2.0 . Here are the results:
```
Total tests: 5692
Passed:      4747
Failed:       408
Skipped:      488
Not run:       49
```

Which is roughly what is reported above against SQL Server but a bit lower pass rate.

I tried analying the failed tests a bit and didn't see a clear failure pattern.

@ngbrown Is it possible for you to characterize what code functionality still needs work in the PR?

Is there a specific type of functionality that hasn't been worked on yet? Or is there a bunch of issues in a variety of areas?

I am trying to gauge if I could pick up a build based on this PR in order to start seeing if NHibernate on .NET Core 2.0 will work for us and am unsure at a glance if the functionality we need intersects with the unfinished areas.

ngbrown:
@shallowtempo, hello, and thanks for showing interest in helping. I'm guessing that with so few tests failing, there's a high likelihood it will work fine in a real-world application.  If you could try it out and if you come across any problems, let me know.

src/NHibernate/NHibernate.csproj
```diff
@@ -1,1837 +1,41 @@
...
-    <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
-    <DocumentationFile>bin\Release-2.0\NHibernate.XML</DocumentationFile>
-    <Prefer32Bit>false</Prefer32Bit>
-  </PropertyGroup>
-  <PropertyGroup>
-    <SignAssembly>true</SignAssembly>
-  </PropertyGroup>
-  <PropertyGroup>
-    <AssemblyOriginatorKeyFile>..\NHibernate.snk</AssemblyOriginatorKeyFile>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```
hazzik: I don't understand why do we need to do .netstandard instead of .netcore

ngbrown: @hazzik, I would like to understand what you think the benefits of targeting .netcore would be?

From what I understand, `netcoreapp2.0` is for the resulting application or executable (and a specific runtime), while .netstandard is for libraries that will be compatible with current and future .netstandard compliant runtimes...  That means not just .NET Framework 4.6.1 and .NET Core 2, but also other implementations such as Mono, Unity, Tizen, etc.

All libraries I know of target .netstandard.  Entity Framework Core does, ASP.Net Core 2 will (they had some issues with the preview 1, which is why they didn't initially).

Often when I use NHibernate in an application, I'm not linking it just to the final Web project, but to an intermediate class library that contains the domain layer and database mappings that is used in a variety of ways, including for unit tests and adjacent console task projects.  I think it would be in the best interest to have that layer as portable as possible.

ngbrown: This might interest you; Scott Hanselman [just addressed a very similar question](https://www.hanselman.com/blog/HowToReferenceANETCoreLibraryInWinFormsOrNETStandardExplained.aspx).

fredericDelaporte: So in short, we could theoretically even only target netstandard, no more net461. But since we still have a bunch of missing features in netstandard, better keep the explicit target on the full framework for now.

hazzik: Thanks, I understand what is the difference between .NetStandard and .NetCore. 

>From what I understand, netcoreapp2.0 is for the resulting application or executable (and a specific runtime)

Not quite. There are plenty of libraries targeting .NetCoreApp. 

>I think it would be in the best interest to have that layer as portable as possible.
 
It will not if we have to cut a lot of things because they are not supported in a standard. 

>other implementations such as Mono, Unity, Tizen, etc

I doubt if anyone would use NHibernate on these platforms.

hazzik:
>This version is different than @hazzik's NH-3807 branch in than it aims to fully delineate what parts are available on .Net Core, and what parts are going to require the full .NET Framework. 

In my branch there are 2 separate builds: .NET461 and .NetCoreApp. the .NET Core build does not have any dependency on a full framework.

hazzik:
> Decide if this should avoid a rollback of commit 5c44833. That means antlr/antlrcs#71 will need to be resolved. What if it still won't run on Linux? Do the ANTLR definition files change all that often?

I have compiled a private version of [ANTLR](https://www.nuget.org/packages/Antlr3.core). It works and is capable of building the NHibernate assembly... But `dotnet build` fails because VB is not yet supported in CLI

hazzik:
> Integrate with Ant build system.

Prefer to completely purge N/Ant and switch to MSBuild

hazzik:
> Build SharedAsemblyInfo.cs on Linux.

Need to switch to dotnet native versioning

src/NHibernate/NHibernate.csproj
oskarb: > > other implementations such as Mono, Unity, Tizen, etc
> 
> I doubt if anyone would use NHibernate on these platforms.

I do use NHibernate on Mono on Linux. Thouch I have not yet investigated/analyzed if that would or should be changed to NH on .NetCore on Linux eventually.

hazzik: Since Mono is compatible with full .NET framework it does not count. (I meant to exclude it from this list)

fredericDelaporte: Mono compatibility with full .Net framework is imperfect. Compliance to .NetStandard seems more achievable. Being able of using .NetStandard compatible libraries instead of relying on the imperfect compatibility with full framework should give more predictable results (stability).

> There are plenty of libraries targeting .NetCoreApp.

Which is maybe an unfortunate choice from them, unless they had no other choice.

> It will not [be of interest] if we have to cut a lot of things because they are not supported in a standard.

Of course, but does .NetStandard lacks things NHibernate requires and are available in .NetCore?

src/NHibernate/NHibernate.csproj
hazzik: System.Reflection.Emit is available only for netstandard1.0 + platform extensions (whatever it does mean)

src/NHibernate/Context/WcfOperationSessionContext.cs
```diff
@@ -1,3 +1,5 @@
+#if !NETSTANDARD2_0
```
hazzik: This is supported through [System.ServiceModel.Primitives](https://www.nuget.org/packages/System.ServiceModel.Primitives) (.NetStandard + Platform Extensions)

src/NHibernate/NHibernate.csproj
hazzik: >Of course, but does .NetStandard lacks things NHibernate requires and are available in .NetCore?

We have some dependencies which are not pure .NetStandard, but .NetStandard + platform extensions (System.Reflection.Emit & System.Reflection.Emit.Lightweight)

ngbrown: @hazzik 

> It will not if we have to cut a lot of things because they are not supported in a standard.

Very little has been cut out, and none on full .NET Framework. What is not lost is the richness that NHibernate has for legacy databases and having that available on upcoming platforms like in Docker Containers (both Windows and Linux) is what my goals are in doing this.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -1,3 +1,4 @@
+#if !NETSTANDARD2_0
```
hazzik: Why? 

As I can see the System.Transaction is supported on .netstandard.

ngbrown: Having this class enabled makes my failure count go from 67 to 1,676.

Simply calling `new TransactionScope()` causes: `System.Transactions.TransactionAbortedException : The transaction has aborted.`.

Calling `transactionContext.AmbientTransation.EnlistVolatile(transactionContext, EnlistmentOptions.EnlistDuringPrepareRequired);` results in: `System.Transactions.TransactionException : The operation is not valid for the state of the transaction.`.

Calling `System.Transactions.Transaction.Current.EnlistDurable(Guid.NewGuid(), this, EnlistmentOptions.None);` results in: `System.PlatformNotSupportedException : This platform does not support distributed transactions.`.

Maybe it's supported, but it doesn't seem to be working for me.  And this is Windows.  What does a distributed transaction even look like on Linux?

This is the implementation: dotnet/corefx#10089 Maybe they are missing something we need?

hazzik: Start reading from here: https://github.com/dotnet/corefx/issues/15046 => https://github.com/dotnet/standard/issues/157 => https://github.com/dotnet/standard/pull/168 

Basically it is supported in a standard, but not yet implemented everywhere

ngbrown: @hazzik, so what do you suggest? From what I gather, there is a very narrow scenario that may work, but lots of NHibernate's tests fail with this factory.

I have never used distributed transactions, but at this point I would prefer to say that they are only supported in the 4.6.1 framework version, not in .net core. Mainly because I don't know what it will take for distributed transactions to work.

fredericDelaporte: Please mind that `TransactionScope` is not only about distributed transaction. I believe its most frequent usage is not distributed at all. Its main purpose is to be a transaction handling pattern. It additionally supports distributed transaction with full framework, but as far as I know, not within the standard for now. Distributed transaction are dependent on MS DTC, which is an old Windows only technology. Your test with the durable ressource showcases this, since enlisting a 2PC only durable ressource always enforces promotion to distributed. Only SinglePhasePromotable durable ressource can be used without enforcing promotion to distributed, provided their are not enlisted along with another durable ressource. Database connection are normally enlisting as durable ressource. Some data provider (like Npgsql) do enlist them as volatile, but that is not a normal case. Maybe some other data provider enlist them as 2PC only durable ressource, and then those ones will not be usable with .Net Standard.

Tests about distributed transaction (those ones forcing escalation to distributed) should be disabled if not running under the full framework. (In #627, I have made a bunch of renaming for ceasing to imply distributed where nothing was distributed: this should help having a clearer view.)

But we should have all the other `TransactionScope` use cases working.

ngbrown: @fredericDelaporte I got it mostly working.  Do you have any idea how to implement `ITransactionPromoter.Promote()` or if it even needs implemented?  It's part of the `IPromotableSinglePhaseNotification` interface.

ngbrown:
I've made some progress and am now down to only 64 failures (on Windows).

A big chuck of the previous failures were from `System.Configuration.ConfigurationManager` not being able to determine the path of the test assembly because it relies on `Assembly.GetEntryAssembly()` which returns the unit test loader, and there's no way around that because there are no app domains in .NET Core.  So I have to manually load and extract the XML out of the config file (see `TestsContext.cs`).

I also have enlisted transactions working, that means most of the `TransactionScope` tests are back, with the exception of those that rely on MSDTC (distributed transactions).

What is remaining are mostly grouped into the following categories:
* Lingering `ConfigurationManager` problems.
* Queries and table schema operations that aren't escaping keywords, because `GetSchema()` isn't supported and the dialect files aren't complete.
* Serialization of a session fails because `System.Type`,  `MethodInfo`, and `FieldInfo` are not serializable.
* `BeginInvoke` calls used in tests are not supported https://github.com/dotnet/corefx/issues/5940

src/NHibernate/Context/WcfOperationSessionContext.cs
ngbrown: I re-enabled it, However, I don't think there are any tests that cover it.

src/NHibernate/NHibernate.csproj
danilobreda: @hazzik the correct way to go is .NetStandard... the problem is that if i would like to use .net 462, i cant because its only .net core. With standard its possible. If you see the packages on nugget, 90% is targeting netstandard. Its a point for no problems on the future.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
fredericDelaporte: `Promote` is for going distributed. In the mono case, it seems likely to me it has to throw `NotSupportedException`. But for which use do you want to implement `IPromotableSinglePhaseNotification`? If it is in the "DtcFailures" tests, this could cause them to be no more distributed, which is not the aim of those tests. If it is in the `AdoNetWithDistributedTransactionFactory`, it should not be needed, and I do not believe it can be done while preserving the "flush on commit" feature. The 2 PC volatile ressource NHibernate is enlisting there should not cause promotion to distributed (only a durable is supposed to do that).

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -90,7 +90,7 @@ public void ExecuteWorkInIsolation(ISessionImplementor session, IIsolatedWork wo
 			}
 		}
 
-		public class DistributedTransactionContext : ITransactionContext, IEnlistmentNotification
+		public class DistributedTransactionContext : ITransactionContext, IEnlistmentNotification, IPromotableSinglePhaseNotification
```
fredericDelaporte: Has it really changed anything to add that interface while still enlisting the resource the same way at line 71? I do not think so. I think this change is unnecessary. Please tell me if undoing it causes failure for non-distributed cases.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -104,6 +104,79 @@ public DistributedTransactionContext(ISessionImplementor sessionImplementor, Sys
 				IsInActiveTransaction = true;
 			}
 
+			#region IPromotableSinglePhaseNotification Members
+
+			void IPromotableSinglePhaseNotification.Initialize()
```
fredericDelaporte: It does not look to me as a valid place for flushing before commit. It flushes too early here, right at the beginning of the transaction I believe, not "just before actual commit". Anyway currently I think this code here has no actual usage (never called).

fredericDelaporte:
> Queries and table schema operations that aren't escaping keywords, because GetSchema() isn't supported and the dialect files aren't complete.

Not complete or "plain wrong", as the Sybase one which registers keywords in caps while the matching of keywords is done against lower case tokens. Currently keywords are to be registered in lower case, otherwise they are ignored. (The base dialect could just do a to lower on them, but that is not currently the case.)

src/NHibernate/NHibernate.csproj
hazzik: @danilobreda if you want to use 462. You reference 461. No problemo here.

hazzik: The problem with targeting .NetStandard is that we depend on some packages which are not .NetStandard, but .NetStandard+Platform Extensions. It means that it would fail on the platforms which do not have these extensions.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
ngbrown: You are correct about `IPromotableSinglePhaseNotification` not actually being needed.  When I added the cleanup `CheckEnlistedTransactionsWereDisposed` in `TestCase.cs` is when most of the errors went away.

I'll revert the changes to `DistributedTransactionContext`.

ngbrown:
Here's a summary of the failures left:

| Type of error                                  | Number of failing tests |
|--------------------------------------|--- |
| Remote Transaction                       | 9  |
| No System.Web.HttpRequest DLL | 1  |
| Unable to Serialize type                | 22 |
| Unknown                                       | 7  |

I'm going to concentrate on finding the root cause of the unknown category.  The others are going to end up `#if !DOTNETAPP2_0` out unless someone has a way to move forward on them, especially the serialization.

To continue including session serialization, it basically has to be able to serialize as JSON (as an example). The .NET Core binary serializer no longer supports serializing complex internal types such as `System.Type`, `MethodInfo`, and `FieldInfo`.  Their reasons for this, is that they want to be able to evolve those types without having the internals locked down.

ngbrown:
Also, I'm re-writing the build scripts with [Cake](http://cakebuild.net/), as it can be run cross-platform, enabling building and testing on all supported .NET Core platforms.

hazzik:
I would like to stick with msbuild. It's cross platform. And a part of a
platform

ngbrown:
@hazzik, something needs to drive the `dotnet` command line with version string and the framework being tested.  A lot of what I've written so far is interfacing between the CI server and the `dotnet` command line, which would then use msbuild for actual building, packaging, testing, etc.

NUnit is making quite successful use of it, while still using msbuild for building and package creation: https://github.com/nunit/nunit/blob/master/build.cake

I was also going to look into branching out with a Linux build on something like [Travis CI](https://travis-ci.org/), as it is possible to have a container configured with the various database servers desired for testing.  Including [Microsoft SQL Server 2017](https://hub.docker.com/r/microsoft/mssql-server-linux/).

fredericDelaporte:
Why are those last three commit prefixed NH-4008? This Jira does not seems related to those commits.

Rather than putting compilation directive for disabling tests, I wonder if it could be easier, at least for some cases, to use the ignore logic already in place for some databases. For instance, for distributed tests, if #627 was merged, it could simply be one compilation directive "overriding" `SupportsDistributedTransactions` to false in dialects, for having distributed tests disabled.

About the last commit "question", I do not believe .Net Core supports interacting with MSDTC which is a very windows specific technology and is required for distributed transaction. So testing remotely or not should not change anything to that. Distributed transactions are not supported by any target other than the .Net framework currently. (Transaction scopes are supported by many more, but provided they stay non-distributed.)

This last commit appears to have some change related to weak-references troubles too, but does not mention them in its commit message.

ngbrown:
Sorry, they should have been NH-3807, I fixed it.  I was trying something with `WeakReference` but it didn't work out and accidentally included it in the commit.  It should be removed.

I've been trying to work toward zero test failures on .NET Core, but am not there yet.  There are some `WeakReference` failures (I opened an issue with dotnet/coreclr) and some `Trim` failures.

On the distributed transactions; I thought that .NET Core was supposed to support single-phase `TransactionScope` but as far as I can tell, most tests trying to use `TransactionScope` at all fail.

I had been running Visual Studio 2017.3 Preview from a VM, then accessing SQL Server across the network to the main machine, so I thought that was a possible source of failing, but I ran against a local SQL Server, and they still fail, so it's not a question any more.

I'm going to go around again and figure out a better way to disable tests, maybe by setting `Ignore` or asserting inconclusive.

fredericDelaporte:
Currently most tests involving transaction scopes are distributed in NHibernate. In the previously linked PR I am adding a bunch of tests which are not supposed to be distributed (but ODBC seems to easily promote to distributed, so ADBC may still cause issues there).

ngbrown:
Both Windows and Linux (Ubuntu 14.04) are now passing with 0 unit test failures against SQL Server.

src/NHibernate/NHibernate.csproj
RUSshy: .net core is 2x more performant than mono, it'll disapear soon, and MS is working on AOT compilation for (https://github.com/dotnet/corert) so porting to .net core makes a LOT of sense

looking forward to it, thanks for initiative, i was about to use EF 6 for my next project, but looks like ill just wait for nhibernate core come :)

ngbrown:
[.NET Standard 2.0](https://blogs.msdn.microsoft.com/dotnet/2017/08/14/announcing-net-standard-2-0/) and [.NET Core 2.0](https://blogs.msdn.microsoft.com/dotnet/2017/08/14/announcing-net-core-2-0/) have been announced as released.

I've updated my branch to use release NuGet packages.

MSSQL and PostgreSQL are 100% working on Linux.  MySQL has a few failures under .netstandard (both Windows and Linux) because something about the driver is different, but it will probably work for you.

Continous integration builds from my repository are running on [AppVeyor](https://ci.appveyor.com/project/ngbrown/nhibernate-core/history) for Windows and and on [Circle CI](https://circleci.com/gh/ngbrown/nhibernate-core/tree/NH-3807-CI) for Linux.  A custom [Docker image](https://hub.docker.com/r/ngbrown/nhibernate-build-dotnet-sdk/) was used for building on Circle CI, while standard database images were used for the DB server.

A NuGet feed of these builds is available at: https://ci.appveyor.com/nuget/nhibernate-core-6cm11v2xdhud

penderi:
Firstly - huge congrats on getting tests passing on linux. 
I'm looking to test and see the results too.. but I found that the Tests project is targeting 461. So I wonder, how are the tests being run ?

ngbrown:
@penderi thanks for your interest in reviewing the code.  The .NET Core code is still in my fork on it's own branch.  This is the line you are looking for:

https://github.com/ngbrown/nhibernate-core/blob/NH-3807/src/NHibernate.Test/NHibernate.Test.csproj#L4
```xml
    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
```

The tests are then ran by selecting a target framework.  Something like this:

```
dotnet test -f netcoreapp2.0 ./src/NHibernate.Test/NHibernate.Test.csproj
```

303248153:
Would you consider use [MySqlConnector](https://www.nuget.org/packages/MySqlConnector) instead of [MySql.Data](https://www.nuget.org/packages/MySql.Data/8.0.8-dmr)?
It's MIT license and it provides better .Net Core support.
Or provider another DataDriver for it.

robertorp:
@ngbrown 

Hi, could you make a nhibernate.dll available for me to be testing? I'm having trouble compiling

penderi:
Hi @ngbrown 
I'm not sure if this is useful but I've the downloaded built and used in some of our more simple apps.
It certainly seems to be working, with a few caveats:
- nHibernate.Cfg needs a reference to System.Configuration.ConfigurationManager. I was a little surprised as this seems more of a backwards compatible step than a dotnetcore-like configuration style. Still it works
- the code works fine under .NetStandard but doesn't run for me under .net 461:
`System.TypeLoadException : Type 'NHibernate.Tuple.Component.PocoComponentTuplizer' in assembly 'NHibernate, Version=5.0.0.0, Culture=neutral, PublicKeyToken=aa95f207798dfdb4' has method 'OnDeserialized' with an incorrect signature for the serialization attribute that it is decorated with.
`
this seems to be thrown from NHibernate.Tuple.Component.ComponentMetamodel. Any ideas what this could be?

Hoping this is helpful. I've not used it under load yet, but simple connect/CRUD scenarios are good under .NetStandard

ngbrown:
@penderi, while I prefer using just configure-by-code and mapping-by-code, I'm not sure how the larger user-base feels about that.

Parts of NHibernate that are linked to legacy technologies such as `System.Configuration.ConfigurationManager` could be pulled into optional NuGet packages to minimize existing code changes when updating to a new version, but so far, the path has been to keep NHibernate together as a monolithic library.  This means all the dependencies need to be pulled in, even if not being used.

I haven't seen any problems like you described under .net 4.61.  NHibernate should be dual-building against both `netstandard20` and `net461`.  If you build all the assemblies for your actual target platform and the NuGet packages are used, then the correct assembly should be selected at compile time.  If you are going through an intermediate library, make sure that library is correctly targeted or at least dual-targeted as well:

```xml
<TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```

Unfortunately, binary serialization has so little support in .netstandard, that NHibernate probably shouldn't have any `[Serializable]` attributes at all for the .netstandard build.  I just haven't gone through and `#if` them all out.

penderi:
@ngbrown maybe the problem is I'm building NH myself and adding a direct reference... perhaps the issue is different bindings for 461/Standard. I'll try and look as well as evaluating the async stuff we may have now.
Moreover, any feel as to how close we are to a beta/RC ?

mattzink:
The NuGet packages produced by this branch include unwanted content files (LICENSE.txt, nhibernate-configuration.xsd, and nhibernate-mapping.xsd). This causes them to be further propagated as content in derived packages, where the license won't probably match the derived package's license.

Also, if you reference both the main package and a driver one, you get a build warning say that multiple dependencies are attempting to add the content with the same filename (LICENSE.txt).

Please remove these unnecessary content files from the NuGet packages.

ngbrown:
@mattzink Good points.  I fixed the readme and *.xsd files to be named and embedded in the .nupkg file the same way as 4.x.  They no longer get included in the project when being included.

Try the new builds in the [Appveyor NuGet feed](https://ci.appveyor.com/nuget/nhibernate-core-6cm11v2xdhud): 5.0.0-ci-00138-dbg-NH-3807.

mattzink:
@ngbrown Thanks, but I don't see any changes in this PR branch. Are you not keeping this PR up to date with your changes? We are patching NH ourselves on top of your patch, so using nuget packages directly isn't an option for us.

ngbrown:
@mattzink I re-based the changes into the ".netstandard 2.0 project" commit.  The .csproj files should have `<PackagePath>...</PackagePath>` for the readme and *.xsd now.

penderi:
Hi @ngbrown,
I'm looking to try and integrate again, and I've a cou0le of questions:
- I've reset my source to your PR (git reset origin pull/633/head) . The build is green (good) but I don't see any output in ~\nhibernate-core\src\NHibernate\bin\Debug. Editing the csproj shows only     <TargetFramework>net461</TargetFramework> . I'm manually adding netstandars2.0, but wondered if I'm missing something?
- Do we have any idea on the nHibernate community/contributors/council of wisdom's view of this PR? It seems the shortest route for NH on NetStandard by far... and we're keen to adopt.

Many thanks again for all your efforts. Again we're green on .NetStandard20 in a vanilla SqlServer env ...

robertorp:
I got The type initializer for 'NHibernate.Context.ReflectiveHttpContext' threw an exception. when configuring property.Add ("current_session_context_class", "web");
.net core 2.0 WebApi
Postgresql

ngbrown:
@robertorp ASP.Net Core is quite a bit different, and I don't think is has a global HTTP Context any more.  Any lifetime control of NHibernate sessions should be managed through manually working with the ASP.Net Core dependency injection system.

robertorp:
@ngbrown  thanks. 
and async await, was looking at your branch NH-3807-CI still does not have the correct async support await?

I have a project to do, I do not need async await. But you doing this support would recommend using which database? What is the support so far and more guaranteed?

danilobreda:
ASP.NET core does not have a global context... I dont recommend the use of ASP.Net Core dependency injection system, its very simple right now. A lot of third party injectors like Simpleinjector have a good way to integrate with the asp.net core and manage scopes... im using it right now with the 4.1 nhibernate and asp.net core(.net framework) on production.
Another question... How can i participate with the testing of this branch/issue? (withpostgresql), how i install the nuget? Thanks.

x2764tech:
@ngbrown is there any chance you can bump the version number in the nuget feed? The `5.0.0` release is taking precedence over the version in the feed

ngbrown:
@danilobreda see an earlier comment for the Appveyor feed: https://github.com/nhibernate/nhibernate-core/pull/633#issuecomment-329678965

ngbrown:
@x2764tech are you having a problem selecting a specific version?  NuGet supports setting to a specific version in the `.csproj` file, from the command line, and from the Visual Studio GUI.

x2764tech:
@ngbrown sorted it out, thanks! CI server not respecting/finding Nuget.config with your feed in it!

mstancombe:
Hey, just tried browsing the appveyor nuget feed, and it looks empty.  Is this expected?  I'm trying to reference it from a net standard 2.0 library, could that be the problem?

ngbrown:
@mstancombe, The NuGet browser in Visual Studio should show something when the URL is added as a custom feed.

mstancombe:
@ngbrown All good thanks, I just hadn't checked 'Include prerelease' 😜    Thanks so much for working on this port 👍

mstancombe:
@ngbrown Some feedback: I've done a rough port of fluent nhib to core (don't judge me.. 😛 ) , and it all worked perfectly together with this branch, even on linux subsystem (using mssql provider) 😄 Huge thanks!

danilobreda:
@mstancombe i know this is a little offtopic, but can you share with us the port of the fluent nhib. ? The fluent nhibernate with the nhibernate is the stack that we use here. I think is the stack of a lot of enterprise projects too... this will be perfect.. thanks

default.build
```diff
@@ -235,65 +240,75 @@
...
-  <target name="nuspec" depends="init nuget.set-properties" description="Create nuspec files">
-    <nant target="nuspec">
-      <buildfiles refid="nugetfiles.all" />
-    </nant>
-  </target>
-  
-  <target name="nuget" depends="init binaries nuget.set-properties nuspec" 
-          description="Creates files for the release on nuget gallery.">
+	<fileset id="nugetfiles.all" basedir="src">
+		<include name="NHibernate/NHibernate.build" />
```
hazzik: These should be built by standard tools (eg nuget via MsBuild's PackageaProject(dont remember exact name of a property))

ngbrown: @hazzik This line is an artifact of the #626 pull request needing to build and test within the current framework.  If you could review and approve that first, it would make this pull request easier.

I don't use these NAnt build files in my [CI branch](https://github.com/ngbrown/nhibernate-core/commits/NH-3807-CI) and do exactly that so that I can build for multiple platforms and multiple CI providers (right now CircleCI and AppVeyor).

There seemed to be resistance to switching away from NAnt, so that's why it's in a separate commit (and branch).  I can replace the NAnt build scripts entirely if there is a feeling that the effort would be accepted into the project.

ngbrown: See https://github.com/ngbrown/nhibernate-core/blob/1644ce59001378ba18f863f03935eb77b753e0b6/src/NHibernate/NHibernate.csproj#L15 for the use of `GeneratePackageOnBuild` from MsBuild.

ngbrown: I've updated #626 so it is now on top of the new #1404 (Use MsBuild for packing .nupkg files) so there will be no more per-project `.build` files and the `.nupkg` files are built as part of the main MsBuild task.

I'm going to rebase this pull request, but it will take some time.  If the pre-requisite pull requests can be reviewed first, that would certainly help.

erikbra:
Any hope of progress on this issue? Personally I don't have the necessary knowledge of NHibernate do do any good here, I fear. But is there somehow I can help getting this forward? We have a .NET standard port of RoundhousE that is kindof waiting for NHibernate to be .NETStandard compatible.

fernandolguevara:
Hi, is there any news on this issue?

teonivalois:
I do see some progress on #626, which is an active dependency... maybe a good idea would be to follow up on the dependencies as well

poojabirla:
Any idea as to when #626 will be merged in? It will be great if you can provide a timeline as to when this will be ready. No rush, thanks!

hazzik:
This PR has broken configuration. Instead of trying to fix the configuration it has hidden the fact that the configuration is broken.

ngbrown:
> This PR has broken configuration. Instead of trying to fix the configuration it has hidden the fact that the configuration is broken by tampering the tests.

`App.config` is mostly broken in .NET Core 2.0.  I'm working around it by manually specifying the `NHibernate.Test.dll.config` file name and loading it from the correct directory.

I'm still in the middle of re-building this on top of the Async and other changes that have happened since the pull request was made.

jpenniman:
@ngbrown, @hazzik , it's not so much that app.config is broken in core, it's virtually non-existent.  It's been replaced with Microsoft.Extensions.Configuration.  When we use Asp.Net Core with FullFramework and NHibernate, we either use hibernate.cfg.xml or set the properties manually using a custom ConfigOptions block in appSettings.json, depending on the use-case.

@hazzik, in @ngbrown's defense, the tests are no longer valid since Core does not support app.config at all.

My sense is, if the goal is pure NetStandard with no fallbacks or compiler constants, support for app.config will need to be removed or replaced with a IOptions model using the new Microsoft.Extensions.Configuration bits.

I'd be happy to share our code for brainstorming.

hazzik:
@jpenniman you are wrong. App.config works well in .NetCore 2. 

The problem is that when test run with `dotnet test`/VS/R# the entry assembly is `testhost.dll`.

The logic has been broken in Environment.cs, as `Configuration.GetSection` requires the implementation of `ConfigurationSection` as opposite to `IConfigurationSectionHandler` for `ConfigurationManager.GetSection`.

ngbrown:
@hazzik I'm not really planning on doing anything differently around `app.config` than what I did in this earlier pull request.  I didn't disable any test because of it, so is there something specific that was the problem with manually locating the `.config` file?

ondravondra:
@ngbrown Hi, would it be possible to rebase this PR onto master? I would like to try the appveyor nuget in our company project on linux, but it requires the latest nhibernate features.

LightCZ:
I quite does not understand whats the current state of this merge request. This merge request is getting really really old and have some really major changes to NHibernate. It would be nice, if someone from NHibernate team would at least make an announcement about current plans for NHibernate support .NET Core 2.0 since its out for a long time now.

ngbrown:
@LightCZ check out my [NH-3807-Async branch](https://github.com/ngbrown/nhibernate-core/tree/NH-3807-Async) where I was rebuilding this pull request.  I haven't had time to go through the few failing tests and ensure their is either a fix or it needs to be excluded.

For .NET Framework there are no breaking changes, and the legacy NAnt build scripts "worked" in the TeamCity CI.  In a separate branch I had added a second set of build scripts for Circle-CI (Linux) and Appveyor (Windows).  That was how I had Appveyor nuget packages.

ngbrown:
I've rebuilt this and re-based this pull request.  It relies on #1425 and #626 so exclude those when looking at the diff.

It has ran on Linux (with my [CI branch](https://github.com/ngbrown/nhibernate-core/tree/NH-3807-CI)) with 0 failures for [postgres](https://circleci.com/gh/ngbrown/nhibernate-core/322) and [firebird](https://circleci.com/gh/ngbrown/nhibernate-core/321).

junalmeida:
i guess that an eye should be put at this. dotnet core is on its version 2, the other ORMs are bad. The sooner this get released, more adoption will come up. all checks have passed, this is exiting but I cannot see any code review yet.  
maybe is a good idea to release a "beta" version.

LightCZ:
Yeah I feel it the same way. .NET Core 2 was a long time in RC and is released for a long time now. Releasing this even in BETA/ ALPHA / PRE-Release etc. would take some .NET community attention. After all nhiberante is one of the most mature frameworks in .NET world and its kinda shame there is no .NET CORE version of it available.

gudmundurm:
Releasing nhibernate with .net core 2.0 support is a major thing and would surely get massive attention. 

It could also be smart to create some documentation along with it, show how it's best to wire it up and etc.

giggio:
NHibernate not being available to .NET Core is surely a big impediment to the platform adoption. I am 100% sure that as soon as a beta, or heck, even an alpha, is released, people will start adopting it.
I really hope this is reviewed quickly. And the ones this depends on. They are already almost a year old, and we are all really anxious to get our hands on it.

iancooper:
A lot of people would be willing to help, as this gets them onto Core. A lot of folks would be happy with a beta that just supported SQL Server.

EF Core is still not ready for prime time, so I agree with others that this would make a LOT of difference to the .NET community.

Can we help?

ngbrown:
@iancooper @erikbra @penderi I think a way to help would be to submit a separate pull request that totally replaces the NAnt build scripts with either Cake (see [my CI branch](https://github.com/ngbrown/nhibernate-core/tree/NH-3807-CI) for a start) or with PowerShell and Psake (see @hazzik and [his branch](https://github.com/hazzik/nhibernate-core/blob/9af0923f50d9fbe70d42d5e24ce0ae274dd42762/psake.ps1) for that).

The goal is to build and test in Linux (and macOS) in addition to Windows, so I think that Cake has an advantage here as it has fewer dependencies by not needing PowerShell in addition to the .NET Core SDK.

This hypothetical pull request would be just for Windows and would need need to interoperate with the current way that TeamCity interacts with the builds (all tests passing, artifacts being picked up).

iancooper:
@ngbrown We are talking to the Cake community for you. Will copy you in on Twitter

gep13:
@iancooper @ngbrown we, the Cake Team, are certainly willing to help if we can.  Is there a specific problem that you are trying to address, or do you have any questions?

penderi:
@gep13 looking above Nathan and the team need to produce Linux and MacOS binaries from a .NetStandard codebase. The thinking seems to be fewer dependencies (ie no Powershell) should make it easier.
BUT it would mean replacing the current NANT build scripts with Cake.  Which could be a big job.

gep13:
@penderi I haven't looked to see everything that the NAnt Script is currently doing, but in terms of creating a Cake Build Script that orchestrates the `dotnet` cli, across multiple platforms, that is definitely something that Cake can help with.  I would recommend starting small, and getting Cake to do the basic things first, and then build up to include all the required functionality.  Feel free to reach out and ask questions as when you run into any problems.

gep13:
i.e. there would be no harm in having both the NAnt and Cake scripts in place during the transition period.

ngbrown:
@gep13 @iancooper  Thanks for reaching out.  I think the current maintainers of NHibernate (@hazzik, @fredericDelaporte, @gliljas) would benefit from a discussion around cross-platform build.

@hazzik had previously pushed for an MSBuild only solution, but must have had trouble coming through as he started using PSake.  @fredericDelaporte had started a pull request around using AppVeyor (with PowerShell embedded in the AppVeyor YAML file).

I had gotten the basic build and testing working [with Cake on Windows and Linux using only .NET Core](https://github.com/ngbrown/nhibernate-core/blob/NH-3807-CI/build.cake), but it does not include building the documentation or distribution ZIP.  My solution isn't included in this pull request and is just used privately to test this pull request and provide private NuGet packages.  The build platform is a decision that should probably be made separately which is why I threw the request for help out there in the hopes that someone would take on the task to completely replace NAnt. A ready-to-go solution would make the maintainer's decision easier.

Would everyone be open for a teleconference or a chat session?

gep13:
>@ngbrown said...
Would everyone be open for a teleconference or a chat session?

Yes, I would be happy to jump on a call.

penderi:
I'm not sure I can add to the call I'm just trying to connect the dots... if I can help in anyway please shout and I'd be more than happy to join the call.

x2764tech:
> ...it does not include building the documentation or distribution ZIP.

Does the documentation need to be built on Linux?

LightCZ:
I think that chat session is much better, because you can keep history in a text form. You can reach some conclusions in Tel-Co, but still someone has to made some kind of written overview. I ll be happy to be part of that chat.

Also I think that in the past msBuild was used because there used to be some Teamcity builds. Currently teamcity does not natively support psake or cake - you have to use command line build steps.

fredericDelaporte:
> Currently teamcity does not natively support psake or cake - you have to use command line build steps.

I do not think this is an issue in the case of the NHibernate TeamCity builds, because it uses a build agent provided by @hazzik. He can install whatever he wants on it. (I also have remote access to this build machine.)

> @fredericDelaporte had started a pull request around using AppVeyor (with PowerShell embedded in the AppVeyor YAML file).

I have done this (#1508) neither for cross platform compatibility, nor in a .Net Core perspective. I have done this because our TeamCity builds currently relies on a build machine provided by @hazzik, and I wanted to investigate the possibility to cease needing it. I have not used the Nant build scripts at all in it because it was looking to me as more work than simply rewriting what they do with PS.
As explained in this PR, I am not convinced striving to get a cross-platform CI build solution is worth it. I mean, I doubt it will be overall simpler than having a short dedicated script for each CI, like the one I have put in this AppVeyor PR. The Nant scripts we currently have handle both the CI and the developer workstation, but are at least five time as big. And they have still hard coded dependencies on how the things are setup on the machine, causing them to be quite far from being cross-platform too.

By the way, taking the approach of ad-hoc scripts for each CI platform would avoid changing anything to the current CI, and so it would remove the TeamCity build out of the equation for getting a .Net Core/Standard version build under Linux or Macintosh.

But as you can read in #1508, no one was agreeing with me so I have just closed it. (I have not deleted its branch for letting anyone who would like see how that AppVeyor build was working.)

> Does the documentation need to be built on Linux?

I do not think we should have this requirement. Furthermore, its build has two main dependency: [SHFB](https://github.com/EWSoftware/SHFB) and [DocBook](http://docbook.org/). DocBook being Java based, it should be possible to build it under Linux or Macintosh. SHFB uses the Microsoft HTML Help Workshop, I doubt there is a cross-platform version of this.

hazzik:
>@hazzik had previously pushed for an MSBuild only solution, but must have had trouble coming through as he started using PSake. @fredericDelaporte had started a pull request around using AppVeyor (with PowerShell embedded in the AppVeyor YAML file).

MSBuild is still a prefered option. PSake has come to a radar because the changes I've played with were based on @fredericDelaporte PowerShell changes. 

>Also I think that in the past msBuild was used because there used to be some Teamcity builds. 

MSBuild was used because it's a native build system of .NET. NAnt was used because the project has lived longer than the MSBuild existed.

> you have to use command line build steps.

It is always a better option than vendor-locking to a build system either it would be TeamCity, AppVeyor, Travis-CI, Jenkins or any other system.

The incentive to move away from NAnt has come from several points: 

- NAnt is not actively maintained anymore.
- It does no run cross-platform [anymore](https://github.com/nant/nant/issues/161)
- The current script is not flexible enough to produce cross-platform build (it can be dealt with thou)
- The current script is not flexible enough to be run on different environments (eg. it has hard-coded connections strings in the `teamcity.build` file).


>I have not used the Nant build scripts at all in it because it was looking to me as more work than simply rewriting what they do with PS.
>As explained in this PR, I am not convinced striving to get a cross-platform CI build solution is worth it. I mean, I doubt it will be overall simpler than having a short dedicated script for each CI, like the one I have put in this AppVeyor PR. The Nant scripts we currently have handle both the CI and the developer workstation, but are at least five time as big. And they have still hard coded dependencies on how the things are setup on the machine, causing them to be quite far from being cross-platform too.

This is simply not accurate. The only thing is vendor-locked and environment-locked is a `teamcity.build` file. But, in reality the only thing it does is population of `current-configuration` folder for the chosen environment. On "a developer machine" this is usually done by `ShowBuildMenu.bat`. I was disagreeing with putting everything into AppVeyor.yml file. These vendor-locked solutions should contain _only_ code to prepare the environment and then call our build system. 

>By the way, taking the approach of ad-hoc scripts for each CI platform would avoid changing anything to the current CI, and so it would remove the TeamCity build out of the equation for getting a .Net Core/Standard version build under Linux or Macintosh.

No, it would vendor-lock us and stop migrating to other platform when the current platform will die. [And it WILL DIE eventually](http://disq.us/p/7id2cx).

fredericDelaporte:
> No, it would vendor-lock us and stop migrating to other platform when the current platform will die. And it WILL DIE eventually.

I do not see a vendor lock in using a single 335 lines build file specific to that platform. Migrating to another platform when NHibernate will have would to will be a similar, quite short work. And it will be an opportunity to upgrade the build by the way, since by the time AppVeyor come to pass away, it is very likely new practices and tooling, potentially better, would be available.

This would be true for a cross-platform, "non-vendor specific" solution too, excepted it will be more complex to evolve since it will have to retain that cross-platform capability. This means it will likely have more inertia, till it become too obsolete, a bit like our Nant scripts nowadays, causing us to have to rewrite it fully.

And by the way, "non-vendor specific" is an utopia in my mind: there will be always some parts needed to get run by the platform which will be specific to it. Maybe the specific part can be reduced to a one liner just launching NHibernate build scripts, which will then doing all the work without using any tooling supplied by the platform, up to the point of installing and setting up all the db we need for our tests. But I do not see how going this far could overall save us work compared to being fully vendor specific with a script as simple and short the platform allows for what we build.

I am still not convinced aiming at having a cross-platform, "non-vendor specific" build solution is a valid prerequisite to conceiving the .Net Core version of NHibernate. I think that renders the process more complicated, and that it will not have added value, causing more work overall.

iancooper:
Out of interest we are moving away from psake to cake. Partially that was motivated by the crisis in maintainers (I know Brandon has taken over from Gary now), but also because we perceive it is easier to maintain C# over Powershell. As we go x-platform we are busy unwinding our position on PS in favor of CLI written in Python or C#.

I have no idea if others are doing the same.

It is easier to go x-platform with Cake today, and it appears to have most of the community momentum.

Gary Ewan Park is willing to talk this through with you guys.

We'd be happy to help out with some Huddle resources, if we can get clear agreement on what work it would help if we took on.

iancooper:
That said, from an outsider's point of view I agree with @fredericDelaporte that you may want to determine if the Core milestone needs x-platform builds, or if you make that a later milestone. Iterative over big-bang. But I'm a chicken not a pig in this conversation, so...

jrgcubano:
We are back with FluentNHibernate and planing to follow the update to NetCore when NHibernate release the new version. We move the build system to build.cake too and AppVeyor.

Let me know if I can help on anything!

Thanks for the amazing work!

baumann74:
Hi guys. Any ideas about a release date? One month, three months, six months?

gep13:
As @iancooper mentioned, if there is anything that we the Cake Team can do to help with this, let me know.

Lionhunter3k:
Just tried this version with the latest oracle odp DLL (which supports .netstandard 2.0) and it works :D Had to use a separate Oracle driver, because the Reflection driver loaded the full framework one and kept getting DbFactoryProvider exceptions. In short, you can connect to an oracle DB through a .net core app, at this moment. Thanks for this library and this pull request, I still think nHibernate has something to say in 2018, can't wait to see it officially in a nuget package.

stephan-simondshealth-com-au:
Awesome work Guys and Girls !

Reading through this ticket, is "merging" all whats left to do?
Are there any takers from the people with write access?

(FYI: I'd be happy to chip in, but don't have the required street cred to unlock that level.... )

fredericDelaporte:
An alternate PR for .Net Core has been merged: #1523. It has less changes: drivers are not split in separated projects, binary serialization is not fully supported.

giggio:
@fredericDelaporte is this PR being abandoned or will it be updated to support these other things?

fredericDelaporte:
The two things I have cited already exist as separated PR. About this PR here, @ngbrown will likely check at some point if there is still things to do with it or if it should be just closed.

Diaskhan:
Could somebody provide demo example with net  core 2.0 ?

giggio:
I have one here @Diaskhan https://github.com/giggio/exemplonhdnetstandard2

ngbrown:
@Diaskhan the included web example works as well: <https://github.com/nhibernate/nhibernate-core/tree/master/src/NHibernate.Example.Web>

hazzik:
@ngbrown is there anything else from this PR that is not included into other PRs?

penderi:
@ngbrown I want to say thanks. That probably means little but you've struggled for a year or more with little feedback from the main NH team on progress or the fact other efforts were underway. 
Community involvement in OSS projects should be encouraged and helped. I'm not sure that's happened here. For over a year. I just want to say your gargantuan efforts have been appreciated. Thanks for trying and here's hoping the other PR and release is successful. 
Many thanks @ngbrown  I think we all owe you a beer!

ngbrown:
@penderi Thanks for the kind words.

ngbrown:
Since the core of this was implemented by #954, and I have pull requests for most outstanding differences, I'll close this.

The major outstanding difference is #626 which would split each database provider off into individual NuGet packages.  These NuGet packages and the built-in dependency resolution would enforce at build time what drivers can be used with what frameworks.  Without this, all the drivers have become reflection-based for the .NET Core framework builds.

The focus should shift to getting this final difference into a release.

PR #636, opened 2017-05-30
NH-4021 - enforce session tracking in tests

fredericDelaporte:
[NH-4021](https://nhibernate.jira.com/browse/NH-4021) - Track all opened session in tests

As illustrated with DtcFailuresFixture, some session "leaks" (indeed delayed releases) are going unnoticed due to the test by-passing the session tracking.
Adapt the base test class for ensuring this is no more possible, at least for sessions opened from session factory.

This is done by wrapping the session factory in delegating session factory. Some tests doing assumptions on the concrete session factory had to be adapted.

Lot of tests which were directly using the `TestBase` session factory field have been changed for using the property instead. Some tests which were changing the session factory were adapted too.

This does not handle sessions opened from other sessions. It would require a similar wrapping, and much more tests could be impacted by it.

### DIFF
--- a/src/NHibernate.Test/TestCase.cs
+++ b/src/NHibernate.Test/TestCase.cs
@@ -12,12 +12,8 @@ using NHibernate.Tool.hbm2ddl;
 using NHibernate.Type;
 using NUnit.Framework;
 using NHibernate.Hql.Ast.ANTLR;
-using System.Collections.Concurrent;
-using System.IO;
 using NUnit.Framework.Interfaces;
 using System.Text;
-using log4net.Util;
-using static NUnit.Framework.TestContext;
 
 namespace NHibernate.Test
 {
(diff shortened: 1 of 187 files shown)


src/NHibernate.Test/TestCase.cs
```diff
@@ -42,17 +38,7 @@ protected TestDialect TestDialect
...
-			get
-			{
-				return sessions.Settings.QueryTranslatorFactory is ASTQueryTranslatorFactory;
-			}
-		}
-
-		private ConcurrentBag<ISession> _openedSessions = new ConcurrentBag<ISession>();
-		private ISession _lastOpenedSession;
-		private DebugConnectionProvider connectionProvider;
+		protected bool IsAntlrParser => Sfi.Settings.QueryTranslatorFactory is ASTQueryTranslatorFactory;
```
hazzik: This is always true

fredericDelaporte: Was thinking it was very likely obsolete, but had not checked. Thanks for the cleanup.

PR #637, opened 2017-05-31
NH-4023 - Passing session to type getter and setter

fredericDelaporte:
[NH-4023](https://nhibernate.jira.com/browse/NH-4023) for solving time issues for some databases.
[NH-1851](https://nhibernate.jira.com/browse/NH-1851) MySql
[NH-3078](https://nhibernate.jira.com/browse/NH-3078) Sybase Anywhere
[NH-4017](https://nhibernate.jira.com/browse/NH-4017) Npgsql 3
[NH-4024](https://nhibernate.jira.com/browse/NH-4024) ODBC

Maybe a replacement for #635 and #631.

But has a heavy impact on types, which may hinders UserType as shown in test project: it seems frequent to delegate the set/get of an user type to a NHibernate type. But with that change, that will require the session. ~And the user type will not have it.~

### DIFF
--- a/src/NHibernate.DomainModel/NHSpecific/NullInt32UserType.cs
+++ b/src/NHibernate.DomainModel/NHSpecific/NullInt32UserType.cs
@@ -54,7 +54,7 @@ namespace NHibernate.DomainModel.NHSpecific
 			}
 			else
 			{
-				_int32Type.Set(cmd, value, index);
+				_int32Type.Set(cmd, value, index, null);
 			}
 		}
 
@@ -65,7 +65,7 @@ namespace NHibernate.DomainModel.NHSpecific
 
 		public object NullSafeGet(DbDataReader rs, string[] names, object owner)
 		{
-			return _int32Type.NullSafeGet(rs, names);
+			return _int32Type.NullSafeGet(rs, names, null);
 		}
 
 		public bool IsMutable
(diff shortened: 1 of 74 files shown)


hazzik:
I think ISessionImplementor can be too much.

hazzik:
ODBC driver also requires TimeSpan

hazzik:
On Hibernate side UserType has `SharedSessionContractImplementor session` parameter in `nullSafeGet` and `nullSafeSet` methods, so I think we should do the same.

fredericDelaporte:
> On Hibernate side `UserType` has `SharedSessionContractImplementor session`

I have seen that Hibernate was not shy about passing that around, enforcing my decision for `NullableType`. But I have had not checked what is was for `IUserType`, thanks

I am ok with changing it too. Sure breaking change, but not doing it would be worst I think.

fredericDelaporte:
> ODBC driver also requires TimeSpan

Thanks. Should I update the tests result files by the way, for avoiding hiding those newly fixed cases?

hazzik:
>Should I update the tests result files by the way, for avoiding hiding those newly fixed cases?

I think the goal shall be to have no red tests at all. So, we either need to fix them, or ignore them.

src/NHibernate.DomainModel/NHSpecific/NullInt32UserType.cs
```diff
@@ -63,9 +63,9 @@ public System.Type ReturnedType
 			get { return typeof(Int32); }
 		}
 
-		public object NullSafeGet(DbDataReader rs, string[] names, object owner)
+		public object NullSafeGet(DbDataReader rs, string[] names, ISessionImplementor session, object owner)
```
hazzik: :+1:

PR #639, opened 2017-06-02
NH-4026 - upgrade Firebird driver and use server

fredericDelaporte:
[NH-4026](https://nhibernate.jira.com/browse/NH-4026) - upgrading Firebird driver and using server instead of embedded mode.


hazzik:
I think we need to replace whole `teamcity/lib` folder with the NuGet sourced packages. It's not easy achievable, but it will allow greater manageability.

fredericDelaporte:
> I think we need to replace whole `teamcity/lib` folder with the NuGet sourced packages. It's not easy achievable, but it will allow greater manageability.

Could be easily done if we just add them all bluntly to the test project, provided they do not conflict.

fredericDelaporte:
Ok, I had forgotten about the `TestDatabaseSetup` which is run only on TeamCity.

hazzik:
> Could be easily done if we just add them all bluntly to the test project, provided they do not conflict.

No, it's not so easy. There is also ShowBuildMenu.bat and "available-test-configurations"/"current-test-configuration"

fredericDelaporte:
It will just have to move around `hibernate.cfg`, and no more libs, if the test project have directly all references.

ngbrown:
I had been working on why huge numbers of tests for Firebird were failing in #626.  I had updated to the most recent Firebird NuGet package.

It was being caused by either 

```
TearDown : NHibernate.HibernateException : unsuccessful metadata update
object INVERSED_STUFF is in use
```

or sometimes

```
TearDown : NHibernate.HibernateException : lock conflict on no wait transaction
unsuccessful metadata update
object HIBERNATE_UNIQUE_KEY is in use
```

It wasn't consistent which specific test would fail to tear down, but once it happened, it would start a massive chain reaction and fail lots of tests after that. 

I had narrowed it down to a change in the driver from 2.7.7 to 3.0.0.  It looks like you encountered the same and solved it with clearing the connection pool.

ngbrown:
As for the embedded server, I don't see why not to continue to use it for unit tests... It's easier to setup that way, much like how SQLite is.

fredericDelaporte:
I am doing that for testing DTC issues with as much as possible databases. I am not sure Firebird will behave the same about distributed transactions in embedded mode. True, due to a bunch of others Firebird issues, it will likely stay un-testable for DTC.

Moreover, SQLite is meant to be used only as embedded, but not Firebird. Firebird is a more classical database server, likely more frequently used in server mode than in embedded mode. I rather test its main use case than what is, I think, a "side" use case.

I have posted this [question](https://stackoverflow.com/q/44353607/1178314) on Stack Overflow about the Firebird trouble with dropping table and connection pooling. If you have spotted the exact change having caused that, maybe could it be of interest you tell it in a comment.

Now there are still three newly failing test (and two old failures), will probably check them tomorrow.

ngbrown:
@fredericDelaporte I hadn't figured out the exact cause yet.  It seems you are a bit ahead of me already having a solution.  I had just narrowed it down to what version had the change.  I have the suspicion it has something to do with this change [DNET-337](http://tracker.firebirdsql.org/browse/DNET-337), but not really sure.

A note on your stack overflow question; the problems I'm testing against is that it's not actually the server version that breaks the tests, but the driver version that makes the difference.  I've been testing with various versions of the driver, still with the same embedded server that's been part of the NHibernate repository.

hvlad:
@ngbrown i'm not expert in .Net but it seems that change of wait mode (from "wait" to "nowait") by DNET-337 could lead to such behaviour in DDL transactions (if tests uses default transactions mode). 
Workaround is easy (i think) - use explicit parameteres for (at least) DDL transaction.
Hope it helps

fredericDelaporte:
@hvlad, maybe not that easy to do. I do not really understand what are those parameters you are referring too. They are likely Firebird concepts, but NHibernate schema creation/drop logic is database agnostic. We are limited in what we can additionally do specifically for a database. Currently for this trouble we have [this workaround](https://github.com/nhibernate/nhibernate-core/pull/639/commits/96ac0d395de99ad2b142876b17f7eb88fad127d0#diff-808ab68d694e3b0427ed6c179f2f6be7R274) which seems to do the job.

fredericDelaporte:
Now:
- upgraded (driver from 2.5.2 to 9.5.1) + server (from embedded 2.5.2 + a (unused I think) 3.0.1 to a 3.0.2 server)
- using server mode instead of embedded mode
- no more additional libraries with driver
- no more failing test at all
- previous test result removed and check against previously failed removed

src/NHibernate.Test/NHSpecificTest/NH3374/FixtureByCode.cs
```diff
@@ -15,9 +13,11 @@ protected override HbmMapping GetMappings()
 			mapper.Class<Document>(rc =>
 			{
 				rc.Id(x => x.Id, idMapper => idMapper.Generator(Generators.Identity));
-				rc.ManyToOne(x => x.Blob, m =>
+				rc.ManyToOne(x => x.Blob,
+					m =>
 					{
 						m.Cascade(Mapping.ByCode.Cascade.All);
+						m.Column("`Blob`");
```
hazzik: I think the correct <tm> way to do this is to mark "blob" as [a keyword for Firebird](http://www.firebirdsql.org/refdocs/langrefupd25-reskeywords-full-keywords.html) dialect.

fredericDelaporte:
On Teamcity side, should the failure condition `at least one test failed` be enabled? I have compared with other build not having test results comparison, only `NHibernate` build has it enabled. `Sql Server 2012`, `SQLite` and `PostgreSQL` do not have it enabled.

teamcity.build
```diff
@@ -88,30 +88,24 @@
...
-		<property name="nhibernate.connection.connection_string"	value="Database=NHibernate.fdb;ServerType=1;UserID=SYSDBA" />
+		<property name="nhibernate.connection.connection_string"	value="DataSource=localhost;Database=nhibernate;User ID=SYSDBA;Password=masterkey;MaxPoolSize=200;" />
 		<copy todir="${bin.dir}">
-			<fileset basedir="${root.dir}/lib/teamcity/firebird/x64">
+			<fileset basedir="${root.dir}/lib/teamcity/firebird">
 				<include name="*.dll"/>
-				<include name="*.msg"/>
 			</fileset>
 		</copy>
-		<property name="NHibernate.Test.IgnoreFail" value="true" />
```
hazzik: 👍

src/NHibernate.Test/NHSpecificTest/NH3374/FixtureByCode.cs
fredericDelaporte: protected virtual void RegisterKeywords()
		{
			RegisterKeyword("date");
		}

Maybe a lot more are missing...

fredericDelaporte: And does not work anyway. Test failing again.

fredericDelaporte: Those keywords are seldom used in the code base. Only the `Template` class uses them, and not for quoting.

hazzik: Why "date"? it's "blob"

fredericDelaporte: That is the current only registered word. For my test I have put "blob".

fredericDelaporte:
There is a trouble with server mode. It seems damn slow... Previous failures are due to execution timeout. There is one build in server mode which has succeeded, in 45 minutes...

fredericDelaporte:
Digging a bit more for checking the embedded v3 perf: in v3, embedded mode has change. We now have to take different files. And "embedded" may still connect to the local server. Currently I am stuck having it connecting my local server then losing connection. Here is the [most complete instructions](https://www.riedmann.it/blog/?p=270) I have found about using embedded mode with Firebird v3. (On official site, there is just a small note in release notes from which we are supposed to guess the steps for using embedded mode with v3. Doc for embedded mode itself is still referring to the old, incompatible way.)

This database is a nightmare to get running and up to date. Is it worth it?

We may stay with this PR current state, but with an increased timeout for Firebird build. I suggest at least two hours...

hazzik:
Yeah, this is why I have installed the server to the test agent - it seemed to be required when I was testing Firebird locally.

hazzik:
Try some of these before increasing the limit: https://ib-aid.com/en/articles/45-ways-to-speed-up-firebird-database/

It's not really acceptable have a timeout of 2 hours.

hazzik:
@amroel can you also take a look?

ngbrown:
For embedded mode, I think you are probably missing the `ClientLibrary=fbclient.dll;` part of the connection string.

I've gotten v3.0.2 working in embedded mode.  See https://github.com/ngbrown/nhibernate-core/commit/c9304718730d0728142a94f0b727979def22601a

It's currently down to 49 failures and taking 25 minutes (on my laptop).

fredericDelaporte:
@ngbrown, I have tested with `ClientLibrary=fbclient`, not sure adding `.dll` will change anything. Without that param, the error is it tries to load `embed` instead, and I am not having that error. Maybe a specific trouble with my own local server. I will check you branch anyway, thanks.

On my own laptop in server mode, the test takes 40 minutes, so no that far from your 25 minutes in embedded mode: it could be just hardware differences. While it was taking 6 to 7 minutes on build agent with 2.5.2 version. Have you ascertain it was actual embedded? I test that by stopping Firebird service. In such case, in pseudo-embedded mode where the `fbclient.dll` just acts as a native client to the server, I get a meaningful error: `Unable to complete network request to host "xnet://Global\FIREBIRD".`. When I have the service up, I was having instead `connection lost to database`. But now I have found "why": my test runner had switch back to 64bits without me noticing... Now I get it working, and with Firebird service down, so very likely as actually embedded. Waiting to have some timings.

@hazzik, Yes a big timeout would not be a good thing. If we had to do that, we should then probably schedule such build only for running daily at most, for reducing its impact on build agent availability.

fredericDelaporte:
> Try some of these before increasing the limit: https://ib-aid.com/en/articles/45-ways-to-speed-up-firebird-database/

On my laptop, the bottleneck in embedded mode seems to be the disk (although it is a SSD, M.2 sata 3 port). So their parameters for increasing cache and reducing writes may help, if we can apply them in embedded mode. I will check them in server mode too.

It seems that in embedded mode, the provider has some differences in behavior, as [Nathan fix for boolean](https://github.com/ngbrown/nhibernate-core/commit/c9304718730d0728142a94f0b727979def22601a#diff-36c41df4e56bd3cac4dfc0da0ff73cfcR285) in its branch seems to tell. This may require adding an embedded dialect...

fredericDelaporte:
None of their parameters were having significant impact, excepted changing the database itself for disabling forced writes: duration drops to 7 minutes on my laptop. Committing that change. Forced writes are enabled by default since long in Firebird. I guess the old data provider was disabling it instead of following Firebird default.

Firebird advice strongly against disabling it. They consider its disabling as the major source of corrupted databases. Since we recreate it at each run, that is not an issue for us.

fredericDelaporte:
Still 40 minutes. Trying some tweaking directly on server.

fredericDelaporte:
15 minutes now: I have manually deleted the database. It looks like overwrite option is not able of changing forced writes. So for staying on the safe side, I will remove my simplification and go back to "try drop then create". I will try to gain even more performances server side by moving the db file to `D:`: it is supposed to be a local drive according to Azure documentation, while `C:` is in the network. (Such a move cannot be done for much over databases, because they will likely not support losing unexpectedly a database after a reboot, while Firebird does not care.)

@hazzik, I think we have an additional trouble on the build agent. Oracle take 1Gb of memory alone (counting Java in it, maybe I am wrong doing so, but well, then it takes 800Mb). The server has only 3.5Gb, so that is a bit problematic. On my laptop, using the same XE 11 I believe, it takes only 200Mb. I have already restarted the service on the agent, no better. Maybe some maintenance is needed on that Oracle database, but I do not know much about that. 
Side note: SQL-Server 2012 was failing most tests with timeouts. Restarting its service seems to have solved that issue.

lib/teamcity/firebird/firebird_installation.txt
```diff
@@ -8,7 +8,7 @@ Installation steps for Firebird for NH TeamCity:
 5.  Go into Firebird folder (c:\program files\firebird\) and create a folder named Data;
 6.  Go in Firebird installation directory and open databases.conf;
 7.  Add in "Live Databases" section:
-nhibernate = C:\Program Files\Firebird\Data\nhibernate.fdb
+nhibernate = D:\SqlData\Firebird\nhibernate.fdb
```
hazzik: Why not make it local?

src/NHibernate.TestDatabaseSetup/TestDatabaseSetup.cs
```diff
@@ -102,8 +102,17 @@ private static void SetupSqlServerOdbc(Cfg.Configuration cfg)
 
 		private static void SetupFirebird(Cfg.Configuration cfg)
 		{
+			Directory.CreateDirectory(@"D:\SqlData\Firebird");
```
hazzik: I don't like references to a specific folder.  I think we need to use a local (relative) folder. Building with VS and NAnt+MSBuild will produce different results. When building with NAnt the output folder would be /build/version/net40 (or something like this). Building with VS will produce tests assemblies in different folders (/src/NHibernate.Test/bin/Debug-2.0/, /src/NHibernate.TestDatabaseSetup/bin/Debug-2.0/, etc). We need to choose some common folder, probably `"curent-test-configuration"`, or a repository root.

I think we shall/should parse connection string to find where the file sits.

hazzik:
>On my laptop, using the same XE 11 I believe, it takes only 200Mb.

It's 12c

lib/teamcity/firebird/firebird_installation.txt
fredericDelaporte: Local to the database server? Would be on `C:`, which is not local to the host. Local to the NHibernate test client? The path seems not frozen, only embedded mode can do that, where the client is the server too. I think Firebird mainstream usage is as an actual server, not as embedded mode, so I would rather test it as an actual server. And as showcase by Nathan branch, Firebird v3 embedded mode seems to need a new dialect.

When using a database server, does it really make sens to locate the database relatively to the client? That should be the server choice, not the client. Currently this is done with a configured alias in FireBird_3_0\Databases.conf file, which decides where to put the file for the alias `nhibernate`.

src/NHibernate.TestDatabaseSetup/TestDatabaseSetup.cs
fredericDelaporte: The connection string does currently only hold the alias `nhibernate`. The actual path is in Databases.conf file, in Firebird installation directory. Yes this code line is a crutch for accounting the possible loss of D: content.

When using a database server, it does not really make sens for me to locate the database relatively to the client, that is usually the server choice. Firebird allows to specify the database location from the connection string, which is a strange model for a server, not at all common. So we may do that anyway, by detecting where the test assemblies are then tweaking the connection string for telling the server where to put the file. But that would depart from what is and can be done for other actual servers like Oracle, Sql-Server, PostgreSql, which gives no option to the client about where to locate the database.
And we would have to do that both in `TestDatabaseSetup` and the test project.

Otherwise we should go back to testing Firebird in embedded mode, with all its additional binaries and apparently some new behaviors in v3 requiring a dedicated dialect. (There are seven hundred failing tests when running in embedded mode with Firebird v3, most solved by redefining the boolean type as have done Nathan. I have not yet investigated the other cases, since I do not consider we should test Firebird as an embedded server.)

fredericDelaporte:
> It's 12c

The test settings for Oracle are using the 10g dialect, while a 12c one is available. That why I was believing it was the free XE 11 edition.

fredericDelaporte:
About Firebird, locating that db on `D:` seems worst than on `C:`, although the temporary drive is supposed to be more performant than the system drive accoding to [msdn](https://blogs.msdn.microsoft.com/mast/2013/12/06/understanding-the-temporary-drive-on-windows-azure-virtual-machines/):

> This temporary storage drive is present on the physical machine which is hosting your VM and hence can have higher IOPS and lower latency when compared to the persistent storage like data disk.

The current run has taken 34 minutes to complete. Going to launch a new one in case that was some temporary degradation.

src/NHibernate.TestDatabaseSetup/TestDatabaseSetup.cs
hazzik: The problem, that people might not necessarily have a "D:" drive, and this would fail. I usually use TestDb to create a fresh copy of a database (including Firebird) myself.

fredericDelaporte: So I may just locate the file on d:\ root if that prove to be any better than putting it within Firebird installation folder. That would avoid that line. Actually I was seeing `TestDatabaseSetup` as a Teamcity exclusive thing, managing myself my own local db. So I would avoid to do such changes now.

hazzik: It still be executed if you run from ShowBuildMenu (release option, for ex)

fredericDelaporte: Ok, I have missed that, sorry.

fredericDelaporte:
Last run on D:, 7'36'', now launching another run on C: (best time was 15' for C:).

fredericDelaporte:
Checking the 34' run on D:, publishing artifacts had taken 22' alone, so that was mostly unrelated to Firebird. Current run waiting the merge builds I have triggered from NH-4013. It seems you have just taken the session I was having on the build agent.

hazzik:
You can take the session back:)

fredericDelaporte:
IMO, this is now ready for squashing&merging.

hazzik:
Merged

fredericDelaporte:
Thanks Alexsander.
I have just disconnected (instead of log-off) from a session on the build agent where a PostgreSql install was finished. I have left it as is.

PR #640, opened 2017-06-04
NH-4027 - Missing disposals of enumerators.

fredericDelaporte:
[NH-4027](https://nhibernate.jira.com/browse/NH-4027) - Missing disposals of enumerators.

It appears NHibernate heavily uses `IEnumerator` directly, and does not dispose them in many cases.
This is likely a side effect of Java porting code. But in .Net, many cases can be replaced by LINQ calls or `foreach` loops, and should be.
Otherwise, extra care must be taken in order to ensure their are disposed of.
`IEnumerator<T>` is directly disposable, but the old `IEnumerator` is not, while the underlying concrete type will be `IDisposable` most of the time. The `foreach` construct handles it by try-casting it to `IDisposable`, and we should do the same.

Done in NHibernate project. Test project also have many cases, not fixed.

### DIFF
--- a/src/NHibernate/AdoNet/Util/BasicFormatter.cs
+++ b/src/NHibernate/AdoNet/Util/BasicFormatter.cs
@@ -59,14 +59,15 @@ namespace NHibernate.AdoNet.Util
 
 		public virtual string Format(string source)
 		{
-			return new FormatProcess(source).Perform();
+			using (var fp = new FormatProcess(source))
+				return fp.Perform();
 		}
 
 		#endregion
 
 		#region Nested type: FormatProcess
 
-		private class FormatProcess
+		private class FormatProcess : IDisposable
 		{
 			private readonly List<bool> afterByOrFromOrSelects = new List<bool>();
 			private readonly List<int> parenCounts = new List<int>();
@@ -440,6 +441,11 @@ namespace NHibernate.AdoNet.Util
 					result.Append(IndentString);
 				}
 				beginLine = true;
+			}
+
+			public void Dispose()
+			{
+				tokens.Dispose();
 			}
 		}
 
--- a/src/NHibernate/Dialect/MsSql2012Dialect.cs
+++ b/src/NHibernate/Dialect/MsSql2012Dialect.cs
@@ -58,31 +58,33 @@ namespace NHibernate.Dialect
 
 		public override SqlString GetLimitString(SqlString querySqlString, SqlString offset, SqlString limit)
 		{
-			var tokenEnum = new SqlTokenizer(querySqlString).GetEnumerator();
-			if (!tokenEnum.TryParseUntilFirstMsSqlSelectColumn()) return null;
-
-			var result = new SqlStringBuilder(querySqlString);
-			if (!tokenEnum.TryParseUntil("order"))
+			using (var tokenEnum = new SqlTokenizer(querySqlString).GetEnumerator())
 			{
-				result.Add(" ORDER BY CURRENT_TIMESTAMP");
-			}
+				if (!tokenEnum.TryParseUntilFirstMsSqlSelectColumn()) return null;
 
-			result.Add(" OFFSET ");
-			if (offset != null)
-			{
-				result.Add(offset).Add(" ROWS");
-			}
-			else
-			{
-				result.Add("0 ROWS");
-			}
+				var result = new SqlStringBuilder(querySqlString);
+				if (!tokenEnum.TryParseUntil("order"))
+				{
+					result.Add(" ORDER BY CURRENT_TIMESTAMP");
+				}
 
-			if (limit != null)
-			{
-				result.Add(" FETCH FIRST ").Add(limit).Add(" ROWS ONLY");
-			}
+				result.Add(" OFFSET ");
+				if (offset != null)
+				{
+					result.Add(offset).Add(" ROWS");
+				}
+				else
+				{
+					result.Add("0 ROWS");
+				}
 
-			return result.ToSqlString();
+				if (limit != null)
+				{
+					result.Add(" FETCH FIRST ").Add(limit).Add(" ROWS ONLY");
+				}
+
+				return result.ToSqlString();
+			}
 		}
 	}
 }
--- a/src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
+++ b/src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
@@ -81,7 +81,7 @@ namespace NHibernate.Event.Default
 					.Append(" removals to ").Append(persistenceContext.CollectionEntries.Count).Append(" collections");
 
 				log.Debug(sb.ToString());
-				new Printer(session.Factory).ToString(persistenceContext.EntitiesByKey.Values.ToArray().GetEnumerator());
+				new Printer(session.Factory).ToString(persistenceContext.EntitiesByKey.Values.ToArray());
 			}
 		}
 
--- a/src/NHibernate/Hql/QuerySplitter.cs
+++ b/src/NHibernate/Hql/QuerySplitter.cs
@@ -113,7 +113,7 @@ namespace NHibernate.Hql
 				templateQuery.Append(token);
 			}
 			string[] results =
-				StringHelper.Multiply(templateQuery.ToString(), placeholders.GetEnumerator(), replacements.GetEnumerator());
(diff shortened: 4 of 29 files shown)


src/NHibernate/AdoNet/Util/BasicFormatter.cs
```diff
@@ -441,6 +442,11 @@ private void Newline()
 				}
 				beginLine = true;
 			}
+
+			public void Dispose()
+			{
+				tokens.Dispose();
```
fredericDelaporte: This one is a bit nit-picking: concrete implementation of this `Dispose` is currently no-op.

src/NHibernate/Dialect/MsSql2012Dialect.cs
```diff
@@ -58,31 +58,33 @@ protected override void RegisterFunctions()
 
 		public override SqlString GetLimitString(SqlString querySqlString, SqlString offset, SqlString limit)
 		{
-			var tokenEnum = new SqlTokenizer(querySqlString).GetEnumerator();
-			if (!tokenEnum.TryParseUntilFirstMsSqlSelectColumn()) return null;
-
-			var result = new SqlStringBuilder(querySqlString);
-			if (!tokenEnum.TryParseUntil("order"))
+			using (var tokenEnum = new SqlTokenizer(querySqlString).GetEnumerator())
```
fredericDelaporte: Only significant change, but with the indent impact, Github mix it all.

src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
```diff
@@ -81,7 +81,7 @@ protected virtual void FlushEverythingToExecutions(FlushEvent @event)
 					.Append(" removals to ").Append(persistenceContext.CollectionEntries.Count).Append(" collections");
 
 				log.Debug(sb.ToString());
-				new Printer(session.Factory).ToString(persistenceContext.EntitiesByKey.Values.ToArray().GetEnumerator());
+				new Printer(session.Factory).ToString(persistenceContext.EntitiesByKey.Values.ToArray());
```
fredericDelaporte: Pattern change: take an `IEnumerable` instead of a disposable thing in "fire&forget" mode.

src/NHibernate/Hql/QuerySplitter.cs
```diff
@@ -113,7 +113,7 @@ public static string[] ConcreteQueries(string query, ISessionFactoryImplementor
 				templateQuery.Append(token);
 			}
 			string[] results =
-				StringHelper.Multiply(templateQuery.ToString(), placeholders.GetEnumerator(), replacements.GetEnumerator());
+				StringHelper.Multiply(templateQuery.ToString(), placeholders, replacements);
```
fredericDelaporte: Same as previous comment.

src/NHibernate/Impl/EnumerableImpl.cs
```diff
@@ -15,7 +15,11 @@ namespace NHibernate.Impl
 	/// Provides an <see cref="IEnumerable"/> wrapper over the results of an <see cref="IQuery"/>.
 	/// </summary>
 	/// <remarks>
-	/// This is the IteratorImpl in H2.0.3
+	/// <para>This is the IteratorImpl in H2.0.3</para>
+	/// <para>This thing is scary. It is an <see cref="IEnumerable" /> which returns itself as a <see cref="IEnumerator" />
+	/// when <c>GetEnumerator</c> is called, and EnumerableImpl is disposable. Iterating over it with a <c>foreach</c>
+	/// will cause it to be disposed, probably unexpectedly. (https://stackoverflow.com/a/11179175/1178314)
+	/// "Fortunately", it does not currently support multiple iterations anyway.</para>
```
fredericDelaporte: Yuck.

src/NHibernate/Impl/FilterImpl.cs
```diff
@@ -93,10 +93,19 @@ public IFilter SetParameterList(string name, ICollection values)
 			if (values.Count > 0)
 			{
 				var e = values.GetEnumerator();
-				e.MoveNext();
-				if (!type.ReturnedClass.IsInstanceOfType(e.Current))
+				try
```
hazzik: This looks like Any or FirstOrDefault

hazzik: I think it shall be Any

src/NHibernate/Mapping/ForeignKey.cs
```diff
@@ -122,12 +122,14 @@ private void AlignColumns(Table referencedTable)
...
 			}
-			IEnumerator<Column> fkCols = ColumnIterator.GetEnumerator();
-			IEnumerator<Column> pkCols = referencedTable.PrimaryKey.ColumnIterator.GetEnumerator();
 
-			while (fkCols.MoveNext() && pkCols.MoveNext())
+			using (var fkCols = ColumnIterator.GetEnumerator())
+			using (var pkCols = referencedTable.PrimaryKey.ColumnIterator.GetEnumerator())
 			{
-				fkCols.Current.Length = pkCols.Current.Length;
+				while (fkCols.MoveNext() && pkCols.MoveNext())
```
hazzik: This looks like Zip

src/NHibernate/Impl/FilterImpl.cs
fredericDelaporte: Both `FirstOrDefault` and `Any` would change the exact technical semantic of this code. It is currently "if there are any elements (`values.Count > 0`), take the first and check it has an appropriate type". `FirstOrDefault` does not let us know whether there was no element or there was one but the it was the default one. I have not put a `First` here because I should have issue a `Cast<object>()` before for using `First`.

Now if we change that to a test of all supplied values, not only of the first one if any, why not using `Cast<object>().Any(...)`.

src/NHibernate/Mapping/ForeignKey.cs
fredericDelaporte: Granted, I am not familiar with `Zip`. Using it would be awkward. It does produce a third sequence we do not need here, and we would have to artificially iterate over it while using a result operator having some side effects. Something like:

    fkCols.Zip(pkCols, (fkc, pkc) => fkc.Length = pkc.Length).Count();

That should work but looks quite ugly and counter-intuitive to me (and will probably cause some warnings).

src/NHibernate/Impl/FilterImpl.cs
fredericDelaporte: I think we should modernize the whole method: cease taking a non generic collection parameter, but take a generic one.

src/NHibernate/Dialect/MsSql2012Dialect.cs
hazzik: You can add `?w=1` to the query string to ignore whitespace changes

fredericDelaporte: Thanks for te trick, I did not know.

src/NHibernate/Impl/FilterImpl.cs
fredericDelaporte: It looks like is doable without it being a sure breaking-changes: most users are probably already using something implicitly castable to `ICollection<T>`. Changing it does not require any other code change in NHibernate, test project included. Maybe is-it time to purge all our interfaces from old non generic `IEnumerable`, `ICollection`, `IList`. It would be fine to remove them from internal implementations too, but this could be heavier to do.

src/NHibernate/Mapping/SimpleValue.cs
```diff
@@ -1,6 +1,6 @@
 using System;
-using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
```
fredericDelaporte: Fix an unexpected use of `Util.EnumerableExtension.First` instead of the Linq one. Those extensions are crutches for the use of non generic `IEnumerable`, which are near from not being used anymore. Mainly need to convert some other `SetParameterList` to a generic form for getting rid of them.

src/NHibernate/Impl/FilterImpl.cs
fredericDelaporte: This change is done. If it is ok for you, I will merge, and add another Jira for doing same on other SetParameterList currently not generic. (And maybe on all other non generic case.)

PR #641, opened 2017-06-05
NH-4028 - Supports inconclusive tests in result comparison

fredericDelaporte:
[NH-4028](https://nhibernate.jira.com/browse/NH-4028) - Supports inconclusive tests in result comparison

Within #627, I have started issuing NUnit warning about problematic database behaviors we cannot fix. This causes the test to be inconclusive: it executes without errors but exhibit abnormal behaviors.
Current comparison result logic classifies them as failing. Teamcity classifies them as succeeded. The build log contains their warning.
Adjust comparison result logic to classify them as inconclusive and report them accordingly in comparison result. Do not cause the build to fail on them.

An example of such behavior causing a test to be inconclusive is the delayed commit with Oracle when the transaction is distributed, occurring after transaction scope disposal, instead of being committed before the disposal is left. The commit succeeded, but a query done right after the scope may not be able to see it. The test in #627 sleep a bit to check if the data finally gets committed. If no, it fails, if yes, it warns about the delayed commit, causing the test to be inconclusive.

### DIFF
--- a/teamcity.build
+++ b/teamcity.build
@@ -369,7 +369,7 @@
 				report.AppendLine("==================");
 
 				var before = Result.ParseFile(lastResult);
-				var beforeByName = before.ToDictionary(b => b.Name);
+				var beforeByName = before.GroupBy(b => b.Name).ToDictionary(g => g.Key, g => g.First());
 				var after = Result.ParseFile(currentResult);
 
 				var newFailingTests = new List<Result>();
--- a/teamcity.build
+++ b/teamcity.build
@@ -236,6 +236,7 @@
 	<script language="C#" prefix="testResult">
 		<references>
 			<include name="System.dll" />
+			<include name="System.Core.dll" />
 			<include name="System.Data.dll" />
 			<include name="System.Xml.dll" />
 		</references>
--- a/teamcity.build
+++ b/teamcity.build
@@ -362,14 +362,28 @@
 		{
 			try
 			{
-				var outputFile = Path.GetDirectoryName(currentResult) + "/Comparison.txt";
+				var outputFile = Path.Combine(Path.GetDirectoryName(currentResult), "Comparison.txt");
 				var report = new StringBuilder();
 
 				report.AppendLine("Comparison Results");
 				report.AppendLine("==================");
 
 				var before = Result.ParseFile(lastResult);
-				var beforeByName = before.GroupBy(b => b.Name).ToDictionary(g => g.Key, g => g.First());
+				var beforeByName = before
+					// Some multi-value tests may be duplicated
+					.GroupBy(b => b.Name)
+					.ToDictionary(
+						g => g.Key,
+						// Take one, preferably not failing first, thus ensuring an after failing test
+						// will be match as broken in case one of the dup was not failing.
+						g => g
+							// Succeeded
+							.OrderByDescending(r => r.Success)
+							// Else ignored
+							.ThenBy(r => r.Executed)
+							// Else inconclusive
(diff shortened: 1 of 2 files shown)


teamcity.build
```diff
@@ -245,231 +246,278 @@
...
+			public string Name;
+			public ResultStatus Status;
+			public bool Executed;
+			public bool Success;
+
+			public bool Inconclusive
+			{
+				get
+				{
+					return Executed && !Success && Status == ResultStatus.Success;
```
fredericDelaporte: Issuing a NUnit warning causes the test to have `Success` as `result` attribute value with `false`  as `success` attribute value. So I have added an `Inconclusive` property matching these on the `Result` class.

teamcity.build
```diff
@@ -245,231 +246,278 @@
...
+			try
+			{
+				var outputFile = Path.GetDirectoryName(currentResult) + "/Comparison.txt";
+				var report = new StringBuilder();
+
+				report.AppendLine("Comparison Results");
+				report.AppendLine("==================");
+
+				var before = Result.ParseFile(lastResult);
+				var beforeByName = before.GroupBy(b => b.Name).ToDictionary(g => g.Key, g => g.First());
```
fredericDelaporte: Some existing result file have duplicated tests. The previous code was matching new tests to each of them, with those new tests ending up potentially duplicated in the results lists below.

Here I simply take the first. The dup test is currently always `NHibernate.Test.Linq.WhereTests.CanUseCompareInQuery(p => (p.ShippingWeight.CompareTo(4.98) <= 0),17,False)`. I am going to fix its test case, it does it twice. All outcome were same, but just in case, il will change the logic above for considering non-failing results first.

teamcity.build
```diff
@@ -245,231 +246,278 @@
...
+				var after = Result.ParseFile(currentResult);
+
+				var newFailingTests = new List<Result>();
+				var newInconclusiveTests = new List<Result>();
+				var newNotFailingTests = new List<Result>();
+				var fixedTests = new List<Result>();
+				var ignoredTests = new List<Result>();
+				var brokenTests = new List<Result>();
+				var inconclusiveTests = new List<Result>();
+				foreach (var afterResult in after)
```
fredericDelaporte: Now aggregating results in one pass. Previous was looping on `after`, then on `before` for matching, then adding in one result list, and start over for the next list... With the growing number of tests, a O(n²) logic may end up some day taking a significant duration.

teamcity.build
```diff
@@ -245,231 +246,278 @@
...
+				if (brokenTests.Count > 0)
+				{
+					foreach (var result in brokenTests)
+						report.AppendLine(result.ToString());
+				}
+				else
+					report.AppendLine("None");
+
+				report.AppendLine();
+				report.AppendLine("*** Tests new (failed) since last recorded results ***");
```
fredericDelaporte: I have moved the failed tests at start of file. Seeking them in the middle of other was not very practical.

fredericDelaporte:
Any thoughts about this? That is probably the last piece missing for ending my work on #627. (Since I will not get Firebird into dtc tests unless at least one of its data provider issues ([DNET-764](http://tracker.firebirdsql.org/browse/DNET-764), [DNET-766](http://tracker.firebirdsql.org/browse/DNET-764)) is fixed fast.)

Of course I may instead give up that "inconclusive" status instead and ignore corresponding tests for Oracle, but that will cause Oracle to not be thoroughly tested with DTC. And if we do that, I think I will keep this PR, expurge from that inconclusive status, but keeping the other changes done for avoiding a O(n²) logic with `n` being the number of tests.

PR #588, opened 2017-03-31
[WIP] Async generator

hazzik:
Some things left on the NHibernate side:

- [x] We need to define how the cancellation token parameter will be generated ([current proposal](https://github.com/maca88/AsyncGenerator/pull/11))
- [x] Some warnings are produced because of the `<include />` tag ([example](https://github.com/nhibernate/nhibernate-core/blob/7ba52d2a6a4e2ea702c329470a27444bcbdb2466/src/NHibernate/Type/IType.cs#L16)), why not use `<inheritdoc />` tag instead?
- [x] Create a tool for generating the code (what options should the tool have?)
- [x] Add missing `ConfigureAwait(false)` for the manually written code
- [x] Replace `[ThreadStatic]` with `AsyncLocal<>` on the `SessionIdLoggingContext`
- [x] Add missing `CancellationToken` parameters to manually written async methods (linq extensions and future)
- [x] Fix broken tests for Firebird
- [x] Fix broken tests for Oracle
- [x] Fix broken tests for PostgreSQL (https://github.com/nunit/nunit/pull/2046)
- [x] Fix broken tests for SqlServer 2012 ([NH-4022](https://nhibernate.jira.com/browse/NH-4022))
- [x] Fix broken tests for MySQL

PR #642, opened 2017-06-12
NH-4030 - Refactoring future extension methods.

fredericDelaporte:
[NH-4030](https://nhibernate.jira.com/browse/NH-4030) - Cleanup and xml doc of Linq Future extension

Seen in #588, the Linq future methods lack xml comments and have not very sound preconditions checks.
Some clean-up and commenting would be welcome.

### DIFF
--- a/src/NHibernate/Impl/AbstractQueryImpl.cs
+++ b/src/NHibernate/Impl/AbstractQueryImpl.cs
@@ -886,7 +886,7 @@ namespace NHibernate.Impl
 		{
 			if (!session.Factory.ConnectionProvider.Driver.SupportsMultipleQueries)
 			{
-				return List<T>();
+				return new DelayedEnumerator<T>(List<T>);
 			}
 
 			session.FutureQueryBatch.Add<T>(this);
--- a/src/NHibernate/Linq/DefaultQueryProvider.cs
+++ b/src/NHibernate/Linq/DefaultQueryProvider.cs
@@ -13,7 +13,8 @@ namespace NHibernate.Linq
 {
 	public interface INhQueryProvider : IQueryProvider
 	{
-		object ExecuteFuture(Expression expression);
+		IEnumerable<TResult> ExecuteFuture<TResult>(Expression expression);
+		IFutureValue<TResult> ExecuteFutureValue<TResult>(Expression expression);
 		void SetResultTransformerAndAdditionalCriteria(IQuery query, NhLinqExpression nhExpression, IDictionary<string, Tuple<object, IType>> parameters);
 	}
 
@@ -59,12 +60,32 @@ namespace NHibernate.Linq
 			return new NhQueryable<T>(this, expression);
 		}
 
-		public virtual object ExecuteFuture(Expression expression)
+		public virtual IEnumerable<TResult> ExecuteFuture<TResult>(Expression expression)
 		{
-			IQuery query;
-			NhLinqExpression nhQuery;
-			NhLinqExpression nhLinqExpression = PrepareQuery(expression, out query, out nhQuery);
-			return ExecuteFutureQuery(nhLinqExpression, query, nhQuery);
+			PrepareQuery(expression, out var query, out var nhQuery);
+
+			var result = query.Future<TResult>();
+			SetupFutureResult(nhQuery, (IDelayedValue)result);
+
+			return result;
+		}
+
+		public virtual IFutureValue<TResult> ExecuteFutureValue<TResult>(Expression expression)
+		{
+			PrepareQuery(expression, out var query, out var nhQuery);
+
+			var result = query.FutureValue<TResult>();
+			SetupFutureResult(nhQuery, (IDelayedValue)result);
+
+			return result;
+		}
+
+		private static void SetupFutureResult(NhLinqExpression nhQuery, IDelayedValue result)
+		{
+			if (nhQuery.ExpressionToHqlTranslationResults.PostExecuteTransformer == null)
+				return;
+
+			result.ExecuteOnEval = nhQuery.ExpressionToHqlTranslationResults.PostExecuteTransformer;
 		}
 
 		protected virtual NhLinqExpression PrepareQuery(Expression expression, out IQuery query, out NhLinqExpression nhQuery)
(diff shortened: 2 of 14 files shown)


fredericDelaporte:
I have some additional simplifications to do, removing a bunch of useless reflection and casting. But they are currently blocked by a bug I am going to write a test case for and a fix.

src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -886,7 +886,7 @@ public IEnumerable<T> Future<T>()
 		{
 			if (!session.Factory.ConnectionProvider.Driver.SupportsMultipleQueries)
 			{
-				return List<T>();
+				return new DelayedEnumerator<T>(List<T>);
```
fredericDelaporte: Latent bug fix in `Future` method. In case of lack of multi-query supports, a raw list was yielded instead of the future list internal implementation.
But Linq Future handling casts results to the internal interface `IDelayedValue` for defining their post execute transformer. Previous code was doing that only if such a transformer was defined, and they are currently defined only on aggregating results using the `FutureValue` method, which does not have that flaw.

This also adjusts a "detail": previous implementation was causing an immediate execution of the query in case of lack of multiple query support, contrary to what do `FutureValue`. Now it will be delayed too, so a discarded future query will not have its query run at all when multiple queries are not supported, it will be run only if the future query is actually enumerated.

src/NHibernate/Linq/DefaultQueryProvider.cs
```diff
@@ -59,12 +60,32 @@ public virtual IQueryable<T> CreateQuery<T>(Expression expression)
...
+		public virtual IEnumerable<TResult> ExecuteFuture<TResult>(Expression expression)
 		{
-			IQuery query;
-			NhLinqExpression nhQuery;
-			NhLinqExpression nhLinqExpression = PrepareQuery(expression, out query, out nhQuery);
-			return ExecuteFutureQuery(nhLinqExpression, query, nhQuery);
+			PrepareQuery(expression, out var query, out var nhQuery);
+
+			var result = query.Future<TResult>();
+			SetupFutureResult(nhQuery, (IDelayedValue)result);
```
fredericDelaporte: Previously the cast was done only if a post execute transformer was defined, hiding the potential failure of the lack of multi-query fallback. There were not actual bug currently because there is up to now no way to get a post result transformer with something else than a future value, and `FutureValue` own fallback was already yielding a `IDelayedValue`.

src/NHibernate/Linq/DefaultQueryProvider.cs
```diff
@@ -80,24 +101,6 @@ protected virtual NhLinqExpression PrepareQuery(Expression expression, out IQuer
 			return nhLinqExpression;
 		}
 
-		private static readonly MethodInfo Future = ReflectHelper.GetMethodDefinition<IQuery>(q => q.Future<object>());
-		private static readonly MethodInfo FutureValue = ReflectHelper.GetMethodDefinition<IQuery>(q => q.FutureValue<object>());
```
fredericDelaporte: No need anymore to reflect on those.

src/NHibernate/Linq/LinqExtensionMethods.cs
```diff
@@ -77,51 +81,81 @@ public static IQueryable<T> Timeout<T>(this IQueryable<T> query, int timeout)
...
+			{
+				throw new NotSupportedException($"Source {nameof(source.Provider)} must be a {nameof(INhQueryProvider)}");
+			}
+			var future = provider.ExecuteFuture<TSource>(source.Expression);
+			return new FutureValue<TSource>(() => future);
+		}
 
-			var provider = (INhQueryProvider) nhQueryable.Provider;
-			var future = provider.ExecuteFuture(nhQueryable.Expression);
-			if (future is IEnumerable<T>)
```
fredericDelaporte: This `if` was non-sens: result is always `IEnumerable<T>` since the expression is extracted from an `IQueryable<T>` without applying any aggregation on it. The implicit `else` part was indeed dead code.

src/NHibernate/Linq/LinqExtensionMethods.cs
```diff
@@ -77,51 +81,81 @@ public static IQueryable<T> Timeout<T>(this IQueryable<T> query, int timeout)
...
+			{
+				throw new ArgumentNullException(nameof(source));
+			}
+			if (!(source.Provider is INhQueryProvider provider))
 			{
-				return new FutureValue<T>(() => ((IEnumerable<T>) future));
+				throw new NotSupportedException($"Source {nameof(source.Provider)} must be a {nameof(INhQueryProvider)}");
 			}
 
-			return (IFutureValue<T>) future;
```
fredericDelaporte: This line was the implicit `else`, belonging to `ToFutureValue` without selector. The `IFutureValue<T>` case can only happen with an aggregation selector indeed.

src/NHibernate/Linq/DefaultQueryProvider.cs
```diff
@@ -80,24 +101,6 @@ protected virtual NhLinqExpression PrepareQuery(Expression expression, out IQuer
...
-
-		protected virtual object ExecuteFutureQuery(NhLinqExpression nhLinqExpression, IQuery query, NhLinqExpression nhQuery)
-		{
-			var method = (nhLinqExpression.ReturnType == NhLinqExpressionReturnType.Sequence ? Future : FutureValue)
```
fredericDelaporte: This condition was guaranteeing the `ToFutureValue` without selector case could only ever have a `IEnumerable<T>` result, causing the later commented `if` below to be moot. `ReturnType` is always sequence when the expression evaluate to `IQueryable`, see [here](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Linq/NhLinqExpression.cs#L56). Expression as only been [`PreTransform`](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Linq/NhRelinqQueryParser.cs#L56) at that point, and that does not change its type.

fredericDelaporte:
The aim of those simplifications is also to simplify async implementation of future, which are not looking nice IMO currently, with mostly 4x copy-paste + small change, more casts and reflection.

fredericDelaporte:
Haviing added a test case to a fixture which is globally failing for ODBC, I have fixed it here.

PR #645, opened 2017-06-13
NH-4003 - Refactor session constructor, some missing cleanup.

fredericDelaporte:
[https://nhibernate.jira.com/browse/NH-4003](NH-4003) - Refactor session constructor

Some cleanup was missing, due to that `ThreadLocalSessionContext` thing. This appears to be a partial port from Hibernate, left semi-functional. It is documented as closing automatically session at end of transaction, but this `autoClose` feature is not implemented. For hiding that from "innocent user", the `OpenSession` overload allowing to specify this useless `autoClose` is on `ISessionFactoryImplementor` instead of `ISessionFactory`, and I have missed marking it obsolete in the interface, while obsoleting the implementation.

I have also added some xml comments about things actually implemented.

Maybe should we instead drop them completely. At least drop the `AutoClose`, which is now visible by any user on the `ISessionBuilder`. Having publicly available non internal members with a "Not yet implemented" comment is not very nice.

The `ThreadLocalSessionContext` could not be that easy to drop: it has some special semantics implemented which some users may rely on: 

- Auto-open a session if none was there, instead of yielding null.
- Auto-close previous session if any when binding a new one.
- Unless #644 is merged, supports multiple factories, contrary to `ThreadStaticSessionContext`.

### DIFF
--- a/src/NHibernate/Impl/SessionFactoryImpl.cs
+++ b/src/NHibernate/Impl/SessionFactoryImpl.cs
@@ -505,7 +505,6 @@ namespace NHibernate.Impl
 		{
 			return WithOptions()
 				.Connection(connection)
-				.Interceptor(interceptor)
 				.AutoClose(autoCloseSessionEnabled)
 				.ConnectionReleaseMode(connectionReleaseMode)
 				.OpenSession();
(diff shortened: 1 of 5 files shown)


src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -505,7 +505,6 @@ public ISession OpenSession(IInterceptor sessionLocalInterceptor)
 		{
 			return WithOptions()
 				.Connection(connection)
-				.Interceptor(interceptor)
```
fredericDelaporte: Is a no-op, `interceptor` is the default.

PR #646, opened 2017-06-15
NH-4033 - upgrading MySql connector.

fredericDelaporte:
[NH-4033](https://nhibernate.jira.com/browse/NH-4033) - Upgrade MySql connector used in tests

I have not yet installed a local MySql on my laptop, so I use this PR for the tests.


hazzik:
Do we actually need to create JIRAs for this?

fredericDelaporte:
Well, for the whole test part, no Jira? So we should remove the "Test" component from Jira? Why not, tests are not supposed to change anything to actual features.

PR #648, opened 2017-06-20
NH-4034 - flushing sessions sharing transaction on commit.

fredericDelaporte:
[NH-4034](https://nhibernate.jira.com/browse/NH-4034) - Flush all sessions participating in a transaction

With [NH-4003](https://nhibernate.jira.com/browse/NH-4003), a transaction sharing mechanism between sessions has been introduced. It works by opening new sessions from another one, while using the `Connection()` option on the session builder.

When committing the NHibernate transaction with `FlushMode.Commit`, only the originating session get flushed. The other sessions should be flushed too.

### DIFF
--- a/src/NHibernate.Test/SystemTransactions/TransactionFixture.cs
+++ b/src/NHibernate.Test/SystemTransactions/TransactionFixture.cs
@@ -1,26 +1,23 @@
-using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
 using System.Transactions;
-using NHibernate.DomainModel;
+using NHibernate.Linq;
+using NHibernate.Test.TransactionTest;
 using NUnit.Framework;
 
 namespace NHibernate.Test.SystemTransactions
 {
 	[TestFixture]
-	public class TransactionFixture : TestCase
+	public class TransactionFixture : TransactionFixtureBase
 	{
-		protected override IList Mappings
-		{
-			get { return new string[] { "WZ.hbm.xml" }; }
-		}
-
 		[Test]
 		public void CanUseSystemTransactionsToCommit()
 		{
-			object identifier;
+			int identifier;
 			using(ISession session = Sfi.OpenSession())
 			using(TransactionScope tx = new TransactionScope())
 			{
-				W s = new W();
+				var s = new Person();
 				session.Save(s);
 				identifier = s.Id;
 				tx.Complete();
@@ -29,10 +26,83 @@ namespace NHibernate.Test.SystemTransactions
 			using (ISession session = Sfi.OpenSession())
 			using (TransactionScope tx = new TransactionScope())
 			{
-				W w = session.Get<W>(identifier);
+				var w = session.Get<Person>(identifier);
 				Assert.IsNotNull(w);
 				session.Delete(w);
 				tx.Complete();
+			}
+		}
+
+		[Test]
+		public void FlushFromTransactionAppliesToDisposedSharingSession()
+		{
+			var flushOrder = new List<int>();
+			using (var s = OpenSession(new TestInterceptor(0, flushOrder)))
+			{
+				var builder = s.SessionWithOptions().Connection();
+
+				using (var t = new TransactionScope())
+				{
+					var p1 = new Person();
+					var p2 = new Person();
+					var p3 = new Person();
+					var p4 = new Person();
+
+					using (var s1 = builder.Interceptor(new TestInterceptor(1, flushOrder)).OpenSession())
+						s1.Save(p1);
+					using (var s2 = builder.Interceptor(new TestInterceptor(2, flushOrder)).OpenSession())
+					{
+						s2.Save(p2);
+						using (var s3 = s2.SessionWithOptions().Connection().Interceptor(new TestInterceptor(3, flushOrder)).OpenSession())
+							s3.Save(p3);
+					}
+					s.Save(p4);
+					t.Complete();
+				}
+			}
+
+			Assert.That(flushOrder, Is.EqualTo(new[] { 1, 2, 3, 0 }));
+
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				Assert.That(s.Query<Person>().Count(), Is.EqualTo(4));
+				t.Commit();
+			}
+		}
+
+		[Test]
+		public void FlushFromTransactionAppliesToSharingSession()
+		{
+			var flushOrder = new List<int>();
+			using (var s = OpenSession(new TestInterceptor(0, flushOrder)))
+			{
+				var builder = s.SessionWithOptions().Connection();
+
+				using (var s1 = builder.Interceptor(new TestInterceptor(1, flushOrder)).OpenSession())
+				using (var s2 = builder.Interceptor(new TestInterceptor(2, flushOrder)).OpenSession())
+				using (var s3 = s2.SessionWithOptions().Connection().Interceptor(new TestInterceptor(3, flushOrder)).OpenSession())
+				using (var t = new TransactionScope())
+				{
+					var p1 = new Person();
+					var p2 = new Person();
+					var p3 = new Person();
+					var p4 = new Person();
+					s1.Save(p1);
+					s2.Save(p2);
+					s3.Save(p3);
+					s.Save(p4);
+					t.Complete();
+				}
+			}
+
+			Assert.That(flushOrder, Is.EqualTo(new[] { 1, 2, 3, 0 }));
+
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				Assert.That(s.Query<Person>().Count(), Is.EqualTo(4));
+				t.Commit();
 			}
 		}
 	}
--- a/src/NHibernate.Test/SystemTransactions/TransactionFixture.cs
+++ b/src/NHibernate.Test/SystemTransactions/TransactionFixture.cs
@@ -61,7 +61,7 @@ namespace NHibernate.Test.SystemTransactions
 				}
 			}
 
-			Assert.That(flushOrder, Is.EqualTo(new[] { 1, 2, 3, 0 }));
+			Assert.That(flushOrder, Is.EqualTo(new[] { 0, 1, 2, 3 }));
 
 			using (var s = OpenSession())
 			using (var t = s.BeginTransaction())
@@ -96,7 +96,7 @@ namespace NHibernate.Test.SystemTransactions
 				}
 			}
 
-			Assert.That(flushOrder, Is.EqualTo(new[] { 1, 2, 3, 0 }));
+			Assert.That(flushOrder, Is.EqualTo(new[] { 0, 1, 2, 3 }));
 
 			using (var s = OpenSession())
 			using (var t = s.BeginTransaction())
--- a/src/NHibernate.Test/TransactionTest/Person.cs
+++ b/src/NHibernate.Test/TransactionTest/Person.cs
@@ -1,0 +1,11 @@
+using System;
+
+namespace NHibernate.Test.TransactionTest
+{
+	public class Person
+	{
+		public virtual int Id { get; set; }
+
+		public virtual DateTime CreatedAt { get; set; } = DateTime.Now;
+	}
+}
--- a/src/NHibernate.Test/TransactionTest/Person.hbm.xml
+++ b/src/NHibernate.Test/TransactionTest/Person.hbm.xml
@@ -1,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<hibernate-mapping xmlns="urn:nhibernate-mapping-2.2"
+    namespace="NHibernate.Test.TransactionTest"
+    assembly="NHibernate.Test">
+
+  <class name="Person">
+    <id name="Id">
+      <generator class="hilo"/>
+    </id>
+    <property name="CreatedAt"/>
+  </class>
+</hibernate-mapping>
--- a/src/NHibernate.Test/TransactionTest/TransactionFixture.cs
+++ b/src/NHibernate.Test/TransactionTest/TransactionFixture.cs
@@ -163,7 +163,7 @@ namespace NHibernate.Test.TransactionTest
 				}
 			}
 
-			Assert.That(flushOrder, Is.EqualTo(new[] { 1, 2, 3, 0 }));
+			Assert.That(flushOrder, Is.EqualTo(new[] { 0, 1, 2, 3 }));
 
 			using (var s = OpenSession())
 			using (var t = s.BeginTransaction())
--- a/src/NHibernate.Test/TransactionTest/TransactionFixture.cs
+++ b/src/NHibernate.Test/TransactionTest/TransactionFixture.cs
@@ -1,19 +1,15 @@
 using System;
 using System.Collections;
-using System.Data.Common;
+using System.Collections.Generic;
+using System.Linq;
+using NHibernate.Linq;
 using NUnit.Framework;
 
 namespace NHibernate.Test.TransactionTest
 {
 	[TestFixture]
-	public class TransactionFixture : TestCase
+	public class TransactionFixture : TransactionFixtureBase
 	{
-		protected override IList Mappings
-		{
-			// The mapping is only actually needed in one test
-			get { return new string[] {"Simple.hbm.xml"}; }
-		}
-
 		[Test]
 		public void SecondTransactionShouldntBeCommitted()
 		{
@@ -74,25 +70,25 @@ namespace NHibernate.Test.TransactionTest
 		{
 			using (ISession s = OpenSession())
 			{
-				using (ITransaction t = s.BeginTransaction())
+				using (s.BeginTransaction())
 				{
 				}
 
-				s.CreateQuery("from Simple").List();
+				s.CreateQuery("from Person").List();
 
 				using (ITransaction t = s.BeginTransaction())
 				{
 					t.Commit();
 				}
 
-				s.CreateQuery("from Simple").List();
+				s.CreateQuery("from Person").List();
 
 				using (ITransaction t = s.BeginTransaction())
 				{
 					t.Rollback();
 				}
 
-				s.CreateQuery("from Simple").List();
+				s.CreateQuery("from Person").List();
 			}
 		}
 
@@ -139,6 +135,41 @@ namespace NHibernate.Test.TransactionTest
 					Assert.IsFalse(t.IsActive);
 					Assert.IsFalse(s.Transaction.IsActive);
 				}
+			}
+		}
+
+		[Test]
+		public void FlushFromTransactionAppliesToSharingSession()
+		{
+			var flushOrder = new List<int>();
+			using (var s = OpenSession(new TestInterceptor(0, flushOrder)))
+			{
+				var builder = s.SessionWithOptions().Connection();
+
+				using (var s1 = builder.Interceptor(new TestInterceptor(1, flushOrder)).OpenSession())
+				using (var s2 = builder.Interceptor(new TestInterceptor(2, flushOrder)).OpenSession())
+				using (var s3 = s1.SessionWithOptions().Connection().Interceptor(new TestInterceptor(3, flushOrder)).OpenSession())
+				using (var t = s.BeginTransaction())
+				{
+					var p1 = new Person();
+					var p2 = new Person();
+					var p3 = new Person();
+					var p4 = new Person();
+					s1.Save(p1);
+					s2.Save(p2);
+					s3.Save(p3);
+					s.Save(p4);
+					t.Commit();
+				}
+			}
+
+			Assert.That(flushOrder, Is.EqualTo(new[] { 1, 2, 3, 0 }));
+
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				Assert.That(s.Query<Person>().Count(), Is.EqualTo(4));
+				t.Commit();
 			}
 		}
 	}
--- a/src/NHibernate.Test/TransactionTest/TransactionFixtureBase.cs
+++ b/src/NHibernate.Test/TransactionTest/TransactionFixtureBase.cs
@@ -1,0 +1,39 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace NHibernate.Test.TransactionTest
+{
+	public abstract class TransactionFixtureBase : TestCase
+	{
+		protected override IList Mappings => new[] { "TransactionTest.Person.hbm.xml" };
+
+		protected override string MappingsAssembly => "NHibernate.Test";
+
+		protected override void OnTearDown()
+		{
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				s.Delete("from System.Object");
+				t.Commit();
+			}
+		}
+
+		public class TestInterceptor : EmptyInterceptor
+		{
+			private readonly int _numero;
+			private readonly List<int> _flushOrder;
+
+			public TestInterceptor(int numero, List<int> flushOrder)
+			{
+				_numero = numero;
+				_flushOrder = flushOrder;
+			}
+
+			public override void PreFlush(ICollection entitites)
+			{
+				_flushOrder.Add(_numero);
+			}
+		}
+	}
+}
--- a/src/NHibernate/AdoNet/ConnectionManager.cs
+++ b/src/NHibernate/AdoNet/ConnectionManager.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
 using System.Runtime.Serialization;
@@ -41,6 +42,17 @@ namespace NHibernate.AdoNet
 		private readonly ISessionImplementor session;
 		private readonly ConnectionReleaseMode connectionReleaseMode;
 		private readonly IInterceptor interceptor;
+		private readonly List<ISessionImplementor> _dependentSessions = new List<ISessionImplementor>();
+
+		/// <summary>
+		/// The session responsible for the lifecycle of the connection manager.
+		/// </summary>
+		public ISessionImplementor Session => session;
+
+		/// <summary>
+		/// The sessions using the connection manager of the session responsible for it.
+		/// </summary>
+		public IReadOnlyCollection<ISessionImplementor> DependentSessions => _dependentSessions;
 
 		[NonSerialized]
 		private bool _releasesEnabled = true;
@@ -61,6 +73,16 @@ namespace NHibernate.AdoNet
 			batcher = session.Factory.Settings.BatcherFactory.CreateBatcher(this, interceptor);
 
 			ownConnection = suppliedConnection == null;
+		}
+
+		public void AddDependentSession(ISessionImplementor session)
+		{
+			_dependentSessions.Add(session);
+		}
+
+		public void RemoveDependentSession(ISessionImplementor session)
+		{
+			_dependentSessions.Remove(session);
 		}
 
 		public bool IsInActiveTransaction
--- a/src/NHibernate/ISharedSessionBuilder.cs
+++ b/src/NHibernate/ISharedSessionBuilder.cs
@@ -9,6 +9,8 @@ namespace NHibernate
 	{
 		/// <summary>
 		/// Signifies that the connection from the original session should be used to create the new session.
+		/// The original session remains responsible for it and its closing will cause sharing sessions to be no
+		/// more usable.
 		/// </summary>
 		/// <returns><see langword="this" />, for method chaining.</returns>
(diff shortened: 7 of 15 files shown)


src/NHibernate.Test/TransactionTest/TransactionFixture.cs
```diff
@@ -1,19 +1,13 @@
...
 namespace NHibernate.Test.TransactionTest
 {
 	[TestFixture]
-	public class TransactionFixture : TestCase
+	public class TransactionFixture : TransactionFixtureBase
 	{
-		protected override IList Mappings
-		{
-			// The mapping is only actually needed in one test
-			get { return new string[] {"Simple.hbm.xml"}; }
```
fredericDelaporte: This `Simple` entity is anything but simple. It does not have a mapped id property, just a column in database, causing weird patterns when actually willing to use it.

src/NHibernate.Test/TransactionTest/TransactionFixtureBase.cs
```diff
@@ -0,0 +1,19 @@
...
+		protected override IList Mappings => new[] { "TransactionTest.Person.hbm.xml" };
+
+		protected override string MappingsAssembly => "NHibernate.Test";
+
+		protected override void OnTearDown()
+		{
+			using (var s = OpenSession())
+			{
+				s.CreateQuery("delete from System.Object").ExecuteUpdate();
+			}
```
fredericDelaporte: I wonder whether such `Teardown` should be directly added in `TestCase`. It would catch all cases not requiring ordering for deletes, and could still be overridden by test needing a specific delete ordering.

src/NHibernate/AdoNet/ConnectionManager.cs
```diff
@@ -63,6 +68,16 @@ public interface Callback
...
 		}
 
+		public void AddSessionSharingManager(ISessionImplementor session)
+		{
+			_sessionsSharingManager.Add(session);
+		}
+
+		public void RemoveSessionSharingManager(ISessionImplementor session)
+		{
+			_sessionsSharingManager.Remove(session);
```
fredericDelaporte: `O(n)`, true, but well, we are not supposed to open a lot of sessions from a single one. Otherwise, we would have to use the `OrderedHashSet` from Iesi, since we need to retain the ordering.

src/NHibernate/ISharedSessionBuilder.cs
```diff
@@ -9,6 +9,8 @@ public interface ISharedSessionBuilder : ISessionBuilder<ISharedSessionBuilder>
 	{
 		/// <summary>
 		/// Signifies that the connection from the original session should be used to create the new session.
+		/// The original session remains responsible for it and its closing will cause sharing sessions to be no
+		/// more usable.
```
fredericDelaporte: Documenting that pitfall by the way.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -316,6 +317,12 @@ public DbConnection Close()
 		/// </summary>
 		public override void AfterTransactionCompletion(bool success, ITransaction tx)
 		{
+			if (!_transactionCoordinatorShared)
+				foreach (var sharingSession in ConnectionManager.SessionsSharingManager)
+				{
+					sharingSession.AfterTransactionCompletion(success, tx);
```
fredericDelaporte: I could have put that in callers, but then it was duplicated between `AdoTransaction` and `AdoNetWithDistributedTransactionFactory`.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -324,7 +331,12 @@ public override void AfterTransactionCompletion(bool success, ITransaction tx)
 					Factory.StatisticsImplementor.EndTransaction(success);
 				}
 
-				connectionManager.AfterTransaction();
+				// Let the originating session notify the connection manager
+				if (!_transactionCoordinatorShared)
+				{
+					connectionManager.AfterTransaction();
+				}
```
fredericDelaporte: In the transaction scope case, the connection manager was getting notified by each session using it. Better let only the originating session do that. This further justifies the ordering: better notify sharing session then originating session last.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -2092,12 +2104,28 @@ public override void AfterTransactionBegin(ITransaction tx)
 			using (new SessionIdLoggingContext(SessionId))
 			{
 				CheckAndUpdateSessionStatus();
-				Interceptor.AfterTransactionBegin(tx);
+
+				if (!_transactionCoordinatorShared)
+				{
+					Interceptor.AfterTransactionBegin(tx);
+				}
```
fredericDelaporte: Bit muddy subject here.

In the case of `AdoTransaction`, the transaction events were not called on sharing sessions, so this check keeps that behavior for the interceptor.

In the case of transaction scope, each session was still enlisting itself, and was receiving events. So the interceptor was call for each sharing session and for the originating one, excepted for the after transaction event which was already guarded for being fired only from originating session... I have decided to align on that behavior for all events.

Now, why not firing the call on the sharing session interceptor? Maybe because by default, it is the same than the originating session. So it would get notified multiples times, likely causing issues. Maybe we should check if it is the same, and if not, then firing the events. Going to add that change.

src/NHibernate/Transaction/AdoNetWithDistributedTransactionFactory.cs
```diff
@@ -29,7 +30,18 @@ public void EnlistInDistributedTransactionIfNeeded(ISessionImplementor session)
...
+			var originatingSession = session.ConnectionManager.Session;
+			if (originatingSession != session)
+			{
+				session.TransactionContext = new DependentContext();
+			}
+
+			if (originatingSession.TransactionContext != null)
+				return;
+
+			session = originatingSession;
```
fredericDelaporte: Previous logic was causing each sharing session to enlist its own resource within the transaction. Then each was having its own prepare/transaction complete/... In non distributed case, that was working quite well, because it appears all those volatile resources are called in turn according to their enlistment order.

But with distributed transaction, they would have been called asynchronously, and then what a mess...

So I have changed the logic for having only the originating session enlisting itself.

src/NHibernate.Test/TransactionTest/Person.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace NHibernate.Test.TransactionTest
+{
+	public class Person
+	{
+        public virtual int Id { get; set; }
```
hazzik: White space

src/NHibernate/AdoNet/ConnectionManager.cs
hazzik: LinkedList would be enough.

src/NHibernate/AdoNet/ConnectionManager.cs
```diff
@@ -41,6 +42,10 @@ public interface Callback
 		private readonly ISessionImplementor session;
 		private readonly ConnectionReleaseMode connectionReleaseMode;
 		private readonly IInterceptor interceptor;
+		private readonly List<ISessionImplementor> _sessionsSharingManager = new List<ISessionImplementor>();
```
hazzik: Can this be renamed to something more obvious? I don't understand who is managing whom from the name.

fredericDelaporte: Granted... I have some difficulties with that name.

src/NHibernate/AdoNet/ConnectionManager.cs
fredericDelaporte: `LinkedList` is O(n) as `List` since it still does involve a non indexed search. It just saves on the list rewriting.

src/NHibernate/Transaction/AdoTransaction.cs
```diff
@@ -186,6 +186,13 @@ public void Commit()
 
 				log.Debug("Start Commit");
 
+				foreach (var sharingSession in session.ConnectionManager.SessionsSharingManager)
```
hazzik: Maybe move list of connections directly to the transaction?

hazzik: Sorry, list of sessions, not connections.

fredericDelaporte: It serves for transaction scope too. It would be strange to involve the `AdoTransaction` for transaction scope handling. And at each transaction renewal, we would have to pass that list from the old to the new.

I am renaming that list list `DependentSessions`.

src/NHibernate.Test/TransactionTest/Person.hbm.xml
```diff
@@ -0,0 +1,14 @@
...
+				   namespace="NHibernate.Test.TransactionTest"
+				   assembly="NHibernate.Test">
+
+	<class name="Person">
+		<id name="Id">
+			<generator class="hilo"/>
+		</id>
+    <property name="CreatedAt"/>
+    <property name="NotNullData"/> <!-- NOT NULL will be set on created schema directly to avoid NH checks. -->
+    <many-to-one name="Related" class="Person" />
```
fredericDelaporte: white spaces still^^

src/NHibernate/AdoNet/ConnectionManager.cs
```diff
@@ -63,6 +75,16 @@ public interface Callback
 			ownConnection = suppliedConnection == null;
 		}
 
+		public void AddSessionSharingManager(ISessionImplementor session)
```
hazzik: The method name shall be renamed as well ;)

hazzik:
Can you test a case with a grandchild session (a child session opened from a child session)?

src/NHibernate/AdoNet/ConnectionManager.cs
fredericDelaporte: Of course!

hazzik:
I did not run the code myself, so I would like to know: in the case of grand-child session who is the main session is it immediate parent or a grand-parent?

src/NHibernate/Transaction/AdoTransaction.cs
```diff
@@ -186,6 +186,13 @@ public void Commit()
 
 				log.Debug("Start Commit");
 
+				foreach (var dependentSession in session.ConnectionManager.DependentSessions)
```
hazzik: Shouldn't this be reversed?

fredericDelaporte:
The main session is always the one responsible for the connection manager. The one which has created it itself instead of getting it from its constructor options.

There is actually no such things as "grand children", their are just all dependent sessions if they have taken their connection manager from "outside". Otherwise their are just sessions, without any dependency or relationship with the one having spawn them.

 A "grand children" parent could even be closed without this affecting the grand children: only the session responsible for the connection manager must ot be closed before dependent sessions.

src/NHibernate/Transaction/AdoTransaction.cs
fredericDelaporte: Reverted? No, that is precisely the main point of this PR.
Reversed order? I do not get why.

hazzik:
I think the flush order shall be reverse of open order (LIFO), this would make much more sense. This way you can, probably, use only one stack and eliminate the "Session" property on ConnectionManager

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -2092,17 +2104,33 @@ public override void AfterTransactionBegin(ITransaction tx)
 			using (new SessionIdLoggingContext(SessionId))
 			{
 				CheckAndUpdateSessionStatus();
-				Interceptor.AfterTransactionBegin(tx);
+
+				if (!_transactionCoordinatorShared || Interceptor != ConnectionManager.Session.Interceptor)
```
hazzik: This condition is repeated several times, I think it shall be moved to a method.

hazzik: Why you do not call interceptors for "child" sessions? I think it is perfectly valid scenario.

fredericDelaporte:
Flushing in reverse order is not very intuitive for the user I think. It is likely users will expect the dependent sessions to be flush in their opening order.

I am a bit reluctant to put the session owning the connection manager in the same list of others by the way. It really has a special role. I am not willing to rely on the list ordering for acting on that particular session at the right moment.

I do not see this list as a stack, as if there were a sort of nested transaction concept. Those dependent sessions are more a heap of things acting independently, eventually overlapping their usage. That is way I handle them FIFO.

src/NHibernate/Impl/SessionImpl.cs
fredericDelaporte: I do call it for "children", excepted if that is the same interceptor than the originating session. It does not make sens to call it for each children session in such case, since the originating session will anyway call it, and nothing allow the interceptor to understand why it has many "begin tran" called: it does not know which session has called that.

Previous code was never calling for "children" session, either because the session event was not call at all, or because a test on `_transactionCoordinatorShared` was already there). I have adjusted that for always calling dependent session transaction events and also calling the interceptor only if the dependent session has another interceptor than the originating session.

"Child session" not sharing the connection will have independent transaction handling and will call their interceptor in all cases. Their are just usual sessions indeed, without any knowledge of having been spawned from another session.

hazzik:
I don't see the original session as an "the special one". It was just the first one. So, I believe, it shall be a stack. The "speciality" of the session is that it causes the "dependent" session to do the same, but I think it is wrong. This causality shall happen in a connection manager itself (or, some other class). 

So, for example, calls of `session.BeforeTransactionCompletion` shall be updated to calls to `connectionManager.BeforeTransactionCompletion`, which would be implemented as:

```
foreach(var session in this.sessions) { session.BeforeTransactionCompletion(...); }
```

src/NHibernate/Impl/SessionImpl.cs
hazzik: Ok, I gave it another though. I think that this code makes `ISharedSessionBuilder.Interceptor()` to have no sense at all, as shared interceptor would never be called.

fredericDelaporte: Interceptor is not here only for transaction. All the other cases are still called. But when the transaction is shared, does it make sens to notifies the same interceptor multiple time of the same transaction beginning? If the session was supplied as an argument, why not, but that is not the case.

hazzik:
Can we port these tests? https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/test/java/org/hibernate/sharedSession/SessionWithSharedConnectionTest.java

hazzik:
Some reading: https://hibernate.atlassian.net/browse/HHH-7273

hazzik:
https://hibernate.atlassian.net/browse/HHH-7017

fredericDelaporte:
Yes I thought about that job would be better centralized somewhere else, but only the connection manager seems currently available for taking this. That would add to it a bunch of new methods, while changing the semantics of other: it already has its own AfterTransaction, it would to be merged with the after transaction notifying the session.

But I still disagree for the rest.

> I don't see the original session as an "the special one"

This one still own the connection manager. Only this one will close it, regardless of whether spawned session are still around, un-closed, or not. Maybe do you wish to change that. I have just stuck to Hibernate implementation here. That would mean something like, any call to RemoveDependentSession causing the list to get empty would then dispose the connection manager. No more needs to track the sharing, provided all the notification handling has been moved into connection manager.

> I believe, it shall be a stack.

I really am unable to understand what is the benefit of working with dependent sessions as a stack: it really looks counter-intuitive for me, and from a user point of view. Those sessions do not care of each others, excepted they will start failing if the connection manager owning one is closed. Why should we ensure they work as if they were a stack? Moreover, in term of lifecycle, the user is not at all mandated to work with them in LIFO, closing first the last opened. It is even likely that in many cases, they will use them in FIFO mode, spawning one dependent session, working with it, disposing of it, then spawning another one, ...

Flushing them in reverse order just look to me as the path for maximizing troubles. I think it is way more likely that the later sessions work ends up having some dependency on the previous sessions work, than the opposite.

(I have not yet read your links, it is almost 3 in the morning here.)

hazzik:
>(I have not yet read your links, it is almost 3 in the morning here.)

Get some sleep:) Maybe in the morning you'll agree with me. 

We, probably, shall not implement this feature at all.

hazzik:
We need to port this feature first: https://github.com/hibernate/hibernate-orm/commit/b476094d43e24db36d970e6c0c21607fe5d42832

fredericDelaporte:
A 192 files change without much explanations in the ticket. Here we are only on a 12 files change, for fixing a specific issue.

I still do not understand why it would be worst to do the changes I propose rather than to stay in current situation. Especially I am still completely unable to understand why should it be treated as a stack, even by reading the above links, which I am unable to relate to the subject. Why not porting the tests, but they almost all rely on a non ported feature: autoclose. So it may not make much sens. I will probably have a look at what impact it has to move the transaction event raising into the connection manager.

The links chain to a bunch of issues which I believe NHibernate does not suffer or do not care at its current state. Much of those issues are due to complex interactions between all the many components they currently use to get the session working: `JdbcCoordinator` (NHibernate `ConnectionManager`), `TransactionCoordinator` (which NHibernate do not actually have, its transaction factory and transaction context do not do the same job), `LogicalConnection` (merged in `ConnectionManager` for NHibernate).

One of their issues is to have the session holding the transaction coordinator itself holding the jdbc coordinator, and that creates them issues for allowing concurrent usage of sessions sharing the same transaction coordinator, because connection does not allow multi-threaded usage. So they want to invert that relationship, having the session holding the jdbc coordinator which in turn hold the transaction coordinator.  
Fine, NHibernate is somewhat already in that state, session holds directly the connection manager. (There is no transaction coordinator.) And since currently NHibernate explicitly shares the connection, not the transaction, it looks like a no-brainer that session sharing the same connection should not be used concurrently (by different threads), and that NHibernate does not support it.

They have others issues, like connection leaks when sharing (with current NHibernate model, I do not see how that could be happening, unless not closing the "originating" session, and if I move all to connection manager, unless not closing all sessions, or using them on different threads), like "originating" session getting closed by spawned sessions own close (due to the shared transaction coordinator which, when closed, closes its "main session", behavior we do not have currently in NHibernate), ...

It looks to me that porting all their stuff is likely to creates way more issues than what we actually have now, all that for maybe supporting a scenario I have seen asked by no-one, allowing concurrent (multi-threaded) usage of sessions sharing the same transaction. Anyway `DbTransaction` is not thread safe... We will not be able to implement a transaction coordinator supporting multi-threaded usage unless mandating use of system transactions instead of ado transaction.

Nowhere I have understood what it was having to do with considering spawned sessions as a stack of sessions.

fredericDelaporte:
I had forgotten, they also have issues with "temporary sessions". NHibernate does not have these currently. Maybe those one need to be treated in a stacked way, but they appear to me as really another beast that this feature of spawning sessions sharing the same connection.

The envers case maybe desserve a stacked treatment, but are those spawned sessions the right tool for it? (And well, stacked treatment for envers sound strange for me: flushing logs before the actual changes they correspond too?)

hazzik:
Ok, let's forget about the stacked sessions, but why the "parent" session needs to be flushed last? For me, it either does not matter or otherwise, all sessions shall be flushed in the opposite order of their opening.

fredericDelaporte:
Well, I am quite agree with you on that later point, if we change the implementation for having the originating session to really be just another session, no more exclusively responsible for the connection manager. So I definitely need to move those transaction events raising into the connection manager, then handle its disposal at last session removal instead of at originating session closing.

hazzik:
Ok, I've finally found how Hibernate does flush with shared transaction coordinator. There is indeed no "primary"/"parent" session. They handle multiple transactions flushes by "ISomethingObserver". (It looks the same as ISynchronization (but they also have ISynchronization). There is an anonymous observer class which calls beforeTransactionCompletion and afterTransactionCompletion. Flush is called from flushBeforeTransactionCompletion, which is called from beforeTransactionCompletion.

hazzik:
The observer is added from a session constructor.

hazzik:
Ok, I was wrong. There IS a "parent" session. But, it is called first.

fredericDelaporte:
Yes I have already seen those observers in their code, without checking all their responsibilities. Hibernate is definitely way more "meat" heavy than NHibernate currently is. Maybe I am slightly biased toward Java, but I sometime fear they fell in the [hammer factory factory](http://discuss.joelonsoftware.com/default.asp?joel.3.219431.12) trap.

So on Hibernate side, they still have the "originating" session treated as a special one. I am not sure this is very desirable.

hazzik:
They do not restrict calling interceptors from "child" sessions. This is how the observer is implemented:

	@Override
	protected void addSharedSessionTransactionObserver(TransactionCoordinator transactionCoordinator) {
		transactionCoordinator.addObserver(
				new TransactionObserver() {
					@Override
					public void afterBegin() {
					}

					@Override
					public void beforeCompletion() {
						if ( isOpen() && getHibernateFlushMode() !=  FlushMode.MANUAL ) {
    // Hazzik: this is strange, they do not call "beforeTransactionCompletion" for some reasons.
							managedFlush();
						}
						actionQueue.beforeTransactionCompletion();
						try {
							getInterceptor().beforeTransactionCompletion( getCurrentTransaction() );
						}
						catch (Throwable t) {
							log.exceptionInBeforeTransactionCompletionInterceptor( t );
						}
					}

					@Override
					public void afterCompletion(boolean successful, boolean delayed) {
						afterTransactionCompletion( successful, delayed );
						if ( !isClosed() && autoClose ) {
							managedClose();
						}
					}
				}
		);
	}

fredericDelaporte:
What is your comment in this code about? Do you think before completion should be called before flush on commit? NHibernate too, does flush before the "before completion" event for ado transaction. (But currently not for system transaction, which is an oddity to have it different, and I believe I have changed it in #627.)

hazzik:
Observer's `beforeCompletion` looks the same as SessionImpl's `beforeTransactionCompletion``. The only difference that in former one they do not check if transaction is in progress (not sure if this is an optimization, or not.)

hazzik:
@Override
	public void beforeTransactionCompletion() {
		log.tracef( "SessionImpl#beforeTransactionCompletion()" );
		flushBeforeTransactionCompletion();
		actionQueue.beforeTransactionCompletion();
		try {
			getInterceptor().beforeTransactionCompletion( getCurrentTransaction() );
		}
		catch (Throwable t) {
			log.exceptionInBeforeTransactionCompletionInterceptor( t );
		}
	}

	@Override
	public void flushBeforeTransactionCompletion() {
		final boolean doFlush = isTransactionFlushable()
				&& getHibernateFlushMode() != FlushMode.MANUAL;

		try {
			if ( doFlush ) {
				managedFlush();
			}
		}
		catch (RuntimeException re) {
			throw exceptionMapper.mapManagedFlushFailure( "error during managed flush", re, this );
		}
	}

	private boolean isTransactionFlushable() {
		if ( getCurrentTransaction() == null ) {
			// assume it is flushable - CMT, auto-commit, etc
			return true;
		}
		final TransactionStatus status = getCurrentTransaction().getStatus();
		return status == TransactionStatus.ACTIVE || status == TransactionStatus.COMMITTING;
	}

fredericDelaporte:
Ok, looks like either they have dead code, or more likely different code paths depending on whether the transaction is JTA, "native jdbc" (no more sure of the naming), or yet something else (JPA brings its own transaction handling too, or am I wrong?).

hazzik:
No, it's not dead code, it's called from different paths.

hazzik:
Our `AdoTransaction` is their `JdbcResourceLocalTransactionCoordinatorImpl` + `TransactionDriverControlImpl`

Our on commit (including this PR):

- flush "children" sessions
- flush "parent" session"
- notify synchronizations
- beginTransactionCompletion of "parent" session
  - beginTransactionCompletion of "children" sessions
  - actionQueue.beforeTranactionCompletion
  - interceptor.beforeTransactionCompletion

Their flow:

- "parent" session beforeTransactionCompletion
  - flush 
  - action queue
  - interceptor
-  notify synchronizations
- observer's beforeCompletion ("children" sessions")
  - flush
  - action queue
  - interceptor

hazzik:
Hibernate does not call `AfterTransactionBegin` for "children" sessions

hazzik:
@fredericDelaporte I'll try to make changes which resemble the current Hibernate state (I'll try to avoid hammer factory) and will make a PR to your branch shortly.

fredericDelaporte:
By the way, I do not see that as a priority. I was doing it for removing that lack of flush from transaction commit, believing it could be done with few code changes. It could stay lacking otherwise. After all there is no regression here, compared to 4.1.

I am not very keen on replicating Hibernate implementation when its rational are not clear to me. It could be that their choices make sens in a Java context, but less in a .Net one.

hazzik:
>rational are not clear to me

It seems that sole purpose of TransactionCoordinator is to manage the flushes on transaction commit

src/NHibernate.Test/SystemTransactions/TransactionFixture.cs
```diff
@@ -29,11 +25,84 @@ public void CanUseSystemTransactionsToCommit()
...
 		}
+
+		[Test]
+		public void FlushFromTransactionAppliesToDisposedSharingSession()
+		{
+			using (var s = OpenSession())
+			{
+				var builder = s.SessionWithOptions().Connection();
+
+				using (var t = new TransactionScope())
```
hazzik: I think these tests are not valid.

src/NHibernate.Test/SystemTransactions/TransactionFixture.cs
```diff
@@ -29,11 +25,84 @@ public void CanUseSystemTransactionsToCommit()
...
+		public void FlushFromTransactionAppliesToSharingSession()
+		{
+			using (var s = OpenSession())
+			{
+				var builder = s.SessionWithOptions().Connection();
+
+				using (var s1 = builder.OpenSession())
+				using (var s2 = builder.OpenSession())
+				using (var s3 = s2.SessionWithOptions().Connection().OpenSession())
+				using (var t = new TransactionScope())
```
hazzik: Same as above

src/NHibernate.Test/TransactionTest/TransactionFixture.cs
```diff
@@ -141,5 +135,40 @@ public void WasCommittedOrRolledBack()
...
+		public void FlushFromTransactionAppliesToSharingSession()
+		{
+			using (var s = OpenSession())
+			{
+				var builder = s.SessionWithOptions().Connection();
+
+				using (var s1 = builder.OpenSession())
+				using (var s2 = builder.OpenSession())
+				using (var s3 = s1.SessionWithOptions().Connection().OpenSession())
+				using (var t = s.BeginTransaction())
```
hazzik: Same as above

src/NHibernate.Test/SystemTransactions/TransactionFixture.cs
hazzik: http://nhibernate.info/doc/nhibernate-reference/transactions.html

> You should observe the following practices when creating NHibernate Sessions:
> 
> * **Never create more than one concurrent `ISession` or `ITransaction` instance per database connection.**
> 
> * Be extremely careful when creating more than one ISession per database per transaction. The ISession itself keeps track of updates made to loaded objects, so a different ISession might see stale data.
> 
> * The ISession is not threadsafe! Never access the same ISession in two concurrent threads. An ISession is usually only a single unit-of-work!

hazzik: At least it is not valid to have an entity graph with entities from different sessions. The valid approach testing that flush has happened would be to supply a listener or an interceptor.

fredericDelaporte: > Never create more than one concurrent ISession or ITransaction instance per database connection.

A bit contradictory with the connection sharing option indeed. By the way, previously was doable only by supplying a user connection I guess.

About having an entity graph from different sessions, yes that does not look as a good practice, I will change that.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -430,6 +431,7 @@ protected void AfterOperation(bool success)
 				if (!ConnectionManager.IsInActiveTransaction)
 				{
 					ConnectionManager.AfterNonTransactionalQuery(success);
+					ConnectionManager.AfterTransaction();
```
fredericDelaporte: Could be moved into `ConnectionManager.AfterNonTransactionalQuery`. This is not the Hibernate way but well, Hibernate does not have that `AfterNonTransactionalQuery`, which in NHibernate just do a log currently.

hazzik: I think it's better to leave it as is.

fredericDelaporte:
Thanks,

I guess you have seen some such typos/misses in my recent previous merges. Sorry to have not spot them, although I have read them.

PR #649, opened 2017-06-20
NH-4035 - cleanup even if tear down fails.

fredericDelaporte:
[NH-4035](https://nhibernate.jira.com/browse/NH-4035) - Teardown failure should not prevent cleanup

When a test teardown fails, the database cleanup, session closing, connection closing is no more done.
Ensure they are done whatever happens.
Otherwise this could cause chained failures of unrelated tests.

### DIFF
--- a/src/NHibernate.Test/TestCase.cs
+++ b/src/NHibernate.Test/TestCase.cs
@@ -151,20 +151,27 @@ namespace NHibernate.Test
 			string badCleanupMessage = null;
 			try
 			{
-				OnTearDown();
-				var wereClosed = _sessionFactory.CheckSessionsWereClosed();
-				var wasCleaned = CheckDatabaseWasCleaned();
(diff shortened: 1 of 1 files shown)


src/NHibernate.Test/TestCase.cs
```diff
@@ -151,20 +151,27 @@ public void TearDown()
...
-				var wasCleaned = CheckDatabaseWasCleaned();
-				var wereConnectionsClosed = CheckConnectionsWereClosed();
-				fail = !wereClosed || !wasCleaned || !wereConnectionsClosed;
-
-				if (fail)
+				try
+				{
+					OnTearDown();
+				}
+				finally
```
fredericDelaporte: A cleanup failure would hide the tear down failure. But cleanup has far less risk of failures unless the server is down (case where hiding the teardown failure does not matter much). And trying to guard against cleanup failure for not hiding tear down failure causes the code to be a bit convoluted.

PR #651, opened 2017-06-21
NH-4003 - avoiding session override on interceptor 

fredericDelaporte:
NH-4003 lack: avoiding session override on interceptor when a spawned session re-use its "parent" interceptor.

Consequences of a NH specific feature I have not sufficiently analyzed for the session builder from session case.

### DIFF
--- a/src/NHibernate/Impl/SessionFactoryImpl.cs
+++ b/src/NHibernate/Impl/SessionFactoryImpl.cs
@@ -1330,15 +1330,17 @@ namespace NHibernate.Impl
 			{
 				_log.Debug("Opening Hibernate Session.");
 				var session = new SessionImpl(_sessionFactory, this);
-				if (_interceptor != null)
-				{
-					// NH specific feature
-					// _interceptor may be the shared accros threads EmptyInterceptor.Instance, but that is
-					// not an issue, SetSession is no-op on it.
-					_interceptor.SetSession(session);
-				}
+				SetSessionOnInterceptor(session);
 
 				return session;
+			}
+
+			protected virtual void SetSessionOnInterceptor(SessionImpl session)
+			{
+				// NH specific feature
+				// _interceptor may be the shared accros threads EmptyInterceptor.Instance, but that is
+				// not an issue, SetSession is no-op on it.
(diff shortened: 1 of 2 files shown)


src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1330,17 +1330,19 @@ public virtual ISession OpenSession()
...
-					// _interceptor may be the shared accros threads EmptyInterceptor.Instance, but that is
-					// not an issue, SetSession is no-op on it.
-					_interceptor.SetSession(session);
-				}
+				SetSessionOnInterceptor(session);
 
 				return session;
 			}
 
+			protected virtual void SetSessionOnInterceptor(SessionImpl session)
```
hazzik: I think the correct approach is to drop this "feature". Currently the intention of `IInterceptor.SetSession` is not clear. I think it was an attempt to implement logging, which shall be done with a new "child" session instead. 

This "feature" was sneakily added without leaving any trace in the issue tracker.

fredericDelaporte: This feature had always looks as code smell to me. User needing it should probably be able to replace it by themselves, adding their own set session on their interceptors and calling it right after having opened their sessions. Slightly less convenient maybe, but relieving NHibernate for handling a thing having conflicts potentials (when an interceptor is used by many sessions, when using the session inside some events, potentially messing up the session internal state, ...).

Or maybe should we not care at all of the conflict potential and leave it to the user to avoid re-using an interceptor needing the session with many sessions.

The fact it has been added without a Jira ten years ago is a detail. And in facts, it has one: [NH-947](https://nhibernate.jira.com/browse/NH-947).

As explained by Ayende:

> The only usage is in the rare cases where you have interceptor per session, and it needs to be aware of the session that it is attached to.

So we may instead do not care of potentially overriding the session and let the user cope with that.

hazzik: >The fact it has been added without a Jira ten years ago is a detail. And in facts, it has one: NH-947.

Sure, I've found this JIRA later

PR #652, opened 2017-06-22
NH-2241 - IStatelessSession is accidentally 2LC enabled in some cases

igitur:
[NH-2241](https://nhibernate.jira.com/browse/NH-2241) - When inserting an entity that references an entity with assigned id causes exception if that entity as a cache property.

Example:
```xml
<class name="Document" >
    <cache region="ShortTerm" usage="read-write" />
    <id name="Name" >
        <generator class="assigned"/>
    </id>
    <property name="Text"/>
</class>
<class name="Paper">
    <id name="Id">
        <generator class="native"/>
    </id>
    <property name="Color"/>
    <many-to-one name="Document" column="DocId" cascade="none" />
</class>
```
When inserting a `Paper` entity using `BatchInsert` in the stateless session throws `NotSupported` exception from the `Timestamp` property on the session (`ISessionImplementor`).

This happens in the end of the `IsTransient` method (of the `AbstractEntityPersister`):

```c#
// check to see if it is in the second-level cache
if (HasCache)
{
    CacheKey ck = new CacheKey(id, IdentifierType, RootEntityName, session.EntityMode, session.Factory);
    if (Cache.Get(ck, session.Timestamp) != null)
        return false;
}
```

Also logged at https://nhibernate.jira.com/browse/NH-2241

This PR adds a unit test, but unfortunately I couldn't find a solution for this bug.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH2241/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH2241/Fixture.cs
@@ -1,0 +1,44 @@
+﻿using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH2241
+{
+	[TestFixture]
+	public class Fixture : BugTestCase
+	{
+		protected override void OnSetUp()
+		{
+			using (var session = OpenSession())
+			using (var tran = session.BeginTransaction())
+			{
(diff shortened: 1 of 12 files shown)


igitur:
Thank you for the quick fix. Much appreciated.

PR #653, opened 2017-06-23
NH-3488 - Strongly typed DML

hazzik:
This is a rewrite of #391 

Following syntax is supported:

#### Delete
No changes here
````
s.Query<Dog>().Where(x=>x.Name == "Jack").Delete();
````
#### Update
Syntax with `.As(x=>new ...)`/`.Assign(x => x.Set(...))` has been removed
```
s.Query<Dog>().Where(x=>x.Name == "Jack").Update(x=>new Dog { Age=Age+1 });
```
#### Insert
Syntax with `.As(x=>new ...)`/`.Assign(x => x.Set(...))` has been removed
```
s.Query<Cat>().Where(x=>x.Name="Puss").Insert(x=>new Dog { Name="Jack", Friend=x });
```
### TODO:
- [x] Update documentation
- [ ] ? Do not rewrite expression into dictionary
- [ ] Allow anonymous types
  - In order to support UPDATE for abstract classes and protected member, I think we need to add support to write following
    ``` s.Query<Animal>().Update(x=>new { Name = "Tiger" }); ```

/cc @fredericDelaporte @gliljas


gliljas:
I like the cleanup, but the removal of the syntax worries me. It was put there for a reason. Requiring a "new X" makes it necessary to use a real constructor, and that may not be possible in all scenarios. When it's possible, it's convenient, as just one little ".As" away.

/G

hazzik:
@gliljas I have plans to add support for anonymous types for these cases (see the 3rd point in a TODO). Also, I was thinking about Dictionary instead of anonymous types.

fredericDelaporte:
It looks nicer to me this way too, good job.

I think the anonymous/dictionary syntax would be needed for insert too. But that would cause the generic type argument to be mandatory for it to works.

hazzik:
Closed in favor of #391

PR #654, opened 2017-06-28
NH-4043 - Complete keyword registration in dialects.

ngbrown:
This is [NH-4043](https://nhibernate.jira.com/browse/NH-4043).

  - [x] Need to use results from CI Build to finish dialects.
  - [x] Baseline all dialects with ANSI SQL 2003 keywords
  - [x] Lowercase all remaining dialect keywords.


src/NHibernate/Dialect/MsSql2000Dialect.cs
```diff
@@ -66,35 +66,412 @@ protected virtual void RegisterDefaultProperties()
 
 		protected virtual void RegisterKeywords()
 		{
-			RegisterKeyword("top");
-			RegisterKeyword("int");
-			RegisterKeyword("integer"); // a commonly-used alias for 'int'
-			RegisterKeyword("tinyint");
-			RegisterKeyword("smallint");
+			RegisterKeyword("absolute");
```
hazzik: Can this be done in a loop somehow? Like following:

```
var keywords = new [] { ... };  
foreach (var keyword in keywords) 
{
    RegisterKeyword(keyword);
}
```

hazzik: Actually I would like to have following function on base dialect:

```
protected void RegisterKeywords(IEnumerable<string> keywords) //maybe also with params
{
    Keywords.UnionWith(keywords);
}

Comment (ngbrown) [inline src/NHibernate/Dialect/MsSql2000Dialect.cs]: Were you concerned by the number of lines it takes, or by the slowness of hundreds of function calls?

If I change a ReSharper array initializer wrap setting, I get something like this:


		private static readonly string[] keywordsToRegister =
		{
			"absolute", "action", "ada", "add", "admin", "after", "aggregate", "alias", "all", "allocate", "alter", "and", "any",
			"are", "array", "as", "asc", "assertion", "at", "authorization", "avg", "backup", "before", "begin", "between",
			"bigint", "binary", "bit", "bit_length", "blob", "boolean", "both", "breadth", "break", "browse", "bulk", "by",
			"call", "cascade", "cascaded", "case", "cast", "catalog", "char", "char_length", "character", "character_length",
			"check", "checkpoint", "class", "clob", "close", "clustered", "coalesce", "collate", "collation", "column", "commit",
			"completion", "compute", "connect", "connection", "constraint", "constraints", "constructor", "contains",
			"containstable", "continue", "convert", "corresponding", "count", "create", "cross", "cube", "current",
			"current_date", "current_path", "current_role", "current_time", "current_timestamp", "current_user", "cursor",
			"cycle", "data", "database", "date", "datetime", "day", "dbcc", "deallocate", "dec", "decimal", "declare", "default",
			"deferrable", "deferred", "delete", "deny", "depth", "deref", "desc", "describe", "descriptor", "destroy",
			"destructor", "deterministic", "diagnostics", "dictionary", "disconnect", "disk", "distinct", "distributed",
			"domain", "double", "drop", "dummy", "dump", "dynamic", "each", "else", "end", "end-exec", "equals", "errlvl",
			"escape", "every", "except", "exception", "exec", "execute", "exists", "exit", "external", "extract", "false",
			"fetch", "file", "fillfactor", "first", "float", "for", "foreign", "fortran", "found", "free", "freetext",
			"freetexttable", "from", "full", "function", "general", "get", "global", "go", "goto", "grant", "group", "grouping",
			"having", "holdlock", "host", "hour", "identity", "identity_insert", "identitycol", "if", "ignore", "image",
			"immediate", "in", "include", "index", "indicator", "initialize", "initially", "inner", "inout", "input",
			"insensitive", "insert", "int", "integer", "intersect", "interval", "into", "is", "isolation", "iterate", "join",
			"key", "kill", "language", "large", "last", "lateral", "leading", "left", "less", "level", "like", "limit", "lineno",
			"load", "local", "localtime", "localtimestamp", "locator", "lower", "map", "match", "max", "min", "minute",
			"modifies", "modify", "module", "money", "month", "names", "national ", "natural", "nchar", "nclob", "new", "next",
			"no", "nocheck", "nonclustered", "none", "not", "ntext", "null", "nullif", "numeric", "nvarchar", "object",
			"octet_length", "of", "off", "offsets", "old", "on", "only", "open", "opendatasource", "openquery", "openrowset",
			"openxml", "operation", "option", "or", "order", "ordinality", "out", "outer", "output", "over", "overlaps", "pad",
			"parameter", "parameters", "partial", "pascal", "path", "percent", "plan", "position", "postfix", "precision",
			"prefix", "preorder", "prepare", "preserve", "primary", "print", "prior", "privileges", "proc", "procedure",
			"public", "raiserror", "read", "reads", "readtext", "real", "reconfigure", "recursive", "ref", "references",
			"referencing", "relative", "replication", "restore", "restrict", "result", "return", "returns", "revoke", "right",
			"role", "rollback", "rollup", "routine", "row", "rowcount", "rowguidcol", "rows", "rule", "save", "savepoint",
			"schema", "scope", "scroll", "search", "second", "section", "select", "sequence", "session", "session_user", "set",
			"sets", "setuser", "shutdown", "size", "smalldatetime", "smallint", "smallmoney", "some", "space", "specific",
			"specifictype", "sql", "sql_variant", "sqlca", "sqlcode", "sqlerror", "sqlexception", "sqlstate", "sqlwarning",
			"start", "state", "statement", "static", "statistics", "structure", "substring", "sum", "system_user", "table",
			"temporary", "terminate", "text", "textsize", "than", "then", "time", "timestamp", "timezone_hour",
			"timezone_minute", "tinyint", "to", "top", "trailing", "tran", "transaction", "translate", "translation", "treat",
			"trigger", "trim", "true", "truncate", "tsequal", "under", "union", "unique", "uniqueidentifier", "unknown",
			"unnest", "update", "updatetext", "upper", "usage", "use", "user", "using", "value", "values", "varbinary",
			"varchar", "variable", "varying", "view", "waitfor", "when", "whenever", "where", "while", "with", "without", "work",
			"write", "writetext", "year", "zone",
		};

		protected virtual void RegisterKeywords()
		{
			RegisterKeywords(keywordsToRegister);
		}

The nice thing about one-per-line was that it could be easily sorted.

Comment (hazzik) [inline src/NHibernate/Dialect/MsSql2000Dialect.cs]: By this: "by the slowness of hundreds of function calls"

Comment (hazzik) [inline src/NHibernate/Dialect/MsSql2000Dialect.cs]: >The nice thing about one-per-line was that it could be easily sorted.
  You can keep them one by line in the array initializer

Comment (hazzik) [inline src/NHibernate/Dialect/MsSql2000Dialect.cs]: Dont need to put this into the static field

Comment (hazzik) [inline src/NHibernate/Dialect/MsSql2000Dialect.cs]: Ok, does not really matter, leave it as many functions. It seems that C# does not bother with optimizing string arrays (as it does for numbers)

Comment (hazzik) [inline src/NHibernate/Dialect/MsSql2000Dialect.cs]: We could optimize this only by providing this array to the `HashSet<>` constructor

Comment (fredericDelaporte) [inline src/NHibernate/Dialect/MsSql2000Dialect.cs]: So maybe a `RegisterKeywords(params string[] keywords)`?

File: src/NHibernate.Test/Tools/hbm2ddl/SchemaMetadataUpdaterTest/SchemaMetadataUpdaterFixture.cs [hunk 2]
```diff
@@ -56,10 +57,94 @@ public void UpdateReservedWordsInDialect()
...
+				{
+					Console.WriteLine("  RegisterKeyword(\"{0}\");", keyword);
+				}
+			}
+
 			Assert.That(match, Is.EquivalentTo(reservedDb));
 		}
 
 		[Test]
+		public void CheckForExcessReservedWordsHardCodedInDialect()
```
fredericDelaporte: Is this test relevant? It looks like the approach taken to solve this issue is to set keywords on old dialect like SQL2000, while using indeed a more recent server likely having more keywords, like SQL2016 (what we have on the build agent). So indeed testing this against an actual SQL Server 2000 would likely yield some excess words. Does it matter much? I do not think. So I do not think this test is useful.

(Do not try to emulate this with SQL Server compatibility level: that is just a backward compatibility setting, re-enabling obsolete features but not disabling newer ones. For actually testing a SQL Server 2000, only an actual SQL Server 2000 is suitable.)

ngbrown: @fredericDelaporte I did check against an actual SQL Server 2000... yes, checking every database and version is a pain.

ngbrown:
One option is to do what Hibernate did, and automatically register all [ANSI SQL 2003 keywords](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/engine/jdbc/env/spi/AnsiSqlKeywords.java), and then give [each dialect a chance to register additional ones](https://github.com/hibernate/hibernate-orm/blob/f85737cb0b5825fc9491dc1fdedd4ffbac4a4e87/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java#L1883).

src/NHibernate.Test/Tools/hbm2ddl/SchemaMetadataUpdaterTest/SchemaMetadataUpdaterFixture.cs
```diff
@@ -140,7 +147,7 @@ public void CheckForExcessReservedWordsHardCodedInDialect()
 			}
 
 			// Don't fail incase the driver returns nothing.
-			//Assert.That(sf.Dialect.Keywords, Is.EquivalentTo(reservedDb));
+			// This is an info-only test.
```
fredericDelaporte: You may issue an `Assert.Warn` instead. Since a recent PR it will not cause the build to fail.
`Assume.That` would be slightly better, but it currently causes a failure, excepted in a PR where I have started using it and [fixed it](https://github.com/nhibernate/nhibernate-core/pull/627/files#diff-628b6f12bc5afe5f072a2e2f1b01f576).

src/NHibernate/Tool/hbm2ddl/SchemaMetadataUpdater.cs
```diff
@@ -29,16 +37,9 @@ public static void QuoteTableAndColumns(Configuration configuration)
...
-		{
-			var dialect = Dialect.Dialect.GetDialect(cfgProperties);
-			var connectionHelper = new ManagedProviderConnectionHelper(cfgProperties);
-			return GetReservedWords(dialect, connectionHelper);
-		}
-
 		private static ISet<string> GetReservedWords(Dialect.Dialect dialect, IConnectionHelper connectionHelper)
 		{
-			ISet<string> reservedDb = new HashSet<string>();
+			ISet<string> reservedDb = new HashSet<string>(dialect.Keywords);
```
fredericDelaporte: Causes `Update` code to try re-add all the already known words. Would be better to avoid this for `Update` case.

src/NHibernate.Test/Tools/hbm2ddl/SchemaMetadataUpdaterTest/SchemaMetadataUpdaterFixture.cs
```diff
@@ -56,10 +59,98 @@ public void UpdateReservedWordsInDialect()
...
+				Console.WriteLine("Update Dialect {0} with RegisterKeyword:", sf.Dialect.GetType().Name);
+				foreach (var keyword in differences.OrderBy(x => x))
+				{
+					Console.WriteLine("  RegisterKeyword(\"{0}\");", keyword);
+				}
+			}
+
+			if (sf.ConnectionProvider.Driver is OdbcDriver)
+			{
+				Assert.Inconclusive("ODBC has excess keywords reserved");
```
fredericDelaporte: Not supported by the "result comparison tool" currently, causing a build failure. You already need this [change in teamcity.build file](https://github.com/nhibernate/nhibernate-core/pull/627/files#diff-628b6f12bc5afe5f072a2e2f1b01f576).

Feel free to replicate it in your PR, I will adjust mine if your get merged before.

fredericDelaporte: Now available as a PR on its own. #655

src/NHibernate/Tool/hbm2ddl/SchemaMetadataUpdater.cs
```diff
@@ -8,17 +9,24 @@ namespace NHibernate.Tool.hbm2ddl
...
+			var connectionHelper = new SuppliedConnectionProviderConnectionHelper(sessionFactory.ConnectionProvider);
+			sessionFactory.Dialect.Keywords.UnionWith(GetReservedWords(dialect, connectionHelper));
 		}
 
 		public static void QuoteTableAndColumns(Configuration configuration)
 		{
-			ISet<string> reservedDb = GetReservedWords(configuration.GetDerivedProperties());
+			// Instatiates a new instance of the dialect so doesn't benefit from the Update call.
+			var dialect = Dialect.Dialect.GetDialect(configuration.GetDerivedProperties());
+			QuoteTableAndColumns(configuration, dialect);
```
ngbrown: So, the problem was, that the `QuoteTableAndColumns` wasn't even using the keywords loaded into the Dialect.  This was a major problem on the .NET Standard branch because there is no functioning auto-importing, so schema names weren't getting quoted at all, because the DbConnection based GetSchema had to return an empty list.

The auto-import is enabled if auto-quoting is enabled, and auto-import takes place first during the session factory construction.

I think the only paths that use auto-quoting without going through the session factory construction, and thus the auto-importing, are some unit tests that are trying to test just that feature.  For those the auto-quoting to conflate auto-importing, it's not allowing the tests the isolation they may desire.

fredericDelaporte: What about [here](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Tool/hbm2ddl/SchemaExport.cs#L63) and [here](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Tool/hbm2ddl/SchemaUpdate.cs#L144) ? 

Those are not code from NHibernate.Test, and your change removes for them the addition of reserved word by querying the db.

src/NHibernate.Test/Tools/hbm2ddl/SchemaMetadataUpdaterTest/SchemaMetadataUpdaterFixture.cs
fredericDelaporte: This one a merge, you can rebase/merge master for avoiding `Assert.Inconclusive` causing a failure for builds using results comparison.

src/NHibernate/Tool/hbm2ddl/SchemaMetadataUpdater.cs
ngbrown: Ok, thanks for the feedback.

I reverted the original signature for `QuoteTableAndColumns` to it's previous behavior, marked it obsolete, and nothing in NHibernate now uses it.  Everywhere that calls the new `QuoteTableAndColumns` overload calls `Update` beforehand (except for the two explicit unit tests).

I also refactored the `GetReservedWords` to better encapsulate the dispose pattern they are using (finally { Release() }).

src/NHibernate/Tool/hbm2ddl/SchemaUpdate.cs
```diff
@@ -141,7 +141,8 @@ public void Execute(Action<string> scriptAction, bool doUpdate)
 			autoKeyWordsImport = autoKeyWordsImport.ToLowerInvariant();
 			if (autoKeyWordsImport == Hbm2DDLKeyWords.AutoQuote)
 			{
-				SchemaMetadataUpdater.QuoteTableAndColumns(configuration);
+				SchemaMetadataUpdater.Update(configuration, dialect);
```
fredericDelaporte: Need to be moved in some initialization method (in constructors indeed). Here it will needlessly reload the keywords at each `Execute` call after the first.

fredericDelaporte:
Here is the output of `EnsureReservedWordsHardCodedInDialect` for a Sap SQL Anywhere 17 (previously Sybase):

```
Update Dialect SybaseSQLAnywhere12Dialect with RegisterKeyword:
  RegisterKeyword("aes_decrypt");
  RegisterKeyword("asc");
  RegisterKeyword("attach");
  RegisterKeyword("backup");
  RegisterKeyword("bit");
  RegisterKeyword("bottom");
  RegisterKeyword("break");
  RegisterKeyword("capability");
  RegisterKeyword("cascade");
  RegisterKeyword("char_convert");
  RegisterKeyword("checkpoint");
  RegisterKeyword("clear");
  RegisterKeyword("comment");
  RegisterKeyword("compressed");
  RegisterKeyword("conflict");
  RegisterKeyword("convert");
  RegisterKeyword("dbspace");
  RegisterKeyword("deleting");
  RegisterKeyword("desc");
  RegisterKeyword("detach");
  RegisterKeyword("elseif");
  RegisterKeyword("encrypted");
  RegisterKeyword("endif");
  RegisterKeyword("exception");
  RegisterKeyword("executing");
  RegisterKeyword("executing_user");
  RegisterKeyword("existing");
  RegisterKeyword("externlogin");
  RegisterKeyword("extract");
  RegisterKeyword("force");
  RegisterKeyword("forward");
  RegisterKeyword("goto");
  RegisterKeyword("holdlock");
  RegisterKeyword("identified");
  RegisterKeyword("index");
  RegisterKeyword("inserting");
  RegisterKeyword("install");
  RegisterKeyword("instead");
  RegisterKeyword("integrated");
  RegisterKeyword("invoking");
  RegisterKeyword("invoking_user");
  RegisterKeyword("iq");
  RegisterKeyword("isolation");
  RegisterKeyword("json");
  RegisterKeyword("kerberos");
  RegisterKeyword("key");
  RegisterKeyword("lock");
  RegisterKeyword("login");
  RegisterKeyword("long");
  RegisterKeyword("membership");
  RegisterKeyword("message");
  RegisterKeyword("mode");
  RegisterKeyword("modify");
  RegisterKeyword("noholdlock");
  RegisterKeyword("notify");
  RegisterKeyword("nvarchar");
  RegisterKeyword("off");
  RegisterKeyword("openstring");
  RegisterKeyword("openxml");
  RegisterKeyword("option");
  RegisterKeyword("options");
  RegisterKeyword("others");
  RegisterKeyword("passthrough");
  RegisterKeyword("pivot");
  RegisterKeyword("print");
  RegisterKeyword("privileges");
  RegisterKeyword("proc");
  RegisterKeyword("procedure_owner");
  RegisterKeyword("publication");
  RegisterKeyword("raiserror");
  RegisterKeyword("readtext");
  RegisterKeyword("reference");
  RegisterKeyword("refresh");
  RegisterKeyword("remote");
  RegisterKeyword("remove");
  RegisterKeyword("rename");
  RegisterKeyword("reorganize");
  RegisterKeyword("resource");
  RegisterKeyword("restore");
  RegisterKeyword("restrict");
  RegisterKeyword("rowtype");
  RegisterKeyword("save");
  RegisterKeyword("session");
  RegisterKeyword("session_user");
  RegisterKeyword("setuser");
  RegisterKeyword("share");
  RegisterKeyword("spatial");
  RegisterKeyword("sqlcode");
  RegisterKeyword("stop");
  RegisterKeyword("subtrans");
  RegisterKeyword("subtransaction");
  RegisterKeyword("synchronize");
  RegisterKeyword("temporary");
  RegisterKeyword("timeline");
  RegisterKeyword("tinyint");
  RegisterKeyword("truncate");
  RegisterKeyword("tsequal");
  RegisterKeyword("unbounded");
  RegisterKeyword("uniqueidentifier");
  RegisterKeyword("unpivot");
  RegisterKeyword("unsigned");
  RegisterKeyword("updating");
  RegisterKeyword("validate");
  RegisterKeyword("varbinary");
  RegisterKeyword("varbit");
  RegisterKeyword("variable");
  RegisterKeyword("varray");
  RegisterKeyword("view");
  RegisterKeyword("wait");
  RegisterKeyword("waitfor");
  RegisterKeyword("work");
  RegisterKeyword("writetext");
  RegisterKeyword("xml");
```

ngbrown:
@fredericDelaporte I had downloaded SQL Anywhere 16 and gotten similar results, but which Dialect should that be added to?  To a new Dialect for 16/17 or to one of the older ones, 10, 11, or 12?

fredericDelaporte:
Adding a new dialect without testing if it needs some changes should not be done I think. (That is the reason I will likely not contribute the specific driver I have locally written for accessing Anywhere 17, which has once again change its assembly name.) I rather let actual Anywhere users contribute these.

I am more for adding them to an old dialect. Having more keyword than needed should not be harmful, while having less is surely harmful.

src/NHibernate/Tool/hbm2ddl/SchemaUpdate.cs
ngbrown: I'm not a big fan of doing remote IO in a constructor.  There will be no improvement in this case because everywhere `Execute` is called, it is immediately after `new`ing  up a new `SchemaUpdate` instance.

fredericDelaporte: For consistency with `SchemaExport`, it could be done on a `Initialize` method, called only once from method requiring it. `SchemaUpdate` can be used by users, we cannot assume they will never call many time its execute.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2246,6 +2248,11 @@ protected void RegisterKeyword(string word)
 			Keywords.Add(word);
 		}
 
+		protected void RegisterKeywords(params string[] keywords)
+		{
+			Keywords.UnionWith(keywords);
```
ngbrown: It should be  higher performance to not generate new strings, so I made the `HashSet`'s case insensitive.

src/NHibernate/Dialect/AnsiSqlKeywords.cs
```diff
@@ -0,0 +1,267 @@
...
+	public class AnsiSqlKeywords
+	{
+		/// <summary>
+		/// Retrieve all keywords defined by ANSI SQL:2003
+		/// </summary>
+		public static readonly IReadOnlyCollection<string> Sql2003;
+
+		static AnsiSqlKeywords()
+		{
+			var keywordsSql2003 = new HashSet<string>
```
hazzik: Shouldn't this be case insensitive as well?

ngbrown: @hazzik, It's not necessary.  The AnsiSqlKeywords are iterated over as an `IEnumerable` when initializing the keyword `HashSet`.

src/NHibernate/Dialect/Schema/AbstractDataBaseSchema.cs
```diff
@@ -99,6 +99,13 @@ public virtual ISet<string> GetReservedWords()
 			{
 				result.Add(row["ReservedWord"].ToString());
 			}
+
+			DataTable dtTypes = connection.GetSchema(DbMetaDataCollectionNames.DataTypes);
+			foreach (DataRow row in dtTypes.Rows)
+			{
+				result.Add(row["TypeName"].ToString());
```
ngbrown: Done with `IncludeDataTypesInReservedWords`.... squashed into previous commit.

src/NHibernate/Dialect/AnsiSqlKeywords.cs
hazzik: Ok, why this is even a set? we just need an array, dont we?

ngbrown: I was following what Hibernate had.  They had a HashSet too, even though I assume Java unions HashSets the same way.

ngbrown: Either way, I changed it to an array for lower initialization cost.

src/NHibernate/Dialect/Dialect.cs
fredericDelaporte: I do not know if this a really more performant overall, token lookups included, since an `Ordinal` comparison is maybe faster than an `OrdinalIgnoreCase`. But at least it is more robust, so I prefer having that ignore case.

src/NHibernate/Dialect/Schema/AbstractDataBaseSchema.cs
```diff
@@ -99,6 +101,16 @@ public virtual ISet<string> GetReservedWords()
 			{
 				result.Add(row["ReservedWord"].ToString());
 			}
+
+			if (IncludeDataTypesInReservedWords)
+			{
+				DataTable dtTypes = connection.GetSchema(DbMetaDataCollectionNames.DataTypes);
+				foreach (DataRow row in dtTypes.Rows)
```
fredericDelaporte: Anywhere 17 supports this too, without a different name (using `TypeName` works for it). But its schema specific class overrides that method, and has not be updated for including data types. Maybe should it be updated.

src/NHibernate/Dialect/SybaseSQLAnywhere10Dialect.cs
```diff
@@ -63,17 +63,17 @@ public SybaseSQLAnywhere10Dialect()
...
 			RegisterColumnType(DbType.AnsiString, 2147483647, "LONG VARCHAR");
-			RegisterColumnType(DbType.StringFixedLength, "NCHAR(1)");
+			RegisterColumnType(DbType.StringFixedLength, "NCHAR(255)");
 			RegisterColumnType(DbType.StringFixedLength, 32767, "NCHAR($l)");
-			RegisterColumnType(DbType.String, "NVARCHAR(1)");
+			RegisterColumnType(DbType.String, "NVARCHAR(255)");
 			RegisterColumnType(DbType.String, 32767, "NVARCHAR($l)");
 			RegisterColumnType(DbType.String, 2147483647, "LONG NVARCHAR");
-			RegisterColumnType(DbType.Binary, "BINARY(1)");
+			RegisterColumnType(DbType.Binary, "BINARY(8000)");
```
fredericDelaporte: Huh, I understand now why basic tests were failing with string truncation now... But this fix deserves a Jira on its own.

I have created [NH-4046](https://nhibernate.jira.com/browse/NH-4046). I think it should be fixed at least as a separated commit, maybe even a separated PR. `SybaseASE15Dialect` has the issue too.

We should be able to use `SybaseASA9Dialect` default lengths safely, since this is an old name for Anywhere. In this regard, `Binary(8000)` is maybe too big. `SybaseASA9Dialect` declares `Binary(255)` instead.

src/NHibernate/Tool/hbm2ddl/SchemaMetadataUpdater.cs
```diff
@@ -29,41 +44,54 @@ public static void QuoteTableAndColumns(Configuration configuration)
...
 			finally
 			{
 				connectionHelper.Release();
 			}
-			return reservedDb;
+		}
+
+		private static ISet<string> GetReservedWords(Dialect.Dialect dialect, IConnectionHelper connectionHelper)
+		{
+			ISet<string> reservedWords = new HashSet<string>(dialect.Keywords, StringComparer.OrdinalIgnoreCase);
```
hazzik: This HashSet is completely useless:

1. We create this hash set from `dialect.Keywords`.
2. We add another hash set from `metaData.GetReservedWords()`.
3. We merge this hash set into `dialect.Keywords`;

src/NHibernate/Tool/hbm2ddl/SchemaUpdate.cs
```diff
@@ -42,6 +43,24 @@ public SchemaUpdate(Configuration cfg, Settings settings)
...
+		private void Initialize()
+		{
+			if (wasInitialized)
+			{
+				return;
+			}
+
+			string autoKeyWordsImport = PropertiesHelper.GetString(Environment.Hbm2ddlKeyWords, configuration.Properties, "not-defined");
+			autoKeyWordsImport = autoKeyWordsImport.ToLowerInvariant();
+			if (autoKeyWordsImport == Hbm2DDLKeyWords.AutoQuote)
```
hazzik: Here shall be the same check as in SessionFactoryImpl.

src/NHibernate/Tool/hbm2ddl/SchemaMetadataUpdater.cs
hazzik: >A number of code paths do call QuoteTableAndColumns without updating the dialect before.

@fredericDelaporte I think this is intentional. We do not need to update the dialect on SchemaExport/SchemaUpdate as it is a throw-away dialect. However, it does not harm.

hazzik:
I have cleaned the code:

- Dialect encapsulates Keywords
- AbstractDataBaseSchema and SybaseAnywhereColumnMetaData return case insensitive hash set (in fact, they can return just an array/IEnumerable<string>)
- SchemaMetadataUpdater does not do unnecessary hash sets recast
- SchemaMetadataUpdater just sets IsQuoted for table and column names, so it does not do unnecessary string operations

ngbrown:
@hazzik Thanks for the cleanup.  I assume I still need to split off the type changes into NH-4046?

hazzik:
yes.

ngbrown:
I just added a commit to undo it.  I could have re-written the history, but maybe this whole thing will be squashed on merge anyways?

hazzik:
There is JIRA issue will be fixed by this PR: https://nhibernate.jira.com/browse/NH-3158

hazzik:
I'm going to rewrite the history. Split into 2 commits.

hazzik:
I've regrouped changes into 2 commits. One is NH-3158, another one is NH-4043. There were no tests that checking that dialect keywords were used.

ngbrown:
Looks good.

src/NHibernate/Tool/hbm2ddl/SchemaUpdate.cs
fredericDelaporte: > Here, probably, shall be the same check as in SessionFactoryImpl.

It would look more consistent, but would complicate code for nothing, in its current state. The check in `SessionFactoryImpl` treats `AutoQuote` and `KeywordImport` as two separated options. But they indeed are only one configuration setting, currently not allowing having auto-quoting without having keyword-import.

To have things actually consistent, we would need to put a helper somewhere for getting those two booleans from the actual configuration setting, or refactor `SchemaUpdate` and `SchemaExport` for taking/building a session factory setting object. I do not think it is worth the hassle.

PR #655, opened 2017-07-02
NH-4028 - support another kind of inconclusive result.

fredericDelaporte:
[NH-4028](https://nhibernate.jira.com/browse/NH-4028) - Supports inconclusive tests in result comparison.

Follow up on #641. As demonstrated in Nathan's #654 and in my #627, another kind of inconclusive results are not supported by the result comparison tool, and classified as failure.

### DIFF
--- a/teamcity.build
+++ b/teamcity.build
@@ -289,9 +289,10 @@
 		public enum ResultStatus
 		{
 			None,
+			Failure,
 			Ignored,
-			Success,
-			Failure
+			Inconclusive,
+			Success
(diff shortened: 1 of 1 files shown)


fredericDelaporte:
Or should I just bypass review for such minor changes? (Or do direct commit? But when that is not as trivial as a `?.` in a test file for fixing a null ref exception, I prefer to check the build results before.)

fredericDelaporte:
Thanks Alexander.

PR #657, opened 2017-07-06
NH-4003 - removing from configuration schema a dropped property after…

fredericDelaporte:
… session opening port from hibernate

I have forgotten about nhibernate-configuration.xsd file. Removing from it the old obsolete and now dropped by NH-4003 parameter. (On NH-4011 I have updated the xsd too, for its own new parameter.)


fredericDelaporte:
Any thoughts against merging this?

PR #659, opened 2017-07-11
NH-4009 - Handling db-only methods without exceptions.

fredericDelaporte:
[NH-4009](https://nhibernate.jira.com/browse/NH-4009) - Allow marking a Linq extension as db only

As per Alexander's request, available directly on `LinqExtensionMethodAttribute`, but still available as a separated `DBOnlyAttribute` (inherited by `LinqExtensionMethodAttribute`) too for method not being simply converted to function calls, like `MappedAs` or other custom extensions.

They have a `DBOnly` boolean property, `true` by default for `DBOnlyAttribute`, `false` by default for `LinqExtensionMethodAttribute`.

If a user combine both, its method will be considered "DB only" if at least one of them has `DBOnly` being `true`.

This change allows to more silently swallow unexpected partial evaluation failures, state in which the master branch was since [NH-3961](https://nhibernate.jira.com/browse/NH-3961).

With this change, [NH-3386](https://nhibernate.jira.com/browse/NH-3386) (accidentally "fixed" by NH-3961) will be supported by setting `DBOnly = true` on the `LinqExtensionMethodAttribute`, instead of being supported by throwing any exception (or even by trying to implement an in-memory implementation but failing and accidentally throwing at that).

### DIFF
--- a/doc/reference/modules/query_linq.xml
+++ b/doc/reference/modules/query_linq.xml
@@ -67,6 +67,10 @@ using NHibernate.Linq;]]></programlisting>
     session.Query<Cat>()
         .Where(c => c.BirthDate == DateTime.Today.MappedAs(NHibernateUtil.Date))
         .ToList();]]></programlisting>
+    <programlisting><![CDATA[IList<Cat> cats =
+    session.Query<Cat>()
+        .Where(c => c.Name == "Max".MappedAs(TypeFactory.Basic("AnsiString(200)")))
+        .ToList();]]></programlisting>
   </sect1>
 
   <sect1 id="querylinq-supportedmethods">
@@ -656,11 +660,20 @@ IList<Cat> cats =
     }
 }]]></programlisting>
       <para>
-        The method call will always be translated to SQL if at least one of the parameters of the
-        method call has its value originating from an entity. Otherwise, the Linq provider will try to
-        evaluate the method call with .Net runtime instead. Since NHibernate 5.0, if this runtime
-        evaluation fails (throws an exception), then the method call will be translated to SQL too.
+        Since NHibernate v5.0, the Linq provider will no more evaluate in-memory the method call
+        even when it does not depend on the queried data. If you wish to have the method call evaluated
+        before querying whenever possible, and then replaced in the query by its resulting value, specify
+        <literal>LinqExtensionPreEvaluation.AllowPreEvaluation</literal> on the attribute.
       </para>
+      <programlisting><![CDATA[public static class CustomLinqExtensions
+{
+    [LinqExtensionMethod("dbo.aCustomFunction", LinqExtensionPreEvaluation.AllowPreEvaluation)]
+    public static string ACustomFunction(this string input, string otherInput)
+    {
+        // In-memory evaluation implementation.
+        return input.Replace(otherInput, "blah");
+    }
+}]]></programlisting>
     </sect2>
 
     <sect2 id="querylinq-extending-generator">
@@ -766,6 +779,13 @@ cfg.LinqToHqlGeneratorsRegistry<ExtendedLinqToHqlGeneratorsRegistry>();
         .Min();]]></programlisting>
       <para>
         (Of course, the same result could be obtained with <literal>(DateTime?)(c.BirthDate)</literal>.)
+      </para>
+      <para>
+        By default, the Linq provider will try to evaluate the method call with .Net runtime
+        whenever possible, instead of translating it to SQL. It will not do it if at least one
+        of the parameters of the method call has its value originating from an entity, or if
+        the method is marked with the <literal>NoPreEvaluation</literal> attribute (available
+        since NHibernate 5.0).
       </para>
     </sect2>
   </sect1>
--- a/src/NHibernate.Test/Linq/ExtensionMethods.cs
+++ b/src/NHibernate.Test/Linq/ExtensionMethods.cs
@@ -1,19 +1,38 @@
-﻿using NHibernate.Linq;
+﻿using System;
+using NHibernate.Linq;
 
 namespace NHibernate.Test.Linq
 {
-    static class ExtensionMethods
-    {
-        [LinqExtensionMethod("Replace")]
-        public static string ReplaceExtension(this string subject, string search, string replaceWith)
-        {
-            return null;
-        }
+	static class ExtensionMethods
+	{
+		[LinqExtensionMethod("Replace")]
+		public static string ReplaceExtension(this string subject, string search, string replaceWith)
+		{
+			throw new InvalidOperationException("To be translated to SQL only");
+		}
 
-        [LinqExtensionMethod]
-        public static string Replace(this string subject, string search, string replaceWith)
-        {
-            return null;
-        }
-    }
+		[LinqExtensionMethod]
+		public static string Replace(this string subject, string search, string replaceWith)
+		{
+			throw new InvalidOperationException("To be translated to SQL only");
+		}
+
+		// NH4 default behavior
+		[LinqExtensionMethod("Replace", LinqExtensionPreEvaluation.AllowPreEvaluation)]
+		public static string ReplaceWithEvaluation(this string subject, string search, string replaceWith)
+		{
+			if (subject == null)
+				throw new ArgumentNullException(nameof(subject));
+			return subject.Replace(search, replaceWith) + " (done in-memory)";
+		}
+
+		// Just for checking weird combination
+		[LinqExtensionMethod("Replace", LinqExtensionPreEvaluation.AllowPreEvaluation), NoPreEvaluation]
+		public static string ReplaceWithEvaluation2(this string subject, string search, string replaceWith)
+		{
+			if (subject == null)
+				throw new ArgumentNullException(nameof(subject));
+			return subject.Replace(search, replaceWith) + " (done in-memory)";
+		}
+	}
 }
--- a/src/NHibernate.Test/Linq/FunctionTests.cs
+++ b/src/NHibernate.Test/Linq/FunctionTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Text.RegularExpressions;
 using NHibernate.DomainModel;
 using NHibernate.DomainModel.Northwind.Entities;
 using NHibernate.Linq;
@@ -114,17 +115,58 @@ namespace NHibernate.Test.Linq
 		[Test]
 		public void ReplaceFunction()
 		{
+			var suppliedName = "Anne";
 			var query = from e in db.Employees
 						where e.FirstName.StartsWith("An")
 						select new
-								{
-									Before = e.FirstName,
-									AfterMethod = e.FirstName.Replace("An", "Zan"),
-									AfterExtension = ExtensionMethods.Replace(e.FirstName, "An", "Zan"),
-									AfterExtension2 = e.FirstName.ReplaceExtension("An", "Zan")
-								};
+							{
+								Before = e.FirstName,
+								// This one call the standard string.Replace, not the extension. The linq registry handles it.
+								AfterMethod = e.FirstName.Replace("An", "Zan"),
+								AfterExtension = ExtensionMethods.Replace(e.FirstName, "An", "Zan"),
+								AfterNamedExtension = e.FirstName.ReplaceExtension("An", "Zan"),
+								AfterEvaluableExtension = e.FirstName.ReplaceWithEvaluation("An", "Zan"),
+								AfterEvaluable2Extension = e.FirstName.ReplaceWithEvaluation2("An", "Zan"),
+							BeforeConst = suppliedName,
+								// This one call the standard string.Replace, not the extension. The linq registry handles it.
+								AfterMethodConst = suppliedName.Replace("An", "Zan"),
+								AfterExtensionConst = ExtensionMethods.Replace(suppliedName, "An", "Zan"),
+								AfterNamedExtensionConst = suppliedName.ReplaceExtension("An", "Zan"),
+								AfterEvaluableExtensionConst = suppliedName.ReplaceWithEvaluation("An", "Zan"),
+								AfterEvaluable2ExtensionConst = suppliedName.ReplaceWithEvaluation2("An", "Zan")
+						};
+			var results = query.ToList();
+			var s = ObjectDumper.Write(results);
 
-			var s = ObjectDumper.Write(query);
+			foreach (var result in results)
+			{
+				var expectedDbResult = Regex.Replace(result.Before, "An", "Zan", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+				Assert.That(result.AfterMethod, Is.EqualTo(expectedDbResult), $"Wrong {nameof(result.AfterMethod)} value");
+				Assert.That(result.AfterExtension, Is.EqualTo(expectedDbResult), $"Wrong {nameof(result.AfterExtension)} value");
+				Assert.That(result.AfterNamedExtension, Is.EqualTo(expectedDbResult), $"Wrong {nameof(result.AfterNamedExtension)} value");
+				Assert.That(result.AfterEvaluableExtension, Is.EqualTo(expectedDbResult), $"Wrong {nameof(result.AfterEvaluableExtension)} value");
+				Assert.That(result.AfterEvaluable2Extension, Is.EqualTo(expectedDbResult), $"Wrong {nameof(result.AfterEvaluable2Extension)} value");
+
+				var expectedDbResultConst = Regex.Replace(result.BeforeConst, "An", "Zan", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+				var expectedInMemoryResultConst = result.BeforeConst.Replace("An", "Zan");
+				var expectedInMemoryExtensionResultConst = result.BeforeConst.ReplaceWithEvaluation("An", "Zan");
+				Assert.That(result.AfterMethodConst, Is.EqualTo(expectedInMemoryResultConst), $"Wrong {nameof(result.AfterMethodConst)} value");
+				Assert.That(result.AfterExtensionConst, Is.EqualTo(expectedDbResultConst), $"Wrong {nameof(result.AfterExtensionConst)} value");
+				Assert.That(result.AfterNamedExtensionConst, Is.EqualTo(expectedDbResultConst), $"Wrong {nameof(result.AfterNamedExtensionConst)} value");
+				Assert.That(result.AfterEvaluableExtensionConst, Is.EqualTo(expectedInMemoryExtensionResultConst), $"Wrong {nameof(result.AfterEvaluableExtensionConst)} value");
+				Assert.That(result.AfterEvaluable2ExtensionConst, Is.EqualTo(expectedInMemoryExtensionResultConst), $"Wrong {nameof(result.AfterEvaluable2ExtensionConst)} value");
+			}
+
+			// Should cause ReplaceWithEvaluation to fail
+			suppliedName = null;
+			var failingQuery = from e in db.Employees
+						where e.FirstName.StartsWith("An")
+						select new
+						{
+							Before = e.FirstName,
+							AfterEvaluableExtensionConst = suppliedName.ReplaceWithEvaluation("An", "Zan")
+						};
+			Assert.That(() => failingQuery.ToList(), Throws.InstanceOf<HibernateException>().And.InnerException.InstanceOf<ArgumentNullException>());
 		}
 
 		[Test]
--- a/src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
+++ b/src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
@@ -2,17 +2,106 @@ using System;
 
 namespace NHibernate.Linq
 {
-	public class LinqExtensionMethodAttribute: Attribute
+	/// <summary>
+	/// Flag a method as being a SQL function call for the linq-to-nhibernate provider. Its
+	/// parameters will be used as the function call parameters.
+	/// </summary>
+	public class LinqExtensionMethodAttribute : LinqExtensionMethodAttributeBase
 	{
+		/// <summary>
+		/// Default constructor. The method call will be translated by the linq provider to
+		/// a function call having the same name than the method.
+		/// </summary>
 		public LinqExtensionMethodAttribute()
-		{
-		}
+			: base(LinqExtensionPreEvaluation.NoEvaluation) { }
 
+		/// <summary>
+		/// Constructor specifying a SQL function name.
+		/// </summary>
+		/// <param name="name">The name of the SQL function.</param>
 		public LinqExtensionMethodAttribute(string name)
+			: this(name, LinqExtensionPreEvaluation.NoEvaluation) { }
+
+		/// <summary>
+		/// Constructor allowing to specify a <see cref="LinqExtensionPreEvaluation"/> for the method.
+		/// </summary>
+		/// <param name="preEvaluation">Should the method call be pre-evaluated when not depending on
+		/// queried data? Default is <see cref="LinqExtensionPreEvaluation.NoEvaluation"/>.</param>
+		public LinqExtensionMethodAttribute(LinqExtensionPreEvaluation preEvaluation)
+			: base(preEvaluation) { }
+
+		/// <summary>
+		/// Constructor for specifying a SQL function name and a <see cref="LinqExtensionPreEvaluation"/>.
+		/// </summary>
+		/// <param name="name">The name of the SQL function.</param>
+		/// <param name="preEvaluation">Should the method call be pre-evaluated when not depending on
+		/// queried data? Default is <see cref="LinqExtensionPreEvaluation.NoEvaluation"/>.</param>
+		public LinqExtensionMethodAttribute(string name, LinqExtensionPreEvaluation preEvaluation)
+			: base(preEvaluation)
 		{
 			Name = name;
 		}
 
-		public string Name { get; private set; }
+		/// <summary>
+		/// The name of the SQL function.
+		/// </summary>
+		public string Name { get; }
+	}
+
+	/// <summary>
+	/// Can flag a method as not being callable by the runtime, when used in Linq queries.
+	/// If the method is supported by the linq-to-nhibernate provider, it will always be converted
+	/// to the corresponding SQL statement.
+	/// Otherwise the linq-to-nhibernate provider evaluates method calls when they do not depend on
+	/// the queried data.
+	/// </summary>
+	public class NoPreEvaluationAttribute : LinqExtensionMethodAttributeBase
+	{
+		/// <summary>
+		/// Default constructor.
+		/// </summary>
+		public NoPreEvaluationAttribute()
+			: base(LinqExtensionPreEvaluation.NoEvaluation) { }
+	}
+
+	/// <summary>
+	/// Base class for Linq extension attributes.
+	/// </summary>
+	public abstract class LinqExtensionMethodAttributeBase : Attribute
+	{
+		/// <summary>
+		/// Should the method call be pre-evaluated when not depending on queried data? If it can,
+		/// it would then be evaluated and replaced by the resulting (parameterized) constant expression 
+		/// in the resulting SQL query.
+		/// </summary>
+		public LinqExtensionPreEvaluation PreEvaluation { get; }
+
+		/// <summary>
+		/// Default constructor.
+		/// </summary>
+		/// <param name="preEvaluation">Should the method call be pre-evaluated when not depending on queried data?</param>
+		protected LinqExtensionMethodAttributeBase(LinqExtensionPreEvaluation preEvaluation)
+		{
+			PreEvaluation = preEvaluation;
+		}
+	}
+
+	/// <summary>
+	/// Possible method call behaviors when the linq to NHibernate provider pre-evaluates
+	/// expressions before translating them to SQL.
+	/// </summary>
+	public enum LinqExtensionPreEvaluation
+	{
+		/// <summary>
+		/// The method call will not be evaluated even if its arguments do not depend on queried data.
+		/// It will always be translated to the corresponding SQL statement.
+		/// </summary>
+		NoEvaluation,
+		/// <summary>
+		/// If the method call does not depend on queried data, the method call will be evaluated and replaced
+		/// by the resulting (parameterized) constant expression in the resulting SQL query. A throwing
+		/// method implementation will cause the query to throw.
+		/// </summary>
+		AllowPreEvaluation
 	}
 }
--- a/src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
+++ b/src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
@@ -1,3 +1,5 @@
+using System;
+using System.Linq;
 using System.Linq.Expressions;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Parsing;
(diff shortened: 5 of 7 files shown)


doc/reference/modules/query_linq.xml
```diff
@@ -67,6 +67,10 @@ using NHibernate.Linq;]]></programlisting>
     session.Query<Cat>()
         .Where(c => c.BirthDate == DateTime.Today.MappedAs(NHibernateUtil.Date))
         .ToList();]]></programlisting>
+    <programlisting><![CDATA[IList<Cat> cats =
+    session.Query<Cat>()
+        .Where(c => c.Name == "Max".MappedAs(TypeFactory.Basic("AnsiString(200)")))
+        .ToList();]]></programlisting>
```
fredericDelaporte: Sneakily added another example here by the way... Not utterly unrelated, since `MappedAs` benefits from `DBOnly` (avoids having its implementation relying on throwing exceptions), but a bit remote maybe.

src/NHibernate/Linq/DBOnlyAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace NHibernate.Linq
+{
+	/// <summary>
+	/// Can flag a method as not being callable by the runtime, when used in Linq queries.
+	/// If the method is supported by the linq-to-nhibernate provider, and <see cref="DBOnly"/>
+	/// is <see langword="true" />, it will always be converted to the corresponding SQL statement.
+	/// Otherwise the linq-to-nhibernate provider evaluates method calls when they do not depend on
+	/// the queried data. <see cref="DBOnly"/> default value is <see langword="true" />.
+	/// </summary>
+	public class DBOnlyAttribute : Attribute
```
fredericDelaporte: Stating the default value in class comment instead of on property, since this property changes its default value when inherited.

hazzik:
I don't think that DbOnly is required at all. Any method explicitly registered with a generator/LinqExtensionMethodAttribute shall be treated as "db only" and evaluated on db site.

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
```diff
@@ -1,18 +1,34 @@
...
+	/// </summary>
+	public class LinqExtensionMethodAttribute: DBOnlyAttribute
 	{
+		/// <summary>
+		/// Default constructor. The method call will be translated by the linq provider to
+		/// a function call having the same name than the method.
+		/// </summary>
 		public LinqExtensionMethodAttribute()
 		{
+			DBOnly = false;
```
hazzik: What is the purpose of having `DBOnly = false`?

fredericDelaporte: Having the method by default evaluated in memory whenever possible.

For performances reasons, it is usually better to evaluate in memory, once per request, all computation which can be, rather than letting the SQL server doing that. Things that are purely iterative computations are usually a job more suitable for iterative processing engines like .Net than for data querying engines.

Especially when it is a function call: for SQL servers, most functions are "black-boxes", and if they are not correctly flagged as deterministic, they will be evaluated for each row even if their actual result with same arguments will always be the same. Causing extension calls which were previously in-memory evaluated to always be evaluated in SQL as function could be a major breaking change for performances.

When writing Linq queries, it is frequent to write expressions that can be evaluated before hand. Changing the Linq provider for not evaluating them would cause anyone willing to optimize them to extract them in local variables before writing the query, which could be quite a hassle. That is the very purpose of the partial evaluating processor, evaluating all sub-expression which can be.

fredericDelaporte:
> I don't think that DbOnly is required at all. Any method explicitly registered with a generator/LinqExtensionMethodAttribute shall be treated as "db only" and evaluated on the db side.

Easy for `LinqExtensionMethodAttribute` if we really deem that preferable, not so for method registered. It would require the filter to check each method call against registered methods, with signature checks. Quite heavier than checking an attribute.

And here too, I am not convinced this is a good default choice. I consider that all that can be evaluated before hand in memory should be evaluated, instead of transmitting them to SQL. .Net is best as iterative processing, SQL at data processing. Better do iterative processing on .Net side whenever possible.

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
hazzik: >Having the method by default evaluated in memory whenever possible

We need a test for this. But I don't see any practical application of this behavior. 

It seems that "DBOnly=false" is ... just a normal method.

fredericDelaporte: `DBOnly = false` is current NHibernate v4 behavior. Evaluated in memory if call arguments do not depends on queried data, otherwise evaluated by SQL.

hazzik:
I just don't understand what you are going to achieve with "DbOnly=false". If method is instance then linq will throw an unknown method exception. If method is static it will be evaluated and substituted with a result, so it will never make into a query. As it was a just normal method.

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
hazzik: It's not true. Any static method regardless accepting the argument from the query or not is tried be evaluated.

fredericDelaporte:
I do not understand your points either. That just works that way with NHibernate v4, as if dbOnly was false. Are you telling `LinqExtensionMethodAttribute` has currently no purpose? But it has already served me well, and other people too.

fredericDelaporte:
> If method is instance then linq will throw an unknown method exception. If method is static it will be evaluated and substituted with a result, so it will never make into a query. As it was a just normal method.

Never seen that. No, it just works if the method has a working in-memory implementation and a working SQL translation. When it is called on constants, it gets evaluated by the partial evaluating processor, and is replaced by a constant in the resulting query. When it is called on queried entities, the partial evaluating processor will not try to evaluate them and will let the NHibernate linq extension translate them to sql.

hazzik:
I go sleep, but changes without tests do not go into master. There is no sensible application for dbonly=false.

fredericDelaporte:
> I go sleep, but changes without tests do not go into master. There is no sensible application for dbonly=false.

Good night, but maybe reconsider later. `DBOnly = false` is no change, `DBOnly = false` is preserving NHibernate v4 behavior.

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
fredericDelaporte: I will add tests demonstrating that. But here that will be tests about how things were already working in NHibernate v4.

hazzik:
If the behavior will be preserved, why did you change the attributes on existing methods (newid for example)?

hazzik:
Ok, you've made me wake up from the bed. It's bad.

Let's consider these two methods in NH4:
```
	[LinqExtensionMethod(Name="NEWID")]
	public static Guid NewID1_1()
	{
		throw new InvalidOperationException("To be translated to SQL only");
	}
	public static Guid NewID1_2()
	{
		throw new InvalidOperationException("To be translated to SQL only");
	}
	[LinqExtensionMethod(Name="NEWID")]
	public static Guid NewID2_1()
	{
		return Guid.NewGuid();
	}
	public static Guid NewID2_2()
	{
		return Guid.NewGuid();
	}
```
NewID1_1 will throw an exception and will not be evaluated in memory. Later the HqlGenerator will find the [LinqExtension] attribute and will substitute it with "NEWID" sql function call. NewID1_2 will throw an exception and will not be evaluated in memory. Later the hql generator will not be able to find a function substitution for a method and throw NotSupportedException (HqlGeneratorExpressionVisitor.cs, L515). NewID2_1 and NewID2_2 will not throw any exception and will be evaluated by .NET and will never make into the query.

Now consider the same methods with the proposed implementations:

```
	[LinqExtensionMethod("NEWID", DbOnly=true)]
	public static Guid NewID1_1()
	{
		throw new InvalidOperationException("To be translated to SQL only");
	}
	[LinqExtensionMethod("NEWID", DbOnly=false)]
	public static Guid NewID1_2()
	{
		throw new InvalidOperationException("To be translated to SQL only");
	}
	public static Guid NewID1_3() //no attribute
	{
		throw new InvalidOperationException("To be translated to SQL only");
	}
	[LinqExtensionMethod("NEWID", DDOnly=true)]
	public static Guid NewID2_1()
	{
		return Guid.NewGuid();
	}
	[LinqExtensionMethod("NEWID", DDOnly=false)]
	public static Guid NewID2_2()
	{
		return Guid.NewGuid();
	}
	public static Guid NewID2_3() //no attribute
	{
		return Guid.NewGuid();
	}
```

- NewID1_1 will be evaluated into the query as a NEWID function
- NewID1_2 will throw "System.NotSupportedException: PartialEvalException (InvalidOperationException ("To be translated to SQL only"), NewID1_2())"
- NewID1_3 will throw "System.NotSupportedException: PartialEvalException (InvalidOperationException ("To be translated to SQL only"), NewID1_2())" **(sic!)**
- NewID2_1 will be evaluated into the query as a NEWID function
- NewID2_2 will be evaluated into the query as a parameter (constant)
- NewID2_3 will be evaluated into the query as a parameter (constant) **(sic!)**

So as we can see, "DDOnly=false" behaves exactly as there was no attribute applied at all - and therefore it does not make sense to have it. This is the results of the tests I just wrote.

fredericDelaporte:
> If the behavior will be preserved, why did you change the attributes on existing methods (newid for example)?

This one was failing in NHibernate v4, and was ignored. As written at the start of this PR, it was unexpectedly fixed when fixing NH-3961 (the parameter value bad "caching" trouble). Prior to that change, there were no mean to get translated to SQL a method not having any argument, or only called on constants. The first partial evaluation was ignoring its failure and treating it as having to stay in the expression for later processing, but a second partial evaluation was trying to run it again, and was then rejecting the method failure.

NH-3961 has removed that second partial evaluation, causing any failing method to be ignored and let in the tree for being translated to SQL. This has caused NH-3386 to be "fixed" in v5 by the way. But this is not a desirable state, because that would mean that any unexpected failure is swallowed in the hope that the failing method is translatable to SQL. If not, the query parsing will then fail with an unrecognized method, instead of telling the user what was the evaluation failure.

That is why we anyway need to do something, for avoiding swallowing potential actual errors, not only "exceptions raised from methods for forcing them to be translated to SQL". And better implement a way not relying at all on exception being thrown, contrary to what do master since NH-3961 for extension methods like newid, and NHibernate since `MappedAs` is around. This PR does that.

The possibility to disable that new `DBOnly` is not required for doing that. It is there for still allowing calls to be evaluated in memory when they can be, as it is currently done by NHibernate v4. Not having that would mean by example that a string concatenation in a lambda would be translated to SQL even if it could have been done before emitting the query.

Like
```C#
public IList<Person> Search(nameStart)
{
    // ...
    queryable = queryable.Where(e => e.Name.Like(nameStart + "%"));
}
```
Say the method is called with "Bo"; `Like` is another extension method already working (and tested, without any attribute, so treated as if db only was false).
The current partial evaluation cause the SQL query to be like:
```sql
where e.Name like @p1
```
With `@p1 = 'Bo%'`
Treating `String.Concat` as dbonly because it would be registered in recognized linqs method (not currently the case) would cause the SQL query to be then:
```sql
where e.Name like @p1 + '%'
```
With `@p1 = 'Bo'`

Here that is just a simple case, which do not result in a function call on sql side and should not have much impact. But for other extension cases, having them ending in SQL while they could have been evaluated in memory instead can be more detrimental.

fredericDelaporte:
> Ok, you've made me wake up from the bed. It's bad.

Sorry, maybe mute your phone, at least for mails.

The newid example prove nothings: it is not the general case. Newid is typically a non-deterministic method, the kind which requires being evaluated at each row, so the one deserving being always "dbonly".

The general case is more like the string.Concat case I have showcased in my previous comment. There, dbonly changes things.

hazzik:
Just tested: instance methods will never be checked with `IsEvaluatableMethodCall`. They always treated as evaluatable.

hazzik:
Replace newid method with any other method. The behavior will be the same. "DBOnly=false" = no attribute.

fredericDelaporte:
I have told you I am going to add test for demonstrating that. Just have a bit of patience. No need to close. I will not merge that during your sleep.

hazzik:
I just demonstrated my point. Sorry, but there is nothing left to demonstrate.

fredericDelaporte:
I am sure I will prove it otherwise. I am just not on that subject yet.

fredericDelaporte:
And this pr is not only about dbonly=false. There is more into it.

hazzik:
>public IList<Person> Search(nameStart)

This is an imaginary example, which will never be true. string.Concat is treated as a special case by the relinq itself.

>And this pr is not only about dbonly=false. There is more into it.

It just wrong implementation because of wrong unproven assumption.

I can not just imagine the case where DBOnly=false behaves differently to not having attribute at all. Making the defaut DBOnly to false for LinqExtensionMethodAttribute is a breaking change, but not "preserving the current behaviour" as showed by the example above.

fredericDelaporte:
It seems to me your are not trying to understand what I have written, this discussion is going nowhere currently. Better go on a bit later, mind you?

hazzik:
For the methods accepting the current IQueryable (as an instance or as an argument), the `IsEvaluatableMethodCall` will not be called as they deemed to be not evaluatable. So, the DbOnly= argument would not play any role. 

So, changing the above example to following:

		[LinqExtensionMethod("NEWID", DBOnly = true)]
		public static Guid NewID1_1(Entity entity)
		{
			throw new InvalidOperationException("To be translated to SQL only");
		}

		[LinqExtensionMethod("NEWID", DBOnly = false)]
		public static Guid NewID1_2(Entity entity)
		{
			throw new InvalidOperationException("To be translated to SQL only");
		}

		public static Guid NewID1_3(Entity entity)
		{
			throw new InvalidOperationException("To be translated to SQL only");
		}

		[LinqExtensionMethod("NEWID", DBOnly = true)]
		public static Guid NewID2_1(Entity entity)
		{
			return Guid.NewGuid();
		}

		[LinqExtensionMethod("NEWID", DBOnly = false)]
		public static Guid NewID2_2(Entity entity)
		{
			return Guid.NewGuid();
		}

		public static Guid NewID2_3(Entity entity)
		{
			return Guid.NewGuid();
		}

Brings us to the following result:

- NewID1_1 will be evaluated into the query as a NEWID function
- NewID1_2will be evaluated into the query as a NEWID function **(sic!)**
- NewID1_3 will throw "System.NotSupportedException: System.Guid NewID1_3(NHibernate.Test.NHSpecificTest.NH3386.Entity)" becuase there is no generator
- NewID2_1 will be evaluated into the query as a NEWID function
- NewID2_2 will be evaluated into the query as a NEWID function **(sic!)**
- NewID2_3 will throw "System.NotSupportedException: System.Guid NewID2_3(NHibernate.Test.NHSpecificTest.NH3386.Entity)" becuase there is no generator

No calls to IsEvaluatableMethodCall has been made.

hazzik:
It seems that I finally understood what you want to achieve, but I am still opposing this change. There shall be no functions which we might evaluate by DB or not based on a moon phase.

fredericDelaporte:
> There shall be no functions which we might evaluate by DB or not based on a moon phase.

If you consider the behavior "evaluate db side things depending on queried data, in-memory the remaining" to be like "moon phase", this is the way the partial evaluating processor works. So we should indeed completely remove that partial evaluating processor for having the whole expression evaluated db side. But that will likely be a major breaking change for the user code base.

src/NHibernate.Test/Linq/ExtensionMethods.cs
```diff
@@ -2,18 +2,24 @@
 
 namespace NHibernate.Test.Linq
 {
-    static class ExtensionMethods
-    {
-        [LinqExtensionMethod("Replace")]
-        public static string ReplaceExtension(this string subject, string search, string replaceWith)
-        {
-            return null;
```
fredericDelaporte: I have replaced `null` by `subject?.Replace(search, replaceWith) + " (done in-memory)";` for allowing better asserts. (There were none, the original test was not validating the query was not throwing.)

src/NHibernate.Test/Linq/ExtensionMethods.cs
```diff
@@ -2,18 +2,24 @@
...
-        }
-    }
+		[LinqExtensionMethod]
+		public static string Replace(this string subject, string search, string replaceWith)
+		{
+			return subject?.Replace(search, replaceWith) + " (done in-memory)";
+		}
+
+		[LinqExtensionMethod("Replace", DBOnly = true)]
+		public static string ReplaceInDb(this string subject, string search, string replaceWith)
```
fredericDelaporte: Added for demonstrating the behavior change between false and true.

src/NHibernate.Test/Linq/FunctionTests.cs
```diff
@@ -114,17 +115,43 @@ public void LeftFunction()
...
+			var results = query.ToList();
+			var s = ObjectDumper.Write(results);
+
+			foreach (var result in results)
+			{
+				var expectedDbResult = Regex.Replace(result.Before, "An", "Zan", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+				Assert.That(result.AfterMethod, Is.EqualTo(expectedDbResult), "Wrong AfterMethod value");
+				Assert.That(result.AfterExtension, Is.EqualTo(expectedDbResult), "Wrong AfterExtension value");
+				Assert.That(result.AfterExtension2, Is.EqualTo(expectedDbResult), "Wrong AfterExtension2 value");
+				Assert.That(result.AfterExtension3, Is.EqualTo(expectedDbResult), "Wrong AfterExtension3 value");
```
fredericDelaporte: When called on an entity property, they are all evaluated in Db.

src/NHibernate.Test/Linq/FunctionTests.cs
```diff
@@ -114,17 +115,43 @@ public void LeftFunction()
...
+
+				var expectedDbResultConst = Regex.Replace(result.BeforeConst, "An", "Zan", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+				var expectedInMemoryResultConst = result.BeforeConst.Replace("An", "Zan");
+				var expectedInMemoryExtensionResultConst = ExtensionMethods.Replace(result.BeforeConst, "An", "Zan");
+				Assert.That(result.AfterMethodConst, Is.EqualTo(expectedInMemoryResultConst), "Wrong AfterMethodConst value");
```
fredericDelaporte: When called on constant expressions, the actual `string.Replace` is always evaluated in-memory. This has always be the case with the NHibernate Linq provider.

src/NHibernate.Test/Linq/FunctionTests.cs
```diff
@@ -114,17 +115,43 @@ public void LeftFunction()
...
+				Assert.That(result.AfterExtensionConst, Is.EqualTo(expectedInMemoryExtensionResultConst), "Wrong AfterExtensionConst value");
+				Assert.That(result.AfterExtension2Const, Is.EqualTo(expectedInMemoryExtensionResultConst), "Wrong AfterExtension2Const value");
```
fredericDelaporte: When called on constant expressions, extension methods without `dbOnly = true` are called in-memory. This has always be the case with the NHibernate Linq provider.

(Since NH-3961 committed for v5, if the in-memory evaluation fails with an exception, the call stays in the expression instead of throwing. Even if the failure was not expected by the method developer.)

src/NHibernate.Test/Linq/FunctionTests.cs
```diff
@@ -114,17 +115,43 @@ public void LeftFunction()
...
+				Assert.That(result.AfterExtension3Const, Is.EqualTo(expectedDbResultConst), "Wrong AfterExtension3Const value");
```
fredericDelaporte: Even when called on constant expressions, extension methods with `dbOnly = true` are no more called in-memory. This is a new behavior. They are always translated to SQL (at least, translation attempted).

hazzik:
>So we should indeed completely remove that partial evaluating processor for having the whole expression evaluated db side.

Good luck.

hazzik:
>DBOnly = false is current NHibernate v4 behavior.

It is not. And this is why you have added "DbOnly=true" to some existing methods. Meaning this is a breaking change. "DBOnly = false" nor "DbOnly=true" represent current NH4 behavior.

src/NHibernate.Test/NHSpecificTest/NH3386/Fixture.cs
```diff
@@ -60,7 +60,7 @@ public void ShouldSupportNonRuntimeExtensionWithoutEntityReference()
 
 	public static class SqlServerFunction
 	{
-		[LinqExtensionMethod]
+		[LinqExtensionMethod(DBOnly = true)]
```
hazzik: Requiring users to add additional flag for existing methods is a breaking change.

fredericDelaporte:
> It is not. And this is why you have added "DbOnly=true" to some existing methods. Meaning this is a breaking change. "DBOnly = false" nor "DbOnly=true" represent current NH4 behavior.

Please apply 8fb2bfa10fbbdd9870b4dd362cc019240f5c4781 (test changes only, no dbonly) to a 4.x branch and see by yourself. DBOnly = false is actually current NH4 behavior.

And well, I have done it [here](https://github.com/fredericDelaporte/nhibernate-core/tree/LinqExtensionBehavior). Only failing assert is the one needing dbonly true. This demonstrates current NH4 is working the "dbOnly false" way.

I still have to add "method throwing exception cases". For this cases, I am convinced too that the "dbonly false" accurately represent NH4 behavior too.

src/NHibernate.Test/NHSpecificTest/NH3386/Fixture.cs
fredericDelaporte: No requirement here. Their working code will still work without it. Their failing code could start to work if they add it.

hazzik: This code was working and stopped working without it

hazzik:
> I still have to add "method throwing exception cases". For those cases, I am convinced too that the "dbonly false" accurately represent NH4 behavior too.
--
I don't understand what are you trying to say here.

src/NHibernate.Test/NHSpecificTest/NH3386/Fixture.cs
fredericDelaporte: It was not working with NH4. Third time I answer that, to the third time you do this objection. This code had start working with a v5 commit having accidentally sort of fixed it. With drawbacks.

fredericDelaporte:
> I don't understand what are you trying to say here.

I am referring to a case currently not covered by tests: extension methods which implementation does throw, meaning such extension methods could only be evaluated db side. Having test demonstrating NHibernate behavior with them, be it v4, current v5, or this PR, would be better. And I will add them.

src/NHibernate.Test/NHSpecificTest/NH3386/Fixture.cs
fredericDelaporte: I am going to cherry-pick that newid test into the NH4 branch I have just made to showcase this.

hazzik:
Ok, I've finally understood what is the problem here. The problem is that 5.0 is long overdue, and I did not realize that NEWID wasn't possible in 4.0. My apologies. I should have stayed in the bed yesterday:)

hazzik:
It's all fine then, except the naming. Having the ability to write `[DbOnly(DbOnly=false)]` is weird.

fredericDelaporte:
Thanks!!!

I was really having a hard time with this discussion. Not to the point of not sleeping but maybe not that far!

I have pushed the newid test in my [nh4 branch](https://github.com/fredericDelaporte/nhibernate-core/tree/LinqExtensionBehavior) if you wish to ascertain all that just in case. Avoiding undue regression is always a valid concern anyway.

About `[DbOnly(DbOnly=false)]`, yes that is non-sens, and better avoid non-sens possibilities if doable without drawbacks. I have the following solutions:

- Have `LinqExtensionMethodAttribute` no more inheriting db only, remove the property from `DBOnlyAttribute`, have `[DBOnly]` only being db only, and required to be added for extension method on top of linq extension attribute if that method is to be "dbonly". A bit unfortunate to have to put two attributes for simple function extension method when they must be db-only, as you were objecting in my NH-3944 own pr.
- Same as previous but adding back a `DBOnly` property directly on `LinqExtensionMethodAttribute`, and checking both attributes in the partial evaluation filter. (The db only attribute stays required for things like `MappedAs`, in order to support them without swallowing exceptions.) I am not fond of having to check two things for the same purpose.
- Change that `DBOnly` property as something like `internal protected DBOnly`, and put a `public new DBOnly` in `LinqExtensionMethodAttribute` mapping back to `base.DBOnly`. A bit hackish I find. But maybe still better than all other possibilities, including staying in this PR current state.

By the way I still intend to add more tests here, for the cases which were told about in the discussion.

hazzik:
Here are some thoughts which are still valid despite:

The contract (however not bound by the code) was that methods marked as [LinqExtensionMethod] should not have an actual implementation and should throw `InvalidOperationException`, `NotSupportedException`, `NotImplementedException`, or any other exception by the preference of the user. The fact that these methods could have an implementation evaluable by CLR is an undocumented feature. The users who leverage this undocumented feature I believe is the minority.

We shall treat `[LinqExtensionMethod]` attribute as always db-only. We can declare this as a possible breaking change if you wish. 

Currently, EF Core treats some non-deterministic methods as not evaluable by CLR and always substitute them with the DB equivalent if they happen to be in the query. These methods are Random.Next(), Guid.NewGuid(), DateTime.Now, and DateTime.NowUtc. We shall do the same. This will make queries idempotent to the query provider. 

MappedAs (and any methods of such nature in the future) shall be treated as a special method. As, in fact, it is not DBOnly per se.

hazzik:
> and put a public new DBOnly

Just a constructor argument is enough.

fredericDelaporte:
> The contract (however not bound by the code) was that methods marked as [LinqExtensionMethod] should not have an actual implementation and should throw InvalidOperationException, NotSupportedException, NotImplementedException, or any other exception by the preference of the user. The fact that these methods could have an implementation evaluable by CLR is an undocumented feature. The users who leverage this undocumented feature I believe is the minority.

I am not aware of this contract being documented anywhere. It would be limiting to mandate throwing for users having the need of a custom method both for runtime usage and db-usage, forcing them to have two distinct methods for the same job, depending on whether they want to use it in runtime or in SQL.
But yes it is likely those users are a minority, so why not having "dbonly true" a default, but still allowing reverting back to old behavior for who would wish it. I would then rename the option in the constructor to something like `evaluate` or `canBeEvaluated`, ...

For Random.Next(), Guid.NewGuid(), DateTime.Now, and DateTime.NowUtc, agreed, possible breaking change too, but better. Maybe should it be done for `DateTimeOffset` too. But for me it needs to be done with another Jira. The attribute can not be applied to them, another way of filtering them is required. And they are not currently registered for translation, so there is more work to do than just filtering them out of partial evaluation. (They will also require a matching standard hql function: not sure we have that for all of them currently.)

`MappedAs` is not actually an in-db thing, granted. But it cannot be evaluated either. (And maybe could it be extended later for allowing casting an entity property to another type in-db.) The `DBOnly` can still be useful for custom extensions not simply mapping to a function and requiring a registry. That will be the sole way to get them filtered out of partial evaluation. (Unless we change the registration for allowing specifying that the registered method needs to be filtered out of partial evaluation. But we are not there yet, it have to be done in another Jira I think.)

hazzik:
I've updated my comment, it already has the answers/suggestions for the first part. 

`evaluate`/`canBeEvaluated` does not mean anything for the user who is not familiar with the internals of NHibernate/Re-Linq. But DbOnly does not make sense for an uneducated user as well. Something like "AlwaysTranslateToSql". Or, even better (in my sense at least): make this property an enum. 

As for non-deterministic methods, obviously, this requires a separate JIRA.

fredericDelaporte:
So I have written [NH-4048](https://nhibernate.jira.com/browse/NH-4048) and redesigned this PR.

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
```diff
@@ -31,4 +47,58 @@ public LinqExtensionMethodAttribute(string name) : this()
...
+	}
+
+	/// <summary>
+	/// Possible method call behaviors when the linq to NHibernate provider partially evaluates
+	/// expressions before parsing them.
+	/// </summary>
+	public enum LinqExtensionEvaluation
+	{
+		/// <summary>
+		/// The method call will no be evaluated even if it arguments do not depend on the queried data.
```
hazzik: "no" => "not"

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
```diff
@@ -31,4 +47,58 @@ public LinqExtensionMethodAttribute(string name) : this()
...
+		/// </summary>
+		NoEvaluation,
+		/// <summary>
+		/// If the method call do not depend on queried data, the method call will be evaluated and replaced
+		/// by the resulting constant expression in the expression tree. This was the default behavior
+		/// prior to NHibernate v5.
+		/// </summary>
```
hazzik: I think this comment shall be reworded. First, need to remove mentions of the previous behavior, as it does not matter. Second, if method throws - it will throw.

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
```diff
@@ -31,4 +47,58 @@ public LinqExtensionMethodAttribute(string name) : this()
...
 	}
+
+	/// <summary>
+	/// Can flag a method as not being callable by the runtime, when used in Linq queries.
+	/// If the method is supported by the linq-to-nhibernate provider, it will always be converted
+	/// to the corresponding SQL statement.
+	/// Otherwise the linq-to-nhibernate provider evaluates method calls when they do not depend on
+	/// the queried data.
+	/// </summary>
+	public class NotEvaluatableAttribute : LinqExtensionMethodAttributeBase
```
hazzik: https://english.stackexchange.com/questions/304948/evaluable-vs-evaluatable

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -21,17 +22,22 @@ protected override Expression VisitConstant(ConstantExpression expression)
...
 	{
+		public override bool IsEvaluatableMethodCall(MethodCallExpression node)
+		{
+			if (node == null)
+				throw new ArgumentNullException(nameof(node));
+
+			return node.Method
+				.GetCustomAttributes(typeof(LinqExtensionMethodAttributeBase), false)
+				.Cast<LinqExtensionMethodAttributeBase>()
+				.All(a => a.Evaluation == LinqExtensionEvaluation.AllowPartialEvaluation);
```
hazzik: I think it shall be "Any", not all.

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
```diff
@@ -2,17 +2,103 @@
...
+
+	/// <summary>
+	/// Base class for Linq extension attributes.
+	/// </summary>
+	public abstract class LinqExtensionMethodAttributeBase : Attribute
+	{
+		/// <summary>
+		/// Should the method call be partially evaluated?
+		/// </summary>
+		public LinqExtensionEvaluation Evaluation { get; }
```
hazzik: Talking about evaluation here is to disclose the internals of the implementation, which do not matter in this context. This needs to be close to the context. In this context if something is not evaluatable it will be forcible translated into a method call; if it is partially evaluatable, then it will be calculated before the call. I don't know how to express this as a name of the property.

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
```diff
@@ -2,17 +2,103 @@
...
+		/// <see cref="LinqExtensionEvaluation.NoEvaluation"/>.</param>
+		public LinqExtensionMethodAttribute(string name, LinqExtensionEvaluation evaluation)
+			: base(evaluation)
 		{
 			Name = name;
 		}
 
-		public string Name { get; private set; }
+		/// <summary>
+		/// The name of the SQL to call.
```
hazzik: SQL _function_ to call

fredericDelaporte:
You have trapped me! Putting `Any` causes all methods without any attribute to be considered non-evaluable, causing a bunch of failures. (I have launched the test suit locally, but forgot about it and pushed without checking.)

src/NHibernate/Linq/LinqExtensionMethodAttribute.cs
```diff
@@ -2,17 +2,106 @@
...
+	/// </summary>
+	public enum LinqExtensionPreEvaluation
+	{
+		/// <summary>
+		/// The method call will not be evaluated even if its arguments do not depend on queried data.
+		/// It will always be translated to the corresponding SQL statement.
+		/// </summary>
+		NoEvaluation,
+		/// <summary>
+		/// If the method call do not depend on queried data, the method call will be evaluated and replaced
```
hazzik: does not

hazzik:
>You have trapped me! 

I did not. I just suggested that this shall be Any.

hazzik:
Removed arguing

hazzik:
Can we get a second opinion on the naming of the property?

fredericDelaporte:
> I did not. I just suggested that this shall be Any.

I was just kidding of course.

fredericDelaporte:
To make things a bit easier, the current state of this PR is:

- Causes methods flagged with `[LinqExtensionMethod]` to not be partially evaluated anymore by default. This allows such methods to always be left in the expression tree for being translated to HQL, even if they call do not depend on any queried data. This allows to use things like a custom `NewID()` method in Linq and have it translated into SQL, which was not previously possible.
- `LinqExtensionMethodAttribute` constructors take an additional `LinqExtensionPreEvaluation preEvaluation` parameter, allowing to specify `AllowPreEvaluation`, default being `NoEvaluation`. This allows who would want it to get back to previous behavior, having the method evaluated in-memory when called on constants, not on queried data.
- A new `NoPreEvaluationAttribute` is added, currently used for the `MappedAs` extension, allowing to avoid it to be evaluated. (Instead of catching its evaluation failure.) It is public, because it could be useful for custom Linq extension too.
- The partial evaluation processor will no more swallow evaluation exceptions silently.

fredericDelaporte:
Rebased.
We are one month later, so it seems @gliljas has not much time available for NHibernate. We may have to decide without his view.

fredericDelaporte:
👍

PR #663, opened 2017-07-14
NH-4052 - Collect schema validation exceptions

DMDTT:
[NH-4052](https://nhibernate.jira.com/browse/NH-4052) - Collect schema validation exceptions

+ add list of exceptions to HibernateException
+ create HibernateException on validation fail and add to the exception list
+ return list of exceptions to throw one common HibernateException to be able to evaluate it on application side

### DIFF
--- a/src/NHibernate.Test/Async/Tools/hbm2ddl/SchemaValidator/SchemaValidateFixture.cs
+++ b/src/NHibernate.Test/Async/Tools/hbm2ddl/SchemaValidator/SchemaValidateFixture.cs
@@ -19,16 +19,44 @@ namespace NHibernate.Test.Tools.hbm2ddl.SchemaValidator
 	[TestFixture]
 	public class SchemaValidateFixtureAsync
 	{
-		[Test]
-		public async Task ShouldVerifySameTableAsync()
+		private const string _resourcesPrefix = "NHibernate.Test.Tools.hbm2ddl.SchemaValidator.";
+		private const string _version1Resource = _resourcesPrefix + "1_Version.hbm.xml";
+		private const string _version2Resource = _resourcesPrefix + "2_Version.hbm.xml";
+		private const string _version3Resource = _resourcesPrefix + "3_Version.hbm.xml";
+		private Configuration _configuration1;
+		private SchemaExport _export1;
+
+		[OneTimeSetUp]
+		public void OneTimeSetUp()
 		{
-			const string resource = "NHibernate.Test.Tools.hbm2ddl.SchemaValidator.1_Version.hbm.xml";
-			var cfg = BuildConfiguration(resource);
+			_configuration1 = BuildConfiguration(_version1Resource);
+		}
 
-			await (new SchemaExport(cfg).ExecuteAsync(true, true, false));
+		[SetUp]
+		public void SetUp()
+		{
+			_export1 = new SchemaExport(_configuration1);
+			_export1.Create(true, true);
+		}
 
-			var validator = new Tool.hbm2ddl.SchemaValidator((cfg));
-			await (validator.ValidateAsync());
+		[TearDown]
+		public void TearDown()
+		{
+			_export1.Drop(true, true);
+		}
+
+		[Test]
+		public Task ShouldVerifySameTableAsync()
+		{
+			try
+			{
+				var validator = new Tool.hbm2ddl.SchemaValidator((_configuration1));
+				return validator.ValidateAsync();
+			}
+			catch (System.Exception ex)
+			{
+				return Task.FromException<object>(ex);
+			}
 		}
 
 		[Test, SetCulture("tr-TR"), SetUICulture("tr-TR")]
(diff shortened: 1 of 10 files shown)


src/NHibernate/Mapping/Table.cs
```diff
@@ -972,30 +972,34 @@ public override string ToString()
...
 			foreach (Column column in iter)
 			{
 				IColumnMetadata columnInfo = tableInfo.GetColumnMetadata(column.Name);
 
 				if (columnInfo == null)
 				{
-					throw new HibernateException(string.Format("Missing column: {0} in {1}", column.Name,
-															   dialect.Qualify(tableInfo.Catalog, tableInfo.Schema, tableInfo.Name)));
+					exceptions.Add(new HibernateException(string.Format("Missing column: {0} in {1}", column.Name,
+															   dialect.Qualify(tableInfo.Catalog, tableInfo.Schema, tableInfo.Name))));
```
fredericDelaporte: Here lies a bug due to the change. A throw stops code execution, protecting in this case the code from reaching a null ref condition. You should make sure that everywhere you have replaced a throw by just an add in a list, the code following that will support the state for which an exception was supposed to be thrown or will be appropriately bypassed.

src/NHibernate/HibernateException.cs
```diff
@@ -12,6 +13,7 @@ namespace NHibernate
 	[Serializable]
 	public class HibernateException : Exception
 	{
+		public List<HibernateException> Exceptions { get; }
```
fredericDelaporte: Better follow usual conventions, like the one of the standard [`AggregateException`](https://docs.microsoft.com/en-us/dotnet/api/system.aggregateexception): name this `InnerExceptions`, and expose this as a `ReadOnlyCollection<Exception>`.

src/NHibernate/HibernateException.cs
```diff
@@ -12,6 +13,7 @@ namespace NHibernate
 	[Serializable]
 	public class HibernateException : Exception
```
fredericDelaporte: I am not for adding this to the hibernate exception base class. If we add that, it should be on a specialized class, like adding `class HibernateAggregateException : HibernateException`.

DMDTT:
@fredericDelaporte Thanks for your review.
Keeping the validation errors as a readonlycollection seems a more apropriate way then keeping one string.

I updated the code

src/NHibernate.Test/App.config
```diff
@@ -7,7 +7,7 @@
 	</configSections>
 
 	<connectionStrings>
-		<add name="TestConnectionString" connectionString="Server=localhost\sqlexpress;Database=nhibernate;Integrated Security=SSPI" />
+		<add name="TestConnectionString" connectionString="Server=localhost;Database=nhibernate;Integrated Security=SSPI" />
```
fredericDelaporte: Please revert. You should not need to change that. Use ShowBuildMenu and setup your test configuration with it.

src/NHibernate/HibernateException.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
```
fredericDelaporte: Change now unneeded, please revert too.

src/NHibernate.TestDatabaseSetup/App.config
```diff
@@ -9,7 +9,7 @@
     <session-factory name="NHibernate.TestDatabaseSetup">
 			<property name="connection.driver_class">NHibernate.Driver.SqlClientDriver</property>
 			<property name="connection.connection_string">
-        Server=.\SQLExpress;initial catalog=master;Integrated Security=SSPI
+        Server=localhost;initial catalog=master;Integrated Security=SSPI
```
fredericDelaporte: Please revert too.

src/NHibernate/SchemaValidationException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace NHibernate
+{
+	public class SchemaValidationException : HibernateException
+	{
+		public ReadOnlyCollection<string> ValidationErrors { get; }
+
+		public SchemaValidationException(string msg, IList<string> validationErrors) : base(msg)
```
fredericDelaporte: `IEnumerable<string>` should be enough.

src/NHibernate.Test/NHSpecificTest/NH1274ExportExclude/NH1274ExportExcludeFixture.cs
```diff
@@ -7,6 +7,7 @@
 using NHibernate.Criterion;
 using NHibernate.Engine;
 using NHibernate.Tool.hbm2ddl;
+using NHibernate.Util;
```
fredericDelaporte: We wish to remove NHibernate collection custom extensions, so please do not add more usages of them.

src/NHibernate.Test/NHSpecificTest/NH1274ExportExclude/NH1274ExportExcludeFixture.cs
```diff
@@ -84,9 +85,10 @@ public void SchemaExport_Validate_CausesValidateException()
 			{
 				validator.Validate();
 			}
-			catch (HibernateException he)
+			catch (SchemaValidationException e)
 			{
-				Assert.IsTrue(he.Message.Contains("Home_Validate"));
+				Assert.IsTrue(e.Message.Contains("Schema validation failed: see list of validation errors"));
+				Assert.IsTrue(e.ValidationErrors.Contains("Missing table: Home_Validate"));
```
fredericDelaporte: Better use Fluent syntax: `Assert.That`. NUnit does no more add features to the others, and may obsolete them.
By the way, just having "expected true but was false" as an error message while there is more than one test on `true` is not practical : better at least add an explicit error message.

Here the whole thing could be simplified too:

```C#
Assert.That(() => validator.Validate(),
    Throws.TypeOf<SchemaValidationException>()
        .And.Property("ValidationErrors").Contain("Missing table: Home_Validate"));
```
(Checking the type and the error list without checking the exception message seems enough to me, the type is already sufficiently specialized.)

src/NHibernate.Test/Tools/hbm2ddl/SchemaValidator/SchemaValidateFixture.cs
```diff
@@ -60,9 +61,10 @@ public void ShouldNotVerifyModifiedTable()
 			{
 				validatorV2.Validate();
 			}
-			catch (HibernateException e)
+			catch (SchemaValidationException e)
 			{
-				Assert.That(e.Message, Does.StartWith("Missing column: Name"));
+				Assert.That(e.Message, Does.StartWith("Schema validation failed: see list of validation errors"));
+				Assert.That(e.ValidationErrors[0], Does.StartWith("Missing column: Name"));
```
fredericDelaporte: Better but could be simplified as previous assert.

src/NHibernate/SchemaValidationException.cs
hazzik: ReadOnlyCollection (next line) requires `IList<T>`

hazzik:
Mentioned issues resolved. Squashed, rebased and Async tests regenerated.

src/NHibernate.Test/Async/Tools/hbm2ddl/SchemaValidator/SchemaValidateFixture.cs
```diff
@@ -67,14 +68,12 @@ public class SchemaValidateFixtureAsync
...
-			catch (HibernateException e)
-			{
-				Assert.That(e.Message, Does.StartWith("Missing column: Name"));
-			}
+
+			Assert.That(
+				() => validatorV2.ValidateAsync(),
+				Throws.TypeOf<SchemaValidationException>()
+				      .And.Message.EqualTo("Schema validation failed: see list of validation errors")
+				      .And.Property("ValidationErrors").Contains("Missing column: Name in nhibernate.dbo.Version"));
```
fredericDelaporte: As showcased by failing tests, this column naming is specific to SQL Server and causes the test to fail for other db.

hazzik:
I'm not doing it right now. So yes, if you wish

DMDTT:
@fredericDelaporte @hazzik 
Thanks for your work guys, 
I didn't manage it to work on it in the last month, sry for that.

But nethertheless the work of yours is great. Thanks again.

PR #664, opened 2017-07-20
Better logging on StaleObjectStateException and Interceptor.OnSave with new Version

adoris:
Better logging on StaleObjectStateException : NH-4055
Interceptor.OnSave with new Version: NH-4056


adoris:
StaleObjectStateException need extended

PR #665, opened 2017-07-24
NH-4057 - fix tests for MySQL

fredericDelaporte:
[NH-4057](https://nhibernate.jira.com/browse/NH-4057) - fix tests for MySQL


src/NHibernate.Test/NHSpecificTest/NH2302/Fixture.cs
```diff
@@ -152,6 +153,8 @@ public void BlobWithLength()
 				[Test]
 				public void BlobWithoutLength()
 				{
+					if (Dialect is MySQLDialect)
```
hazzik: We need to fix it for all dialects actually. Currently we have a huuuuuuuuge mess in how we handle string types.

fredericDelaporte: It seems this one test fails only for MySql. There is another test here failing for another db, but for all other db it works. So maybe it is not that far of being fixed for all.

PR #666, opened 2017-07-25
NH-4058 - Fix Oracle (12c) Managed 32 failing tests.

fredericDelaporte:
[NH-4058](https://nhibernate.jira.com/browse/NH-4058) - Fix Oracle (12c) Managed 32 failing tests.


src/NHibernate.Test/SecondLevelCacheTest/QueryCacheFixture.cs
```diff
@@ -301,11 +301,13 @@ public void SimpleProjections()
...
 
-		public class CustomTransformer: IResultTransformer
+		public class CustomTransformer : IResultTransformer
 		{
 			public object TransformTuple(object[] tuple, string[] aliases)
 			{
-				return new AnotherItem {Name = tuple[0].ToString(), Description = tuple[1].ToString()};
+				// Some db change the empty string to null, causing .ToString() to blow.
+				// https://stackoverflow.com/a/203536/1178314
+				return new AnotherItem { Name = tuple[0]?.ToString(), Description = tuple[1]?.ToString() };
```
fredericDelaporte: That is an old Oracle behavior. '' => null.

src/NHibernate.Test/TestCase.cs
```diff
@@ -292,7 +292,12 @@ protected virtual void DropSchema()
 
 		protected virtual DebugSessionFactory BuildSessionFactory()
 		{
-			return new DebugSessionFactory(cfg.BuildSessionFactory());
+			var factory = new DebugSessionFactory(cfg.BuildSessionFactory());
+			// Avoid ORA-12704 in tests, due to NHibernate defaulting String to NVarchar2 when creating tables while
+			// ODP.Net defaults it to Varchar2 (which does support Unicode too, depending on server) for parameters.
+			if (factory.ConnectionProvider?.Driver is OracleManagedDataClientDriver oracleManagedDriver)
+				oracleManagedDriver.UseNVarchar2ForStringParameter = true;
```
fredericDelaporte: I have added that setting on the oracle managed driver. There are ore detail on the parameter comments. The culprit is what I see as a mismatch between NHibernate Oracle Dialect, matching String to `NVarchar2`, while ODP matches it to `Varchar2` instead (which can be Unicode too, unlike with SQL Server). Oracle refuses to mix those string types for a number of operations, causing issues in some queries, because NHibernate creates columns as `NVarchar2` while string parameters are transmitted as `Varchar2`.

Maybe we should change that in Oracle Dialect for re-aligning on ODP defaults. But that would then requires another Jira I think.

fredericDelaporte:
I have fixed only the failing one on TeamCity. On my local setup, with Oracle 11g XE, more are failing.

src/NHibernate.Test/Linq/MiscellaneousTextFixture.cs
```diff
@@ -86,16 +86,17 @@ public void ReferenceToOuter()
...
-            IQueryable<Product> page2 = q.Skip(5).Take(5);
-            IQueryable<Product> page3 = q.Skip(10).Take(5);
-            IQueryable<Product> page4 = q.Skip(15).Take(5);
+			var q =
+				from p in db.Products
+				where p.ProductId > 1
+				orderby p.ProductId
+				select p;
+
+			// ToList required otherwise the First call alters the paging and test something else than paging three pages.
```
hazzik: I don't quite agree with this change.

fredericDelaporte: `Skip(5).Take(5).First()` does not make sens, it should be `Skip(5).First()`. Do you really think the test writer intended the previous code?

For me, this test is testing something else that what it was intended for. Something that bugs with Oracle by the way. But that bug is a bit corner case, the `Take` has no purpose anymore when calling `First` on it.

hazzik: Anyway, I think we shall not change the test. This is a valid linq expression, even if it does not make much sense for you. We shall either skip this test for oracle unchanged or/and create a JIRA issue for a specific case (with a minor/trivial priority).

src/NHibernate.Test/Legacy/SQLFunctionsTest.cs
```diff
@@ -76,10 +76,13 @@ public void DialectSQLFunctions()
 				rset = s.CreateQuery("select abs(round(s.Pay)) from s in class Simple").List();
 				Assert.AreEqual(46f, rset[0], "abs(round(-45.8)) result was incorrect");
 
+				/* As of NH-3893, left and right functions are broken. Seemed confused with join keyword, and not
```
hazzik: Should not comment a test. Need to exclude it with if statement on some condition

hazzik: Also, JIRA is required

fredericDelaporte: That is [NH-3893](https://nhibernate.jira.com/browse/NH-3893).

There are no other conditions than `if (false)` for this part of the test. It is broken, whatever the database. But since it was nested in an `if (Dialect is Oracle8iDialect)`, it was failing only for Oracle.

hazzik: Ok, can we then remove this lines or move them outside of `if oracle`?

src/NHibernate/Driver/OracleManagedDataClientDriver.cs
```diff
@@ -47,8 +48,31 @@ public OracleManagedDataClientDriver()
...
+		/// <para>
+		/// So the default NHibernate choice in Oracle dialects to map <see cref="DbType.String"/> to <c>nvarchar2</c> is a bit
+		/// unfortunate, because this is not the default mapping for Oracle. For tests, since the database is created by NHibernate,
+		/// string columns get typed <c>nvarchar2</c>, but parameters are still typed <c>varchar2</c>, causing an
+		/// <c>ORA-12704: character set mismatch</c> for some queries. This being not even reported, I guess most NHibernate Oracle
+		/// users handles their tables themselves and types them with <c>varchar2</c>, avoiding the trouble. So changing the behavior
+		/// for parameters type would be a major breaking change.
+		/// </para>
+		/// </remarks>
+		public bool UseNVarchar2ForStringParameter { get; set; }
```
hazzik: This is the wrong assumption, I belive.

We have 4 different mappings of strings in ADO.NET:
1. AnsiString - VARCHAR(n) (VARCHAR2(n) in the case of Oracle) - the single byte string collation dependent.
2. AnsiStringFixedLength - CHAR(n) - fixed length string.
3. String - NVARCHAR(n) (NVARCHAR2(n) in the case of Oracle) - always Unicode strings (UTF16, I think)
4. StringFixedLength - NCHAR(n) - fixed length Unicode string (UTF16, I think).

We shall really fix the provider to obey this rule.

fredericDelaporte: I am convinced that NHibernate is wrong assuming Oracle Varchar2 is meant for single byte character encoding. See this [Oracle](https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch6unicode.htm#CACHCAHF) page.

They have not implemented Unicode as a new string type, but simply as a new encoding, usable with any already existing string and character types.

The default recommended way of using Unicode with Oracle is to use a varchar2 with an adequate database encoding.

> You can create a Unicode database that enables you to store UTF-8 encoded characters as SQL `CHAR` datatypes (`CHAR`, `VARCHAR2`, `CLOB`, and `LONG`).

Using the `N` prefixed types are second option, notably meant for people wanting to transition incrementally to Unicode instead of globally switching to it.

> If you prefer to implement Unicode support incrementally or if you need to support multilingual data only in certain columns, then you can store Unicode data in either the UTF-16 or UTF-8 encoding form in SQL `NCHAR` datatypes (`NCHAR`, `NVARCHAR2`, and `NCLOB`). The SQL `NCHAR` datatypes are called Unicode datatypes because they are used only for storing Unicode data.

This effectively means Oracle has two Unicode support models. And NHibernate implements the one which looks like being the less used. If we really want to fix this mess, we need to give an option for letting the user tell which model he uses. This could be that `UseNVarchar2ForStringParameter` option, but for completeness, it needs to be taken into account by the dialect too.

The first recommended one, by the way, does not support anymore single byte encoding at all.

hazzik: Ok. Can we then separate the wheat from the chaff? I mean skip tests which are failing and create a bug to be fixed later? As I'm not convinced with the solution of sneakily fixing such bugs

PR #667, opened 2017-08-01
NH-4062 - Oracle Unicode support dual model.

fredericDelaporte:
[NH-4062](https://nhibernate.jira.com/browse/NH-4062) - Properly handle Oracle Unicode support dual model

~~And reverting [NH-3620](https://nhibernate.jira.com/browse/NH-3620): this was a workaround for a now obsolete bug in ODP.Net managed driver. Its test still pass without it now. It should probably have been solved as "external issue" with instruction on how to extend the driver for working around the bug till Oracle fixes it.~~ Removing this workaround causes another test to fail: `BinaryBlobTypeFixture.ReadWriteZeroLen`. Without it, the blob come back as `null`.

Managed and un-managed drivers require the same code, but it was duplicated. Now it is refactored in a single base class.

### DIFF
--- a/src/NHibernate/Cfg/Environment.cs
+++ b/src/NHibernate/Cfg/Environment.cs
@@ -208,6 +208,20 @@ namespace NHibernate.Cfg
 		/// </summary>
 		public const string OdbcDateTimeScale = "odbc.explicit_datetime_scale";
 
+		/// <summary>
+		/// <para>Oracle has a dual Unicode support model.</para>
+		/// <para>Either the whole database use an Unicode encoding, and then all string types
+		/// will be Unicode. In such case, Unicode strings should be mapped to non <c>N</c> prefixed
+		/// types, such as <c>Varchar2</c>. This is the default.</para>
+		/// <para>Or <c>N</c> prefixed types such as <c>NVarchar2</c> are to be used for Unicode strings.</para>
+		/// </summary>
+		/// <remarks>
+		/// See https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch6unicode.htm#CACHCAHF
+		/// https://docs.oracle.com/database/121/ODPNT/featOraCommand.htm#i1007557
+		/// This setting applies only to Oracle dialects and ODP.Net managed or unmanaged driver.
+		/// </remarks>
+		public const string OracleUseNPrefixedTypesForUnicode = "oracle.use_n_prefixed_types_for_unicode";
+
 		private static readonly Dictionary<string, string> GlobalProperties;
 
 		private static IBytecodeProvider BytecodeProviderInstance;
--- a/src/NHibernate/Dialect/Oracle8iDialect.cs
+++ b/src/NHibernate/Dialect/Oracle8iDialect.cs
@@ -64,9 +64,23 @@ namespace NHibernate.Dialect
 			get { return false; }
 		}
 
+		/// <summary>
+		/// <para>Oracle has a dual Unicode support model.</para>
+		/// <para>Either the whole database use an Unicode encoding, and then all string types
+		/// will be Unicode. In such case, Unicode strings should be mapped to non <c>N</c> prefixed
+		/// types, such as <c>Varchar2</c>. This is the default.</para>
+		/// <para>Or <c>N</c> prefixed types such as <c>NVarchar2</c> are to be used for Unicode strings.</para>
+		/// <para>This property is set according to <see cref="Cfg.Environment.OracleUseNPrefixedTypesForUnicode"/>
+		/// configuration parameter.</para>
+		/// </summary>
+		/// <remarks>
+		/// See https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch6unicode.htm#CACHCAHF
+		/// https://docs.oracle.com/database/121/ODPNT/featOraCommand.htm#i1007557
+		/// </remarks>
+		public bool UseNPrefixedTypesForUnicode { get; private set; }
+
 		public Oracle8iDialect()
 		{
-			RegisterCharacterTypeMappings();
 			RegisterNumericTypeMappings();
 			RegisterDateTimeTypeMappings();
 			RegisterLargeObjectTypeMappings();
@@ -78,6 +92,16 @@ namespace NHibernate.Dialect
 			RegisterKeywords();
 
 			RegisterDefaultProperties();
+		}
+
+		/// <inheritdoc/>
+		public override void Configure(IDictionary<string, string> settings)
+		{
+			base.Configure(settings);
+
+			// If changing the default value, keep it in sync with OracleDataClientDriverBase.Configure.
+			UseNPrefixedTypesForUnicode = PropertiesHelper.GetBoolean(Environment.OracleUseNPrefixedTypesForUnicode, settings, false);
+			RegisterCharacterTypeMappings();
 		}
 
 		#region private static readonly string[] DialectKeywords = { ... }
@@ -137,10 +161,14 @@ namespace NHibernate.Dialect
 			RegisterColumnType(DbType.AnsiStringFixedLength, 2000, "CHAR($l)");
 			RegisterColumnType(DbType.AnsiString, "VARCHAR2(255)");
 			RegisterColumnType(DbType.AnsiString, 4000, "VARCHAR2($l)");
-			RegisterColumnType(DbType.StringFixedLength, "NCHAR(255)");
-			RegisterColumnType(DbType.StringFixedLength, 2000, "NCHAR($l)");
-			RegisterColumnType(DbType.String, "NVARCHAR2(255)");
-			RegisterColumnType(DbType.String, 4000, "NVARCHAR2($l)");
+			RegisterColumnType(DbType.AnsiString, 2147483647, "CLOB"); // should use the IType.ClobType
+
+			var prefix = UseNPrefixedTypesForUnicode ? "N" : string.Empty;
+			RegisterColumnType(DbType.StringFixedLength, prefix + "CHAR(255)");
+			RegisterColumnType(DbType.StringFixedLength, 2000, prefix + "CHAR($l)");
+			RegisterColumnType(DbType.String, prefix + "VARCHAR2(255)");
+			RegisterColumnType(DbType.String, 4000, prefix + "VARCHAR2($l)");
+			RegisterColumnType(DbType.String, 1073741823, prefix + "CLOB");
 		}
 
 		protected virtual void RegisterNumericTypeMappings()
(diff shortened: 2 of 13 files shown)


src/NHibernate/Dialect/Oracle8iDialect.cs
```diff
@@ -80,6 +94,15 @@ public Oracle8iDialect()
 			RegisterDefaultProperties();
 		}
 
+		/// <inheritdoc/>
+		public override void Configure(IDictionary<string, string> settings)
+		{
+			base.Configure(settings);
+
+			UseNPrefixedTypesForUnicode = PropertiesHelper.GetBoolean(Cfg.Environment.OracleUseNPrefixedTypesForUnicode, settings, false);
+			RegisterCharacterTypeMappings();
```
fredericDelaporte: Possible breaking change for those using hbm2ddl: now Unicode string types will be mapped by default to non `N` prefixed types, while NH 4 was mapping to `N` prefixed one.

Doing it the other way round would be a possible breaking change for queries instead, because they would start using `NVarchar2` for Unicode string parameters instead of `Varchar2` (NH 4 currently maps that to `Varchar2`).

Other possible breaking change, character types will not be registered if the dialect is not configured, which may happen if not getting it through `Dialect.GetDialect`.

src/NHibernate/Cfg/Environment.cs
```diff
@@ -186,6 +186,20 @@ public static string Version
...
+		/// will be Unicode. In such case, Unicode strings should be mapped to non <c>N</c> prefixed
+		/// types, such as <c>Varchar2</c>. This is the default.</para>
+		/// <para>Or <c>N</c> prefixed types such as <c>NVarchar2</c> are to be used for Unicode strings.</para>
+		/// </summary>
+		/// <remarks>
+		/// See https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch6unicode.htm#CACHCAHF
+		/// https://docs.oracle.com/database/121/ODPNT/featOraCommand.htm#i1007557
+		/// This setting applies only to Oracle dialects and ODP.Net managed or unmanaged driver.
+		/// </remarks>
+		public const string OracleUseNPrefixedTypesForUnicode = "oracle.use_n_prefixed_types_for_unicode";
```
fredericDelaporte: New setting just for Oracle.

src/NHibernate/Dialect/Oracle8iDialect.cs
```diff
@@ -80,6 +94,16 @@ public Oracle8iDialect()
...
 		}
 
+		/// <inheritdoc/>
+		public override void Configure(IDictionary<string, string> settings)
+		{
+			base.Configure(settings);
+
+			// If changing the default value, keep it in sync with OracleDataClientDriverBase.Configure.
+			UseNPrefixedTypesForUnicode = PropertiesHelper.GetBoolean(Cfg.Environment.OracleUseNPrefixedTypesForUnicode, settings, false);
+			RegisterCharacterTypeMappings();
```
fredericDelaporte: Possible breaking change for those using hbm2ddl: now Unicode string types will be mapped by default to non `N` prefixed types, while NH 4 was mapping to `N` prefixed one.

Doing it the other way round would be a possible breaking change for queries instead, because they would start using `NVarchar2` for Unicode string parameters instead of `Varchar2` (NH 4 currently maps that to `Varchar2`).

Other possible breaking change, character types will not be registered if the dialect is not configured, which may happen if not getting it through `Dialect.GetDialect`.

src/NHibernate/Driver/OracleDataClientDriverBase.cs
```diff
@@ -0,0 +1,176 @@
...
+			_oracleDbTypeNChar = Enum.Parse(oracleDbTypeEnum, "NChar");
+		}
+
+		/// <inheritdoc/>
+		public override void Configure(IDictionary<string, string> settings)
+		{
+			base.Configure(settings);
+
+			// If changing the default value, keep it in sync with Oracle8iDialect.Configure.
+			UseNPrefixedTypesForUnicode = PropertiesHelper.GetBoolean(Cfg.Environment.OracleUseNPrefixedTypesForUnicode, settings, false);
```
fredericDelaporte: Default value `false` avoid a possible breaking change for queries, by keeping the default typing of the ODP.Net provider.

src/NHibernate/Dialect/Oracle8iDialect.cs
```diff
@@ -80,6 +94,16 @@ public Oracle8iDialect()
...
 		}
 
+		/// <inheritdoc/>
+		public override void Configure(IDictionary<string, string> settings)
+		{
+			base.Configure(settings);
+
+			// If changing the default value, keep it in sync with OracleDataClientDriverBase.Configure.
+			UseNPrefixedTypesForUnicode = PropertiesHelper.GetBoolean(Environment.OracleUseNPrefixedTypesForUnicode, settings, false);
+			RegisterCharacterTypeMappings();
```
fredericDelaporte: (Moving the comment on non-outdated file version.)
Possible breaking change for those using hbm2ddl: now Unicode string types will be mapped by default to non `N` prefixed types, while NH 4 was mapping to `N` prefixed one.

Doing it the other way round would be a possible breaking change for queries instead, because they would start using `NVarchar2` for Unicode string parameters instead of `Varchar2` (NH 4 currently maps that to `Varchar2`).

Other possible breaking change, character types will not be registered if the dialect is not configured, which may happen if not getting it through `Dialect.GetDialect`.

fredericDelaporte:
Rebased with async regeneration embedded in "NH-4062 - Oracle Unicode support dual model." commit. (It does just alter one async test.)

src/NHibernate/Driver/OracleDataClientDriverBase.cs
```diff
@@ -0,0 +1,181 @@
...
+		/// <summary>
+		/// Default constructor.
+		/// </summary>
+		/// <param name="managed"><see langword="true"/> for loading the managed driver, <see langword="false"/> for loading the unmanaged driver.</param>
+		/// <exception cref="HibernateException">
+		/// Thrown when the <c>Oracle.DataAccess</c> assembly can not be loaded.
+		/// </exception>
+		protected OracleDataClientDriverBase(bool managed)
+			: this(
+				managed ? _managedProviderInvariantName : _unmanagedProviderInvariantName,
```
hazzik: Can we move these back to the constructor instead?

fredericDelaporte: I do not get it.

Would you like this to be removed and having manage and un-managed classes providing these values themselves? Or should the `ReflectionBasedDriver` be changed for allowing specifying these values without a constructor call (which forces to put that inside `: this(...)` or `: base(...)`), by example with some protected settable properties or a protected initialize method?
Or something else?

hazzik: What I meant is to receive these strings from the child constructor instead of a bool argument. Eg. move this constructor down the inheritance chain:
```
public OracleManagedDataClientDriver : base(_managedProviderInvariantName, ...)
```

Also, it seems that we need to have only 2 variable strings: "Oracle.ManagedDataAccess" and "Oracle.DataAccess", all other are derivatives

fredericDelaporte: Done, with only assembly name as a parameter. My previous idea was to be able to adapt to eventual weird naming changes done on Oracle side for some future version. But well, they may not do same than Sap with its Sybase drivers.

src/NHibernate/Dialect/Oracle8iDialect.cs
```diff
@@ -137,10 +161,14 @@ protected virtual void RegisterCharacterTypeMappings()
...
 			RegisterColumnType(DbType.AnsiString, 4000, "VARCHAR2($l)");
-			RegisterColumnType(DbType.StringFixedLength, "NCHAR(255)");
-			RegisterColumnType(DbType.StringFixedLength, 2000, "NCHAR($l)");
-			RegisterColumnType(DbType.String, "NVARCHAR2(255)");
-			RegisterColumnType(DbType.String, 4000, "NVARCHAR2($l)");
+			RegisterColumnType(DbType.AnsiString, 2147483647, "CLOB"); // should use the IType.ClobType
+
+			var prefix = UseNPrefixedTypesForUnicode ? "N" : string.Empty;
+			RegisterColumnType(DbType.StringFixedLength, prefix + "CHAR(255)");
+			RegisterColumnType(DbType.StringFixedLength, 2000, prefix + "CHAR($l)");
```
hazzik: Not sure why this is 2000.

src/NHibernate/Driver/OracleDataClientDriverBase.cs
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		private OracleDataClientDriverBase(string driverAssemblyName, string clientNamespace)
+			: base(clientNamespace, driverAssemblyName, clientNamespace + ".OracleConnection", clientNamespace + "." + _commandClassName)
+		{
+			var oracleCommandType = ReflectHelper.TypeFromAssembly(clientNamespace + "." + _commandClassName, driverAssemblyName, true);
+			_oracleCommandBindByName = oracleCommandType.GetProperty("BindByName");
+
+			var parameterType = ReflectHelper.TypeFromAssembly(clientNamespace + ".OracleParameter", driverAssemblyName, true);
+			_oracleDbType = parameterType.GetProperty("OracleDbType");
```
hazzik: Can we build delegates for this properties? Like in `MySqlClientSqlCommandSet`?

src/NHibernate/Driver/OracleDataClientDriverBase.cs
```diff
@@ -0,0 +1,171 @@
...
+		}
+
+		private OracleDataClientDriverBase(string driverAssemblyName, string clientNamespace)
+			: base(clientNamespace, driverAssemblyName, clientNamespace + ".OracleConnection", clientNamespace + "." + _commandClassName)
+		{
+			var oracleCommandType = ReflectHelper.TypeFromAssembly(clientNamespace + "." + _commandClassName, driverAssemblyName, true);
+			_commandBindByNameSetter = DelegateHelper.BuildPropertySetter<bool>(oracleCommandType, "BindByName");
+
+			var parameterType = ReflectHelper.TypeFromAssembly(clientNamespace + ".OracleParameter", driverAssemblyName, true);
+			_parameterOracleDbTypeSetter = DelegateHelper.BuildPropertySetter<object>(parameterType, "OracleDbType");
```
fredericDelaporte: We can not use the actual set type here for the delegate. This has required adapting `DelegateHelper`.

src/NHibernate/Util/DelegateHelper.cs
```diff
@@ -19,9 +19,15 @@ internal static class DelegateHelper
 			var parameter = Expression.Parameter(typeof(object), "x");
 			var instance = Expression.Convert(parameter, type);
 			var property = Expression.Property(instance, propertyName);
-			var value = Expression.Parameter(typeof (T), "value");
+			var valueParameter = Expression.Parameter(typeof (T), "value");
+			Expression value = valueParameter;
+			// Cast value if required
+			if (!property.Type.IsAssignableFrom(typeof(T)))
```
fredericDelaporte: Changes required for supporting setting without being able to provide the actual property type as a generic parameter.

PR #668, opened 2017-08-04
NH-4064 - Unmerge Antlr3.Runtime

hazzik:


src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -67,7 +67,7 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Antlr3.Runtime, Version=3.5.0.2, Culture=neutral, PublicKeyToken=eb42632606e9261f, processorArchitecture=MSIL" Condition=" '$(Configuration)' == 'Debug' ">
+    <Reference Include="Antlr3.Runtime, Version=3.5.0.2, Culture=neutral, PublicKeyToken=eb42632606e9261f, processorArchitecture=MSIL">
       <HintPath>..\packages\Antlr3.Runtime.3.5.1\lib\net40-client\Antlr3.Runtime.dll</HintPath>
       <Private>True</Private>
```
fredericDelaporte: This `private` thing was removed from ReLinq when doing NH-3750. It does not seem it changes anything to the reference property window though. (Copy local stays same.) And it is still there for Iesi too. So well, should it be kept or removed, I do not know.

PR #670, opened 2017-08-06
NH-3990 - Upgrade VB test project to VS2017 project structure

ngbrown:
This is a continuation of #605.

* Adds the ability on all test projects to be ran from `dotnet test`.
* Accounts for some changes in the VB compiler around `CompareString` methods being generated from LINQ queries.
* Also, some test behavior is different on my machine in release mode when build with `dotnet build`, so that is addressed.  Specifically, `IEnlistmentNotification.Prepare` doesn't seem to be guaranteed to be called from a different thread like it asserts.

### DIFF
--- a/src/NHibernate.Test.VisualBasic/NHibernate.Test.VisualBasic.vbproj
+++ b/src/NHibernate.Test.VisualBasic/NHibernate.Test.VisualBasic.vbproj
@@ -1,160 +1,51 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProductVersion>
-    </ProductVersion>
-    <SchemaVersion>
-    </SchemaVersion>
-    <ProjectGuid>{7C2EF610-BCA0-4D1F-898A-DE9908E4970C}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <RootNamespace>NHibernate.Test.VisualBasic</RootNamespace>
-    <AssemblyName>NHibernate.Test.VisualBasic</AssemblyName>
-    <FileAlignment>512</FileAlignment>
-    <MyType>Windows</MyType>
-    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
-    <TargetFrameworkProfile />
-    <StartAction>Program</StartAction>
-    <StartProgram>$(MSBuildProjectDirectory)\..\..\Tools\nunit\nunit-x86.exe</StartProgram>
-    <StartArguments>NHibernate.Test.VisualBasic.dll</StartArguments>
+    <TargetFramework>net461</TargetFramework>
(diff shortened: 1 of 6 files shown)


src/NHibernate/Linq/Functions/CompareGenerator.cs
```diff
@@ -44,9 +44,17 @@ internal static bool IsCompareMethod(MethodInfo methodInfo)
...
 
 			// This is .Net 4 only, and in the System.Data.Services assembly, which we don't depend directly on.
-			return methodInfo != null && methodInfo.Name == "Compare" &&
+			if (methodInfo != null && methodInfo.Name == "Compare" &&
 				   methodInfo.DeclaringType != null &&
-				   methodInfo.DeclaringType.FullName == "System.Data.Services.Providers.DataServiceProviderMethods";
+				   methodInfo.DeclaringType.FullName == "System.Data.Services.Providers.DataServiceProviderMethods")
+				return true;
+
+			if (methodInfo != null && methodInfo.Name == "CompareString" &&
```
hazzik: This is normally handled by `VBStringComparisonExpression`. The VB Core/New Roslyn case will be fixed in re-linq 2.2 (it's in alpha-3 now)

ngbrown: Ok. Is this work-around fine for now, or should NHibernate depend on the alpha package?  Also, include `https:\\` in your links...

hazzik: I think, test with 2.2-alpha3 and wait for the official release.

ngbrown: This work-around will have to stay till re-linq 2.2.0-alpha-004.

hazzik:
Don't want to wait as VS2017.3 is released.

src/NHibernate.Test.VisualBasic/NHibernate.Test.VisualBasic.vbproj
```diff
@@ -1,160 +1,57 @@
...
   <ItemGroup>
-    <Import Include="Microsoft.VisualBasic" />
-    <Import Include="System" />
-    <Import Include="System.Collections" />
-    <Import Include="System.Collections.Generic" />
-    <Import Include="System.Data" />
-    <Import Include="System.Diagnostics" />
-    <Import Include="System.Linq" />
-    <Import Include="System.Xml.Linq" />
+    <Compile Remove="Issues\NH3302\**" /> <!--"Like" not supported in new dotnet compiler, not implemented anyways in .net framework-->
```
hazzik: Just noticed this. It's not actually true.

ngbrown: I think I meant that it's not implemented in the full .NET Framework version of NHibernate.  At least it's ignored in [the tests](https://github.com/nhibernate/nhibernate-core/blob/e68bcb8e06c73aab7403f723d07132f40f848b32/src/NHibernate.Test.VisualBasic/Issues/NH3302/Fixture.vb#L9):

```vb.net
        <TestFixture(), Ignore("Not fixed yet.")> _
```

I don't actually know if it has been fixed, as I didn't try running it.  Either way, the whole file doesn't compile with `dotnet build`.

hazzik: It's because Microsoft.VisualBaaic assembly is not referenced

fredericDelaporte:
> Don't want to wait as VS2017.3 is released.

Anyway it is already released, but not detected as an update from Visual Studio IDE. Launching the Visual Studio Installer detects it and allows installing it (at least from a 17.2).

PR #671, opened 2017-08-09
Purge test warnings

fredericDelaporte:
For more easily avoiding adding new compilation warnings to any project, better avoid letting any anywhere.

This PR removes compilations warning from the test project.

An additional commit enables "warnings as errors" for release build. So the release build should detect if any warning is added back. I have not enabled it for debug for avoiding hindering debug session and temporarily fiddling with code.

It appears the Hql.g warnings are not considered by this setting, so they should not block the release build.


src/NHibernate.Test/Generatedkeys/Identity/MyEntityIdentity.cs
```diff
@@ -2,7 +2,10 @@ namespace NHibernate.Test.Generatedkeys.Identity
 {
 	public class MyEntityIdentity
 	{
+		// Assigned by reflection
+#pragma warning disable CS0649 // Field is never assigned to, and will always have its default value
 		private int id;
+#pragma warning restore CS0649 // Field is never assigned to, and will always have its default value
```
fredericDelaporte: Likely most of those cases could have been fixed by using a setter instead. But this would require analyzing each test case for ensuring their current way of mapping the property was not having any bound with the tested feature.

src/NHibernate.Test/NHSpecificTest/NH2188/AppDomainWithMultipleSearchPath.cs
```diff
@@ -24,6 +24,10 @@ public void WhenSerchInMultiplePathsThenNotThrows()
 			string binPath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "bin");
 			var expected = Path.Combine(binPath, Configuration.DefaultHibernateCfgFileName);
 
+			// Test uses an obsolete .Net feature, changing the private path of an ongoing AppDomain.
+			// If that feature is dropped, the tests will likely need to be dropped too. (Or heavily changed
+			// for running in a newly created AppDomain.)
+#pragma warning disable CS0618 // Type or member is obsolete
```
fredericDelaporte: I have not found a way to do that with non-obsolete methods and without spawning a new app domain for the test. It seems the used methods have been obsoleted precisely because they were altering private path for already created app domain, which the .Net designers seem to advise against.

src/NHibernate.Test/NHSpecificTest/NH3850/Fixture.cs
```diff
@@ -2,11 +2,9 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Data;
-using System.Diagnostics;
```
fredericDelaporte: Duplicated namespace probably due to a rebase miss.

src/NHibernate.DomainModel/NHibernate.DomainModel.csproj
```diff
@@ -12,6 +12,11 @@
     <NoWarn>$(NoWarn);3001;3002;3003;3005</NoWarn>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
hazzik: I think it's good to treat warnings as errors for all configurations.

fredericDelaporte: I would like to give it a try with only the release build checking that. If that generates frequently PR failing at this but detected a bit too late (especially already merged), it would be easy and fast to change that.

PR #672, opened 2017-08-10
NH-3990 - Fixes for upgrade to VS2017 project structure

ngbrown:
This corrects a few issues with #605 

* NHibernate.Tool.HbmXsd.csproj wasn't generating an .EXE.
* Because `ABC.hbm.xml` was copied to the `bin` output folders, it was being re-included as an EmbeddedResource.  This will cause problems when the VB project is migrated to VS 2017 and depends on `NHibernate.Test.csproj`.  Solution is to exclude the `bin` folders from being considered when finding files for embedding.
* Certain references are auto-included, so they can be removed.

### DIFF
--- a/src/NHibernate.Test/NHibernate.Test.csproj
+++ b/src/NHibernate.Test/NHibernate.Test.csproj
@@ -21,7 +21,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <EmbeddedResource Include="**\*.hbm.xml" />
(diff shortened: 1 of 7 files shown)


hazzik:
Also, the tool needs the default generation path be changed as the tool is being compiled to a different folder now.

fredericDelaporte:
And it needs a bit more indeed. It was embedding the mapping schema, but lost it during conversion to new format. Going to add a commit once I have get it running again.

PR #675, opened 2017-08-16
NH-4073 - Replace NHibernate.Web.Example with modern version.

ngbrown:
This is [NH-4073](https://nhibernate.jira.com/browse/NH-4073)


hazzik:
Some white spaces are inconsistent

ngbrown:
@hazzik It's because the templates started out at 4 space characters, but the `.editorconfig` dictates tabs for any new lines I create, and then GitHub shows tabs as 8 spaces.

Is the solution to convert every file to tabs then?  Can I set *.cshtml and *.json to 2 character spaces in `.editorconfig`?

hazzik:
Yes, `.editorconfig` would work. I think the options shall be as following:
.xml: spaces, 2 spaces
.json: spaces, 2 spaces
.cshtml: spaces, 4 spaces

ngbrown:
Fixed.

Evidently one can set the hard-tab width in GitHub by setting the `indent_size` in `.editorconfig`.  https://github.com/isaacs/github/issues/170#issuecomment-150489692  But then you wouldn't see when I screw it up. 😄

PR #679, opened 2017-08-30
NH-4075 - Remove code obsolete in 4.x

fredericDelaporte:
[NH-4075](https://nhibernate.jira.com/browse/NH-4075) - Remove old obsolete

And flag recent ones with the version in which they were obsoleted.

### DIFF
--- a/src/NHibernate/Hql/Ast/ANTLR/ASTQueryTranslatorFactory.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/ASTQueryTranslatorFactory.cs
@@ -14,11 +14,6 @@ namespace NHibernate.Hql.Ast.ANTLR
 	/// </summary>
 	public class ASTQueryTranslatorFactory : IQueryTranslatorFactory
 	{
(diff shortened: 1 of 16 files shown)


src/NHibernate/Hql/Ast/ANTLR/ASTQueryTranslatorFactory.cs
```diff
@@ -14,11 +14,6 @@ namespace NHibernate.Hql.Ast.ANTLR
 	/// </summary>
 	public class ASTQueryTranslatorFactory : IQueryTranslatorFactory
 	{
-		public IQueryTranslator[] CreateQueryTranslators(string queryString, string collectionRole, bool shallow, IDictionary<string, IFilter> filters, ISessionFactoryImplementor factory)
```
fredericDelaporte: This one was obsolete in `IQueryTranslatorFactory` but not in its concrete implementation. If we want to be conservative, then we should flag it obsolete instead of removing it.

hazzik: That's fine.

PR #680, opened 2017-08-30
NH-4076 - Do not resurrect session

fredericDelaporte:
[NH-4076](https://nhibernate.jira.com/browse/NH-4076) - Do not resurrect session

The Linq provider holds a `WeakReference` to the session because the provider may end up cached in query cache. This should allow the session to be garbage collected when it only remains referenced from query cache.

But this weak reference is set with an option allowing the session object to be resurrected. In such case trying to use it can only failed since it has been disposed. Better just directly throw.

### DIFF
--- a/src/NHibernate.Test/Linq/ExpressionSessionLeakTest.cs
+++ b/src/NHibernate.Test/Linq/ExpressionSessionLeakTest.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Reflection;
 using NHibernate.DomainModel.Northwind.Entities;
 using NHibernate.Linq;
 using NUnit.Framework;
@@ -20,10 +21,12 @@ namespace NHibernate.Test.Linq
 
 		private WeakReference DoLinqInSeparateSession()
 		{
+			// Using SessionFactory of another session dodge TestCase session tracking, which is needed for having session garbage collected
+			// Otherwise DebugSessionFactory session tracking should be changed to use WeakReference too.
 			using (var leakTest = session.SessionFactory.OpenSession())
 			{
 				// It appears linq expressions will (or might) contain a reference to the
-				// IQueryable. At time of writing, linq expressions are helt within NhLinqExpression,
+				// IQueryable. At time of writing, linq expressions are held within NhLinqExpression,
 				// which in turn will be held in the query plan cache. Since the IQueryable will
 				// be an NhQueryable, which holds a reference to the SessionImpl instance,
 				// we will be leaking session instances.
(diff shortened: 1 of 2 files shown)


src/NHibernate/Linq/DefaultQueryProvider.cs
```diff
@@ -23,16 +23,24 @@ public class DefaultQueryProvider : INhQueryProvider
...
+			_session = new WeakReference<ISessionImplementor>(session);
 		}
 
 		protected virtual ISessionImplementor Session
 		{
-			get { return _session.Target as ISessionImplementor; }
+			get
+			{
+				if (!_session.TryGetTarget(out var target))
+					throw new InvalidOperationException("Session has already been garbaged collected");
```
hazzik: I think there shall be no "d" at the end of "garbage"

hazzik:
I think we need a test which will produce the query out of the query cache and see how it'll behave.

src/NHibernate.Test/Linq/ExpressionSessionLeakTest.cs
```diff
@@ -33,5 +34,36 @@ private WeakReference DoLinqInSeparateSession()
 				return new WeakReference(leakTest, false);
 			}
 		}
+
+		private static readonly PropertyInfo _sessionProperty = typeof(DefaultQueryProvider).GetProperty("Session", BindingFlags.NonPublic | BindingFlags.Instance);
+
+		[Test]
+		public void SessionIsNotResurrected()
```
hazzik: This test does not proof anything. It does not fail, with the old code.

hazzik: To make this test valid you need to remove `using` around session variable. Why? When you call `Dispose` the session is marked with suppress finilizer flag, so there is no "long" tracking then, no finilizer => nothing to await for.

fredericDelaporte: With the old code this was yielding `null` in case the session was disposed (and I had tested that), which would have always cause a null ref failure in normal use cases, because `Session` property is only called by code using it without checking for null. Due to the way this property is currently used, having `null` yielded should be considered a failure, because a null ref exception is in most case an almost meaningless error. (It is way better yielding the `InvalidOperationException`.)

Yes I was aware that resurrecting was requiring not disposing the session. But well, with the session tracking mechanism put in the test base class, we have a difficulty here. (Maybe time to solve it by the way, it could be related to troubles those weak-references tests causes with .Net Core.)

fredericDelaporte:
> I think we need a test which will produce the query out of the query cache and see how it'll behave.
This trouble concerns the `HqlQueryPlan` cache which stores NhExpression which references expression which may hold a reference to a Queryable (I guess when using sub-queries).

So extracting from that the queryable looks a bit challenging.

Yes this PR is about fixing something that is a theoretical semantic error but is likely to not cause any actual bug in all normal use cases.

A way to have troubles with that error is to reuse a queryable after having let its session been de-referenced without having closed/disposed it. In such case the provider would resurrect the session and use it, with quite unpredictable behavior. Going to add a test for that.

fredericDelaporte:
For testing the old behavior, I had simulated it by changing default query provider to:

```C#
public DefaultQueryProvider(ISessionImplementor session)
{
	// Short reference (no trackResurrection). If the session gets garbage collected, it will be in an unpredictable state:
	// better throw rather than resurrecting it.
	// https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/weak-references
	_session = new WeakReference<ISessionImplementor>(session, true);
}

protected virtual ISessionImplementor Session
{
	get
	{
		if (!_session.TryGetTarget(out var target))
			return target;//throw new InvalidOperationException("Session has already been garbage collected");
		return target;
	}
}
```

hazzik:
First of all, the PR and JIRA is about "resurrecting" of the session, meaning returning the session for which the finalizer (sic!) is pending. In a "normal" use case, where people do `using` the `Displose` method will mark session with suppress finalizer flag. Which mean that the `WeakReference` would be short-lived. In the old user case (pre-.NET 2.0) when the pattern is to call `session.Close()` without `using` the session would be "finalized", so the `WeakReference` will be long-lived. And you do not have a test for this case.

>With the old code this was yielding null in case the session was disposed (and I had tested that), which would have always cause a null ref failure in normal use cases, because Session property is only called by code using it without checking for null. Due to the way this property is currently used, having null yielded should be considered a failure, because a null ref exception is in most cases an almost meaningless error. (It is way better yielding the `InvalidOperationException`.)

This code does not yield any issues, In fact the only case this is possible is following code:

```
IQueryable<Product> q;
using (var s = OpenSession()) {
  q = s.Query<Product>();
}
```
But again, this case is untested, and I would consider this usage an invalid. 

So, there are 2 tests missing:

1) The test of actually long lived session (the same test, but without using). The added test was not failing with the old code (if we accept that `null` instead of an exception was the expected behavior)
2) The test of using the query outside of session.

hazzik:
I did a quick test with old provider:
1) The long-lived session yields `ObjectDisposedException` 
2) The short-lived session yields `NullReferenceException`

hazzik:
>This trouble concerns the HqlQueryPlan cache which stores NhExpression which references expression which may hold a reference to a Queryable (I guess when using sub-queries).

No it does not. The `IQueryable` inside the plan is always replaced by the true query to fix the issue with expanding parameters (`IEnumerable<>` which is generated into `IN (@p1_1, @p1_2, ...)` statement). But this case is also untested...

hazzik:
I added a case when we close session instead of disposing. Also, removed code duplication.

fredericDelaporte:
> And you do not have a test for this case.

> So, there are 2 tests missing

I guess you had not seen that I have adjusted tests after your comment for covering that, meaning I was acknowledging the long-live case and the invalid usage case were not tested.

> I would consider this usage an invalid.

Precisely what I was referring to in this [comment](https://github.com/nhibernate/nhibernate-core/pull/680#issuecomment-326357913):

>Yes this PR is about fixing something that is a theoretical semantic error but is likely to not cause any actual bug in all normal use cases.
>
>A way to have troubles with that error is to reuse a queryable after having let its session been de-referenced without having closed/disposed it. In such case the provider would resurrect the session and use it, with quite unpredictable behavior. Going to add a test for that.

About:

> > unpredictable behavior.
>
> This's not true

I have seen that in my own testing but that it is still what [documents Microsoft](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/weak-references#short-and-long-weak-references):

> This allows the object to be recreated, but the state of the object remains unpredictable.

So for such case I would rather not rely on the behavior seen on a few tests for our resurrected session.

Overall it seems to me we mostly understand the same things but were misunderstanding each-other.

Just one point:

> No it does not. The IQueryable inside...

Does that means there is no more any actual reason to have that `WeakReference` instead of a strong one in `DefaultQueryProvider`? The [test case seems to imply this was needed only to avoid leaking session in the plan cache](https://github.com/nhibernate/nhibernate-core/pull/680/commits/477244836f1d27a9c48a79fdd44127954148bf1d#diff-1451064b9ad5a97df745f19eaac41016R28):

> ```C#
> // It appears linq expressions will (or might) contain a reference to the
> // IQueryable. At time of writing, linq expressions are held within NhLinqExpression,
> // which in turn will be held in the query plan cache. Since the IQueryable will
> // be an NhQueryable, which holds a reference to the SessionImpl instance,
> // we will be leaking session instances.
> ```

We could remove that weak reference added complexity if all that was not true.

hazzik:
> guess you had not seen that ...
Yes, I did not check the new commits when was writing the comment.

>Does that means there is no more any actual reason to have that WeakReference instead of a strong one in DefaultQueryProvider?

The query plan still holds the expression, but the expression itself is not used: it gets replaced by the actual expression.

PR #682, opened 2017-09-04
NH-4077 - do not auto-flush while already flushing

fredericDelaporte:
[NH-4077](https://nhibernate.jira.com/browse/NH-4077).

The trouble was due to auto-flushing getting triggered from a flush event, altering the flush action list while it was iterating over it.


src/NHibernate/Engine/ISessionImplementor.cs
```diff
@@ -225,7 +225,17 @@ public partial interface ISessionImplementor
...
+		/// Is auto-flush suspended?
+		/// </summary>
+		bool DontFlushFromFind { get; }
+
+		/// <summary>
+		/// Suspend auto-flushing, yielding a disposable to dispose when auto flush should be restored. Supports
+		/// being called multiple times.
+		/// </summary>
+		/// <returns>A disposable to dispose when auto flush should be restored.</returns>
+		IDisposable SuspendAutoFlush();
```
fredericDelaporte: Needing it from an other internal class than the session, and factoring it by the way.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1611,25 +1602,26 @@ private void Filter(object collection, string filter, QueryParameters queryParam
...
+					}
+					catch (Exception e)
+					{
+						throw Convert(e, "could not execute query");
+					}
+					finally
+					{
+						AfterOperation(success);
+					}
 				}
```
fredericDelaporte: Same as previous `AfterOperation` case.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -536,25 +533,26 @@ public override void List(IQueryExpression queryExpression, QueryParameters quer
...
+					}
+					catch (Exception e)
+					{
+						throw Convert(e, "Could not execute query");
+					}
+					finally
+					{
+						AfterOperation(success);
+					}
 				}
```
fredericDelaporte: Discrepancy: the `AfterOperation(success);` was previously put after decrementing back `dontFlushFromFind`, meaning after operations could trigger auto-flush. Now it is done before leaving the new using, so before re-enabling auto-flushes.
But `AfterOperation` does in fact nothing if we are inside a transaction. And when we are not inside a transaction, auto-flush is anyway already disabled. So this change has no consequences with current code.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1793,30 +1785,30 @@ public override void List(CriteriaImpl criteria, IList results)
...
-				}
-				catch (Exception sqle)
-				{
-					throw Convert(sqle, "Unable to perform find");
-				}
-				finally
-				{
-					dontFlushFromFind--;
-					AfterOperation(success);
 				}
```
fredericDelaporte: Same as previous `AfterOperation` case.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1891,16 +1883,17 @@ public override void ListCustomQuery(ICustomQuery customQuery, QueryParameters q
...
+					try
+					{
+						ArrayHelper.AddAll(results, loader.List(this, queryParameters));
+						success = true;
+					}
+					finally
+					{
+						AfterOperation(success);
+					}
 				}
```
fredericDelaporte: Same as previous `AfterOperation` case.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -293,6 +292,49 @@ protected bool IsAlreadyDisposed
...
 
+		public bool DontFlushFromFind => _suspendAutoFlushCount != 0;
+
+		[NonSerialized]
+		private int _suspendAutoFlushCount;
+
+		/// <inheritdoc/>
+		public IDisposable SuspendAutoFlush()
+		{
+			_suspendAutoFlushCount++;
```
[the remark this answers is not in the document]
fredericDelaporte: Make sens moving to constructor. But as session are inherently non-thread safe and so are supposed to be used only by one thread at a time, I will not introduce a thread safety guard here.

hazzik:
I think it would make sense to move `SuspendAutoFlush` to `IEventSource` interface and the implementation into `SessionImpl`. As none of these matters for stateless sessions

fredericDelaporte:
So, moving by the way the `AutoFlushSuspended` property (previously named `DontFlushFromFind`) too.

hazzik:
>So, moving by the way the AutoFlushSuspended
Yes, I meant to write it, but forgot

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -293,6 +292,50 @@ protected bool IsAlreadyDisposed
...
+			public SuspendAutoFlushHelper(AbstractSessionImpl session)
+			{
+				_session = session;
+				_session._suspendAutoFlushCount++;
+			}
+
+			#region IDisposable Support
+			private bool _disposedValue; // To detect redundant calls
+
+			protected virtual void Dispose(bool disposing)
```
hazzik: This class is private, so I dont see the reasons to have virtual members and/or implement Disposable "correctly"

fredericDelaporte: Default standard snippet from VS, including the anti-redundant call (which is a must for this case). 

And well, we could consider than when `disposing` is false (meaning being called by GC, explicit dispose lacking) we should left the counter incremented for avoiding "fixing" the lack of dispose call through GC and hiding somewhat it.

fredericDelaporte: Now moved to `SessionImpl`, pointless `virtual` removed.

hazzik: But, then we miss finilizer, which should call `Dispose(false)` then

fredericDelaporte: No more needed since Fx 2 I believe, handled by GC without any finalizer. The fact that `SessionImpl` still do that is an old remnant of Fx 1.

hazzik: But anyway no one is going to call `Dispose(false)`, so I don't think we need to follow this obsolete pattern. Also, I don't think that framework will call `Dispose(false)` or `Dispose()` on the finalization stage. We need to prevent calling `Dispose()` twice, but anyway, without this check we'll get `NullReferenceException` as session would be unset by the time of second time `Dispose()` is called.

fredericDelaporte: From where is the `protected virtual Dispose(boolean disposing)` pattern said to be obsolete? Only implementing explicitly a finalizer is obsolete for most cases. This [doc](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) seems quite current.

Now for such a little class it should not ever have consequences, but I just rather stick to recommended patterns when they are strongly recommended as in previously linked documentation.

hazzik:
I would implement the helper as following:

```csharp
class SuspendAutoFlushHelper : IDisposable
{
	SessionImpl _session;
	public SuspendAutoFlushHelper(SessionImpl session)
	{
		_session = session;
		_session._suspendAutoFlushCount++;
	}

	public void Dispose()
	{
		if (_session == null)
			throw new ObjectDisposedException("The SuspendAutoFlushHelper has been disposed already");

		_session._suspendAutoFlushCount--;
		_session = null;
	}
}
```
This just fit for purpose. The `ObjectDisposedException` has been just added for those who do not like `NullReferenceException` in general.

src/NHibernate/Impl/AbstractSessionImpl.cs
hazzik: I don't see where it says that finalize is optional:

What I see is:
> You **must** override Finalize to ensure that unmanaged resources are disposed of if your System.IDisposable.Dispose implementation is not called by a consumer of your type.

But again, we implement the `IDisposable` interface not for it's purpose, but because of convenience of `using`.

hazzik:
Also, there is no point of `protected` `Dispose(bool)` as the class is private and don't have any inheritors, so, make it private, and we're good to go.

src/NHibernate/Impl/AbstractSessionImpl.cs
fredericDelaporte: The fact is my implementation is invalid for the disposing pattern because I should not touch the managed session when disposing is false: if both were dereferenced, the session could have been GC already and should no more be used in any way.

I am probably wrong by telling the GC call disposing(false) itself, it is likely only useful if some later inheritor decide to implement its own finalizer though this should no more be the way to go (use SafeHandle instead for not having to handle directly the releasing of unmanaged resources).

But well, following the recommended practice just avoids having to think about such details while not being a significant overhead.

fredericDelaporte:
I not really agree with the `ObjectDisposedException`. An usual dispose should avoid throwing (although some well known classes depart from that, like `TransactionScope`). This exception is for other methods of the class. The dispose itself should just silently do nothing when called multiple times.

Now for this hijacked dispose, ok, I will code it your way.

src/NHibernate/Impl/AbstractSessionImpl.cs
fredericDelaporte: > This is the recommended variation and doesn't require overriding the System.Object.Finalize method.

This is in the first point above. The "must" is in the second way of implementing dispose, while the recommended way is the first.

fredericDelaporte:
Thanks for the reviewing,

I have seen in some previous merge to master than updating the branch (merging from master) then merging back to master was causing some noise like [here](https://github.com/nhibernate/nhibernate-core/commits/master?after=5d8219b39c3f904ee0b3a04e6b692c1bbd478270+30) for the tests warning.

So I am going to rebase instead, discarding the review by the way.

hazzik:
np

PR #684, opened 2017-09-07
NH-3578 - Subcriteria.UniqueResult<T> for value types should return default(T), same as CriteriaImpl.UniqueResult<T>

hazzik:
rework of #243

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -34,3 +34,4 @@ SharedAssemblyInfo.*
 hibernate.cfg.xml
 .vs/
 packages/
+.idea/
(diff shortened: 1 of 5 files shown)


hazzik:
Fixed compilation issue. Should be good now.

src/.gitignore
```diff
@@ -33,4 +33,5 @@ _ReSharper*/
 SharedAssemblyInfo.*
 hibernate.cfg.xml
 .vs/
-packages/
\ No newline at end of file
+packages/
+.idea/
```
fredericDelaporte: IntelliJ? And maybe unintended, I do not see how it relates to this PR.

hazzik:
Rider. Not related:)

hazzik:
Probably was captured during git amendment

fredericDelaporte:
Probably no harm to have it and more convenient. For who is using this IDE.

VS is terribly slow on NHibernate solution since project format change. Slow on load, freezes many seconds frequently after file changes by an external process, "reloading" changes for compiling during half a minute... Maybe I should check Rider.

PR #685, opened 2017-09-08
NH-3905 - Replacing more manual async implementation by generation.

fredericDelaporte:
Upgrading to async generator 0.4.2 and using it to generate previously manually added async overrides of methods having sync overrides.

The current configuration restrict this generation for the test project to types from NHibernate assembly.
And this generation is not enabled for the NHibernate project itself.

I think we should enable it for all types and for the NHibernate project itself. It would help avoiding forgetting to override manually an async counterpart not taken into account by current configuration. I have tested it, it currently generates nothing more, so currently: we have not forgotten anything more; it does not harm by generating unneeded things.


fredericDelaporte:
Any thoughts about following?

> I think we should enable it for all types and for the NHibernate project itself. It would help avoiding forgetting to override manually an async counterpart not taken into account by current configuration. I have tested it, it currently generates nothing more, so currently: we have not forgotten anything more; it does not harm by generating unneeded things.

And maca88 has just published a new version promising to be faster, I will probably change this PR to use it.

hazzik:
I think it's ok to enable proposed configs, but I'm not sure what are the implications.

fredericDelaporte:
Done,

The latest version runs in less than 160" on my setup, while previously it was requiring at least 250". Good job @maca88

PR #686, opened 2017-09-12
[WIP] NH-4036 - Cleaning up non generic usages

fredericDelaporte:
[NH-4036](https://nhibernate.jira.com/browse/NH-4036) - Cleaning up non generic usages

Many features implementation relies on non generic types, especially `IEnumerable`/`ICollection`/`IDictionary`, where they could be using the generic counterpart.

This causes NHibernate to have a bunch of helpers for handling non generic collections, since they tend to have less utilities available natively (lack of Linq methods by example).

There is also some API redundancy, NHibernate continuing to provide non-generic API along generic ones.

The implied code-base seems a bit huge, so this will likely require to be done in many PR. This one targets removing the non-generic helper by migrating code using it to generic whenever sensible. Those utilities could be a nuisance, especially when they are public extensions methods, because they can pollute NHibernate's users with unwanted extensions, while that is not the business of NHibernate to provide such utilities.

In its current state, this PR obsolete `NHibernate.Util.EnumerableHelper`. It aims at obsoleting some more such classes.


src/NHibernate/Util/WeakHashtable.cs
```diff
@@ -131,124 +117,154 @@ public bool MoveNext()
...
-			get { return Entry; }
+			// Do not change this code. Put cleanup code in Dispose(bool disposing) above.
+			Dispose(true);
 		}
+		#endregion
 	}
 
 	[Serializable]
-	public class WeakHashtable : IDictionary
+	public class WeakHashtable : IDictionary<object, object>
```
fredericDelaporte: Minimal switch, the next commit turns this to a fully generic type `WeakHashtable<TKey, TValue>`.

But I question whether it is worth it or not. On Java side, they have ceased using this for query plan caching. (Java side, they have start using that weak hashtable for first level cache it seems.) They are using a 2048 max entries (configurable) plan cache instead with an algorithm for evicting coldest entries. And moreover, they use something more efficient than weak references for their remaining "caching" purpose. (Usage of weak reference for caching is by the way advises against by Microsoft documentation.) Weak references may be reclaimed while being far from reaching available memory limits. Java use soft references instead, designed to be reclaimed only when under memory stress. They do not have equivalent in .Net. 

We may have better throwing that away for using a `System.Runtime.Caching.MemoryCache` instead (available since fx4).

src/NHibernate/DebugHelpers/DictionaryProxy.cs
```diff
@@ -1,34 +1,8 @@
...
-using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 
 namespace NHibernate.DebugHelpers
 {
-	/// <summary>
-	/// Used to show a better debug display for dictionaries
-	/// </summary>
-	public class DictionaryProxy
```
fredericDelaporte: Was having one invalid usage left: applied on a non-dictionary collection.

src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -24,7 +24,7 @@ public class QueryPlanCache
 		private readonly SimpleMRUCache sqlParamMetadataCache = new SimpleMRUCache();
 
 		// the cache of the actual plans...
-		private readonly SoftLimitMRUCache planCache = new SoftLimitMRUCache(128);
+		private readonly SoftLimitMRUCache<QueryPlanKey, object> planCache = new SoftLimitMRUCache<QueryPlanKey, object>(128);
```
fredericDelaporte: Values are different classes here without a common ancestor. I had introduced one first, just to realize it was having almost no added value, because all the cached values currently do not override any `Equals`, causing them to only compare by reference.

src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -157,16 +157,15 @@ private ParameterMetadata BuildNativeSQLParameterMetadata(string sqlString)
...
-					new NamedParameterDescriptor(name, null, description.JpaStyle);				
+					new NamedParameterDescriptor(name, null, description.JpaStyle);
 			}
 
 			return new ParameterMetadata(ordinalDescriptors, namedParamDescriptorMap);
 		}
 
 		[Serializable]
-		private class HQLQueryPlanKey : IEquatable<HQLQueryPlanKey>
+		private class HQLQueryPlanKey : QueryPlanKey
```
fredericDelaporte: `IEquatable` was having no usage here because the class was stored in a non generic dictionary. Furthermore, now it is stored in a generic dictionary of `QueryPlanKey`, which implements `IEquatable<QueryPlanKey>`. I am wrong believing it is useless to add `IEquatable` for subclasses not stored in collection specifically using their type, not the base one? Adding the specialized `IEquatable` for subclasses along with base equatable implementation is incurring some code duplication or inefficiencies at least.

src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -200,46 +199,24 @@ protected HQLQueryPlanKey(System.Type queryTypeDiscriminator, string query, bool
...
-					hash = 29*hash + queryTypeDiscriminator.GetHashCode();
+					hash = 29 * hash + (shallow ? 1 : 0);
+					hash = 29 * hash + CollectionHelper.GetHashCode(filterNames);
+					hash = 29 * hash + queryTypeDiscriminator.GetHashCode();
 					hashCode = hash;
 				}
 			}
 
-			public override bool Equals(object obj)
+			public override bool Equals(QueryPlanKey other)
```
fredericDelaporte: Overrides the base one for first calling it and ruling out simple inequality cases (not same hashcode or not same query) then if deemed equal by base implementation, checking the additional constraints of the subclass.

src/NHibernate/Util/SoftLimitMRUCache.cs
```diff
@@ -22,33 +22,33 @@ namespace NHibernate.Util
 	/// enqueuement.
 	/// </remarks>
 	[Serializable]
-	public class SoftLimitMRUCache : IDeserializationCallback
+	public class SoftLimitMRUCache<TKey, TValue> : IDeserializationCallback where TKey : class where TValue : class
```
fredericDelaporte: As said on first commit review, I question whether it is worth it or not. On Java side, they have ceased using this for query plan caching. (Java side, they have start using that weak hashtable for first level cache it seems.) They are using a 2048 max entries (configurable) plan cache instead with an algorithm for evicting coldest entries. And moreover, they use something more efficient than weak references for their remaining "caching" purpose. (Usage of weak reference for caching is by the way advises against by Microsoft documentation.) Weak references may be reclaimed while being far from reaching available memory limits. Java use soft references instead, designed to be reclaimed only when under memory stress. They do not have equivalent in .Net. 

We may have better throwing that away for using a `System.Runtime.Caching.MemoryCache` instead (available since fx4).

src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -701,14 +727,34 @@ public IQuery SetParameterList(string name, IEnumerable vals, IType type)
 			{
 				throw new ArgumentNullException("type","Can't determine the type of parameter-list elements.");
 			}
+#pragma warning disable 618 // Any(taking non generic IEnumerable) is obsolete for turning it internal
 			if(!vals.Any())
+#pragma warning restore 618
 			{
```
fredericDelaporte: `SetParameterList(string name, IEnumerable vals, IType type)` is kept, because the element type is already specified as NH type through the third argument, and switching to `IEnumerable<T>` would require to use reflection for many callers, which have a non typed parameters list source possibly mixing different `IEnumerable<T>` values. (Detached query, multi query, ...) Their only common interface is `IEnumerable`. Covariancy to  `IEnumerable<object>` does not help because `IEnumerable<int>` by example can not vary to `object`, it is a `struct`, while `IEnumerable<string>` can.

fredericDelaporte:
Going to be done through some other PR (some already merged).

PR #687, opened 2017-09-13
NH-3669 - Make Query<> methods interface members of ISession and IStatelessSession

hazzik:


src/NHibernate/Linq/LinqExtensionMethods.cs
```diff
@@ -2395,26 +2395,6 @@ public static Task<List<TSource>> ToListAsync<TSource>(this IQueryable<TSource>
 
 		#endregion
 
-		public static IQueryable<T> Query<T>(this ISession session)
```
hazzik: NHibernate v5 assembly is binary incompatible with previous version, so just remove these methods. In the majority of cases the change is source compatible and would not cause any problems. The only case it breaks if an user has used `LinqExtensionMethods.Query` as a regular static method, but I expect this to be a really minor case.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -444,5 +446,26 @@ internal IOuterJoinLoadable GetOuterJoinLoadable(string entityName)
 		public abstract IEnumerable<T> Enumerable<T>(IQueryExpression queryExpression, QueryParameters queryParameters);
 
 		public abstract int ExecuteUpdate(IQueryExpression queryExpression, QueryParameters queryParameters);
+		
+		/// <summary>
+		/// Creates a new Linq <see cref="IQueryable{T}"/> for the entity class.
+		/// </summary>
+		/// <typeparam name="T">The entity class</typeparam>
+		/// <returns>An <see cref="IQueryable{T}"/> instance</returns>
+		public IQueryable<T> Query<T>()
```
fredericDelaporte: I now tend to just put `/// <inheritdoc />` for concrete implementation, in order to avoid duplicating doc, although it is not supported by all tools (actually for what NHibernate use, only sandcastle supports it I believe). What do you think of that?

hazzik: It doesn't work this way in this case as AbstractSessionImpl does not implement either of these interfaces.

PR #688, opened 2017-09-14
NH-3669 - documenting move of Linq Query from extension

fredericDelaporte:
NH-3669 - documenting move of Linq Query from extension to plain method of session

I have forgot the doc in my review.

### DIFF
--- a/doc/reference/modules/query_linq.xml
+++ b/doc/reference/modules/query_linq.xml
@@ -6,9 +6,11 @@
     for querying with NHibernate.
   </para>
   <para>
-    The Linq provider works as an extension of the <literal>ISession</literal>. It is defined in the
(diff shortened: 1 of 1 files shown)


doc/reference/modules/query_linq.xml
```diff
@@ -65,7 +67,9 @@ using NHibernate.Linq;]]></programlisting>
     <para>&nbsp;</para>
 
     <para>
-      A client timeout for the query can be defined.
+      A client timeout for the query can be defined. As most others NHibernate specific features for
+      Linq, this is available through an extension defined in <literal>NHibernate.Linq</literal>
+      namespace.
```
fredericDelaporte: Most other features were at least already mentioning they were extensions methods without precising again the namespace, but I think it should be enough this way.

PR #689, opened 2017-09-15
NH-3700 - Remove CodeDom BytecodeProviderImpl

ngbrown:
This is [NH-3700](https://nhibernate.jira.com/browse/NH-3700).

The CodeDom BytecodeProviderImpl was intended for .Net 1.1 and has not been the default for many years. It doesn't actually work currently.

This was slated for v5.0.  It's not really a breaking change as it hasn't been usable; see [NH-2798](https://nhibernate.jira.com/browse/NH-2798).


hazzik:
>It doesn't actually work currently.
This isn't true. It works.

ngbrown:
There's currently zero tests that cover the results of the CodeDom code generation, and the code it generates is incorrect.  The first error in the generated code is `using NHibernate.Property;` because there is no matching namespace.

src/NHibernate/nhibernate-configuration.xsd
```diff
@@ -15,7 +15,6 @@
 						<xs:documentation>
 							There are 3 default short-cut values
 							- lcg     : default for .NET2.0 and higher.
-							- codedom : CodeDOM-based bytecode provider (mostly for .NET1.1).
 							- null    : Disable the reflection optimization completely.
```
fredericDelaporte: `There are 3` should be changed to `There are 2`.

ngbrown: Fixed.

PR #674, opened 2017-08-15
NH-3927 - Switch to SemVer version generation

hazzik:
- SharedAssemblyInfo.?? are retired

PR #690, opened 2017-09-15
Doc fixes

fredericDelaporte:


doc/reference/modules/nhibernate_caches.xml
```diff
@@ -480,17 +488,6 @@
...
 		</section>
 
-		<section>
-			<title>Patches</title>
-
-			<para>
-				There is a known issue where some SQL Server 2005 notifications might not be received when an
-				application subscribes to query notifications by using ADO.NET 2.0. To fix this problem install
-				<ulink url="http://support.microsoft.com/Default.aspx?kbid=913364">SQL hotfix for kb 913364</ulink>.
-			</para>
```
fredericDelaporte: Link dead, kb not found anymore, probably merge in a SQL 2005 SP since long.

doc/reference/modules/performance.xml
```diff
@@ -659,7 +659,13 @@ ICat fritz = (ICat) iter.Current;]]></programlisting>
             <emphasis>In NHibernate 1.x the second level cache does not work correctly in combination with
             distributed transactions.</emphasis>
         </para>
-        
+
+        <para>
+          The second level cache requires the use of transactions, be it through transaction scopes or NHibernate
+          transactions. Interacting with the data store without an explicit transaction is discouraged, and will
+          not allow the second level cache to work as intended.
+        </para>
```
fredericDelaporte: One frequent cause of complains about second level cache not working.

doc/reference/modules/nhibernate_caches.xml
```diff
@@ -153,7 +153,15 @@
...
 			</para>
+			<para>
+				The second level cache requires the use of transactions, be it through transaction scopes or NHibernate
+				transactions. Interacting with the data store without an explicit transaction is discouraged, and will
+				not allow the second level cache to work as intended.
+			</para>
 		</formalpara>
+		<para>
+			See also <xref linkend="performance-cache" />.
+		</para>
```
fredericDelaporte: We have two separated section documenting second level cache, each with valuable information not available in the other. It would be better to merge them. At least I add a cross link here. (The other side already does it.)

fredericDelaporte:
Building SHFB was failing due to two reasons: 
* missing build property `project.version.numeric`, removed in #674.
* [seems incompatible with dotnet.exe](https://github.com/EWSoftware/SHFB/issues/517), which has replaced msbuild since #605.

The file `NHibernate.ndoc.in` referenced another missing property (`project.version.short`), but it does not seems to cause issues. Is this file an obsolete artifact? I do not find any reference to it.

fredericDelaporte:
The doc was referring to a bunch of non existent method, like `ISession.Find`, `ISession.Enumerable`, `ISession.Filter` which were likely short-hands for creating queries/filters and executing them. I have replaced them. And done many other fixes and modernizations.

PR #692, opened 2017-09-19
NH-3905 - upgrade async generator and regen

fredericDelaporte:
Speeds up generation by 30%, removes <content> generated comments which sometimes interfere with some tooling (Rider), and clean some undue generated imports.

The <content> comment removal (maca88/AsyncGenerator#55) is responsible of most changes, while some async test files also have some undue `import System.Threading` removed (maca88/AsyncGenerator#51).


fredericDelaporte:
Hi-jacked this PR for dropping a note on async in documentation. (I will not squash commits together.)

PR #553, opened 2017-02-14
NH-3845, NH-3946 - Fix Linq queries with `.OfType<T>()` and `where is T`, where T is a mapped base type

alexDevBR:
Patch allows queries like `session.Query<Animal>().Where(x => x is Mammal)` and `session.Query<Animal>().OfType<Mammal>()` to get all dogs and cats, for example.

PR #693, opened 2017-09-19
NH-4000 - Prepare release of v5.0

fredericDelaporte:
[NH-4000](https://nhibernate.jira.com/browse/NH-4000) - Proposed changes to release notes for v5.0.

~~[NH-3845](https://nhibernate.jira.com/browse/NH-3845) not included as it does not seem to be fully fixed, see #553.~~

[NH-3945](https://nhibernate.jira.com/browse/NH-3945) (Antlr3 upgrade) is labeled with "possible breaking changes", but I was unable to determine them.

Elaborated from the Jira list of fixed for 5.0.0. Opened issues (excepted NH-4000) have been removed from the release. (Mostly moved to 5.1, some to 6.0 or nothing.)

Test related items are identified as such, ~~one has been dropped from the list (support of inconclusive results)~~.

Some minor titles adjustment has been done for some issues in order for them to be more explicit. Some others should maybe be reworded, especially bugs which titles are sometimes remotely related to their actual cause.

Previous week I had review all PR with milestone:5.0 and checked their corresponding Jira, then additionally check all closed PR by merged date down to november 2016, so it should not stay some missed change.

Update: Well, thinking about it, there was a constructor change somewhere likely for some factory I have not mention, and do not remember what it was.


fredericDelaporte:
Since it seems not an usual practice, I have added as a separated commit some highlights. To be dropped if considered undesirable there.

About documentation, I am thinking about maybe one lacking point with async: how to mix async and lazy-loading. Users may believe lazy-loading can not be async if they do not know about `NHibernateUtil.InitializeAsync` (although indirectly documented since `NHibernateUtil.Initialize` is documented and all IO bound methods are said to have async counterparts). But adding an async section just for that (and maybe re-warning about awaiting each call) would be a bit overkill. And furthermore using such method before each potential lazy-load would be a bit "bloaty".

fredericDelaporte:
The `ISession.GetSession(entityMode)` is still in 5.0 indeed, but obsolete. So I do not mention it in breaking changes.

hazzik:
>NH-3845 not included as it does not seem to be fully fixed, see #553.

The not fixed part has own JIRA issue now: [NH-3947](https://nhibernate.jira.com/browse/NH-3947).

hazzik:
>Test related items are identified as such, one has been dropped from the list (support of inconclusive results).

Why?

hazzik:
@fredericDelaporte I've updated several titles of issues, so it would need to update in the release notes. Also, some issues were moved from Tasks into Improvements category.

hazzik:
@fredericDelaporte I completely forgot about NH-3919. We have to make it into 5.0

doc/reference/modules/configuration.xml
```diff
@@ -115,7 +115,7 @@ Configuration cfg = new Configuration()
             obtain ADO.NET connections wherever it pleases:
         </para>
 
-        <programlisting><![CDATA[IDbConnection conn = myApp.GetOpenConnection();
+        <programlisting><![CDATA[DbConnection conn = myApp.GetOpenConnection();
```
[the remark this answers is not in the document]
fredericDelaporte: Why not. I have mixed feeling on that in documentation context: there is no intellisens there to check what type is the var. But in most cases it should not matter much.

fredericDelaporte:
> Why?

I was even considering dropping all tests only changes/fixes as they do not add or change features of the library. User could consider them as just noise for a release changes list. Still they give some more confidence in the library reliability. This one was not doing that. Now we can also let them all in the list: one more does not matter. Or maybe extract them to a separated list at the end (but that will cause more manual work for updating the list).

fredericDelaporte:
> @fredericDelaporte I completely forgot about [NH-3919](https://nhibernate.jira.com/browse/NH-3919). We have to make it into 5.0

Does not look as a small one, at least it implies to do some choices as to whether we re-align on hibernate (automatically switch to datetime 2 for sql server 2008+ instead of having to use a specific not portable type) or not. The automatic switch may cause a bunch of issues for column still using datetime, not datetime2.

If fixing that one we should fix [NH-3792 ](https://nhibernate.jira.com/browse/NH-3792 ) by the way, and maybe the two other linked issues to.

hazzik:
I think it is fine then. 

At least, I think, we need to replace DateTimeType with TimestampType, as there are a lot of confusion with cutting the milliseconds part. Because this is, potentially, a huge breaking change and we would need to postpone to 6.0 otherwise.

We will need to register Timestamp as a default type for DateTime (in TypeFactory.RegisterDefaultTypes) and change some of the dialect functions wrongly registered with DateTime, when in fact they have higher precision.

fredericDelaporte:
I may still give a try at [NH-3919](https://nhibernate.jira.com/browse/NH-3919), with checking what it gives with old datetime columns. Maybe [NH-3884](https://nhibernate.jira.com/browse/NH-3884) if implemented could give a way to revert back for users who would need it.

releasenotes.txt
```diff
@@ -3062,7 +3283,7 @@ Alpha Build 0.3.0.0
 - BatcherImpl and PreparerImpl were combined and code cleaned up thanks to problems found when using Ngpsql (Martijn Boland).
 - ITransaction is now responsible for joining IDbCommand to IDbTransaction instead of IBatcher - if applicable.
 - Modified code to help improve performance of Drivers that don't support multiple Open DataReaders on a single IDbConnection.
-- Fixed bug with hbm2net and VelocityRenderer throwing Exception (Carlos Guzm�n �lvarez & Peter Smulovics).
+- Fixed bug with hbm2net and VelocityRenderer throwing Exception (Carlos Guzm�n �lvarez & Peter Smulovics).
```
hazzik: Som unicode screw-ups

fredericDelaporte: VS Code gotcha: Microsoft/vscode#23997, they read a limited number of bytes before guessing the encoding. (At first only 512, they have increased that, but of course if they do not read the whole file they will still screw some cases...) Going to fix by switching the file to UTF8+BOM.

fredericDelaporte:
Rebased, squashed, corruption fixed, preventing it to occur again by re-encoding  to UTF8 with BOM.

doc/reference/modules/basic_mapping.xml
```diff
@@ -106,9 +106,8 @@
                      <area id="hm5" coords="6 55"/>
                    <area id="hm6" coords="7 55"/>
                    <area id="hm7" coords="8 55"/>
-                 </areaspec>   
-                 <programlisting>
-           <![CDATA[<hibernate-mapping
+                 </areaspec>
+                 <programlisting><![CDATA[<hibernate-mapping
```
fredericDelaporte: This is what wrecks numbering [here](http://nhibernate.info/doc/nhibernate-reference/mapping.html#mapping-declaration-mapping).

fredericDelaporte:
Fixes squashed.

hazzik:
@fredericDelaporte are you happy with the state of the release?

igitur:
Congratulations on this milestone.

hazzik:
thanks @igitur

PR #696, opened 2017-09-20
[WIP] NH-3990 - Reverts workaround in CompareGenerator for VB.

ngbrown:
This should finish up NH-3990, sorry to continue to linger on that issue...

It is Work-in-Progress because Remotion.Linq is using alpha build, waiting for final build.  It might be appropriate to slate this for v5.1.

To support VS2017 project for the VB project, a workaround was in place until Remotion.Linq was updated.

This updates Microsoft.NET.Test.Sdk for correct binding redirect generation. For dotnet SDK style projects that use NHibernate, they may need to do the same or add to their .csproj: `<AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>` and `<GenerateBindingRedirectsOutputType>true<GenerateBindingRedirectsOutputType>`.


ngbrown:
For more background and links to various discussions about the binding redirect issues in VS 2017 SDK style projects, see https://github.com/dotnet/sdk/issues/1595#issuecomment-330859618

fredericDelaporte:
Removing `Blocked` label since it does no more relies on a beta Remotion.Linq.

hazzik:
Something else broke.

ngbrown:
`Remotion.Linq` 2.2.0 now has support for Queryable.AsQueryable extension method.  See [RMLNQ-28](https://www.re-motion.org/jira/browse/RMLNQ-28).

Without implementing the `AsQueryableResultOperator.ISupportedByQueryModelVisitor` marker interface it seems like there should have been no change needed because the visitor shouldn't be called.  However, the tree has `query` `IASTNode` that `NHibernate.Hql.Ast.ANTLR.PolymorphicQuerySourceDetector.GetClassName` can't handle:

```
(( query ( select_from ( from ( range ( . category Products ) <generated>_1 ) ) ( select <generated>_1 ) ) ))
```

ngbrown:
This is ready for review now.  Both `Remotion.Linq` and `Remotion.Linq.EagerFetching` are updated to 2.2.0. 

NHibernate was directly accessing the `QueryModel` generated by Remotion Linq and not relying solely on visitors, so the `ResultOperatorRewriter` needed to support the new `AsQueryableResultOperator`.

hazzik:
Assigned to 5.2 because upgrading reference required to release a minor version.

cremor:
@ngbrown @hazzik Is it normal that a minor version update of ReLinq needs code changes in NHibernate? If yes, shouldn't the version contraint of the NuGet package be changed to "< 2.3.0"?

With 5.0.x and 5.1.x we now already have the situation that users of NHibernate that use the "old" style packages.config (so nearly everyone that doesn't target .NET Standard or .NET Core) are offered updates to ReLinq 2.2.0. So maybe it would also make sense to update the version constraint for the next patch releases of that versions?

fredericDelaporte:
Sometimes a minor version introduces new features which do not play well with old code emulating it for previous versions. Maybe it is this is the case with the `AsQueryable` support added in 2.2. (I have not actually analyzed why there are issues with NH5.1 used with ReLinq 2.2.)

Ideally it should not happen. We cannot predict whether this will happen again or not, be it with ReLinq or another dependency. So better stick to SemVer.

PR #697, opened 2017-09-21
Add Sql2008ClientDriver TeamCity configuration

hazzik:


fredericDelaporte:
Shouldn't we instead switch all `SqlClientDriver` to `Sql2008ClientDriver`? Having the old one prevents testing NH-1752, and what sens does it makes to test Sql2008+ with the old driver instead of its corresponding one?

hazzik:
I'm, actually, considering make `Sql2008ClientDriver` the default one, and the old one rename as `LegacySqlClientDriver`

fredericDelaporte:
#698 has switched all remaining `SqlClientDriver` configured usages in TeamCity to `Sql2008ClientDriver`. So the dedicated configuration for `Sql2008ClientDriver` should be removed now in my opinion, and its teamcity build too.

PR #698, opened 2017-09-21
NH-1752 - NHibernate Date type converts to NULL

fredericDelaporte:
[NH-1752](https://nhibernate.jira.com/browse/NH-1752) - NHibernate Date type converts to NULL

Occurs when date is lower than year 1753.

This behavior is removed. This behavior was handy when nullable types were lacking, and a test was still relying on this.

For testing the change with SqlServer, I have migrated all SqlClient driver to Sql2008Client.


doc/reference/modules/configuration.xml
```diff
@@ -1002,12 +1002,6 @@ ISession session = sessions.OpenSession(conn);
   <session-factory>
 
     <!-- properties -->
-    <property name="connection.provider">
-      NHibernate.Connection.DriverConnectionProvider
-      </property>
-    <property name="connection.driver_class">
-      NHibernate.Driver.SqlClientDriver
-    </property>
```
fredericDelaporte: Our doc examples incite users to use an obsolete driver for SqlServer: I have purged unneeded parameters.

src/NHibernate.Test/Async/NHSpecificTest/Logs/LogsFixture.cs
```diff
@@ -51,7 +51,7 @@ protected override string MappingsAssembly
 
 				await (s.GetAsync<Person>(1));//will execute some sql
 
-				var loggingEvent = spy.Events[0];
+				var loggingEvent = spy.GetWholeLog();
```
fredericDelaporte: @ngbrown, it seems you have forgotten to regen Async files.

ngbrown: Not sure what you mean.  I run the `ShowBuildMenu` option H, and no files change.  Also, there's no documentation at https://github.com/maca88/AsyncGenerator, so what does it even do?

ngbrown: Sorry, I was getting an error in the nuget restore, and didn't see that.

ngbrown: #699 if you want to make a separate merge for that.

fredericDelaporte: [Contributing](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md#regenerate-async-code) have been completed about that. But you now knows about it.

That is not a big change so I would not have required a new PR. I will see if Alexander will rather merge your PR first or just mine.

src/NHibernate/Driver/IDriver.cs
```diff
@@ -158,5 +158,10 @@ public interface IDriver
 		/// crash due to this, because they re-use the connection in the second phase.
 		/// </remarks>
 		bool HasDelayedDistributedTransactionCompletion { get; }
+
+		/// <summary>
+		/// The minimal date supported by this driver.
+		/// </summary>
+		DateTime MinDate { get; }
```
hazzik: What about other drivers?

The datetime range for [MySQL](https://dev.mysql.com/doc/refman/5.7/en/datetime.html), [Oracle](https://docs.oracle.com/cd/B19306_01/olap.102/b14346/dml_datatypes005.htm)  is 1000-01-01 to 9999-12-31

For [Postgres](https://www.postgresql.org/docs/current/static/datatype-datetime.html) is 4713 BC to 294276 AD

src/NHibernate.Test/TypesTest/DateTypeTest.cs
```diff
@@ -55,36 +56,30 @@ public void ReadWriteNormal()
...
+		{
+			var expected = Sfi.ConnectionProvider.Driver.MinDate;
+
+			ReadWrite(expected);
+		}
+
+		private void ReadWrite(DateTime expected)
 		{
-			var basic = new DateClass { DateValue = new DateTime(1899,1,1) };
+			var basic = new DateClass { DateValue = expected.AddHours(1) };
```
hazzik: Adding one hour does not make sense here as we only store date.

src/NHibernate.Test/TypesTest/DateTypeTest.cs
```diff
@@ -55,36 +56,30 @@ public void ReadWriteNormal()
 		{
 			var expected = DateTime.Today.Date;
```
hazzik: Do not need to call `.Date` on `.Today` as it is already a date

hazzik:
I don't understand how the tests work for MySQL and Oracle as their datetime range starts from year one thousand.

src/NHibernate/Driver/IDriver.cs
fredericDelaporte: I have just put .Net `MinValue` and let the test tell me if it was ok. So that is strange for MySql, and for Postgres, as .Net native type does not allow BC dates, it is limited by .Net rather than by the DB, but that still mean a limit at `MinValue`.

src/NHibernate.Test/TypesTest/DateTypeTest.cs
fredericDelaporte: Got screwed first too, I could have added a comment.

src/NHibernate.Test/TypesTest/DateTypeTest.cs
fredericDelaporte: Sure I could have cleaned that, just had refactored previous code in a common method without changing the logic.

src/NHibernate.Test/NHSpecificTest/NH2207/SampleTest.cs
```diff
@@ -9,10 +10,14 @@ namespace NHibernate.Test.NHSpecificTest.NH2207
...
 	{
+		protected override bool AppliesTo(Dialect.Dialect dialect)
+		{
+			return dialect is MsSql2008Dialect;
+		}
+
 		protected override bool AppliesTo(ISessionFactoryImplementor factory)
 		{
 			return factory.ConnectionProvider.Driver is Sql2008ClientDriver;
-
```
fredericDelaporte: Indeed, this one is checked after schema creation. So if the schema cannot be created for the test, we have to test the driver by checking `TestCase.Cfg` property in the dialect `AppliesTo`, somewhat like [here](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate.Test/Pagination/CustomDialectFixture.cs#L33).

It is coded as if creating the factory was requiring the schema to be there.
```C#
Configure();
if (!AppliesTo(Dialect))
{
	Assert.Ignore(GetType() + " does not apply to " + Dialect);
}

CreateSchema();
try
{
	_sessionFactory = BuildSessionFactory();
	if (!AppliesTo(_sessionFactory))
	{
		Assert.Ignore(GetType() + " does not apply with the current session-factory configuration");
	}
}
catch
{
	DropSchema();
	throw;
}
```

We could rewrite it:

```C#
Configure();
if (!AppliesTo(Dialect))
{
	Assert.Ignore(GetType() + " does not apply to " + Dialect);
}

_sessionFactory = BuildSessionFactory();
if (!AppliesTo(_sessionFactory))
{
	Assert.Ignore(GetType() + " does not apply with the current session-factory configuration");
}
CreateSchema();
```

src/NHibernate.Test/TypesTest/DateTypeTest.cs
```diff
@@ -55,36 +56,41 @@ public void ReadWriteNormal()
...
+			var expected = Sfi.ConnectionProvider.Driver.MinDate;
+
+			ReadWrite(expected);
+		}
+
+		[Test]
+		public void ReadWriteYear750()
+		{
+			var expected = new DateTime(750, 5, 13);
+			if (Sfi.ConnectionProvider.Driver.MinDate < expected)
```
fredericDelaporte: Shouldn't it be reversed?

src/NHibernate/Driver/IDriver.cs
fredericDelaporte: According to this [other doc](https://docs.oracle.com/database/122/SQLRF/Data-Types.htm#GUID-5405B652-C30E-4F4F-9D33-9A4CB2110F1B), Oracle `DATE` support date down to 4712 BC. So I think we can keep `MinDate` for Oracle.

About MySql, they also write:
> For the DATE and DATETIME range descriptions, “supported” means that although earlier values might work, there is no guarantee.

So it just happens to work in our case, but we should not complain if it ceases too. I will set its `MinDate` accordingly.

src/NHibernate.Test/TypesTest/DateTypeTest.cs
hazzik: Yes, you are correct

fredericDelaporte: I am going to do some changes, I will do it by the way.

fredericDelaporte:
May we merge this? It may allow me to do more clean-up in ongoing #703.

hazzik:
Sure. Can you also rebase/squash it before merging?

PR #702, opened 2017-09-22
NH-3884 - Default DateTime type can not be overridden

fredericDelaporte:
[NH-3884](https://nhibernate.jira.com/browse/NH-3884) (#844) - Default DateTime type can not be overridden

Here is a proposition for allowing overriding default clr type mapping to NHibernate type.

Such a method is quite low level, and for having a coherent session factory, it must be used before configuring and building the session factory. So I have put proper warnings in its comment.

Such a method could become handy if by doing [NH-3919](https://nhibernate.jira.com/browse/NH-3919) as proposed [here](https://github.com/nhibernate/nhibernate-core/pull/693#issuecomment-330808981), we cause issues to some users.

### DIFF
--- a/doc/reference/modules/basic_mapping.xml
+++ b/doc/reference/modules/basic_mapping.xml
@@ -2633,7 +2633,7 @@
                 types, <literal>System.Object</literal> types, and <literal>System.Object</literal> types for large objects.  Just like
                 Columns for System.ValueType types can handle <literal>null</literal> values only if the entity property is properly
                 typed with a <literal>Nullable&lt;T&gt;</literal>. Otherwise <literal>null</literal> will be replaced by the default
-                value for the type when reading, and when be overwritten by it when persisting the entity, potentially leading to
+                value for the type when reading, and then will be overwritten by it when persisting the entity, potentially leading to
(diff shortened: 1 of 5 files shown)


fredericDelaporte:
@hazzik, I think it would be better to merge this for v5.0 if ongoing #703 makes it too. This would give a way to revert the .Net `DateTime` default NHibernate type mapping semantic change for those who would need it.

hazzik:
I don't want to merge this at this stage. There is a really high chance that these methods would be made obsolete in a possible future type refactoring. I think what we can do now is to make some of RegisterType methods public (anyway these methods do the same as your SetDefaultType does).

fredericDelaporte:
The nearest is `RegisterType(System.Type systemType, IType nhibernateType, IEnumerable<string> aliases)` which takes an additional argument that will be unneeded for most users and also re-register the type alias (which should be unneeded for an override, but should not harm).

fredericDelaporte:
I will likely split that in many PR.

fredericDelaporte:
Rebased, the split in many PR and merge of the new PRs (#1483, #1484) having been done.

PR #703, opened 2017-09-24
NH-3919 - Clean up and harmonize datetime types with regards to different dialects

fredericDelaporte:
[NH-3919](https://nhibernate.jira.com/browse/NH-3919) - Clean up and harmonize datetime types with regards to different dialects.

Possible breaking changes:
- NHibernate type DateTimeType, which is the default for a .Net DateTime, does no more cut milliseconds. Use DateTimeNoMsType if you wish to have milliseconds cut. It applies to its Local/Utc counterparts too.
- LocalDateTimeType and UtcDateTimeType do no more accept being set with value having a non-matching kind, they throw instead.
- DbTimestamp will now round the retrieved value according to Dialect.TimestampResolutionInTicks.
- When an object typed property is mapped to a NHibernate timestamp, setting an invalid object in the property will now throw at flush instead of replacing it with DateTime.Now.
- SQL Server 2008+ dialects now use datetime2 instead of datetime for all date time types, including timestamp. This can be reverted with sql_types.keep_datetime setting.
- SQL Server 2008+ timestamp resolution is now 100ns in accordance with datetime2 capabilities, down from 10ms previously. This can be reverted with sql_types.keep_datetime setting.
- Oracle 9g+ dialects now use timestamp(7) for all date time types, instead of timestamp(4).
- Oracle 9g+ timestamp resolution is now 100ns in accordance with timestamp(7) capabilities, down from 100µs previously. 
- String parameter length will no more be specified by the OdbcDriver.
- IMapping interface has an additional Dialect member. ISessionFactoryImplementor has lost it, since it gains it back through IMapping.
- IDriver.ExpandQueryParameters and DriverBase.CloneParameter take an additional argument.


src/NHibernate.Test/NHSpecificTest/NH283/Fixture.cs
```diff
@@ -21,11 +21,11 @@ public void ForeignKeyNames()
 
 
 			string script = string.Join("\n",
-			                            cfg.GenerateSchemaCreationScript(new MsSql2000Dialect()));
+			                            cfg.GenerateSchemaCreationScript(new MsSql2008Dialect()));
```
fredericDelaporte: 2000 was failing due to "unknow type datetime2"...

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -1,26 +1,21 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../../build-common/NHibernate.props" />
-
```
fredericDelaporte: Should I revert those riders glitches, or just let it do that once and hopefully for all?

src/NHibernate/Dialect/MsSql2008Dialect.cs
```diff
@@ -36,5 +74,26 @@ protected override void RegisterDefaultProperties()
 			base.RegisterDefaultProperties();
 			DefaultProperties[Environment.ConnectionDriver] = typeof(Sql2008ClientDriver).AssemblyQualifiedName;
 		}
+
+		public override string CurrentTimestampSQLFunctionName =>
+			"SYSDATETIME()";
```
fredericDelaporte: Previously it was using [`CURRENT_TIMESTAMP`](https://docs.microsoft.com/en-us/sql/t-sql/functions/current-timestamp-transact-sql), calling indeed `getdate()`, so yielding `datetime` instead of `datetime2`.

src/NHibernate/Hql/Ast/ANTLR/Tree/IdentNode.cs
```diff
@@ -39,7 +39,7 @@ public override IType DataType
 					return fe.DataType;
 				}
 				ISQLFunction sf = Walker.SessionFactoryHelper.FindSQLFunction(Text);
-				return sf == null ? null : sf.ReturnType(null, null);
+				return sf?.ReturnType(null, Walker.SessionFactoryHelper.Factory);
```
fredericDelaporte: Unique case I have spotted where a methods chain of calls ending by calling `IType.SqlTypes` was not supplying an `IMapping` (the factory in this case).

src/NHibernate/Type/ClassMetaType.cs
```diff
@@ -16,7 +16,8 @@ public partial class ClassMetaType : AbstractType
 	{
 		public override SqlType[] SqlTypes(IMapping mapping)
 		{
-			return new SqlType[] { NHibernateUtil.String.SqlType };
+			var types = new[] { NHibernateUtil.String.SqlType };
+			return mapping?.Dialect.RefineSqlTypes(this, types) ?? types;
```
fredericDelaporte: I have added that logic for each type, excepted those yielding no types and those building them from other types.

src/NHibernate/Type/NullableType.cs
```diff
@@ -34,9 +34,9 @@ private IInternalLogger Log
 
 		/// <summary>
 		/// Initialize a new instance of the NullableType class using a 
-		/// <see cref="SqlType"/>. 
+		/// <see cref="NHibernate.SqlTypes.SqlType"/>.
```
fredericDelaporte: It was referencing the `NullabeType.SqType` property otherwise.

src/NHibernate/Type/NullableType.cs
```diff
@@ -291,9 +291,10 @@ public virtual SqlType SqlType
 		/// column.  All of their implementation should be in <see cref="NullableType.SqlType" />.
 		/// </para>
 		/// </remarks>
-		public override sealed SqlType[] SqlTypes(IMapping mapping)
+		public sealed override SqlType[] SqlTypes(IMapping mapping)
 		{
-			return new SqlType[] {SqlType};
+			var types = new[] { SqlType };
+			return mapping?.Dialect.RefineSqlTypes(this, types) ?? types;
```
fredericDelaporte: That the main change for supporting the feature: this allow switching the NHibernate type default corresponding sql type for another.

Hibernate does not do that. They do not need. Java does not know DbType.DateTime and DbType.DateTime2, it only knows `Types.TIMESTAMP`. Hibernate just need to tell this one should be mapped to sql type `datetime2` in SQL 2008 [dialect](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/SQLServer2008Dialect.java#L29).

src/NHibernate.Test/TypesTest/DateTimeTypeFixture.cs
```diff
@@ -37,21 +47,296 @@ public void DeepCopyNotNull()
...
+				s.CreateQuery("delete from DateTimeClass").ExecuteUpdate();
+				t.Commit();
+			}
+		}
+
+		[Test]
+		public void DbHasExpectedType()
+		{
+			var validator = new SchemaValidator(cfg);
+			validator.Validate();
```
fredericDelaporte: Delegating to `Validate` for checking we have the expected type in database. Which means being confident `Validate` correctly knows if it should be `DateTime` or `DateTime2`.
It appeared to me overly complicated to test this manually for all dialects, so I delegate to `Validate`. I have furthermore checked in a debug session directly in the db what type it was to ascertain we were now having datetime2 with SQL2008.

src/NHibernate.Test/TypesTest/DateTimeTypeFixture.cs
```diff
@@ -193,10 +196,10 @@ public void QueryUseExpectedSqlType()
 				t.Commit();
 			}
 
-			AssertSqlType(driver);
+			AssertSqlType(driver, 5);
 		}
 
-		private void AssertSqlType(ClientDriverWithParamsStats driver)
+		private void AssertSqlType(ClientDriverWithParamsStats driver, int expectedCount)
```
fredericDelaporte: Checking the exact usage count by the way.

src/NHibernate.Test/TypesTest/DateTimeTypeFixture.cs
```diff
@@ -167,12 +167,15 @@ public void UpdateUseExpectedSqlType()
...
 
-			AssertSqlType(driver);
+			AssertSqlType(driver, 3);
 		}
 
 		[Test]
 		public void QueryUseExpectedSqlType()
 		{
+			if (!TestDialect.SupportsNonDataBoundCondition)
+				Assert.Ignore("Dialect does not support the test query");
```
fredericDelaporte: Debuging the failure, the expected DbType and SqlType usages count is found (5), but Firebird does not accept the last condition. Removing it allows the query to be executed, but I want this condition for avoiding having the Hql parser determining the type for all parameters from the properties they are compared to, and instead use the type supplied by the parameter. So I ignore this test for Firebird.

src/NHibernate/Type/NullableType.cs
hazzik: They, actually, do something similar: https://github.com/hibernate/hibernate-orm/blob/9f2edb654dd55cee67d1b0445bdc0f6ae716f5e9/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java#L179

hazzik: They call it on nullSafeGet/nullSafeSet/extract(no equivalent on our side)

fredericDelaporte: Ok. It does not seems near to be ported currently, since it depends on their type descriptor which is not ported currently, and uses instance overrides for `doBind` which relies on members of the calling class, pattern not doable that way in .Net. (We could likely use method delegates instead.) And I have not checked if their override is taken into account also for hbm2ddl, as mine.

fredericDelaporte:
Last commit is part 2. I intend to add more tests though.

src/NHibernate.Test/Async/Operations/MergeFixture.cs
```diff
@@ -189,9 +190,21 @@ private void Cleanup()
 		{
 			ClearCounts();
 
-			var root = new Node {Name = "root"};
-			var child = new Node {Name = "child"};
-			var grandchild = new Node {Name = "grandchild"};
+			var root = new Node
+			{
+				Created = RoundForDialect(DateTime.Now),
```
fredericDelaporte: Just fixing the trouble in test. It could be fixed at the type level by overriding `IsEqual` (and preferably setter and getter too), but this would be a possible breaking change for `timestamp` when used a as simple property, because this will cause it to use the dialect declared accuracy instead of the actual database accuracy.

src/NHibernate.Test/NHSpecificTest/NH3567/Mappings.hbm.xml
```diff
@@ -6,23 +6,23 @@
     <id name="Id" type="Int32">
       <generator class="assigned" />
     </id>
-    <property name="Name" length="2000" />
+    <property name="Name" length="1999" />
```
fredericDelaporte: With below change to expand parameters, the size defined by OdbcDriver now actually ends in final parameters, and it causes issues for nvarchar(2000) and above: they get converted to NText by ODBC, causing:
```
ERROR [42000] [Microsoft][SQL Server Native Client 11.0][SQL Server]The data types nvarchar and ntext are incompatible in the equal to operator.
```

So here this is just a workaround for this test which was not about long string. But I fear there is a serious trouble with Odbc, and we should consider removing that size specification which was not actually taken into accont until below change.

src/NHibernate/Driver/DriverBase.cs
```diff
@@ -231,7 +231,7 @@ public void RemoveUnusedCommandParameters(DbCommand cmd, SqlString sqlString)
 				.ForEach(unusedParameterName => cmd.Parameters.RemoveAt(unusedParameterName));
 		}
 
-		public virtual void ExpandQueryParameters(DbCommand cmd, SqlString sqlString)
+		public virtual void ExpandQueryParameters(DbCommand cmd, SqlString sqlString, SqlType[] parameterTypes)
```
fredericDelaporte: This method in the case of ODBC was replacing the original parameters by new one, partially clone, leaving aside `Scale`, `Precision` and `Size` set by `OdbcDriver.InitializeParamater`. So I wonder if those changes in `InitializeParamater` were tested. It does not seem to be the case. When looking at [NH-3895](https://nhibernate.jira.com/browse/NH-3895) which supposed to be fixed by one of those changes, there are no test in the commit.

Due to `ExpandQueryParameters`, `Scale` is lost. Due to the change in `DateTimeType`, we now have milliseconds, and this has triggered the NH-3895 bug for numerous tests. So I have changed `ExpandQueryParameters` for it to create parameters the same way as original one, calling `InitializeParamater`. This fixes NH-3895. But it has triggered the trouble explained in previous comment.

src/NHibernate/AdoNet/Util/SqlStatementLogger.cs
```diff
@@ -104,11 +104,7 @@ public string GetCommandLineWithParameters(DbCommand command)
 
 		private static string GetParameterLoggableType(DbParameter dataParameter)
 		{
-			//TODO: Fix me after 4.6.2 update. Size and Precision has been added to DbParameter
-			var p = dataParameter as IDbDataParameter;
-			if (p != null)
-				return p.DbType + " (" + p.Size + ":" + p.Scale + ":" + p.Precision + ")";
```
fredericDelaporte: I have missed something or this todo was wrong on the version number? `Precision` was added in 4.5.1, and others are older.

src/NHibernate.Test/NHSpecificTest/NH3567/Mappings.hbm.xml
fredericDelaporte: I have searched about this trouble of obc switching to ntext for nvarchar >= 2000. Only to find someone confirming it [here](https://stackoverflow.com/q/8569844/1178314), but no one seem to have a clue about why is it like this.

src/NHibernate/Driver/DriverBase.cs
fredericDelaporte: Having dug into files history, the size setting has been added with 03eb58acae867e2d5620446c3b07cb91e7d74422 in January 2010, while the `ExpandQueryParameters` was not yet there. This one has been added in July 2010 with 131b97e9971608e5f90b31cd57e09b8d7935c4a8. So, not having Size/Precision/Scale actually specified for Odbc parameters (and all other drivers not handling named parameters - DB2, Ifx, Ingres, OleDb, OracelLite, SybaseAsa - if they do set them in `InitializeParameter`,) is an old regression.

fredericDelaporte: But later on [NH-3036](https://nhibernate.jira.com/browse/NH-3036) reverted the size setting for sql client due to issues with `like` (which may requires larger size than the actual column size), so I think we should remove this `Size` setting from Odbc parameters too.

src/NHibernate.Test/NHSpecificTest/NH3818/Fixture.cs
```diff
@@ -40,31 +38,33 @@ public void SelectConditionalValuesTest()
                         .Select(o => new
                         {
                             o.Color,
-                            AliveDays = (int)(DateTime.Now - o.Birthdate).TotalDays,
+                            AliveDays = (now - o.Birthdate).TotalDays,
```
fredericDelaporte: Test had start being flaky with `DateTimeType` no more truncating to seconds. The diff followed by `TotalDays` does not count boundaries but a fractional days count. We should not truncate it to int but better instead compare with a tolerance.

src/NHibernate.Test/TypesTest/DateTime2TypeFixture.cs
```diff
@@ -23,14 +24,14 @@ public void Next()
 		[Test]
 		public void Seed()
 		{
-			DateTimeType type = NHibernateUtil.DateTime;
+			var type = NHibernateUtil.DateTime2;
```
fredericDelaporte: Yuck. Was not testing what it was supposed to.

src/NHibernate.Test/TypesTest/DateTimeNoMsTypeFixture.cs
```diff
@@ -0,0 +1,73 @@
...
+using NHibernate.Util;
+using NUnit.Framework;
+
+namespace NHibernate.Test.TypesTest
+{
+	/// <summary>
+	/// TestFixtures for the <see cref="DateTimeNoMsType"/>.
+	/// </summary>
+	[TestFixture]
+	public class DateTimeNoMsTypeFixture
```
fredericDelaporte: In fact this is what was `DateTimeTypeFixture`.

src/NHibernate.Test/TypesTest/StringTypeWithLengthFixture.cs
```diff
@@ -145,164 +204,92 @@ private void AssertFailedInsertExceptionDetailsAndEmptyTable(Exception ex)
...
 
+				var aaItems =
+					s.CreateCriteria<StringClass>()
+					 .Add(Restrictions.Like("StringValue", "%AAAAAAAAA%"))
+					 .List();
 
-			// This test fails against the ODBC driver. The driver would need to be override to allow longer parameter sizes than the column.
-			AssertExpectedFailureOrNoException(
-				exception,
-				(Sfi.ConnectionProvider.Driver is OdbcDriver));
```
fredericDelaporte: Was no more failing since 131b97e9971608e5f90b31cd57e09b8d7935c4a8 which was losing the `Size` set on parameter. But since "expand parameter" is now fixed, it would fail again if we do not do NH-4083.

src/NHibernate/Type/AbstractDateTimeType.cs
```diff
@@ -0,0 +1,160 @@
...
+using NHibernate.Engine;
+using NHibernate.SqlTypes;
+
+namespace NHibernate.Type
+{
+	/// <summary>
+	/// Base class for date time types.
+	/// </summary>
+	[Serializable]
+	public abstract partial class AbstractDateTimeType : PrimitiveType, IIdentifierType, ILiteralType, IVersionType
```
fredericDelaporte: New base type for types yielding .Net `DateTime`, including `DateType`.

It includes the logic of the removed `AbstractDateTimeSpecificKindType`, which was applied only to type with kind `DateTimeKind.Local` or `DateTimeKind.Utc`, not to other types (non prefixed with `Local` or `Utc`). Thus here, we take into `Kind` only if it not `Unspecified`, its default value in this base class.

src/NHibernate/Type/AbstractDateTimeType.cs
```diff
@@ -0,0 +1,160 @@
...
+		/// </summary>
+		protected AbstractDateTimeType() : base(SqlTypeFactory.DateTime)
+		{
+		}
+
+		/// <summary>
+		/// Constructor for overriding the default <see cref="SqlType"/>.
+		/// </summary>
+		/// <param name="sqlTypeDateTime">The <see cref="SqlType"/> to use.</param>
+		protected AbstractDateTimeType(SqlType sqlTypeDateTime) : base(sqlTypeDateTime)
```
fredericDelaporte: Used by obsoleted `DateTime2Type` but also by `DateType`.

src/NHibernate/Type/AbstractDateTimeType.cs
```diff
@@ -0,0 +1,160 @@
...
+		{
+		}
+
+		/// <summary>
+		/// Adjust the date time value for this type from an arbitrary date time value.
+		/// </summary>
+		/// <param name="dateValue">The <see cref="System.DateTime" /> to adjust.</param>
+		/// <returns>A <see cref="System.DateTime" />.</returns>
+		protected virtual DateTime AdjustDateTime(DateTime dateValue) =>
+			Kind == DateTimeKind.Unspecified ? dateValue : DateTime.SpecifyKind(dateValue, Kind);
```
fredericDelaporte: From `AbstractDateTimeSpecificKindType`, but applied only when specified. This method `AdjustDateTime` take the purpose of `CreateDateTime` which was in `DateTimeType`.

src/NHibernate/Type/AbstractDateTimeType.cs
```diff
@@ -0,0 +1,160 @@
...
+		public override void Set(DbCommand st, object value, int index, ISessionImplementor session) =>
+			st.Parameters[index].Value = GetDateTimeToSet(value, session);
+
+		/// <summary>
+		/// Get the <see cref="DateTime" /> from the value to set in a parameter.
+		/// </summary>
+		/// <param name="value">The original value.</param>
+		/// <param name="session">The session for which the operation is done.</param>
+		/// <returns>The <see cref="DateTime" /> to set as parameter value.</returns>
+		protected virtual DateTime GetDateTimeToSet(object value, ISessionImplementor session) =>
```
fredericDelaporte: Overridden by `TimestampType` for its old logic replacing non datetime values by `Now`. Intended to replace `null` it seems, it was never doing that since `Set` is never called on `null` values. But it may replace invalid values set through a property typed `object` .

src/NHibernate/Type/AbstractDateTimeType.cs
```diff
@@ -0,0 +1,160 @@
...
+		/// <summary>
+		/// Compares two <see cref="DateTime" /> object and also compare its Kind if needed, which is not used by the
+		/// .Net Framework <see cref="DateTime.Equals(object)"/> implementation.
+		/// </summary>
+		/// <param name="x">The first date time to compare.</param>
+		/// <param name="y">The second date time to compare.</param>
+		/// <returns><see langword="true" /> if they are equals, <see langword="false" /> otherwise.</returns>
+		public override bool IsEqual(object x, object y) =>
+			base.IsEqual(x, y) &&
+				(Kind == DateTimeKind.Unspecified || x == null || ((DateTime) x).Kind == ((DateTime) y).Kind);
```
fredericDelaporte: `GetHashCode` is not overridden but that should not cause issues: this `IsEqual` further restrict the equality, meaning the contract will not be broken. Two equal object will still have the same hashcode. `GetHashCode` is overridden for thos date types only when the equality is relaxed, like in `DateType` or the new `DateTimeNoMsType` (old `DateTimeType`).

src/NHibernate/Type/DateTimeNoMsType.cs
```diff
@@ -0,0 +1,95 @@
...
+	/// </remarks>
+	[Serializable]
+	public partial class DateTimeNoMsType : AbstractDateTimeType
+	{
+		/// <inheritdoc />
+		public override string Name => "DateTimeNoMs";
+
+		/// <inheritdoc />
+		protected override DateTime AdjustDateTime(DateTime dateValue) =>
+			base.AdjustDateTime(new DateTime(dateValue.Year, dateValue.Month, dateValue.Day, dateValue.Hour, dateValue.Minute, dateValue.Second));
```
fredericDelaporte: Still calling `base.AdjustDateTime` for Utc/Local switch if required.

src/NHibernate/Type/TimestampType.cs
```diff
@@ -1,142 +1,50 @@
...
-using System.Data.Common;
 using NHibernate.Engine;
-using NHibernate.SqlTypes;
 
 namespace NHibernate.Type
 {
 	/// <summary>
-	/// This is almost the exact same type as the DateTime except it can be used
-	/// in the version column, stores it to the accuracy the database supports, 
-	/// and will default to the value of DateTime.Now if the value is null.
```
fredericDelaporte: `default to the value of DateTime.Now if the value is null`: wrong, since it was "implemented" into `Set` which is not called when the value is `null`.
Moreover since when this comment was written, it seems `DateTimeType` has gained the ability to serve as version too.

src/NHibernate/Type/TimestampType.cs
```diff
@@ -1,142 +1,50 @@
...
+	/// This is almost the exact same type as the <see cref="DateTimeType" /> excepted it can be used
+	/// in the version column, stores it to the accuracy the database supports
+	/// and will default to the value of <see cref="TimestampType.Now" /> if the value is not valid.
```
fredericDelaporte: We may remove that since I do not think it was intended, the aim for the code being to replace `null` values, but in a method which is never called on `null`. It coud be a breaking change for edge cases, where a `timestamp` is mapped to an `object` property in which the app may put non `DateTime` data...

I have left the accuracy part in case for some reason we were to reduce `DateTimeType` one again.

src/NHibernate/Type/TimestampType.cs
```diff
@@ -1,142 +1,50 @@
...
-			return Get(rs, rs.GetOrdinal(name), session);
-		}
-
-		public override System.Type ReturnedClass
-		{
-			get { return typeof(DateTime); }
-		}
+		/// <inheritdoc />
+		protected override DateTime GetDateTimeToSet(object value, ISessionImplementor session) =>
+			(value as DateTime?) ?? Now;
```
fredericDelaporte: This override should be removed if we want to clean the failed attempt at replacing null values by `Now`. By the way if keeping it, we should probably call `Seed` instead of `Now`.

src/NHibernate/Type/TimestampType.cs
```diff
@@ -1,142 +1,50 @@
...
-		{
-			get { return DateTime.MinValue; }
-		}
-
-		public override string ObjectToSQLString(object value, Dialect.Dialect dialect)
-		{
-			return '\'' + value.ToString() + '\'';
-		}
+		public override string ToString(object val) =>
+			((DateTime) val).ToString("O");
```
fredericDelaporte: Maybe the only significant change of `Timestamp` compared to the new `DateTimeType`.

src/NHibernate/Type/TimestampUtcType.cs
```diff
@@ -13,58 +38,16 @@ public class TimestampUtcType : TimestampType
...
-		/// </summary>
-		public override string ToString(object val)
+		/// <inheritdoc />
+		protected override DateTime GetDateTimeToSet(object value, ISessionImplementor session)
 		{
-			return ((DateTime) val).ToString("o");
+			var dbValue = base.GetDateTimeToSet(value, session);
+			if (dbValue.Kind != DateTimeKind.Utc)
+				throw new ArgumentException("Kind is NOT Utc", nameof(value));
+			return dbValue;
```
fredericDelaporte: With `Timestamp.ToString`, that is all the differences with `UtcDateTimeType`. `UtcDateTimeType` does not throw if it receives a non UTC value: it silently switch its kind to UTC.

src/NHibernate/Cfg/Environment.cs
```diff
@@ -208,6 +208,14 @@ public static string Version
 		/// </summary>
 		public const string OdbcDateTimeScale = "odbc.explicit_datetime_scale";
 
+		// Obsolete Since v5.0
+		/// <summary>
+		/// Disable switching built-in NHibernate date-time types from DbType.DateTime to DbType.DateTime2
+		/// for dialects supporting datetime2.
+		/// </summary>
+		[Obsolete("Migrate SQL datetime type to datetime2 instead.")]
+		public const string SqlTypesKeepDateTime = "sql_types.keep_datetime";
```
fredericDelaporte: `DateTime2` being unsupported by OdbcDriver, we may use it for allowing using 2008 dialect with Odbc. So I think I will remove the obsolete attribute.

The subject is a bit deeper since Odbc does not understand `Time` either, and likely `Date` too: we should have a mean to switch those types too. So we can also decide that only dialects up to 2005 are compatible with ODBC.
(By the way this [page](https://docs.microsoft.com/en-us/sql/relational-databases/native-client-odbc-date-time/data-type-support-for-odbc-date-and-time-improvements) states ODBC does support SQL2008 types, but how are we supposed to do in a .Net context, mystery.)

An other option would be to try moving the type override from dialect to driver. But currently `IMapping` interface which is where I pull the dialect from for refining types is not much suitable for having the driver, unless we add the ability for `Configuration.Mapping` to instantiate a driver.

I think I will just remove the `Obsolete` attribute.

fredericDelaporte:
All date time types are now tested. One more bug to fix by the way.
I intend to test what it gives using timestamp as version on a SQL Server db having datetime columns instead of datetime2: it will surely cause stale update exceptions, which will be the main breaking change for this switch from datetime to datetime2.

hazzik:
Such a Pandora's box.

fredericDelaporte:
If it is not 5.0 it should be 6, too much changes for a minor.

hazzik:
>If it is not 5.0 it should be 6, too much changes for a minor.

I think this is better to leave it for 6. A good exercise for a types refactoring. At least much better than trying to rush in 5.0

fredericDelaporte:
I am not that rushing anyway. I take time to check untested cases, refactor a bit more than what was asked in NH-3919, ...

About the Local/Utc types, I would like your view on a point: the non-timestamp ones do "abruptly" change incoming dates (dates to set in the db) kind if it is not the "right" one, using `DateTime.SpecifyKind()`. Good for me for `Unspecified`, we can assume it is already a time matching the expected kind. But when an `LocalDateTime` property is set with an UTC date, switching it to local without adjusting its time part looks terribly as a data corruption. I would tend to change that for converting instead. Same when an utc property is set with a local one.
(`UtcTimestamp` does not have the trouble, it throws if it receives a non Utc date, forcing to always set such properties with Utc dates. Maybe not very convenient though.)

src/NHibernate/Type/DbTimestampType.cs
```diff
@@ -46,47 +45,55 @@ public override object Seed(ISessionImplementor session)
...
+		protected virtual DateTime GetCurrentTimestamp(ISessionImplementor session)
 		{
-			Dialect.Dialect dialect = session.Factory.Dialect;
-			string timestampSelectString = dialect.CurrentTimestampSelectString;
-			return UsePreparedStatement(timestampSelectString, session);
+			var dialect = session.Factory.Dialect;
+			// Need to round notably for Sql Server DateTime with Odbc, which has a 3.33ms resolution,
+			// causing stale data update failure 2/3 of times if not rounded to 10ms.
+			return Round(
+				UsePreparedStatement(dialect.CurrentTimestampSelectString, session),
```
fredericDelaporte: In fact I am fixing there the trouble which was causing Odbc to be ignored for NH-1756 tests and DbVersionFixture. Going to re-enable them.

src/NHibernate.Test/Async/TypesTest/TimestampTypeFixture.cs
```diff
@@ -21,4 +24,42 @@ public class TimestampTypeFixtureAsync : AbstractDateTimeTypeFixtureAsync
 		protected override string TypeName => "Timestamp";
 		protected override AbstractDateTimeType Type => NHibernateUtil.Timestamp;
 	}
+
+	// Testing SQL Server 2008 with datetime in db instead of datetime2
+	[TestFixture]
+	public class TimestampTypeMixedDateTimeFixtureAsync : TimestampTypeFixtureAsync
+	{
```
fredericDelaporte: No troubles if using `SqlTypesKeepDateTime`.

src/NHibernate.Test/Async/TypesTest/TimestampTypeFixture.cs
```diff
@@ -21,4 +24,42 @@ public class TimestampTypeFixtureAsync : AbstractDateTimeTypeFixtureAsync
...
+		{
+			base.Configure(configuration);
+
+			configuration.SetProperty(Environment.SqlTypesKeepDateTime, "true");
+		}
+
+		/* Uncomment for testing how it fails if NHibernate uses DateTime2 while the db column is DateTime.
+		 * It fails two thirds of times on update with an undue optimistic locking exception, due to the datetimes
+		 * ending with 3.333ms (transmitted as 3) or 6.666ms (transmitted as 7), which does not match the datetime2
+		 * sent by NHibernate for checking the revision has not changed. Demonstrated by UpdateUseExpectedSqlType.
```
fredericDelaporte: If using a db with `datetime` but letting NHibernate use `datetime2`, versioning is broken. `SqlTypesKeepDateTime` must be used for supporting datetime based versioning. (Or better of course, migrate the db to `datetime2`.)

doc/reference/modules/basic_mapping.xml
```diff
@@ -2750,15 +2789,22 @@
                         <row>
                             <entry><literal>Timestamp</literal></entry>
                             <entry><literal>System.DateTime</literal></entry>
-                            <entry><literal>DbType.DateTime</literal> - as specific as database supports.</entry>
```
fredericDelaporte: > as specific as database supports

This was a lie. It just does not cut milliseconds as the old `DateTime` was doing. But by example with Oracle, it will use a `timestamp(4)` while Oracle supports `timestamp(9)` (at least since 10g).

doc/reference/modules/basic_mapping.xml
```diff
@@ -2656,9 +2673,14 @@
...
-                            <entry><literal>type="DbTimestamp"</literal> must be specified.  When used as a <literal>version</literal>
-                            field, uses the database's current time rather than the client's current time.</entry>
+                            <entry>
+                                <literal>DbType.DateTime</literal> / <literal>DbType.DateTime2</literal><coref linkend="basic_mapping.datetime-co" />
+                            </entry>
+                            <entry>
+                                <literal>type="DbTimestamp"</literal> must be specified.  When used as a
+                                <literal>version</literal> field, uses the database's current time retrieved
+                                in dedicated queries, rather than the client's current time.
```
fredericDelaporte: Further hinting at the performance consequences it may have by mentioning it does that in dedicated queries.

doc/reference/modules/basic_mapping.xml
```diff
@@ -2750,15 +2789,22 @@
...
-                            <entry><literal>DbType.DateTime</literal> - as specific as database supports.</entry>
-                            <entry><literal>type="TimestampUtc"</literal> must be specified.  Ensures the
-                            <literal>DateTimeKind</literal> is set to <literal>DateTimeKind.Utc</literal></entry>
+                            <entry>
+                                <literal>DbType.DateTime</literal> / <literal>DbType.DateTime2</literal><coref linkend="basic_mapping.datetime-co" />
+                            </entry>
+                            <entry>
+                                <literal>type="TimestampUtc"</literal> must be specified. Ensures the
+                                <literal>DateTimeKind</literal> is set to <literal>DateTimeKind.Utc</literal>.
+                                Throws if setted with another kind.
```
fredericDelaporte: Documenting its throwing behavior by the way.

src/NHibernate/NHibernateUtil.cs
```diff
@@ -33,10 +33,11 @@ static NHibernateUtil()
 				clrTypeToNHibernateType[type.ReturnedClass] = type;
 			}
 
-			// There are multiple possibilites for boolean and strings.
+			// There are multiple possibilites for boolean, strings and datetime.
 			// Override so that we use the most natural mapping.
 			clrTypeToNHibernateType[Boolean.ReturnedClass] = Boolean;
 			clrTypeToNHibernateType[String.ReturnedClass] = String;
+			clrTypeToNHibernateType[DateTime.ReturnedClass] = DateTime;
```
fredericDelaporte: In case #702 does not get merge, at least fix NHibernateUtil.GuesType yielding a random type for `DateTime`.

hazzik:
Do I understand correctly, that following types are equal now?

DateTimeType <=> TimestampType <= DateTime2Type
UtcDateTimeType <=> TimestampUtcType

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -292,6 +292,14 @@ protected void RegisterColumnType(DbType code, string name)
 			_typeNames.Put(code, name);
 		}
 
+		/// <summary>
+		/// Refine provided <see cref="SqlType"/>s.
+		/// </summary>
+		/// <param name="types">The original <see cref="SqlType"/>s.</param>
+		/// <returns>Refined <see cref="SqlType"/>s.</returns>
+		public virtual SqlType[] RefineSqlTypes(SqlType[] types) =>
```
hazzik: I would like this method to accept one type, not array.

src/NHibernate/Dialect/MsSql2008Dialect.cs
```diff
@@ -36,5 +68,27 @@ protected override void RegisterDefaultProperties()
...
+		public override SqlType[] RefineSqlTypes(SqlType[] types)
+		{
+			types = base.RefineSqlTypes(types);
+
+			if (KeepDateTime)
+				return types;
+
+			return types
+				.Select(t => SqlTypeFactory.DateTime.Equals(t) ? SqlTypeFactory.DateTime2 : t)
+				.ToArray();
```
hazzik: This needs to be [`Array.ConvertAll`](https://msdn.microsoft.com/en-us/library/exc45z53)

src/NHibernate/Type/UtcDateTimeType.cs
```diff
@@ -3,16 +3,22 @@
...
-			get { return "UtcDateTime"; }
-		}
+		/// <inheritdoc />
+		public override string Name => "UtcDateTime";
 	}
-}
\ No newline at end of file
+
+	[Serializable]
+	public class UtcDateTimeNoMsType : DateTimeNoMsType
```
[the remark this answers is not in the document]
fredericDelaporte: Not sure, at least they allow reverting back to old behavior. Oskar seems to consider it has some advantages, reading NH-3919. They also perfectly the Oracle {{date}} type.

src/NHibernate.Test/DialectTest/MsSql2008DialectFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+			AssertSqlType(NHibernateUtil.DbTimestamp, SqlTypeFactory.DateTime, mapping);
+			AssertSqlType(NHibernateUtil.LocalDateTime, SqlTypeFactory.DateTime, mapping);
+			AssertSqlType(NHibernateUtil.UtcDateTime, SqlTypeFactory.DateTime, mapping);
+#pragma warning disable 618 // DateTime2 is obsolete
+			AssertSqlType(NHibernateUtil.DateTime2, SqlTypeFactory.DateTime2, mapping);
+#pragma warning restore 618
+		}
+
+		private static readonly FieldInfo _mappingField =
+			typeof(Configuration).GetField("mapping", BindingFlags.Instance | BindingFlags.NonPublic);
```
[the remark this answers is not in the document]
hazzik: Apparently, it is not:(

src/NHibernate/Type/TypeFactory.cs
```diff
@@ -259,7 +259,8 @@ private static void RegisterBuiltInTypes()
 			
 			RegisterType(NHibernateUtil.StringClob, EmptyAliases,
 						 l => GetType(NHibernateUtil.StringClob, l, len => new StringClobType(SqlTypeFactory.GetStringClob(len))));
-			
+
+			RegisterType(NHibernateUtil.DateTimeNoMs, new[] { "datetimenoms" });
 			RegisterType(NHibernateUtil.Date, new[] { "date" });
 			RegisterType(NHibernateUtil.Timestamp, new[] { "timestamp" });
```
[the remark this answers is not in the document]
fredericDelaporte: No, it is [there](https://github.com/fredericDelaporte/nhibernate-core/blob/e168e694d7270a13a31756c3dabb302f9473d0f6/src/NHibernate/Type/TypeFactory.cs#L202), in the other method, the one for default types.

hazzik:
@fredericDelaporte  Do I understand correctly, that following types are equal now?

`DateTimeType` <=> `TimestampType` <= `DateTime2Type`
`UtcDateTimeType` <=> `TimestampUtcType`

I think we need to do following:
- Obsolete `TimestampType` and `DateTime2Type` in favor of `DateTimeType`
- Obsolete newly created NoMs compatibility types
- Remove `TimestampUtcType` as it was never released, and close the issue as obsoleted by NH-3919
- Add log warning for mappings resolution for obsolete types
- Change Oracle's DbType.DateTime binding to maximum precision supported by .NET's DateTime type (instead of `TIMESTAMP(4)`).
- (Not sure about this) add support for specifying length (precision? scale?) for date-time based types

fredericDelaporte:
> Do I understand correctly, that following types are equal now?
>
> DateTimeType <=> TimestampType <= DateTime2Type
> UtcDateTimeType <=> TimestampUtcType

As said in a review above, `Timestamp` differs now from `DateTime` only by its `ToString` method, which use iso format instead of a more "natural" one. `TimestampUtc` differs furthermore by only accepting being set with UTC datetime, throwing otherwise. `UtcDateTime` does not throw and instead takes all as Utc, disregarding the actual kind (which looks to me as a bit data corruption as said in a previous review, proposing to convert instead if kind is local).

`DateTime2` is identical to `DateTime` only if backward compatibility setting `keep_datetime` is not set to true. Otherwise, it stays mapped to `DbType.DateTime2` while `DateTime` is reverted to `DbType.DateTime`.

If we remove `Timestamp` and `TimestampUtc`,  we then at least need to map their aliases back to `DateTime` and `UtcDateTime`, for smoothing the upgrade. Same for `DateTime2`, but that would mean that those having "hybrid" database using both `DateTime` and `DateTime2` will no more be able to map both with their exact `DbType`. In case of use of both as version, they will be stuck.

About limiting the override ability only to `NullableType`, why not. But custom types can still benefit of it, we just do not know. My implementation was giving the choice to implementors to opt in or out of the override, through the `AllowSqlTypeOverride` property. With your change they will have to add themselves the call to dialect override if they need it. Not a big deal but slightly less convenient I think.

fredericDelaporte:
> - Obsolete `TimestampType` and `DateTime2Type` in favor of `DateTimeType`
> - Obsolete newly created NoMs compatibility types
> - Remove `TimestampUtcType` as it was never released, and close the issue as obsoleted by NH-3919
> - Add log warning for mappings resolution for obsolete types

Ok. If removing them in 6.0, we should probably still think about mapping their old alias to the replacing types, for convenience of users used to Hibernate.

> - Change Oracle's DbType.DateTime binding to maximum precision supported by .NET's DateTime type (instead of `TIMESTAMP(4)`).

Sounds as timestamp(7). (100ns, the datetime tick.) Supported at least since Oracle 9.

> - (Not sure about this) add support for specifying length (precision? scale?) for date-time based types

I do not know what it implies to code that. Would be good to have it, but maybe better to do that in another PR.

hazzik:
>As said in a review above, Timestamp differs now from DateTime only by its ToString method, which use iso format instead of a more "natural" one. 

Not essential, so, I think we need to make them identical (and replace the one with other. I would prefer DateTimeType to survive)

>TimestampUtc differs furthermore by only accepting being set with UTC datetime, throwing otherwise. UtcDateTime does not throw and instead takes all as Utc, disregarding the actual kind (which looks to me as a bit data corruption as said in a previous review, proposing to convert instead if kind is local).

Need to unify semantic, otherwise it's confusing. I would prefer either options where data loss does not occur (eg. throwing. Or converting known kinds to a target kind, meaning throwing if kind is `Unspecified`)

>DateTime2 is identical to DateTime only if backward compatibility setting keep_datetime is not set to true. Otherwise, it stays mapped to DbType.DateTime2 while DateTime is reverted to DbType.DateTime.

But this is not what I see in code (?). As I understand the code - it is promoting DateTime to DateTime2 if `KeepDateTime == false`

>If we remove Timestamp and TimestampUtc, we then at least need to map their aliases back to DateTime and UtcDateTime, for smoothing the upgrade. 
>Ok. If removing them in 6.0, we should probably still think about mapping their old alias to the replacing types, for convenience of users used to Hibernate.

Yes, the Timestamp's aliases needs to be remapped to DateTimeType in 6.0. We need to be clever here, as for example `"NHibernate.Type.TimestamptType"` should render a warning in a log, but `"timestamp"` shall not. the TimestampUtc needs to be killed, as it has never made into a released version.

> for DateTime2, but that would mean that those having "hybrid" database using both DateTime and DateTime2 will no more be able to map both with their exact DbType. In case of use of both as version, they will be stuck.

I think we need to leave the ability to specify a SqlType in a `DateTimeType` (through a constructor argument and/or TypeFactory)

>I do not know what it implies to code that. Would be good to have it, but maybe better to do that in another PR.

I believe simply register as `"TIMESPAN($l)"` (same as for strings, byte arrays and decimals) in addition to the default precision. Don't mind doing this in another PR/after 5.0

>But custom types can still benefit of it, we just do not know. 

As I said in a commit message - the user shall care about themselves, we do not want to mess with their intentions. So, I would favor the explicit types for IUserType.

fredericDelaporte:
> But this is not what I see in code (?). As I understand the code - it is promoting DateTime to DateTime2 if KeepDateTime == false

`DateTime2Type` is hard-coded to always use `DbType.DateTime2` (constructor), whatever the value of `KeepDateTime`. So if `KeepDateTime` is `true`, `DateTime2Type` will still use `DbType.DateTime2` while `DateTimeType` will be reverted to `DbType.DateTime`. 

> I think we need to leave the ability to specify a SqlType in a DateTimeType (through a constructor argument and/or TypeFactory)

Why not, but since `DateTime2Type` is released since quite some time now, many users probably already use it, so better jut obsolete it as this PR already does. And furthermore if the user explicitly specifies the `DbType` for a `DateTimeType`, it should then no more be overridden, meaning we need to put back in place `AllowSqlTypeOverride` and set it to `false` when the type is specified.

 > As I said in a commit message - the user shall care about themselves, we do not want to mess with their intentions. So, I would favor the explicit types for IUserType.

Yes I have seen the commit message, and yes my code was not interfering with their intention: `AllowSqlTypeOverride` was `false` by default, meaning no override unless users change it themselves. But as I said that was just a small added convenience, they can still one line re-code it themselves if needed.

src/NHibernate/Type/AbstractDateTimeType.cs
```diff
@@ -91,8 +92,38 @@ protected virtual DateTime GetDateTime(DbDataReader rs, int index, ISessionImple
...
+				if (Kind == DateTimeKind.Utc)
+				{
+					if (TimeZoneInfo.Local.IsInvalidTime(dateValue))
+						throw new ArgumentException(
+							$"Date is local and has an invalid time ({dateValue:o}), unable to convert to UTC.",
+							nameof(value));
+					if (TimeZoneInfo.Local.IsAmbiguousTime(dateValue))
+						throw new ArgumentException(
+							$"Date is local and has an ambiguous time ({dateValue:o}), unable to convert to UTC.",
+							nameof(value));
```
fredericDelaporte: Conversion from Local to Utc may cause issue when daylight savings change hour, going back one hour. Having code failing one hour per year is not really nice and easy to test, so maybe should we just not support converting from local to utc, and only support converting from utc to local (which can always be done). Or not support any conversion at all.

hazzik: 👍 for not doing any conversion

fredericDelaporte:
Rebased, scale specification added but not yet tested (only checked non-regression locally).

> I believe simply register as "TIMESPAN($l)" (same as for strings, byte arrays and decimals) in addition to the default precision.

`TypeFactory` requires changes for handling new type having options. And it is not a length but a scale, alone without precision. This case was not handled.

I will add tests later.

fredericDelaporte:
> I think we need to leave the ability to specify a SqlType in a DateTimeType (through a constructor argument and/or TypeFactory)

That would mean relaxing the last commit added `DateTimeType`/`LocalDateTimeType`/`UtcDateTimeType` constructor with type for accepting any `SqlType` (or adding an overload taking `DateTime2SqlType`).
It appears to me constructors are not supposed to be used directly, some types have them internal, and they are usable only through `TypeFactory`. That was not the case of date time types already existing constructor so I have left them public. But maybe should we instead switch all types constructor to internal, breaking change telling to go through `TypeFactory`.
So we should always add adequate methods in `TypeFactory`.
But I wonder how a constructor allowing many types not only differing by length/precision/scale is supposed to be useful when specifying a type in mappings: I do not think there is support for this. So it does not look to me as an easy pattern to put in place

fredericDelaporte:
Scale tests added.

src/NHibernate/Type/TimestampUtcType.cs
```diff
@@ -1,70 +0,0 @@
-﻿using System;
-using System.Data.Common;
-using NHibernate.Engine;
-
-namespace NHibernate.Type
-{
-	[Serializable]
-	public class TimestampUtcType : TimestampType
```
hazzik: Need to remove from mapping schema

fredericDelaporte: Had not thought it was even added as a specif versioning element shortcut for `<version type="timestamputc"`

src/NHibernate/Cfg/MappingSchema/Hbm.generated.cs
```diff
@@ -2,7 +2,7 @@ namespace NHibernate.Cfg.MappingSchema {
     
     
     /// <remarks/>
-    [System.CodeDom.Compiler.GeneratedCodeAttribute("HbmXsd", "5.0.0.Alpha1")]
+    [System.CodeDom.Compiler.GeneratedCodeAttribute("HbmXsd", "5.0.0-Alpha1")]
```
fredericDelaporte: Mass change due to the NHibernate versioning change.

Going to squash the fixup before TC starts building.

src/NHibernate/Type/TimestampUtcType.cs
hazzik: Maybe remove it in a separate commit?

src/NHibernate/Dialect/PostgreSQL81Dialect.cs
```diff
@@ -34,6 +33,13 @@ namespace NHibernate.Dialect
 	/// <seealso cref="PostgreSQLDialect" />
 	public class PostgreSQL81Dialect : PostgreSQLDialect
 	{
+		protected override void RegisterDateTimeTypeMappings()
+		{
+			base.RegisterDateTimeTypeMappings();
+			RegisterColumnType(DbType.DateTime, 6, "timestamp($s)");
```
hazzik: What will be the resolution of `$s=7`?

fredericDelaporte: Not handled by postgresql.

hazzik: But will it resolve to just `timestamp`?

fredericDelaporte: No idea, missing test.

src/NHibernate/Type/TimestampUtcType.cs
fredericDelaporte: Remove the whole timestamputc stuff with a commit like "[NH-2520](https://nhibernate.jira.com/browse/NH-2520) - Revert TimestampUtc in favor of UtcDateTime"

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2123,6 +2131,11 @@ public virtual bool SupportsSubSelects
 		/// </remarks>
 		public virtual bool SupportsDistributedTransactions => true;
 
+		/// <summary>
+		/// Does this dialect handles date and time types scale (fractional seconds precision)?
+		/// </summary>
+		public virtual bool SupportsDateTimeScale => false;
```
hazzik: Is it only for tests?

fredericDelaporte: Yes.

src/NHibernate/Dialect/PostgreSQL81Dialect.cs
fredericDelaporte: Still sent in creation script as `timestamp(7)`, PostgreSql switches it by itself to `timestamp(6)`. Other db may yield an error instead.

The 6 in `RegisterColumnType(DbType.DateTime, 6, "timestamp($s)")` in facts works only for size/length, it has no effect on scale. There are no mechanism for this currently.

It could be easily adapted by considering [here](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Dialect/TypeNames.cs#L86) that for `DbType.DateTime/DateTime2`, the size does not matter and we should consider the scale. The current implementation would then yield `timestamp` for column creation. Db compliant to the SQL standard should then yield 6 (according to [MySql doc](https://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html)). But MySql would yield 0, Sql Server 7, Postgres writes "no explicit bound" (what does it mean? May be more than 6 in such case?) and Oracle 6.

So if adapting the code in `TypeNames`, maybe should we default to explicit max scale instead of the bare type without scale.

src/NHibernate/Type/TimestampUtcType.cs
hazzik: yes

src/NHibernate/Dialect/PostgreSQL81Dialect.cs
fredericDelaporte: > The current implementation would then yield timestamp for column creation.

That what is doing Postgres dialect because I have forgotten to override default type with `timestamp(6)` instead. So no needs for complicating things, just need to handle the scale types case, yielding default if above scale, and taking care of having default suitably defined to max scale in dialect.

fredericDelaporte: Done.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -274,7 +274,7 @@ public virtual string GetCastTypeName(SqlType sqlType)
 		/// length (if appropriate)
 		/// </summary>
 		/// <param name="code">The typecode</param>
-		/// <param name="capacity">Maximum length of database type</param>
+		/// <param name="capacity">Maximum length or scale of database type</param>
 		/// <param name="name">The database type name</param>
 		protected void RegisterColumnType(DbType code, int capacity, string name)
```
fredericDelaporte: In fact it appears `capacity` is also used for transmitting max precision for `decimal` types, but the retrieval code does not switch to precision for it. Since the retrieval code would switch to default in case of over capacity, which in the case of `decimal` is far from the maximum, that seems better it does not, although not very satisfactory.

Moreover, current `capacity` values transmitted for decimal are currently plain wrong, at least for sql server: `19`, which neither match its maximal precision `38` (at least since 2000, although some current docs write without much precision that it was `28` in some earlier version) nor its maximal size `17`.

src/NHibernate/Dialect/MySQL57Dialect.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace NHibernate.Dialect
+{
+	public class MySQL57Dialect : MySQL55Dialect
+	{
+		public MySQL57Dialect()
+		{
+			RegisterColumnType(DbType.DateTime, "DATETIME(6)");
+			RegisterColumnType(DbType.DateTime, 6, "DATETIME($s)");
+			RegisterColumnType(DbType.Time, "TIME(6)");
+			RegisterColumnType(DbType.Time, 6, "TIME($s)");
```
fredericDelaporte: Has now minimal tests checking expected sql string is yielded for those types.

src/NHibernate/Dialect/TypeNames.cs
```diff
@@ -72,44 +72,65 @@ public string Get(DbType typecode)
...
 		/// </returns>
 		public string Get(DbType typecode, int size, int precision, int scale)
 		{
-			SortedList<int, string> map;
-			weighted.TryGetValue(typecode, out map);
+			weighted.TryGetValue(typecode, out var map);
 			if (map != null && map.Count > 0)
 			{
-				foreach (KeyValuePair<int, string> entry in map)
+				var requiredCapacity = IsScaleType(typecode) ? scale : size;
```
fredericDelaporte: Now handling types for which capacity means scale, not size.
Something similar could be done for precision, but it would requires to fix the dialect current decimal registrations, which do not appear to transmit valid capacity for `decimal`. Moreover, the default for decimal would not suit when a "too big" capacity is asked, because this default has usually not the maximal precision. It would then be required to have some additional logic here for yielding the maximal actual precision here. But it would have to account scale too, and that would be quite messy. (Should it reduces the scale as much as precision is reduced? Or just trims it to max too? Or something in between?)

src/NHibernate/Mapping/ForeignKey.cs
```diff
@@ -123,12 +123,22 @@ private void AlignColumns(Table referencedTable)
...
+		{
+			using (var fkCols = fk.GetEnumerator())
+			using (var pkCols = pk.GetEnumerator())
 			{
 				while (fkCols.MoveNext() && pkCols.MoveNext())
 				{
-					fkCols.Current.Length = pkCols.Current.Length;
+					if (pkCols.Current.IsLengthDefined() || fkCols.Current.IsLengthDefined())
+						fkCols.Current.Length = pkCols.Current.Length;
+					if (pkCols.Current.IsPrecisionDefined() || fkCols.Current.IsPrecisionDefined())
```
fredericDelaporte: Previous column alignment logic between fk and pk was defining a length on fk even if it was not defined on pk, because those properties have default values (`255` for length). Now defining it only if it is defined on one side. I guess that if the case "defined on fk but not on pk" is possible, it will still fail, and it would be better to have a way to un-define it on fk. I have just made minimal changes here for at least ceasing specifying bad length on fk when it is actually not defined on pk.

And I have added alignement for scale and precision too.

Previous code was causing a bunch of tests to fail since max scale handling: the defined length was causing the column to convert `datetime` type to sql type through the method taking into account length/precision/scale, with default values for them (255 for length, 2 for scale). This was not having consequences when the capacity was only considered to be for size, since 255 was the out of range of datetime max scale, causing it to still use datetime default. But with the handling of scale as capacity instead, this logic was then causing fk to be generated with `datetime2(2)` sql type instead of `datetime2`.

src/NHibernate/Dialect/TypeNames.cs
```diff
@@ -72,44 +72,63 @@ public string Get(DbType typecode)
...
+		{
+			switch (typecode)
+			{
+				case DbType.DateTime:
+				case DbType.DateTime2:
+				case DbType.DateTimeOffset:
+				case DbType.Time:
+					return true;
+			}
+			return false;
```
fredericDelaporte: Right, I have been a bit quick with the "refactor extract method here".

hazzik:
I just realized that "NoMs" types are not needed anymore.

fredericDelaporte:
Why, because we have scale? Scale is not understood and handled by every database.

PR #705, opened 2017-09-25
NH-4080 - Load virtual assembly for mapping

Triflepl:
#961 - 5.0 commit with included tests.

Add functionality to load virtual assembly created in application memory for get types from loaded assemblies. Assembles need to be add before use function to add mapping to configuration.

Class "\util\ReflectHelperMemoryAssembly.cs" for store added assembly.
Used by method "TypeFromAssembly()" in "ReflectHelper.cs".

Use new functionality by adding assembly in Configuration method "AddMemoryAssembly()".

It is similar method in Configuration "AddAssembly()", but that method don't work with virtual assembly.


fredericDelaporte:
On the same branch, better just update the existing PR. You can force push to override previous commits if needed. (Now that this is done as a new PR, we will go on on this PR.)

Triflepl:
Ok, i will remember that tips. 
I thinking that way, new commit, then new PR.
So, it is all done, i will be waiting for new info about that PR.

src/NHibernate.Test/CfgTest/LoadVirtualAssembyTest.cs
```diff
@@ -0,0 +1,100 @@
...
+			CompilerParameters CompilerParams = new CompilerParameters();
+			CompilerParams.GenerateInMemory = true;
+			CompilerParams.TreatWarningsAsErrors = false;
+			CompilerParams.GenerateExecutable = false;
+			CompilerParams.CompilerOptions = "/optimize";
+			CompilerParams.ReferencedAssemblies.AddRange(new String[] {
+				"System.dll",
+                "System.Core.dll",
+                "System.Data.dll",
+				"C:\\NHibernate.dll" //load compiled version of NHibernate
```
fredericDelaporte: Not really portable, thus the CI failures. Can't you load from assembly containing a NHibernate type?

src/NHibernate/Cfg/Configuration.cs
```diff
@@ -532,6 +532,18 @@ public void AddDeserializedMapping(HbmMapping mappingDocument, string documentFi
 			}
 		}
 
+		/// <summary>
+		///  Adds assembly created in memory as a resource names for mappning
```
fredericDelaporte: typo in `mappning`.

src/NHibernate/Util/ReflectHelperMemoryAssembly.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NHibernate.Util
+{
+	public class ReflectHelperMemoryAssembly
+    {
+		private static readonly ReflectHelperMemoryAssembly instance = new ReflectHelperMemoryAssembly();
+		private Dictionary<string, Assembly> _assemblyList;
```
fredericDelaporte: A singleton must be thread safe, you should use a [`ConcurrentDictionary<string, Assembly>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2) here, stored in a readonly field.

src/NHibernate/Util/ReflectHelperMemoryAssembly.cs
```diff
@@ -0,0 +1,114 @@
...
+				return instance;
+			}
+		}
+
+		/// <summary>
+		///  Remove all stored assembly in instance.
+		/// </summary>
+		public void Reset()
+		{
+			_assemblyList = new Dictionary<string, Assembly>();
```
fredericDelaporte: Once switched to a concurrent dictionary, `Clear` it instead, because assigning a new one is not thread safe. Of course clearing it all is still a bit debatable if other threads may add assemblies.

src/NHibernate/Util/ReflectHelperMemoryAssembly.cs
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		public ReflectHelperMemoryAssembly()
+		{
+			Reset();
```
fredericDelaporte: Will be unneeded if previous comments proposed changes are done.

src/NHibernate/Util/ReflectHelperMemoryAssembly.cs
```diff
@@ -0,0 +1,114 @@
...
+		///  Add new assembly to store.
+		/// </summary>
+		/// <param name="assembly">Assembly object.</param>
+		/// <returns>True if assembly was stored</returns>
+		public bool AddMemoryAssembly(Assembly assembly)
+		{
+			string name = assembly.FullName;
+			if (name.Length > 0 && !_assemblyList.ContainsKey(name))
+			{ 
+				_assemblyList.Add(name, assembly);
```
fredericDelaporte: Use `TryAdd` instead of `if not contains then add`, once switched to concurrent.

src/NHibernate/Util/ReflectHelperMemoryAssembly.cs
```diff
@@ -0,0 +1,114 @@
...
+		/// <summary>
+		///  Get type from specific assembly.
+		/// </summary>
+		/// <param name="type">Specify name class to get.</param>
+		/// /// <param name="assemblyName">Specify assembly name.</param>
+		public System.Type GetType(string type, string assemblyName)
+		{
+			return _assemblyList.ContainsKey(assemblyName)
+				?
+					_assemblyList[assemblyName].GetType(type)
```
fredericDelaporte: Use `TryGetValue` instead of `if contains then get`, once switched to concurrent.

src/NHibernate/Util/ReflectHelperMemoryAssembly.cs
```diff
@@ -0,0 +1,114 @@
...
+				:
+					_assemblyList.First(s => GetAssemblyName(s.Key) == assemblyName).Value.GetType(type)
```
fredericDelaporte: With a non concurrent dictionary this will throw in case a concurrent thread add or remove an assembly. With a concurrent dictionary it will be ok.

src/NHibernate/Util/ReflectHelperMemoryAssembly.cs
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		/// <summary>
+		///  Remove specify stored assembly.
+		/// </summary>
+		/// <param name="name">Assembly fullname.</param>
+		/// <returns>True if assembly was removed</returns>
+		public bool RemoveAssembly(string name)
+		{
+			return _assemblyList.Remove(name);
```
fredericDelaporte: Use `TryRemove` instead, once switched to concurrent.

src/NHibernate.Test/CfgTest/LoadVirtualAssembyTest.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.CSharp;
+using System.Reflection;
+using NHibernate.Mapping.ByCode;
+using NHibernate.Mapping.ByCode.Conformist;
+
+namespace NHibernate.Test.CfgTest
+{
+	[TestFixture]
+	class LoadVirtualAssembyTest
+    {
```
fredericDelaporte: Space glitch

src/NHibernate.Test/CfgTest/LoadVirtualAssembyTest.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.CodeDom.Compiler;
+using Microsoft.CSharp;
+using System.Reflection;
+using NHibernate.Mapping.ByCode;
+using NHibernate.Mapping.ByCode.Conformist;
+
+namespace NHibernate.Test.CfgTest
+{
+	[TestFixture]
+	class LoadVirtualAssembyTest
```
fredericDelaporte: More tests should be added for covering other methods of `ReflectHelperMemoryAssembly`, like `RemoveAssembly`, `GetType(string)`, ... Or those methods should be removed if they have no usage (seems to the case).

Triflepl:
Yes, i think about that implementation and that function in "src/NHibernate/Util/ReflectHelper.cs" get assembly types more then one time. So i decided to use singleton, for easy store assemblies.
Do you have any suggestion, how to make better implementation of that contribution? 

I can add more test subjects for other methods.

For that assembly compile method i search for better way for use "ReferencedAssemblies", but don't find it. Only strings names for existing references work. 

I will make change what you suggest on the next day.

hazzik:
Can someone explain why the existing method would not work with virtual assemblies?

src/NHibernate.Test/CfgTest/LoadVirtualAssembyTest.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+
+		[Test]
+		public void VirtualAssembyLoadTest()
+		{
+			int assemblyGenerateCount = 10;
+			var config = new Configuration();
+			var mapper = new ModelMapper();
+			var module = Compile(assemblyGenerateCount);
+			config.AddMemoryAssembly(module.Assembly);
```
hazzik: When I comment this line the test still pass.

hazzik: Ok, this behaves differently on Mono and full .NET Framework

hazzik:
No new feature is required, this can be improved by supplying custom assembly resolver to System.Type.GetType.

Replaced by #1699

PR #707, opened 2017-10-03
NH-3787 - Decimal truncation in Linq ternary expression

### LINKED ISSUE #1335: NH-3787 - Decimal truncation in Linq ternary expression
**Erik jansson** created an [issue](https://nhibernate.jira.com/browse/NH-3787) — <time datetime="2015-05-06T10:10:01.783Z" title="2015-05-06T10:10:01.783Z">6th May 2015, 10:10:01</time>:
> My table has two `decimal` values with precision 15 and scale 9.
> 
> 
> ```c#
> public partial class TestEntityMap : ClassMap<TestEntity>
> {
> 	public TestEntityMap()
> 	{
>		Id(x => x.Id).GeneratedBy.Identity();
> 		Map(x => x.UsePreviousRate).Not.Nullable();
> 		Map(x => x.Rate).Precision(15).Scale(9).Not.Nullable();
> 		Map(x => x.PreviousRate).Precision(15).Scale(9).Not.Nullable();
> 	} 
> } 
> ```
> 
> 
> In my query i want to select which value to select based on a boolean (`UsePreviousRate`).
> 
> 
> ```c#
> var queryResult = (from test in _session.Query<TestEntity>()
>                                select new RateDto { Rate = test.UsePreviousRate ? test.PreviousRate : test.Rate }).ToList();
> ```
> 
> 
> If `PreviousRate` is `12345.123456789` i expect the property `Rate` in `RateDto` to be the same but the value is rounded to 5 decimals (`12345.12346`)
> 
> Generated SQL looks like
> 
> ```sql
> select cast(case when testentity0*.UsePreviousRate=1 then testentity0_.PreviousRate else testentity0_.Rate end as DECIMAL(19,5)) as col_0_0_ from <TestEntity> testentity0*
> ```
> 
> 
> The same query with queryover and conditional projection returns expected value
> 
> 
> ```c#
> var query =
>     _session.QueryOver<TestEntity>(() => testEntity)
>         .Select(
>             Projections.Alias(Projections.Conditional(Restrictions.Eq(Projections.Property(() => testEntity.UsePreviousRate), true), Projections.Property(() => testEntity.PreviousRate), Projections.Property(() => testEntity.Rate)), "Rate")
>         .WithAlias(() => rateDto.Rate));
> 
> var queryResult = query.TransformUsing(Transformers.AliasToBean<RateDto>()).List<RateDto>();
> 
> Assert.AreEqual(12345.123456789M, queryResult<0>.Rate);
> ```
> 
> 
> Generated sql
> 
> ```sql
> exec sp*executesql N'SELECT (case when this_.UsePreviousRate = @p0 then this_.PreviousRate else this_.Rate end) as y0_ FROM <TestEntity> this*',N'@p0 bit',@p0=1
> ```
> 
> 
> 
> 
---
**Erik jansson** added a comment — <time datetime="2017-10-02T08:34:07.959Z" title="2017-10-02T08:34:07.959Z">2nd October 2017, 8:34:07</time>:
> This bug has been fixed in NH 4.1.1.4000.
> 
> Generated SQL
> 
> 
> ```sql
> select testentity0*.UsePreviousRate as col_0_0_, testentity0_.Rate as col_1_0_, testentity0_.PreviousRate as col_2_0_ from TestEntity testentity0*
> ```
> 
---
**Frédéric Delaporte** added a comment — <time datetime="2017-10-02T16:36:44.116Z" title="2017-10-02T16:36:44.116Z">2nd October 2017, 16:36:44</time>:
> Binaries removed from test case zip.
> 
---
**Frédéric Delaporte** added a comment — <time datetime="2017-10-02T17:25:11.190Z" title="2017-10-02T17:25:11.190Z">2nd October 2017, 17:25:11</time>:
> Still failing with a variation:
> 
> ```c#
> [Test]
> public void TestLinqQueryOnExpression()
> {
> 	using (var s = OpenSession())
> 	using (var t = s.BeginTransaction())
> 	{
> 		var queryResult = s
> 			.Query<TestEntity>()
> 			.Where(e => (e.UsePreviousRate ? e.PreviousRate : e.Rate) == _testRate)
> 			.ToList();
> 
> 		Assert.That(queryResult.Count, Is.EqualTo(1));
> 		Assert.That(queryResult[0].PreviousRate, Is.EqualTo(_testRate));
> 		t.Commit();
> 	}
> }
> ```
> 
> 
> Yields:
> ```sql
> exec sp_executesql 
> N'select 
> 		testentity0*.Id as id1_0*, 
> 		testentity0*.UsePreviousRate as usepre2_0*, 
> 		testentity0*.PreviousRate as previo3_0*, 
> 		testentity0*.Rate as rate4_0*
> 	from TestEntity testentity0_ 
> 	where cast(
> 		case when testentity0_.UsePreviousRate=1 
> 			then testentity0_.PreviousRate 
> 			else testentity0_.Rate end
> 		as DECIMAL(19,5)) = @p0',
> N'@p0 decimal(28,5)', @p0=12345.12345
> ```
> 
> 
> And that is a double bug: 
> 
> * An optimization for SQL Server query plan cache set decimal parameters precision and scale to their "max" values in [`SqlClientDriver`](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Driver/SqlClientDriver.cs#L133) if they do not specify those themselves. And max scale is only `5`. Any way, max scale is non-sens, since it is indeed equal to precision, in which case their are no more digits before the dot. It is more a default scale. And such a default scale is quite a nasty thing. It causes the value to be truncated (and not rounded) to 5 digits after the dot, when the query parsing was not able to infer its precision/scale from compared entity property (due to the expression). This issue has been isolated in [NH-4087](https://nhibernate.jira.com/browse/NH-4087) (#1196).
> * An undue cast to the NHibernate default decimal (`decimal(19, 5)`) is done on the expression, causing a round. So this undue cast may compensate the first bug when truncation and rounding are equivalent... But fixing the first bug will leave us with the second anyway.
> 
> Tests currently pushed [here](https://github.com/fredericDelaporte/nhibernate-core/tree/NH-3787/src/NHibernate.Test/NHSpecificTest/NH3787), I may PR that later.
>

fredericDelaporte:
Test and fix #1335 ([NH-3787](https://nhibernate.jira.com/browse/NH-3787))~~, with the one still failing set as ignored~~.

~~I will not work on that linq casting trouble soon, so better merge the tests alone.~~


fredericDelaporte:
I shouldn't had a look in this. Now I have added a fix commit.

The trouble is due to a hack in `IIf` expression handling, which translates to a `case`, for which HQL may be unable to determine the type if all its inner expressions are parameters. In such case the query preparation fails when the case type is seek for typing a result column. Adding an extra `cast` around the case solves the trouble, but this is just a workaround for the HQL trouble, and is unneeded for the SQL query. It pollutes it, and may corrupt the data as showcase by this NH-3787 bug.

Since I have not found any easy way to transmit the type to the hql case (or to its inner parameters, which in HQL are un-typed till binding with actual parameters, which occurs too late for the column typing), I have introduced a `transparentcast` hql function which is here just for supplying the type to HQL, without actually casting in SQL.

fredericDelaporte:
Going to fail, I have forgot I have tested with a modified driver from #708, which is required due to NH-4087 being another cause of this bug. So it needs #708 to be merged first.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -540,7 +540,7 @@ protected HqlTreeNode VisitConditionalExpression(ConditionalExpression expressio
 
 			return (expression.Type == typeof (bool) || expression.Type == (typeof (bool?)))
 					   ? @case
-					   : _hqlTreeBuilder.Cast(@case, expression.Type);
+					   : _hqlTreeBuilder.TransparentCast(@case, expression.Type);
```
fredericDelaporte: Still a crutch, but at least one that causes less bugs.

I wonder if some other cast introduced by other method in this file should be switched to transparent too or not. Maybe the `Avg` by example, for it output cast (input casts should clearly stay actual casts). Or the `Sum` cast.

fredericDelaporte:
Rebased and finalized.

This case looks like #708 and #709, but this yet another case. Linq expression translated to a `case when` statement currently introduces a hql `cast` for helping the parser determining the type. But this also causes a sql `cast` which is unnecessary for most databases and causes decimal to eventually lose the adequate scale.

This PR avoids this additional `cast` in SQL, excepted for SQLite which requires it.

fredericDelaporte:
This thing is good to go in my opinion. This should be the last fix of the "decimal fix" series.

hazzik:
@fredericDelaporte need to rename the PR as it is not just "tests" anymore.

PR #708, opened 2017-10-03
NH-4087 - Fix decimal parameter with scale above 5

fredericDelaporte:
[NH-4087](https://nhibernate.jira.com/browse/NH-4087) (#1196) - Fix decimal parameter with scale above 5

 * ~~Cease defining default precision/scale in SQL Server driver unless explicitly configured~~
 * ~~Allow Firebird to adjust its default parameter type cast~~
 * Use dialect configurable default scale/precision


src/NHibernate.Test/DriverTest/OdbcDriverFixture.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Collections;
+using NHibernate.Dialect;
+using NHibernate.Driver;
+using NHibernate.Engine;
+using NUnit.Framework;
+
+namespace NHibernate.Test.DriverTest
+{
+	[TestFixture]
+	public class OdbcDriverFixture : TestCase
```
fredericDelaporte: Copy paste of `SqlClientDriverFixture` which was executed for ODBC too, but with some specific test asserting ignore when it was ODBC. Now separated in its own fixture, with only non-ignored tests.

src/NHibernate.Test/DriverTest/SqlClientDriverFixture.cs
```diff
@@ -49,50 +61,70 @@ public void Crud()
...
-			using (ITransaction t = s.BeginTransaction())
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
 			{
 				var m = s.Get<MultiTypeEntity>(savedId);
+
+				Assert.That(m.StringProp, Is.EqualTo("a"), "StringProp");
+				Assert.That(m.StringClob, Is.EqualTo("a"), "StringClob");
+				Assert.That(m.BinaryBlob, Is.EqualTo(new byte[] { 1, 2, 3 }), "BinaryBlob");
+				Assert.That(m.Binary, Is.EqualTo(new byte[] { 4, 5, 6 }), "BinaryBlob");
```
fredericDelaporte: The test was previously just saving/updating without checking values.

src/NHibernate.Test/DriverTest/SqlClientDriverFixture.cs
```diff
@@ -109,19 +141,73 @@ public void QueryPlansAreReused()
...
+					afterCount - beforeCount,
+					Is.LessThan(insertCount - 1),
+					$"Excessive query plans created: before={beforeCount} after={afterCount}");
 
 				t.Rollback();
 			}
 		}
+
+		[Test]
+		public void DefaultPrecisionScale()
```
fredericDelaporte: Added test for new decimal precision/Scale settings.

src/NHibernate.Test/TestDialect.cs
```diff
@@ -55,6 +58,11 @@ public TestDialect(Dialect.Dialect dialect)
 		/// </summary>
 		public virtual bool SupportsEmptyInserts => true;
 
+		/// <summary>
+		/// Supports condition not bound to any data, like "where @p1 = @p2".
+		/// </summary>
+		public virtual bool SupportsNonDataBoundCondition => true;
```
fredericDelaporte: Copied from #703, I will have to resolve a conflict on one of those PR depending on merge order. But I was needing this.

src/NHibernate.Test/TypesTest/DecimalTypeFixture.cs
```diff
@@ -22,50 +64,100 @@ protected override string TypeName
...
-			object mappedValue = type.StringToObject("0");
-			Assert.AreEqual(0m, mappedValue);
-			Assert.IsTrue(type.IsEqual(mappedValue, 0m), "'0' in the mapping file should have been converted to a 0m");
+			var mappedValue = _type.StringToObject("0");
+			Assert.That(mappedValue, Is.EqualTo(0m));
+			Assert.IsTrue(_type.IsEqual(mappedValue, 0m), "'0' in the mapping file should have been converted to a 0m");
+		}
+
+		[Test]
+		public void HighScaleParameterSelect()
```
fredericDelaporte: Starting with this one, new tests for checking "high" scale decimal are not truncated.

src/NHibernate.Test/TypesTest/DecimalTypeFixture.cs
```diff
@@ -10,9 +11,50 @@ namespace NHibernate.Test.TypesTest
...
+		}
+
+		protected override void Configure(Configuration configuration)
+		{
+			base.Configure(configuration);
+
+			if (Dialect is FirebirdDialect)
+			{
+				configuration.SetProperty(Environment.ParameterDefaultDecimalPrecision, "18");
+				configuration.SetProperty(Environment.ParameterDefaultDecimalScale, "9");
```
fredericDelaporte: Due to Firebird needing an explicit cast in some queries, injected by its driver, we must set its default scale/precision according to what the application needs. (Applications having varying needs will be stuck by this Firebird limitation, unless they use separated session factories just for those varying needs...)

src/NHibernate/Driver/FirebirdClientDriver.cs
```diff
@@ -99,15 +115,27 @@ private HashSet<string> GetCastCandidates(string statement)
...
+			{
+				case DbType.Currency:
+					// Take the dialect binding for currency: decimal(18,4)
+					return _fbDialect.GetTypeName(sqlType);
+				case DbType.Decimal:
+					if (sqlType.PrecisionDefined)
+						return _fbDialect.GetTypeName(sqlType);
+					return _fbDialect.GetTypeName(SqlTypeFactory.GetSqlType(sqlType.DbType, _dbDecimalPrecision, _dbDecimalScale));
+			}
+			return _fbDialect.GetCastTypeName(sqlType);
```
fredericDelaporte: This `GetCastTypeName` is quite a terrible method, hopefully having few usages. It disregards the type `Length`/`Precision`/`Scale` and use instead `Column.DefaultLength`/`Column.DefaultPrecision`/`Column.DefaultScale`, which are `255`/`19`/`2`.
This causes all string types to be cast to a `255` length. Decimal type are trim down to `18` precision by dialect, but anyway, since the type resolution currently compared provided length with the max for the type, which is well below `255` for `decimal`, it always uses the default for `decimal` type, which for Firebird is `decimal(18,5)`. Otherwise we would have got a truncation at 2 digits after dot.
I will fill another Jira for that method.

src/NHibernate/Driver/SqlServerCeDriver.cs
```diff
@@ -104,8 +109,8 @@ protected override void InitializeParameter(DbParameter dbParam, string name, Sq
 			AdjustDbParamTypeForLargeObjects(dbParam, sqlType);
 			if (prepareSql)
 			{
-				SqlClientDriver.SetVariableLengthParameterSize(dbParam, sqlType);
-		}
+				SqlClientDriver.SetVariableLengthParameterSize(dbParam, sqlType, _dbDecimalPrecision, _dbDecimalScale);
+			}
```
fredericDelaporte: Does it really makes sens to set length for SqlServerCe? I believe its sole purpose it to optimize query plan cache usage, but does SqlServerCe have a query plan cache?

fredericDelaporte:
> Allow Firebird to adjust its default parameter type cast

This part would be mostly obsoleted by what I am doing for [NH-4088](https://nhibernate.jira.com/browse/NH-4088).

fredericDelaporte:
Rebased, changed due to `prepare_sql` issues if decimal scale and precision are not always set: now SQL Server driver follows the defaults settable in dialect, instead of some hard-coded defaults. So it allows working around the issue by adjusting those defaults.

src/NHibernate/Driver/SqlClientDriver.cs
```diff
@@ -97,13 +105,16 @@ public override bool SupportsMultipleOpenReaders
...
-			SetVariableLengthParameterSize(dbParam, sqlType);
+			if (_dialect == null)
+				throw new InvalidOperationException("Dialect not available, is this driver used without having been configured?");
+			SetVariableLengthParameterSize(dbParam, sqlType, _dialect.DefaultCastPrecision, _dialect.DefaultCastScale);
 		}
 
 		// Used from SqlServerCeDriver as well
-		public static void SetVariableLengthParameterSize(DbParameter dbParam, SqlType sqlType)
+		public static void SetVariableLengthParameterSize(
+			DbParameter dbParam, SqlType sqlType, byte defaultDecimalPrecision, byte defaultDecimalScale)
```
fredericDelaporte: If doing that in 5.1, I guess we will have to add back previous method but flagged obsolete, for avoiding a possible breaking change.

fredericDelaporte:
Rebased and adjusted for having no more breaking changes.

fredericDelaporte:
Rebased, reworded.

fredericDelaporte:
Anything against merging this ?

hazzik:
@fredericDelaporte absolutely nothing apart I don't understand how it works.

fredericDelaporte:
The trouble: currently SqlClientDriver set a hard coded scale (5) and precision (28) on decimal parameters when their parameter definition does not specify it. (They may be deduced and set by the HQL query parsing in some cases, but not in all cases.) This causes issues when the parameter value need a higher scale than 5.

With 5.0, a configurable default scale (10) and default precision (28) have been introduced, currently used when a cast is issued in HQL or Linq.

This PR uses them instead of the hard coded non configurable ones from SqlClientDriver. So now the user can control the default scale of decimal parameter, and adjust it to its needs, instead of being limited to 5.

This PR also adds tests for checking for all dialects we can now use decimal with a scale higher than 5. The Sql client driver test fixture is also split in two, because it was actually also testing the Odbc driver, which was complexifying it and causing issues for the sql client driver changes tests. Now there is an odbc driver fixture on its own, along with the Sql client driver fixture purged from Odbc specific bits.

PR #709, opened 2017-10-04
NH-4088 - Dialect.GetCastTypeName is buggy

fredericDelaporte:
[NH-4088](https://nhibernate.jira.com/browse/NH-4088) - Dialect.GetCastTypeName is buggy

Fix TypeNames.Get for decimal capacity
 * Compare capacity against precision for precision based types
 * Fix dialects declarations for max precision
 * Fix dialects declarations for precision based types which were using length as precision or scale

Fix GetCastTypeName
 * Use type length/precision/scale when defined
 * Use maximal capacity types otherwise when it makes sens
 * Use configurable default length/precision/scale otherwise

### DIFF
--- a/src/NHibernate.Test/DriverTest/FirebirdClientDriverFixture.cs
+++ b/src/NHibernate.Test/DriverTest/FirebirdClientDriverFixture.cs
@@ -14,157 +14,160 @@ namespace NHibernate.Test.DriverTest
 		private string _connectionString;
 		private FirebirdClientDriver _driver;
 
-		[Test]
-		public void ConnectionPooling_OpenThenCloseThenOpenAnotherOne_OnlyOneConnectionIsPooled()
-		{
-			MakeDriver();
-
-			_driver.ClearPool(_connectionString);
-
-			var allreadyEstablished = GetEstablishedConnections();
-
-			var connection1 = MakeConnection();
-			var connection2 = MakeConnection();
-
-			//open first connection
-			connection1.Open();
-			VerifyCountOfEstablishedConnectionsIs(allreadyEstablished + 1, "After first open");
-
-			//return it to the pool
-			connection1.Close();
-			VerifyCountOfEstablishedConnectionsIs(allreadyEstablished + 1, "After first close");
-
-			//open the second connection
-			connection2.Open();
-			VerifyCountOfEstablishedConnectionsIs(allreadyEstablished + 1, "After second open");
-
-			//return it to the pool
-			connection2.Close();
-			VerifyCountOfEstablishedConnectionsIs(allreadyEstablished + 1, "After second close");
-		}
-
-		[Test]
-		public void ConnectionPooling_OpenThenCloseTwoAtTheSameTime_TowConnectionsArePooled()
-		{
-			MakeDriver();
-
-			_driver.ClearPool(_connectionString);
-
-			var allreadyEstablished = GetEstablishedConnections();
-
-			var connection1 = MakeConnection();
-			var connection2 = MakeConnection();
-
-			//open first connection
-			connection1.Open();
-			VerifyCountOfEstablishedConnectionsIs(allreadyEstablished + 1, "After first open");
-
-			//open second one
-			connection2.Open();
-			VerifyCountOfEstablishedConnectionsIs(allreadyEstablished + 2, "After second open");
-
-			//return connection1 to the pool
-			connection1.Close();
-			VerifyCountOfEstablishedConnectionsIs(allreadyEstablished + 2, "After first close");
-
-			//return connection2 to the pool
-			connection2.Close();
-			VerifyCountOfEstablishedConnectionsIs(allreadyEstablished + 2, "After second close");
-		}
-
-		[Test]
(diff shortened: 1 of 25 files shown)


src/NHibernate.Test/DriverTest/FirebirdClientDriverFixture.cs
```diff
@@ -14,157 +14,160 @@ public class FirebirdClientDriverFixture
...
+		public void OneTimeSetup()
+		{
+			var cfg = TestConfigurationHelper.GetDefaultConfiguration();
+
+			var dlct = cfg.GetProperty("dialect");
+			if (!dlct.Contains("Firebird"))
+				Assert.Ignore("Applies only to Firebird");
+
+			_driver = new FirebirdClientDriver();
+			_driver.Configure(cfg.Properties);
```
fredericDelaporte: Test was failing due to used driver not being configured.

src/NHibernate.Test/DriverTest/FirebirdClientDriverFixture.cs
```diff
@@ -14,157 +14,160 @@ public class FirebirdClientDriverFixture
...
-
-			_driver.AdjustCommand(cmd);
+			using (var cmd = BuildSelectCaseCommand(SqlTypeFactory.GetString(255)))
+			{
+				_driver.AdjustCommand(cmd);
 
-			var expectedCommandTxt = "select (case when col = @p0 then cast(@p1 as VARCHAR(255)) else cast(@p2 as VARCHAR(255)) end) from table";
-			Assert.That(cmd.CommandText, Is.EqualTo(expectedCommandTxt));
+				var expectedCommandTxt =
+					"select (case when col = @p0 then cast(@p1 as VARCHAR(4000)) else cast(@p2 as VARCHAR(4000)) end) from table";
```
fredericDelaporte: It was also failing because still testing for `VARCHAR(255)` in resulting string, while the driver has been changed to use the default length instead of the supplied one for variable length string, due to the sql `like` issue otherwise.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -199,6 +199,9 @@ private static Dialect InstantiateDialect(string dialectName, IDictionary<string
 		/// <param name="settings">The configuration settings.</param>
 		public virtual void Configure(IDictionary<string, string> settings)
 		{
+			DefaultCastLength = PropertiesHelper.GetInt32(Environment.QueryDefaultCastLength, settings, 4000);
+			DefaultCastPrecision = PropertiesHelper.GetByte(Environment.QueryDefaultCastPrecision, settings, null) ?? 28;
+			DefaultCastScale = PropertiesHelper.GetByte(Environment.QueryDefaultCastScale, settings, null) ?? 10;
```
fredericDelaporte: There are some inconsistencies in those `PropertiesHelper` methods.

I think the right signature is the `GetInt32` one, but completed with an overload yielding a nullable and not taking any default.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -215,17 +218,10 @@ public virtual string GetTypeName(SqlType sqlType)
 		{
 			if (sqlType.LengthDefined || sqlType.PrecisionDefined || sqlType.ScaleDefined)
 			{
-				string resultWithLength = _typeNames.Get(sqlType.DbType, sqlType.Length, sqlType.Precision, sqlType.Scale);
-				if (resultWithLength != null) return resultWithLength;
+				return _typeNames.Get(sqlType.DbType, sqlType.Length, sqlType.Precision, sqlType.Scale);
```
fredericDelaporte: Those `TypeNames` methods are never yielding `null`. (Unless a `Put` was done with `null`, but now `Put` throws null argument exception if it is attempted.)

src/NHibernate/Dialect/MySQLDialect.cs
```diff
@@ -64,7 +63,7 @@ public MySQLDialect()
 			RegisterColumnType(DbType.String, 16777215, "MEDIUMTEXT");
 			//todo: future: add compatibility with decimal???
 			//An unpacked fixed-point number. Behaves like a CHAR column; 
-			//�unpacked� means the number is stored as a string, using one character for each digit of the value.
+			//“unpacked” means the number is stored as a string, using one character for each digit of the value.
```
fredericDelaporte: Encoding change. The file was not UTF-8.

src/NHibernate/Dialect/TypeNames.cs
```diff
@@ -72,30 +72,46 @@ public string Get(DbType typecode)
...
+					// Math.Min((int) Math.Round(scale * adjustedPrecision / (double) precision), adjustedPrecision);
+					// But if the type is used for storing amounts, this may cause losing the ability to store cents...
+					// So better just reduce as few as possible.
+					var adjustedScale = Math.Min(scale, adjustedPrecision);
+					return Replace(maxEntry.Value, size, adjustedPrecision, adjustedScale);
+				}
 			}
 			//Could not find a specific type for the capacity, using the default
-			return Replace(Get(typecode), size, precision, scale);
+			return Get(typecode);
```
fredericDelaporte: Default typenames are never supposed to have formatting placeholders.

PR #710, opened 2017-10-09
Fixes with prepare_sql

hazzik:
prepare_sql requires all sizes/scales to be specified upfront.

~This PR to be dropped if all tests are successful.~

### DIFF
--- a/src/NHibernate.Test/Async/NHSpecificTest/SqlConverterAndMultiQuery/Fixture.cs
+++ b/src/NHibernate.Test/Async/NHSpecificTest/SqlConverterAndMultiQuery/Fixture.cs
@@ -9,7 +9,8 @@
 
 
 using NHibernate.Cfg;
-using NHibernate.Dialect;
+using NHibernate.Driver;
+using NHibernate.Engine;
 using NUnit.Framework;
 
 namespace NHibernate.Test.NHSpecificTest.SqlConverterAndMultiQuery
@@ -25,11 +26,13 @@ namespace NHibernate.Test.NHSpecificTest.SqlConverterAndMultiQuery
 			configuration.DataBaseIntegration(x => x.ExceptionConverter<SqlConverter>());
 		}
 
-		protected override bool AppliesTo(Dialect.Dialect dialect)
+		protected override bool AppliesTo(ISessionFactoryImplementor factory)
 		{
-			// MsSqlCe throws InvalidOperationException instead of a DbException for these tests, preventing
-			// the test SqlConverter to do its job.
-			return !(Dialect is MsSqlCeDialect);
+			// Test current implementation allows to test mmostly SQL Server. Other databases
+			// tend to (validly) send InvalidOperationException during prepare phase due to the closed
+			// connection, which get not converted. For testing other case, maybe a failure caused by a
+			// schema mismatch (like done in transaction tests) would be better.
+			return factory.ConnectionProvider.Driver is SqlClientDriver;
 		}
 
 		[Test]
--- a/src/NHibernate/Driver/Sql2008ClientDriver.cs
+++ b/src/NHibernate/Driver/Sql2008ClientDriver.cs
@@ -2,12 +2,13 @@ using System;
 using System.Data;
 using System.Data.Common;
 using System.Data.SqlClient;
-using System.Linq;
 
 namespace NHibernate.Driver
 {
 	public class Sql2008ClientDriver : SqlClientDriver
 	{
+		const byte MaxTime = 5;
+
 		protected override void InitializeParameter(DbParameter dbParam, string name, SqlTypes.SqlType sqlType)
 		{
 			base.InitializeParameter(dbParam, name, sqlType);
@@ -15,6 +16,7 @@ namespace NHibernate.Driver
 			{
 				case DbType.Time:
 					((SqlParameter) dbParam).SqlDbType = SqlDbType.Time;
+					dbParam.Size = MaxTime;
 					break;
 				case DbType.Date:
 					((SqlParameter) dbParam).SqlDbType = SqlDbType.Date;
--- a/src/NHibernate/Driver/SqlClientDriver.cs
+++ b/src/NHibernate/Driver/SqlClientDriver.cs
@@ -100,7 +100,6 @@ namespace NHibernate.Driver
 			SetVariableLengthParameterSize(dbParam, sqlType);
 		}
 
-		// Used from SqlServerCeDriver as well
 		public static void SetVariableLengthParameterSize(DbParameter dbParam, SqlType sqlType)
 		{
 			SetDefaultParameterSize(dbParam, sqlType);
--- a/src/NHibernate/Driver/SqlServerCeDriver.cs
+++ b/src/NHibernate/Driver/SqlServerCeDriver.cs
@@ -4,8 +4,6 @@ using System.Data;
 using System.Data.Common;
 using System.Reflection;
 using NHibernate.SqlTypes;
-using NHibernate.Util;
-using Environment = NHibernate.Cfg.Environment;
 
 namespace NHibernate.Driver
 {
@@ -25,13 +23,11 @@ namespace NHibernate.Driver
 		{
 		}
 
-		private bool prepareSql;
 		private PropertyInfo dbParamSqlDbTypeProperty;
 
 		public override void Configure(IDictionary<string, string> settings)
 		{
 			base.Configure(settings);
-			prepareSql = PropertiesHelper.GetBoolean(Environment.PrepareSql, settings, false);
 
 			using (var cmd = CreateCommand())
 			{
@@ -102,10 +98,6 @@ namespace NHibernate.Driver
 			base.InitializeParameter(dbParam, name, AdjustSqlType(sqlType));
 
 			AdjustDbParamTypeForLargeObjects(dbParam, sqlType);
-			if (prepareSql)
-			{
-				SqlClientDriver.SetVariableLengthParameterSize(dbParam, sqlType);
-		}
 		}
 
 		private static SqlType AdjustSqlType(SqlType sqlType)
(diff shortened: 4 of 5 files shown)


fredericDelaporte:
It seems to demonstrate older issues, mainly with `Time`. That is for Odbc that some size specifications were removed. Not for SQL Client driver. `Time` was never sized.

hazzik:
Ok, then we need to add these sizings

fredericDelaporte:
Currently testing that, going to add a Jira. Any way good catch. Since this seems a SQL Client specific trouble, do we add this prepare_sql to the 2012 build by example? Testing 2008 without, 2012 with? Or do we duplicate all SQL Server builds, so long for the total build time?

fredericDelaporte:
I have overwritten your merge with a rebase + a quick fix to check if there are other issues than time with SQL Server.

src/NHibernate/Driver/SqlClientDriver.cs
```diff
@@ -144,6 +145,9 @@ protected static void SetDefaultParameterSize(DbParameter dbParam, SqlType sqlTy
 				case DbType.DateTimeOffset:
 					dbParam.Size = MaxDateTimeOffset;
 					break;
+				case DbType.Time:
```
hazzik: This should be in 2008 driver, as 2005 does not support time and reverts it to DateTime. Same for DateTime2.

fredericDelaporte: And same for `DateTimeOffset`. Defining defaults for those type supported only for 2008+ in the driver for previous versions is not very sound, excepted that the current pattern of sharing that method with SQL Server Ce, causing it to be static, causes it to be hard to be overridden in 2008 driver.

The reverting for `DbType.Time` and  `DbType.Date` seems to be a specific "feature" of `DbParameter` interface, when setting types through `DbType` property, forcing us to set them through `SqlDbType`, and only into 2008 driver.

To clean-up that maybe we should cease sharing with SQL Server CE, switch to instance methods and override them. But it would be a breaking change for those using `DateTime2` and `DateTimeOffset` with the old SQL Server driver.

fredericDelaporte:
ODBC is going to fail due to to four tests, like `NormalHqlShouldThrowUserException`, which get another failure than the one expected with ODBC when prepare SQL is enabled.

src/NHibernate/Driver/SqlClientDriver.cs
fredericDelaporte: Another option could be to check the dialect in `SqlClientDriver` for flagging whether we are using 2008+ or not, and if yes port and use corresponding code from `SqlClient2008Driver`. That way, no need to override anything anymore in  `SqlClient2008Driver`, which would be then set as obsolete.

fredericDelaporte:
SQL Server CE fails with insufficient memory errors on blob types tests. Seems to be yet another issue.

fredericDelaporte:
> SQL Server CE fails with insufficient memory errors on blob types tests. It seems to be yet another issue.

This occurs only with 32 bits version. The 64 bits has not the trouble. The root of the trouble is the default sizing of binary parameters to 2Gb when `prepare_sql` is enabled... This wrecks the 32 bits versions, as if it was trying to allocate such a size two or more times, hitting the 4Gb limit for 32 bits processes. The 64 bits surely does same but with my 16Gb setup, it does not get out of memory.
Disabling parameters sizing removes the trouble for SQL Server CE and let all tests work, although according to its documentation sizing parameters is mandatory when preparing SQL.

So we could fix this [NH-4091](https://nhibernate.jira.com/browse/NH-4091) by ceasing specifying parameter sizes even when `prepare_sql` is enabled, which would by the way allow to switch this sizing method to instance and override it for sql 2008. Or we need to change it for not specifying blob sizes when called by SQL Server CE.

src/NHibernate.Test/Async/NHSpecificTest/SqlConverterAndMultiQuery/Fixture.cs
```diff
@@ -27,9 +28,12 @@ protected override void Configure(Configuration configuration)
...
 		protected override bool AppliesTo(Dialect.Dialect dialect)
 		{
-			// MsSqlCe throws InvalidOperationException instead of a DbException for these tests, preventing
+			// If prepare_sql is enabled, with most databases prepare will fail with an InvalidOperationException,
+			// turned into ADOException by prepare code and then not passed to the exception converter.
+			// MsSqlCe throws InvalidOperationException during Prepare (even when disabled), preventing
 			// the test SqlConverter to do its job.
-			return !(Dialect is MsSqlCeDialect);
+			var prepareSql = PropertiesHelper.GetBoolean(Environment.PrepareSql, cfg.Properties, false);
+			return !(prepareSql || Dialect is MsSqlCeDialect);
```
hazzik: Maybe it shall be only for SqlClientDriver?

fredericDelaporte: `return !(prepareSql && !(driver is SqlClientDriver) || Dialect is MsSqlCeDialect);` ? Or just `return driver is SqlClientDriver` ?

src/NHibernate/Driver/SqlServerCeDriver.cs
```diff
@@ -102,10 +98,6 @@ protected override void InitializeParameter(DbParameter dbParam, string name, Sq
 			base.InitializeParameter(dbParam, name, AdjustSqlType(sqlType));
 
 			AdjustDbParamTypeForLargeObjects(dbParam, sqlType);
-			if (prepareSql)
-			{
-				SqlClientDriver.SetVariableLengthParameterSize(dbParam, sqlType);
-		}
```
fredericDelaporte: What do you think of this removal? It works but is in contradiction with [documentation](https://msdn.microsoft.com/en-us/library/system.data.sqlserverce.sqlcecommand.prepare(v=vs.100).aspx#Anchor_2):

> Before you call Prepare, specify the data type of each parameter in the statement to be prepared. For each parameter that has a variable-length data type, you must set the Size property to the maximum size needed. Prepare returns an error if these conditions are not met.

hazzik: It is strange that it works.

src/NHibernate.Test/Async/NHSpecificTest/SqlConverterAndMultiQuery/Fixture.cs
hazzik: Second

src/NHibernate/Driver/Sql2008ClientDriver.cs
```diff
@@ -2,19 +2,21 @@
 using System.Data;
 using System.Data.Common;
 using System.Data.SqlClient;
-using System.Linq;
 
 namespace NHibernate.Driver
 {
 	public class Sql2008ClientDriver : SqlClientDriver
 	{
+		public const byte MaxTime = 5;
```
hazzik: Why is this 5? It shall be either 7 or 8 (same as datetime2)

fredericDelaporte: Yes, the same for `datetime2` and `datetimeoffset`, where it wants to knows their size instead of their  scale.

src/NHibernate.Test/Async/NHSpecificTest/SqlConverterAndMultiQuery/Fixture.cs
fredericDelaporte: I do not really get why it should be considered a Sql Server only test. For me it is more like this tests reveal there is some "too wide" catching in place causing the feature to be unavailable for most databases when prepare is enabled. Anyways, that is a quite secondary feature in my mind, and the code lacks indications as to why this "wide" catch is there, so I do not think we should investigate/fix this point, at least not for 5.0.

hazzik: Ok, leave it as is

fredericDelaporte: Or maybe more accurately, reveals that the test implementation (closing the connection) is not suitable for most databases when prepare is enabled. Maybe should it instead use a trick (like the nullable no null column in some transactions tests for failures) to alter the schema and use that to cause an actual db side failure.

fredericDelaporte: Well, already changed, with comment corresponding to my last comment just above.

src/NHibernate/Driver/SqlServerCeDriver.cs
fredericDelaporte: I dislike it quite. But that is the simplest change to get it work... Otherwise I could move the code from `SqlClientDriver` into Sql CE driver and adjust it to its needs, meaning removing 2008 stuff and not specifying blob sizes, but letting all the other been set, since it will be more compliant with doc and should work too.

src/NHibernate.Test/Async/NHSpecificTest/SqlConverterAndMultiQuery/Fixture.cs
hazzik: Need to regen the async

fredericDelaporte: Done. Thanks for the reminder.

src/NHibernate/Driver/Sql2008ClientDriver.cs
fredericDelaporte: About those `MaxSizes`, I have put it public for mimicking what was done for others. But well, I do not think there is need for that. Maybe time to clean-up a bit the driver by switching back to protected those parameter sizing members.

src/NHibernate/Driver/SqlServerCeDriver.cs
hazzik: I don't think that SqlServer CE requires having all parameter sizes to be set when `Prepare` is used.

fredericDelaporte: So we just keep that change as is. This PR could then be merged after dropping the first commit.

hazzik:
The first commit dropped.

PR #711, opened 2017-10-10
Switch doc generation to UTF-8.

fredericDelaporte:
Not found how to add the html 5 dtd, it seems to supports only XHTML dtd. And I have not searched for meta.

### DIFF
--- a/doc/reference/styles/html.xsl
+++ b/doc/reference/styles/html.xsl
@@ -23,12 +23,13 @@
                 xmlns="http://www.w3.org/TR/xhtml1/transitional"
                 exclude-result-prefixes="#default">
                 
-<xsl:import href="&db_xsl_path;/html/docbook.xsl"/>
+<xsl:import href="&db_xsl_path;/html/onechunk.xsl"/>
 
 <!--###################################################
                      HTML Settings
     ################################################### -->   
 
+    <xsl:param name="chunker.output.encoding">UTF-8</xsl:param>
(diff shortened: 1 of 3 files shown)


doc/reference/styles/html.xsl
```diff
@@ -23,12 +23,13 @@
                 xmlns="http://www.w3.org/TR/xhtml1/transitional"
                 exclude-result-prefixes="#default">
                 
-<xsl:import href="&db_xsl_path;/html/docbook.xsl"/>
+<xsl:import href="&db_xsl_path;/html/onechunk.xsl"/>
```
hazzik: I don't know anything about this, but why has this changed?

fredericDelaporte: Because the standard one page xsl `docbook` does not handle any encoding option, and so we have to use the chunker, but with a one page chunker since we do not really want to chunk. See [here](http://www.sagehill.net/docbookxsl/OneChunk.html):

> The chunking stylesheet gives you several options you do not have with the nonchunking `docbook.xsl` stylesheet that generates a single HTML output file.

...

> The DocBook XSL distribution includes a special stylesheet named `onechunk.xsl` that uses the same extension function to generate a single HTML file. This is most useful when you want your output in a single file, but you want to set one or more of these options:

...

> - Specify the output encoding. See [the section “Output encoding for chunk HTML”](http://www.sagehill.net/docbookxsl/Chunking.html#ChunkEncoding).

PR #712, opened 2017-10-10
NH-4092 - AsyncGenerator creates unused private static event handler in SQLite20Driver

ngbrown:
Addresses [NH-4092](https://nhibernate.jira.com/browse/NH-4092).

Exclude `SQLite20Driver.Connection_StateChange` in AsyncGenerator.yml and regenerate async files.

If it is ok for the `Connection.StateChange` event to call a fire-and-forget event handler, then the primary event handler method can be modified to an `async void`.


src/NHibernate/Async/Connection/DriverConnectionProvider.cs
```diff
@@ -1,51 +0,0 @@
...
-		/// the <see cref="NHibernate.Driver.IDriver"/>.
-		/// </summary>
-		/// <param name="cancellationToken">A cancellation token that can be used to cancel the work</param>
-		/// <returns>
-		/// An Open <see cref="DbConnection"/>.
-		/// </returns>
-		/// <exception cref="Exception">
-		/// If there is any problem creating or opening the <see cref="DbConnection"/>.
-		/// </exception>
-		public override async Task<DbConnection> GetConnectionAsync(CancellationToken cancellationToken)
```
hazzik: I don't understand how this went away. It still uses `conn.OpenAsync`

ngbrown: I don't really know.  There may be some issues in the AsyncGenerator.  I tried various ways to try to exclude `SQLite20Driver` from being generated and it always made all these changes.

hazzik: @maca88 can you please investigate?

ngbrown: @maca88 There is this warning in the generate-async task log on the current master branch (before these changes):

```
     [exec] INFO  ProjectAnalyzer:0 - Analyzing documents started
     [exec] WARN  ProjectAnalyzer:0 - Cannot attach an async method to an event (void async is not an option as cannot be awaited):
     [exec] connection.StateChange += Connection_StateChange
     [exec] 
     [exec] INFO  ProjectAnalyzer:0 - Analyzing documents completed
```

What is the correct way to keep the un-used/un-needed private static method from being generated?

maca88: > What is the correct way to keep the un-used/un-needed private static method from being generated?

Currently, the generator is not smart enough to skip generating un-used private methods, I will try to add support for it.
For un-needed methods, you have to add them in the configuration as you did, but it seems that for this particular method something is not right as `GetConnectionAsync` is removed even if there is still one async method used (`OpenAsync`). I will investigate it when I get home.

maca88: [I've added](https://github.com/maca88/AsyncGenerator/commit/5e9fb5b38b47e9f62c0873e073a42875e7becdc4) support for skipping the generation of unused methods, just by updating to the new version `0.5.1` the method `Connection_StateChange` won't be generated anymore.

Here is why the `GetConnectionAsync` goes await when ignoring `Connection_StateChange`:

With the current settings `GetConnection` has the conversion set to `Unknown` when the generator starts scanning, which means that the generator will convert this method to async only if one of the async marked method is called/referenced inside `GetConnection` method.
Before this change, the `GetConnection` was generated because of the following dependencies:

1. `IBatcher.ExecuteNonQuery` which has the conversion set to `Async` within the configuration
2. Inside `IBatcher.ExecuteNonQuery` method body, `DbCommand.ExecuteNonQuery` was found to have an async counterpart
3. `SQLite20Driver.Connection_StateChange` was found because uses the `DbCommand.ExecuteNonQuery` method
4. `SQLite20Driver.CreateConnection` was found because uses the `Connection_StateChange` method
5. `DriverConnectionProvider.GetConnection` was found beause uses `IConnectionProvider.CreateConnection`
6. Inside `DriverConnectionProvider.GetConnection` method body, `DbConnection.Open` was found to have an async counterpart

By ignoring `SQLite20Driver.Connection_StateChange` method, the generator will stop at point 3, which means that `DriverConnectionProvider.GetConnection` will never be scanned and consequently not generated. So the solution (if we ignore the new version) would be, to set the conversion of `GetConnection` method to `Async` or `Smart` in order to have its method body scanned.

ngbrown: Thanks @maca88 for the quick turn-around on this.  Updating to the new version found a few more un-used private functions that were being generated.

PR #983, opened 2017-10-12
Fix forgotten CDATA closure.

fredericDelaporte:
Here is the glitch...


fredericDelaporte:
Of course I should have compiled it before PR, but I just forgot... Sorry for the inconvenience.

PR #1374, opened 2017-10-13
Prevent SQLite from messing with timezone offset of DateTime.

ngbrown:
This addresses #1362 by writing all `DateTime` parameter values to SQLite as `DateTimeKind.Unspecified`.

This should work even when the connection string has `;DateTimeKind=Utc`, as the ADO.Net driver uses that to mung all `DateTime` values to Utc on reading back, irregardless of how they were saved.

### DIFF
--- a/src/NHibernate/Type/AbstractDateTimeType.cs
+++ b/src/NHibernate/Type/AbstractDateTimeType.cs
@@ -52,7 +52,7 @@ namespace NHibernate.Type
 		/// <param name="dateValue">The <see cref="System.DateTime" /> to adjust.</param>
 		/// <returns>A <see cref="System.DateTime" />.</returns>
 		protected virtual DateTime AdjustDateTime(DateTime dateValue) =>
-			Kind == DateTimeKind.Unspecified ? dateValue : DateTime.SpecifyKind(dateValue, Kind);
+			DateTime.SpecifyKind(dateValue, Kind);
 
 		/// <inheritdoc />
 		public override object Get(DbDataReader rs, int index, ISessionImplementor session) =>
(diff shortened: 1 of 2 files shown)


src/NHibernate/Type/AbstractDateTimeType.cs
```diff
@@ -52,7 +52,7 @@ protected AbstractDateTimeType(SqlType sqlTypeDateTime) : base(sqlTypeDateTime)
 		/// <param name="dateValue">The <see cref="System.DateTime" /> to adjust.</param>
 		/// <returns>A <see cref="System.DateTime" />.</returns>
 		protected virtual DateTime AdjustDateTime(DateTime dateValue) =>
-			Kind == DateTimeKind.Unspecified ? dateValue : DateTime.SpecifyKind(dateValue, Kind);
+			DateTime.SpecifyKind(dateValue, Kind);
```
fredericDelaporte: Is really this change required? SQLite users working with `DateTimeKind=UTC` would have a legitimate grievance with this change, as it would wreck their reading of dates, causing them to be unspecified instead of being said as UTC, and forcing them to use NHibernate Utc types instead.

I do not think we should force unspecified unless really having a compelling reason to do it. I think the NHibernate "unspecified" types should have as semantic "no change at all to read date kind", as was doing `TimestampType` previously (which is now become `DateTimeType`.) The fact that previous `DateTimeType`, cutting fractional seconds, is losing the kind, was likely accidental design. (`DateTimeNoMsType` probably still does that indeed.)

ngbrown: I think it should stay the way it is without any configuration.  See the [unit test I created](https://gist.github.com/ngbrown/195e09681572371011fb2c78c5fcf04d) and [my comment on the issue](https://github.com/nhibernate/nhibernate-core/issues/1362#issuecomment-336591202).

hazzik:
Replaced by #1378

PR #1377, opened 2017-10-14
Logging refactoring

ngbrown:
Using structured log processing can greatly benefit application development.  Most logging libraries support sinks that can handle the structured nature of log messages.  This means each of the parameters in a `string.Format()` can be become meaningful.  The [.NET Core logger](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging?tabs=aspnetcore2x), [NLog](https://github.com/NLog/NLog), [Serilog](https://github.com/serilog/serilog), and others support sinking to structured logging systems.

This is how the NHibernate.Example.Web logging can look with the changing values colored to be noticeable (this is the default settings for the Serilog console sink):
screenshot - 2017-10-13 17_07_23

The legacy `IInternalLogger` is directly modeled after the log4net interface and limits the interaction with these modern logging systems.  For example, any logging of an exception precludes also passing the string.Format `params` array into the logging system.

There were also many areas that either string concatenation was taking place to generate the log message or string.Format was called then a single string passed into the logger.  These have been fixed.

This pull request adds the ability to implement and set an alternative `INHibernateLoggerFactory` which then returns implementations of the new `IInternalLogger2` interface.  It is not a breaking change for applications that are providing a custom `ILoggerFactory` and `IInternalLogger` implementation.  The old factory will still be accepted.

There are some obsolete extension methods that are intended to catch any merges that are using the old logging function style with the exception as the second parameter.  They can probably be removed at some point, because they generate an error if referenced.


hazzik:
I think we should utilize FormattableString instead of message + params.

ngbrown:
Using `FormattableString` might be possible with extension methods. It shouldn't be implemented directly on the interface. The structured logging frameworks, including Microsoft's, would prefer to name the parameters (something besides 0, 1, 2, etc).

hazzik:
Sorry, I don't follow.

ngbrown:
I mean, that when your application is using NLog or log4net with a structured log sink, then in your logging system, the name each of the parameters shows as '0', '1', '2', etc.  If you were using Serilog or Microsoft's `ILogger` (from `Microsoft.Extensions.Logging.Abstractions`) they both allow the use of a name in the message string that is used in the structured output.  See [example from here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging?tabs=aspnetcore2x#log-message-format-string):

```cs
_logger.LogInformation("Getting item {ID} at {RequestTime}", id, DateTime.Now);
```

For output at the logging server, I use Seq ([see example of logged message](https://docs.getseq.net/docs)), but I'm sure it's similar in most other systems such as Application Insights or Elasticsearch.

As for not adding more methods to the interface, it might actually be beneficial to collapse it down to just a few methods that have the level passed in as a parameter and do the per-level logging with extension methods.  The Microsoft's `ILogger` interface only has [three methods](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=aspnetcore-2.0), and one of them is scope oriented, so NHibernate only needs two.

Alternatively, we could just directly rely on `Microsoft.Extensions.Logging.Abstractions`, but I didn't know if you wanted to take on that dependency.

ngbrown:
Anyways, if `FormattableString` is desired instead of a string + params, then using extension methods + a raw string conversion class seems to be the way to do it.  Otherwise, there are precedence issues with standard `string`.

ngbrown:
I've moved most logging methods to extension  methods so anything can be done with them in the future without breaking clients implementing custom logging.  The interface in now just this:

```cs
public interface IInternalLogger2
{
	void Log(InternalLogLevel logLevel, InternalLogValues state, Exception exception);
	bool IsEnabled(InternalLogLevel logLevel);
}
```

The state object is built in the extension methods and it would be straightforward to use `FormattableString` in the future without changing the clients again.

mstancombe:
Nice! log4net can use named attributes which can then be exposed to appenders such as app insights for richer analytics, so maybe this ticket should be 'structured logging' rather than 'modern logging', but nitpicking aside, it's a great addition to the base logging in nhib.👍

hazzik:
I have a concern that this change can bring performance regression. Could you please check and doublecheck, that this is not the case?

ngbrown:
The build times on TeamCity are equivalent, so they don't show any major change.

I wrote some performance tests that simulate 20 million log calls and timed them repeatedly for a statistically valid sample.  Each test consists of making 20 million log calls, half formatted, and counting the characters in a mock logger (381 MiB worth).

```cs
var iterationCount = 10000000;
for (int i = 0; i < iterationCount; i++)
{
    logger2.Warn("message");
    logger2.Warn("message with parameters {0}, {1}", "string", 5);
}
```

This takes `3,677 ms` with the old interfaces, and `3,922 ms` with the new interfaces.

If the 20 million log calls are all disabled (and the formatting and counting don't take place), then the calls took `244 ms` with the old interface and `451 ms` with the new interface.

These times are normalized to 95 percent confidence numbers.  See [logger analysis.xlsx](https://github.com/nhibernate/nhibernate-core/files/1473654/logger.analysis.xlsx).

I did notice that the times are slightly helped by making `InternalLogValues` a `struct`, so I made that change.

fredericDelaporte:
I have checked with Frans bench too:

Test | NH v5.0.1 | NH altered by this PR  | % (base is unaltered 5.0.1)
Change tracking fetches, set fetches (50 runs), no caching | 2.477,37ms, Enum: 1,59ms, 828.952KB | 2.513,55ms, Enum: 1,52ms, 833.132KB | +1,5%ms, Enum: -4,4%, +0.5%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | 0,42ms, 96KB | 0,39ms, 96KB | -2,4%ms, 0%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | 381,94ms, 123.000KB | 412,59ms, 123.737KB | +8%ms, +0,6%KB

All differences are smaller than the standard deviation of each test, so this PR impact on this benchmark performance is not significant in my opinion.

src/NHibernate/Logging.cs
```diff
@@ -101,135 +138,183 @@ private static string GetNhibernateLoggerClass()
...
 
-		public bool IsDebugEnabled
+		private LoggerProvider(INHibernateLoggerFactory loggerFactory)
 		{
-			get { return false; }
+			this._loggerFactory = loggerFactory;
 		}
 
-		public bool IsInfoEnabled
+		public static IInternalLogger2 LoggerFor(string keyName)
```
fredericDelaporte: This is a breaking change, it will breaks dependent packages like NHibernate.Caches which use LoggerProvider.LoggerFor.

If keeping this as is, it will mandate a 6.0.

For it to be 5.1, we would need to either have `IInternalLogger2` extend `IInternalLogger` (confusing, bloating) or to keep `LoggerProvider.LoggerFor` yielding `IInternalLogger`, adding for migrated logging either a `LoggerFor2` method or a `LoggerProvider2`. And of course, for keeping non migrated logging working, we would need to change `LoggerFor` for it to use an `IInternalLogger` adapter over an `IInternalLogger2` logger.

src/NHibernate/Logging.cs
```diff
@@ -101,135 +138,183 @@ private static string GetNhibernateLoggerClass()
...
 		{
-			get { return false; }
+			return _instance._loggerFactory.LoggerFor(keyName);
 		}
 
-		public bool IsWarnEnabled
+		public static IInternalLogger2 LoggerFor(System.Type type)
```
fredericDelaporte: Same as previous comment.

src/NHibernate/Logging.cs
```diff
@@ -430,4 +518,133 @@ public void WarnFormat(string format, params object[] args)
...
+			logger.Log(InternalLogLevel.Debug, new InternalLogValues(format, args), exception);
+		}
+
+		public static void Debug(this IInternalLogger2 logger, string format, params object[] args)
+		{
+			logger.Log(InternalLogLevel.Debug, new InternalLogValues(format, args), null);
+		}
+
+
+		// catch any method calls with an Exception argument second as they would otherwise silently be consumed by `parms object[] args`.
```
fredericDelaporte: Typo: `parms` instead of `params`.

src/NHibernate/Logging.cs
```diff
@@ -430,4 +518,133 @@ public void WarnFormat(string format, params object[] args)
...
+
+		/// <summary>
+		/// Don't use or implement, simply throw new NotImplementedException();
```
fredericDelaporte: May be confusing. I would write instead:

> Throws <c>NotImplementedException</c>. Calling this method is an error. Please use methods taking the exception as first argument instead.

Same for the four next methods.

src/NHibernate/Logging.cs
```diff
@@ -101,135 +138,183 @@ private static string GetNhibernateLoggerClass()
...
+					throw new ArgumentOutOfRangeException(nameof(logLevel), logLevel, null);
+			}
 		}
+	}
 
-		public void Info(object message, Exception exception)
+	public class NoLoggingLoggerFactory: INHibernateLoggerFactory
+	{
+		private static readonly IInternalLogger2 Nologging = new NoLoggingInternalLogger();
+		public IInternalLogger2 LoggerFor(string keyName)
```
fredericDelaporte: Same as previous comment. Applies to next method too. Applies to next class.

src/NHibernate/Stat/StatisticsImpl.cs
```diff
@@ -424,34 +424,34 @@ public QueryStatistics GetQueryStatistics(string queryString)
...
-			log.Info("collections updated: " + collectionUpdateCount);
-			log.Info("collections removed: " + collectionRemoveCount);
-			log.Info("collections recreated: " + collectionRecreateCount);
-			log.Info("collections fetched (minimize this): " + collectionFetchCount);
-			log.Info("queries executed to database: " + queryExecutionCount);
-			log.Info("query cache puts: " + queryCachePutCount);
-			log.Info("query cache hits: " + queryCacheHitCount);
-			log.Info("query cache misses: " + queryCacheMissCount);
-			log.Info("max query time: " + queryExecutionMaxTime.TotalMilliseconds.ToString("0") + " ms");
+			log.Info("start time: {0}", startTime.ToString("o"));
```
fredericDelaporte: By the way could be changed to `log.Info("start time: {0:o}", startTime);`

src/NHibernate/Util/ADOExceptionReporter.cs
```diff
@@ -19,20 +19,25 @@ public static void LogExceptions(Exception ex)
...
 				{
 					message = StringHelper.IsNotEmpty(message) ? message : DefaultExceptionMsg;
-					log.Debug(message, ex);
+					log.Debug(ex, message);
 				}
+
+				// Pass full exception on highest call
+				if (log.IsWarnEnabled()) log.Warn(ex, ex.ToString());
+				log.Error(ex, ex.Message);
+				ex = ex.InnerException;
```
fredericDelaporte: Why those changes and the next?
That changes what is was actually doing. Previously if warn was enabled it would have fully logged inner exceptions. Now it does no more.

src/NHibernate/Util/ADOExceptionReporter.cs
```diff
@@ -19,20 +19,25 @@ public static void LogExceptions(Exception ex)
...
 				while (ex != null)
 				{
-					log.Warn(ex);
+					if (log.IsWarnEnabled()) log.Warn(ex.ToString());
```
fredericDelaporte: I would instead change that to `log.Warn(ex, ex.Message)` and revert the other changes.

src/NHibernate/Util/ReflectHelper.cs
```diff
@@ -407,35 +407,34 @@ public static System.Type TypeFromAssembly(AssemblyQualifiedTypeName name, bool
 				if (name.Assembly == null)
 				{
 					// No assembly was specified for the type, so just fail
-					string message = "Could not load type " + name + ". Possible cause: no assembly name specified.";
-					log.Warn(message);
-					if (throwOnError) throw new TypeLoadException(message);
+					log.Warn("Could not load type {0}. Possible cause: no assembly name specified.", name);
+					if (throwOnError) throw new TypeLoadException("Could not load type " + name + ". Possible cause: no assembly name specified.");
```
fredericDelaporte: Here previous code was better in my opinion. Now the message may be constructed two times by two different means.

fredericDelaporte:
I have updated some comments of my review, which were just forgetting the point of the PR.

src/NHibernate/Logging.cs
ngbrown: I didn't know about NHibernate.Caches using `LoggerProvider.LoggerFor`.  I guess the best option would be to create a new `LoggerProvider.LoggerFor2`.  Rev'ing major versions has too long of a timeline.

Alternatively, and a better option if NHibernate is OK taking on a new external dependency, there is [Microsoft.Extensions.Logging.Abstractions](https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions/) and it already has adapters to many logging framework.

fredericDelaporte: Why not. Now it appears that this package is not CLS compliant, contrary to NHibernate and its current dependencies. So this would be another breaking change, going non CLS compliant.

src/NHibernate/Util/ReflectHelper.cs
hazzik: Still. The message in exception needs to be updated to use string.Format

ngbrown:
This now addresses preventing a breaking change for dependent packages.

I added a second thunk to in reverse to supply a non-changing public interface for libraries dependent on `NHibernate.LogProvider.LoggerFor()`, such as NHibernate.Caches.  The public implementation should now be non-breaking for a minor version.

At the next major version, all the obsolete classes, methods, and interfaces can be removed. Any dependent library that has no warnings in 5.x should continue work with no changes when the obsolete methods are removed.

@fredericDelaporte had an alternative way that built upon all but the last commit of this pull request [here](https://github.com/fredericDelaporte/nhibernate-core/blob/modern-logging/src/NHibernate/Logging.cs). However in that alternative, because implementer of the new `INHibernateLogger` would also need to implement `IInternalLogger` (with the help of a new base class), I think it would require libraries implementing the new interfaces to eventually make an additional change at a major version of NHibernate (see https://github.com/ngbrown/nhibernate-core/pull/4#issuecomment-348357621).

I think either way meets the goals of this pull request.

fredericDelaporte:
My proposition was for avoiding playing with names, like putting a `LoggerFor2` and maybe later once `LoggerFor` is dropped, obsoleting `LoggerFor2` and going back to the old name but with new interface.

So if instead we have names that sounds ok for all new interfaces methods, the approach in Nathan PR is ok too. So now we have `LoggerProvider.For()` for getting the new logger, while the old one stays on `LoggerProvider.LoggerFor()`. Why not putting a real method name for the new one? Normally a method starts with a verb, so it could be `LoggerProvider.GetLogger()`.

(Note: I have not taken time to review the last changes and I will probably not have much time for reviewing before next monday.)

fredericDelaporte:
Checking again the Frans benchmark with latest version of this PR, it gives:

Test | NH v5.0.1 | NH altered by this PR v2  | % (base is unaltered 5.0.1)
Change tracking fetches, set fetches (50 runs), no caching | 2.477,37ms, Enum: 1,59ms, 828.952KB | 2.493,94ms, Enum: 1,51ms, 833.058KB | +0,7%ms, Enum: -5%, +0.5%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | 0,42ms, 96KB | 0,38ms, 96KB | -9,5%ms, 0%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | 381,94ms, 123.000KB | 427,76ms, 123.745KB | +12%ms, +0,6%KB

The impact stays low.

src/NHibernate/Logging.cs
```diff
@@ -101,27 +202,330 @@ private static string GetNhibernateLoggerClass()
...
+		public bool IsEnabled(InternalLogLevel logLevel)
+		{
+			if (logLevel == InternalLogLevel.None) return true;
+
+			return false;
+		}
+	}
+
+	// Since 5.1
+	[Obsolete("To set no-logging, use LoggerProvider.SetLoggersFactory((INHibernateLoggerFactory)null)")]
```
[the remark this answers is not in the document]
ngbrown: I changed to `default(INHibernateLoggerFactory)` and also merged master in.

hazzik:
Sorry, I might be late. But I would like to propose to use a class instead of interface `INHibernateLogger`. A class where all methods are no-op virtual. This way it will prevent any further breaking changes. (Still did not look into the PR in details, sorry).

src/NHibernate/Logging.cs
```diff
@@ -316,118 +738,254 @@ static Log4NetLogger()
...
+		[Obsolete("Use Debug(Exception, string, params object[])", true)]
+		public static void Debug(this INHibernateLogger logger, string message, Exception ex) => ThrowNotImplemented();
+
+		private static void ThrowNotImplemented()
+		{
+			throw new NotImplementedException("Should not have compiled with call to this method");
+		}
+	}
+
+	public struct InternalLogValues
```
hazzik: This is almost identical tothe ` FormattableString`

src/NHibernate/Logging.cs
```diff
@@ -316,118 +738,254 @@ static Log4NetLogger()
...
+	/// </summary>
+	public static class NHibernateLoggerExtensions
+	{
+		public static bool IsDebugEnabled(this INHibernateLogger logger) => logger.IsEnabled(InternalLogLevel.Debug);
+		public static bool IsInfoEnabled(this INHibernateLogger logger) => logger.IsEnabled(InternalLogLevel.Info);
+		public static bool IsWarnEnabled(this INHibernateLogger logger) => logger.IsEnabled(InternalLogLevel.Warn);
+		public static bool IsErrorEnabled(this INHibernateLogger logger) => logger.IsEnabled(InternalLogLevel.Error);
+		public static bool IsFatalEnabled(this INHibernateLogger logger) => logger.IsEnabled(InternalLogLevel.Fatal);
+
+		public static void Fatal(this INHibernateLogger logger, Exception exception, string format, params object[] args)
```
hazzik: `params` always allocates an array. Better to provide several overloads, which take no args, one arg, two args, etc.

src/NHibernate/Logging.cs
```diff
@@ -316,118 +738,254 @@ static Log4NetLogger()
...
+		{
+			logger.Log(InternalLogLevel.Info, new InternalLogValues(format, args), null);
+		}
+
+		public static void Debug(this INHibernateLogger logger, Exception exception, string format, params object[] args)
+		{
+			logger.Log(InternalLogLevel.Debug, new InternalLogValues(format, args), exception);
+		}
+
+		public static void Debug(this INHibernateLogger logger, string format, params object[] args)
```
hazzik: There are 229 calls to this method with no arguments. but the array will be allocated anyway.
401 calls with 1 argument
88 calls with 2 arguments
And 32 calls with 3 or more arguments

src/NHibernate/Logging.cs
fredericDelaporte: > `params` always allocates an array. Better to provide several overloads, which take no args, one arg, two args, etc.

An overload with "no args" would save the array, but I do not see how overloads for one args, two args, ..., would save it.
Unless the underlying structure `InternalLogValues` handles additional properties for storing separately individual values of arguments for those other overloads (which would quite complicate the struct and its usage), it would anyway have to create an array for storing them.

So why not adding overload for the "no args" cases, but for the other cases it seems to me quite harder to get benefits from coding them.

By the way, the previous interface is not attempting to optimize these other cases. It is having "no args" methods, but not "one arg", "two args", ... Introducing optimization for these other args cardinality cases may ask for another PR.

src/NHibernate/Logging.cs
fredericDelaporte: > This is almost identical to the `FormattableString`

Do you mean `InternalLogValues` should be a class deriving from `FormattableString` instead? Why not. Looking at the concrete implementation available in .Net Framework, it would require slightly more code though.

From `System.Runtime.CompilerServices`:
```c#
private sealed class ConcreteFormattableString : FormattableString
{
	private readonly string _format;

	private readonly object[] _arguments;

	public override string Format
	{
		get
		{
			return _format;
		}
	}

	public override int ArgumentCount
	{
		get
		{
			return _arguments.Length;
		}
	}

	internal ConcreteFormattableString(string format, object[] arguments)
	{
		_format = format;
		_arguments = arguments;
	}

	public override object[] GetArguments()
	{
		return _arguments;
	}

	public override object GetArgument(int index)
	{
		return _arguments[index];
	}

	public override string ToString(IFormatProvider formatProvider)
	{
		return string.Format(formatProvider, _format, _arguments);
	}
}
```

hazzik:
The problem with this PR is that it trying to sit on two chairs. We need to clearly separate 2 concepts here: a logger adapter and a logger. A logger shall be internal to the library (it's now represented by the NHibernateLoggerExtensions). There are a lot of unnecessary changes to how we use the logger in-house.

fredericDelaporte:
> The problem with this PR is that it trying to sit on two chairs. We need to clearly separate 2 concepts here: a logger adapter and a logger. A logger shall be internal to the library (it's now represented by the NHibernateLoggerExtensions). A logger adapter (what's now INHibernateLogger) and it's factory interface shall be public. Because these concepts are mixed together there are a lot of unnecessary changes to how we use the logger in-house.

Isn't that the actual situation without this PR? For me this situation is not changed by this PR. It looks to me as a related subject, not as the subject of this PR. Should it be a prerequisite for this PR? I do not think so because it looks to me it would be quite heavy prerequisite, with breaking changes or at least obsoleting things used by dependent libraries like NHibernate.Caches without supplying a replacement (but telling instead something like "log by yourself").

ngbrown:
> In current commits state, it would be merged as a single squashed commit.

A single squashed commit is fine.  There's back and forth noise in the commits, but nothing that really adds value.

> But I would like to propose to use a class instead of interface `INHibernateLogger`. A class where all methods are no-op virtual. This way it will prevent any further breaking changes.

I see your point, but I think the expectation is generally that it would be an interface.  There is the proposal for c# interfaces which would allow providing default implementations for new methods: https://github.com/dotnet/csharplang/issues/52.

> This is almost identical to the `FormattableString`

Using `FormattableString` is a `string.Format()` replacement concern, not for transport to the host application for logging.  This is structured so that `FormattableString` could be used within NHibernate at some future point by logging methods with no change to the `INHibernateLogger` contract.  (something like this `log.Info($"Entity Id={entity.Id}");`).  The logging extension would be responsible for passing the standard String.Format composite formatting string and array to the adaptor.  The composite formatting string and array is what what `FormattableString` provides.  In fact, it doesn't appear that it's possible to get the original string with the variable name out of it.

> `params` always allocates an array. Better to provide several overloads, which take no args, one arg, two args, etc.

The no args case is a simple addition to `NHibernateLoggerExtensions` with no changes anywhere else.  One arg would require changes to the log value transport and the logging libraries may not care and allocate an array anyways when passed to their formatter.

> The problem with this PR is that it trying to sit on two chairs. We need to clearly separate 2 concepts here: a logger adapter and a logger....

The goal of this pull request was to separate all varying parts of the logging messages into params so that structured logging at the client side works correctly.  This means not having a unique format string for every variation of a parameter changing because `string.Format()` was ran before it got to the logging library.

These may be two separate concepts, but the previous interface was limited in what it could do (no args for Exception logging) and needed breaking changes, so it was better to leap to the end state with the interface in one step.

I was informed that it wasn't really internal and the logging functions were used externally by other libraries such as NHibernate.Caches.  This required more compatibility and keeping the logging function externally visible.  Ideally the `NHibernateLoggerExtensions` would be internal (with specific friend assemblies for test).

hazzik:
@fredericDelaporte it surely can be made without breaking changes. That approach, in fact, will bring much fewer changes to the code base, providing, however, greater benefits. I'll generally be fine with leaving `LoggerProvider.For` as public to server something like Caches or other "adapter" libraries.

>I see your point, but I think the expectation is generally that it would be an interface. 

Whose expectation? My expectation is to give people this simple interface (what is currently INHibernateLogger) to implement wrap this implementation into the adapter class and use that class. The inability to add a default implementation is the least of the problems with the interface. The major problem is that all methods are public.

>This is almost identical to the `FormattableString`
I was just thinking out-loud. 

>it doesn't appear that it's possible to get the original string with the variable name out of it [`FormattableString`].
We're not doing that anyway. And the reality is that as we are implementing the baseline adaptor we're limited by the most common functionality. There are logging frameworks who do not support the "parametrized" messages or whose syntax for such "parametrizing" is different. So, we really limited by the base string format with index-based "parametrization"

>There is the proposal for c# interfaces which would allow providing default implementations for new methods: dotnet/csharplang#52.

Not really related. I'm 200% sure that the "default implementations" would be implemented as a syntactic sugar for extension/static methods. Otherwise, it will require changes to the runtime as an addition to the compiler, which MS would resist as hard as they can.

> The goal of this pull request was to separate all varying parts of the logging messages into params so that structured logging at the client side works correctly.

> Isn't that the actual situation without this PR? For me this situation is not changed by this PR. It looks to me as a related subject, not as the subject of this PR.

Yeah, we are changing here apples to apples. It does not really help to solve the root problem: interfaces are inflexible.

ngbrown:
This is the interface in question:

```cs
public interface INHibernateLogger
{
	void Log(InternalLogLevel logLevel, InternalLogValues state, Exception exception);
	bool IsEnabled(InternalLogLevel logLevel);
}
```

It is two functions, so there's also the `InternalLogLevel` enumerable and `InternalLogValues` struct, but those are extensible if needed.

Is this pull request acceptable if those two methods are virtual with default implementations?  There's no way to make them non-public because the log adapters need to implement them.

hazzik:
No, not this exact interface is in question, but what's returning by `LoggerProvider.For` is.

ngbrown:
So are we talking about the factory then?

```cs
public interface INHibernateLoggerFactory
{
    INHibernateLogger LoggerFor(string keyName);
    INHibernateLogger LoggerFor(System.Type type);
}
```

It is a variation on the previous version which was `ILoggerFactory`, just with different return types.  An application using NHibernate wouldn't care that there are two variations and just implement one; the legacy one or the new one.

gliljas:
I would agree that the expectation is that such things are exposed as interfaces. Whether that expectation is always valid is a different matter. It has to be decided on a case by case basis. IMHO a logger should be a very stable contract, since it really doesn't do much. 

Speaking of expectations, I would expect a `LoggerProvider.For` method to return a LoggerProvider (implementation), not a logger. That's the convention for such method names, in my experience, and that's lingustically what it says.

ngbrown:
In the example web project there is an example of a factory and logging framework adapter:

https://github.com/ngbrown/nhibernate-core/tree/8947da537191d5313415d61087b3a5b20d938d98/src/NHibernate.Example.Web/Infrastructure

`Logger.For(typeof(SessionImpl));` would be more linguistically correct for what is going on with that, but `LoggerProvider` is what already existed.

gliljas:
> Not really related. I'm 200% sure that the "default implementations" would be implemented as a syntactic sugar for extension/static methods. Otherwise, it will require changes to the runtime as an addition to the compiler, which MS would resist as hard as they can.

It is somewhat related, since the reason they are adding it is to make interfaces more change friendly. If they go ahead with it, they will change the runtime, according to Mads Torgensen.

hazzik:
> I would agree that the expectation is that such things are exposed as interfaces.

This expectation would be valid if we were doing a logging framework, but we are _using_ logging, and so that expectation does not apply here. It might be exposed as whatever we please as it is not intended by the consumers (except by, probably, extension libraries as Caches). Using the same interface which a client implemented feels wrong. 

Major logging frameworks do segregate "internal" logger and "public" logger. log4net for example: "[public contract](https://github.com/apache/logging-log4net/blob/develop/src/ILog.cs)" that looks exctly as IIniternalLogger and "[private contract](https://github.com/apache/logging-log4net/blob/develop/src/Core/ILogger.cs)" which looks exactly as new INHibernateLogger. The only exception that in our case these contracts are reversed.

hazzik:
> If they go ahead 

Let's not implement things based on some hope, assumptions or predictions for some future. My gut feeling says that there will be no default methods for the interfaces in the next 5 years.

> IMHO a logger should be a very stable contract

Apparently, it is not as demonstrated by this PR.

>but LoggerProvider is what already existed. [And similar comments]

At this point, we have an opportunity to improve things, trying to hold back does not make it better. If we are doing this change, it needs to be designed and implemented properly (with a compatibility level of course).

hazzik:
>So are we talking about the factory then?

No, about `LoggerProvider.For`

ngbrown:
So in summary, these are the changes I've made:

- Split the logging implementation into multiple files for easier understanding:
  - `Logging.cs` has the new implementations
  - `Logging.log4net.cs` has the factory and implementation for reflection based log4net logger.
  - `Logging.extensions.cs` is the new extension methods
  - `Logging.obsolete.cs` is the obsolete interfaces and implementations.
- `LoggerProvider` is now fully obsolete and exists just for compatibility.  It contains just enough logic to stay synchronized with the active logging factory.
- `NHibernateLogger` is the new in-library logger and `NHibernateLogger.For(typeof(SessionImpl))` would return an `INHibernateLogger` from the currently active `INHibernateLoggerFactory`.  It is static and not meant to be inherited.
- `INHibernateLogger` and `INHibernateLoggerFactory` have not changed except the following rename:
  - `InternalLogLevel` and `InternalLogValues` are renamed to `NHibernateLogLevel` and `NHibernateLogValues`.

Previously `IInternalLogger` mirrored the external logging interface of one specific log library and that was lacking.  Seeing how `INHibernateLogger` is now similar to the internal logging interfaces of various log libraries means that this is probably on the right track for what is the better format for converting to other loggers.

There can always be more learning the future.

src/NHibernate/Logging.extensions.cs
```diff
@@ -0,0 +1,154 @@
...
+		{
+			logger.Log(NHibernateLogLevel.Fatal, new NHibernateLogValues(format, null), exception);
+		}
+
+		public static void Fatal(this INHibernateLogger logger, string format, params object[] args)
+		{
+			logger.Log(NHibernateLogLevel.Fatal, new NHibernateLogValues(format, args), null);
+		}
+
+		public static void Fatal(this INHibernateLogger logger, string format)
```
fredericDelaporte: A bit strange to have a `format` parameter without argument. Maybe would be better to rename it `message` for this no args overload.

src/NHibernate/Logging.cs
```diff
@@ -101,333 +119,121 @@ private static string GetNhibernateLoggerClass()
...
-			ParameterExpression repositoryAssemblyParam = Expression.Parameter(typeof(Assembly), "repositoryAssembly");
-			MethodCallExpression methodCall = Expression.Call(null, method, repositoryAssemblyParam, nameParam);
-			return Expression.Lambda<Func<Assembly, string, object>>(methodCall, repositoryAssemblyParam, nameParam).Compile();
+		private readonly string _format;
+		private readonly object[] _args;
+
+		/// <summary>
+		/// Instantiates a new instance of the <see cref="NHibernateLogValues"/> structure.
+		/// </summary>
+		/// <param name="format">a composite format string</param>
```
fredericDelaporte: For casing and punctuation consistency, should be `A composite format string.`.

src/NHibernate/Logging.extensions.cs
hazzik: Agree

ngbrown:
Minor typo and parameter name adjustments made.  Merge with master done.

This can be squash merged since I haven't been rebasing it.

mxmissile:
Where is `FormattedLogValues` coming from, I'm unable to resolve? Greenfield project, asp.net core 3.1.

ngbrown:
@mxmissile You're talking about an example, and the target for that example was .net core 2.0:

https://github.com/nhibernate/nhibernate-core/blob/ac39173567b31bcfad475ca32687c9faf0d37f87/src/NHibernate.Example.Web/Infrastructure/NHibernateToMicrosoftLogger.cs#L28-L31

The available extension methods for the `Microsoft.Extensions.Logging.ILogger` interface has changed slightly since then.  In .net core 3.1, there's an extension overload that directly takes the format and args parameters:

```cs
public void Log(NHibernateLogLevel logLevel, NHibernateLogValues state, Exception exception)
{
    _msLogger.Log(MapLevels[logLevel], exception, state.Format, state.Args);
}
```

PR #1379, opened 2017-10-14
Doc fixes

### LINKED ISSUE #788: NH-1947 - Undocumented attributes on sql-query element
**Richard Dingwall** created an issue — <time datetime="2009-08-30T21:06:40.712Z" title="2009-08-30T21:06:40.712Z">30th August 2009, 21:06:40</time>:
> Nhibernate-mapping.xsd has revealed the following attributes on <sql-query> that are not documented in http://nhibernate.info/doc/nh/en/index.html:
> 
> cache-mode
> callable
> flush-mode
> read-only
> 
> They sound like they might be relevant to my use case, but I don't know exactly what they do. It would be really great if someone could document them!
> 
---
**Fabio Maulo** added a comment — <time datetime="2009-09-11T14:58:58.397Z" title="2009-09-11T14:58:58.397Z">11th September 2009, 14:58:58</time>:
> we'll really happy too.
>

fredericDelaporte:
Fixes #788, #823, #824, #1165 and #1179. (This time I have tested generating the doc locally before issuing the PR.)


src/NHibernate/Cfg/Environment.cs
```diff
@@ -78,6 +79,7 @@ public static string Version
 		public const string ConnectionStringName = "connection.connection_string_name";
 
 		// Unused, Java-specific
+		// But has many code usage though.
 		public const string SessionFactoryName = "session_factory_name";
```
fredericDelaporte: Not switched to obsolete because its many code usages imply a lot of changes, and so I am not sure it has really not any use. (And I do not intend to investigate further the subject.)

src/NHibernate/Dialect/Oracle8iDialect.cs
```diff
@@ -309,12 +309,6 @@ protected virtual void RegisterFunctions()
 		protected internal virtual void RegisterDefaultProperties()
 		{
 			//DefaultProperties[Environment.DefaultBatchFetchSize] = DefaultBatchSize; It can break some test and it is a user matter
-
-			// Oracle driver reports to support GetGeneratedKeys(), but they only
-			// support the version taking an array of the names of the columns to
-			// be returned (via its RETURNING clause).  No other driver seems to
-			// support this overloaded version.
-			DefaultProperties[Environment.UseGetGeneratedKeys] = "false";
```
fredericDelaporte: This thing was set but never read. Its corresponding feature is not implemented.

Issue #750, opened 2017-10-12
Transformers.AliasToBean: Value cannot be null. Parameter name: key 


mazharqayyum:
Just test upgraded from version 3.3.1 to 5.0.0 from nuget. Some of our test cases are broken after upgrade and they all seems to be making use of Transformers.AliasToBean. I am not able to open latest code in Visual Studio 2015 so at the moment can't do further debugging.

```
at NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters, IResultTransformer forcedResultTransformer)
at NHibernate.Loader.Loader.ListIgnoreQueryCache(ISessionImplementor session, QueryParameters queryParameters)
at NHibernate.Loader.Loader.List(ISessionImplementor session, QueryParameters queryParameters, ISet`1 querySpaces, IType[] resultTypes)
at NHibernate.Impl.SessionImpl.List(CriteriaImpl criteria, IList results)
at NHibernate.Impl.CriteriaImpl.List(IList results)
at NHibernate.Impl.CriteriaImpl.List[T]()
at NHibernate.Impl.CriteriaImpl.Subcriteria.List[T]()`
```

fredericDelaporte:
There has been some work on this result transformer in #597. There are some details about it in [NH-3693](https://nhibernate.jira.com/browse/NH-3693).

NHibernate is using the VS2017 csproj format, so you need a compatible IDE, like JetBrains Rider or Visual Studio 2017 Community (the later can be used freely by anyone for open source projects).

fredericDelaporte:
Without more information, this will likely be closed. We have many tests passing with this transformer, so we need a failing test case for your trouble. (What is the bean, the query, the database used, its schema involved for this case, the complete failure (not just a stack trace excerpt),...?) We do not ask you to debug NHibernate code, just your own failing code should allow to give way more information.

mazharqayyum:
I will post the details. I was trying to sum up with a failing test case to demonstrate the issue.

mazharqayyum:
`ICriteria criteria = NHibernateHelper.CreateCriteria<CommerceBuilder.Orders.Order>("O", maximumRows, startRowIndex, sortExpression)
                .CreateCriteria("Items", "OI", NHibernate.SqlCommand.JoinType.InnerJoin)
                .CreateCriteria("OI.Product", "P", NHibernate.SqlCommand.JoinType.InnerJoin)
[...]

PR #1380, opened 2017-10-15
#750 - AliasToBean failure, test case and fix

mazharqayyum:
…ues/750

I have created the unit test to demonstrate exception happening due to AliasToBean result transformer. The related issue is https://github.com/nhibernate/nhibernate-core/issues/750


fredericDelaporte:
Regression introduced by 9491f57aca2264e293889dc960e49ddfb20350f8, the check on `null` alias has been removed, causing queries having no (criteria) alias for some of the columns they return to fail. This happen with grouping in criteria by example.

src/NHibernate.Test/Async/NHSpecificTest/GH0750/Fixture.cs
```diff
@@ -0,0 +1,115 @@
...
+using NHibernate.Transform;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.GH0750
+{
+	using System.Threading.Tasks;
+	using System.Threading;
+
+	[TestFixture]
+	public class FixtureAsync : BugTestCase
```
fredericDelaporte: Its generation was forgotten. See [this section of contributing](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md#regenerate-async-code).

src/NHibernate.Test/NHSpecificTest/GH0750/Fixture.cs
```diff
@@ -0,0 +1,113 @@
...
+				session.Save(blackTeaOrder);
+
+				blackTeaOrder.OrderLines.Add(new OrderLine() { Order = blackTeaOrder, Product = blackTea, Quantity = 5, UnitPrice = blackTea.UnitPrice ?? 0 });
+
+				session.Save(blackTeaOrder);
+				tx.Commit();
+			}
+		}
+
+		protected override void OnTearDown()
```
fredericDelaporte: This was missing. Deleting all data is required by our tests, otherwise it is considered as a teardown failure.

src/NHibernate/Transform/AliasToBeanResultTransformer.cs
```diff
@@ -121,6 +121,10 @@ public override IList TransformList(IList collection)
 		/// same visibility and inheritance depth.</exception>
 		private void SetProperty(string alias, object value, object resultObj)
 		{
+			if (alias == null)
+				// Grouping properties in criteria are selected without alias, just ignore them.
+				return;
```
fredericDelaporte: The fix. Reintroduces the check removed [here](https://github.com/nhibernate/nhibernate-core/pull/597/commits/9491f57aca2264e293889dc960e49ddfb20350f8#diff-e431ffcf2f1f91c671a967d74bcb1252L85) without us spotting it and questioning it.

src/NHibernate.Test/Async/NHSpecificTest/GH0750/Fixture.cs
mazharqayyum: OOPS! It was very hectic to work with VS2017 since it keeps throwing itself into NOT RESPONDING state for my install while I add/edit/rename files in test project. Would keep it in mind if needed to write another one.

fredericDelaporte: No troubles. Yes VS2017 is currently quite bad at handling large projects migrated to its new csproj format. That is why I had mentioned Rider, which does quite better. But maybe things will start to be better with changes like Microsoft/msbuild#2572.

About writing another one, you will need to clean your master branch. But you should do it only after this PR is closed, otherwise you will need to open a new one from a dedicated branch this time. 

For cleaning, in case you need instructions, it could be something like:

 * Branch away changes done on master if you want to keep them. Assuming you are currently on master, `git branch newBranch`.
 * Still on master, reset it hard `git reset --hard head~x`, where x is the number of commits to remove (2 currently `git reset --hard head~2`).
 * Force push it to your remote `git push -f`.

mazharqayyum: Thanks for the detailed instructions. I will wait for this PR to close.

PR #1382, opened 2017-10-16
Jira to GitHub: change issue naming in tests

fredericDelaporte:
We need to change our naming since the recent switch to GitHub issues from Jira. Otherwise we will have soon conflicts between old Jira test cases and new GitHub ones. I propose to replace `NH` by `GH`, since `GH-` is an alternate to `#` for referencing issues.


src/NHibernate.Test/NHSpecificTest/NH2218/Fixture.cs
```diff
@@ -1,6 +1,5 @@
 ﻿using System.Linq;
 using NHibernate.Linq;
-using NHibernate.Test.NHSpecificTest.NH0000;
```
fredericDelaporte: This test case was using directly th NH0000 entity.

PR #1388, opened 2017-10-17
Fix #1387 - Linq Sum() with ToFutureValue fails

### LINKED ISSUE #1387: Linq Sum() with ToFutureValue fails
I have the following code to get a sum of values using futures:

```csharp
var sum = s.Query<AnEntity>()
	.Select(x => x.AnInt)
	.ToFutureValue(x => x.Sum());

Assert.NotZero(sum.Value);
```

This fails in NHibernate 5 with the following exception:

```
System.ArgumentException: Object of type 'System.Linq.EnumerableQuery`1[System.Int32]' cannot be converted to object of type 'System.Collections.Generic.IEnumerable`1[System.Nullable`1[System.Int32]]'.
   at System.RuntimeType.TryChangeType(Object value, Binder binder, CultureInfo culture, Boolean needsSpecialCast)
   at System.RuntimeType.CheckValue(Object value, Binder binder, CultureInfo culture, BindingFlags invokeAttr)
   at System.Reflection.MethodBase.CheckArguments(Object[] parameters, Binder binder, BindingFlags invokeAttr, CultureInfo culture, Signature sig)
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Delegate.DynamicInvokeImpl(Object[] args)
   at NHibernate.Impl.FutureValue`1.get_Value()
```

`s.Query<AnEntity>()` is not empty. Just using `Sum()` without futures works as expected.

I already analyzed this bug and created a pull request where I explain it in more detail: #1388

nkreipke:
This fixes issue #1387.

To cope for the mismatch between LINQ to Objects and SQL regarding sum behavior for an empty set (SQL will return NULL, LINQ will return 0), the result type of a `Sum` query is changed to `Nullable<int>` (or another nullable value type). A `PostExecuteTransformer` is applied which converts the resulting `IList<Nullable<int>>` to `int`. This happens in `NHibernate.Linq.Visitors.QueryModelVisitor.AddPostExecuteTransformerForSum()`.

I found an issue with `NHibernate.Impl.FutureValue<T>` where it can only apply post execute transformers which accept the future's result type as the parameter. This is because `NHibernate.Impl.FutureBatch<T1, T2>` already applies a `Cast<TResult>` operation in `GetCurrentResult<TResult>` before the post execute transformer is invoked.

In the PR, I removed type information from `FutureBatch.GetCurrentResult` and subsequently `FutureValue.GetResult` and `FutureValue.GetResultAsync`. If `ExecuteOnEval` is used, the result list is cast to the correct parameter type of the delegate.

### DIFF
--- a/src/NHibernate.Test/Futures/LinqFutureFixture.cs
+++ b/src/NHibernate.Test/Futures/LinqFutureFixture.cs
@@ -360,36 +360,5 @@ namespace NHibernate.Test.Futures
 				}
 			}
 		}
-
-		[Test(Description = "https://github.com/nhibernate/nhibernate-core/issues/1387")]
-		public void ToFutureValueWithSumReturnsResult()
-		{
-			// Prepare data
-			using (var s = OpenSession())
-			using (var tx = s.BeginTransaction())
-			{
-				s.Save(new Person());
-				s.Save(new Person());
-				tx.Commit();
-			}
-
-			using (var s = OpenSession())
-			{
-				var personsSum = s.Query<Person>()
-					.Select(x => x.Id)
-					.ToFutureValue(x => x.Sum());
-
-				Assert.IsNotNull(personsSum);
-				Assert.NotZero(personsSum.Value);
-			}
-
-			// Cleanup
-			using (var s = OpenSession())
-			using (var tx = s.BeginTransaction())
-			{
-				s.Delete("from Person");
-				tx.Commit();
-			}
-		}
 	}
 }
--- a/src/NHibernate.Test/Futures/LinqFutureFixture.cs
+++ b/src/NHibernate.Test/Futures/LinqFutureFixture.cs
@@ -360,5 +360,36 @@ namespace NHibernate.Test.Futures
 				}
 			}
 		}
+
+		[Test(Description = "https://github.com/nhibernate/nhibernate-core/issues/1387")]
+		public void ToFutureValueWithSumReturnsResult()
+		{
+			// Prepare data
+			using (var s = OpenSession())
+			using (var tx = s.BeginTransaction())
+			{
+				s.Save(new Person());
+				s.Save(new Person());
+				tx.Commit();
+			}
+
+			using (var s = OpenSession())
+			{
+				var personsSum = s.Query<Person>()
+					.Select(x => x.Id)
+					.ToFutureValue(x => x.Sum());
+
+				Assert.IsNotNull(personsSum);
+				Assert.NotZero(personsSum.Value);
+			}
+
+			// Cleanup
+			using (var s = OpenSession())
+			using (var tx = s.BeginTransaction())
+			{
+				s.Delete("from Person");
+				tx.Commit();
+			}
+		}
 	}
 }
--- a/src/NHibernate/Util/MakeQueryableHelper.cs
+++ b/src/NHibernate/Util/MakeQueryableHelper.cs
@@ -1,0 +1,23 @@
+﻿using System;
+using System.Collections;
+using System.Linq;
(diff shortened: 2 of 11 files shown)


src/NHibernate.Test/Futures/LinqFutureFixture.cs
```diff
@@ -360,5 +360,36 @@ public void CanExecuteMultipleQueriesOnSameExpression()
 				}
 			}
 		}
+
+		[Test(Description = "https://github.com/nhibernate/nhibernate-core/issues/1387")]
+		public void ToFutureValueWithSumReturnsResult()
```
fredericDelaporte: Since it is a `FutureValue` only test (not mixed with `Future`), it may be better placed in `LinqToFutureValueFixture`.

src/NHibernate/Util/MakeQueryableHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections;
+using System.Linq;
+using System.Reflection;
+
+namespace NHibernate.Util
+{
+	internal static class MakeQueryableHelper
+    {
+		private static readonly Lazy<MethodInfo> _implMethod = new Lazy<MethodInfo>(() =>
+			typeof(MakeQueryableHelper).GetMethod(nameof(MakeQueryableImpl), BindingFlags.Static | BindingFlags.NonPublic));
```
fredericDelaporte: NHibernate has reflection helpers. Such reflection is usually done as:

```C#
private static readonly MethodInfo _implMethodDefinition =
	ReflectHelper.GetMethodDefinition(() => MakeQueryableHelper.MakeQueryableImpl<object>(null));
```

Better follow the NHibernate usual pattern.

src/NHibernate.Test/Futures/LinqFutureFixture.cs
fredericDelaporte: We likely need to also check the behavior of future with empty sum, and same on nullables.

nkreipke:
Thanks for your remarks. I updated the PR and added additional test cases.

fredericDelaporte:
I have put future tests in all NH-3850 tests, and it appears we need more to fix the issue.

fredericDelaporte:
I think we need to run the post execute transformer [here](https://github.com/nhibernate/nhibernate-core/blob/5.0.0/src/NHibernate/Impl/MultiQueryImpl.cs#L480).

fredericDelaporte:
Failing tests for your tests are due to different multi query implementation depending on the target database : the throw is not always the same...

nkreipke:
```
======= Failed test run #1 ==========
  Expected: <System.ArgumentNullException>
  But was:  <NHibernate.Exceptions.GenericADOException: Could not execute query[SQL: SQL not available] ---> System.ArgumentNullException: Value cannot be null.
```

Is this exception mismatch between different DB backends intentional? This makes exception handling for this case quite difficult.

fredericDelaporte:
> Is this exception mismatch between different DB backends intentional? This makes exception handling for this case quite difficult.

It is accidental. Databases not supporting multi-queries just run them one by one, and the thing handling that does not behave the same for exception handling... Needs an issue on its own I think.

I am going to PR another way of fixing the trouble, moving the post execute result transformer to multi-query.

fredericDelaporte:
#1393 is the another way to fix the issue I propose.

fredericDelaporte:
Thanks for the report and the diagnosis Nico Kreipke.

PR #1392, opened 2017-10-18
Add support for query space synchronization

### LINKED ISSUE #1381: NHibernate's IQuery is missing AddSynchronizedQuerySpace
...and AddSynchronizedEntityName and AddSynchronizedEntityClass. This is crucial when 2nd level cache is enabled and ExecuteUpdate is performed on a native SQL query. Currently this causes a massive cache eviction, when it's often unnecessary.

gliljas:
Ok, this was a bit of a Pandora's box...
Fixes #1381

The problem was that using an `ISQLQuery `and `ExecuteUpdate` causes the entire (sort of) 2nd level cache to be cleared. Not a problem with an in-memory cache provider, but with a distributed cache provider (Redis in our case), it's a massive problem. We ended up having to create ugly wrappers, with native `SqlCommand `etc.

I found this...

https://stackoverflow.com/questions/19054673/2nd-level-cache-invalidation-with-native-queries

...which indicated that Hibernate had a solution for this, but these methods were missing in NHibernate.

The first thing to do was a unit test, to verify the behavior. This is the first assert in 
[NativeSQLBulkOperationsWithCache.cs](https://github.com/gliljas/nhibernate-core/blob/43625d296428a5107a0757d395591f757e80d39c/src/NHibernate.Test/BulkManipulation/NativeSQLBulkOperationsWithCache.cs
) .

My expectation was that it would call the `ICache.Clear` method once (only one queryspace/table in the mapping), but found that it called the method 10 times (!). That's quite costly in a distributed scenario, especially if you have more than one mapped class. We do. :)

The reasons for this are two.

1. There's a method in [NativeSQLQueryPlan.cs](https://github.com/nhibernate/nhibernate-core/blob/4d99b2b9258cb46ba1ba345369b79d83e87f1044/src/NHibernate/Engine/Query/NativeSQLQueryPlan.cs) named `CoordinateSharedCacheCleanup`, called from `PerformExecuteUpdate`, which creates a `BulkOperationCleanupAction `and calls `Init()` on it. This call causes a first wave of `ICache.Clear`:s. I don't understand why a native `ExecuteUpdate `would need to clear the cache before it's run, but it does. The Hibernate version of this looks about the same, but never calls Init(), so the cache is never touched. I haven't touched this yet, since I wanted your input, but it seems like the Init call should be removed.
2.  `BulkOperationCleanupAction `calls `EvictEntity `for every queryspace it finds, but this may very well end up with the same `IPersister `many times, and the same `ICache `just as many times.

Since the unit test has a mapped base class, subclassed into 4, we ended up with 2*5 calls to the `Vehicle `cache.

This was fixed by adding an overload of `EvictEntity `(and its siblings) which takes an `IEnumerable<string>` (entity names), which can be handled as a batch, calling each `ICache `only once. Async versions were also added. Please take a look at the `Task.WhenAny(Task.WhenAll....)` construct, which is used to exit via cancellation. It may be worth discussing.

The `AddSynchronizedQuerySpace `method was added via an `ISynchronizableQuery `interface, just as Hibernate uses [SynchronizeableQuery](https://github.com/hibernate/hibernate-orm/blob/0a2a5c622e3eb30724e80bc8661c0ac55ebfb2be/hibernate-core/src/main/java/org/hibernate/SynchronizeableQuery.java) (notice the spelling, which I think is wrong). Java does things with its generics which I don't think is possible in C#, but I replicated the behavior using a self referencing generic interface. It's somewhat elegant, but also a bit weird. Just using a plain `ISynchronizableQuery `interface would prevent method chaining, but an option could be to simply add the methods to `ISQLQuery`.

I replicated the Hibernate unit test I found for this, which checks whether the session is flushed unnecessarily.

As I mentioned on the developer list, I've added NSubstitute to the project.

1. For the `ICacheProvider `test it wasn't super elegant, but that's also an indication that things in the core code is a bit coupled. 
2. The `BulkOperationCleanupAction `tests were a much better fit, and such tests could be used a lot more.
3. NSubstitute 3 (I'm using v2 in my own projects) apparently has a dependency on NETStandard. It adds a boat load of subdependencies, but it seems to work just fine, after fixing a couple of ambiguous names. If it causes problems, v2 works fine too.


fredericDelaporte:
It may be 5.1 if it has no breaking changes, otherwise 6.0. Currently I am focusing on 5.0.1 and also 5.0 releases of some dependent projects like caches.

gliljas:
5.1 would be great, since it's a performance drag. As far as breaking changes go, it adds new methods to public interfaces (ISQLQuery, ISessionFactory), but some of these could be hidden/internalized in a first release.

src/NHibernate/Impl/SqlQueryImpl.cs
```diff
@@ -170,10 +172,12 @@ public override IList<T> List<T>()
 
 		public NativeSQLQuerySpecification GenerateQuerySpecification(IDictionary<string, TypedValue> parameters)
 		{
+			var allQuerySpaces=new HashSet<string>(querySpaces ?? new List<string>());
+			allQuerySpaces.UnionWith(GetSynchronizedQuerySpaces());
```
hazzik: `NativeSQLQuerySpecification` already does deduplication, so we dont' need to do it here.

src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -42,7 +42,7 @@ public abstract partial class AbstractQueryImpl : IQuery
 		private CacheMode? cacheMode;
 		private CacheMode? sessionCacheMode;
 		private string comment;
-
+
```
fredericDelaporte: Accidental white-space addition.

fredericDelaporte:
The `BulkOperationCleanupAction` has quite changed on Hibernate side. The `Init` method is gone. That was done in [HHH-4034](https://hibernate.atlassian.net/browse/HHH-4034) [commit](https://github.com/hibernate/hibernate-orm/commit/a41506f4041accb707135bd3e77942ba6fb79c83).

The `Init` removal is completed by en `else` for ensuring the cleanup is done, see [here](https://github.com/hibernate/hibernate-orm/commit/a41506f4041accb707135bd3e77942ba6fb79c83#diff-e56a9f545ab9c4e1af92d60fe8d9e39cL158). We should likely do the same, for the native sql case and for the abstract executor case too (see [here](https://github.com/hibernate/hibernate-orm/commit/a41506f4041accb707135bd3e77942ba6fb79c83#diff-2f26a163bcfb21681cac18eb9d55ba60L243)).

gliljas:
@fredericDelaporte  Any suggestion on how to make this suitable for inclusion?

hazzik:
@gliljas at least resolve the conflicts and made the build green.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -967,6 +992,25 @@ public void EvictCollection(string roleName)
...
+		{
+			if (roleNames == null)
+				throw new ArgumentNullException(nameof(roleNames));
+
+			foreach (var cacheGroup in roleNames.Select(GetCollectionPersister).Where(x => x.HasCache).GroupBy(x => x.Cache))
+			{
+			
+				if (log.IsDebugEnabled())
+				{
+					foreach (var p in cacheGroup)
```
hazzik: I would love to see only one line in logs in this case.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -912,6 +919,24 @@ public void EvictEntity(string entityName)
...
+			if (entityNames == null)
+				throw new ArgumentNullException(nameof(entityNames));
+
+			foreach (var cacheGroup in entityNames.Select(GetEntityPersister).Where(x => x.HasCache).GroupBy(x => x.Cache))
+			{
+				if (log.IsDebugEnabled())
+				{
+					foreach (var p in cacheGroup)
+					{
+						log.Debug("evicting second-level cache: " + p.EntityName);
```
hazzik: I would love to see only one line in logs in this case.

src/NHibernate/Impl/SqlQueryImpl.cs
```diff
@@ -320,5 +327,37 @@ protected internal override IEnumerable<ITranslator> GetTranslators(ISessionImpl
...
+			foreach (var querySpace in persister.QuerySpaces)
+			{
+				addedQuerySpaces.Value.Add(querySpace);
+			}
+			return this;
+		}
+
+		public IReadOnlyCollection<string> GetSynchronizedQuerySpaces()
+		{
+			return addedQuerySpaces.IsValueCreated ? addedQuerySpaces.Value : new HashSet<string>();
```
hazzik: The `new HashSet<string>()` kills the purpose of the `Lazy<>` here.

fredericDelaporte: I think the `Lazy<>` is not worth it here. For having it useful, `GetSynchronizedQuerySpaces()` would have to yield `null` instead when it is not initialized. Would it be a private method, why not, but as an interface method, better yield an empty collection.

fredericDelaporte:
I propose gliljas/nhibernate-core#2 for avoiding the breaking changes this PR currently has.

src/NHibernate.Test/BulkManipulation/BulkOperationCleanupActionFixture.cs
```diff
@@ -0,0 +1,71 @@
...
+
+			var target = new BulkOperationCleanupAction(_session, new HashSet<string>(querySpaces.Split(new []{','},StringSplitOptions.RemoveEmptyEntries)));
+
+			target.AfterTransactionCompletionProcess(true);
+
+			Assert.AreEqual(expectedPropertySpaceLength, target.PropertySpaces.Length);
+
+			if (expectedEntityEvictionCount > 0)
+			{
+				_factory.Received(1).EvictEntity(Arg.Is<IEnumerable<string>>(x => x.Count() == expectedEntityEvictionCount));
```
fredericDelaporte: This fails compiling with [Travis](https://travis-ci.org/nhibernate/nhibernate-core/jobs/383615826#L2046).

```
BulkManipulation/BulkOperationCleanupActionFixture.cs(54,38): error CS0012: The type 'Expression<>' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Linq.Expressions, Version=4.0.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. [/home/travis/build/nhibernate/nhibernate-core/src/NHibernate.Test/NHibernate.Test.csproj]
BulkManipulation/BulkOperationCleanupActionFixture.cs(54,66): error CS1660: Cannot convert lambda expression to type 'IEnumerable<string>' because it is not a delegate type [/home/travis/build/nhibernate/nhibernate-core/src/NHibernate.Test/NHibernate.Test.csproj]
```

I have suspected that a reference was missing for .Net Core target, but it does compile fine under Windows. (AppVeyor build is successful so far, it has ended its SQL-Server and PostgreSQL build as of now.)

Moreover [the package it asks to add](https://www.nuget.org/packages/System.Linq.Expressions) looks to me as an obsolete thing: it does still depend on `.NETCore 5.0`, the old naming before they decided to number .Net Core from 1.0.

So this looks as a linux/mono specific issue.

gliljas:
I ran the build locally on WSL, and I think I managed to fix it by updating NSubstitute to 3.1.0.

fredericDelaporte:
The Travis build is fixed. I had thought about upgrading NSubstitue, but looking at the references of both versions, I was having the impression it will not change a thing.

fredericDelaporte:
I have run an async regen just to verify the test due to the async upgrade and configuration change on master, which may impact them.
Looking at them, nothing should have change. But in fact a change was missing in NHibernate code instead, I do not have analyzed why yet.

src/NHibernate/Async/Action/BulkOperationCleanupAction.cs
```diff
@@ -57,37 +59,46 @@ public Task ExecuteAsync(CancellationToken cancellationToken)
...
+			}
+			catch (Exception ex)
+			{
+				return Task.FromException<object>(ex);
 			}
 		}
 
 		#endregion
-
-		public virtual async Task InitAsync(CancellationToken cancellationToken)
```
hazzik: "Since the class is not internal, this is a breaking change. It needs to be added back, obsoleted, in the non-generated partial."

src/NHibernate/Async/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
```diff
@@ -34,19 +34,6 @@ public abstract partial class AbstractStatementExecutor : IStatementExecutor
 
 		public abstract Task<int> ExecuteAsync(QueryParameters parameters, ISessionImplementor session, CancellationToken cancellationToken);
 
-		protected virtual async Task CoordinateSharedCacheCleanupAsync(ISessionImplementor session, CancellationToken cancellationToken)
```
fredericDelaporte: Yes, and same for this one.

src/NHibernate/ISQLQuery.cs
```diff
@@ -1,7 +1,69 @@
...
+	{
+		/// <summary>
+		/// Adds a query space for auto-flush synchronization and second level cache invalidation.
+		/// </summary>
+		/// <param name="sqlQuery">The query.</param>
+		/// <param name="querySpace">The query space.</param>
+		/// <returns>The query.</returns>
+		public static ISQLQuery AddSynchronizedQuerySpace(this ISQLQuery sqlQuery, string querySpace)
+		{
+			if (!(sqlQuery is ISynchronizableSQLQuery ssq))
```
hazzik: @fredericDelaporte I think then it shall be `CastOrThrow`

src/NHibernate/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
```diff
@@ -47,12 +49,25 @@ protected virtual void CoordinateSharedCacheCleanup(ISessionImplementor session)
...
 			}
+			else
+			{
+				action.AfterTransactionCompletionProcess(true);
+			}
+		}
+
+		// Since v5.2
+		[Obsolete("This method has no more actual async calls to do, use its sync version instead.")]
+		protected virtual Task CoordinateSharedCacheCleanupAsync(ISessionImplementor session, CancellationToken cancellationToken)
```
hazzik: Need either `async` keyword, or wrapping into try-catch.

src/NHibernate/ISynchronizableQuery.cs
```diff
@@ -0,0 +1,34 @@
+using System.Collections.Generic;
+
+namespace NHibernate
+{
+	public interface ISynchronizableQuery<out T> where T : ISynchronizableQuery<T>
```
hazzik: Do we plant to have other types of `ISynchronizableQuery<>` other than `ISQLQuery`?

fredericDelaporte: Looking at Hibernate code, it seems not. They use it for `SQLQuery` (obsoleted on their side) and `NativeQuery` (replacement for `SQLQuery`). So why not merging that in `ISynchronizableSQLQuery` without the generic param trick.

PR #1393, opened 2017-10-19
Fix Linq Future aggregates failures, fixes #1387

### LINKED ISSUE #1387: Linq Sum() with ToFutureValue fails

fredericDelaporte:
This is an alternate fix for #1387, aiming at replacing #1388.

It includes the tests added in #1388, but fix the trouble with another way. It is required for fixing all cases, especially all the NH-3850 tests once rewritten for testing both non-future and future queries.

### DIFF
--- a/src/NHibernate/Impl/FutureBatch.cs
+++ b/src/NHibernate/Impl/FutureBatch.cs
@@ -1,15 +1,21 @@
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
-using System.Threading.Tasks;
+using NHibernate.Transform;
 
 namespace NHibernate.Impl
 {
 	public abstract partial class FutureBatch<TQueryApproach, TMultiApproach>
 	{
-		private readonly List<TQueryApproach> queries = new List<TQueryApproach>();
-		private readonly IList<System.Type> resultTypes = new List<System.Type>();
+		private class BatchedQuery
+		{
+			public TQueryApproach Query { get; set; }
+			public System.Type ResultType { get; set; }
+			public IDelayedValue Future { get; set; }
+		}
+
+		private readonly List<BatchedQuery> queries = new List<BatchedQuery>();
 		private int index;
 		private IList results;
 		private bool isCacheable = true;
@@ -29,8 +35,7 @@ namespace NHibernate.Impl
 				cacheRegion = CacheRegion(query);
 			}
 
-			queries.Add(query);
-			resultTypes.Add(typeof(TResult));
+			queries.Add(new BatchedQuery { Query = query, ResultType = typeof(TResult) });
 			index = queries.Count - 1;
 			isCacheable = isCacheable && IsQueryCacheable(query);
 			isCacheable = isCacheable && (cacheRegion == CacheRegion(query));
@@ -44,26 +49,44 @@ namespace NHibernate.Impl
 		public IFutureValue<TResult> GetFutureValue<TResult>()
 		{
 			int currentIndex = index;
-			return new FutureValue<TResult>(() => GetCurrentResult<TResult>(currentIndex), cancellationToken => GetCurrentResultAsync<TResult>(currentIndex, cancellationToken));
+			var future = new FutureValue<TResult>(
+				() => GetCurrentResult<TResult>(currentIndex),
+				cancellationToken => GetCurrentResultAsync<TResult>(currentIndex, cancellationToken));
+			var query = queries[currentIndex];
+			query.Future = future;
+			return future;
 		}
 
 		public IFutureEnumerable<TResult> GetEnumerator<TResult>()
 		{
 			var currentIndex = index;
-			return new DelayedEnumerator<TResult>(() => GetCurrentResult<TResult>(currentIndex), cancellationToken => GetCurrentResultAsync<TResult>(currentIndex, cancellationToken));
+			var future = new DelayedEnumerator<TResult>(
+				() => GetCurrentResult<TResult>(currentIndex),
+				cancellationToken => GetCurrentResultAsync<TResult>(currentIndex, cancellationToken));
+			var query = queries[currentIndex];
+			query.Future = future;
+			return future;
 		}
 
 		private IList GetResults()
 		{
 			if (results != null)
-			{
 				return results;
-			}
+
 			var multiApproach = CreateMultiApproach(isCacheable, cacheRegion);
-			for (int i = 0; i < queries.Count; i++)
+			var needTransformer = false;
+			foreach (var query in queries)
 			{
-				AddTo(multiApproach, queries[i], resultTypes[i]);
+				AddTo(multiApproach, query.Query, query.ResultType);
+				if (query.Future?.ExecuteOnEval != null)
+					needTransformer = true;
 			}
+
+			if (needTransformer)
+				AddResultTransformer(
+					multiApproach, 
+					new FutureResultsTransformer(queries));
+
 			results = GetResultsFrom(multiApproach);
 			ClearCurrentFutureBatch();
 			return results;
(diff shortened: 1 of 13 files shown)


src/NHibernate/Impl/FutureBatch.cs
```diff
@@ -80,5 +116,88 @@ private IEnumerable<TResult> GetCurrentResult<TResult>(int currentIndex)
...
+					{
+						results.Add(element);
+					}
+					return results;
+				}
+
+				// When not null on a future value, ExecuteOnEval is fetched with PostExecuteTransformer from
+				// IntermediateHqlTree through ExpressionToHqlTranslationResults, which requires a IQueryable
+				// as input and directly yields the scalar result when the query is scalar.
+				var resultElement = postExecute.ExecuteOnEval.DynamicInvoke(collection.AsQueryable());
```
fredericDelaporte: No need to type the collection for `AsQueryable` if its underlying type is already the good one, which is done in multi-query.

src/NHibernate/Impl/MultiQueryImpl.cs
```diff
@@ -458,30 +458,37 @@ protected void After()
...
-					resultCollections.Add(new List<object>());
-				}
-				else
-				{
-					resultCollections.Add(Activator.CreateInstance(typeof(List<>).MakeGenericType(resultCollectionGenericType[i])));
-				}
+				var query = queries[i] as ExpressionQueryImpl;
+				// Linq queries may override the query type, finishing the work with a post execute transformer,
+				// which with multi queries are executed through the multi-query result trasformer.
+				var rawElementType = query?.QueryExpression?.Type ?? resultCollectionGenericType[i];
```
fredericDelaporte: Grabbing here the Linq result type override if any, for typing the list before transformation with nullable if required.

src/NHibernate/Impl/MultiQueryImpl.cs
```diff
@@ -458,30 +458,37 @@ protected void After()
...
+			}
+
+			var multiqueryHolderInstatiator = GetMultiQueryHolderInstatiator();
+			var resultCollections = new List<object>(resultCollectionGenericType.Count);
+			for (var i = 0; i < queries.Count; i++)
+			{
+				// Once polymorpic queries aggregated in one result per query (previous loop), use the
+				// MultiQueryTransformer using, as source, the aggregated result.
+				var resultList = GetTransformedResults(rawResultCollections[i], multiqueryHolderInstatiator);
+				resultCollections.Add(resultList);
```
fredericDelaporte: We already have a list of target type here, no need to redo list instanciation + `ArrayHelper.AddAll`.

src/NHibernate/Impl/FutureBatch.cs
```diff
@@ -80,5 +116,88 @@ private IEnumerable<TResult> GetCurrentResult<TResult>(int currentIndex)
...
+
+			public IList TransformList(IList collection)
+			{
+				if (_currentIndex >= _postExecutes.Count)
+					throw new InvalidOperationException(
+						$"Transformer have been called more times ({_currentIndex + 1}) than it has queries to transform.");
+
+				var postExecute = _postExecutes[_currentIndex];
+				_currentIndex++;
+				if (postExecute.ExecuteOnEval == null)
```
hazzik: I would like to encapsulate this logic into PostExecute class

hazzik: I’ve made a slight refactoring. I’ll update the branch shortly.

hazzik:
Hi @fredericDelaporte please review my commit, and if it is ok - integrate

src/NHibernate/Impl/FutureBatch.cs
```diff
@@ -154,38 +133,14 @@ public object TransformTuple(object[] tuple, string[] aliases)
...
-				}
-
-				// When not null on a future value, ExecuteOnEval is fetched with PostExecuteTransformer from
-				// IntermediateHqlTree through ExpressionToHqlTranslationResults, which requires a IQueryable
-				// as input and directly yields the scalar result when the query is scalar.
-				var resultElement = postExecute.ExecuteOnEval.DynamicInvoke(collection.AsQueryable());
-				results.Add(resultElement);
 
-				return results;
+				return batchedQuery.Future?.TransformList(collection);
```
fredericDelaporte: In fact I do not think we may have cases where `Future` will be null, but I have not checked all code paths to be sure, so for a patch I have chosen to be on the safe side.
But then we requires here to yield the collection, not `null`.
I will integrate by changing this to:
```c#
return batchedQuery.Future?.TransformList(collection) ?? collection;
```
This may hide/obfuscate a failure case (`TransformList` yielding `null`), but that is a quite unlikely failure.

hazzik:
I reworded the commit messages to make them more natural as we are not dependant on JIRA now. The idea is to mention "Fixes #1387" in a merge commit.

src/NHibernate/Impl/MultiQueryImpl.cs
```diff
@@ -458,58 +458,54 @@ protected void After()
...
+				// Once polymorpic queries aggregated in one result per query (previous loop), use the
+				// MultiQueryTransformer using, as source, the aggregated result.
+				var resultList = GetTransformedResults(rawResultCollections[i]);
+				resultCollections.Add(resultList);
 			}
 
 			return resultCollections;
 		}
 
-		private IList GetTransformedResults(IList source, HolderInstantiator holderInstantiator)
```
hazzik: @fredericDelaporte I've removed HolderInstantiator here as it just adds the overhead without providing much value

src/NHibernate/Impl/MultiQueryImpl.cs
```diff
@@ -458,58 +458,54 @@ protected void After()
...
+		private  IList GetTransformedResults(IList source)
 		{
-			if (!holderInstantiator.IsRequired)
-			{
+			if (resultTransformer == null)
 				return source;
-			}
-			for (int j = 0; j < source.Count; j++)
+
+			//MultiCriteria does not call TransformTuple here
```
fredericDelaporte: MultiCriteria and MultiQuery looks as almost dups which have diverged on some points. Probably we need the issue on having a common query interface to be done for refactoring this.

hazzik: It's in #792

src/NHibernate/Impl/FutureBatch.cs
```diff
@@ -80,5 +116,88 @@ private IEnumerable<TResult> GetCurrentResult<TResult>(int currentIndex)
...
+			public System.Type ResultType { get; set; }
+			public Delegate ExecuteOnEval { get; set; }
+			public bool IsValue { get; set; }
+		}
+
+		// ResultTransformer are usually re-usable, this is not the case of this one, which will
+		// be built for each multi-query requiring it.
+		// It also usually ends in query cache, but this is not the case either for multi-query.
+		[Serializable]
+		private class FutureResultsTransformer : IResultTransformer
```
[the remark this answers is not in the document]
ngbrown: Since `FutureResultsTransformer` isn't actually serializable because `BatchedQuery` isn't serializable, can the test in `DetachedCriteriaSerializable` be adjusted to exclude this type instead?  `DetachedCriteria` doesn't expose a `Future`, so not all `IResultTransformer` technically need to be serializable.

Why is serializing a `DetachedCriteria` so important anyway?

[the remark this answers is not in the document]
fredericDelaporte: Originally I have made that class truly serializable, but that was requiring an additional "data holding" class just for that.

We could instead try to adjust the serializable test, but since this class is private, we will have to add some more reflection logic matching by name to do so. Does that `Serializable` attribute hurts somewhere that it would be better to remove it?

About `DetachedCriteria`, I have not checked why it actually needs to be serializable, but I think it is likely because it may ends in the query cache too.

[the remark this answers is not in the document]
ngbrown: Since serialization is the biggest compatibility break with .NET Core, I am writing some Roslyn analyzers to find issues so I can try to resolve them ahead of time.  On types that are required to be serialized for some reason, and use `Type` or `MethodInfo`, etc., I'll probably have to implement `ISerializable`.

Of course this `FutureResultsTransformer ` is triggering a full-framework warning which is based on FxCop rule [CA2235](https://docs.microsoft.com/visualstudio/code-quality/ca2235-mark-all-non-serializable-fields).  Marking the field as `[NonSerializable]` is really just masking the problem because the entire type wouldn't work if it were deserialized like that.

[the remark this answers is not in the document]
fredericDelaporte: It seems to me most of those `Serializable` things everywhere in NHibernate code are here for allowing to serialize sessions and session factories. But is there any practical common use case in serializing sessions? Do a significant parts of NHibernate user need this?

Maybe should we consider dropping the possibility of serializing sessions and session factories.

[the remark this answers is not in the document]
fredericDelaporte: I found [this](https://nhibernate.jira.com/browse/NH-3487), which shows there are users using session serialization. It looks like we need to keep that.

Issue #1391, opened 2017-10-18
Performance regression in SessionIdLoggingContext


maca88:
By doing some profiling, I discovered that the introduction of [`AsyncLocal<>`](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Impl/SessionIdLoggingContext.cs#L8) inside the `SessionIdLoggingContext` caused a performance regression of approximately 6.5% when retrieving a bigger result set.

Here is the test that I ran, which uses the data provided by the `LinqTestCase` class.
```
[Test]
public void PerformanceTest()
{
    var list =
            db.Orders
              .FetchMany(o => o.OrderLines)
              .ThenFetch(o => o.Product)
              .ThenFetch(o => o.Category)
              .ToList();
    Assert.AreEqual(830, list.Count);
}
```
The test completed in 1557ms, where 97ms was spent inside `SessionIdLoggingContext`:
![asynclocal](https://user-images.githubusercontent.com/5311349/31735987-baa5efc4-b443-11e7-92c1-757a328b053b.png)

By replacing `AsyncLocal` with `ThreadLocal` as it was prior the async addition, the `SessionIdLoggingContext` disappears from the radar, which proves that `AsyncLocal` is way slower than `ThreadLocal` and we should use it only when necessary.

In order to avoid those penalities for consumers that don't use it, I think that we should provide a way to disable it (e.g. by adding a simple flag on the class itself).

hazzik:
@maca88 there are some issues with the generation of async tests for that PR. Could you please take a look?

maca88:
Upgrading to `0.6.2` will fix [the problem](https://github.com/maca88/AsyncGenerator/issues/71).

hazzik:
Thanks

PR #1394, opened 2017-10-19
Added the ability to disable capturing the current session id

maca88:
Here is one way to solve #1391.


src/NHibernate/Impl/SessionIdLoggingContext.cs
```diff
@@ -11,11 +11,23 @@ public class SessionIdLoggingContext : IDisposable
...
 			SessionId = id;
 		}
 
 		/// <summary>
+		/// Determines whether the capturing of the current session is enabled or not.
+		/// When false, <see cref="SessionIdLoggingContext.SessionId"/> will always be null.
+		/// In case the current session id won't be used, is strongly recommended to disable it, in order to increase performance.
+		/// By default is set to true.
+		/// </summary>
+		public static bool Enabled { get; set; } = true;
```
fredericDelaporte: So we should change the default to `false` for 6.0 (small breaking change).

src/NHibernate.Test/NHSpecificTest/GH1391/Fixture.cs
```diff
@@ -0,0 +1,71 @@
...
+	{
+		[Test]
+		public void MultipleThreadsNested()
+		{
+			Parallel.For(0, 100, i =>
+			{
+				var guid = Guid.NewGuid();
+				using (new SessionIdLoggingContext(guid))
+				{
+					Assert.AreEqual(guid, SessionIdLoggingContext.SessionId);
```
fredericDelaporte: This is NUnit classic model of assertion. We are favoring the constraint model for new tests. See [two models](https://github.com/nunit/docs/wiki/Assertions#two-models).

maca88:
@fredericDelaporte I've modified the configuration parameter to be set in an usual way. The difference is that now the setting is set per session factory instead beign a global one, which is IMO more correct, but it introduced a [new public member](https://github.com/nhibernate/nhibernate-core/pull/1394/commits/b8198448c5f014e48962ab4b99787e5d28bb35a5#diff-4f54cd4cf1deac5a24237ca0e0af5bb5R265) in `ISessionImplementor.` Also, `SessionId` will be now `Guid.Empty` instead of `null`, when tracking is disabled.

src/NHibernate.Test/NHSpecificTest/GH1391/Fixture.cs
maca88: [Fixed](https://github.com/nhibernate/nhibernate-core/pull/1394/commits/b8198448c5f014e48962ab4b99787e5d28bb35a5#diff-1e66d0f683fb89f2119cd220860a4402R29)

src/NHibernate/Engine/ISessionImplementor.cs
```diff
@@ -258,6 +258,11 @@ public partial interface ISessionImplementor
 		FlushMode FlushMode { get; set; }
 
 		string FetchProfile { get; set; }
+		
+		/// <summary>
+		/// Determines whether <see cref="SessionId"/> is tracked by <see cref="SessionIdLoggingContext"/> or not.
+		/// </summary>
+		bool IdTracking { get; }
```
fredericDelaporte: I was not seeing that this way. `SessionId` only serves for logging, just set it to `Guid.Empty` in the session itself when the option disable it. No massive change. No potential breaking change by adding an interface member (although probably very few implements their own `ISessionImplementor`).

src/NHibernate/Impl/SessionIdLoggingContext.cs
```diff
@@ -8,10 +8,12 @@ public class SessionIdLoggingContext : IDisposable
...
 
 		private readonly Guid? _oldSessonId;
+		private static bool _tracking;
 
-		public SessionIdLoggingContext(Guid id)
+		public SessionIdLoggingContext(Guid id, bool track)
 		{
-			if (!Enabled)
+			_tracking = track;
+			if (!_tracking)
```
fredericDelaporte: I would have just tested `id == Guid.Empty` instead of adding the tracking flag.

src/NHibernate/Engine/ISessionImplementor.cs
maca88: My bad, I didn't read your comment good enough. I made an [another version](https://github.com/nhibernate/nhibernate-core/pull/1394/commits/540eb1813e0df36e6b232dcef69481d2b0607be1) that takes into account your suggestions.

src/NHibernate/Impl/SessionIdLoggingContext.cs
maca88: The tracking flag is needed for the `Dispose` method, otherwise we have to check the `SessionId` that is slow due to `AsyncLocal`.

fredericDelaporte: Yes for the one in SessionIdLoggingContext, I should have been more precise and write `track`, the one coming from the added boolean on `ISessionImplementor`.

src/NHibernate/Impl/SessionIdLoggingContext.cs
```diff
@@ -8,9 +8,15 @@ public class SessionIdLoggingContext : IDisposable
 		private static readonly AsyncLocal<Guid?> _currentSessionId = new AsyncLocal<Guid?>();
 
 		private readonly Guid? _oldSessonId;
+		private static bool _tracking;
```
fredericDelaporte: `static` is probably forgotten from previous version, it needs to be removed.

src/NHibernate/nhibernate-configuration.xsd
```diff
@@ -193,6 +193,18 @@
...
 									</xs:annotation>
 								</xs:enumeration>
+								<xs:enumeration value="track_session_id">
+									<xs:annotation>
+										<xs:documentation>
+											Set whether tracking the session id or not. When true, each session will have an unique Guid
+											that can be retrieved by ISessionImplementor.SessionId, otherwise ISessionImplementor.SessionId will
+											always be Guid.Empty. In addition, when tracking is enabled, the current session id will be stored in
+											SessionIdLoggingContext.SessionId. In case the current session id won't be used, is strongly recommended
+											to disable it, in order to increase performance.
```
fredericDelaporte: `, is strongly recommended` should be `, it is strongly recommended`, and I would personally just write `, it is recommended`.

src/NHibernate/Cfg/Environment.cs
```diff
@@ -258,6 +258,17 @@ public static string Version
 		/// </remarks>
 		public const string OracleUseNPrefixedTypesForUnicode = "oracle.use_n_prefixed_types_for_unicode";
 
+		/// <summary>
+		/// <para>Set whether tracking the session id or not. When <see langword="true"/>, each session 
+		/// will have an unique <see cref="Guid"/> that can be retrieved by <see cref="ISessionImplementor.SessionId"/>,
+		/// otherwise <see cref="ISessionImplementor.SessionId"/> will always be <see cref="Guid.Empty"/>.</para>
+		/// In addition, when tracking is enabled, the current session id will be stored in 
+		/// <see cref="NHibernate.Impl.SessionIdLoggingContext.SessionId"/>.
+		/// In case the current session id won't be used, is strongly recommended to disable it, in order to increase performance.
```
fredericDelaporte: `, is strongly recommended` should be `, it is strongly recommended`, and I would personally just write `, it is recommended`.

src/NHibernate/Cfg/Environment.cs
```diff
@@ -258,6 +258,17 @@ public static string Version
 		/// </remarks>
 		public const string OracleUseNPrefixedTypesForUnicode = "oracle.use_n_prefixed_types_for_unicode";
 
+		/// <summary>
+		/// <para>Set whether tracking the session id or not. When <see langword="true"/>, each session 
+		/// will have an unique <see cref="Guid"/> that can be retrieved by <see cref="ISessionImplementor.SessionId"/>,
+		/// otherwise <see cref="ISessionImplementor.SessionId"/> will always be <see cref="Guid.Empty"/>.</para>
```
fredericDelaporte: I would add something like "Session id is used for logging purpose."

src/NHibernate/Cfg/Loquacious/IDbIntegrationConfigurationProperties.cs
```diff
@@ -25,6 +25,7 @@ public interface IDbIntegrationConfigurationProperties
 		void Batcher<TBatcher>() where TBatcher : IBatcherFactory;
 		short BatchSize { set; }
 		bool OrderInserts { set; }
+		bool TrackSessionId { set; }
```
fredericDelaporte: I would more see that among session factory settings, on IFluentSessionFactoryConfiguration, as a method taking a boolean.

I have to say I never updated the loquacious cfg when adding parameters, and it is likely there are a bunch of them lacking.

Moreover looking at it, it seems that we should reflect changes done on the extension based `IDbIntegrationConfigurationProperties` to the fluent `IDbIntegrationConfiguration` too.

Since it is a public interface by the way, this is technically a breaking change. Should we do it for a revision fix?

src/NHibernate/Cfg/Environment.cs
maca88: [Added](https://github.com/nhibernate/nhibernate-core/pull/1394/commits/8ca4ba8e4a32a800bf2a48de9b371387e4f034ae#diff-2063778a671420f6e1bceee0dd202b1bR264).

src/NHibernate/Cfg/Environment.cs
maca88: Corrected.

src/NHibernate/Impl/SessionIdLoggingContext.cs
maca88: [Replaced](https://github.com/nhibernate/nhibernate-core/pull/1394/commits/8ca4ba8e4a32a800bf2a48de9b371387e4f034ae#diff-582ed89a24df0b05c4a1198ffb0f8e1cR11) static with readonly.

src/NHibernate/Cfg/Loquacious/IDbIntegrationConfigurationProperties.cs
maca88: I removed the member from `IDbIntegrationConfigurationProperties` in order to avoid a breaking change. IMO , the loquacious cfg should be updated in a separate PR, where the missing parameters would be added.

fredericDelaporte:
It seems "Allow edit from maintainers" is not enabled. I was willing to add
```xml
                    <row>
                        <entry>
                            <literal>track_session_id</literal>
                        </entry>
                        <entry>
                            Set whether the session id should be tracked in logs or not. When <literal>true</literal>, each
                            session will have an unique <literal>Guid</literal> that can be retrieved with
                            <literal>ISessionImplementor.SessionId</literal>, otherwise <literal>ISessionImplementor.SessionId</literal>
                            will be <literal>Guid.Empty</literal>.
                            <para>
                                Session id is used for logging purpose and can also be retrieved on the static property
                                <literal>NHibernate.Impl.SessionIdLoggingContext.SessionId</literal>, when tracking is enabled.
                            </para>
                            <para>
                                Disabling tracking by setting <literal>track_session_id</literal> to <literal>false</literal>
                                increases performances. Default is <literal>true</literal>.
                            </para>
                            <para>
                                <emphasis role="strong">eg.</emphasis>
                                <literal>true</literal> | <literal>false</literal>
                            </para>
                        </entry>
                    </row>
```

In `doc/reference/modules/configuration.xml`, at line 1030.

maca88:
> It seems "Allow edit from maintainers" is not enabled.
Strange, the checkbox is checked. I've double clicked it, so please try again if it works.

fredericDelaporte:
```
E:\Projets\nhibernate\nhibernate-core> git push
remote: Permission to maca88/nhibernate-core.git denied to fredericDelaporte.
fatal: unable to access 'https://github.com/maca88/nhibernate-core/': The requested URL returned error: 403
```
Issuing a `git remote -v` does not show me anything special with your remote, compared to my other remotes.

So maybe just commit it yourself.

maca88:
> I wonder, does the get on SessionId underlying AsyncLocal have no performance hit when it has not yet been set?

You will always get the performance hit when accessing the `Value` property from `AsyncLocal,` even if it was never been set. `AsyncLocal` is trying to get the value from the current execution context, which is expensive to get. The getter of the `Value` property is faster that the setter as seen in #1394 (6ms vs 19ms), so there is a smaller penality for it.

> Otherwise we may need something more elaborated.

The problem here is, that at least I don't see how we could achieve having a static getter without the `AsyncLocal` penalities.

fredericDelaporte:
>The problem here is, that at least I don't see how we could achieve having a static getter without the AsyncLocal penalities.

It could be:

```c#
private static volatile AsyncLocal<Guid?> _currentSessionId;
private static readonly object _lock = new object();

...

public static Guid? SessionId
{
	get => _currentSessionId?.Value;
	set
	{
		if (_currentSessionId == null)
		{
			lock (_lock)
			{
				if (_currentSessionId == null)
				{
					_currentSessionId = new AsyncLocal<Guid?>();
				}
			}
		}
		_currentSessionId.Value = value;
	}
}
```

But is it worth it?

maca88:
This would solve most of our problems, but not all :). One example where this wouldn't work, is when the consumer is using two session factories, where one has the tracking enabled and the other disabled, which is very rare to occur. 

> But is it worth it?

Probably yes, the majority of consumers are using just one session factory in their applications, so by setting the `track_session_id` to `false`, without removing the usages of the `SessionId` from the logs, they will not suffer the penalities.

fredericDelaporte:
> Probably yes

I am also concerned by having to use a `volatile`, since we read it without memory barriers (no lock). Otherwise it could be possible during `AsyncLocal` instanciation in release mode, that the variable already get assigned while the `AsyncLocal` constructor has not yet finished its initialization. (Some devs even use an intermediate local variable for constructing, but without the `volatile`, it can be optimized out.) I have read somewhere this trouble with that "double check locking" pattern can no more occur with .Net Framework 4, but well not ascertained. What is certain is that I have already witnessed that trouble many years ago.

maca88:
`volatile` seems to be required for [some processors](https://stackoverflow.com/a/1964832), even in the offical [documentation](https://msdn.microsoft.com/en-us/library/ff650316.aspx?f=255&MSPPError=-2147217396) the `volatile` is used. However using a `Lazy<>` with `isThreadSafe` parameter set to `true,` seems to be [faster](https://dotnetfiddle.net/ssA4F0) that the double check locking pattern. By using `Lazy<>`:

```
private static readonly Lazy<AsyncLocal<Guid?>> _currentSessionId = 
    new Lazy<AsyncLocal<Guid?>>(() => new AsyncLocal<Guid?>(), true);
...
public static Guid? SessionId
{
    get => _currentSessionId.Value.Value;
    set => _currentSessionId.Value.Value = value;
}

```
If you agree, I've would change the code to use `Lazy<>`.

fredericDelaporte:
I was discarding `Lazy<>` believing falsely the get would forcibly initialize it, but your right, way better to use it.

src/NHibernate.Test/NHSpecificTest/GH1391/Fixture.cs
```diff
@@ -0,0 +1,154 @@
...
+				}
+				else
+				{
+					Disabled();
+				}
+			});
+		}
+
+		[Test]
+		public void ConcurrentAsync()
```
hazzik: Shouldn't the "...Async" tests be generated?

maca88: Currently, the generator is not able to convert a [Parallel.For](https://github.com/maca88/AsyncGenerator/issues/63) and [Thread.Sleep](https://github.com/maca88/AsyncGenerator/issues/29), but I will make sure that those will be possible with the next release.

hazzik:
<🤓>technically this is not a bug, so it could not be included in 5.0.1</🤓>

hazzik:
To be honest I quite dislike the `SessionIdLoggingContext` itself. I think this feature shall be implemented using the different means, probably on a logger level.

maca88:
> To be honest I quite dislike the SessionIdLoggingContext itself

I can agree here, NHibernate doesn't need to have a class like `SessionIdLoggingContext.` The consumer could use [GetCurrentSession](https://github.com/nhibernate/nhibernate-core/blob/4d99b2b9258cb46ba1ba345369b79d83e87f1044/src/NHibernate/ISessionFactory.cs#L219) method from `ISessionFactory` to get the current session or create its own `SessionIdLoggingContext`.

By looking at Hibernate source, [I didn't found](https://github.com/hibernate/hibernate-orm/search?utf8=%E2%9C%93&q=getSessionIdentifier&type=) anything like `SessionIdLoggingContext`, so maybe we should get rid of it, somewhere in the future.

hazzik:
The history behind `SessionIdLoggingContext` is that @ayende has added it for his own needs of NHProf.

ayende:
The idea is that you can correlate the log messages to the right session.
This is done so offline log analysis can be done, without having access to the process.

fredericDelaporte:
Having an option to disable it does sound rather good, and it should even be disabled by default.
(But strictly speaking this would be a breaking change: should we forbid us such changes within a minor/patch, or should we show some flexibility about them, especially when there is a way to restore the behavior we have change?)

fredericDelaporte:
Regenerating async now generates an async test case for GH1391. This is probably due to the 0.6 upgrade of AsyncGenerator, which has happened before merging this PR but was not taken into account in this PR.

maca88:
This is [now fixed](https://github.com/maca88/AsyncGenerator/issues/71) in version `0.6.2`.

PR #1395, opened 2017-10-20
Documentation fixes

### LINKED ISSUE #1386: Lack of custom logging documentation (was Serilog Tracing)
How can I configure Serilog for logging NHibernate?

fredericDelaporte:
Remove references to an obsolete configuration section since NHibernate 2.0.0 (NH-1087), and clarifies system-level properties.

Fixes #1386.

### DIFF
--- a/doc/reference/modules/configuration.xml
+++ b/doc/reference/modules/configuration.xml
@@ -1091,6 +1091,33 @@ in the parameter binding.</programlisting>
                             </para>
                         </entry>
                     </row>
+                    <row>
+                        <entry>
+                            <literal>nhibernate-logger</literal>
+                        </entry>
+                        <entry>
+                            The class name of an <literal>ILoggerFactory</literal> implementation. It allows using another
+                            logger than log4net.
+                            <para>
+                                The default is not defined, which causes NHibernate to search for log4net assembly. If this
+                                search succeeds, NHibernate will log with log4net. Otherwise, its internal logging will
+                                be disabled.
+                            </para>
+                            <para>
+                                This is a very special system-level property. It can only be set through an
+                                <ulink url="https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/file-schema/appsettings/">appSetting</ulink>
+                                named <literal>nhibernate-logger</literal> in the application configuration file. It cannot
+                                be set neither with <literal>NHibernate.Cfg.Environment</literal> class, nor be defined in the
+                                <literal>&lt;hibernate-configuration&gt;</literal> section of the application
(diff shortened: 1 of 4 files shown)


fredericDelaporte:
The cache settings are documented as of NHibernate.Caches 5.1.0, which I intend to publish soon.

doc/reference/modules/configuration.xml
```diff
@@ -1092,6 +1091,33 @@ in the parameter binding.</programlisting>
...
+                        <entry>
+                            The class name of an <literal>ILoggerFactory</literal> implementation. Allows to use another
+                            logger than log4net.
+                            <para>
+                                Default is not defined, which causes NHibernate to search for log4net assembly. If this
+                                search succeed, NHibernate will log with log4net. Otherwise, its internal logging will
+                                be disabled.
+                            </para>
+                            <para>
+                                This is a very special system-level property. It can only bet set through an
```
hazzik: `bet`?

fredericDelaporte: Oops...

fredericDelaporte: No bet anymore.

fredericDelaporte:
Rebased, some more typo fixed, backported css changes.

hazzik:
@fredericDelaporte I think the PRs need to be as targeted as possible (eg, try to fix one issue at a time)

fredericDelaporte:
I tend to group by subjects (here a bunch of documentation fixes) due to those factors:
* Builds are long. We have currently more than 24h of builds queued according to TeamCity. Each PR triggers its build, while commits grouped in a PR may not do that (if not pushed one by one with waiting for the build to start on each).
* Stale PR merges are blocked, so we have to update them at each merge of other PR (or use admin privilege). The more we have PR, the more updates we do on them, the more builds we add.

Having separated commits still allows to distinguish the sub-subjects.

Doing 4 to 6 PR instead of this single one looks to me as a bit overkill for those documentation fixes.

ngbrown:
Maybe the TeamCity server could skip all the per-db-server tests if the changes are just documentation.  [This could be done with VCS trigger rules](https://confluence.jetbrains.com/display/TCD10/Configuring+VCS+Triggers#ConfiguringVCSTriggers-TriggerRulesExamples).  The GitHub "Required" tests may have to be scaled back to prevent merges from being blocked.

hazzik:
I've excluded `doc/**` from the DB builds as @ngbrown suggested

fredericDelaporte:
Thanks for the fixes, rebased and squashed.

PR #1401, opened 2017-10-24
Clean up db tests dependencies

fredericDelaporte:
Aims at simplifying dependencies management.

Use NuGet for database tests when possible
 * FirebirdSql.Data.FirebirdClient
 * Microsoft.SqlServer.Compact
 * MySql.Data
 * Npgsql (3.2.4.1, due to 3.2.5 no more supporting NHibernate closing connection from distributed transaction second phase: maybe we should add a condition for not testing that instead)
 * Oracle.ManagedDataAccess (12.1.2400, newer versions (at least up to 12.2.1100) fail NH-1171 tests, apparently due to some old Oracle bug striking back https://stackoverflow.com/a/7542670/1178314 )
 * System.Data.SQLite.Core

Folder `available-test-configurations` of NH devs boxes need to be cleaned accordingly. (Remove binaries matching those packages.)


src/NHibernate.sln.DotSettings
```diff
@@ -19,6 +19,7 @@
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateConstants/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="aaBb" /&gt;</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateInstanceFields/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="_" Suffix="" Style="aaBb" /&gt;</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateStaticReadonly/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb" /&gt;</s:String>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpAttributeForSingleLineMethodUpgrade/@EntryIndexedValue">True</s:Boolean>
```
fredericDelaporte: Rider insists at doing that since its last update.

ShowBuildMenu.bat
```diff
@@ -29,117 +30,102 @@ echo --- Exit ---
...
 set LIB_FILES=
 set LIB_FILES2=
 goto test-setup-generic
 
 :test-setup-sqlservercex86
 set CONFIG_NAME=SqlServerCe32
-set PLATFORM=AnyCPU
+set TEST_PLATFORM=AnyCPU
 set LIB_FILES=lib\teamcity\SqlServerCe\*.dll
 set LIB_FILES2=lib\teamcity\SqlServerCe\X86\*.dll
```
ngbrown: You can remove these lines for the SQL Server Compact native files by linking to the files from the test .csproj, and copying to output, and also including a PackageReference to Microsoft.SqlServer.Compact.

Alternatively, a custom `.target` file can be used to pull the files from the `$(NuGetPackageRoot)` and copy them to output.

ngbrown: Never mind about the `.target`, that's just how I automate the whole thing for subsequent dependencies.  For just a single test project, include this:

```xml
  <ItemGroup>
    <PackageReference Include="Microsoft.SqlServer.Compact" Version="4.0.8876.1" />
  </ItemGroup>
  <ItemGroup '$(NuGetPackageRoot)' != '' ">
      <NativeBinaries Include="$(NuGetPackageRoot)microsoft.sqlserver.compact\4.0.8876.1\NativeBinaries\**\*.*" />
      <Content Include="@(NativeBinaries)">
        <Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
      </Content>
  </ItemGroup>
```

fredericDelaporte:
Npgsql 3.2.5 fails tests involving using the connection from prepare phase. Outside of this PR, we use 3.2.2. I have checked all intermediate versions, they all work in this case, only 3.2.5 (current latest) fails. So upgrading to 3.2.4.1 instead.

The 3.2.5 failure may lead to pending prepared transaction on PostgreSQL db, causing it to prevent nhibernate database drop, wrecking all subsequent builds whatever the PR. It is then required to use the pgadmin tool on build agent for rollbacking pending prepared transactions (see [here](https://stackoverflow.com/q/25077581/1178314)) and dropping the db.

fredericDelaporte:
Testing SqlServerCe NuGet package (I was falsely presuming it was not having an official one) and Oracle Managed driver.
If all work, this will mean all currently running teamcity builds will rely on NuGet rather than on manually deployed binaries. (Not counting the Oracle non managed build since it is suspended.)

ShowBuildMenu.bat
fredericDelaporte: Thanks,

Added to NHibernate.Test.csproj, seems taken into account by NHibernate.TestDatabaseSetup.csproj too, so it should fix last CE failure.

fredericDelaporte:
As said on another PR too, Oracle versions greater than 12.1.2400 fail [NH-1171](https://nhibernate.jira.com/browse/NH-1171) tests, it seems it has its own bug with comments in SQL. And it is a known bug since quite long as shown [here](https://stackoverflow.com/a/7542670/1178314), but apparently with some versions suffering less of it. So just using its NuGet package without upgrading it.

hazzik:
I have slightly different view on who and how shall reference these packages, but I think it is good for now.

fredericDelaporte:
I agree my approach is a bit "brute force" approach. Works for now, till no data providers have conflicting requirements.

fredericDelaporte:
Rebased.

PR #1404, opened 2017-10-25
Use MsBuild for packing .nupkg files

ngbrown:
Use `dotnet msbuild` tasks to pack NuGet `.nupkg` files.  Done as part of the primary build in the current NAnt scripts.

* The `README.md` file isn't transformed to `.html` for the `.nupkg` file, but is copied as is.  I doubt many people need it there when they should refer to the most recent version on github.
* ~Removed the unused and broken NHibernate.Setup WiX project.~


src/NHibernate/NHibernate.csproj
```diff
@@ -1,42 +1,72 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../../build-common/NHibernate.props" />
+
   <PropertyGroup>
-    <Description>An object persistence library for relational databases.</Description>
+    <Description>NHibernate is a mature, open source object-relational mapper for the .NET framework. It is actively developed, fully featured and used in thousands of successful projects.</Description>
+    <PackageTags>ORM; O/RM; DataBase; DAL; ObjectRelationalMapping; NHibernate; ADO.Net; Core</PackageTags>
```
hazzik: Tags are space separated

src/NHibernate/NHibernate.csproj
```diff
@@ -1,42 +1,72 @@
...
+
   <ItemGroup>
     <PackageReference Include="Antlr3.CodeGenerator" Version="3.5.2-beta1">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Antlr3.Runtime" Version="3.5.1" />
-    <PackageReference Include="iesi.collections" Version="4.0.2" />
-    <PackageReference Include="Remotion.Linq" Version="2.1.2" />
-    <PackageReference Include="Remotion.Linq.EagerFetching" Version="2.1.0" />
+    <PackageReference Include="Antlr3.Runtime" Version="[3.5.1, 4.0)" />
```
hazzik: I doubt there will be v4 of Antlr3

src/NHibernate/NHibernate.csproj
ngbrown: https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#packagetags

hazzik:
Can you please drop d55ffdfc0c21f0878252963339b0a9773218a231 out of this PR?

ngbrown:
> Can you please drop d55ffdf out of this PR?

Do you want it as a separate pull request?  Is there a reason for leaving it?

src/NHibernate/NHibernate.csproj
```diff
@@ -1,42 +1,72 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../../build-common/NHibernate.props" />
+
   <PropertyGroup>
-    <Description>An object persistence library for relational databases.</Description>
+    <Description>NHibernate is a mature, open source object-relational mapper for the .NET framework. It is actively developed, fully featured and used in thousands of successful projects.</Description>
```
hazzik: Please put this into `<PackageDescription>`

ngbrown: @hazzik  what are you referencing?  `<Description>` is what Visual Studio fills out based on the project package property page, and what is documented here: https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#description

hazzik:
>Is there a reason for leaving it?
There are a lot of changes not related to the topic of the PR

build-common/NHibernate.props
```diff
@@ -12,5 +12,17 @@
     <Product>NHibernate</Product>
     <Company>NHibernate.info</Company>
     <Copyright>Licensed under LGPL.</Copyright>
+    <Authors>NHibernate community, Hibernate community</Authors>
+    <Company>NHibernate.info</Company>
```
hazzik: Already here

src/NHibernate/NHibernate.csproj
hazzik: I don't care about this documentation. I'm talking about this one: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks.Pack/Pack.targets#L28

ngbrown: I agree that the naming of properties intended to be set in `.csproj` for generating `.nuget` packages are not altogether consistent, but they do work as documented.

If you set the description in `<PackageDescription>`, it may override that pack.targets property, but it no longer shows up in Visual Studio's project property page for the Package description, so `<PackageDescription>` clearly is not the property that is intended to be set in the `.csproj` file.  The line you reference is by default set by `$(Description)`, which we were already setting in the `.csproj`. I just adjusted the text so the output `.nuspec` matches the old templated one.

If you compile and compare the extracted `.nuget` packages, it all works fine.

build-common/NHibernate.props
ngbrown: Removed duplicate property.

ngbrown:
I've separated out the d55ffdf commit into #1405 and fixed the errors and pushed a replacement commit.

src/NHibernate/NHibernate.csproj
ngbrown: Excluding v4 was how the previous `.nuspec` had it.  It doesn't hurt to keep it.

src/NHibernate/NHibernate.csproj
hazzik: The problem that `Description` also meant to set `AssemblyDescriptionAttribute`. The `PackageDescription` is what transferred into the `PackTask` (L211). So, it is a correct way to set the separate descriptions for a package and an assembly.

ngbrown: ok... but where does that ever show up in normal use?  It doesn't show up in the `NHibernate.dll` file Properties dialog.

Is there a problem if the text on `AssemblyDescriptionAttribute` does match the NuGet package description?

hazzik: https://github.com/NuGet/NuGet.Client/pull/1648

ngbrown: I'm not saying that they don't collide, obviously they do.  I was asking why it matters?  Also, the NuGet/NuGet.Client#1648 pull request was only 2 months ago, so it wasn't included in the .net core 2.0 release.  On my computer, the file at:

`C:\Program Files\dotnet\sdk\2.0.0\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets` doesn't have `<PackageDescription>` and instead relies on `<Description>`:

```xml
    <Description Condition="'$(Description)'==''">Package Description</Description>
```

hazzik: It's in 2.0.2

hazzik:
I've removed unrelated changes

ngbrown:
I think my changes were better.  Now with the `<GeneratePackageOnBuild>true</GeneratePackageOnBuild>` in the project instead set at build time, the `NHibernate.5.0.0.nupkg` and `NHibernate.5.0.0.symbols.nupkg` are going to be built every time you build in Visual Studio.  That means zipping up some 2,000 source files every time.

hazzik:
But this is only in release configuration. Do you often build that?

ngbrown:
No, I suppose I don't build release in visual studio often.

Too many common items such as PackageProjectUrl, PackageIconUrl, PackageRequireLicenseAcceptance, and PackageLicenseUrl moved out of the common `NHibernate.props` and into the `.csproj`.  There's no reason they would change between packages released from this repository.

ngbrown:
After force pushing on top of each other... the current version (e7a59798e748adf95a3f2dda11ab60e67951eb11) is probably the best it's been.  Thanks for the feedback.

hazzik:
Thanks

Issue #1385, opened 2017-10-17
SecondLevelCache CreateSQLQuery().UniqueResult<int>() throws Exception Specified cast is not valid.


joesun99:
Hi

Please check if this is a bug i get `Exception: System.InvalidCastException : Specified cast is not valid`.

For test case these two flags must be enabled:
```
 cache.use_second_level_cache=true 
 cache.use_query_cache=true 
```

TestCase:

```c#
[Test]
public void QueryCacheWithScalarFunctionThatReturnsInt()
{
	using (ISession s = OpenSession())
	{
		int result =  s
			.CreateSQLQuery("SELECT [dbo].[TestQueryCache] ()")
			.SetCacheable(true)
			.UniqueResult<int>();

		Assert.IsNotNull(result);
	}
}
```

Mssql scalar function create:

```sql
USE [nhibernate]
GO

/****** Object:  UserDefinedFunction [dbo].[TestQueryCache]    Script Date: 17.10.2017 14:11:13 ******/
SET ANSI_NULLS ON
GO

SET QUOTED_IDENTIFIER ON
GO

-- =============================================
-- Author:		<Author,,Name>
-- Create date: <Create Date, ,>
-- Description:	<Description, ,>
-- =============================================
CREATE FUNCTION [dbo].[TestQueryCache] 
()
RETURNS INT
AS
BEGIN
	
	RETURN DATEPART(YEAR, GETUTCDATE()) * 100 + DATEPART(MONTH, GETUTCDATE());

END

GO
```

jaxx:
I've got exactly the same problem.

hazzik:
Probably a `Regression`

fredericDelaporte:
Still worth confirming: do you have this issue specifically with 5.0? It is written nowhere currently.

jaxx:
I have NHibernate 5.0 now, but this bug was also in earlier versions. 4.1.1 for example.

jaxx:
@hazzik Wow, that you left this one out from the 5.0.1 release. I've been waiting this issue to be resolved for 2-3 years for now. So we can't upgrade NH in our project and continue to use 3.3.5.4000.

fredericDelaporte:
5.0.1 main purpose is to fix 5.0.0 regressions. This trouble is not a recent regression, and its PR (#1421) is not yet ready. One month after the major release, it is reasonable to have that release own regressions fixed, without waiting more for including unrelated bugs in the fix.

[...]

PR #1406, opened 2017-10-25
SecondLevel QueryCache fix

joesun99:
Hi

Please check my solution for #1385. Problem is in Loader.cs method GetResultSet where is call AutoDiscoverTypes. The resultTypes are assigned to CustomerLoader object. The class variable IType[] resultTypes is not used by reference.

Regards


src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
```diff
@@ -60,5 +60,19 @@ public void QueryCacheWithNullParameters()
...
 			}
 		}
+
+		[Test]
+		public void QueryCacheWithScalarFunctionThatReturnsInt()
+		{
+			using (ISession s = OpenSession())
+			{
+				int result = s
+					.CreateSQLQuery("SELECT [dbo].[TestQueryCache] ()")
```
gliljas: The TestQueryCache method is not in the test database. Use a simple "SELECT 1" or similar.

src/NHibernate/Loader/Loader.cs
```diff
@@ -1533,17 +1533,17 @@ private IList ListIgnoreQueryCache(ISessionImplementor session, QueryParameters
...
 
 			QueryKey key = GenerateQueryKey(session, queryParameters);
 
-			IList result = GetResultFromQueryCache(session, queryParameters, querySpaces, resultTypes, queryCache, key);
+			IList result = GetResultFromQueryCache(session, queryParameters, querySpaces, ref resultTypes, queryCache, key);
 
 			if (result == null)
 			{
-				result = DoList(session, queryParameters, key.ResultTransformer);
+				result = DoList(session, queryParameters, queryParameters.ResultTransformer);
```
gliljas: Why this change?

hazzik:
I don't understand why types shall be passed by ref. I don't see any assignments.

src/NHibernate/Loader/Loader.cs
joesun99: Hi

This change is to be reverted, was inside for testing purposes.

Thx

LG

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
joesun99: Hi

I will change test case so it will be  testable.

Regards

joesun99: Hi @gliljas 

I have made new feature branch https://github.com/joesun99/nhibernate-core/tree/SecondLevelQueryCacheFixOld 
Should I make new pull request and close current?

Thx

Regards

fredericDelaporte: Or update the current PR, unless that is two different approaches and you prefer keep the old one in case it finally should be chosen.

joesun99:
New pull request https://github.com/nhibernate/nhibernate-core/pull/1406 in new feature branch

PR #1408, opened 2017-10-26
Refactored SetOptions

### LINKED ISSUE #1372: NH-3982 - Simple query with Cacheable, Fetch and SingleOrDefault throws exception (regression from 3.3.0)
**Lee Timmins** created an [issue](https://nhibernate.jira.com/browse/NH-3982) — <time datetime="2017-04-12T09:29:54.780Z" title="2017-04-12T09:29:54.780Z">12th April 2017, 9:29:54</time>:
> This is related to https://nhibernate.jira.com/browse/NH-3186. However if you say:
> 
> var query = db.Users
> **.Cacheable()**
> .Fetch(x => x.Role)
> .Where(x => x.Name == "xxx")
> .SingleOrDefault();
> 
> Then you get the same error thrown in the ticket above. I am currently upgrading from version 3.3.4 and this used to work fine.
>

gliljas:
As discussed a long time ago....

https://groups.google.com/d/topic/nhibernate-development/XqhDhDrILhc/discussion

...I think that things that don't affect the query, but only how it's executed, should be dealt with as soon as possible, and not complicate things further for everything else. By doing that we fix #1372 and issues we've been seeing with Cacheable and complex queries. 

A couple of things to consider.

1. If something here is considered to be breaking API changes, I'm sure we can internalize them.
2. I added NSubstitute. It made sense and came to good use.
3. The instantiation of the OptionsResultOperator now handles cases where the call is still not reshaped into a ConstantExpression, which could happen quite easily. In the current version this throws an error. However, this should perhaps be dealt with in the first expression simplification process, so that misuses (such as passing an argument to SetCacheable from an IQueryable) can be thrown away.
4. From what I could gather, the calls to SetOptions were included in the cache key. They shouldn't be. 

Also, I think that the change in syntax was a bit unfortunate, but I only blame myself for not being involved enough.

SetOptions(x => x.SetCacheable(true))

...is a bit "un-LINQy", where fluent, immutable words like "With" and "As" are more the convention.

the old way or...

WithOptions(x => x.Cacheable(true))

would be better IMHO. Perhaps in 6.0?

The syntax also adds a new interface (IQueryableOptions) which, at least very strictly speaking, blocks extensibility, in a way that the extension methods didn't. There's an issue #968 which requests a LockMode extension. This can't be handled without modifying the interface. OK for 5.1?


hazzik:
@gliljas you can add `WithOptions` and deprecate the `SetOptions` method, it shall be fine for 5.1

hazzik:
Unfortunately, it is not possible to rename SetCacheable to Cacheable as it collides with a property with the same name.

hazzik:
4 tests are failing.

hazzik:
`WithOptions` would allow you to change the argument to `Action<NhQueryableOptions>`

gliljas:
I will prioritize the failing tests for now.

gliljas:
I found the problem, which had to do with Translate, which is called twice, and that didn't work well with having a single, mutable QueryModel to work with. That's all fine. We should able to call it twice, although it should be avoided. I tried to just check whether the translationresults had already been generated, but got "recognition errors" then. Solved with a Lazy, which will only be invoked for cached queries. Other queries recreate the QueryModel, when needed.

gliljas:
SetCacheable vs Cacheable? What does it collide with? It's on a specific interface.

hazzik:
No, if you want to name the method `Cacheable` (without Set) it would
collide with the property.

gliljas:
Ah, now I see what you mean. Well, that should be a non issue, if it becomes internalized into NhQueryableOptions. 

Anyway, the syntax proposal was more of a side discussion. I don't want it to stop the bugfix.

fredericDelaporte:
The idea behind the interface and its method names was to be able at some later point to have it in common with other querying API. So the names match those of other API like IQuery and ICriteria.

src/NHibernate.Test/Linq/ByMethod/SetOptionsTests.cs
```diff
@@ -0,0 +1,50 @@
...
+using NHibernate.Engine;
+using NHibernate.Linq;
+using NSubstitute;
+using NUnit.Framework;
+
+namespace NHibernate.Test.Linq
+{
+	[TestFixture]
+    public class SetOptionsTests
+    {
```
fredericDelaporte: There are some whitespace glitches here and at other places.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -37,6 +37,8 @@ public Expression VisitPartialEvaluationException(PartialEvaluationExceptionExpr
 
 	internal class NhEvaluatableExpressionFilter : EvaluatableExpressionFilterBase
 	{
+		
+
```
fredericDelaporte: Undue change.

src/NHibernate/Linq/Visitors/QueryOptionsExtractor.cs
```diff
@@ -0,0 +1,80 @@
...
+			{
+				current = ((SubQueryExpression)current.MainFromClause.FromExpression).QueryModel;
+			}
+			return current.MainFromClause;
+		}
+
+
+
+
+
```
fredericDelaporte: Single blank line should be enough.

gliljas:
Yes, I know there's an issue calling for that kind of similarity.

https://github.com/nhibernate/nhibernate-core/pull/377

But for IQueryable I don't see the point. It's such a different beast, with syntax, limitations and possibilities differing a lot from (the hidden) IQuery and ICriteria.

src/NHibernate/Linq/Visitors/QueryOptionsExtractor.cs
gliljas: Will fix

fredericDelaporte:
> But for IQueryable I don't see the point. It's such a different beast, with syntax, limitations and possibilities differing a lot from (the hidden) IQuery and ICriteria.

Not agreed. The current set of Linq options has the same features than for criteria and query (cache options and timeout). Linq readonly option if added should behave the same than for criteria and query. If some API happens to have specific options, it can still be added to its specific interface.

Not having a common interface for handling options common across APIs is not a sound design in my opinion.

gliljas:
In theory, I agree with you, but in practice, I don't see any benefit other than terminology. Maybe I'm missing something?

fredericDelaporte:
If someone wants to use some "common options setter" across API, it would avoid him having to duplicate it for each API.

gliljas:
Yes, but....

1. Is that really a plausible scenario? We're talking about query specific settings here.  I'm not saying that it's impossible, but is it a common scenario? 
2. If someone wants to do that, introducing your own abstraction would be a simple task, and probably make more sense as an instruction to an abstracted data layer.   
3. Is it worth sacrificing LINQ syntax fluency for this?

@rjperes , you created the PR mentioned above. Did that come from an actual use case?

fredericDelaporte:
Note that I do not think it is an important subject. It just looks to me nicer that way.

1. For caching options, since it frequently requires already two calls per query, one for enabling cache for the query, the other for specifying its cache region.
2. That will still be different code paths depending on the querying API.
3. I do not get you on that point. The Linq fluency is still there, `SetOptions` (renamed `WithOptions`, why not) yields an `IQueryable`. And even the `SetOptions` inner argument is fluent.

Would you rather keep separated `IQueryable` extension methods for each options? I feel it bloats a bit the `IQueryable` interface.

But the main reason I have chosen to consolidate them inside a single `SetOptions` method was implementation. That was the way I have used to cease having as many Linq result operators as options to handle, allowing to add new options in a way less intrusive way. Adding an option does no more requires handling a new result operator, the change stays confined into these option handling interface and classes. I have not checked the detail of your new implementations, but I guess it provides that same benefit.

The main drawback of that interface based implementation is the "breaking change" aspect of adding new options due to changing an interface, either mandating going through extension methods if willing to add it into a minor release, or relaxing the SemVer semantic on that point by considering this option interface is not meant to be implemented outside of NHibernate. (So adding members to it would not be a breaking change for normal use cases.)

gliljas:
It's not super important for me either, but these little things are worth discussing :)

1. I agree that the API should steer the user right in requiring cache enabling when one of the other two are set.
2. Yes
3. I was speaking in more linguistic terms. Also, it wasn't a very clever comment from my side :)

Just for fun I made a branch off of this, with a different take:

https://github.com/gliljas/nhibernate-core/commit/f7ea47e4b3ccb0e2433a60fed082f716d5c49626

The objective was first and foremost to encapsulate the logic of the options in their own classes. It would have to be tweaked to allow LockMode to work, since that needs to see which "table" to lock.

While dabbling I made a syntax that uses an abstract QueryOptions class, and static methods to create them.

The syntax becomes:

`query.WithOptions(QueryCache.Enabled)`

or

`query.WithOptions(QueryCache.Disabled)` (no chaining after Disabled)

or

`query.WithOptions(QueryCache.Enabled.InRegion("someregion"))` In or With?

or 

`query.WithOptions(CommandTimeout.InSeconds(30))`

or, since it's a params array... 

`query.WithOptions(QueryCache.Enabled, CommandTimeout.InSeconds(30))`

Quite nice, but with the disadvantage that no Intellisense will guide the developer to available options. That would of course be possible with extension methods, although that requires "methods" as an entry point.

`query.WithOptions(x => x.QueryCache().Enabled)`

...which kind of destroys the neatness...

Perhaps, just perhaps, such a QueryOption could be allowed to do more this, such as inserting rewriters.

query.WithOptions(Localize.Strings)

Nah...

src/NHibernate/Linq/NhLinqExpression.cs
```diff
@@ -34,7 +36,8 @@ public class NhLinqExpression : IQueryExpression
 
 		private readonly Expression _expression;
 		private readonly IDictionary<ConstantExpression, NamedParameter> _constantToParameterMap;
-
+		private Lazy<QueryModelWithExtractedOptions> _queryModelAndOptions;
```
hazzik: I don't understand why do you need this lazy.

hazzik:
I have some idea. It is to have QueryableOptions on DefaultQueryProvider and do not propagate them to NhQueryable. I will try to make an alternate PR.

src/NHibernate/Linq/NhLinqExpression.cs
gliljas: The idea was to make sure that the extraction process was run as few times as possible. When the query plan is cached the first Lazy instance will be invoked, since Translate will not be called, but as soon as it is called, the Lazy is replaced with a direct reference. Perhaps it wasn't necessary. It would also be good to avoid the double Translate.

fredericDelaporte:
> Quite nice, but with the disadvantage that no Intellisense will guide the developer to available options. 

Yes, bad discover-ability hurts, that is what cause me to have long favored the NUnit [classic model](https://github.com/nunit/docs/wiki/Classic-Model) over the [constraint model](https://github.com/nunit/docs/wiki/Constraint-Model). The later requires to know which static classes are exposing constraints (`Is`, `Does`, `Throws`, `Has` and maybe some other I do not know). They are not easily discovered (and seem not even documented).

Maybe could this point be somewhat addressed by having those options classes being sub-classes of `QueryOptions`, yielding a syntax like:
```c#
query.WithOptions(QueryOptions.Cache.Disabled)
```

hazzik:
@gliljas this changes would not enable for lock modes (#968) as lock modes need to be attached to a "query sources"

gliljas:
@hazzik I know. That's why I mentioned that it would have to be tweaked a bit for that. The "callback" would need more than the IQuery or the QueryOptions instance could just act as a factory class for the real callback class, so that it could contain more than one callback and maintain necessary state (since, theoretically, the QueryOptions class could be reused). In the prototype implementation, that is..  It's definitely not "ready for production".

gliljas:
Ok, that was a confusing sentence. But I think you get the point :)

hazzik:
> relaxing the SemVer semantic

Not an option.

hazzik:
Closing in favor of #1412

PR #1412, opened 2017-10-31
Store Linq query options in a query provider instead of a queryable

hazzik:
I've written this while on a bus. The tests are taken from #1408. I did not run them, so they could fail.


gliljas:
This is nice and it works. I'm just not 100% confident that the options will always be set on an NhQueryable. But perhaps that could be a reasonable sacrifice.

hazzik:
> that the options will always be set on an NhQueryable.

Yeah, made them lazy.

gliljas:
I'm generally fine with this approach, at least for now. It solves the main issue of ResultOperators that end up complicating the query parsing. Perhaps there is still a case to be made for the "removed ResultOperator" approach, but that could be evaluated for a possible syntax change. I think the code may now include a couple of things (just one actually) that was only required for that approach to work. The QueryModel will not need to be evaluated before the Translate now.

hazzik:
>The QueryModel will not need to be evaluated before the Translate now.

@gliljas but it is not

gliljas:
Ok, didn't look close enough.

hazzik:
@fredericDelaporte could you please review?

hazzik:
Yeah, true. Did not think about that scenario. Thanks @fredericDelaporte

fredericDelaporte:
We are maybe missing a "no option contamination test" to prevent this to be done someday. When I had done that `SetOptions` change for 5.0, I had thought about consolidating options on provider too, but did not done it due to the provider being potentially shared. I should have added a test guarding against that.

hazzik:
I'll make some separate PRs to replace `SetOptions` with `WithOptions`, fix tests and tests for "contamination"

src/NHibernate.Test/Linq/QueryCacheableTests.cs
```diff
@@ -76,21 +76,23 @@ public void QueryIsCacheable3()
 		public void QueryIsCacheableWithRegion()
 		{
 			Sfi.Statistics.Clear();
-			Sfi.QueryCache.Clear();
+			Sfi.EvictQueries();
```
hazzik: `Sfi.QueryCache.Clear()` cleans only default cache region. Named cache regions remain untouched. So need to clear them separatly

src/NHibernate.Test/Linq/QueryCacheableTests.cs
```diff
@@ -76,21 +76,23 @@ public void QueryIsCacheable3()
 		public void QueryIsCacheableWithRegion()
 		{
 			Sfi.Statistics.Clear();
-			Sfi.QueryCache.Clear();
```
hazzik: `Sfi.QueryCache.Clear()` cleans only default cache region. Named cache regions remain untouched. So need to clear them separatly

hazzik:
@gliljas I like the proposal to just accept `Action<NhQueryableOptions>` instead of `Action<IQueryableOptions>`, eg concrete class instead of an interface.

hazzik:
@fredericDelaporte it does not more contaminate the sibling queryables.

src/NHibernate/Linq/DefaultQueryProvider.cs
```diff
@@ -81,9 +88,19 @@ public TResult Execute<TResult>(Expression expression)
...
+			return WithOptions(setOptions)
+				.CreateQuery<T>(queryable.Expression);
+		}
+
+		public IQueryProvider WithOptions(Action<NhQueryableOptions> setOptions)
+		{
+			var options = _options != null
+				? _options.Clone()
+				: new NhQueryableOptions();
+			setOptions(options);
```
fredericDelaporte: We should guard against a `null` `setOptions` here, throwing `ArgumentNullException` instead of null ref.

src/NHibernate.Test/Linq/ByMethod/WithOptionsTests.cs
```diff
@@ -71,5 +102,37 @@ public void AppliesOptionsToQuery()
 			query.Received(1).SetCacheRegion("testregion");
 			query.Received(1).SetTimeout(10);
 		}
+		
+		[Test]
+		public async Task DoNotCantaminateQueryWithOptionsAsync()
```
fredericDelaporte: Typo, `Cantaminate` instead of `Contaminate`, same in commit message.

src/NHibernate.Test/Linq/ByMethod/WithOptionsTests.cs
```diff
@@ -42,6 +42,37 @@ public void AppliesOptionsToQuery()
 			query.Received(1).SetCacheRegion("testregion");
 			query.Received(1).SetTimeout(10);
 		}
+		
+		[Test]
+		public void DoNotCantaminateQueryWithOptions()
```
fredericDelaporte: Typo, `Cantaminate` instead of `Contaminate`, same in commit message.

hazzik:
>About WithOptions alternate syntax, we will still be able to do it later as an overload if we want to, so it should not be troublesome to already use the name with current syntax.

I did not get what this means. What's your proposal?

fredericDelaporte:
This is about Gunnar comment:

> Shouldn't we hold off on changing to WithOptions before settling on a final syntax? SetOptions is fine for now and the corresponding bugs can be fixed without changing the API.

It just means I do not think introducing now `WithOptions` should be an issue for the syntax changes he proposes, if we later choose to do his additional syntax change.

hazzik:
Ok, cool

ngbrown:
This broke the release build.

fredericDelaporte:
An obsolete warning has not been handled, which the release built rejects.

hazzik:
@fredericDelaporte that is why I proposed to enable warnings for all builds and not only release.

Issue #1413, opened 2017-11-01
Make ISession.IsDirty() method safe, effective and useful


bahusoid:
ISession.IsDirty() method description:
> Does this ISession contain any changes which must be synchronized with the database? Would any SQL be executed if we flushed this session?

**So expected behavior:** method should not persist any entities (work in read-only mode) and doesn't throw exceptions for dirty session.

**And current behavior:** It's very unexpected that this call can lead to exceptions or even data modifications in DB:
1) Exception is thrown when transient ManyToOne object  is present in session.
2) And data modifications - all cascaded entities are saved (which I think is very dangerous behavior)

I will try to provide failing tests today.

My use case for using this ISession.IsDirty() is the following:
```C#
var entity = LoadEntity();
//Do some modifications to entity like entity.Children.Add(child); which is mapped with Cascade.All
entity.Children.Add(CreateNewChild());

//... And at the end of the request:
if(session.IsDirty())
{
	//do some resource consuming stuff like open DTC transaction or something else that needs to be done only if database modifications are requried
	using(var dtcTransaction = OpenDTCTransaction())
	{
		session.Flush();
		//...do some stuff with legacy VB objects
	}
}

```

With current implementation I can't use `IsDirty()` as it will save child `entity.Children` elements without transaction context - very dangerous and unexpected.

If there is a valid scenario with current implementation - could you please explain it?

I understand why  cascade updates are executed - current implementation call full PreFlush life-cycle. But is it really necessary? Can't we simply check for any queued action and for any first dirty object in session?

fredericDelaporte:
Seems related to [NH-2727](https://nhibernate.jira.com/browse/NH-2727) and [NH-2136](https://nhibernate.jira.com/browse/NH-2136) which were closed as "Not an issue" without much explanations.
About NH-2136, the case happens with identity generator. Maybe there is some good technical constraint mandating this behavior with identity, but it should then be documented on `IsDirty` (and added to the list of reasons it is not recommended to use identity for generating ids).

hazzik:
Maybe some information is in the mailing lists (I’m away sleeping, so could not check now)

hazzik:
Read the linked issues. I don’t know why this is an “expected behavior”, @fabiomaulo. Could you please elaborate? I agree that the behavior should be as described by @bahusoid.

fabiomaulo:
I can't remember which was the discussion around this matter... let me check...

fabiomaulo:
[...]

PR #1414, opened 2017-11-01
Test ISession.IsDirty() should not trigger cascade saving

bahusoid:
Added failing test cases for #1413.


src/NHibernate.Test/NHSpecificTest/GH1413/Entities.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace NHibernate.Test.NHSpecificTest.GHTODO
```
fredericDelaporte: Should be `GH1413` instead of `GHTODO`, here and in other files.

src/NHibernate.Test/NHSpecificTest/GH1413/FixtureByCode.cs
```diff
@@ -0,0 +1,114 @@
...
+			{
+				EntityParent parent = GetParent(session);
+
+				//parent.Child entity is not cascaded, I want to save it explictilty later
+				parent.Child = new EntityChild { Name = "NewManyToOneChild" };
+
+
+				bool isDirty = false;
+				Assert.DoesNotThrow(() => isDirty = session.IsDirty(), "ISession.IsDirty() call should not fail for transient  many-to-one object referenced in session.");
+				Assert.IsTrue(isDirty, "ISession.IsDirty() call should return true.");
```
fredericDelaporte: For new tests, we favor NUnit [constraint model](https://github.com/nunit/docs/wiki/Constraint-Model) over its [classic model](https://github.com/nunit/docs/wiki/Classic-Model), see [two models](https://github.com/nunit/docs/wiki/Assertions#two-models).

bahusoid:
@fredericDelaporte  Done. I'm not very comfortable with this "constraint model" - so let me know if anything is not right. 
Note: Exception handling seems better to me in classic model. At least Assert.DoesNotThrow looks more concise/clear to me than Assert.That(.., Throw.Nothing,...)

src/NHibernate.Test/NHSpecificTest/GH1413/FixtureByCode.cs
```diff
@@ -0,0 +1,114 @@
...
+			using (ISession session = OpenSession())
+			using (session.BeginTransaction())
+			{
+				EntityParent parent = GetParent(session);
+
+				//parent.Child entity is not cascaded, I want to save it explictilty later
+				parent.Child = new EntityChild { Name = "NewManyToOneChild" };
+
+				bool isDirty = false;
+				Assert.That(() => isDirty = session.IsDirty(), Throws.Nothing, () => "ISession.IsDirty() call should not fail for transient  many-to-one object referenced in session.");
```
fredericDelaporte: I was not knowing it could take a delegate for the error message (third argument). A string argument is enough for such cases indeed. (That is just a small detail, not important.)

bahusoid:
Added fix for SessionIsDirtyShouldNotFailForNewManyToOneObject  test case:
https://github.com/nhibernate/nhibernate-core/pull/1414/files#diff-1c250a6682b3027be99e132c9257cb4dR72

src/NHibernate/Type/ManyToOneType.cs
```diff
@@ -224,5 +204,31 @@ public override bool[] ToColumnNullness(object value, IMapping mapping)
...
+				return true;
+			}
+
+			object oldid = GetIdentifier(old, session);
+			object newid = GetIdentifier(current, session);
+			IType identifierType = GetIdentifierType(session);
+
+			return checkable == null
+				? identifierType.IsDirty(oldid, newid, session)
+				: identifierType.IsDirty(old, current, checkable, session);
```
fredericDelaporte: Shouldn't it be `identifierType.IsDirty(oldid, newid, checkable, session);`?

bahusoid: Ouch.. Indeed it should..

bahusoid:
PR updated (I cleaned up all not related stuff)

bahusoid:
@hazzik  Sorry I won't be able to do any coding in the following week or two.
And yes - current test doesn't generate any SQL inserts. But it surely will If I change EnityChild.Id mapping from Generators.GuidComb to Generators.Identity.

The main point of the test - IsDirty() should not modify nor entities in memory nor DB state. It should be immutable operation. Don't you agree?

As @fredericDelaporte  already pointed out [here](https://github.com/nhibernate/nhibernate-core/issues/1413#issuecomment-341415777):
> is it legit for IsDirty to raise SaveOrUpdate events?

So this test checks this part - and I think there is no need to do additional checks if SQL is really generated. Let me know if you disagree - I will update tests as you think is right.

fredericDelaporte:
I have added statistics instead, with native generator. See commit details.

hazzik:
@fredericDelaporte  this way this issue becomes a duplicate of NH-2136. Which is closed "By design". The problem is that if we want to fix this issue - we need completely rewrite IsDirty.

fredericDelaporte:
The point is, was it really reasonable to close this as "by design" without more actions? I do not think so. The bare minimum would be to document the possible side effects of `IsDirty`, at least in its xml comments.

Currently it just states:
> Does this `ISession` contain any changes which must be synchronized with the database? Would any SQL be executed if we flushed this session?

So a method explicitly documented as allowing to check if some SQL may be triggered by another operation may itself triggers some SQL without giving any clue about that... Quite unexpected. And it causes `IsDirty` to be somewhat moot for those falling in the cases triggering SQL. (`identity` generator => insert + all generator db-sided => id generation triggered.)

fredericDelaporte:
I have checked the Hibernate implementation, it is still the same than NHibernate on this point.
About reported issues on Hibernate side, none have been rejected as "by design", they are closed by lack of test case or information instead. It seems they have never been seriously studied:

- [HHH-404](https://hibernate.atlassian.net/browse/HHH-404) - issue with a "please help" quality level, it was not really deserving more than a close.
- [HHH-3963](https://hibernate.atlassian.net/browse/HHH-3963) - it was providing a test case, but it was not handled until a new major version of Hibernate, then closed because the opener did not confirm the issue was still there with the new version. Hum, interesting way of handling issues: applying it to NHibernate bugs would likely close a good bunch of them!
- [HHH-6611](https://hibernate.atlassian.net/browse/HHH-6611) - closed because not providing a test case.
- [Forum post](https://forum.hibernate.org/viewtopic.php?f=1&t=949705&sid=fd7bdcd7c4ab3dc8bbd152055bc7b940) with an invalid suggestion. At least in NHibernate case, changing the flush mode does not dodge the issue. There are other forum posts on the subject, without any useful answer.

> we need completely rewrite IsDirty

If it was about rewriting all the `IsDirty` methods, it would be a bit overwhelming. But it looks to me we "only" need to rewrite the `DefaultDirtyCheckEventListener` class, which seems to be only used by `SessionImpl.IsDirty`. So it does not look to me as a too big deal, although it would likely imply to somewhat rewrite a good deal of `AbstractFlushingEventListener` into `DefaultDirtyCheckEventListener` but in a way without side effects. It could have two drawbacks: de-synchronization risk between the `IsDirty` checking logic and the actual flush changes ; loss of the optimization reducing the number of dirty check to be done after an `IsDirty` call (the `ClearFromFlushNeededCheck` call currently done in `DefaultDirtyCheckEventListener`). But the `IsDirty` could also gain some speed through early exit as soon as it finds something dirty.

fredericDelaporte:
Having looked a bit more into it, it would require a lot of changes in the cascading logic for allowing a "cascade simulation" just telling whether a save would be done or not. So I think we should likely just document the potential side effects of `ISession.IsDirty`.

bahusoid:
If I try to go this  way:
>  we need completely rewrite IsDirty
Will you consider it? Current implementation seems like not very useful as identity generators are commonly used (at least in projects where I was involved). And processing all the objects in session - doesn't look as effective solution.

fredericDelaporte:
The dirty checking mechanism is injectable at many levels, including for the global dirty check on the session. You can change the default isDirty event handler for the session by setting another one, implementing `IDirtyCheckEventListener`, and set by using `Configuration.SetListener`. Its default implementation is [`DefaultDirtyCheckEventListener`](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Event/Default/DefaultDirtyCheckEventListener.cs).
A custom one not triggering the save should likely not inherit from `AbstractFlushingEventListener` but directly implements `IDirtyCheckEventListener` on its own.
Now for reducing code duplication it would likely instead require to adapt `AbstractFlushingEventListener` and the cascade mechanism, but doing so without breaking changes is challenging (otherwise it would mandate a new major release).
The advantage of the custom dirty check event listener would be to be doable as a separated project, usable with current NHibernate v5.

hazzik:
This can "easily" be fixed by enabling this branch of code:

https://github.com/nhibernate/nhibernate-core/blob/0b667429315ec7979a5e3290280ea20c210df678/src/NHibernate/Event/Default/AbstractSaveEventListener.cs#L256-L266

Unfortunately due to "NH Different behavior (shouldDelayIdentityInserts=false anyway)" it never executes.

Hope this helps.

bahusoid:
I think that proper dirty check shouldn't call `PerformReplicateOrSave` at all. But it looks like a huge refactoring I'm not ready to start right now. And if #1754  fixes the main issue (unexpected db inserts) maybe it's not worth implementing the "proper" way.

hazzik:
> And if #1754 fixes the main issue (unexpected db inserts)  

It does not. We need the other means to enable delayed inserts for the dirty checks.

hazzik:
Rebased, squashed, commit reworded, test is marked as known bug.

PR #1415, opened 2017-11-01
Correct MaxAliasLength for various dialects

### LINKED ISSUE #879: NH-4006 - Provide a correct MaxAliasLength for various dialects
**Alexander Zaytsev** created an [issue](https://nhibernate.jira.com/browse/NH-4006) — <time datetime="2017-05-02T21:45:50.155Z" title="2017-05-02T21:45:50.155Z">2nd May 2017, 21:45:50</time>:
> After implementation of [NH-3899](https://nhibernate.jira.com/browse/NH-3899) the correct MaxAliasLength was not specified.
>

fredericDelaporte:
Fix #879.

### DIFF
--- a/src/NHibernate/Dialect/Dialect.cs
+++ b/src/NHibernate/Dialect/Dialect.cs
@@ -2282,10 +2282,11 @@ namespace NHibernate.Dialect
 			get { return "lower"; }
 		}
 
-		public virtual int MaxAliasLength
-		{
-			get { return 10; }
-		}
+		// 18 is the smallest of all dialects we handle.
+		/// <summary>
+		/// The maximum length a SQL alias can have.
+		/// </summary>
+		public virtual int MaxAliasLength => 18;
 
 		/// <summary>
 		/// The syntax used to add a column to a table. Note this is deprecated
--- a/src/NHibernate/Dialect/FirebirdDialect.cs
+++ b/src/NHibernate/Dialect/FirebirdDialect.cs
@@ -520,6 +520,12 @@ namespace NHibernate.Dialect
 			RegisterFunction("tanh", new StandardSQLFunction("tanh", NHibernateUtil.Double));
 		}
 
+		// As of Firebird 2.5 documentation, limit is 30/31 (not all source are concordant), with some
+		// cases supporting more but considered as bugs and no more tolerated in v3.
+		// It seems it may be extended to 63 for Firebird v4.
+		/// <inheritdoc />
+		public override int MaxAliasLength => 30;
+
 		#region Informational metadata
 
 		/// <summary>
--- a/src/NHibernate/Dialect/IngresDialect.cs
+++ b/src/NHibernate/Dialect/IngresDialect.cs
@@ -53,6 +53,15 @@ namespace NHibernate.Dialect
 			DefaultProperties[Environment.ConnectionDriver] = "NHibernate.Driver.IngresDriver";
 		}
 
+		// Ingres 10.2 supports 256 bytes (so worst unicode case would mean 64 characters), but I am unable to find
+		// the limit for older versions, excepted many various sites mention a 32 length limit.
+		// https://unifaceinfo.com/docs/0906/Uniface_Library_HTML/ulibrary/INS_NAMING_RULES_8EEFC1A489331BF969D2A8AA36AF2832.html
+		// There are traces of a ticket for increasing this in version 10: http://lists.ingres.com/pipermail/bugs/2010-May/000052.html
+		// This dialect seems to target version below 9, since there is Ingres9Dialect deriving from it.
+		// So sticking to 32.
+		/// <inheritdoc />
+		public override int MaxAliasLength => 32;
+
 		#region Overridden informational metadata
 
 		public override bool SupportsEmptyInList => false;
--- a/src/NHibernate/Dialect/Oracle12cDialect.cs
+++ b/src/NHibernate/Dialect/Oracle12cDialect.cs
@@ -3,7 +3,7 @@ using NHibernate.SqlCommand;
 namespace NHibernate.Dialect
 {
 	/// <summary> 
-	/// A dialect specifically for use with Oracle 10g.
+	/// A dialect specifically for use with Oracle 12c.
 	/// </summary>
 	/// <remarks>
 	/// The main difference between this dialect and <see cref="Oracle12cDialect"/>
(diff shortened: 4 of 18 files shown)


src/NHibernate/Dialect/SQLiteDialect.cs
```diff
@@ -395,6 +395,10 @@ public override bool SupportsForeignKeyConstraintInAlterTable
 		/// </remarks>
 		public override bool SupportsConcurrentWritingConnections => false;
 
+		// Said to be unlimited. http://sqlite.1065341.n5.nabble.com/Max-limits-on-the-following-td37859.html
+		/// <inheritdoc />
+		public override int MaxAliasLength => 100;
```
hazzik: I think it needs to be aligned with others. Say 128.

src/NHibernate/Dialect/IngresDialect.cs
```diff
@@ -53,6 +53,11 @@ public IngresDialect()
 			DefaultProperties[Environment.ConnectionDriver] = "NHibernate.Driver.IngresDriver";
 		}
 
+		// Ingres 10.2 supports 256 bytes (so worst unicode case would mean 64 characters), but I am unable to find
+		// the limit for older versions, excepted many various sites mention a 32 length limit. Being conservative.
+		/// <inheritdoc />
+		public override int MaxAliasLength => 32;
```
hazzik: I think we need not specify if we’re not sure.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2282,10 +2282,10 @@ public virtual string LowercaseFunction
...
 		}
 
-		public virtual int MaxAliasLength
-		{
-			get { return 10; }
-		}
+		/// <summary>
+		/// The maximal length a SQL alias can have.
+		/// </summary>
+		public virtual int MaxAliasLength => 10;
```
hazzik: Take the minimum here - 18

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2282,10 +2282,10 @@ public virtual string LowercaseFunction
 			get { return "lower"; }
 		}
 
-		public virtual int MaxAliasLength
-		{
-			get { return 10; }
-		}
+		/// <summary>
+		/// The maximal length a SQL alias can have.
```
hazzik: It seems it shall be "The maximum".

src/NHibernate/Dialect/Oracle12cDialect.cs
```diff
@@ -38,5 +38,9 @@ public override SqlString GetLimitString(SqlString querySqlString, SqlString off
 
 			return result.ToSqlString();
 		}
+
+		// 128 since 12.2. https://stackoverflow.com/a/756569/1178314
+		/// <inheritdoc />
+		public override int MaxAliasLength => 128;
```
fredericDelaporte: OK, our own CI is 12.1, and fails due to that, should have stayed on the safe side here.

src/NHibernate/Dialect/FirebirdDialect.cs
```diff
@@ -520,6 +520,12 @@ private void RegisterTrigonometricFunctions()
 			RegisterFunction("tanh", new StandardSQLFunction("tanh", NHibernateUtil.Double));
 		}
 
+		// As of Firebird 2.5 documentation, limit is 30/31 (not all source are concordant), with some
+		// cases supporting more but considered as bugs and no more tolerated in v3.
+		// It seems it may be extended to 63 for Firebird v4.
+		/// <inheritdoc />
+		public override int MaxAliasLength => 30;
```
fredericDelaporte: It seems we still have a bug. A test has generated the alias `requestcompletionstatus3_8691_0_` for a query, that is 32 characters long.

hazzik: It's because the identifier has 2 "unique" parts for some reasons.

hazzik: It's in "Column.cs", line 173.

hazzik: Or not. I'm confused.

fredericDelaporte: I am going to have a look. I am currently adjusting lengths and sources for those lengths.

hazzik: It is from `SelectFragment.cs` suffix field. Which is a "collectionSuffix" which is a table number + `_`.

fredericDelaporte: Run one by one, the tests succeed. It seems there is some coupling between tests.

fredericDelaporte: The coupling is very likely coming from the thread static `tableCounter` in `Table` class. It seems it endlessly grows as we build/discard session factories.
Either this field should be reset when starting to build a session factory (if it has no more usages after build) or it should be bound to the session factory instead of the thread on which it was built.

The `SelectFragment` suffix is an additional suffix which is meant (at least in the cases I have checked) to differentiate columns of the same table in case it is auto-joined. It does not detect auto-join but simply always add the suffix according to query joins count.

It then means the `Column.GetAlias` should account for that potential additional suffix by not exhausting the full maximum length. At least 3 characters should be left in my opinion, for supporting more than 9 joins (up to 99) and the additional `_` of the suffix.

If we want want to support more than 99 joins (up to 999), we should keep 4 characters left, but I am not used to have such monstrous queries with NHibernate.

fredericDelaporte:
Added fix for #1418. It could be put as a separated PR, but the trouble it fixes is revealed only in this current PR, and is related. I may still drop that commit and put it in another PR if you prefer.

fredericDelaporte:
#879 commits squashed, I have pushed the Oracle build up in its queue and intend to merge after it rather than right now, even if there is no reason for the squash to introduce a failure.

hazzik:
Are you going to squash or merge commits?

fredericDelaporte:
Merge, since it is two issues. What I have squashed was the three commits concerning 879.

hazzik:
Then you need to reverse order of the commits.

src/NHibernate/Mapping/Column.cs
```diff
@@ -115,10 +115,14 @@ public string GetQuotedName(Dialect.Dialect d)
 			return IsQuoted ? d.QuoteForColumnName(_name) : _name;
 		}
 
+		// Accomodate the one character suffix appended in AbstractCollectionPersister and
```
hazzik: Typo should be "Accommodate" (double "m")

hazzik:
I've changed commits order, reworded messages and fixed a typo.

fredericDelaporte:
Sorry, done the same right now excepted reword of commit messages...

hazzik:
@fredericDelaporte force push war is fun 😂

fredericDelaporte:
Just reforce your push^^.

hazzik:
Ok, done.

PR #1420, opened 2017-11-02
Fix #1419 - ISession.IsDirty() shouldn't throw exception for transient many-to-one object in a session

### LINKED ISSUE #1419: ISession.IsDirty() shouldn't throw exception for transient many-to-one object in a session
As suggested in https://github.com/nhibernate/nhibernate-core/issues/1413#issuecomment-341455364
I extracted separate issues.

**Current behavior:** TransientObjectException is thrown by ISession.IsDirty() call  when transient many-to-one object is present in session
**Expected behavior:** ISession.IsDirty() should not throw exception

Technical details of the problem is [here](https://github.com/nhibernate/nhibernate-core/issues/1413#issuecomment-341442122).

bahusoid:
Fixes #1419


src/NHibernate/Type/ManyToOneType.cs
```diff
@@ -224,5 +204,31 @@ public override bool[] ToColumnNullness(object value, IMapping mapping)
...
+			{
+				return false;
+			}
+
+			if (old == null || current == null)
+			{
+				return true;
+			}
+
+			if (ForeignKeys.IsTransientFast(GetAssociatedEntityName(), current, session) == true)
```
fredericDelaporte: Need to use `IsTransientSlow`, otherwise the bug will still be there for entities having an assigned id already assigned or having an `unsaved-value` set as `undefined`. `IsTransientSlow` hits the db only if `IsTransientFast` was not able of answering, which occurs in aforementioned cases.

fredericDelaporte: Need to use `IsTransientSlow`, otherwise the bug will still be there for entities having an assigned id already assigned or having an `unsaved-value` set as `undefined`. `IsTransientSlow` hits the db only if `IsTransientFast` was not able of answering, which occurs in aforementioned cases.

hazzik: Need to have a test for this case.

hazzik: @fredericDelaporte, the `GetEntityIdentifierIfNotUnsaved` calls `IsTransientFast` as well.

hazzik: Ok, what will happen in case of an entity with assigned ID - the "IsDirty" method will not fail, but it will return `false`.

bahusoid: @fredericDelaporte 
>Need to use IsTransientSlow, otherwise the bug will still be there
I simply reused check that lead to exception:
https://github.com/nhibernate/nhibernate-core/blob/a6f97d0423b3a57a74e736e9447ca1d0cbb91e88/src/NHibernate/Engine/ForeignKeys.cs#L250-L265

So exception won't be thrown by GetIdentifier method.

@hazzik 
> the "IsDirty" method will not fail, but it will return false

Why false? What IsDirty returns depends on `identifierType.IsDirty` call. And from my understanding it should work fine - we will simply check if identifier value is changed. And if it's new object - it should definitely change.

bahusoid: Just in case I made a quick test with assigned id and it works OK. Should I squash it in to test commit?

bahusoid:
Added new test as separate commit. Let me know if it needs to be rebased and squashed instead.

src/NHibernate/Type/ManyToOneType.cs
fredericDelaporte: Yes, better add it to test commit. You may first add it as a "fixup! Test case for GH1419 (ISession.IsDirty() failed for transient many-to-one" commit and later squash it if you prefer.

Well, already added, never minde, can still be squashed later.

> Ok, what will happen in case of an entity with assigned ID - the "IsDirty" method will not fail, but it will return false.

I believe it yields `null` instead in that case, but @bahusoid test seems to show I were wrong in assuming the whole case would fail. Adding tests for assigned id and unsaved-value "undefined" is definitely the way to go.

hazzik: I was checking with Assigned Guids. Not sure why it works with integers.

bahusoid: @hazzik I changed int to Guid in my test and it still works...

hazzik: Maybe I had the same problem ;)

hazzik:
Thanks. Nice work

PR #1421, opened 2017-11-02
SecondLevel QueryCache fix

### LINKED ISSUE #1385: SecondLevelCache CreateSQLQuery().UniqueResult<int>() throws Exception Specified cast is not valid.

joesun99:
Fixes #1385.


src/NHibernate.Test/App.config
```diff
@@ -7,7 +7,7 @@
 	</configSections>
 
 	<connectionStrings>
-		<add name="TestConnectionString" connectionString="Server=localhost\sqlexpress;Database=nhibernate;Integrated Security=SSPI" />
+    <add name="TestConnectionString" connectionString="Data Source=192.168.1.70\MSSQLSERVER2;Initial Catalog=nhibernate;Integrated Security=True" />
```
gliljas: Undo this change

src/NHibernate/Async/Loader/Custom/CustomLoader.cs
```diff
@@ -52,6 +54,12 @@ public Task<IList> ListAsync(ISessionImplementor session, QueryParameters queryP
 			return rowProcessor.BuildResultRowAsync(row, rs, resultTransformer != null, session, cancellationToken);
 		}
 
+		protected override async Task PutResultInQueryCacheAsync(ISessionImplementor session, QueryParameters queryParameters, IType[] resultTypes,
+									   IQueryCache queryCache, QueryKey key, IList result, CancellationToken cancellationToken)
+		{
+			await base.PutResultInQueryCacheAsync(session, queryParameters, resultTypes, queryCache, key, result, cancellationToken);
```
gliljas: The wrong resultTypes are used.

src/NHibernate/Loader/Criteria/CriteriaLoader.cs
```diff
@@ -93,7 +93,7 @@ protected override bool[] IncludeInResultRow
 
 		public IList List(ISessionImplementor session)
 		{
-			return List(session, translator.GetQueryParameters(), querySpaces, resultTypes);
+			return List(session, translator.GetQueryParameters(), querySpaces, ResultTypes);
```
gliljas: Undo this change

src/NHibernate/Loader/Custom/CustomLoader.cs
```diff
@@ -552,5 +560,15 @@ public IType GetHibernateType(int columnPos)
 				return TypeFactory.Basic(resultSet.GetFieldType(columnPos).Name);
 			}
 		}
+
```
gliljas: Remove

src/NHibernate/Loader/Loader.cs
```diff
@@ -1517,7 +1517,7 @@ public void LoadCollectionBatch(ISessionImplementor session, object[] ids, IType
 		/// <param name="querySpaces"></param>
 		/// <param name="resultTypes"></param>
 		/// <returns></returns>
-		protected IList List(ISessionImplementor session, QueryParameters queryParameters, ISet<string> querySpaces, IType[] resultTypes)
+		protected virtual IList List(ISessionImplementor session, QueryParameters queryParameters, ISet<string> querySpaces, IType[] resultTypes)
```
gliljas: This change isn't necessary

src/NHibernate/Loader/Loader.cs
```diff
@@ -1533,7 +1533,7 @@ private IList ListIgnoreQueryCache(ISessionImplementor session, QueryParameters
 			return GetResultList(DoList(session, queryParameters), queryParameters.ResultTransformer);
 		}
 
-		private IList ListUsingQueryCache(ISessionImplementor session, QueryParameters queryParameters, ISet<string> querySpaces, IType[] resultTypes)
+		protected virtual IList ListUsingQueryCache(ISessionImplementor session, QueryParameters queryParameters, ISet<string> querySpaces, IType[] resultTypes)
```
gliljas: This change isn't necessary

src/NHibernate/Loader/Loader.cs
```diff
@@ -1543,7 +1543,7 @@ private IList ListUsingQueryCache(ISessionImplementor session, QueryParameters q
 
 			if (result == null)
 			{
-				result = DoList(session, queryParameters, key.ResultTransformer);
+				result = DoList(session, queryParameters, queryParameters.ResultTransformer);
```
gliljas: This change isn't necessary

src/NHibernate/Loader/Loader.cs
```diff
@@ -1858,4 +1858,14 @@ protected bool TryGetLimitString(Dialect.Dialect dialect, SqlString queryString,
 
 		#endregion
 	}
+
```
gliljas: Remove

src/NHibernate/Loader/Loader.cs
joesun99: I am still testing this change

src/NHibernate/Async/Loader/Custom/CustomLoader.cs
joesun99: I thought I should use in parameter from PutResultInQueryCacheAsync method? Can u provide me with detail from where to take it. Thx

gliljas: You should do exactly what you've done in the non-async version. Ignore the parameter and send in this.resultTypes instead.

src/NHibernate/Loader/Loader.cs
gliljas: I don't think it's necessary

src/NHibernate/Loader/Custom/CustomLoader.cs
```diff
@@ -368,6 +369,13 @@ public IEnumerable<string> NamedParameters
 			get { return parametersSpecifications.OfType<NamedParameterSpecification>().Select(np=> np.Name ); }
 		}
 
+		protected override void PutResultInQueryCache(ISessionImplementor session, QueryParameters queryParameters, IType[] resultTypes,
+										   IQueryCache queryCache, QueryKey key, IList result)
+		{
+			resultTypes = this.resultTypes;
+			base.PutResultInQueryCache(session, queryParameters, resultTypes, queryCache, key, result);
```
gliljas: You might as well just send this.resultTypes without intermediate assignment.

src/NHibernate/Loader/Custom/CustomLoader.cs
```diff
@@ -553,4 +561,4 @@ public IType GetHibernateType(int columnPos)
 			}
 		}
 	}
-}
```
gliljas: Avoid, if possible

src/NHibernate/Async/Loader/Custom/CustomLoader.cs
```diff
@@ -131,4 +139,4 @@ public Task<object> ExtractAsync(object[] data, DbDataReader resultSet, ISession
 			}
 		}
 	}
-}
```
gliljas: Avoid, if possible

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
```diff
@@ -60,5 +60,19 @@ public void QueryCacheWithNullParameters()
...
+				int result = s
+					.CreateSQLQuery("SELECT 200012")
+					.SetCacheable(true)
+					.UniqueResult<int>();
+
+				Assert.IsNotNull(result);
+			}
+		}
 	}
-}
```
gliljas: Avoid, if possible

src/NHibernate/Type/TypeHelper.cs
```diff
@@ -342,4 +342,4 @@ private static bool Dirty(StandardProperty[] properties, object[] currentState,
 			}
 		}
 	}
-}
```
gliljas: Avoid, if possible

src/NHibernate/Async/Loader/Custom/CustomLoader.cs
fredericDelaporte: In fact Async files should not be manually modified, just run the async generator (see contributing guidelines).

If there is a trouble with what it generates, then discuss about how to handle that.

gliljas:
This issue seems to be a bit more complicated than just fixing the resulttypes issue.

joesun99:
Hi @gliljas 

Why do you think so "This issue seems to be a bit more complicated than just fixing the resulttypes issue."  ?

Regards

src/NHibernate/Type/TypeHelper.cs
joesun99: Is this problem with my IDE and line endings or maybe file encoding.

fredericDelaporte: The .editorconfig mandates a blank end line, and enforces this on all changed files. So when reverting a file, do not just manually revert it, but reset it.

gliljas:
@joesun99, just from testing the changes, it seems there are still issues with the resulttransformers. So your testing of that was indeed a good idea.

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
hazzik: It's .editorconfig setting, so it's fine.

fredericDelaporte:
Hum, there is something wrong in having the commits on master appearing here in this PR.

It seems your master branch is correctly up-to-date (rebased). 

So it should just be a matter of rebasing your PR branch on your master.

joesun99:
Hi @fredericDelaporte 

I wanted to bring latest changes to my feature branch. I am missing something.
Your branch is up-to-date with 'origin/SecondLevelQueryCacheFixOld'.

Regards

fredericDelaporte:
Your master now appears with: `This branch is 6 commits ahead of nhibernate:master.`.
You cannot rebase on this. You need a master even with its upstream. (Yesterday when I have checked it was looking to me even, but that is not the case today. Maybe I have badly checked yesterday but I do not see how I would.)

Recover your master with:
```
>git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'yourRemoteName/master'.

>git reset --hard head~6
HEAD is now at 1e00feff5 Allow naming a property FieldInterceptor when an entity has lazy properties (#1444)
>git push -f
```
Check on https://github.com/joesun99/nhibernate-core that it is now even with nhibernate:master.

Then checkout the PR branch and rebase it:
```
>git checkout SecondLevelQueryCacheFixOld
Switched to branch 'SecondLevelQueryCacheFixOld'
Your branch is up-to-date with 'yourRemoteName/SecondLevelQueryCacheFixOld'.

>git rebase master
First, rewinding head to replay your work on top of it...
Applying: SecondLevel QueryCache fix
Applying: Merge branch 'master' into SecondLevelQueryCacheFixOld
Applying: Merge branch 'SecondLevelQueryCacheFixOld' of https://github.com/joesun99/nhibernate-core into SecondLevelQueryCacheFixOld
Applying: Add test for GH1436
Applying: Fix a stack overflow when merging a transient entity with lazy properties
Applying: Upgrade AsyncGenerator to 0.6.2 and regenerate. (#1445)
Applying: Document IsDirty side effects.
Applying: SecondLevel QueryCache fix
.git/rebase-apply/patch:98: trailing whitespace.

warning: 1 line adds whitespace errors.
error: Failed to merge in the changes.
Using index info to reconstruct a base tree...
M       src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
M       src/NHibernate/Loader/Custom/CustomLoader.cs
M       src/NHibernate/Loader/Hql/QueryLoader.cs
M       src/NHibernate/Loader/Loader.cs
M       src/NHibernate/Type/TypeHelper.cs
Falling back to patching base and 3-way merge...
Auto-merging src/NHibernate/Loader/Loader.cs
CONFLICT (content): Merge conflict in src/NHibernate/Loader/Loader.cs
Auto-merging src/NHibernate/Loader/Hql/QueryLoader.cs
Auto-merging src/NHibernate/Loader/Custom/CustomLoader.cs
Auto-merging src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
CONFLICT (content): Merge conflict in src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
Patch failed at 0008 SecondLevel QueryCache fix
The copy of the patch that failed is found in: .git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".
```

Resolves the conflicts, finish the rebasing.

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
```diff
@@ -60,5 +60,19 @@ public void QueryCacheWithNullParameters()
...
 			}
 		}
+
+		[Test]
+		public void QueryCacheWithScalarFunctionThatReturnsInt()
+		{
+			using (ISession s = OpenSession())
+			{
+				int result = s
+					.CreateSQLQuery("SELECT 200012")
```
fredericDelaporte: This query appears to be not compatible with all tested databases.

I think a fix would be:

 - Add an `override OnSetUp()` and move into it the following code from the other test:
```c#
{
	s.Save(simple, 1L);
	s.Flush();
}
```
- Add an `override OnTearDown()` and move into it the following code from the other test:
```c#
{
	s.Delete("from Simple");
	s.Flush();
}
```
- Change the SQL query to
```SQL
select 200012 from Simple where id_ = 1
```

I am unsure all databases will support the raw selected scalar, maybe you would need to change it to
```SQL
select 200011 + id_ as raw_scalar from Simple where id_ = 1
```
But that will turn it into a `long`, which would then require to adapt the test and check the bug is also demonstrated with that.

Otherwise you may restrict your test to run only on some databases, with something like:
```c#
if (!(Dialect is MsSql2000Dialect))
	Assert.Ignore(
		"Test use a raw SQL query that may not be compatible with currently tested database.");
```
But we favor adjusting tests for them to be run-able on all databases.

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
```diff
@@ -60,5 +60,19 @@ public void QueryCacheWithNullParameters()
...
+					.SetCacheable(true)
+					.UniqueResult<int>();
+
+				Assert.IsNotNull(result);
```
fredericDelaporte: This test is not failing even without your fixes. It needs to run two times the query for validating the fix.

fredericDelaporte: In fact it fails without your changes, but with your changes trying then to re-query (so trying to retrieve the result from the cache) then fails.

fredericDelaporte:
As explained in #1385, the approach here does not look right to me. I am not even sure we should actually fix the trouble, but maybe just yield a better error message.
Anyway, this PR is for now replaced by #1489.

PR #1426, opened 2017-11-05
Upgrade AsyncGenerator to 0.6.0

hazzik:


src/NHibernate.Test/Async/SystemTransactions/ResourceManagerFixture.cs
```diff
@@ -0,0 +1,186 @@
...
+				eventTransaction = e.Transaction;
+			}
+		}
+
+		#endregion
+		#endregion
+		#region Tests setup/teardown/utils
+
+		[TearDown]
+		public async Task TearDownAsync(CancellationToken cancellationToken = default(CancellationToken))
```
hazzik: @maca88 methods wit `TearDownAttribute` are specified as being copied. Why did it generate the Async?

maca88: This is related to [this issue](https://github.com/maca88/AsyncGenerator/issues/68). Copying a method for a partial class would cause a compilation error as the method would be duplicated. Probably, what you want here is to mark all classes with `ExplicitAttribute` to be generated as a new type.

hazzik: Ok, the solution was to apply `[TestFixture]` attribute. Thanks for the tip.

src/NHibernate.Test/Async/SystemTransactions/DistributedSystemTransactionFixture.cs
```diff
@@ -157,6 +157,46 @@ protected override void OnTearDown()
 		}
 
 		[Theory]
+		public void CanRollbackTransaction(bool explicitFlush)
```
fredericDelaporte: A bunch of synchronous tests are now duplicated in the async class.

src/NHibernate.Test/Async/SystemTransactions/DistributedSystemTransactionFixture.cs
```diff
@@ -207,6 +247,38 @@ protected override void OnTearDown()
 		}
 
 		[Theory]
+		[Description("Another action inside the transaction do the rollBack outside nh-session-scope.")]
+		public void RollbackOutsideNh(bool explicitFlush)
```
fredericDelaporte: Duplicated synchronous test.

src/NHibernate.Test/Async/SystemTransactions/DistributedSystemTransactionFixture.cs
```diff
@@ -259,6 +331,37 @@ protected override void OnTearDown()
 		}
 
 		[Theory]
+		[Description("rollback inside nh-session-scope should not commit save and the transaction should be aborted.")]
+		public void TransactionInsertWithRollBackTask(bool explicitFlush)
```
fredericDelaporte: Duplicated synchronous test.

src/NHibernate.Test/Async/SystemTransactions/DistributedSystemTransactionFixture.cs
```diff
@@ -357,6 +460,60 @@ protected override void OnTearDown()
 		}
 
 		[Theory]
+		[Description(@"Two session in two txscope
+ (without an explicit NH transaction)
+ and with a rollback in the second dtc and a ForceRollback outside nh-session-scope.")]
+		public void TransactionInsertLoadWithRollBackTask(bool explicitFlush)
```
fredericDelaporte: Duplicated synchronous test.

This are the one called by `MultiThreadedTransaction` test, but there should not be copied here in async, since MultiThreadedTransaction is not there either. It looks like a regression of the async generator.

maca88: It's a bug, thanks for discovering it. By ignoring the `MultiThreadRunner` class, the calculation for the method conversion fails to calculate the the correct conversion in this case. [I've made](https://github.com/maca88/AsyncGenerator/issues/69) an issue, with a simple test case demonstrating the bug. The old version would also not work in this case. In order to fix it I will have to revisit the method conversion  logic and find a better way to calculate the final conversion. I will post here once it is fixed.

maca88: With the version `0.6.1`, the mentioned synchronous tests are not generated anymore.

hazzik: Thanks @maca88

PR #1428, opened 2017-11-06
Insert underscore in combined parameter name

gliljas:
Simplest possible fix for #1357, which bit us in production today. 

Ideally, this should be extracted to a testable class.

### DIFF
--- a/src/NHibernate.Test/QueryTest/MultiCriteriaFixture.cs
+++ b/src/NHibernate.Test/QueryTest/MultiCriteriaFixture.cs
@@ -487,5 +487,38 @@ namespace NHibernate.Test.QueryTest
 				Assert.That(results[1], Is.InstanceOf<List<int>>());
 			}
 		}
+
+		[Test]
+		public void UsingManyParametersAndQueries_DoesNotCauseParameterNameCollisions()
+		{
+			//GH-1357
+			using (var s = OpenSession())
+			{
+				var item = new Item {Id = 15};
+				s.Save(item);
+				s.Flush();
+			}
+
+			using (var s = OpenSession())
+			{
+				var multi = s.CreateMultiCriteria();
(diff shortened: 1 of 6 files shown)


hazzik:
@gliljas I've changed this to point to the older issue. Also, could you please add tests for "future" queries?

gliljas:
Do you mean multiqueries (and criteria)?

hazzik:
Ah, sorry, it's fine. I wanted for Linq/Criteria Futures. But you've already added them.

hazzik:
Still need for MultiCriteria

src/NHibernate.Test/QueryTest/MultiCriteriaFixture.cs
```diff
@@ -487,5 +487,38 @@ public void CanGetResultInAGenericList()
...
+				{
+					var criteria = s.CreateCriteria(typeof(Item));
+					for (var j = 0; j < 12; j++)
+					{
+						criteria = criteria.Add(Restrictions.Gt("id", j));
+					}
+					multi.Add(criteria);
+				}
+				//Paramater combining is only used for cacheable queries
+				multi.SetCacheable(true);
```
fredericDelaporte: Small typo (Paramater instead of Parameter) and yet another somewhat undue implementation discrepancy between criteria and query. I think the criteria implementation is the right one, the query one creates those parameters which are only used in the query cache key uselessly when the query is not cacheable.

So if we realign/mutualize implementation later, the multi-query test may no more test anything. It would be better to enable caching on it too (by enabling caching on all its queries, with same region).

gliljas: Good idea. I'll fix the typo, set cacheable and rebase.

PR #1431, opened 2017-11-09
Release 5.0.1

fredericDelaporte:
I have not really found a satisfactory solution for the release.

I have manually drafted the [release](https://github.com/nhibernate/nhibernate-core/releases/tag/untagged-c7dece9a78efdef0dae0), it is probably view-able by all owners from the [release list](https://github.com/nhibernate/nhibernate-core/releases).

I have temporarily created another draft with [GitReleaseManager](https://github.com/GitTools/GitReleaseManager) for extracting the issue list. It has not the option to just create the list as a file, nor does it allows an easy text export, and it includes PR within issues.

I have taken the generated markdown and trimmed its code to amend the txt release notes.

That is a lot of manual manipulations.

Furthermore, in case an issue or PR is in the milestone without one of the configured label, the tool fails with an exception. Its message is understandable, but that is not a great error handling.

I have also tried [GitReleaseNotes](https://github.com/GitTools/GitReleaseNotes), but it has too few options. At least GitReleaseManager allows to handle our labels for sorting things into categories.

### DIFF
--- a/GitReleaseManager.yaml
+++ b/GitReleaseManager.yaml
@@ -1,0 +1,23 @@
+create:
+  include-footer: false
+issue-labels-include:
+ - "t: Bug"
+ - "t: New Feature"
+ - "t: Improvement"
+ - "t: Task"
+issue-labels-exclude:
+ - "r: Rejected"
+ - "r: Replaced"
+issue-labels-alias:
+ - name:   "t: Bug"
+   header: Bug
+   plural: Bug
+ - name:   "t: New Feature"
+   header: New Feature
+   plural: New Feature
+ - name:   "t: Improvement"
+   header: Improvement
+   plural: Improvement
+ - name:   "t: Task"
+   header: Task
+   plural: Task
--- a/releasenotes.txt
+++ b/releasenotes.txt
@@ -41,6 +41,7 @@ Release notes - NHibernate - Version 5.0.1
     * #711 Switch doc generation to UTF-8.
 
 ** Task
+    * #1431 Release 5.0.1
     * #1405 Remove unused and broken NHibernate.Setup WiX project
 
 
--- a/releasenotes.txt
+++ b/releasenotes.txt
@@ -4,21 +4,44 @@
 Release notes - NHibernate - Version 5.0.1
 
 ** Bug
+    * #1428 Insert underscore in combined parameter name
+    * #1424 Bad wording and example fixes in cache documentation.
+    * #1420 Fix #1419 - ISession.IsDirty() shouldn't throw exception for transient many-to-one object in a session
     * #1419 ISession.IsDirty() shouldn't throw exception for transient many-to-one object in a session
     * #1418 Column.GetAlias should account for other suffixes
+    * #1415 Correct MaxAliasLength for various dialects
+    * #1393 Fix Linq Future aggregates failures, fixes #1387
+    * #1389 Add support for out/ref Nullable parameters of proxied methods
     * #1387 Linq Sum() with ToFutureValue fails
+    * #1384 Fix a column spec causing missing col in pdf, fix a text overflow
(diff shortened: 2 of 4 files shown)


GitReleaseManager.yaml
```diff
@@ -0,0 +1,23 @@
+create:
```
fredericDelaporte: Used to generate the release issues list as a drafted release.

releasenotes.txt
```diff
@@ -1,4 +1,27 @@
-﻿Build 5.0.0
+﻿Build 5.0.1
+=============================
+
+Release notes - NHibernate - Version 5.0.1
+
+** Bug
+    * #1419 ISession.IsDirty() shouldn't throw exception for transient many-to-one object in a session
```
fredericDelaporte: Manually reformated this way from something like
```
__Bug__

[__#1419__](https://url) ...
```
In which PR were listed too, which I removed, identifying them thanks to their url.

GitReleaseManager.yaml
```diff
@@ -0,0 +1,23 @@
...
+ - "t: New Feature"
+ - "t: Improvement"
+ - "t: Task"
+issue-labels-exclude:
+ - "r: Rejected"
+ - "r: Replaced"
+issue-labels-alias:
+ - name:   "t: Bug"
+   header: Bug
+   plural: Bug
```
fredericDelaporte: Our release notes were not pluralizing the section, so  I have put a singular for plural.

gliljas:
Good idea!

I thinks it's quite OK for now, but clearly none of these tools are good enough. I'm working on a similar tool, which might be a better solution. Well, okay.. it doesn't deal with GitHub, but with Jira,  and it's not exactly feature packed, but I think it may actually be better, since it just pulls down the data and uses a template or a C# lambda (it's used from a Cake file) to generate the actual output. I'm in no rush to finalize that tool, so I might just as well fork it and try to create a Github version instead.

I guess GitReleaseManager would still be good for pushing the releasenotes to GitHub.

fredericDelaporte:
Yes, I have manually removed the PR. I have just somewhat stick to what was done previously with Jira, listing only Jira issues, not PR. That is true my previous comments were not very clear about that, they were just implying it.

Why not taking the PR too, but it creates some redundancies for those having a matching issue.

We may avoid this redundancy by un-setting the milestone on PR having corresponding issues, or by adding to them some sort of "exclude" label we would add to GitReleaseManager.yaml too.

fredericDelaporte:
I have re-created the generated draft, you can see it [here](https://github.com/nhibernate/nhibernate-core/releases/tag/untagged-d7d0e3db911d6f14f0f8).

The command line for this is:
```
GitReleaseManager.exe create -u username -p password -o nhibernate -r nhibernate-core -m 5.0.1
```
Of course username and password have to be changed.

I have adjusted the manual [draft](https://github.com/nhibernate/nhibernate-core/releases/tag/untagged-88ef32419a99d5fb96e7) accordingly.

fredericDelaporte:
I was about to squash that to 5.0.1 branch, but it would have resulted to having this last PR missing in the release notes. Now added.
Moreover I am normally able to publish to NuGet and GitHub, but NHibernate is also still published on SourceForge. I do not have the authorizations for publishing there. (I just discovered I had already an account on it though: `fdelaporte`.)

hazzik:
I’ve added you to SF. Please Run “Package” build on TC and publish from its artifacts.

hazzik:
@fredericDelaporte some thoughts about the number of the issues fixed.

We need to distinguish the PRs w/ and w/o corresponding issues. If PR is for an existing issue, we would mark it with "t: Fix" and not include it in the release log.

fredericDelaporte:
Agreed.

PR #1434, opened 2017-11-15
Replace remaining SetOptions with WithOptions

hazzik:


src/NHibernate/NHibernate.csproj
```diff
@@ -9,14 +9,11 @@
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
     <AssemblyOriginatorKeyFile>..\NHibernate.snk</AssemblyOriginatorKeyFile>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+    <TreatSpecificWarningsAsErrors />
```
fredericDelaporte: Since the trouble was in the test project, changing it only in the main project will not avoid it to happen again in tests.

hazzik: Hahaha. True:)

PR #1437, opened 2017-11-16
Fix StackOverflowException in DefaultDynamicLazyFieldInterceptor

### LINKED ISSUE #1436: StackOverflowException  in Merge  Entity With Lazy Property
When you call `Merge`, you may get a `StackOverflowException`. Test added. May be corrected by a change in the file `PocoEntityTuplizer`, in the method `HasUninitializedLazyProperties`:
```c#
if (EntityMetamodel.HasLazyProperties 
    // add this
    && !(entity is IProxy))
```
But I do not know if it is correct or not.

offan:
Fixes #1436

### DIFF
--- a/src/NHibernate.Test/LazyProperty/LazyPropertyFixture.cs
+++ b/src/NHibernate.Test/LazyProperty/LazyPropertyFixture.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections;
+using System.Linq;
 using NHibernate.Intercept;
 using NHibernate.Tuple.Entity;
 using NUnit.Framework;
@@ -57,7 +58,7 @@ namespace NHibernate.Test.LazyProperty
 			using (var s = OpenSession())
 			using (var tx = s.BeginTransaction())
 			{
-				Assert.That(s.CreateSQLQuery("delete from Book").ExecuteUpdate(), Is.EqualTo(1));
+				s.CreateQuery("delete from Book").ExecuteUpdate();
 				tx.Commit();
 			}
 		}
@@ -132,14 +133,25 @@ namespace NHibernate.Test.LazyProperty
 		public void CanLoadAndSaveObjectInDifferentSessions()
 		{
 			Book book;
+			int bookCount;
 			using (ISession s = OpenSession())
 			{
+				bookCount = s.Query<Book>().Count();
 				book = s.Get<Book>(1);
 			}
+
+			book.Name += " updated";
 
 			using (ISession s = OpenSession())
 			{
 				s.Merge(book);
+				s.Flush();
+			}
+
+			using (ISession s = OpenSession())
+			{
+				Assert.That(s.Query<Book>().Count(), Is.EqualTo(bookCount));
+				Assert.That(s.Get<Book>(1).Name, Is.EqualTo(book.Name));
 			}
 		}
 
@@ -168,8 +180,14 @@ namespace NHibernate.Test.LazyProperty
 		}
 
 		[Test]
-		public void CanMergeWithLazyProperty()
+		public void CanMergeTransientWithLazyProperty()
 		{
+			using (ISession s = OpenSession())
+			{
+				var book = s.Get<Book>(2);
+				Assert.That(book, Is.Null);
+			}
+
 			using (ISession s = OpenSession())
 			using (var tx = s.BeginTransaction())
 			{
@@ -177,15 +195,19 @@ namespace NHibernate.Test.LazyProperty
 				{
 					Name = "some name two",
 					Id = 2,
-					ALotOfText = "a lot of text ..."
+					ALotOfText = "a lot of text two..."
 				};
+				// This should insert a new entity.
 				s.Merge(book);
 				tx.Commit();
 			}
+
 			using (ISession s = OpenSession())
 			{
 				var book = s.Get<Book>(2);
 				Assert.That(book, Is.Not.Null);
+				Assert.That(book.Name, Is.EqualTo("some name two"));
+				Assert.That(book.ALotOfText, Is.EqualTo("a lot of text two..."));
 			}
 		}
 	}
--- a/src/NHibernate.Test/LazyProperty/LazyPropertyFixture.cs
+++ b/src/NHibernate.Test/LazyProperty/LazyPropertyFixture.cs
@@ -166,5 +166,27 @@ namespace NHibernate.Test.LazyProperty
 				Assert.That(book.FieldInterceptor, Is.EqualTo("Why not that name?updated"));
 			}
 		}
+
+		[Test]
+		public void CanMergeWithLazyProperty()
+		{
+			using (ISession s = OpenSession())
+			using (var tx = s.BeginTransaction())
+			{
+				var book = new Book
+				{
+					Name = "some name two",
+					Id = 2,
+					ALotOfText = "a lot of text ..."
+				};
+				s.Merge(book);
+				tx.Commit();
+			}
+			using (ISession s = OpenSession())
+			{
+				var book = s.Get<Book>(2);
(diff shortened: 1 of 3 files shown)


offan:
fixed the test

src/NHibernate.Test/LazyProperty/LazyPropertyFixture.cs
```diff
@@ -50,7 +50,7 @@ protected override void OnTearDown()
 			using (var s = OpenSession())
 			using (var tx = s.BeginTransaction())
 			{
-				Assert.That(s.CreateSQLQuery("delete from Book").ExecuteUpdate(), Is.EqualTo(1));
+				Assert.That(s.CreateSQLQuery("delete from Book").ExecuteUpdate(), Is.EqualTo(2));
```
fredericDelaporte: Since I have commented on the commit (no PR at that time), the PR was not showing the line I was commenting. This was this one.

I consider it should be changed to: 
```c#
s.CreateQuery("delete from Book").ExecuteUpdate();
```

There are no reason to assert the number of inserted things in the teardown. So better remove this undue old assert rather than fixing it. (By the way, it is better to use a HQL delete rather than a SQL one, so I have changed that too in the above code.)

src/NHibernate.Test/LazyProperty/LazyPropertyFixture.cs
```diff
@@ -154,5 +154,22 @@ public void CanUpdateNonLazyWithoutLoadingLazyProperty()
 				Assert.That(book.Name, Is.EqualTo("some nameupdated"));
 			}
 		}
+
+		[Test]
+       		public void CanMergeWithLazyProperty()
+		{
```
fredericDelaporte: Formatting glitch.

offan:
fixed

src/NHibernate.Test/LazyProperty/LazyPropertyFixture.cs
```diff
@@ -50,7 +51,7 @@ protected override void OnTearDown()
 			using (var s = OpenSession())
 			using (var tx = s.BeginTransaction())
 			{
-				Assert.That(s.CreateSQLQuery("delete from Book").ExecuteUpdate(), Is.EqualTo(1));
+				s.CreateQuery("delete from Book").ExecuteUpdate();
```
fredericDelaporte: This assert in teardown is indeed "completing" the `CanLoadAndSaveObjectInDifferentSessions`. Removed and `CanLoadAndSaveObjectInDifferentSessions` test changed for correctly asserting itself all that it needs to assert.

src/NHibernate.Test/LazyProperty/LazyPropertyFixture.cs
```diff
@@ -156,8 +167,14 @@ public void CanUpdateNonLazyWithoutLoadingLazyProperty()
 		}
 
 		[Test]
-		public void CanMergeWithLazyProperty()
+		public void CanMergeTransientWithLazyProperty()
 		{
+			using (ISession s = OpenSession())
+			{
+				var book = s.Get<Book>(2);
+				Assert.That(book, Is.Null);
```
fredericDelaporte: The test fails when merging a not yet persisted entity. Since we use assigned ids, we must ensure that any future changes will not cause this test to start merging an already persisted entity.

src/NHibernate/Event/Default/DefaultMergeEventListener.cs
```diff
@@ -230,6 +230,9 @@ private object MergeTransientEntity(object entity, string entityName, object req
 			else
 			{
 				copy = source.Instantiate(persister, id);
+				// #1436: we use the persister to instanciate a new, transient entity. In case it has lazy properties,
+				// we need to to tell its proxy that the properties are all fetched, since it is a transient entity.
+				persister.AfterInitialize(copy, false, source);
```
fredericDelaporte: Minimal fix. I wonder whether it should be put here or inside `SessionImpl.Instantiate` called just above. The `Instantiate` method semantic does not seem to imply it may only be used on transient entities, that is why I have decided to put the fix here.

The stack overflow was occurring because the proxy `FieldInterceptor` was not initialized. In such case, trying to get it causes a sort of "endless self initialization" loop. Maybe I should try to fix that instead. I will give it a try later.

src/NHibernate/Intercept/DefaultDynamicLazyFieldInterceptor.cs
```diff
@@ -49,6 +49,11 @@ public object Intercept(InvocationInfo info)
 					FieldInterceptor = (IFieldInterceptor)info.Arguments[0];
 					return null;
 				}
+
+				if ("get_FieldInterceptor".Equals(methodName))
+				{
+					return null;
```
fredericDelaporte: When the field interceptor is not set (we are here in the `else` part of an `if (FieldInterceptor != null)`) and is asked for, we should just yield `null` instead of trying to call it on the underlying `Target` (which was causing the stack overflow).

The `"get_FieldInterceptor".Equals(methodName)` test is debatable and causes issues if the entity itself has a `FieldInterceptor` property. But this case is already bugged due to other similar tests done in `DefaultDynamicLazyFieldInterceptor`, so adding it does not cause any regression. It will need a separated fix.

src/NHibernate/Event/Default/DefaultMergeEventListener.cs
```diff
@@ -230,9 +230,6 @@ private object MergeTransientEntity(object entity, string entityName, object req
 			else
 			{
 				copy = source.Instantiate(persister, id);
-				// #1436: we use the persister to instanciate a new, transient entity. In case it has lazy properties,
-				// we need to to tell its proxy that the properties are all fetched, since it is a transient entity.
-				persister.AfterInitialize(copy, false, source);
```
fredericDelaporte: Fix no more needed with the alternate fix. Not having the proxy initialized is then equivalent to having it considering all properties are already fetched.

Ascertained by adding more asserts in the transient merge test.

src/NHibernate/Intercept/DefaultDynamicLazyFieldInterceptor.cs
hazzik: I think checks for `set_FieldInterceptor`/`get_FieldInterceptor` shall be at the beginning of the method, as in anyway the code inside is a duplicate.

fredericDelaporte:
@hazzik, if the refactoring is ok for you, I am going to rebase and squash all that to two commits, by author.

src/NHibernate/Intercept/DefaultDynamicLazyFieldInterceptor.cs
```diff
@@ -13,15 +13,25 @@ public class DefaultDynamicLazyFieldInterceptor : IFieldInterceptorAccessor, Pro
 		public object Intercept(InvocationInfo info)
 		{
 			var methodName = info.TargetMethod.Name;
-			if (FieldInterceptor != null)
+			var isPropertyGet = ReflectHelper.IsPropertyGet(info.TargetMethod);
+			var isPropertySet = !isPropertyGet && ReflectHelper.IsPropertySet(info.TargetMethod);
```
hazzik: `!isPropertyGet` && part isn't needed

hazzik:
@fredericDelaporte I did a further refactoring and merged master. If you want to squash commits, I don't mind if my name would be lost in the process.

fredericDelaporte:
The point on names is mainly to keep commit associated with who has directly worked on the change. For the fix part, being you or me, no trouble, but I do not think it should be @offan since he may not have analyzed and followed the rational behind the fix. So if someone was later asking him explanations on the fix if he was set as the author, it would not be practical.

fredericDelaporte:
Rebased and squashed.

Issue #1439, opened 2017-11-17
ComposedId Entity with Lazy Property issues update to set lazy property to null


MarkPerryBV:
When you query out a ComposedId entity (one with a composite primary key) and that entity also features a property marked as Lazy then NH will try and update that lazy property to "null" at the end of the UOW.

Entity
```
public class CompositeEntity
	{
		public virtual int Id { get; set; }
		public virtual string Name { get; set; }
		public virtual string LazyProperty { get; set; }

		public override int GetHashCode()
		{
			return (Id + "|" + Name).GetHashCode();
		}
		public override bool Equals(object obj)
		{
			if (obj == null)
				return false;
			return obj.GetHashCode() == GetHashCode();
		}
	}
```

Class Mapping:
```
public class CompositeEntityMap : ClassMapping<CompositeEntity>{
    public CompositeEntityMap(){
        ComposedId(c =>{
             c.Property(t => t.Id);
             c.Property(t => t.Name);
         });
        Property(x => x.LazyProperty, x => x.Lazy(true));
    }
}
```

Test
```
using (ISession session = OpenSession())
using (var tran = session.BeginTransaction()){
  var result = (from e in session.Query<CompositeEntity>()
			 where e.Name == "Bob"
			 select e).ToList();
	session.Flush();
	tran.Commit();
}
```

I don't know how to Assert that an update has taken place but I have done enough debugging to know that the entity comparison between current and previous state the lazy property is:
previous: IUnintitalisedLazyproperty
current: null

and then NH issues an update to the DB and makes the lazy property null. This only happens on Composite/ComposedId Entities though.

Any questions please ask. PR with test case on it to follow.

PR #1440, opened 2017-11-17
Test case for ComposedId Entity with Lazy Property is not proxified

MarkPerryBV:
#1439


src/NHibernate.Test/NHSpecificTest/GH1439/Entity.cs
```diff
@@ -0,0 +1,22 @@
...
+		{
+			return (Id + "|" + Name).GetHashCode();
+		}
+
+		public override bool Equals(object obj)
+		{
+			if (obj == null)
+				return false;
+
+			return obj.GetHashCode() == GetHashCode();
```
fredericDelaporte: Even if it is only a test, please use a correct implementation. Hashcode are not guaranteed unique per entity. Hashcode collisions occur and create bugs. Testing hashcode equality can only serves as a shortcut for early exit if they are unequal. If they are equal, the entities may still be different. So for such a simple case, better just test directly that both `Id` and `Name` are equals.

src/NHibernate.Test/NHSpecificTest/GH1439/FixtureByCode.cs
```diff
@@ -0,0 +1,78 @@
+﻿using NHibernate.Test.NHSpecificTest.GH0000;
```
fredericDelaporte: Please do not keep a dependency on the "template" bug. (It seems removable without any other changes here by the way.)

src/NHibernate.Test/NHSpecificTest/GH1439/FixtureByCode.cs
```diff
@@ -0,0 +1,78 @@
...
+			{
+				session.Delete("from System.Object");
+
+				session.Flush();
+				transaction.Commit();
+			}
+		}
+
+		[Test]
+		public void YourTestName()
```
fredericDelaporte: This should be renamed to something more meaningful. Like `LazyPropertyShouldNotBeNullified`.

src/NHibernate.Test/NHSpecificTest/GH1439/FixtureByCode.cs
```diff
@@ -0,0 +1,78 @@
+﻿using NHibernate.Test.NHSpecificTest.GH0000;
+
+namespace NHibernate.Test.NHSpecificTest.GH1439
```
fredericDelaporte: There are a bunch of using lacking here. It does not compile.

src/NHibernate.Test/NHSpecificTest/GH1439/FixtureByCode.cs
```diff
@@ -0,0 +1,78 @@
...
+			using (var tran = session.BeginTransaction())
+			{
+				var result = (from e in session.Query<CompositeEntity>()
+							 where e.Name == "Bob"
+							 select e).ToList();
+				session.Flush();
+				tran.Commit();
+				Assert.AreEqual(1, result.ToList().Count);
+			}
+		}
```
fredericDelaporte: For asserting the bug, you may:
- Enable statistics on the session factory for checking the update count. (See by example [here](https://github.com/nhibernate/nhibernate-core/pull/1414/files#diff-2fd7f1e84082ebfc50ebd804dd007385R93), it does it for an insert case.)
- Reload the entity at the end of the test, and assert the value of the lazy property.

src/NHibernate.Test/NHSpecificTest/GH1439/FixtureByCode.cs
```diff
@@ -0,0 +1,78 @@
...
+		{
+			using (ISession session = OpenSession())
+			using (var tran = session.BeginTransaction())
+			{
+				var result = (from e in session.Query<CompositeEntity>()
+							 where e.Name == "Bob"
+							 select e).ToList();
+				session.Flush();
+				tran.Commit();
+				Assert.AreEqual(1, result.ToList().Count);
```
fredericDelaporte: Redundant `ToList`.

hazzik:
Need to regenerate async tests.

fredericDelaporte:
I have added a test with the id mapped as a component to demonstrate the issue is specific to using a composite-id without a component.

fredericDelaporte:
The failing case, using a composite-id without a component, is now marked as known bug.

PR #1443, opened 2017-11-18
Preserves the stack trace when unwrapping TargetInvocationException

### LINKED ISSUE #1441: Normalize TargetInvocationException unwrapping
Some `TargetInvocationException` unwrapping are done by using `ReflectHelper.UnwrapTargetInvocationException`, which preserves the stack trace by the way.

Others are "manually" done, overriding the stack trace (mainly in Linq provider).

Some unwrapping are missing too. By example in `DefaultDynamicLazyFieldInterceptor` in the same method, one call to `InvokeMethodOnTarget` is guarded for unwrapping while the other is not. (The unwrapping should be moved to `InvokeMethodOnTarget` in this case.)

fredericDelaporte:
* Also follow up to [NH-3500](https://nhibernate.jira.com/browse/NH-3500) which was only partially fixed.
 * Fixes #1441

### DIFF
--- a/src/NHibernate/Intercept/DefaultDynamicLazyFieldInterceptor.cs
+++ b/src/NHibernate/Intercept/DefaultDynamicLazyFieldInterceptor.cs
@@ -51,16 +51,7 @@ namespace NHibernate.Intercept
 				}
 			}
 
-			object returnValue;
-			try
-			{
-				returnValue = info.InvokeMethodOnTarget();
-			}
-			catch (TargetInvocationException ex)
-			{
-				throw ReflectHelper.UnwrapTargetInvocationException(ex);
-			}
-			return returnValue;
+			return info.InvokeMethodOnTarget();
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


src/NHibernate/Intercept/DefaultDynamicLazyFieldInterceptor.cs
```diff
@@ -51,16 +51,7 @@ public object Intercept(InvocationInfo info)
...
-			try
-			{
-				returnValue = info.InvokeMethodOnTarget();
-			}
-			catch (TargetInvocationException ex)
-			{
-				throw ReflectHelper.UnwrapTargetInvocationException(ex);
-			}
-			return returnValue;
+			return info.InvokeMethodOnTarget();
```
fredericDelaporte: Unwrapping moved into `InvokeMethodOnTarget`, which is called two times in NHibernate code: here and a couple of lines above here.

PR #1445, opened 2017-11-18
Upgrade AsyncGenerator to 0.6.2 and regenerate.

fredericDelaporte:
Fixes the invalid generation of async methods for test GH1391, which is currently an annoyance to revert every time async is regenerated.


hazzik:
Can this go after #1437?

fredericDelaporte:
Yes.

PR #1447, opened 2017-11-21
Reduce number of calls to ISessionImplementor.PersistenceContext

hazzik:


src/NHibernate/Proxy/AbstractLazyInitializer.cs
```diff
@@ -273,11 +274,9 @@ private object GetProxyOrNull()
 			return null;
 		}
 		
-		private void SetReadOnly(bool readOnly)
+		private void SetReadOnly(IPersistenceContext persistenceContext, bool readOnly, IEntityPersister entityPersister)
```
fredericDelaporte: Ideally, we should not have to add arguments to a method only for sparing some property calls. That is why I was thinking about removing the check in the property, for turning it back to an innocuous call. (As should be any property call.)

Now it seems a lot more safer to do it this way, as removing the check need a lot of analysis for ensuring the check will in all cases have already be done some point sooner. (And it could imply adding that check to many places where it was previously done through the property call.)

src/NHibernate/Proxy/AbstractLazyInitializer.cs
```diff
@@ -287,9 +286,12 @@ private void SetReadOnly(bool readOnly)
 			if (initialized)
 			{
 				EntityKey key = GenerateEntityKeyOrNull(_id, _session, _entityName);
-				if (key != null && _session.PersistenceContext.ContainsEntity(key))
+				if (key != null)
 				{
-					_session.PersistenceContext.SetReadOnly(_target, readOnly);
+					if (persistenceContext.ContainsEntity(key))
```
fredericDelaporte: Seems a bit undue `if` split in two `if`. (Maybe a first change was to introduce an intermediate variable, but which was later removed due to the method signature change, without recombining the `if`.)

fredericDelaporte:
I am having a look at the stateless session. It does call `CheckAndUpdateSessionStatus` in redundant ways, like calling it just before calling a property or method which call it too. Should we adjust this kind of things in this PR too?

hazzik:
Yes. I'll take a look.

hazzik:
Unfortunately this is not a viable solution as it does not provide the sought benefits. It has halved the number of calls to `PersistenceContex`, but as @fredericDelaporte mentioned earlier this path is responsible of 4% impact. Unfortunately the execution times remained roughly the same.

PR #1452, opened 2017-11-23
Async After-/BeforeTransactionCompletion

gliljas:
While working on #1392 I recognized that the AfterTransactionComplete will always be executed synchronously, even if the rest is async. The only solution I could think of was to do what I've done in this PR. The typed delegates were replaced with interfaces, as it is in Hibernate. The reason for using delegates was clearly to allow for inline definitions of the "process" (cryptically described in the code comments). I made two "wrapper" classes for this, which takes an Action and a Func<Task> to handle such inlining.

I'm not quite sure how to configure AsyncGenerator to do the right things.

Yes, this is a breaking change :)

Update by Frédéric: it is no more breaking thanks to Alexander changes.


gliljas:
Nested partials could be an option too, i.e. create small nested classes, which can access the containing type's members.

src/NHibernate/Action/BulkOperationCleanupAction.cs
```diff
@@ -111,23 +111,30 @@ public void Execute()
...
-				return new AfterTransactionCompletionProcessDelegate((success) =>
-				{
-					this.EvictEntityRegions();
-					this.EvictCollectionRegions();
-				});
+				return new AfterTransactionCompletionProcess(
+					(success) =>
+					{
+						EvictEntityRegions();
+						EvictCollectionRegions();
```
fredericDelaporte: If the sync delegate is defined as a method, the Async Generator should then be able to generate its async counter-part, which could then be referenced as the async delegate here.

So from dev perspective, this is a three steps operation: define the sync method and use it as the sync delegate (along with a dummy async one or by using the sync only constructor), generate async, use the async counterpart as the async delegate.

Same for other places where there is those delegates to supply.

src/NHibernate/Action/CollectionAction.cs
```diff
@@ -104,27 +104,34 @@ public virtual void BeforeExecutions()
...
 				// Only make sense to add the delegate if there is a cache.
 				if (persister.HasCache)
 				{
-					return new AfterTransactionCompletionProcessDelegate((success) =>
-					{
-						CacheKey ck = new CacheKey(key, persister.KeyType, persister.Role, Session.Factory);
-						persister.Cache.Release(ck, softLock);
-					});
+					return new AfterTransactionCompletionProcess(
+						(success) =>
```
fredericDelaporte: Same as previous comment.

src/NHibernate/Action/CollectionUpdateAction.cs
```diff
@@ -113,49 +113,75 @@ private void PostUpdate()
...
 				if (Persister.HasCache)
 				{
 					// NH Different behavior: to support unlocking collections from the cache.(r3260)
-					return new AfterTransactionCompletionProcessDelegate((success) =>
-					{
-						CacheKey ck = Session.GenerateCacheKey(Key, Persister.KeyType, Persister.Role);
-
-						if (success)
+					return new AfterTransactionCompletionProcess(
+						(success) =>
```
fredericDelaporte: Same as previous comment.

src/NHibernate/Action/BulkOperationCleanupAction.cs
hazzik: It's in a backlog https://github.com/maca88/AsyncGenerator/issues/70

maca88: > (along with a dummy async one or by using the sync only constructor)

Another way is to directly use the constructor that has both sync and async actions and use the async method even if it is not generated yet, this way you can save one code generation.

I've made [an example](https://github.com/maca88/nhibernate-core/commit/f17d3fc5b9b1fdb2968b8f01e3252b956d219f5e) for this PR, how the code could look like prior the generation of the async code. 

@hazzik For this PR, the mentioned generator feature wouldn't help, as here we need both sync and async methods to be passed in the constructor method, similar to what we have for [Future](https://github.com/nhibernate/nhibernate-core/blob/f3ccc55904b9d0ddbff6abaeb6e404bf377d1727/src/NHibernate/Impl/CriteriaImpl.cs#L409).

gliljas:
@hazzik @fredericDelaporte 

What are our thoughts on what constitutes a breaking change? The IExecutable interface is public, but... 

1. Apart from BulkOperationCleanupAction all implementations are sealed
2. Actions are explicitly typed in all ActionQueue methods. 
3. Implementing IExecutable is basically useless

fredericDelaporte:
It is still a binary breaking change, so normally not acceptable in a minor release when following SemVer. Is it really inconvenient to follow the pattern seen in some other "would be interface change" cases, where it is has been dodge with a combination of extension methods and cast to known concrete implementation?

src/NHibernate/Action/IAsyncExecutable.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace NHibernate.Action
+{
+	public interface IAsyncExecutable : IExecutable
```
fredericDelaporte: I would add a `// 6.0 TODO: merge into IExecutable`. Or is there any value in keeping a separated interface for async methods?

src/NHibernate/Action/IExecutable.cs
```diff
@@ -1,5 +1,28 @@
...
+	/// <summary>
+	/// Delegate representing some process that needs to occur after transaction completion.
+	/// </summary>
+	/// <param name="success"> Did the transaction complete successfully? True means it did.</param>
+	/// <remarks>
+	/// NH specific: C# does not support dynamic interface proxies so a delegate is used in
+	/// place of the Hibernate interface (see Action/AfterTransactionCompletionProcess). The
+	/// delegate omits the <see cref="ISessionImplementor" /> parameter as it is not used.
+	/// </remarks>
+	public delegate void AfterTransactionCompletionProcessDelegate(bool success);
```
fredericDelaporte: Shouldn't these delegates be obsoleted?

src/NHibernate/Action/IExecutable.cs
```diff
@@ -21,11 +44,11 @@ public partial interface IExecutable
...
 		/// Get the before-transaction-completion process, if any, for this action.
 		/// </summary>
-		IBeforeTransactionCompletionProcess BeforeTransactionCompletionProcess { get; }
+		BeforeTransactionCompletionProcessDelegate BeforeTransactionCompletionProcess { get; }
 
 		/// <summary>
 		/// Get the after-transaction-completion process, if any, for this action.
 		/// </summary>
-		IAfterTransactionCompletionProcess AfterTransactionCompletionProcess { get; }
+		AfterTransactionCompletionProcessDelegate AfterTransactionCompletionProcess { get; }
```
fredericDelaporte: I would obsolete these properties.

src/NHibernate/Engine/ActionQueue.cs
```diff
@@ -132,16 +132,26 @@ public Task AddActionAsync(BulkOperationCleanupAction cleanupAction, Cancellatio
...
-		public void RegisterProcess(IBeforeTransactionCompletionProcess process)
+
+		public void RegisterProcess(BeforeTransactionCompletionProcessDelegate process)
 		{
-			beforeTransactionProcesses.Register(process);
+			RegisterProcess(new BeforeTransactionCompletionDelegatedProcess(process));
 		}
-	
+
+		public void RegisterProcess(AfterTransactionCompletionProcessDelegate process)
```
fredericDelaporte: I would obsolete these methods.

src/NHibernate/Engine/ActionQueue.cs
```diff
@@ -195,8 +205,16 @@ private void InnerExecute(IExecutable executable)
 
 		private void RegisterCleanupActions(IExecutable executable)
 		{
-			beforeTransactionProcesses.Register(executable.BeforeTransactionCompletionProcess);
-			afterTransactionProcesses.Register(executable.AfterTransactionCompletionProcess);
+			if (executable is IAsyncExecutable asyncExecutable)
+			{
+				beforeTransactionProcesses.Register(asyncExecutable.BeforeTransactionCompletionProcess);
+				afterTransactionProcesses.Register(asyncExecutable.AfterTransactionCompletionProcess);
```
fredericDelaporte: It could be:

```c#
RegisterProcess(asyncExecutable.BeforeTransactionCompletionProcess);
RegisterProcess(asyncExecutable.AfterTransactionCompletionProcess);
```

Slightly less code written, and the compiler will likely inline the call anyway.

src/NHibernate/Engine/ActionQueue.cs
```diff
@@ -537,6 +555,36 @@ public void AfterTransactionCompletion(bool success)
...
+				_delegate = @delegate;
+			}
+
+			public void ExecuteBeforeTransactionCompletion()
+			{
+				_delegate?.Invoke();
+			}
+		}
+
+		private partial class AfterTransactionCompletionDelegatedProcess : IAfterTransactionCompletionProcess
```
fredericDelaporte: I would obsolete these classes too.

fredericDelaporte:
This seems all good. Any reasons for not merging it yet?

PR #1453, opened 2017-11-24
Trying to eliminate unnecessary AsyncLocal allocation if SessionId isn't changed

hazzik:
This might fix the unnecessary allocations~, but I've never checked that~.

### DIFF
--- a/src/NHibernate/Impl/SessionIdLoggingContext.cs
+++ b/src/NHibernate/Impl/SessionIdLoggingContext.cs
@@ -1,15 +1,22 @@
 using System;
+#if !NETSTANDARD2_0 && !NETCOREAPP2_0 
+using System.Runtime.Remoting.Messaging;
+#else
 using System.Threading;
+#endif
 
 namespace NHibernate.Impl
 {
 	public class SessionIdLoggingContext : IDisposable
 	{
+#if NETSTANDARD2_0 || NETCOREAPP2_0
 		private static readonly Lazy<AsyncLocal<Guid?>> _currentSessionId =
 			new Lazy<AsyncLocal<Guid?>>(() => new AsyncLocal<Guid?>(), true);
-
+#else
+		private const string LogicalCallContextVariableName = "__" + nameof(SessionIdLoggingContext) + "__";
+#endif
 		private readonly Guid? _oldSessonId;
-		private bool _hasChanged;
+		private readonly bool _hasChanged;
 		
 		public SessionIdLoggingContext(Guid id)
 		{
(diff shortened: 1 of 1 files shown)


hazzik:
After running a profiler it seems that this solution isn't viable.

hazzik:
Ok, this "solution" is performing +1.53% worth than original by time.

hazzik:
I turns out that `CallContext.LogicalSetData`/`CallContext.LogicalGetData` is much more performant than `AsyncLocal<T>`

src/NHibernate/Impl/SessionIdLoggingContext.cs
```diff
@@ -1,18 +1,31 @@
...
 using System.Threading;
+#endif
 
 namespace NHibernate.Impl
 {
 	public class SessionIdLoggingContext : IDisposable
 	{
-		private static readonly AsyncLocal<Guid?> _currentSessionId = new AsyncLocal<Guid?>();
-
+#if NETSTANDARD2_0 || NETCOREAPP2_0
```
hazzik: .net standard and .net core do not support CallContext

hazzik:
I want this change to be introduced in a bugfix release (and release it sooner).

>Whether we accept introducing those target framework directives for this case, or if we prefer [to] avoid them as much as possible (so wait for maca solution to be released or for another one).

I introduced them beforehand as I know we're working on the .net standard version. 

> Is it still worth checking the value has changed before setting it?

Yes.. We already read the old value to store it for the future, so no extra overhead there. Then it saves us the Dispose call. So in the first version, it 1.8% of performance. Apparently _reading_ the context is most expensive operation.

> The main overhead seems to be having the context containing some data, which then gets copied at each context switch anyway.

Yeah, this overhead is still here, but it's much less than the overhead of `AsyncLocal<T>`. Someone on internet noted that `CallContext` works 5 times faster than `AsyncLocal<T>`. And it does not generate  memory traffic as opposed to `AsyncLocal<T>` (at least in _that_ scenario) - the memory consumption is below v4.1 now: 514 269 KB (526 612 072 bytes)

>If putting this solution in place, does maca's solution still provides additional significant savings? 

Yes, as above - the most expensive operation is reading the context. And we still have it here. As you've noted its 18% (here) vs 28% (when completely disabled). So we need to merge these solutions together.

maca88:
@hazzik I did some testing and it's working fine. 👍

hazzik:
Hi @stephencleary I’ve seen [your comment](https://stackoverflow.com/questions/37009413/callcontext-logicalgetdata-vs-threadlocal-threadstatic#comment66193557_37034175) on SO about the performance of `CallContext` vs `AsyncLocal`. It turns out that in our scenario it adds up a very significant footprint: +300MB memory allocated and a lot of CPU wasted.

StephenCleary:
@hazzik From the [source code](https://referencesource.microsoft.com/#mscorlib/system/threading/asynclocal.cs), I just don't see how that's possible.

hazzik:
@StephenCleary most of the CPU is wasted in dictionary operations in [ExecutionContext](https://referencesource.microsoft.com/#mscorlib/system/threading/executioncontext.cs,dcc44d583d37c571). I don't know where the allocations are coming from. But the fact is, that with `AsyncLocal<>` we got +300 MB allocated and about 25% regression by CPU.

hazzik:
Maybe this copying of the dictionary is responsible: https://referencesource.microsoft.com/#mscorlib/system/threading/executioncontext.cs,683

PR #1454, opened 2017-11-24
Fix ProxyFactory cache

hazzik:
- use a correct implementation of cache
- reduce the number of allocations of ProxyCacheEntry
- reduce memory flow

ProxyCache has a weird, inefficient and _incorrect_ implementation. Replaced with just ConcurrentDictionary.
Removed unnecessary collections manipulations to get some perf boost.
Also, previously there were 4 instances of ProxyCacheEntry created per one call to CreateProxyType. But the constructor of ProxyCacheEntry is expensive in terms of both memory and CPU.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3954/EqualsFixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3954/EqualsFixture.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Reflection;
+using System.Runtime.Serialization;
 using NHibernate.Proxy;
 using NHibernate.Proxy.DynamicProxy;
 using NHibernate.Type;
@@ -48,9 +49,9 @@ namespace NHibernate.Test.NHSpecificTest.NH3954
 		[Test]
 		public void InterfaceEquality()
 		{
-			var entry1 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(INHibernateProxy), typeof(IProxy) });
+			var entry1 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(INHibernateProxy), typeof(ISerializable) });
 			// Interfaces order inverted on purpose: must be supported.
-			var entry2 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(IProxy), typeof(INHibernateProxy) });
+			var entry2 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(ISerializable), typeof(INHibernateProxy) });
 			Assert.IsTrue(entry1.Equals(entry2));
 			Assert.IsTrue(entry2.Equals(entry1));
 		}
@@ -59,9 +60,9 @@ namespace NHibernate.Test.NHSpecificTest.NH3954
 		[Test]
 		public void InterfaceEqualityWithLotOfUnordererdAndDupInterfaces()
 		{
-			var entry1 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(INHibernateProxy), typeof(IProxy), typeof(IType), typeof(IDisposable), typeof(IFilter) });
+			var entry1 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(INHibernateProxy), typeof(ISerializable), typeof(IType), typeof(IDisposable), typeof(IFilter) });
 			// Interfaces order inverted and duplicated on purpose: must be supported.
-			var entry2 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(IType), typeof(IProxy), typeof(IFilter), typeof(IDisposable), typeof(IType), typeof(IFilter), typeof(INHibernateProxy) });
+			var entry2 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(IType), typeof(ISerializable), typeof(IFilter), typeof(IDisposable), typeof(IType), typeof(IFilter), typeof(INHibernateProxy) });
 			Assert.IsTrue(entry1.Equals(entry2));
 			Assert.IsTrue(entry2.Equals(entry1));
 		}
@@ -69,8 +70,8 @@ namespace NHibernate.Test.NHSpecificTest.NH3954
 		[Test]
 		public void InterfaceInequality()
 		{
-			var entry1 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(INHibernateProxy), typeof(IProxy) });
-			var entry2 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(IProxy) });
+			var entry1 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(INHibernateProxy), typeof(ISerializable) });
+			var entry2 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(ISerializable) });
 			TweakEntry(entry2, entry1.GetHashCode());
 			Assert.IsFalse(entry1.Equals(entry2));
 			Assert.IsFalse(entry2.Equals(entry1));
--- a/src/NHibernate.Test/NHSpecificTest/NH3954/ProxyCacheFixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3954/ProxyCacheFixture.cs
@@ -1,34 +1,36 @@
-﻿using System.Collections.Concurrent;
+﻿using System;
+using System.Collections.Concurrent;
 using System.Reflection;
+using System.Runtime.Serialization;
 using NHibernate.Proxy;
 using NHibernate.Proxy.DynamicProxy;
 using NUnit.Framework;
 
 namespace NHibernate.Test.NHSpecificTest.NH3954
 {
-	[TestFixture, Explicit("Demonstrates bug impact on cache, but which tests will fail is a bit unpredictable")]
+	[TestFixture, Explicit("Demonstrates bug impact on cache, but which tests will fail is a bit unpredictable"), Obsolete]
 	public class ProxyCacheFixture
 	{
 		private ProxyCache _cache;
-		private ConcurrentDictionary<ProxyCacheEntry, System.Type> _internalCache;
+		private ConcurrentDictionary<ProxyCacheEntry, TypeInfo> _internalCache;
 		private int _hashCode1;
 		private int _hashCode2;
 
 		private static readonly FieldInfo InternalCacheField =
-			typeof(ProxyCache).GetField("cache", BindingFlags.Static | BindingFlags.NonPublic);
+			typeof(ProxyFactory).GetField("_cache", BindingFlags.Static | BindingFlags.NonPublic);
 
 		[SetUp]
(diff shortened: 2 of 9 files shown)


src/NHibernate/Proxy/DynamicProxy/HashSetExtensions.cs
```diff
@@ -2,15 +2,14 @@
 
 namespace NHibernate.Proxy.DynamicProxy
 {
+	//TODO: Mark obsolete in v5.1
 	public static class HashSetExtensions
```
hazzik: Not used anymore

src/NHibernate/Proxy/DynamicProxy/IProxyCache.cs
```diff
@@ -11,6 +11,7 @@
 
 namespace NHibernate.Proxy.DynamicProxy
 {
+	//TOOD: Mark as obsolete in v5.1
 	public interface IProxyCache
```
hazzik: Not used anymore

src/NHibernate/Proxy/DynamicProxy/ProxyCache.cs
```diff
@@ -6,32 +6,27 @@
...
 #endregion
 
-using System.Collections.Concurrent;
+using System;
 using System.Reflection;
 
 namespace NHibernate.Proxy.DynamicProxy
 {
+	//TOOD: Mark as obsolete in v5.1
 	public class ProxyCache : IProxyCache
```
hazzik: Not used anymore

hazzik:
ProxyCache has a weird, inefficient and _incorrect_ implementation. Replaced with just ConcurrentDictioanry of Lazy.
Remove unnecessary collections manipulations to get some perf boost.
Also, previously there were 4 instances of ProxyCacheEntry created per one call to CreateProxyType. But the constructor of ProxyCacheEntry is expensive in terms of both memory and CPU.

hazzik:
Going to include into 5.1 instead.

hazzik:
Would be included in #1451 in some form.

hazzik:
As #1451 now uses completely different approach, I’m going to remove changes related to this PR and move them here.

src/NHibernate.Test/NHSpecificTest/NH3954/EqualsFixture.cs
```diff
@@ -48,9 +49,9 @@ public void TypeInequality()
 		[Test]
 		public void InterfaceEquality()
 		{
-			var entry1 = new ProxyCacheEntry(typeof(Entity1), new[] { typeof(INHibernateProxy), typeof(IProxy) });
```
hazzik: Just wanted to reduce usages of `DynamicProxy.IProxy` as this interface would be dead in a few versions.

hazzik: Just wanted to reduce usages of `DynamicProxy.IProxy` as this interface would be dead in a few versions.

src/NHibernate.Test/NHSpecificTest/NH3954/ProxyCacheFixture.cs
```diff
@@ -1,34 +1,36 @@
...
 using System.Reflection;
+using System.Runtime.Serialization;
 using NHibernate.Proxy;
 using NHibernate.Proxy.DynamicProxy;
 using NUnit.Framework;
 
 namespace NHibernate.Test.NHSpecificTest.NH3954
 {
-	[TestFixture, Explicit("Demonstrates bug impact on cache, but which tests will fail is a bit unpredictable")]
+	[TestFixture, Explicit("Demonstrates bug impact on cache, but which tests will fail is a bit unpredictable"), Obsolete]
```
hazzik: Had to mark it as obsolete as `ProxyCache` class now is marked obsolete

src/NHibernate.Test/NHSpecificTest/NH3954/ProxyCacheFixture.cs
```diff
@@ -1,34 +1,36 @@
...
 	public class ProxyCacheFixture
 	{
 		private ProxyCache _cache;
-		private ConcurrentDictionary<ProxyCacheEntry, System.Type> _internalCache;
+		private ConcurrentDictionary<ProxyCacheEntry, TypeInfo> _internalCache;
```
hazzik: The test fixture itself was failing (but because of `[Explicit]` attribute it did not run and we did not notice)

src/NHibernate.Test/NHSpecificTest/NH3954/ProxyCacheFixture.cs
```diff
@@ -1,34 +1,36 @@
...
 		private int _hashCode1;
 		private int _hashCode2;
 
 		private static readonly FieldInfo InternalCacheField =
-			typeof(ProxyCache).GetField("cache", BindingFlags.Static | BindingFlags.NonPublic);
+			typeof(ProxyFactory).GetField("_cache", BindingFlags.Static | BindingFlags.NonPublic);
```
hazzik: The field has moved to `ProxyFactory`

src/NHibernate.Test/NHSpecificTest/NH3954/ProxyCacheFixture.cs
```diff
@@ -1,34 +1,36 @@
...
 		{
 			_cache = new ProxyCache();
 
-			_internalCache = (ConcurrentDictionary<ProxyCacheEntry, System.Type>)InternalCacheField.GetValue(null);
+			_internalCache = (ConcurrentDictionary<ProxyCacheEntry, TypeInfo>)InternalCacheField.GetValue(null);
 
 			_cache.StoreProxyType(typeof(Entity1FakeProxy).GetTypeInfo(), typeof(Entity1));
 			_cache.StoreProxyType(typeof(Entity2FakeProxy).GetTypeInfo(), typeof(Entity2), typeof(INHibernateProxy));
 			_cache.StoreProxyType(typeof(Entity3FakeProxy).GetTypeInfo(), typeof(Entity3));
-			_cache.StoreProxyType(typeof(Entity4FakeProxy).GetTypeInfo(), typeof(Entity4), typeof(IProxy));
```
hazzik: Again, wanted to reduce usages of `IProxy`

src/NHibernate/Proxy/DynamicProxy/DefaultMethodEmitter.cs
```diff
@@ -10,15 +10,12 @@
...
 using System.Reflection;
 using System.Reflection.Emit;
-using NHibernate.Linq;
 using NHibernate.Util;
 
 namespace NHibernate.Proxy.DynamicProxy
 {
 	internal class DefaultMethodEmitter : IMethodBodyEmitter
 	{
-		private static readonly MethodInfo getInterceptor;
```
hazzik: Use the field directly instead.

src/NHibernate/Proxy/DynamicProxy/HashSetExtensions.cs
```diff
@@ -1,9 +1,12 @@
+using System;
 using System.Collections.Generic;
 
 namespace NHibernate.Proxy.DynamicProxy
 {
+	[Obsolete("This class is not used anymore and will be removed in a next major version")]
 	public static class HashSetExtensions
```
hazzik: Has no more usages

src/NHibernate/Proxy/DynamicProxy/ProxyFactory.cs
```diff
@@ -151,32 +129,14 @@ private TypeInfo CreateUncachedProxyType(System.Type baseType, System.Type[] bas
 			return proxyType;
 		}
 
-		private IEnumerable<System.Type> GetInterfaces(System.Type currentType)
-		{
-			return GetAllInterfaces(currentType);
-		}
-
-		private IEnumerable<System.Type> GetAllInterfaces(System.Type currentType)
```
hazzik: `System.Type.GetInterfaces()` already returns _all_ implemented interfaces, so no need to further traverse the interfaces tree.

src/NHibernate/Async/Engine/Cascade.cs
```diff
@@ -9,7 +9,7 @@
 
 
 using System.Collections;
-
+using System.Collections.Generic;
```
hazzik: Unrelated, but "let eat bee" here

src/NHibernate/Proxy/DynamicProxy/IProxyCache.cs
```diff
@@ -11,13 +11,16 @@
 
 namespace NHibernate.Proxy.DynamicProxy
 {
+	//Since v5.1
+	[Obsolete("This class is not used anymore and will be removed in a next major version")]
```
fredericDelaporte: Seems to be more an interface than a class. But that is just a detail.

src/NHibernate/Proxy/DynamicProxy/ProxyFactory.cs
```diff
@@ -35,62 +37,43 @@ public ProxyFactory(IProxyAssemblyBuilder proxyAssemblyBuilder)
...
-			
-			TypeInfo proxyTypeInfo;
+			if (baseType == null) throw new ArgumentNullException(nameof(baseType));
 
-			// Reuse the previous results, if possible, Fast path without locking.
-			if (Cache.TryGetProxyType(baseType, baseInterfaces, out proxyTypeInfo))
-				return proxyTypeInfo;
+			var typeFactory = _cache.GetOrAdd(
+				new ProxyCacheEntry(baseType, interfaces),
+				k => CreateUncachedProxyType(k.BaseType, k.Interfaces));
```
fredericDelaporte: We have overlooked a potential issue here. `ConcurrentDictionary` does not guarantee the factory method will be called only once per key. It may be called multiple times if concurrent threads attempts to add the same key. `ConcurrentDictionary` only guarantee it will yield the same value to all threads.

So concurrency may leads to try creating the proxy multiple times.
I think we should fix it with this [SO answer](https://stackoverflow.com/a/31637510/1178314).

hazzik: Yes I know about that issue and a solution. No, I did not overlook it. I have considered concurrent dictionary of lazy, end even coded it that way somewhere in between. Unfortunately lazy wasn’t so good for performance reasons. Therefore I decided, that lazy isn’t really necessary there.

fredericDelaporte: Do you mean creating multiple time the same proxy and its assembly is safe? I do not know it surely when reading the code and the documentation of assembly building methods.

hazzik: Yes, it is safe. For each proxy we create a new assembly (I think, it is not very efficient, but I’m not sure if we can create only one assembly) with a songle type. There will be no harm if we create the same twice. As I understand, at the end we will end up with a single type that were created faster and filled the place is a dictionary.

Issue #1446, opened 2017-11-20
A performance regression in v5


FransBouma:
Hi,

Using my raw dataaccess benchmarks (https://github.com/FransBouma/RawDataAccessBencher) which I just updated to include NHv5, latest stable, I saw the performance of NHibernate fell dramatically. v4.1 was slightly faster than EF6, but NHv5 is slower than EF6 and also consumes even more memory. 

Below is a screenshot of the set fetch of 31K rows from AdventureWorks' SalesOrderHeader, synchronously in change tracked entities. I've expanded the tree to a location which illustrates that there's something seriously wrong: with 31K rows in a *single query* it's weird the code ends up 410,000 (!) times in the method which checks whether something has to be enlisted into an ambient transaction. There's just 1 (synchronous) query. 

![nh5setfetchperf](https://user-images.githubusercontent.com/3628530/33012217-8d2dec20-cde0-11e7-946e-3875228d81f1.jpg)

You can easily profile the benchmarked fetches with a profiler, just uncomment this line: https://github.com/FransBouma/RawDataAccessBencher/blob/master/RawBencher/OriginalController.cs#L77 and follow the directions on screen. 

Hope this helps to gain some insight what to fix. Cheers.

gliljas:
@FransBouma Thanks! We'll certainly look into this.

hazzik:
@FransBouma can you make the same screenshot for the NH 4.1?

fredericDelaporte:
There is also the 58.66% part which is not deployed which could be interesting to see.

hazzik:
>I've expanded the tree to a location which illustrates that there's something seriously wrong: with 31K rows in a single query it's weird the code ends up 410,000 (!) times

Official NH's position is that the quote about should be just:

>there's something seriously wrong ... 31K rows in a single query.

My guess is that 410,000 figure is the number of entities being fetched.

fredericDelaporte:
Thanks for the notice.

While the redundant `CheckSessionAndUpdateStatus` can certainly be changed for being way less fired (maybe with some "depth" counter in the session entry points for disabling it for sub-operations), this thing is nothing new. But it does some more checking in v5, like the transaction enlistment.

[...]

PR #1455, opened 2017-11-25
Reduces check session and set context id redundant calls

fredericDelaporte:
* Reduces object allocations.

Two helper methods have been added on the abstract session impl: `BeginProcess` and `BeginContext`.
Both yields an `IDisposable` if we are not already in a "processing", otherwise they yield `null`. `using` supports this. This avoids allocating that small helper object when there is no work to do. ~~The suspend auto flush helper has been changed accordingly~~ (debatable, see review comment to come).
If not already processing, `BeginProcess` does check the status of the session, stores its id in context, then flag the session as "processing". `BeginContext` just stores its id in context if we are not already processing. `BeginContext` neither alters the processing flag, nor has a flag of its own (for simplicity).

For the session id context, skipping setting it again when it has already been set through an ongoing "processing" should not change anything to the feature (excepted its impact on perf): if, through some event or interceptor, it is changed by another session usage in such an event/interceptor while the first session is processing, its current logic will set back the first session id when ending the other session usage, which will forcibly occurs before the event or interceptor ends. (Unless an await on that other session is missing, but that would be a user bug even without this change.)

For the `CheckAndUpdateSessionStatus` status case, not running it again during a processing may weaken the error handling in case the user exploits an event or an interceptor to close the session or start a scope without completing it, while the session is "processing". I think it is acceptable. Closing the session during one of its event or interceptor call is clearly calling for troubles, so long for the user. Starting a scope outside of a using and inside an event or interceptor call is a bit exotic too.

`CheckAndUpdateSessionStatus` is still there and directly called in some cases (mainly properties access), and is now a no-op if the session is already "processing".
`new SessionIdLoggingContext(sessionId)` is now only directly called by `AbstractSessionImpl` and by the `AdoTransaction` which some methods may have to work without a session but with a stored session id.

`BeginProcess` and `BeginContext` are to be added to `ISessionImplementor` at some later point, but are now added as just extension methods of `ISessionImplementor` (avoiding a breaking change).

Many cases who were just setting the session id have been converted to `BeginProcess`, because they were calling some methods ending up doing that. In such case, better call `BeginProcess` the soonest, for reducing the most session id calls.
Some cases were pointlessly setting the id or checking the status, because their were doing nothing susceptible to log anything or actually do some processing, excepted calling a single method which was already doing the `BeginProcess`. (If it calls many methods doing it, then it should still call `BeginProcess` for achieving best savings.) They have been cleaned up of the call.
Some cases were calling the check status or were setting the id while being private and always called by methods doing it themselves.  They have been cleaned up of the call.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/Logs/LogsFixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/Logs/LogsFixture.cs
@@ -2,8 +2,11 @@ using System.Collections;
 using System.Collections.Concurrent;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
+using NHibernate.Cfg;
 using NHibernate.Impl;
-
+using NHibernate.SqlCommand;
+using NHibernate.Type;
 using NUnit.Framework;
 
 namespace NHibernate.Test.NHSpecificTest.Logs
@@ -30,6 +33,33 @@ namespace NHibernate.Test.NHSpecificTest.Logs
 			get { return "NHibernate.Test"; }
 		}
 
+		protected override void Configure(Configuration configuration)
+		{
+			base.Configure(configuration);
+			configuration.SetProperty(Cfg.Environment.UseSecondLevelCache, "false");
+		}
+
+		protected override void OnSetUp()
+		{
+			using (var s = Sfi.OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				s.Save(new Person());
+				s.Save(new Person());
+				t.Commit();
+			}
+		}
+
+		protected override void OnTearDown()
+		{
+			using (var s = Sfi.OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				s.CreateQuery("delete from Person").ExecuteUpdate();
+				t.Commit();
+			}
+		}
+
 		[Test]
 		public void WillGetSessionIdFromSessionLogs()
 		{
@@ -38,12 +68,68 @@ namespace NHibernate.Test.NHSpecificTest.Logs
 			using (var spy = new TextLogSpy("NHibernate.SQL", "%message | SessionId: %property{sessionId}"))
 			using (var s = Sfi.OpenSession())
 			{
-				var sessionId = ((SessionImpl)s).SessionId;
+				var sessionId = ((SessionImpl) s).SessionId;
 
-				s.Get<Person>(1);//will execute some sql
+				s.Get<Person>(1); //will execute some sql
 
 				var loggingEvent = spy.GetWholeLog();
 				Assert.That(loggingEvent.Contains(sessionId.ToString()), Is.True);
+			}
+		}
+
+		[Test]
+		public void WillGetSessionIdFromConsecutiveSessionsLogs()
+		{
+			GlobalContext.Properties["sessionId"] = new SessionIdCapturer();
+
+			using (var spy = new TextLogSpy("NHibernate.SQL", "%message | SessionId: %property{sessionId}"))
+			{
+				var sessions = Enumerable.Range(1, 10).Select(i => Sfi.OpenSession()).ToArray();
+				try
+				{
+					for (var i = 0; i < 10; i++)
+					for (var j = 0; j < 10; j++)
+					{
+						var s = sessions[j];
+						s.Get<Person>(i * 10 + j); //will execute some sql
+					}
+				}
+				finally
+				{
+					foreach (var s in sessions)
+					{
+						s.Dispose();
+					}
+				}
+
+				var loggingEvent = spy.GetWholeLog();
+				for (var i = 0; i < 10; i++)
+				for (var j = 0; j < 10; j++)
+				{
+					var sessionId = sessions[j].GetSessionImplementation().SessionId;
+					Assert.That(loggingEvent, Does.Contain($"p0 = {i * 10 + j} [Type: Int32 (0:0:0)] | SessionId: {sessionId}"));
+				}
+			}
+		}
+
+		[Test]
+		public void WillGetSessionIdFromInterlacedSessionsLogs()
+		{
+			GlobalContext.Properties["sessionId"] = new SessionIdCapturer();
+			var interceptor = new InterlacedSessionInterceptor(Sfi);
+			using (var spy = new TextLogSpy("NHibernate.SQL", "%message | SessionId: %property{sessionId}"))
+			using (var s = Sfi.WithOptions().Interceptor(interceptor).OpenSession())
+			{
+				// Trigger an operation which will fire many interceptor events, before and after s own logging.
+				var persons = s.Query<Person>().ToList();
+
+				var loggingEvent = spy.GetWholeLog();
+				for (var i = 0; i < interceptor.SessionIds.Count; i++)
+				{
+					var sessionId = interceptor.SessionIds[i];
+					Assert.That(loggingEvent, Does.Contain($"p0 = {i + 1} [Type: Int32 (0:0:0)] | SessionId: {sessionId}"));
+				}
+				Assert.That(loggingEvent, Does.Contain($"Person person0_ | SessionId: {s.GetSessionImplementation().SessionId}"));
 			}
 		}
 
@@ -52,13 +138,24 @@ namespace NHibernate.Test.NHSpecificTest.Logs
 		{
 			GlobalContext.Properties["sessionId"] = new SessionIdCapturer();
 
-			var semaphore = new ManualResetEventSlim();
+			// Do not use a ManualResetEventSlim, it does not support async and exhausts the task thread pool in the
+			// async counterparts of this test. SemaphoreSlim has the async support and release the thread when waiting.
+			var semaphore = new SemaphoreSlim(0);
 			var failures = new ConcurrentBag<Exception>();
 			var sessionIds = new ConcurrentDictionary<int, Guid>();
-			var array = Enumerable.Range(1, 10).Select(
-				i => new Thread(
-					() =>
+			using (var spy = new TextLogSpy("NHibernate.SQL", "%message | SessionId: %property{sessionId}"))
+			{
+				Parallel.For(
+					1, 12,
+					i =>
 					{
+						if (i > 10)
+						{
+							// Give some time to threads for reaching the wait, having all of them ready to do most of their job concurrently.
+							Thread.Sleep(100);
+							semaphore.Release(10);
+							return;
+						}
 						try
 						{
 							using (var s = Sfi.OpenSession())
@@ -81,17 +178,9 @@ namespace NHibernate.Test.NHSpecificTest.Logs
 						{
 							failures.Add(e);
 						}
-					})).ToArray();
+					});
 
-			using (var spy = new TextLogSpy("NHibernate.SQL", "%message | SessionId: %property{sessionId}"))
-			{
-				Array.ForEach(array, thread => thread.Start());
-				// Give some time to threads for reaching the wait, having all of them ready to do most of their job concurrently.
-				Thread.Sleep(100);
-				semaphore.Set();
-				Array.ForEach(array, thread => thread.Join());
-
-				Assert.That(failures, Is.Empty, $"{failures.Count} thread(s) failed.");
+				Assert.That(failures, Is.Empty, $"{failures.Count} task(s) failed.");
 
 				var loggingEvent = spy.GetWholeLog();
 				for (var i = 1; i < 11; i++)
@@ -130,7 +219,7 @@ namespace NHibernate.Test.NHSpecificTest.Logs
 					Threshold = Level.All,
 					Writer = new StringWriter(stringBuilder)
 				};
-				loggerImpl = (Logger)LogManager.GetLogger(typeof(LogsFixture).Assembly, loggerName).Logger;
+				loggerImpl = (Logger) LogManager.GetLogger(typeof(LogsFixture).Assembly, loggerName).Logger;
 				loggerImpl.AddAppender(appender);
 				previousLevel = loggerImpl.Level;
 				loggerImpl.Level = Level.All;
@@ -147,7 +236,37 @@ namespace NHibernate.Test.NHSpecificTest.Logs
 				loggerImpl.Level = previousLevel;
 			}
 		}
+
+		public class InterlacedSessionInterceptor : EmptyInterceptor
+		{
+			private readonly ISessionFactory _sfi;
+
+			public System.Collections.Generic.List<Guid> SessionIds { get; } = new System.Collections.Generic.List<Guid>();
+
+			public InterlacedSessionInterceptor(ISessionFactory sfi)
+			{
+				_sfi = sfi;
+			}
+
+			public override SqlString OnPrepareStatement(SqlString sql)
+			{
+				using (var s = _sfi.OpenSession())
+				{
+					SessionIds.Add(s.GetSessionImplementation().SessionId);
+					s.Get<Person>(SessionIds.Count); //will execute some sql
+				}
+				return base.OnPrepareStatement(sql);
+			}
+
+			public override bool OnLoad(object entity, object id, object[] state, string[] propertyNames, IType[] types)
+			{
+				using (var s = _sfi.OpenSession())
+				{
+					SessionIds.Add(s.GetSessionImplementation().SessionId);
+					s.Get<Person>(SessionIds.Count); //will execute some sql
+				}
+				return base.OnLoad(entity, id, state, propertyNames, types);
+			}
+		}
 	}
-
-
 }
--- a/src/NHibernate.Test/NHSpecificTest/Logs/LogsFixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/Logs/LogsFixture.cs
@@ -1,5 +1,7 @@
 using System.Collections;
-
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Threading;
 using NHibernate.Impl;
 
 using NUnit.Framework;
@@ -42,6 +44,62 @@ namespace NHibernate.Test.NHSpecificTest.Logs
 
 				var loggingEvent = spy.GetWholeLog();
 				Assert.That(loggingEvent.Contains(sessionId.ToString()), Is.True);
+			}
+		}
+
+		[Test]
+		public void WillGetSessionIdFromSessionLogsConcurrent()
+		{
+			GlobalContext.Properties["sessionId"] = new SessionIdCapturer();
+
+			var semaphore = new ManualResetEventSlim();
+			var failures = new ConcurrentBag<Exception>();
+			var sessionIds = new ConcurrentDictionary<int, Guid>();
+			var array = Enumerable.Range(1, 10).Select(
+				i => new Thread(
+					() =>
+					{
+						try
+						{
+							using (var s = Sfi.OpenSession())
+							{
+								sessionIds.AddOrUpdate(
+									i,
+									s.GetSessionImplementation().SessionId,
+									(ti, old) => throw new InvalidOperationException(
+										$"Thread number {ti} has already session id {old}, while attempting to set it to" +
+										$" {s.GetSessionImplementation().SessionId}"));
+								semaphore.Wait();
+
+								for (int j = 0; j < 10; j++)
+								{
+									s.Get<Person>(i * 10 + j); //will execute some sql
+								}
+							}
+						}
+						catch (Exception e)
+						{
+							failures.Add(e);
+						}
+					})).ToArray();
+
+			using (var spy = new TextLogSpy("NHibernate.SQL", "%message | SessionId: %property{sessionId}"))
+			{
+				Array.ForEach(array, thread => thread.Start());
+				// Give some time to threads for reaching the wait, having all of them ready to do most of their job concurrently.
+				Thread.Sleep(100);
+				semaphore.Set();
+				Array.ForEach(array, thread => thread.Join());
+
+				Assert.That(failures, Is.Empty, $"{failures.Count} thread(s) failed.");
+
+				var loggingEvent = spy.GetWholeLog();
+				for (var i = 1; i < 11; i++)
+				for (var j = 0; j < 10; j++)
+				{
+					var sessionId = sessionIds[i];
+					Assert.That(loggingEvent, Does.Contain($"p0 = {i * 10 + j} [Type: Int32 (0:0:0)] | SessionId: {sessionId}"));
+				}
 			}
 		}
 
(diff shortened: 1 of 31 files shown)


hazzik:
The purpose of SessionIdLoggingContext was to log a session ID in a multythreaded environments, in a web app for example, where there can be many sessions simultaneously. I have not yet checked your implementation, but based on what I’ve read I’m not sure that it would work an that scenario.

fredericDelaporte:
It does change nothing to its working in such context. That purpose is supported by the usage of `ThreadLocal` (4.x impl)/`AsyncLocal` (5.0 & 5.0.1)/`CallContext` (your patch).
Setting it less frequently will not cause thread contamination.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -380,31 +424,31 @@ public virtual ISQLQuery CreateSQLQuery(string sql)
 
 		protected internal virtual IQueryExpressionPlan GetHQLQueryPlan(IQueryExpression queryExpression, bool shallow)
 		{
-			using (new SessionIdLoggingContext(SessionId))
+			using (BeginContext())
 			{
 				return _factory.QueryPlanCache.GetHQLQueryPlan(queryExpression, shallow, EnabledFilters);
```
fredericDelaporte: For this case I was tempted to put a `BeginProcess`, because it does not really make sens to get the query plan on a closed session. But for sticking to previous semantic, and since it apparently calls nothing that does the session status check or requires it to be opened, I left it as is.
But checking again, `EnabledFilters` does check the session status in the `SessionImpl` override, so I am now switching it to `BeginProcess`.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -380,31 +424,31 @@ public virtual ISQLQuery CreateSQLQuery(string sql)
...
 			}
 		}
 
 		protected internal virtual NativeSQLQueryPlan GetNativeSQLQueryPlan(NativeSQLQuerySpecification spec)
 		{
-			using (new SessionIdLoggingContext(SessionId))
+			using (BeginContext())
 			{
 				return _factory.QueryPlanCache.GetNativeSQLQueryPlan(spec);
```
fredericDelaporte: Similar to previous case, but actually never triggering a check status, so leaving it as is.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -380,31 +424,31 @@ public virtual ISQLQuery CreateSQLQuery(string sql)
...
+			using (BeginContext())
 			{
 				return ADOExceptionHelper.Convert(_factory.SQLExceptionConverter, sqlException, message);
 			}
 		}
 
 		protected void AfterOperation(bool success)
 		{
-			using (new SessionIdLoggingContext(SessionId))
+			using (BeginContext())
```
fredericDelaporte: Transaction completion may occur on a closed session, by example when disposing a session inside a scope. So no session status check here.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -422,15 +466,15 @@ protected void EnlistInAmbientTransactionIfNeeded()
 
 		public void JoinTransaction()
 		{
-			CheckAndUpdateSessionStatus();
-			_factory.TransactionFactory.ExplicitJoinSystemTransaction(this);
+			using (BeginProcess())
+				_factory.TransactionFactory.ExplicitJoinSystemTransaction(this);
```
fredericDelaporte: Missing status check here. Having a closed session joining a transaction is an error.

hazzik:
@fredericDelaporte could you please add a multi-threaded concurrent version of `NHibernate.Test.NHSpecificTest.Logs.Fixture.WillGetSessionIdFromSessionLogs`? `TextLogSpy` should have a unique text pattern including the pre-set ID. At assert it should check that all `SessionId`s logged are the same.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -184,7 +184,7 @@ void IDeserializationCallback.OnDeserialization(object sender)
 		internal SessionImpl(SessionFactoryImpl factory, ISessionCreationOptions options)
 			: base(factory, options)
 		{
-			using (new SessionIdLoggingContext(SessionId))
+			using (BeginContext())
```
fredericDelaporte: This constructor does check status, but as its latest step. So just setting the session id.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -318,7 +318,7 @@ public DbConnection Close()
 		/// </summary>
 		public override void AfterTransactionCompletion(bool success, ITransaction tx)
 		{
-			using (new SessionIdLoggingContext(SessionId))
+			using (BeginContext())
```
fredericDelaporte: As already written, transaction completion may occur on closed session, so just setting the session id here.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1012,7 +977,7 @@ public void PersistOnFlush(object obj)
 
 		public override string BestGuessEntityName(object entity)
 		{
-			using (new SessionIdLoggingContext(SessionId))
+			using (BeginProcess())
```
fredericDelaporte: Does access the `persistenceContext`, thus my choice of switching it to `BeginProcess`. Now what it does likely can work on closed session, so I can revert it to a `BeginContext` if asked to.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -2300,15 +2186,13 @@ public bool DefaultReadOnly
 		public bool IsReadOnly(object entityOrProxy)
 		{
 			ErrorIfClosed();
-			// CheckTransactionSynchStatus();
 			return persistenceContext.IsReadOnly(entityOrProxy);
```
fredericDelaporte: I have hesitated to put this in a `BeginProcess` like the `BestGuessEntityName`. For scenario checking a list of entities, it could have a performance impact, so I have left it as is.

src/NHibernate/Impl/StatelessSessionImpl.cs
```diff
@@ -37,7 +37,7 @@ public partial class StatelessSessionImpl : AbstractSessionImpl, IStatelessSessi
 		internal StatelessSessionImpl(SessionFactoryImpl factory, ISessionCreationOptions options)
 			: base(factory, options)
 		{
-			using (new SessionIdLoggingContext(SessionId))
+			using (BeginContext())
```
fredericDelaporte: Same than the `SessionImpl` constructor.

src/NHibernate/Impl/StatelessSessionImpl.cs
```diff
@@ -251,25 +245,21 @@ public override void AfterTransactionCompletion(bool successful, ITransaction tx
 
 		public override object GetContextEntityIdentifier(object obj)
 		{
-			CheckAndUpdateSessionStatus();
 			return null;
```
fredericDelaporte: I am not seing the point of such a check for yielding `null`. May revert if someone see it better.

src/NHibernate/Impl/StatelessSessionImpl.cs
```diff
@@ -366,7 +358,7 @@ public override FlushMode FlushMode
 
 		public override string BestGuessEntityName(object entity)
 		{
-			using (new SessionIdLoggingContext(SessionId))
+			using (BeginProcess())
```
fredericDelaporte: Switch to `BeginProcess` for matching the method on `SessionImpl` side. Cannot be a breaking change because it is calling `GuessEntityName` which, in the case of the stateless session, was doing a status check for a mysterious reason.

src/NHibernate/Impl/StatelessSessionImpl.cs
```diff
@@ -379,8 +371,10 @@ public override string BestGuessEntityName(object entity)
 
 		public override string GuessEntityName(object entity)
 		{
-			CheckAndUpdateSessionStatus();
-			return entity.GetType().FullName;
+			using (BeginContext())
```
fredericDelaporte: Here that is the other way round: switching to `BeginContext` for matching the method on `SessionImpl`, and since what it does does not rely at all on the session state.

src/NHibernate/Persister/Collection/AbstractCollectionPersister.cs
```diff
@@ -1512,12 +1512,12 @@ private static void CalcPropertyColumnSpan(IType propertyType, ref int count)
...
 		public int GetSize(object key, ISessionImplementor session)
 		{
-			using(new SessionIdLoggingContext(session.SessionId))
+			using (session.BeginProcess())
 			try
 			{
 				if(session.EnabledFilters.Count > 0)
 				{
-					
+					// The above call validates the filters
```
fredericDelaporte: And it is also checking the status, but that is now already done due to `BeginProcess` above instead of just setting the id.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -309,7 +309,7 @@ protected virtual void Unlock()
 			/// <param name="preparingEnlistment">The object for notifying the prepare phase outcome.</param>
 			public virtual void Prepare(PreparingEnlistment preparingEnlistment)
 			{
-				using (new SessionIdLoggingContext(_session.SessionId))
+				using (_session.BeginContext())
```
fredericDelaporte: Due to the possibility of having already closed session, it must here call only `BeginContext`, even if in some case and depending on the configuration, it may do some later operation requiring an opened session. (Those operations will still do the check themselves.)

src/NHibernate/Transaction/AdoTransaction.cs
```diff
@@ -108,7 +108,7 @@ public void Begin()
 		/// </exception>
 		public void Begin(IsolationLevel isolationLevel)
 		{
-			using (new SessionIdLoggingContext(sessionId))
+			using (session.BeginProcess())
```
fredericDelaporte: Here using the stored session id is not useful, on `Begin` or `Commit` the ado transaction call only still have its session. And it must be opened.

hazzik:
Ok, I was trying to break this with following test:

```
[Test]
public void WillGetSessionIdFromSessionLogsConcurrent()
{
	GlobalContext.Properties["sessionId"] = new SessionIdCapturer();

	var semaphore = new ManualResetEventSlim();
	var array = Enumerable.Range(1, 10).Select(
		i => new Thread(
			() =>
			{
				using (var spy = new TextLogSpy("NHibernate.SQL", "%message | SessionId: %property{sessionId}"))
				using (var s = Sfi.OpenSession())
				{
					var sessionId = s.GetSessionImplementation().SessionId;
					semaphore.Wait();

					for (int j = 0; j < 10; j++)
					{
						s.Get<Person>(i * 10 + j); //will execute some sql
					}

					var loggingEvent = spy.GetWholeLog();
					for (int j = 0; j < 10; j++)
					{
						Assert.That(loggingEvent.Contains($"@p0 = {i * 10 + j} [Type: Int32 (0:0:0)] | SessionId: {sessionId}"), Is.True);
					}
				}
			})).ToArray();
	Array.ForEach(array, thread => thread.Start());
	semaphore.Set();
	Array.ForEach(array, thread => thread.Join());
}
```

It works.

fredericDelaporte:
With this PR, here are the changes in Frans benchmark, but run on my machine without a distant SQL Server contrary to what he recommends, for both runs of course. See #1446 for the issue triggering this PR.

Test | NH v5.0.1 | NH v5.0.1 altered by this PR  | % (base is unaltered 5.0.1)
Change tracking fetches, set fetches (50 runs), no caching | 2.477,37ms, Enum: 1,59ms, 828.952KB | 1.792,19ms, Enum: 1,63ms, 489.081KB | -27.6%ms, Enum: +2.5%, -41%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | 0,42ms, 96KB | 0,38ms, 80KB | -9.5%ms, -16.6%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | 381,94ms, 123.000KB | 261,02ms, 62.672KB | -31.7%ms, -49%KB

Thanks @hazzik for the additional test, I will add it.

fredericDelaporte:
While the table may be impressive, it should be note that we have another kind of benchmark: our builds, running thousands of tests. No sensible changes here. But our tests are way more near "normal orm usage" than the benchmark.

hazzik:
@fredericDelaporte I think all "possible breaking changes" shall be reverted, if we want to include this PR into 5.0.2.

fredericDelaporte:
The proposed test is not correct. Assert from spawned threads are not detected. The text log spy get logs from all threads and moreover does not support it. I am checking how to fix it.

fredericDelaporte:
Table updated with a test run on this pr with latest commits.

Test | NH v5.0.1 | NH v5.0.1 altered by this PR, v2  | % (base is unaltered 5.0.1)
Change tracking fetches, set fetches (50 runs), no caching | 2.477,37ms, Enum: 1,59ms, 828.952KB | 1.837,81ms, Enum: 1,53ms, 509.215KB | -25,8%ms, Enum: -0,38%, -38,6%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | 0,42ms, 96KB | 0,41ms, 80KB | -2,4%ms, -16,6%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | 381,94ms, 123.000KB | 254,74ms, 66.960KB | -33,3%ms, -45,6%KB

fredericDelaporte:
I have checked if #1453 or #1391 still give significant gain by applying them above this PR (in fact, by removing all code from the session context for simulating #1391).

Here are the results.

Test | NH v5.0.1 | NH v5.0.1 altered by this PR, v2 + #1453  | % (base is unaltered 5.0.1)
Change tracking fetches, set fetches (50 runs), no caching | 2.477,37ms, Enum: 1,59ms, 828.952KB | 1.770,55ms, Enum: 1,65ms, 490.488KB | -28,5%ms, Enum: +3,8%, -40,8%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | 0,42ms, 96KB | 0,35ms, 72KB | -16,6%ms, -25%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | 381,94ms, 123.000KB | 246,8ms, 62.992KB | -35,4%ms, -48,8%KB

Test | NH v5.0.1 | NH v5.0.1 altered by this PR, v2 + #1391  | % (base is unaltered 5.0.1)
Change tracking fetches, set fetches (50 runs), no caching | 2.477,37ms, Enum: 1,59ms, 828.952KB | 1.716,01ms, Enum: 1,55ms, 489.783KB | -30,8%ms, Enum: -2,5%, -40,9%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | 0,42ms, 96KB | 0,34ms, 72KB | -19%ms, -25%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | 381,94ms, 123.000KB | 250,09ms, 62.832KB | -34,5%ms, -48,9%KB

So, for comparing them all:

Test | NH v5.0.1 altered by this PR, v2 | NH v5.0.1 altered by this PR, v2 + #1453 | NH v5.0.1 altered by this PR, v2 + #1391
Change tracking fetches, set fetches (50 runs), no caching | -25,8%ms, Enum: -0,38%, -38,6%KB | -28,5%ms, Enum: +3,8%, -40,8%KB | -30,8%ms, Enum: -2,5%, -40,9%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | -2,4%ms, -16,6%KB | -16,6%ms, -25%KB | -19%ms, -25%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | -33,3%ms, -45,6%KB | -35,4%ms, -48,8%KB | -34,5%ms, -48,9%KB

Differences between this PR + #1453 and this PR + #1391 are quite minimal. (My first #1391 run was even worst on all tests so I have just guessed that some background process has interfered and I have done another run.) So I do not think we need to do both if the PR here is merged. Since #1391 will be useful for a .Net Standard version too, I would favor just keeping #1391. This would avoid introducing code depending on the target framework.

hazzik:
> Differences between this PR + #1453 and this PR + #1391 are quite minimal. [...] So I do not think we need to do both if the PR here is merged. Since #1391 will be useful for a .Net Standard version too, I would favor just keeping #1391. This would avoid introducing code depending on the target framework.

Don’t forget, that #1391 is just an option to disable the session tracking. If session tracking is enabled (or in .netstandard) you still get reduced performance (as pointed by a table with “bare” tests of this PR). So we need to keep a combined solution.

> The proposed test is not correct.

It was a really quick test.

fredericDelaporte:
Is should have formulated it another way, I was just willing to letknow that is was not yet validating mutli-threaded usage but that I was looking into it.

Why not combined all them too if you thinks its better. #1391 is already in master and #1453 is not a lot of code. #1455 impacts more code but render it a bit cleaner by the way, avoiding the dual "set id & check status" calls that were common place all around.

hazzik:
@fredericDelaporte can you please also add a test when there are several parallel sessions in the same thread?

hazzik:
Also, need to rewrite an async version of the test to be more async-ish (eg using Tasks instead of Threads).

fredericDelaporte:
Rebased, squashed and done.

src/NHibernate.Test/NHSpecificTest/Logs/LogsFixture.cs
```diff
@@ -38,27 +68,94 @@ public void WillGetSessionIdFromSessionLogs()
...
-					() =>
+			using (var spy = new TextLogSpy("NHibernate.SQL", "%message | SessionId: %property{sessionId}"))
+			{
+				Parallel.For(
+					1, 12,
+					i =>
 					{
+						if (i > 10)
+						{
+							// Give some time to threads for reaching the wait, having all of them ready to do most of their job concurrently.
```
fredericDelaporte: I have changed the thread array to a `Parallel.For` for exploiting maca88/AsyncGenerator#63. But since `Parallel.For` is blocking, this forbids releasing the semaphore afterward. So I have added the releasing as the last process in the `Parallel.For`. Since upper bound is excluded, it then needs to be twelve, for having ten threads and an eleventh one releasing them all.
The test for detecting we are the elventh one is `i > 10` instead of `i == 11` for avoiding having an issue in the async counterpart due to maca88/AsyncGenerator#72. Better release "more than needed" the semaphore than waiting too much on it.

src/NHibernate.Test/NHSpecificTest/Logs/LogsFixture.cs
```diff
@@ -30,6 +33,33 @@ protected override string MappingsAssembly
 			get { return "NHibernate.Test"; }
 		}
 
+		protected override void Configure(Configuration configuration)
+		{
+			base.Configure(configuration);
+			configuration.SetProperty(Cfg.Environment.UseSecondLevelCache, "false");
```
fredericDelaporte: `TestCase` base class causes all entities to be cacheable in second level cache by default, and enables it. Quite tricky for a test which relies on emitted SQL.

src/NHibernate.Test/NHSpecificTest/Logs/LogsFixture.cs
fredericDelaporte: Maca has already fixed it but we are on branch which is still using the 0.6 version, not the 0.6.2 version. Upgrading it here to 0.6.3 would cause many additional changes. So I will leave it as is in this PR.

fredericDelaporte:
For completing the comparison, here the table with 5.0.x branch current state (so with #1453 merged in):

Test | NH v5.0.1 | NH v5.0.1 + #1453  | % (base is unaltered 5.0.1)
Change tracking fetches, set fetches (50 runs), no caching | 2.477,37ms, Enum: 1,59ms, 828.952KB | 2.037,07ms, Enum: 1,63ms, 514.175KB | -17,8%ms, Enum: +2,5%, -38%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | 0,42ms, 96KB | 0,38ms, 80KB | -9,5%ms, -16,6%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | 381,94ms, 123.000KB | 410,29ms, 125.291KB | +7,4%ms, +1,9%KB

Overall comparison:

Test | NH v5.0.1 + #1453 | NH v5.0.1 altered by this PR, v2 | NH v5.0.1 altered by this PR, v2 + #1453 | NH v5.0.1 altered by this PR, v2 + #1391
------------ | ------------- | -------------|-------------|------------
Change tracking fetches, set fetches (50 runs), no caching | -17,8%ms, Enum: +2,5%, -38%KB | -25,8%ms, Enum: -0,38%, -38,6%KB | -28,5%ms, Enum: +3,8%, -40,8%KB | -30,8%ms, Enum: -2,5%, -40,9%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | -9,5%ms, -16,6%KB | -2,4%ms, -16,6%KB | -16,6%ms, -25%KB | -19%ms, -25%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | +7,4%ms, +1,9%KB | -33,3%ms, -45,6%KB | -35,4%ms, -48,8%KB | -34,5%ms, -48,9%KB

hazzik:
@fredericDelaporte could you please make a final comparison of 5.0.1 and the current state of the branch?

fredericDelaporte:
That is "NH v5.0.1 altered by this PR, v2 + #1453". But well, just in case, I have re-run it on my 5.0.2 branch (the one for the release PR).

Test | NH v5.0.1 | NH v5.0.2  | % (base is unaltered 5.0.1)
Change tracking fetches, set fetches (50 runs), no caching | 2.477,37ms, Enum: 1,59ms, 828.952KB | 1.726,01ms, Enum: 1,54ms, 490.487KB | -30,3%ms, Enum: -5%, -40.8%KB
Change tracking individual fetches (100 elements, 50 runs), no caching | 0,42ms, 96KB | 0,36ms, 72KB | -14,3%ms, -25%KB
Change tracking fetches, eager load fetches, 3-node split graph, 1000 root elements (50 runs), no caching | 381,94ms, 123.000KB | 247,76ms, 62.992KB | -35,1%ms, -48,8%KB

So, no significant changes compared to "NH v5.0.1 altered by this PR, v2 + #1453", all are below the standard deviation reported in the bench logs.

I have updated the tables of the previous comment: I had taken data from the async benchmark by error for the third line.

PR #302, opened 2014-08-14
NH-3435 - Ability to select the root entity in a criteria projection

rjperes:
https://nhibernate.jira.com/browse/NH-3435

PR #1458, opened 2017-11-28
#948(NH-3435) - Ability to select entities in a criteria projections (rebased PR #302)

### LINKED ISSUE #948: NH-3435 - Ability to select the root entity in a criteria projection
**Jonathan Hynes** created an [issue](https://nhibernate.jira.com/browse/NH-3435) — <time datetime="2013-04-01T13:44:03.748Z" title="2013-04-01T13:44:03.748Z">1st April 2013, 13:44:03</time>:
> The Criteria API need a way to select an Entity as a projection. Suggest the following methods:
> ```c#
> Projection.RootEntity()
> Projection.Entity(string associationPath)
> ```
> This issue may be a duplicate of one of the following, but these issues have been closed and the problem is not resolved:
> [NH-1372](https://nhibernate.jira.com/browse/NH-1372)
> [NH-928](https://nhibernate.jira.com/browse/NH-928)
> 
> Background to the issue can be found here:
> http://stackoverflow.com/questions/15744057/nhibernate-how-to-select-the-root-entity-in-a-projection
> 
---
**Ricardo Peres** added a comment — <time datetime="2014-03-06T20:46:57.344Z" title="2014-03-06T20:46:57.344Z">6th March 2014, 20:46:57</time>:
> I implemented a solution: http://weblogs.asp.net/ricardoperes/archive/2014/03/06/custom-nhibernate-criteria-projections.aspx.
> If anyone can confirm that it works, I may submit a pull request.
> 
---
**Jonathan Hynes** added a comment — <time datetime="2014-03-14T22:42:07.238Z" title="2014-03-14T22:42:07.238Z">14th March 2014, 22:42:07</time>:
> I tested using the `MsSql2008Dialect`. This does not appear to work because aliases are the property name and are not escaped/quoted. This can cause issues if keywords are used as column names. This could also cause problems if multiple projections have the same property name.
> 
---
**Ricardo Peres** added a comment — <time datetime="2014-03-14T22:56:53.768Z" title="2014-03-14T22:56:53.768Z">14th March 2014, 22:56:53</time>:
> What does "does not appear to work" mean? Does it return, in normal cases, what it should, or does it not?
> Not using aliases was a deliberate option, so that we can use `AliasToBean` in order to generate entities from the projection. Quoting is easy to do.
> Anyway, this was a proof of concept, I even made chances to the sample after I published that post. If I ever submit a pull request, I will address some of this issues.
> 
---
**Jonathan Hynes** added a comment — <time datetime="2014-03-14T23:11:03.263Z" title="2014-03-14T23:11:03.263Z">14th March 2014, 23:11:03</time>:
> My test is as follows:
> ```c#
> //Code to persist 20 Person objects here
> 
> var list = Session.CreateCriteria<Person>()
>               .SetFirstResult(5)
>               .SetMaxResults(10)
>               .SetProjection(new RootEntityProjection(), Projections.SqlFunction("rowcount", NHibernateUtil.Int64))
>               .List<object[]>();
> 
> Assert.Equal(10, list.Count);
>             Assert.Equal(20, (long)list<0>[1]);
>             Assert.IsType<Person>(list<0>[0]);
> ```
> The `Person` class has a `User` property. Since `User` is a SQL server keyword, I am getting the following error:
> ```
> System.Data.SqlClient.SqlException : Incorrect syntax near the keyword 'User'.
> ```
> Looking at the generated SQL, given the lack of uniquifying aliases, I also make the conjecture that the generated SQL could have alias conflicts. For instance, suppose `Person` had the following property:
> ```c#
> public virtual Person Father{get;set;}
> ```
> And I modified the query to say:
> ```c#
> .CreateAlias("Father", "f")
> .SetProjection(new RootEntityProjection(), new EntityProjection(Person, "f"))
> ```
> I conjecture that this would create duplicate aliases in the generated SQL for every property in the `Person` class.
> 
---
**Ricardo Peres** added a comment — <time datetime="2014-03-14T23:18:08.259Z" title="2014-03-14T23:18:08.259Z">14th March 2014, 23:18:08</time>:
> The problem with having aliases is that we cannot turn the results into entities. Anyway, I will post a new version that addresses your concerns.
> 
---
**Ricardo Peres** added a comment — <time datetime="2014-03-15T11:22:15.114Z" title="2014-03-15T11:22:15.114Z">15th March 2014, 11:22:15</time>:
> Jonathan, thanks for your feedback! Can I please ask you to try this new version?
> 
> 
---
**Jonathan Hynes** added a comment — <time datetime="2014-03-15T12:07:05.013Z" title="2014-03-15T12:07:05.013Z">15th March 2014, 12:07:05</time>:
> Now I am getting this error:
> ```
> System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection.
> Parameter name: index
> Result StackTrace:	
> at System.ThrowHelper.ThrowArgumentOutOfRangeException()
>    at System.SZArrayHelper.get_Item<T>(Int32 index)
>    at System.Linq.Enumerable.ElementAt<TSource>(IEnumerable`1 source, Int32 index)
>    at NHibernate.Criterion.BaseProjection.NHibernate.Criterion.IProjection.ToSqlString(ICriteria criteria, Int32 position, ICriteriaQuery criteriaQuery, IDictionary`2 enabledFilters) in d:\dev\npf\Drg.M3\Dao\Base\BaseProjection.cs:line 131
>    at NHibernate.Criterion.ProjectionList.ToSqlString(ICriteria criteria, Int32 loc, ICriteriaQuery criteriaQuery, IDictionary`2 enabledFilters)
>    at NHibernate.Loader.Criteria.CriteriaQueryTranslator.GetSelect(IDictionary`2 enabledFilters)
>    at NHibernate.Loader.Criteria.CriteriaJoinWalker..ctor(IOuterJoinLoadable persister, CriteriaQueryTranslator translator, ISessionFactoryImplementor factory, ICriteria criteria, String rootEntityName, IDictionary`2 enabledFilters)
>    at NHibernate.Loader.Criteria.CriteriaLoader..ctor(IOuterJoinLoadable persister, ISessionFactoryImplementor factory, CriteriaImpl rootCriteria, String rootEntityName, IDictionary`2 enabledFilters)
>    at NHibernate.Impl.SessionImpl.List(CriteriaImpl criteria, IList results)
>    at NHibernate.Impl.CriteriaImpl.List(IList results)
>    at NHibernate.Impl.CriteriaImpl.List<T>()
>    at Drg.M3.Tests.Unit.Core.NHibernateFacts.RootEntityProjectionWorks() in d:\dev\npf\Drg.M3.Tests.Unit\Core\NHibernateFacts.cs:line 26
> ```
> FYI, my persister is a `JoinedSubclassEntityPersister`. It's getting this error on the first property and first column name (i==0, c==0), and the property in question is a one-to-one.
> 
---
**Ricardo Peres** added a comment — <time datetime="2014-08-13T14:31:12.201Z" title="2014-08-13T14:31:12.201Z">13th August 2014, 14:31:12</time>:
> Pull request: https://github.com/nhibernate/nhibernate-core/pull/299
> 
---
**Ricardo Peres** added a comment — <time datetime="2014-08-14T10:22:41.085Z" title="2014-08-14T10:22:41.085Z">14th August 2014, 10:22:41</time>:
> New pull request: https://github.com/nhibernate/nhibernate-core/pull/302
> 
---
**Ricardo Peres** added a comment — <time datetime="2014-10-10T13:26:18.398Z" title="2014-10-10T13:26:18.398Z">10th October 2014, 13:26:18</time>:
> Fix version? Maybe 4.1?
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2016-08-12T04:40:11.455Z" title="2016-08-12T04:40:11.455Z">12th August 2016, 4:40:11</time>:
> Move unresolved improvements and new features with minor and trivial priorities to 4.2.0
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2017-09-14T01:47:17.339Z" title="2017-09-14T01:47:17.339Z">14th September 2017, 1:47:17</time>:
> Move issues from 4.2 to 5.0
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2017-09-14T01:50:03.693Z" title="2017-09-14T01:50:03.693Z">14th September 2017, 1:50:03</time>:
> Move new features to 5.1
>

bahusoid:
Fixes #948.
This PR allows to select  entities (lazy/eagerly) via Criteria API (via `Projections.Entity` and `Projections.RootEntity` helper methods).

I rebased #302 pull request +
1) Added more tests (lazy and eager version);
2) Refactored BaseEntityProjection to use `ILoadable` interface instead of `AbstractEntityPersister`;
3) Extended Projections helper class to expose lazy setting and added helper for QueryOver.

Thanks Ricardo Peres for his work!

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/GH948/EntityProjection.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH948/EntityProjection.cs
@@ -1,0 +1,170 @@
+﻿using System;
+using NHibernate.Criterion;
+using NHibernate.Engine;
+using NHibernate.SqlCommand;
+using NHibernate.Type;
+using System.Linq;
+using IQueryable = NHibernate.Persister.Entity.IQueryable;
+
+namespace NHibernate.Test.NHSpecificTest.GH948
+{
+	using TThis = EntityProjection;
+
+	[Serializable]
+	public class EntityProjection : IProjection
+	{
+		private string _entityAlias;
+		private string _columnAliasSuffix;
+		public string _tableAlias;
+		private IType[] _types;
+
+		/// <summary>
+		/// Root entity projection
+		/// </summary>
+		public EntityProjection():this(null, null)
+		{}
+		
+		public EntityProjection(System.Type rootEntity, string entityAlias)
+		{
+			RootEntity = rootEntity;
+			_entityAlias = entityAlias;
+		}
+
+		public bool FetchLazyProperties { get; set; }
+		public bool IsReadOnly { get; set; }
+		public bool Lazy { get; set; }
+
+		internal string[] IdentifierColumnAliases { get; private set; }
+		internal string[][] PropertyColumnAliases { get; private set; }
+		internal IQueryable Persister { get; private set; }
+		internal System.Type RootEntity { get; private set; }
+
+		#region Configuration methods
+
+		public TThis SetLazy(bool lazy = true)
+		{
+			Lazy = lazy;
+			return this;
+		}
+
+		public TThis SetAllPropertyFetch(bool fetchLazyProperties = true)
+		{
+			FetchLazyProperties = fetchLazyProperties;
+			return this;
+		}
+
+		public TThis SetReadonly(bool isReadOnly = true)
+		{
+			IsReadOnly = isReadOnly;
+			return this;
+		}
+
+		#endregion Configuration methods
+
+		#region IProjection implementation
+
+		public string[] Aliases
+		{
+			get;
+			private set;
+		}
+
+		bool IProjection.IsAggregate
+		{
+			get { return false; }
+		}
+
+		bool IProjection.IsGrouped
+		{
+			get { return false; }
+		}
+
+		IType[] IProjection.GetTypes(string alias, ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
+			return new[] { criteriaQuery.GetType(criteria, alias) };
+		}
+
+		IType[] IProjection.GetTypes(ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
+			SetFields(criteria, criteriaQuery);
+			return _types;
+		}
+
+		public string[] GetColumnAliases(int position, ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
+			SetFields(criteria, criteriaQuery);
+			return Aliases;
+		}
+
+		public string[] GetColumnAliases(string alias, int position, ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
+			SetFields(criteria, criteriaQuery);
+			return Aliases;
+		}
+
+		SqlString IProjection.ToSqlString(ICriteria criteria, int position, ICriteriaQuery criteriaQuery)
+		{
+			SetFields(criteria, criteriaQuery);
+
+			string identifierSelectFragment = Persister.IdentifierSelectFragment(_tableAlias, _columnAliasSuffix);
+			if (Lazy)
+				return new SqlString(identifierSelectFragment);
+			
+			return new SqlString(
+				string.Concat(
+					identifierSelectFragment,
+					Persister.PropertySelectFragment(_tableAlias, _columnAliasSuffix, FetchLazyProperties)));
+		}
+
+		SqlString IProjection.ToGroupSqlString(ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
+			throw new NotImplementedException();
+		}
+
+		TypedValue[] IProjection.GetTypedValues(ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
+			throw new NotImplementedException();
+		}
+
+		#endregion IProjection implementation
+
+		private void SetFields(ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
+			//Persister is required, so let's use it as "initialized marker"
+			if (Persister != null)
+				return;
+
+			if (RootEntity == null)
+			{
+				RootEntity = criteria.GetRootEntityTypeIfAvailable();
+			}
+
+			if (_entityAlias == null)
+			{
+				_entityAlias = criteria.Alias;
+			}
+			
+			Persister = (IQueryable) criteriaQuery.Factory.GetEntityPersister(RootEntity.FullName);
+
+			ICriteria subcriteria = criteria.GetCriteriaByAlias(_entityAlias);
+			if (subcriteria == null)
+				throw new HibernateException($"Criteria\\QueryOver alias '{_entityAlias}' for entity projection is not found.");
+
+			_tableAlias = criteriaQuery.GetSQLAlias(
+				subcriteria,
+				Persister.IdentifierPropertyName ?? string.Empty);
+			
+			_columnAliasSuffix = criteriaQuery.GetIndexForAlias().ToString();
+
+			_types = new IType[] {new EntityProjectionType(this)};
+
+			IdentifierColumnAliases = Persister.GetIdentifierAliases(_columnAliasSuffix);
+
+			PropertyColumnAliases = Lazy
+				? new string[][] { }
+				: Enumerable.Range(0, Persister.PropertyNames.Length).Select(i => Persister.GetPropertyAliases(_columnAliasSuffix, i)).ToArray();
+
+			Aliases = IdentifierColumnAliases.Concat(PropertyColumnAliases.SelectMany(alias => alias)).ToArray();
+		}
+	}
+}
--- a/src/NHibernate.Test/NHSpecificTest/GH948/EntityProjectionType.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH948/EntityProjectionType.cs
@@ -1,0 +1,98 @@
+﻿using System;
+using System.Data.Common;
+using NHibernate.Engine;
+using NHibernate.Event;
+using NHibernate.Persister.Entity;
+using NHibernate.Type;
+
+namespace NHibernate.Test.NHSpecificTest.GH948
+{
+	/// <summary>
+	/// Specialized type for retrieving entity from Criteria/QueryOver API projection.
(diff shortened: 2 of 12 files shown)


bahusoid:
Sorry. Somehow I messed up with tests in final commit.... Working on the fix..

bahusoid:
And just realized that eager  version is not really eager - criteria leads to 2 selects ID only and full entity... ( And fixing it doesn't look easy...

fredericDelaporte:
#302/#948 were about choosing the root entity to load. Does this point works as expected? (Loads fully the desired entity, not just uninitialized proxies.) If yes, better re-target this PR at just that. And maybe open new issues/pr for the eager fetching issues. (Check before if none are already existing for this other subject.)

Test failures for Postgres, Oracle and Firebird are all alike: for some reason, it seems it generates bad column names, alias names or badly combine both.

bahusoid:
@fredericDelaporte But is it really useful? 
The root of the issue described in https://nhibernate.jira.com/browse/NH-3435 based on stackoverflow question http://stackoverflow.com/questions/15744057/nhibernate-how-to-select-the-root-entity-in-a-projection.

Both JIRA issue and stackoverflow question by the same author.

And the problem he tries to solve -  implement Criteria API variant of the following HQL query:
```
session.CreateQuery("select b, rowcount() from Blog b")
              .SetFirstResult(5)
              .SetMaxResults(10)
              .List();
```
And HQL does load entities eagerly.

Also see comment in sackoverflow question:
>Yes, I know I could just use CriteriaTransform.TransformToRowCount() but that would require executing the query twice - once for the results and once for the row count. Using Futures may help a little by reducing it to one round-trip, but it's still executing the query twice on the SQL Server. For intensive queries this is unacceptable. **I want to avoid the overhead, and return the row count and the results in the same query.**

So I would say issue was created for supporting projections for fully initialized entities.

And if I understand correctly @rjperes  first [comment](https://nhibernate.jira.com/browse/NH-3435?focusedCommentId=32052&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-32052). His first implementation was exactly about eager loading. And for some reasons final proposed implementation transformed into lazy loading.

 IMHO standard implementation if exists should support eager loading . Otherwise it's not very useful.

fredericDelaporte:
Ok, so the issue (the migrated one on GitHib, #948) should be updated for better matching this. May you add a comment redefining the issue? I would then update its title and first comment.

bahusoid:
@fredericDelaporte It seems we have some kind of misunderstanding (maybe I'm using wrong terminology).

> #302/#948 were about choosing the root entity to load. Does this point works as expected? (Loads fully the desired entity, not just uninitialized proxies.)

Current implementation doesn't fully load entities (no matter root or not) - it loads uninitialized  proxies instead (only ID column retrieved from database).

fredericDelaporte:
Well, there is some mis-understanding. I would not use "eager loading" term for the case where we select an entity along with some additional data (the rowcount here). From what you write, I now understand we happen to have a bug in this PR because instead of actually selecting the entities it just returns uninitialized proxies. Then for me such a point has to be fixed in this PR, and we should not need to specify lazy false or anything like that for having the entity fully loaded by the this query.

When I see "eager loading", I am thinking about related entities, say by example having also a `rootEntity.SomeRelatedEntity` loaded along. If such a case was also needing some work, then it should likely have its own PR.

Now I have to say I am not using Criteria much, so my view may not be the right one.

bahusoid:
As I suspected retrieving initialized entities wasn't an easy fix. I made a draft with such support and there is not much left from initial Ricardo's implementation.

@fredericDelaporte , @hazzik  Could you please review specifically my latest [commit](https://github.com/nhibernate/nhibernate-core/pull/1458/commits/afb281c71d1f65433df0f16289e99a2933f6e76e). I temporarily isolated changes in test solution (see [EntityProjection.cs](https://github.com/nhibernate/nhibernate-core/pull/1458/commits/afb281c71d1f65433df0f16289e99a2933f6e76e#diff-689366c9d925cc6057310a1e0ae0227e) and [EntityProjectionType.cs](https://github.com/nhibernate/nhibernate-core/pull/1458/commits/afb281c71d1f65433df0f16289e99a2933f6e76e#diff-da9d86bb3235599a75b3aea957dc98cb). I also temporarily modified test csproj to include only related tests (new feature doesn't affect any existing functionality - and I hope it will make tests executed on server much quicker)

src/NHibernate.Test/NHSpecificTest/GH948/EntityProjection.cs
```diff
@@ -0,0 +1,171 @@
...
+			{
+				RootEntity = criteria.GetRootEntityTypeIfAvailable();
+			}
+
+			if (_entityAlias == null)
+			{
+				_entityAlias = criteria.Alias;
+			}
+			
+			Persister = (IQueryable) criteriaQuery.Factory.GetEntityPersister(RootEntity.FullName);
```
bahusoid: Is this valid cast? Any more proper way to obtain `IQueryable`

src/NHibernate.Test/NHSpecificTest/GH948/EntityProjectionType.cs
```diff
@@ -0,0 +1,92 @@
...
+				identifier,
+				_projection.PropertyAliases,
+				LockMode.None,
+				_projection.FetchLazyProperties,
+				_projection.IsReadOnly);
+
+			return entity;
+		}
+
+		private static object CreateInitializedEntity(DbDataReader rs, ISessionImplementor session, IQueryable persister, object identifier, string[][] propertyAliases, LockMode lockMode, bool fetchLazyProperties, bool readOnly)
```
bahusoid: The main concern is this method. If/when it's properly implemented I think it should be moved to some utility class. As finding a way to properly initialize entity from data reader wasn't fun.

fredericDelaporte:
I will not have time for reviewing until next week.

bahusoid:
To make things more clear I decided to prepare another pull request.

PR #1460, opened 2017-11-29
NH-2534 - Join-fetching a many-to-one with property-ref results in se…

### LINKED ISSUE #1226: NH-2534 - Join-fetching a many-to-one with property-ref results in select n+1 problem
**cremor** created an [issue](https://nhibernate.jira.com/browse/NH-2534) — <time datetime="2011-02-07T00:58:25.966Z" title="2011-02-07T00:58:25.966Z">7th February 2011, 0:58:25</time>:
> The attached test shows that a `many-to-one` association referencing an unique property (instead of the Id) results in a select n+1 problem. Although the first statement contains the correct join, all associated entities are fetched one by one after the join select. (Entities with the same value in the unique column are even fetched more than once.)
> 
> The interesting point is that this bug only occurs if the referenced entities are already in the session cache. If they are not, no additional select statements are created.
> 
---
**Keith Shannon** added a comment — <time datetime="2011-03-11T13:35:09.298Z" title="2011-03-11T13:35:09.298Z">11th March 2011, 13:35:09</time>:
> I was also able to reproduce this issue in NH 2.1.2, so it's a long-standing issue with join fetching.
> 
---
**Johan Haest** added a comment — <time datetime="2013-11-27T09:38:20.233Z" title="2013-11-27T09:38:20.233Z">27th November 2013, 9:38:20</time>:
> Any workarounds yet? Or even fixes?
> 
---
**Kamil Kliczbor** added a comment — <time datetime="2013-11-29T08:48:41.484Z" title="2013-11-29T08:48:41.484Z">29th November 2013, 8:48:41</time>:
> <~johanhaest>: we had the same issue in our project. The symptom was the same. Finally we gave up with the property-ref reference and switched to normal foreign key relationship (it was possible). And I think that solution is not so trivial.
> 
---
**Johan Haest** added a comment — <time datetime="2013-11-29T09:14:51.185Z" title="2013-11-29T09:14:51.185Z">29th November 2013, 9:14:51</time>:
> Not when your join is not on the primary key of the many to one class right?
> 
---
**Kamil Kliczbor** added a comment — <time datetime="2013-11-29T10:46:28.544Z" title="2013-11-29T10:46:28.544Z">29th November 2013, 10:46:28</time>:
> Right. We had join to other column that Id, but under certain circumstances we could abandon this approach.
>

kfedorchenko:
…lect n+1 problem  (#1226)

fixed  #1226


src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -10,8 +10,11 @@
     <TreatSpecificWarningsAsErrors />
   </PropertyGroup>
   <ItemGroup>
+    <Compile Remove="NHSpecificTest\GH01226\**" />
+    <EmbeddedResource Remove="NHSpecificTest\GH01226\**" />
```
fredericDelaporte: Hi,
I do not think the changes on the test csproj are needed. Please revert them or explain them.

kfedorchenko: Yes, it is wrong lines. I will fix them

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -56,4 +56,7 @@
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="NHSpecificTest\GH1226\" />
+  </ItemGroup>
```
fredericDelaporte: Thanks, but, this change too should not be needed. We have migrated to the new csproj format also for ceasing having changes in it just for adding a test. Is there a specific reason this test would require this change while the other tests do not?

kfedorchenko: I missed this changes in the VS file format, thanks

fredericDelaporte:
I have re-formatted the test case for them to comply with [contributing guideline](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md). Their async counterparts are lacking, but I cannot regenerate the async files because your PR use an out-of-date base, 59 commits behind NHibernate master.

Please rebase your master for having it even with current NHibernate master, then rebase your PR (and squash its four commits together by the way, amend the result with an async regen as explained in contributing).

About changes in `Loader`, I would rather avoid the almost code duplication it represents. But I still have to better analyze this issue.

src/NHibernate/Loader/Loader.cs
```diff
@@ -972,6 +972,35 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
+				{
+					int index = ((IUniqueKeyLoadable) persister).GetPropertyIndex(ukName);
+					IType type = persister.PropertyTypes[index];
+
+					object id = key.Identifier;
+					bool eagerPropertyFetch = IsEagerPropertyFetchEnabled(i);
+					ILoadable entityPersister = (ILoadable) persister;
+					string[][] cols = EntityAliases[i].SuffixedPropertyAliases;
+
+					object[] values = entityPersister.Hydrate(rs, id, obj, entityPersister, cols, eagerPropertyFetch, session);
```
fredericDelaporte: Here, there is a trouble. The referenced property may be mutable. If it has changed in DB but is still loaded in session with the old value, we are then going to associate a unique key on the new value with an entity still holding the old value. Instead of reading its value from the result set, it should be read from the entity in session (`obj`).

fredericDelaporte:
My last commit change the fix for a more proper implementation in my opinion.

kfedorchenko:
@fredericDelaporte Thanks, I will make changes corresponding your recommendations. Also I did not make changes to corresponding async method InstanceAlreadyLoadedAsync.

fredericDelaporte:
> Also I did not make changes to corresponding async method InstanceAlreadyLoadedAsync.

Just in case, I remind you this is the job of the async generator, do not make them manually. Run the async generator instead. But this requires first to update this branch.

fredericDelaporte:
My last commit causes a regression in a test, so it needs more work too.

fredericDelaporte:
I have added the fix, everything is to be squashed in a single one (or will be squashed when merging).

fredericDelaporte:
> Please rebase your master for having it even with current NHibernate master, then rebase your PR (and squash its ~~four~~ seven commits together by the way, amend the result with an async regen as explained in contributing).

Done in #1492. If you take time doing it yourself on your PR, we will merge it. Otherwise my PR will likely be merged and this one closed as replaced.

fredericDelaporte:
Hi,

Commits from master should not appear here. I guess something went wrong in the way this branch was brought up-to-date with master.

It should look like in #1492, where only the changes of this PR appear, while still having the latest commit from master.

May you try re-basing again?

Assuming your local branches are named the same than your remotes one, this should be doable with:

```
git checkout GH1226
git rebase master
```
Here you may have a bunch of conflict to solve. Solve each conflict, re-add each file which were in conflict (stage it), then continue with:
```
git rebase --continue
```
This may lead to another conflicts to resolve, in such case resolve those other ones the same way.
The conflict resolution may also cause the conflicting commit to went empty, which would then prevent continuation: in such case the commit should be dropped with:
```
git rebase --skip
```
And ultimately the rebase will end.
At that point, you should then update your remote with a forced push. A normal push will be rejected, telling you to pull before, but do not do that. When rebasing a branch having its own commits, the remote must be overridden with a forced push.
```
git push -f
```

src/NHibernate/Loader/Loader.cs
hazzik: We need to test this case.

fredericDelaporte: Done in #1492.

hazzik:
Duplicate of #1492

PR #1461, opened 2017-11-29
Disable session tracking by default

fredericDelaporte:
This enables #1391 by default (enabling a deactivation...), as proposed in #1446.

This is a grey area change for a minor: it will breaks NHProf users until they adjust their configuration for re-enabling it. (@ayende, maybe an update to NHProf could hint them about this setting.)

### DIFF
--- a/doc/reference/modules/configuration.xml
+++ b/doc/reference/modules/configuration.xml
@@ -1031,17 +1031,18 @@ var session = sessions.OpenSession(conn);
                             <literal>track_session_id</literal>
                         </entry>
                         <entry>
-                            Set whether the session id should be tracked in logs or not. When <literal>true</literal>, each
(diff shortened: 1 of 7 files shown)


hazzik:
Since the performance of SessionIdLoggingContext has become a norm now, I do not see the urge to include this into 5.1.

hazzik:
@fredericDelaporte why close?

fredericDelaporte:
Not doing it soon, not sure the PR will stay valid till we start 6.0, and this is now only a 1% gain or something like that, so well, is it worth it?

hazzik:
Ok. I just retargeted to 6.0 just in case we still will want to do so.

fredericDelaporte:
Causes a weird progress of 6.0, so going to reopen, and fix the test by the way.

hazzik:
Maybe sometime in the far future.

PR #1462, opened 2017-11-30
Fix disposing SessionIdLoggingContext if CheckAndUpdateSessionStatus is failed

hazzik:


hazzik:
This will go to 5.0.3

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -313,17 +313,26 @@ public ProcessHelper(AbstractSessionImpl session)
...
+				{
+					_context.Dispose();
+					_context = null;
+					throw;
+				}
 			}
 
 			public void Dispose()
 			{
+				_context.Dispose();
```
fredericDelaporte: If disposed is called multiple times, this will cause a null ref. We need a `?.` here.

fredericDelaporte:
Thanks for the fix. The additional point I have mentioned should be fixed in 5.1 I think.

PR #1467, opened 2017-11-30
Reduce the number of calls to UpdateTimestampsCache

gliljas:
This needs a bit of review, since it maybe just goes from wrong to less wrong. The problem is that 2nd level UpdateTimestampCache invalidation is performed twice per action. Insert 200 items, the cache is invalidated 400 times. Unnecessary for "cheap" caches, a significant performance hit for caches with latency.

This PR reduces the number from 400 to 2. It may, due to its design, perform it even on things that fail to execute, but that only has unnecessary invalidation as a side effect.

I see now that a bit of whitespace has been added (probably by R#). The wrong whitespaces?


src/NHibernate.Test/DebugSessionFactory.cs
```diff
@@ -280,7 +281,11 @@ ISession ISessionFactory.GetCurrentSession()
 
 		ITransactionFactory ISessionFactoryImplementor.TransactionFactory => ActualFactory.TransactionFactory;
 
-		UpdateTimestampsCache ISessionFactoryImplementor.UpdateTimestampsCache => ActualFactory.UpdateTimestampsCache;
+		public UpdateTimestampsCache UpdateTimestampsCache
+		{
+			get => _updateTimestampsCache ?? ActualFactory.UpdateTimestampsCache;
+			set { _updateTimestampsCache = value; }
```
hazzik: Please use consistent code style. Either both need to be in brackets or to be lambdas.

src/NHibernate.Test/DebugSessionFactory.cs
```diff
@@ -40,14 +40,15 @@ public partial class DebugSessionFactory : ISessionFactoryImplementor
 		public DebugConnectionProvider DebugConnectionProvider { get; }
 		public ISessionFactoryImplementor ActualFactory { get; }
 
-		public EventListeners EventListeners => ((SessionFactoryImpl)ActualFactory).EventListeners;
```
hazzik: Please revert whitespace changes in the file.

src/NHibernate/Async/Engine/ActionQueue.cs
```diff
@@ -24,15 +24,6 @@ namespace NHibernate.Engine
 	using System.Threading;
 	public partial class ActionQueue
 	{
-
-		public Task AddActionAsync(BulkOperationCleanupAction cleanupAction, CancellationToken cancellationToken)
```
hazzik: Please manually add back this method (into non-async part) as an async-to-sync wrapper to AddAction. And make it obsolete.

fredericDelaporte:
The whitespace changes I have seen match editorconfig, they "good" ones. But for easing reviews, global reformatting of files should be avoided. These changes normally do not occurs without an explicit call to a global reformatting action.

(That annoys me to not be able to fire a global reformat when I want to reformat the code I have just added, but well, global reformat annoys reviewers instead.)

src/NHibernate/Engine/ActionQueue.cs
```diff
@@ -169,6 +187,10 @@ public void ExecuteInserts()
 		/// </summary>
 		public void ExecuteActions()
 		{
+			if (session.Factory.Settings.IsQueryCacheEnabled)
```
hazzik: 1) `ExecuteInserts` (a few lines above) should call this as well.
2) It seems that it shall be called after execute

src/NHibernate/Engine/ActionQueue.cs
```diff
@@ -151,7 +170,6 @@ private void RegisterCleanupActions(IExecutable executable)
 			{
 				string[] spaces = executable.PropertySpaces;
 				afterTransactionProcesses.AddSpacesToInvalidate(spaces);
```
hazzik: I think it is better to utilize `afterTransactionProcesses` instead of creating an additional hashset.

src/NHibernate/Engine/ActionQueue.cs
```diff
@@ -56,6 +57,8 @@ public ActionQueue(ISessionImplementor session)
 
 			afterTransactionProcesses = new AfterTransactionCompletionProcessQueue(session);
 			beforeTransactionProcesses = new BeforeTransactionCompletionProcessQueue(session);
+
+			querySpaces = new HashSet<string>();
```
hazzik: I think it is better to utilize `afterTransactionProcesses.querySpacesToInvalidate` instead of creating an additional hashset.

hazzik: Or opposite.

hazzik: Or opposite. As you have 2 collections which hold exactly the same data.

hazzik: Also, you've forgot to clear the field.

src/NHibernate/Engine/ActionQueue.cs
hazzik: Also `Execute(action)` can be called explicitly.

src/NHibernate.Test/DebugSessionFactory.cs
gliljas: Ok!

src/NHibernate/Async/Engine/ActionQueue.cs
gliljas: It happened automatically, but I'll try what you said.

src/NHibernate/Engine/ActionQueue.cs
gliljas: Bummer. I had made a mental note of it, but forgot.

src/NHibernate/Async/Engine/ActionQueue.cs
gliljas: Unfortunately, making it obsolete didn't stop the async generator from generating calls to it, and since that is treated as an error, it prevents compilation. I fixed it manually, for the sake of the review, but is there a way to prevent this?

src/NHibernate/Engine/ActionQueue.cs
```diff
@@ -42,6 +43,7 @@ public partial class ActionQueue
 
 		private readonly AfterTransactionCompletionProcessQueue afterTransactionProcesses;
 		private readonly BeforeTransactionCompletionProcessQueue beforeTransactionProcesses;
+		private readonly HashSet<string> executedSpaces;
```
hazzik: @gliljas I'm still not sure why you need an additional collection for `executedSpaces` if you can use `AfterTransactionCompletionProcessQueue.querySpacesToInvalidate` (just make it internal) which holds exactly the same data?

gliljas: @hazzik I think that it's a bit wrong to to piggyback on a class that is not involved in the PreInvalidation. However, I have now moved them together, so that the spaces added to 
`AfterTransactionCompletionProcessQueue` comes from the same collection.

hazzik: I still don’t understand why you can not use the same collection

src/NHibernate/Engine/ActionQueue.cs
```diff
@@ -125,34 +136,58 @@ private void ExecuteActions(IList list)
...
 			session.Batcher.ExecuteBatch();
+			
+		}
+
+		private void AfterExecutions()
+		{
+			if (session.Factory.Settings.IsQueryCacheEnabled)
+			{
+				var spaces = executedSpaces.ToArray();
+				afterTransactionProcesses.AddSpacesToInvalidate(spaces);
```
hazzik: I would make the parameter of `AddSpacesToInvalidate` `IReadOnlyCollection<string>` to avoid calling `.ToArray()`. 

Or, actually, remove that method and use `executedSpaces` directly instead of `querySpacesToInvalidate` collection.

hazzik:
@gliljas I'll make a PR into your branch. It is apparently easier to code than explain:)

gliljas:
@hazzik Console.WriteLine("Please do")

src/AsyncGenerator.yml
```diff
@@ -119,6 +119,8 @@
     - name: GetFieldValue
     - name: IsDBNull
     - name: WriteLine
+    - name: AddAction
+      containingTypeName: ActionQueue
```
hazzik: @maca88 we need a way to ignore a method if it's async counterpart is marked as obsolete. This would happen a lot in the future.

hazzik:
@gliljas here is the PR: https://github.com/gliljas/nhibernate-core/pull/1

I have removed the cache invalidation from the `AfterTransactionCompletionProcessQueue` (which is not really a queue, btw). I think this way it's much clear as it's not really `AfterTransactionCompletionProcessQueue`'s job to invalidate caches.

src/AsyncGenerator.yml
maca88: I will [add an option](https://github.com/maca88/AsyncGenerator/issues/73) to make this possible in the following days, most likely tomorrow.

hazzik: Thanks @maca88

maca88: [The new option](https://github.com/maca88/AsyncGenerator/commit/d9be3b7bac99920f3014469f9effa6e6c99a3352) was added and released with the version 0.7.0. To ignore all obsolete async counterparts add the following lines inside the `analyzation` property:
```
ignoreAsyncCounterparts:
- rule: Obsolete
```
where `Obsolete` is a method rule defined as:
```
- filters:
  - hasAttributeName: ObsoleteAttribute
  name: Obsolete
```

hazzik:
There is a bug in NSubstitute, which they seems do not want to acknowledge: https://github.com/nsubstitute/NSubstitute/issues/70

hazzik:
@gliljas haha, I've made almost the same adjustments. Need to regenerate the tests thou.

gliljas:
@hazzik  Ah, crap. Will do.

hazzik:
@gliljas don't worry, I'll handle it.

src/NHibernate.Test/Async/SecondLevelCacheTest/InvalidationTests.cs
```diff
@@ -74,15 +87,25 @@ protected override void Configure(Configuration configuration)
...
-			await (cache.Received(3).InvalidateAsync(Arg.Is<object[]>(x => x.Length == 1 && (string) x[0] == "Item"), CancellationToken.None));
+			Assert.That(preInvalidations.Count,Is.EqualTo(3));
+			Assert.That(preInvalidations.All(x => x.Count == 1 && x.First() == "Item"), Is.True);
+
+			Assert.That(invalidations.Count, Is.EqualTo(3));
+			Assert.That(invalidations.All(x => x.Count == 1 && x.First() == "Item"), Is.True);
 
 		}
 
+		private bool IsRight(HashSet<string> x)
```
gliljas: Hmm...added when debugging

src/AsyncGenerator.yml
```diff
@@ -4,8 +4,6 @@
   applyChanges: true
   analyzation:
     methodConversion:
-    - conversion: Ignore
-      hasAttributeName: ObsoleteAttribute
```
hazzik: As for now we want to generate async counterparts for obsoleted methods (as they were here before the methods were marked obsolete)

src/AsyncGenerator.yml
```diff
@@ -96,6 +94,9 @@
     - conversion: Ignore
       name: Exists
       containingTypeName: AbstractCollectionPersister
+    - conversion: Ignore
+      name: QuoteTableAndColumns
+      containingTypeName: SchemaMetadataUpdater
```
hazzik: This method was obsoleted in 5.0, and did not have an async counterpart, so ignore it explicitly.

src/AsyncGenerator.yml
```diff
@@ -119,6 +120,8 @@
     - name: GetFieldValue
     - name: IsDBNull
     - name: WriteLine
+    ignoreAsyncCounterparts:
+    - rule: Obsolete
```
hazzik: New from 0.7.0. From now on we want to generate async counterparts for obsolete members, but do not want to call them. Also, we would want to add obsolete members by hands where they ceased to be generated.

fredericDelaporte:
I have relaunched the failed builds, they have got a network connectivity issue.

hazzik:
Thanks @fredericDelaporte 

I think the title of the PR needs to be changed to something more descriptive

src/NHibernate/Cache/UpdateTimestampsCache.cs
```diff
@@ -33,34 +34,50 @@ public UpdateTimestampsCache(Settings settings, IDictionary<string, string> prop
...
+		//Since v5.1
+		[Obsolete("Please use PreInvalidate(IReadOnlyCollection<string>) instead.")]
 		public void PreInvalidate(object[] spaces)
+		{
+			//Only for backwards compatibility.
+			PreInvalidate(spaces.OfType<string>().ToList());
+		}
+
+		[MethodImpl(MethodImplOptions.Synchronized)]
+		public virtual void PreInvalidate(IReadOnlyCollection<string> spaces)
```
fredericDelaporte: I do not think this class is meant to be extended. (It should likely be sealed, but it could be done only for 6.0.) Is `virtual` necessary for NSubstitute? If yes I quite dislike to have to do that for testing purpose.
Otherwise why should it be `virtual`?
(Same for the two next `virtual` methods.)

src/NHibernate.Test/TestExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+			var method = property.Body as MemberExpression;
+			var propertyInfo = typeof(T).GetProperty(method.Member.Name);
+			if (propertyInfo!= null && propertyInfo.CanWrite)
+			{
+				propertyInfo.SetValue(instance, value);
+			}
+			else
+			{
+				//camel cased field
+				var name = method.Member.Name.Substring(0, 1).ToLowerInvariant() + method.Member.Name.Substring(1);
```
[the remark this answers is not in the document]
gliljas: This was a method I just copied straight from my own at-work code, where we usually use it to set properties on NH mapped domain classes, where the setters are hidden. It uses the current form to provide chaining, so that settings many properties doesn't become a burden. The lambda is just refactoring friendly (we don't use the field version, it was added here)

I don't mind having a simpler implementation. 

> At the very least if we keep this, it should throw when not finding the field.

Absolutely. That was an omission.

src/NHibernate/Cache/UpdateTimestampsCache.cs
gliljas: I'm more in the "virtual should be the default" camp, but yes, it's for NSubstitute. It's very hard to create real unit tests in NHibernate, since everything is so coupled. I figured making these methods virtual was a small price to pay for coming a bit closer to a unit test. An alternative (which I explored) is to create mocks of the underlying ICache, but that's actually quite hard, since the provider is instantiated using reflection.

fredericDelaporte: Well, not as big as making something more visible just for test. So ok.

src/NHibernate.Test/TestExtensions.cs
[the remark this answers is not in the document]
fredericDelaporte: So maybe adjust it by changing
```c#
if (field != null)
{
	field.SetValue(instance, value);
}
```

To:
```c#
if (field == null)
    var field = typeof(T).GetField("_" + name, BindingFlags.NonPublic | BindingFlags.Instance);
if (field == null)
    throw new InvalidOperationException($"Unable to find a setter or a backing field for property {method.Member.Name}");

field.SetValue(instance, value);
```

src/NHibernate/Cache/UpdateTimestampsCache.cs
hazzik: >Well, not as big as making something more visible just for test. So ok.

@fredericDelaporte I was unsuccessfully trying to comprehend your last comment. Could you please explain?

fredericDelaporte: Switching something to virtual just for tests is not, in most cases, a big concern. (At least, far from being as big as some other "just for tests" changes we have already seen in NHibernate, and which were raising the visibility of some members.)
So let it be for this case here.

hazzik: Ok. Thanks

hazzik:
Removed TestExtensions and made UpdateTimestampsCache.Clear/ClearAsync virtual.

src/NHibernate/Cache/UpdateTimestampsCache.cs
gliljas: Thanks for making the fixes, @hazzik

PR #1469, opened 2017-12-03
Cleanup serialization

ngbrown:
This is a split-off of what was #1425.

This focuses only on the clean up of attributes and overloads.  I used some Roslyn Analyzers that I wrote based on:

* [CA2229](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2229-implement-serialization-constructors)
* [CA2235](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2235-mark-all-non-serializable-fields)
* [CA2239](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2239-provide-deserialization-methods-for-optional-fields)
* [CA2240](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2240-implement-iserializable-correctly)

There are changes to visibility in the public API.

### DIFF
--- a/src/NHibernate/PropertyNotFoundException.cs
+++ b/src/NHibernate/PropertyNotFoundException.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Runtime.Serialization;
+using NHibernate.Util;
 
 namespace NHibernate
 {
@@ -9,9 +10,12 @@ namespace NHibernate
 	[Serializable]
 	public class PropertyNotFoundException : MappingException
 	{
-		private readonly System.Type targetType;
+		private readonly string targetTypeAssemblyQualifiedName;
 		private readonly string propertyName;
 		private readonly string accessorType;
+
+		[NonSerialized]
+		private System.Type targetType;
 
 		/// <summary>
 		/// Initializes a new instance of the <see cref="PropertyNotFoundException" /> class,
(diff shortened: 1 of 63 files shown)


src/NHibernate/PropertyNotFoundException.cs
```diff
@@ -67,11 +75,28 @@ public PropertyNotFoundException(string propertyName, string fieldName, System.T
...
+			info.AddValue("propertyName", propertyName);
+			info.AddValue("accessorType", accessorType);
 		}
 
 		public System.Type TargetType
 		{
-			get { return targetType; }
+			get
+			{
+				if (targetType != null || targetTypeAssemblyQualifiedName == null) return targetType;
```
ngbrown: From #1425 review:

Why not restoring it in the deserialization constructor? (It will allow keeping it `readonly` by the way. The additional field could be avoided too.)

src/NHibernate/Util/ReflectHelper.cs
```diff
@@ -325,7 +325,7 @@ public static System.Type ClassForName(string name)
 		/// the method try to find the System.Type scanning all Assemblies of the <see cref="AppDomain.CurrentDomain"/>.
 		/// </remarks>
 		/// <exception cref="TypeLoadException">If no System.Type was found for <paramref name="classFullName"/>.</exception>
-				public static System.Type ClassForFullName(string classFullName)
+		public static System.Type ClassForFullName(string classFullName)
```
fredericDelaporte: Reformatting this file is quite an unrelated change but well, let it be. (Checked with ?w=1, no unexpected changes.)

PR #1470, opened 2017-12-03
Fix generating coalesce for Boolean expressions

### LINKED ISSUE #1468: Comparison with DateTime? produces wrong SQL
Hi,

```vb
Namespace Issues

    Namespace NA

        Public Class Entity

            Private _id As Guid
            Public Overridable Property Id() As Guid
                Get
                    Return _id
                End Get
                Set(ByVal value As Guid)
                    _id = value
                End Set
            End Property

            Private _date1 As Date?
            Public Overridable Property Date1() As Date?
                Get
                    Return _date1
                End Get
                Set(ByVal value As Date?)
                    _date1 = value
                End Set
            End Property


        End Class

    End Namespace

End Namespace
```

```vb
Using session As ISession = OpenSession()
    Dim d = New Date(2017, 12, 2)
    Dim result = (From e In session.Query(Of Entity)()
    Where e.Date1 >= d
    Select e).ToList()
    Assert.AreEqual(0, result.Count)
End Using
```

produces the following strange SQL:

```sql
[ select entity0_.Id as id1_0_, entity0_.Date1 as date2_0_
    from Entity entity0_
    where coalesce(entity0_.Date1>=@p0, @p1)=1 ]
  Name:p1 - Value:02/12/2017 00:00:00  Name:p2 - Value:False
```

Not sure if this is VB specific or whether it happens in C# as well.

Tested on NHibernate 5.0.3

hazzik:
Fixes #1468

### DIFF
--- a/src/NHibernate.Test.VisualBasic/Issues/GH1468/Fixture.vb
+++ b/src/NHibernate.Test.VisualBasic/Issues/GH1468/Fixture.vb
@@ -1,0 +1,81 @@
+﻿Imports NUnit.Framework
+
+Namespace Issues.GH1468
+    <TestFixture>
+    Public Class Fixture
+        Inherits IssueTestCase
+
+        Protected Overrides Sub OnSetUp()
+
+            Using session As ISession = OpenSession()
+
+                Using transaction As ITransaction = session.BeginTransaction()
+
+                    Dim e1 = New Entity
+                    e1.Date1 = new Date(2017, 12, 3)
+                    session.Save(e1)
+
+                    Dim e2 = New Entity
+                    e2.Date1 = new Date(2017, 12, 1)
+                    session.Save(e2)
+
+                    Dim e3 = New Entity
+                    session.Save(e3)
+
+                    session.Flush()
+                    transaction.Commit()
+
+                End Using
+
+            End Using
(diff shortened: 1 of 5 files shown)


src/NHibernate.Test.VisualBasic/Issues/GH1468/Fixture.vb
```diff
@@ -0,0 +1,81 @@
...
+
+        <Test>
+        Public Sub ShouldBeAbleToCompareNullableDateGetValueOrDefault()
+
+            Using session As ISession = OpenSession()
+                Using session.BeginTransaction()
+
+                    Dim d = New Date(2017, 12, 2)
+                    Dim result = (From e In session.Query (Of Entity)
+                            Where (e.Date1 >= d).GetValueOrDefault()
```
hazzik: This could happen to C# as well.

hazzik:
This is a simple fix, so I'm going to merge it as soon/if all builds finish.

hazzik:
> For query efficiency it would be better to have where entity0_.Date1>=@p0 instead 

Not really doable without Linq expression tree “optimization”: the VB compiler generates such a weird expression tree.

PR #1475, opened 2017-12-07
Document expiration constraint on UpdateTimestampsCache region.

fredericDelaporte:


doc/reference/modules/performance.xml
```diff
@@ -943,11 +943,20 @@ sessionFactory.EvictCollection("Eg.Cat.Kittens");]]></programlisting>
         <para>
             This setting causes the creation of two new cache regions - one holding cached query
             result sets (<literal>NHibernate.Cache.StandardQueryCache</literal>), the other
-            holding timestamps of the most recent updates to queryable tables 
-            (<literal>NHibernate.Cache.UpdateTimestampsCache</literal>). Note that the query
-            cache does not cache the state of any entities in the result set; it caches
-            only identifier values and results of value type. So the query cache should always be
-            used in conjunction with the second-level cache.
+            holding timestamps of the most recent updates to queryable tables
+            (<literal>UpdateTimestampsCache</literal>). Those region names will be prefixed by the
```
fredericDelaporte: It was documented as `NHibernate.Cache.UpdateTimestampsCache` but it is actually just `UpdateTimestampsCache` (while `NHibernate.Cache.StandardQueryCache` is "correct": inconsistency in implementation).

doc/reference/modules/performance.xml
```diff
@@ -943,11 +943,20 @@ sessionFactory.EvictCollection("Eg.Cat.Kittens");]]></programlisting>
...
+            cache region prefix if <literal>cache.region_prefix</literal> setting is configured.
+        </para>
+        <para>
+            If you use a cache provider handling an expiration for cached entries, you should set
+            the <literal>UpdateTimestampsCache</literal> region expiration to a value greater than
+            the expiration of query cache regions. (Or disable its expiration.) Otherwise the query
+            cache may yield stale data.
```
fredericDelaporte: Seen when looking at `UpdateTimestampsCache` xml comment:
```c#
	/// Tracks the timestamps of the most recent updates to particular tables. It is
	/// important that the cache timeout of the underlying cache implementation be set
	/// to a higher value than the timeouts of any of the query caches. In fact, we 
	/// recommend that the the underlying cache not be configured for expiry at all.
	/// Note, in particular, that an LRU cache expiry policy is never appropriate.
```

I thought about introducing an `UpdateTimestampsCache` expiration setting with a default big value, which would be used for overriding the default expiration setting in a properties copy used for building this cache.

But since caches are free to honor or not the default expiration setting, it would be a bit brittle. And if the user cache query for longer, he will still have to knows about this and adjust the setting. So I am just adding some documentation about it.

PR #1477, opened 2017-12-08
Reuse the same generic EmptyMapClass instance across the project

bahusoid:
This is simple "find and replace" kind of pull request to eliminate not needed instantiating of immutable EmptyMapClass instances.

Added `CollectionHelper.EmptyDictionary<TKey, TValue>()` helper mehtod that returns the same instance for given TKey and TValue and use it across the project instead of creating new instances.

### DIFF
--- a/src/NHibernate/Util/CollectionHelper.cs
+++ b/src/NHibernate/Util/CollectionHelper.cs
@@ -214,6 +214,12 @@ namespace NHibernate.Util
 
 		public static readonly IEnumerable EmptyEnumerable = new EmptyEnumerableClass();
 		public static readonly IDictionary EmptyMap = new EmptyMapClass();
+
+		public static IDictionary<TKey, TValue> EmptyDictionary<TKey, TValue>()
+		{
+			return EmptyDictionaryHolder<TKey, TValue>.Instance;
+		}
+
 		public static readonly ICollection EmptyCollection = EmptyMap;
 		// Since v5
 		[Obsolete("It has no more usages in NHibernate and will be removed in a future version.")]
@@ -455,6 +461,11 @@ namespace NHibernate.Util
 			}
 
 			#endregion
+		}
+
+		private static class EmptyDictionaryHolder<TKey, TValue>
+		{
+			public static readonly IDictionary<TKey, TValue> Instance = new EmptyMapClass<TKey, TValue>();
 		}
 
 		/// <summary>
--- a/src/NHibernate/Util/CollectionHelper.cs
+++ b/src/NHibernate/Util/CollectionHelper.cs
@@ -217,7 +217,7 @@ namespace NHibernate.Util
 
 		public static IDictionary<TKey, TValue> EmptyDictionary<TKey, TValue>()
 		{
-			return EmptyDictionaryHolder<TKey, TValue>.Instance;
+			return EmptyMapClass<TKey, TValue>.Instance;
 		}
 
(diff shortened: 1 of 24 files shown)


src/NHibernate/Util/CollectionHelper.cs
```diff
@@ -214,6 +214,12 @@ public IEnumerator GetEnumerator()
 
 		public static readonly IEnumerable EmptyEnumerable = new EmptyEnumerableClass();
 		public static readonly IDictionary EmptyMap = new EmptyMapClass();
+
+		public static IDictionary<TKey, TValue> EmptyMapFor<TKey, TValue>()
```
hazzik: Rename to `EmptyMap` or `EmptyDictionary`

bahusoid: @hazzik Done. I've also moved Instance to separate class - `EmptyDictionaryHolder<TKey, TValue>`. As `EmptyMapClass<TKey, TValue>` is public class and it's better to avoid unnecessary static initialization if it wasn't requested explicitly.

bahusoid:
I didn't touch Test  project. Let me know if it also needs to be updated (for consistency or whatever reasons)

src/NHibernate/Util/CollectionHelper.cs
```diff
@@ -214,6 +214,12 @@ public IEnumerator GetEnumerator()
 
 		public static readonly IEnumerable EmptyEnumerable = new EmptyEnumerableClass();
 		public static readonly IDictionary EmptyMap = new EmptyMapClass();
+
+		public static IDictionary<TKey, TValue> EmptyDictionary<TKey, TValue>()
+		{
+			return EmptyDictionaryHolder<TKey, TValue>.Instance;
```
fredericDelaporte: I would rather have it directly on `EmptyMapClass`. So here it would be
```c#
return EmptyMapClass<TKey, TValue>.Instance;
```

src/NHibernate/Util/CollectionHelper.cs
```diff
@@ -457,6 +463,11 @@ public object Current
 			#endregion
 		}
 
+		private static class EmptyDictionaryHolder<TKey, TValue>
+		{
+			public static readonly IDictionary<TKey, TValue> Instance = new EmptyMapClass<TKey, TValue>();
```
fredericDelaporte: I would move this into `EmptyMapClass` but changed to `internal`.

And then I would introduced an obsolete default public constructor for avoiding future code re-instanciating it directly.
So `EmptyMapClass` would start this way:

```c#
/// <summary>
/// A read-only dictionary that is always empty and permits lookup by <see langword="null" /> key.
/// </summary>
[Serializable]
public class EmptyMapClass<TKey, TValue> : IDictionary<TKey, TValue>
{
#pragma warning disable 618 // Constructor is obsolete, to be switched to non-obsolete but private.
	internal static readonly IDictionary<TKey, TValue> Instance = new EmptyMapClass<TKey, TValue>();
#pragma warning restore 618

	// Since v5.1. To be switched to private.
	[Obsolete("Please use CollectionHelper.EmptyDictionary<TKey, TValue>() instead.")]
	public EmptyMapClass() {}

	private static readonly EmptyEnumerator<TKey, TValue> emptyEnumerator = new EmptyEnumerator<TKey, TValue>();

...
```

bahusoid:
@fredericDelaporte Ok. Did as you suggested

PR #1479, opened 2017-12-08
Fix regression caused by logging refactoring.

### LINKED ISSUE #1478: Exception when using envers with the latest logging changes
We're using envers and now it fails with this error at runtime:
```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'NHibernate.Envers.Configuration.Store.AuditMappedByMetaDataAdder' threw an exception. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at NHibernate.LoggerProvider.LoggerFor(Type type)
   at NHibernate.Envers.Configuration.Store.AuditMappedByMetaDataAdder..cctor()
   at NHibernate.Envers.Configuration.Store.AuditMappedByMetaDataAdder.addToEnti
tyMeta(AuditMappedByAttribute attribute, EntityMeta entityMeta, MemberInfo membe
rInfo)
```
It's the related to the new logging refactoring. I guess it will also fails with other libraries.

```
[Obsolete("Use NHibernateLogger.For() instead.")]
public static IInternalLogger LoggerFor(System.Type type)
{
    return _legacyLoggerFactory.LoggerFor(type);
}
```

it fails because the `_legacyLoggerFactory` is null

edit: I did the changes in envers to use the new logging(`NHibernateLogger.For`') and it wokrs. But in the meantime it should work with the old version wihtout any issues.

ngbrown:
Added back the unit tests for the legacy logging provider obsoleted by #1377.

Fixes #1478.

### DIFF
--- a/src/NHibernate.Test/Logging/LoggerProviderTest.cs
+++ b/src/NHibernate.Test/Logging/LoggerProviderTest.cs
@@ -1,3 +1,4 @@
+using System;
 using NUnit.Framework;
 
 namespace NHibernate.Test.Logging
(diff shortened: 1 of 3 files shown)


hazzik:
@lnu could you please check that it fixes your problem?

lnu:
@hazzik @ngbrown  I ran nhibernate.envers tests without the pull request and all the tests failed. With the pull all the tests are green(and the logs are working). It seems to fix the issue. I'll be able to test with a real app on tuesday but if all tests are green it's ok I guess.

src/NHibernate.Test/Logging/LoggerProviderTest.cs
```diff
@@ -31,5 +44,32 @@ public void WhenConfiguredAsNullThenNoLoggingFactoryIsUsed()
...
 			Assert.That(NHibernateLogger.For("pizza").GetType().Name, Is.EqualTo("NoLoggingNHibernateLogger"));
 		}
+
+		[Test, Explicit("Changes global state."), Obsolete]
+		public void WhenConfiguredAsNullThenNoLoggingFactoryIsUsed_Obsolete()
+		{
+			NHibernateLogger.SetLoggersFactory(default(INHibernateLoggerFactory));
+
+			// NoLoggingNHibernateLogger is internal
+			Assert.That(LoggerProvider.LoggerFor("pizza"), Is.InstanceOf<NoLoggingInternalLogger>());
```
fredericDelaporte: Irrelevant comment. And next one too.

src/NHibernate/Logging.cs
```diff
@@ -52,6 +52,10 @@ public static class NHibernateLogger
 		private const string nhibernateLoggerConfKey = "nhibernate-logger";
 		private static INHibernateLoggerFactory _loggerFactory;
 
+#pragma warning disable 618
+		internal static ILoggerFactory LegacyLoggerFactory { private set; get; }
```
fredericDelaporte: A bit strange to see the setter before the getter. I would rather have it in the usual order.

PR #1480, opened 2017-12-09
Fix cache build for honoring mapped concurrency

fredericDelaporte:
This PR is a follow-up to #1476. It aims at fixing:

- The resolved concurrency strategy. Currently, when two entities share the same region but have a different concurrency strategy, the first having its cache built will "force" the other to use the same concurrency. (This trouble was not occurring for collections caches.)
- The duplication of cache concurrency strategies: currently each cached collection recreate its cache concurrency strategy without checking if one has already been built for its region. (Entities do check instead.)

Possible breaking change: `ICache` caches yielded by the session factory will be a `CacheBase` wrapper around the cache actually provided by the cache provider, if it was not deriving from `CacheBase`.


src/NHibernate/Cache/CacheFactory.cs
```diff
@@ -30,17 +30,41 @@ public static class CacheFactory
...
+		/// <param name="regionCacheGetter">The delegate for obtaining the <see cref="ICache" /> to use for the region.</param>
+		/// <returns>An <see cref="ICacheConcurrencyStrategy"/> to use for this object in the <see cref="ICache"/>.</returns>
+		public static ICacheConcurrencyStrategy CreateCache(
+			string usage,
+			string name,
+			bool mutable,
+			Settings settings,
+			Func<string, ICache> regionCacheGetter)
+		{
+			if (usage == null || !settings.IsSecondLevelCacheEnabled) return null; //no cache
```
fredericDelaporte: It uses a cache "getter" parameter instead of directly the region cache because of this test, in order to avoid creating unneeded region caches.

Otherwise the "usage == null => no cache" rule would have to be moved into the session factory, instead of staying the responsibility of cache factory.

src/NHibernate/Cache/CacheFactory.cs
```diff
@@ -59,18 +83,18 @@ public static class CacheFactory
...
-					//case CacheFactory.Transactional:
-					//	ccs = new TransactionalCache();
-					//	break;
+				//case CacheFactory.Transactional:
+				//	ccs = new TransactionalCache();
+				//	break;
 				default:
 					throw new MappingException(
-						"cache usage attribute should be read-write, read-only, nonstrict-read-write, or transactional");
+						"cache usage attribute should be read-write, read-only or nonstrict-read-write");
```
fredericDelaporte: Fixing the message by the way, transactional is not supported.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1060,10 +1082,9 @@ public void EvictQueries(string cacheRegion)
 			{
 				if (settings.IsQueryCacheEnabled)
 				{
-					IQueryCache currentQueryCache;
-					if (queryCaches.TryGetValue(cacheRegion, out currentQueryCache))
+					if (queryCaches.TryGetValue(cacheRegion, out var currentQueryCache))
 					{
-						currentQueryCache.Clear();
+						currentQueryCache.Value.Clear();
```
hazzik: I think, here we might need to check if the cache was actually created?

src/NHibernate/Cache/UpdateTimestampsCache.cs
```diff
@@ -26,6 +26,8 @@ public virtual void Clear()
 			updateTimestamps.Clear();
 		}
 
+		// Since v5.1
+		[Obsolete("Please use extension overload with an ICache parameter.")]
```
hazzik: This is a constructor, not an extension.

hazzik:
I think this refactoring went a wrong way.

src/NHibernate/Impl/SessionFactoryImpl.cs
fredericDelaporte: Here this is a `TryGetValue`, so it will not create the `Lazy`. If there is one, it means it has been created by some other function calls, which will forcibly also create the lazy value. A concurrency scenario may still causes the call here to create the lazy value before the other call (the one having created the lazy), but that is not an issue, this other call would have created the value itself anyway.

src/NHibernate/Cache/UpdateTimestampsCache.cs
fredericDelaporte: After your class interface, my turn to get caught by "too quick copy/paste"!

hazzik:
>wronger 

Didn't know such a word [exists](https://www.copyediting.com/righting-more-wrong-and-wronger/).

>Can you explain what seems wronger in this PR than current implementation in master?

I did not say that this is worse than the master, I just _feel_ that this is a wrong way, but don't yet know why. I'll take another look later this week.

hazzik:
There is an issue with the lifetime of ICache, previously, because everyone had own copy of ICache, it could kill it as they please. Now, because ICache is shared, some of the cache-holders will try to destroy already destroyed cache.

fredericDelaporte:
Ok, true.

It cannot create issue though with NHibernate or NHibernate.Caches implementations (so no failure testable), because their `Destroy` are either no op or just a call to `Clear` which should be callable as much as we want. Moreover all `Destroy` calls in NHibernate are wrapped in `try catch` logging as a warning any exception then swallowing it.

Still it may be an issue for some external cache implementation.

src/NHibernate/Cache/CacheFactory.cs
```diff
@@ -30,27 +30,50 @@ public static class CacheFactory
...
 			}
 
 			ICacheConcurrencyStrategy ccs;
 			switch (usage)
 			{
 				case ReadOnly:
-					if (mutable)
-					{
-						log.Warn("read-only cache configured for mutable: {0}", name);
-					}
```
fredericDelaporte: Moved into session factory. Only the session factory can reliably issue this warning, since a cache concurrency may be built for an entity then reused for another one having a different mutability.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -845,11 +867,9 @@ public void Close()
 
 			if (settings.IsQueryCacheEnabled)
 			{
-				queryCache.Destroy();
-
-				foreach (IQueryCache cache in queryCaches.Values)
+				foreach (var cache in queryCaches.Values)
```
fredericDelaporte: No need anymore to explicitly destroy `queryCache`, it is now into queryCaches.

fredericDelaporte:
Rebased for resolving conflicts due to this PR being based on #1476 which is now merged.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -845,8 +867,6 @@ public void Close()
 
 			if (settings.IsQueryCacheEnabled)
 			{
-				queryCache.Destroy();
-
 				foreach (var cache in queryCaches.Values)
```
fredericDelaporte: No need anymore to explicitly destroy `queryCache`, it is now into `queryCaches`.

(Comment re-added, it has been lost in last rebase.)

fredericDelaporte:
@hazzik, I believe to have addressed your concerns on this PR.

fredericDelaporte:
Rebased and amended for accounting `TestCase` changes.

(Other PR I have done have just been updated just in case, but should not require changes.)

fredericDelaporte:
Rebased for resolving conflicts with #1788.

fredericDelaporte:

hazzik:
I'm still not happy with the change.

fredericDelaporte:
In March:

> @hazzik, I believe to have addressed your concerns on this PR.

So could you be more explicit about what is still wrong? Otherwise we have no elements to either discard or improve this PR.

fredericDelaporte:
This has been rebased again.

fredericDelaporte:
There is a point this PR could handle but does not handle yet: the duplication of cache regions.

With master current state, each collections creates its own concurrency strategy, each (re)creating its cache region, no matter if it has already been created for another collection or entity. So when some collections are put in the same cache region, they all hold their own instance of the cache region, instead of sharing it like entities are doing. This discrepancy between entities and collections sounds as a bug to me. Depending on the cache implementation, it may do no differences, but it may also do some. (For example, `Clear` may apply only to the cache region instance on which it is called, or may also clear other instances of the same region, depending on how the cache is implemented.)

On the other hand, for the entity case, with master current state, they do not only share their cache region when they are cached in the same region, but also their concurrency strategy, taking the first created, even if it does not match the mapped strategy for the entity. This is clearly a bug.

This PR handles the first point by re-using concurrency strategies among collections, a bit like for entities. But it does re-use them only if they match the concurrency mapped for the collection. So it reduces the cache region duplication. But it does not eliminate it, when a same region is used for different concurrency strategies.

The second point is handled by ceasing to reuse a non-matching concurrency strategy for an entity. It now reuse only strategies matching the mapping. So now entities caching may also duplicate cache region, when a same region is used for different concurrency strategies.

We could eliminate the cache region duplication, as was doing originally this PR, by reusing cache regions for different concurrency strategies. But for correctly handling cache disposal (`Destroy`), this operation should then be no more the responsibility of the concurrency strategy, but be moved into the session factory.

Should this duplication of cache regions be completely addressed by this PR?

hazzik:
I've added some improvements and also I've removed some of the logic which I think unnecessary. You can drop it if you do not like.

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -31,20 +31,13 @@ public partial class StandardQueryCache : IQueryCache, IBatchableQueryCache
...
+				CacheFactory.BuildCacheBase(
+					settings.GetFullCacheRegionName(regionName ?? typeof(StandardQueryCache).FullName),
+					settings,
 					props))
 		{
 		}
 
-		// Since v5.2
-		[Obsolete]
-		private StandardQueryCache(UpdateTimestampsCache updateTimestampsCache, ICache cache)
```
hazzik: Introduced in 5.2, so just remove it.

fredericDelaporte:
Thanks for your changes. After having moved the destroy handling, LGTM.

src/NHibernate/Cache/StandardQueryCache.cs
fredericDelaporte: Moreover, private. But I think than previously, this constructor was used by an obsoleted exposed member. Obviously it is no more the case.

fredericDelaporte:
Rebased, and obsolete comments version updated.

PR #1482, opened 2017-12-11
Refactored DefaultEntityAliases to avoid unnecessary calculations

bahusoid:
DefaultEntityAliases is actively used for generating column aliases  for entities in T-SQL SELECT statements.

My refactorings:
1) Avoided default alias calculation if user provided alias is present;

2) Added early exits if no user provided aliases exist (to avoid lookup for all columns on empty dic);

4) Made RowIdAlias lazy (other fields are used one way or another but this one is dialect specific);

3) Removed `string.IsIntern` calls for each generated alias. I didn't find any places where aliases can be interned inside NHibernate project (and can prepare a pull request that shows that string.IsIntern for aliases always returns null for all tests). According to  b1500ea97f17609866d8e5118f3cd6895c2e2040 `string.IsIntern` was introduced instead of `string.Intern` to avoid excessive memory usage. But I don't understand why NHibernate should expect interned aliases - `string.IsIntern` doesn't look right to me here.

### DIFF
--- a/src/NHibernate/Loader/DefaultEntityAliases.cs
+++ b/src/NHibernate/Loader/DefaultEntityAliases.cs
@@ -12,13 +12,9 @@ namespace NHibernate.Loader
 	/// </summary>
 	public class DefaultEntityAliases : IEntityAliases
 	{
-		private readonly string[] suffixedKeyColumns;
-		private readonly string[] suffixedVersionColumn;
-		private readonly string[][] suffixedPropertyColumns;
-		private readonly string suffixedDiscriminatorColumn;
-		private readonly string suffix;
-		private string rowIdAlias;
-		private readonly IDictionary<string, string[]> userProvidedAliases;
+		private readonly string _suffix;
+		private string _rowIdAlias;
+		private readonly IDictionary<string, string[]> _userProvidedAliases;
 
 		public DefaultEntityAliases(ILoadable persister, string suffix)
 			: this(null, persister, suffix) {}
@@ -28,20 +24,20 @@ namespace NHibernate.Loader
 		/// </summary>
 		public DefaultEntityAliases(IDictionary<string, string[]> userProvidedAliases, ILoadable persister, string suffix)
 		{
-			this.suffix = suffix;
-			this.userProvidedAliases = userProvidedAliases?.Count > 0 ? userProvidedAliases : null;
+			_suffix = suffix;
+			_userProvidedAliases = userProvidedAliases?.Count > 0 ? userProvidedAliases : null;
 
-			suffixedKeyColumns = DetermineKeyAliases(persister, suffix);
-			suffixedPropertyColumns = GetSuffixedPropertyAliases(persister);
-			suffixedDiscriminatorColumn = DetermineDiscriminatorAlias(persister, suffix);
+			SuffixedKeyAliases = DetermineKeyAliases(persister, suffix);
+			SuffixedPropertyAliases = GetSuffixedPropertyAliases(persister);
+			SuffixedDiscriminatorAlias = DetermineDiscriminatorAlias(persister, suffix);
 
-			suffixedVersionColumn = persister.IsVersioned ? suffixedPropertyColumns[persister.VersionProperty] : null;
+			SuffixedVersionAliases = persister.IsVersioned ? SuffixedPropertyAliases[persister.VersionProperty] : null;
 			//rowIdAlias is generated on demand in property
 		}
 
 		private string[] DetermineKeyAliases(ILoadable persister, string suffix)
 		{
-			if (userProvidedAliases == null)
+			if (_userProvidedAliases == null)
 				return GetIdentifierAliases(persister, suffix);
 
 			return GetUserProvidedAliases(persister.IdentifierPropertyName)
@@ -51,7 +47,7 @@ namespace NHibernate.Loader
 
 		private string DetermineDiscriminatorAlias(ILoadable persister, string suffix)
 		{
-			if (userProvidedAliases == null)
+			if (_userProvidedAliases == null)
 				return GetDiscriminatorAlias(persister, suffix);
 
 			return GetUserProvidedAlias(AbstractEntityPersister.EntityClass)
@@ -78,7 +74,7 @@ namespace NHibernate.Loader
 
 		protected virtual string[] GetPropertyAliases(ILoadable persister, int j)
 		{
-			return persister.GetPropertyAliases(suffix, j);
+			return persister.GetPropertyAliases(_suffix, j);
 		}
 
 		private string[] GetUserProvidedAliases(string propertyPath)
@@ -87,7 +83,7 @@ namespace NHibernate.Loader
 				return null;
 
 			string[] result;
-			userProvidedAliases.TryGetValue(propertyPath, out result);
+			_userProvidedAliases.TryGetValue(propertyPath, out result);
 			return result;
 		}
 
@@ -102,7 +98,7 @@ namespace NHibernate.Loader
 		/// </summary>
 		public string[][] GetSuffixedPropertyAliases(ILoadable persister)
 		{
-			if (userProvidedAliases == null)
+			if (_userProvidedAliases == null)
 				return GetPropertiesAliases(persister);
 
 			int size = persister.PropertyNames.Length;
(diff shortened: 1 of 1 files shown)


fredericDelaporte:
About intern string, they seem to be there as a result of the Java port, see [NH-902](https://nhibernate.jira.com/browse/NH-902) comments. (That is the ticket bound to the commit you have linked.)

> About interns - do you have a special reason to use them in C#/CLR? Or is it a Java legacy?

> Yes, interns are there because they were there in Java.

> I implemented sort of a compromise - string.IsInterned will be used to find whether there is an equivalent interned string, and that one will be returned if found. If not, the original string will be used.

So initially they were always interned, not just switched to their intern version if one was there. But apparently reading from NH-902, this was causing high memory consumption due to interning to many various strings, which otherwise would have been garbage collected.

The applied solution to NH-902 does not look to me good. From a situation where too many various strings were interned we are now in a situation where maybe commonly used strings are not interned. That is not a domain where a systematic approach works. Interning should be handled on a case by case basis.

In the case of aliases, can we expect aliases to use only a small set of string values? If yes, they should be interned (reverting NH-902 for them). If no, if instead they have a large set of potential values, no interning should be handled, as your PR do.

If the situation is mixed, meaning if there is a set of common values frequently used along with a large set of various values, then NH-902 solution could have been beneficial if something had been added for interning the set of common values. But that is not the case according to your findings.

I think the safer is to remove interning handling for them. There is a suffix logic for avoiding aliases conflict which causes them to be numerous, so better not revert NH-902 for them.

Ideally it seems to me we should check all the other interning to check if they make sens.

bahusoid:
@fredericDelaporte I agree - the whole b1500ea97f17609866d8e5118f3cd6895c2e2040 looks suspicious. Interning might be useful for long living objects such as various string metadata in session factory (so commonly used property/column names and string metadata between multiple factories are shared).

But IMHO it should not be used in frequently called methods - DefaultEntityAliases instances are created for each query executions. And it's not long living objects  (it's scoped to Loader life)  - we already generated alias (memory is occupied) and obtaining interned alias  here is just a waste of CPU/memory.

fredericDelaporte:
Yes, I think removing interning for aliases is right. For other cases, we may consider opening other issues or PR.

src/NHibernate/Loader/DefaultEntityAliases.cs
```diff
@@ -15,43 +16,26 @@ public class DefaultEntityAliases : IEntityAliases
...
-			if (keyColumnsCandidates == null)
-			{
-				suffixedKeyColumns = GetUserProvidedAliases(EntityPersister.EntityID, GetIdentifierAliases(persister, suffix));
-			}
-			else
-			{
-				suffixedKeyColumns = keyColumnsCandidates;
-			}
-			Intern(suffixedKeyColumns);
+			suffixedKeyColumns = GetSuffixedKeyAliases(persister, suffix);
```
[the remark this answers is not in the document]
bahusoid: @fredericDelaporte  Done. Let me know if I miss something.

src/NHibernate/Loader/DefaultEntityAliases.cs
```diff
@@ -89,27 +100,32 @@ private string[] GetUserProvidedAlias(string propertyPath)
...
+			if (userProvidedAliases == null)
+				return GetPropertiesAliases(persister);
+
 			int size = persister.PropertyNames.Length;
 			string[][] suffixedPropertyAliases = new string[size][];
 			for (int j = 0; j < size; j++)
 			{
-				suffixedPropertyAliases[j] = GetUserProvidedAliases(persister.PropertyNames[j], GetPropertyAliases(persister, j));
-				Intern(suffixedPropertyAliases[j]);
+				suffixedPropertyAliases[j] = GetUserProvidedAliases(persister.PropertyNames[j], () => GetPropertyAliases(persister, j));
```
hazzik: The `() => GetPropertyAliases(persister, j)` part is incorrect. When the lambda is resolved `j` would always be equal to `size`. See here: https://dotnetfiddle.net/wv8xC1

bahusoid: @hazzik I don't think that your example is equal to mine. This lambda is resolved right inside cycle with closured variable. And in you examples you execute action after cycle with modified closured variable.

hazzik: Yeah, it seems that you're right.

hazzik:
@bahusoid I still would like this to be refactored without using lambdas.

bahusoid:
@hazzik What's wrong with lambdas? I can get rid of the "access to the modified closure" and still use lambdas.  I don't understand your concerns - could you please be more specific. How do you prefer to see "deferred value calculation" in some method.

hazzik:
There is no need to have “deferred” calculations. It’s all nice and clean with just simple `??`. Just need to inline some of the private methods and reassemble them into methods.

I actually did the refactoring, but wanted to give you a credit for the idea and the work.

bahusoid:
Ah.. I see :) Ok. I can do it

bahusoid:
@hazzik  Done

fredericDelaporte:
> There is no need to have “deferred” calculations. It’s all nice and clean with just simple ??. Just need to inline some of the private methods and reassemble them into other methods.

Indeed, I should have think about it.

hazzik:
I've allowed myself to cleanup the code. Please review. @bahusoid @fredericDelaporte

src/NHibernate/Loader/DefaultEntityAliases.cs
```diff
@@ -66,85 +66,40 @@ protected virtual string[] GetIdentifierAliases(ILoadable persister, string suff
...
+		private string DetermineDiscriminatorAlias(ILoadable persister)
 		{
-			string[] columns = propertyPath == null ? null : GetUserProvidedAlias(propertyPath);
-			if (columns == null)
-			{
-				return defaultAlias;
-			}
-			else
+			string[] columns = GetUserProvidedAlias(AbstractEntityPersister.EntityClass);
+			if (columns != null)
```
fredericDelaporte: The last refactoring removed by the way a discrepancy in previous refactoring causing a `null` value as first `columns` value to be replaced. So the last refactoring fix an undue change. (That said, this likely concerns an error case.)

hazzik: I don't get it.

fredericDelaporte: The previous state of the PR was somewhat equivalent to
```c#
return columns?[0] ?? GetDiscriminatorAlias(persister, _suffix);
```

While the master code was more like current state.

So on master, if `columns[0]` contains `null`, `null` is yielded as an alias (which is likely an error case). WIth your last commit it will also be the case. But without your last commit, it would have been replaced by `GetDiscriminatorAlias(persister, _suffix)`.

bahusoid:
@hazzik My IMHOs:
1) Now it's really hard to understand what's been changed in this PR. I personally think that fields  renaming  and method orderings should be commited as separate changeset without others changes.

2) @fredericDelaporte 
>and those shortcuts were only quite small savings
Maybe. But the thing is - I believe that user provided aliases are hardly ever used by most users. So those shortcuts are the main route of code for most users - why not to save a little? See example of test with custom provided aliases - `NativeSqlCollectionLoaderFixture.LoadCompositeElementsWithWithCustomAliases`

src/NHibernate/Loader/DefaultEntityAliases.cs
hazzik: I believe that this is a correct behaviour: if we got null in `columns[0]` we should not swallow this case, but fail somewhere.

hazzik:
@bahusoid do you want me to redo this in 2 commits? I don't mind.

bahusoid:
@hazzik Yeah it would be better. And I want my shortcuts back :)

hazzik:
071b7449-84d5-4f26-98de-a0be3bb77a1f

src/NHibernate/Loader/DefaultEntityAliases.cs
fredericDelaporte: Yes, agreed.

hazzik:
@bahusoid done.

bahusoid:
@hazzik Have nothing to complain about.

Though... )) Do we really need both SafeGetUserProvidedAliases/GetUserProvidedAliases? "Unsafe" version is used only twice and only if user aliases are provided. I think code would be cleaner with only safe version.

Just a thought - it's fine by me to keep it.

fredericDelaporte:
> Though... )) Do we really need both SafeGetUserProvidedAliases/GetUserProvidedAliases? "Unsafe" version is used only twice and only if user aliases are provided. I think code would be cleaner with only safe version.

This is always a tricky part with shortcuts/small code optimizations: up to which point should we do them? That is why I try to not debate much on them, unless some benchmark show concrete effects.

About the merge, I was thinking the point of splitting the commits was not only for review, but also for helping understanding the changes in the history. So merging them all squashed to a single commit was not was I was expecting.
Of course it would have required to have this PR commits squashed in a minimal set of commits doing only the final changes but still split between "code renamed, restyled and moved around" and "code logic changed".

src/NHibernate/Loader/DefaultEntityAliases.cs
```diff
@@ -10,48 +9,75 @@ namespace NHibernate.Loader
...
+		public string[] SuffixedVersionAliases { get; }
+
+		public string[][] SuffixedPropertyAliases { get; }
+
+		public string SuffixedDiscriminatorAlias { get; }
+
+		public string[] SuffixedKeyAliases { get; }
+
+		// TODO: not visible to the user!
+		public string RowIdAlias => _rowIdAlias ?? (_rowIdAlias = Loadable.RowIdAlias + _suffix);
```
fredericDelaporte: I think none of us have realized we were changing a class which instances can be used by concurrent threads. In a concurrent usage situation, this micro-optimization can instead be a micro-performance loss! We would have better not have done it in my opinion.

PR #1483, opened 2017-12-12
Clean-up of TypeFactory

fredericDelaporte:
* Removing undue global thread sync and concurrency checks, the concurrent dictionary handles already that.
 * Removing useless singleton pattern, likely a remnant of Java port.

Many static methods of `TypeFactory` and `SqlTypeFactory` for obtaining types were marked with `[MethodImpl(MethodImplOptions.Synchronized)]`, causing a global lock of the type. This was done for protecting the cache of types from concurrent writes. But those methods are since some times implemented with a concurrent dictionary, and are no more needing such a strict synchronization, bad for performances.

This change may allow duplicated types to be build when they are not already cached, but only one will be yielded and cached. Those types are lightweight objects without critical resources, having sometime built a dup will not cause troubles.


hazzik:
> it's really hard to understand what's been changed in this PR. I personally think that fields renaming and method orderings should be commited as separate changeset without others changes.

fredericDelaporte:
> fields renaming and method orderings

There is no such things in `TypeFactory.cs` changes. Three methods have been suppressed because they were useless (just enforcing a bad way of using a concurrent dictionary), the others are just rewritten to correctly use the concurrent dictionary.

If that is the changes in `MultiThreadRunner` which cause issue to you, I can just drop its separated commit, going back to the previous situation: no actual test of parallel usages.

hazzik:
I just cited the neighboring PR ;)

fredericDelaporte:
Yes I know it comes from bahusoid, but does it really apply here? I could remove some syntax update and whitespace fixes, but their are not numerous and not hard to identify.

PR #1484, opened 2017-12-12
Fix default types

fredericDelaporte:
* Remove logic duplication for getting default NHibernate type for a CLR type.
 * Test all defaults for CLR type having many possibilities.
 * Clean-up.

The test for `decimal` was failing, it was yielding `Currency` instead of `Decimal`.


hazzik:
>The test for decimal was failing, it was yielding Currency instead of Decimal.

I think there was an issue about something like that.

hazzik:
It was mentioned in #1052

fredericDelaporte:
> It was mentioned in #1052

Ok, I will add a comment there about this point being resolved once this PR gets merged.

PR #1485, opened 2017-12-13
Add timeouts support to MultiCriteria

### LINKED ISSUE #1403: Add timeouts support to MultiCriteria
From #1303

hazzik:
Fixes #1403

---

The `SetTimeouts` are not tested anywhere in the code. And I'm not sure how we can test it. So, no tests have been added.

### DIFF
--- a/src/NHibernate/Impl/MultiCriteriaImpl.cs
+++ b/src/NHibernate/Impl/MultiCriteriaImpl.cs
@@ -35,6 +35,7 @@ namespace NHibernate.Impl
 		private string cacheRegion;
 		private IResultTransformer resultTransformer;
 		private readonly IResultSetsCommand resultSetsCommand;
+		private int? _timeout;
 
 		/// <summary>
 		/// Initializes a new instance of the <see cref="MultiCriteriaImpl"/> class.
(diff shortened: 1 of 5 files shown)


src/NHibernate/Impl/MultiCriteriaImpl.cs
```diff
@@ -35,6 +35,7 @@ public partial class MultiCriteriaImpl : IMultiCriteria
 		private string cacheRegion;
 		private IResultTransformer resultTransformer;
 		private readonly IResultSetsCommand resultSetsCommand;
+		private int _timeout;
```
fredericDelaporte: For consistency with `MultiQueryImpl`, it should be `int?`.
For correctness, it should be nullable too, because `0` means "no timeout" (it is only a [recommendation](https://docs.microsoft.com/en-us/dotnet/api/system.data.common.dbcommand.commandtimeout?view=netframework-4.7.1#System_Data_Common_DbCommand_CommandTimeout) but likely most data provider will respect this), thus this added feature is going to disable the default timeout.

src/NHibernate/Impl/MultiQueryImpl.cs
```diff
@@ -526,7 +526,7 @@ protected List<object> DoList()
 
 			try
 			{
-				using (var reader = resultSetsCommand.GetReader(commandTimeout != RowSelection.NoValue ? commandTimeout : (int?)null))
+				using (var reader = resultSetsCommand.GetReader(_timeout))
```
fredericDelaporte: Breaking change here: users actively setting the timeout with `NoValue` (or `-1`) will have they code switched from a "use default timeout" behavior to "throw an `ArgumentException`". And so, for consistency, it would be better to put the `NoValue` logic into `MultiCriteriaImpl` too.

hazzik:
Fixed.

PR #1489, opened 2017-12-14
SecondLevel QueryCache causes invalid cast

### LINKED ISSUE #1385: SecondLevelCache CreateSQLQuery().UniqueResult<int>() throws Exception Specified cast is not valid.

fredericDelaporte:
* Fix #1385

As said in #1385, I am not sure we should really try fixing this. Maybe we should instead detect the case (detect attempt to cache a query requiring types auto-discovery) and yield a proper error, instead of letting it fails with an invalid cast.


fredericDelaporte:
> Why not use [Theory] instead of ignoring bunch of tests?

You mean, by example changing:
```c#
[Test]
public void LoadElementCollectionWithCustomLoader()
{
	if (WithQueryCache)
	{
		Assert.Ignore("This test does not use a cacheable query.");
	}
	string[] routes = CreateRoutes();
	Country country = CreateCountry(routes);
	Save(country);
	{
		var c = session.Get<Country>(country.Code);
		Assert.That(c, Is.Not.Null, "country");
		Assert.That(c.Routes, Is.EquivalentTo(routes), "country.Routes");
	}
}
```
To:
```c#
[Theory]
public void LoadElementCollectionWithCustomLoader()
{
	Assume.That(WithQueryCache, Is.False, "This test does not use a cacheable query.");
	string[] routes = CreateRoutes();
	Country country = CreateCountry(routes);
	Save(country);
	{
		var c = session.Get<Country>(country.Code);
		Assert.That(c, Is.Not.Null, "country");
		Assert.That(c.Routes, Is.EquivalentTo(routes), "country.Routes");
	}
}
```
Why not.

hazzik:
No, this is not what I meant.

hazzik:
I meant using `[Theory]` instead of `WithQueryCache` property

hazzik:
Ah, I did not see that `WithQueryCache` is used in `SetUp`.

fredericDelaporte:
> UPD2: Ah, I did not see that `WithQueryCache` is used in `SetUp`.

It could be removed from there since the query cache is activated in our current test configuration, and since it does not really need to be de-activated because it is also handled on a query per query basis.

But then I would have to do a bit more changes for tests using `LoadCountryWithNativeSQL`. That was for minimizing changes to those tests that I have chosen using an overridable property.

Still switching from `Ignore` to `Assume` is less code noisy.

hazzik:
But then you don't need to use `[Theory]` flag. It should work with `[Test]`, shouldn't it?

fredericDelaporte:
`Assume` is meant to be used with `Theory`, not sure `Test` accepts it. Going to check.

src/NHibernate/Transform/CacheableResultTransformer.cs
```diff
@@ -152,13 +172,24 @@ private CacheableResultTransformer(bool[] includeInTuple, bool[] includeInTransf
...
 		public object TransformTuple(object[] tuple, string[] aliases)
 		{
-			if (aliases != null && aliases.Length != _tupleLength)
+			if (_includeInTuple == null)
+				throw new InvalidOperationException("This transformer is not initialized");
+			if (aliases != null && aliases.Length != _includeInTuple.Length)
 			{
 				throw new InvalidOperationException(
-					"aliases expected length is " + _tupleLength +
+					"aliases expected length is " + _includeInTuple.Length +
```
fredericDelaporte: This change is questioning. Tests succeeds, but why should it be so? Moreover it appears here that `aliases` are only used for causing this failure. The actual implementation does not need them. Maybe this check should be removed instead.

src/NHibernate/Transform/CacheableResultTransformer.cs
```diff
@@ -152,22 +172,23 @@ private CacheableResultTransformer(bool[] includeInTuple, bool[] includeInTransf
...
 		{
-			if (aliases != null && aliases.Length != _tupleLength)
-			{
-				throw new InvalidOperationException(
-					"aliases expected length is " + _tupleLength +
-					"; actual length is " + aliases.Length);
-			}
-			// really more correct to pass index( aliases.getClass(), aliases )
-			// as the 2nd arg to the following statement;
-			// passing null instead because it ends up being ignored.
```
fredericDelaporte: Test and comment on `aliases` here were quite obsolete, and the test was causing the auto-discovered cached query to fail. `aliases` are not used by the actual implementation, and the suggested one use a method which seems suppressed.

fredericDelaporte:
@gliljas, @hazzik, do you think #1385 (fixed by this PR) should be in 5.1?

fredericDelaporte:
Rebased.

fredericDelaporte:
There are some people seemingly strongly awaiting this fix (see #1385), so for them it would be great to have it in 5.1.

Now, this fix is still a bit hackish in my view, although I do not think it may creates new issues.

hazzik:
This goes to 5.1 and everything else are going to next release.

PR #1490, opened 2017-12-15
Optimize empty arrays usages

hazzik:


hazzik:
Yeah, I've forgot about `new $T$[] {}`

fredericDelaporte:
I have had a look with Frans's benchmark, no changes, same results with or without commits of this PR.

PR #1492, opened 2017-12-16
Join-fetching with property-ref results in select n+1

### LINKED ISSUE #1226: NH-2534 - Join-fetching a many-to-one with property-ref results in select n+1 problem

fredericDelaporte:
* Fixes #1226 (NH-2534)

This is a rebase of #1460, with async regen, squashed, as it was asked [here](https://github.com/nhibernate/nhibernate-core/pull/1460#issuecomment-350706536).


src/NHibernate/Loader/Loader.cs
```diff
@@ -972,6 +972,29 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
 					entry.LockMode = lockMode;
 				}
 			}
+
+			// #1226: If it is already loaded and can be loaded from an association with a property ref, make
```
hazzik: Can this please be extracted into an own method?

hazzik:
@fredericDelaporte why not do it in the original PR? It seems to have "changes from contributors" enabled.

fredericDelaporte:
> It seems to have "changes from contributors" enabled.

Yes but I have seen on some other PR that some people have difficulties handling forced pushes, so unless the contributor has already shown by its action on his PR that it is not a trouble for him, I now avoid doing forced pushes on their PR.

src/NHibernate/Loader/Loader.cs
fredericDelaporte: Done.

gliljas:
@fredericDelaporte Yes, force pushing on someone else's PR should only be done after communicating with the owner about it.

hazzik:
@fredericDelaporte can the CacheByUniqueKey logic be unifed, eg. called this way:

```
//If the object is already loaded, return the loaded one
obj = session.GetEntityUsingInterceptor(key);
if (obj != null)
{
	//its already loaded so dont need to hydrate it
	InstanceAlreadyLoaded(rs, i, persisters[i], key, obj, lockModes[i], session);
}
else
{
	obj = InstanceNotYetLoaded(rs, i, persisters[i], key, lockModes[i], descriptors[i].RowIdAlias, optionalObjectKey, optionalObject, hydratedObjects, session);
}
CacheByUniqueKey(i, persisters[i], obj, session);
```

And remove this logic from `LoadFromResultSet`?

fredericDelaporte:
> can the CacheByUniqueKey logic be unifed

No, there is a trap here: in the `InstanceNotYetLoaded` case, `obj` is not actually hydrated but just put in a "to be hydrated" list, with its `values` array already put in persistence context. So getting its unique key value from its properties as do `CacheByUniqueKey` will fail. We may try to get `values` back from the persistence context in such case, but this would cause a more complex code and also more redundant unique key caching when the entity appears in many rows.

hazzik:
I've discovered this "trap". 

this code works:

```csharp
obj = session.GetEntityUsingInterceptor(key);
var persister = persisters[i];
object[] values;
if (obj != null)
{
	//its already loaded so dont need to hydrate it
	InstanceAlreadyLoaded(rs, i, persister, key, obj, lockModes[i], session);
	values = persister.GetPropertyValues(obj);
	CacheByUniqueKey(i, persister, obj, session, values);
}
else
{
	obj = InstanceNotYetLoaded(rs, i, persister, key, lockModes[i], descriptors[i].RowIdAlias, optionalObjectKey, optionalObject, hydratedObjects, session);
	var entry = session.PersistenceContext.GetEntry(obj);
	values = entry.LoadedState;
}
CacheByUniqueKey(i, persister, obj, session, values);
```

src/NHibernate/Loader/Loader.cs
```diff
@@ -972,6 +972,34 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
 			}
+
+			CacheByUniqueKey(i, persister, obj, session);
+		}
+
+		private void CacheByUniqueKey(int i, IEntityPersister persister, object obj, ISessionImplementor session)
+		{
+			// #1226: If it is already loaded and can be loaded from an association with a property ref, make
+			// sure it is also cached by its unique key.
+			var ukName = OwnerAssociationTypes?[i]?.RHSUniqueKeyPropertyName;
```
hazzik: can we unfold `?` operator here? It looks unreadable.

src/NHibernate.Test/NHSpecificTest/GH1226/Fixture.cs
```diff
@@ -0,0 +1,84 @@
...
+		protected override void OnTearDown()
+		{
+			base.OnTearDown();
+
+			using (var session = OpenSession())
+			{
+				using (var tx = session.BeginTransaction())
+				{
+					session.Delete("from Account");
+					session.Delete("from Bank");
```
hazzik: `CreateQuery("delete ...").ExecuteUpdate()`?

src/NHibernate.Test/NHSpecificTest/GH1226/Fixture.cs
```diff
@@ -0,0 +1,84 @@
...
+			}
+		}
+
+		protected override void OnTearDown()
+		{
+			base.OnTearDown();
+
+			using (var session = OpenSession())
+			{
+				using (var tx = session.BeginTransaction())
```
hazzik: please align nested usings:

```
using(...)
using(...)
[
}
```

src/NHibernate.Test/Async/NHSpecificTest/GH1226/Fixture.cs
```diff
@@ -0,0 +1,95 @@
...
+	[TestFixture]
+	public class FixtureAsync : BugTestCase
+	{
+		protected override void OnSetUp()
+		{
+			base.OnSetUp();
+
+			using (var session = OpenSession())
+			{
+				using (var tx = session.BeginTransaction())
```
hazzik: Same

src/NHibernate/Loader/Loader.cs
```diff
@@ -972,6 +972,34 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
+			//    is a second case, just test for ukValue null.
+			//  - Entity association is unset in session but not yet persisted, autoflush disabled: ignore. We are
+			//    already in an error case: querying entities changed in session without flushing them before querying.
+			//    So here it gets loaded as if it were still associated, but we do not have the key anymore in session:
+			//    we cannot cache it, so long for the additionnal round-trip this will cause. (Do not fallback on
+			//    reading the key in rs, this is stale data in regard to the session state.)
+			if (ukValue == null)
+				return;
+			var type = persister.PropertyTypes[index];
+			var euk = new EntityUniqueKey(persister.EntityName, ukName, ukValue, type, session.Factory);
```
hazzik: LoadFromRow's version caches by rootPersister.EntityName. Please add a test where the referenced entity is a hierarchical object.

src/NHibernate/Loader/Loader.cs
fredericDelaporte: Just null propagation all the way. Original code was:
```c#
IAssociationType[] ownerAssociationTypes = OwnerAssociationTypes;
if (ownerAssociationTypes != null && ownerAssociationTypes[i] != null)
{
	string ukName = ownerAssociationTypes[i].RHSUniqueKeyPropertyName;
```

Would you rather have:
```c#
var ownerAssociationTypes = OwnerAssociationTypes;
if (ownerAssociationTypes == null)
	return;
var ukName = ownerAssociationTypes[i]?.RHSUniqueKeyPropertyName;
```

src/NHibernate/Loader/Loader.cs
fredericDelaporte: Here `persister` is `rootPersister`. Non-root `persister` is a thing existing only inside `LoadFromResultSet`.

fredericDelaporte:
> this code works:

But it causes, in the already loaded case, to get all the entity properties from its tuplizer even if there is no unique key, which is quite heavier than getting only one property when there is an unique key. I would rather pass `null` as `values` for this case, and fallback on the single property.
Well I consider instead passing an `alreadyLoaded` boolean and handle that inside the method.

src/NHibernate/Loader/Loader.cs
hazzik: Ok

src/NHibernate.Test/NHSpecificTest/GH1226/Fixture.cs
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		[Test]
+		public void BankShouldBeJoinFetched()
+		{
+			using (var session = OpenSession())
+			using (var tx = session.BeginTransaction())
+			{
+				// Bug only occurs if the Banks are already in the session cache.
+				session.CreateQuery("from Bank").List<Bank>();
```
hazzik: Since the old code has changed, could this case please be checked as well?

src/NHibernate.Test/NHSpecificTest/GH1226/Fixture.cs
```diff
@@ -0,0 +1,129 @@
...
+			Sfi.Statistics.IsStatisticsEnabled = true;
+		}
+
+		[Test]
+		public void BankShouldBeJoinFetched()
+		{
+			using (var session = OpenSession())
+			using (var tx = session.BeginTransaction())
+			{
+				// Bug only occurs if the Banks are already in the session cache.
```
hazzik: Since the old code has changed, could this case please be checked as well?

fredericDelaporte: You mean, checking eager fetching works for `Bank` not already in session too? It is likely this is already tested somewhere else, but why not.

fredericDelaporte: Done.

PR #1493, opened 2017-12-18
Fix referencing Oracle ManagedDataAccess native binaries

hazzik:


hazzik:
Stupid thing. Going to roll-back the DB on the build agent.

PR #1494, opened 2017-12-18
Update to Oracle installation instructions.

fredericDelaporte:
Here is a somewhat minimal update for easing setting up Oracle. Feel free to amend it if you know better, as I have not actually checked with a 12.2 version for the server installation part.

### DIFF
--- a/lib/teamcity/oracle/oracle_installation.txt
+++ b/lib/teamcity/oracle/oracle_installation.txt
@@ -1,18 +1,44 @@
 
-Installation steps for Oracle (Express Edition) for NH TeamCity:
-
-1.  Download Oracle Database 10g Express Edition (OracleXE.exe):  http://www.oracle.com/technetwork/database/express-edition/downloads/102xewinsoft-090667.html;
+Installation steps for Oracle for NH TeamCity:
+1.  Download Oracle Database 12c from https://oracle.com/. Please make sure you comply with it's license.
+    12.2 will fail some tests (mainly NH-1171) due to a bug with comments in queries, see
+    https://stackoverflow.com/q/7493028/1178314. So take care of installing 12.1 instead.
         (Note: you'll need to sign-up for a (free) Oracle account to download)
 
-2.  Run the installer ... choose any HTTP port for the listener if asked (it may automatically choose the default 8080 if available);
-3.  Enter 'password' as the password for the SYS and SYSTEM accounts;
-4.  The setup should install Oracle (XE) on the machine;
-5.  Once the installation is complete, leave the option ticked to open the homepage for configuration.
+2.  Run the installer ...
+3.  Choose any HTTP port for the listener if asked (it may automatically choose the default 8080 if available);
+4.  Choose the ASCII character set. You will have to ensure nhibernate.oracle.use_n_prefixed_types_for_unicode
+    parameter is set to true; Choosing instead the Unicode character set will cause some NHibernate tests to fail.
+5.  Enter 'password' as the password for the SYS and SYSTEM accounts;
+6.  The setup should install Oracle on the machine.
 
 Creating the NH Schema:
 
-a.  Login to the homepage using login/password SYSTEM/password;
-b.  Select Administration->Database Users->Create User;
-c.  Enter Username=nhibernate, Password=nhibernate, and select the DBA role;
-d.  Click 'Create' to create the user/schema.
+a.  Run SQL Plus:
+        SQL*Plus: Release XXX Production on YYY
+        Copyright (c) 1982, 2014, Oracle.  All rights reserved.
+        Enter user-name:
+b.  Enter:
+        / as SYSDBA
+    Expected result:
+        Connected to:
+        Oracle Database XXX
+        SQL>
+c.  Enter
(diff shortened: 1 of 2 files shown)


lib/teamcity/oracle/oracle_installation.txt
```diff
@@ -1,12 +1,15 @@
 
-Installation steps for Oracle (Express Edition) for NH TeamCity:
+Installation steps for Oracle for NH TeamCity:
 
-1.  Download Oracle Database 10g Express Edition (OracleXE.exe):  http://www.oracle.com/technetwork/database/express-edition/downloads/102xewinsoft-090667.html;
+1.  Download Oracle Database 12c (12.2):  http://www.oracle.com/technetwork/database/enterprise-edition/downloads/index.html
```
hazzik: 12.2 does not work:
1. 12.2 ManagedDataAccess fails with this stupid bind by name + comments bug (about 8 tests are failing).
2. 12.1 (the one which does not fail tests) ManagedDataAccess driver does not work with 12.2 because of "protocol violations".

lib/teamcity/oracle/oracle_installation.txt
```diff
@@ -16,3 +19,11 @@ b.  Select Administration->Database Users->Create User;
 c.  Enter Username=nhibernate, Password=nhibernate, and select the DBA role;
 d.  Click 'Create' to create the user/schema.
 
+Allow usage of the managed driver NuGet package:
+
+1.  Within a command line, navigate to {Oracle home}\product\{version}\client_64\ODP.NET\managed\x64
```
hazzik: This is required only for 12.1 and below (12.2 does not install these into GAC)

lib/teamcity/oracle/oracle_installation.txt
```diff
@@ -1,12 +1,15 @@
...
+    does not succeed all the NHibernate tests.
         (Note: you'll need to sign-up for a (free) Oracle account to download)
 
 2.  Run the installer ... choose any HTTP port for the listener if asked (it may automatically choose the default 8080 if available);
 3.  Enter 'password' as the password for the SYS and SYSTEM accounts;
-4.  The setup should install Oracle (XE) on the machine;
+4.  The setup should install Oracle on the machine;
 5.  Once the installation is complete, leave the option ticked to open the homepage for configuration.
 
 Creating the NH Schema:
```
hazzik: There is no application express, so need to do with SQL+:
```shell
SQL*Plus: Release XXX Production on YYY
Copyright (c) 1982, 2014, Oracle.  All rights reserved.
Enter user-name: / as SYSDBA
Connected to:
Oracle Database XXX
SQL> create user nhibernate identified by nhibernate;
User created.
SQL> grant dba to nhibernate;
Grant succeeded.
SQL>
```

hazzik: If there is ORA-65096. It means that the DB is installed as a container DB. The container DB wants user names to start with `c##`.  To enable creating users without `c##` need to run `alter session set "_ORACLE_SCRIPT"=true;` to allow creating users without c##.  More info [here](http://www.dba-oracle.com/t_ora_65096_create_user_12c_without_c_prefix.htm).

hazzik:
Also, it is probably required to run the setup in the advanced mode to be able to select max memory consumption, but I don't think it worth mentioning.

lib/teamcity/oracle/oracle_installation.txt
fredericDelaporte: There is still a Database Express (at least with enterprise edition, url https://localhost:5500/em/shell#/security/show_users using an non-trusted certificate) which allows to handle this, but the SQL plus solution seems more widely available, so thanks for the instructions.

fredericDelaporte:
Thanks for all the details Alexander, I have added a commit for taking them into account.

src/NHibernate.Config.Templates/Oracle-Managed.cfg.xml
```diff
@@ -9,10 +9,10 @@ for your own use before compile tests in VisualStudio.
 	<session-factory name="NHibernate.Test">
 		<property name="connection.driver_class">NHibernate.Driver.OracleManagedDataClientDriver</property>
 		<property name="connection.connection_string">
-			User ID=nhibernate;Password=nhibernate;Data Source=(DESCRIPTION = (ADDRESS = (PROTOCOL = TCP)(HOST = localhost)(PORT = 1521)) (CONNECT_DATA = (SERVER = DEDICATED) (SERVICE_NAME = XE)))
+			User ID=nhibernate;Password=nhibernate;Data Source=(DESCRIPTION = (ADDRESS = (PROTOCOL = TCP)(HOST = localhost)(PORT = 1521)) (CONNECT_DATA = (SERVER = DEDICATED) (SERVICE_NAME = ORCL)))
```
fredericDelaporte: `ORCL` seems to be the default service name for 12c. `XE` is very likely the default name when using Express edition, which currently can at best be a 11g. Since instructions are now about installing a 12c, better switch to its default name. (But on the teamcity server, we are still using the `XE` name it seems.)

Issue #1496, opened 2017-12-20
Fix ManyToOneType.IsModified to handle both object instance and identifier passed to the parameter “old”.


mishaurubkov:
# PROBLEM:
ManyToOneType.IsModified won't resolve identifier if object instance (not the identifier) was passed to the “old” parameter.
Similar issue https://hibernate.atlassian.net/browse/HHH-3730
 was reported in Hibernate with pull request https://github.com/hibernate/hibernate-orm/pull/728 

# DESCRIPTION:
The current code of IsModified method accepts 4 parameters:

`public override bool IsModified(object old, object current, bool[] checkable, ISessionImplementor session)`

It assumes that “old” is always Identifier and “current” is the object instance, so it tries to resolve Identifier for the “current” only: 

`return “GetIdentifierOrUniqueKeyType(session.Factory).IsDirty(old, GetIdentifier(current, session), session)”;`

SCENARIO 1: SelectBeforeUpdate

Typically, the method ManyToOneType.IsModified(…) is fired when flushing a session with an entity which contains a "many-to-one" property, but only when the "select-before-update" option is turned on for that class (and performed). For this scenario the identifier is always passed to “old” and it works fine.

SCENARIO 2: Event Listeners
It became common practice to do auditing with NHibernate Event Listeners and call Persister.FindModified() method: 

```
public class AuditEventListener : IPostUpdateEventListener
{
public void OnPostUpdate(PostUpdateEvent @event)
	{
		var modifiedItems = @event.Persister.FindModified(@event.OldState, @event.State, @event.Entity, @event.Session);
		// Process modified items.	
	}
}
```

For this scenario, @event.OldState and @event.State hold the object instances. It ends up going through calls: AbstractEntityPersister.FindModified(..) -> TypeHelper.FindModified(..) -> ManyToOneType. IsModified(..)
Eventually the object instances make into “old” parameter in ManyToOneType.IsModified(..). 
Once it happens, we have two problems:
1.	ManyToOneType.IsModified(..) returns **always true** comparing two ManyToOneType entities with regular (not composite) Ids.
2.	ManyToOneType.IsModified(..) throws **InvalidCastException** exception comparing two ManyToOneType entities with composite Ids

# SOLUTION:
We have to handle 4 possible scenarios for “old” 
1.	Id
2.	Composite Id
3.	Object Instance with Id
4.	Object Instance with Composite Id

A.	Leaving “old” as it is fails for (3) and (4)
[...]

PR #1497, opened 2017-12-21
#1496 Fix ManyToOneType.IsModified to handle both object instance and identifier passed to the parameter “old”. 

mishaurubkov:
Please see unit tests and fix for #1496 in commit:
https://github.com/nhibernate/nhibernate-core/commit/f4e3055052279a45fd0f6d03ca9391a891e83ef3 
I believe ManyToOneType.IsModifiedAsync(...) could be fixed same way but would like to hear the feedback on the issue and proposed solution first.

### DIFF
--- a/src/NHibernate.Test/NHibernateUtilTest.cs
+++ b/src/NHibernate.Test/NHibernateUtilTest.cs
@@ -10,36 +10,56 @@ namespace NHibernate.Test
 		[Test]
 		public void CanGuessTypeOfInt32ByValue()
 		{
-			Assert.AreEqual(NHibernateUtil.Int32, NHibernateUtil.GuessType(15));
+			Assert.That(NHibernateUtil.GuessType(15), Is.EqualTo(NHibernateUtil.Int32));
 		}
 
 		[Test]
 		public void CanGuessTypeOfInt32ByType()
 		{
-			Assert.AreEqual(NHibernateUtil.Int32,
-				NHibernateUtil.GuessType(typeof(int)));
+			Assert.That(NHibernateUtil.GuessType(typeof(int)), Is.EqualTo(NHibernateUtil.Int32));
 		}
 
 		[Test]
 		public void CanGuessTypeOfNullableInt32ByType()
 		{
-			Assert.AreEqual(NHibernateUtil.Int32,
-				NHibernateUtil.GuessType(typeof(int?)));
+			Assert.That(NHibernateUtil.GuessType(typeof(int?)), Is.EqualTo(NHibernateUtil.Int32));
 		}
 
 		[Test]
 		public void CanGuessTypeOfNullableInt32ByValue()
 		{
 			int? val = 15;
-			Assert.AreEqual(NHibernateUtil.Int32,
-				NHibernateUtil.GuessType(val));
+			Assert.That(NHibernateUtil.GuessType(val), Is.EqualTo(NHibernateUtil.Int32));
 		}
 
 		[Test]
 		public void CanGuessTypeOfDateTime()
 		{
-			Assert.AreEqual(NHibernateUtil.DateTime,
-				NHibernateUtil.GuessType(typeof(DateTime)));
+			Assert.That(NHibernateUtil.GuessType(typeof(DateTime)), Is.EqualTo(NHibernateUtil.DateTime));
+		}
+
+		[Test]
+		public void CanGuessTypeOfString()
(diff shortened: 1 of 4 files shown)


fredericDelaporte:
This branch is not clean, it can not be merged. In fact, you should not work directly on your master, but keep it as a synchronization point with NHibernate master, and work on feature branches, as asked in [contributing](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md).

Start by creating a branch from your current master in order to save in it your work.
```
git branch 1496
```
(I tend to use as branch name the issue number if there is an issue, or some short name describing the PR if I am doing a direct PR without issue. But you can name it as you want.)
Then reset the additional commits on your [master](https://github.com/mishaurubkov/nhibernate-core/) (compared to NHibernate master, currently 4 ahead):
```
git checkout master
git reset head~4 --hard
```
The `4` above has to match the number of commits your master is ahead of NHibernate.
(The checkout is done just for ensuring you are on your master, which should be already the case. But better be safe.)
Override your remote:
```
git push -f
```
And then check on GitHub your [master](https://github.com/mishaurubkov/nhibernate-core/) is no more ahead of NHibernate.

From there, better get your master even with NHibernate (no more behind).

You need for this to have a remote set up on NHibernate. If you do not know, check your remotes with:
```
git remote -v
```
It should at least show your origin. Seek if you have a remote on `https://github.com/nhibernate/nhibernate-core` too. On my box, that shows up as:
```
origin  https://github.com/fredericDelaporte/nhibernate-core (fetch)
origin  https://github.com/fredericDelaporte/nhibernate-core (push)
upstream        https://github.com/nhibernate/nhibernate-core (fetch)
upstream        https://github.com/nhibernate/nhibernate-core (push)
```
As you can see, I have named it `upstream`. If you have one, use the name it has on your box. Otherwise, add it:
```
git remote add upstream https://github.com/nhibernate/nhibernate-core
```

From now on, I will assume `upstream` is NHibernate remote. Change this name in next commands if you use a different name.

Ensure your remote is up to date:
```
git fetch upstream
```

Rebase your master on NHibernate:
```
git checkout master
git rebase upstream/master
git push
```
(The checkout is done just for ensuring again you are on your master, not on a feature branch.)
This time, since normally you were no more ahead of NHibernate, the push should not need to be forced (no `-f` parameter).
And then you should see on GitHub that your [branch](https://github.com/mishaurubkov/nhibernate-core/) is now even with NHibernate.

From there, you have two choice for fixing your PR:
 - Rebase your feature branch on your master, created at the beginning. You may have conflicts to solve due to its current state.
    ```
    git checkout 1496
    git rebase master
    ```
 - Create another feature branch and cherry-pick just the commit for this PR.
    ```
    git checkout master
    git branch new1496
    git checkout new1496
    git cherry-pick f4e30550
    ```
    The `f4e30550` is the beginning of the commit hash. 8 characters is usually enough, otherwise you can write it fully (`f4e3055052279a45fd0f6d03ca9391a891e83ef3` in this case).

And then publish the branch and create an new PR from it. This one here will be closed.

src/NHibernate.Test/NHSpecificTest/GH1496/Mappings.hbm.xml
```diff
@@ -0,0 +1,63 @@
...
+      <column name="WorkContactId"/>
+      <formula>'WORK'</formula>
+    </many-to-one>
+  </class>
+
+  <class name="Contact"
+		table="Contact"
+		entity-name="HomeContact"
+		where="TypeName='HOME'"
+		check="TypeName in ('HOME', 'WORK')"
```
fredericDelaporte: This looks unsupported by Firebird. Is having this relevant for the bug? If no, please simplify as much as possible the test for having it as minimal as possible for demonstrating the bug.
Otherwise override the `AppliesTo(Dialect dialect)` in your test case, and add something like
```c#
if (dialect is FireBirdDialect)
    Assert.Ignore("Firebird is not supported by this test");
```

mishaurubkov: Sure, I will simplify it.

mishaurubkov: Also I will follow your instructions and create new branch.

fredericDelaporte:
Replaced by #1499.

PR #1498, opened 2017-12-21
Cease throwing bare Exception

fredericDelaporte:
Throwing the `Exception` class is a bad practice. We should throw at least slightly more specific exceptions.


src/NHibernate/Criterion/ProjectionsExtensions.cs
```diff
@@ -70,7 +70,7 @@ internal static IProjection ProcessDate(System.Linq.Expressions.Expression expre
 		/// </summary>
 		public static double Sqrt(this double numericProperty)
 		{
-			throw new Exception("Not to be used directly - use inside QueryOver expression");
+			throw new InvalidOperationException("Not to be used directly - use inside QueryOver expression");
```
bahusoid: I think it's better to wrap this exception creation in some method to minimize copy-paste for future extensions/refactorings. So it should look something like:
`throw DirectUsageException();`

teamcity.build
```diff
@@ -389,10 +389,10 @@
 				File.WriteAllText(outputFile, output);
 
 				if (brokenTests.Count > 0)
-					throw new Exception("Previously passing tests have been broken\n\n" + output);
+					throw new InvalidOperationException("Previously passing tests have been broken\n\n" + output);
```
hazzik: This is not InvalidOperationException. This is either need to stay as an `Exception` or another exception type, probably from Nant's namespace.

src/NHibernate/Criterion/ProjectionsExtensions.cs
gliljas: Does it have to throw at all? Real methods are replaced in queries all the time, so why not here, i.e call Math.Sqrt and be done with it?

hazzik: @gliljas could do, but in another RP.

src/NHibernate.DomainModel/FooComponent.cs
```diff
@@ -110,7 +110,7 @@ private String NullString
 			set
 			{
 				if (value != null)
-					throw new Exception("null component property");
+					throw new InvalidOperationException("null component property");
```
hazzik: Should be `ArgumentException`/`ArgumentNullException`

teamcity.build
fredericDelaporte: Well, reverting. This script is currently no more used so changing it is not tested.

src/NHibernate/Criterion/ProjectionsExtensions.cs
fredericDelaporte: > Does it have to throw at all? Real methods are replaced in queries all the time, so why not here, i.e call Math.Sqrt and be done with it?

Some cases could be trickier, like the `Projection.Concat` where `string.Concat` `null` handling will not be similar to SQL default `null` handling. So this may cause more troubles than simply throwing.

fredericDelaporte:
I have made more changes, especially in `ExpressionProcessor` where many case were argument validations.

PR #1499, opened 2017-12-22
Fix ManyToOneType.IsModified to handle both object instance and identifier passed to the "old" parameter 

### LINKED ISSUE #1496: Fix ManyToOneType.IsModified to handle both object instance and identifier passed to the parameter “old”.

mishaurubkov:
… identifier passed to the parameter “old”.

The tests `EventListener_Entity_ChangeProperty` and `EventListener_EntityWithCompositeId_ChangeProperty` fail for current code in `ManyToOneType.IsModified(…)` where `old` parameter is not parsed.

The tests `SelectBeforeUpdate_EntityWithCompositeId_ChangeProperty` and `SelectBeforeUpdate_ManyToOneWithCompositeId_ChangeProperty` fail
 if we apply Hibernate fix [HHH-3730](https://hibernate.atlassian.net/browse/HHH-3730) `(var oldIdentifier =GetIdentifier(old, session))`

All tests pass if we apply proposed fix: `var oldIdentifier = IsIdentifier(old, session) ? old : GetIdentifier(old, session);`

I added extra tests to see what data hit the `ManyToOneType.IsModified(…)`. We can leave just 4 tests listed above and remove or set to `[Ignore]` others.

`ManyToOneType.IsModifiedAsync(...)` could be fixed same way but would like to hear the feedback on the issue and proposed solution first.

Fixes #1496.


fredericDelaporte:
> ManyToOneType.IsModifiedAsync(...) could be fixed same way but would like to hear the feedback on the issue and proposed solution first.

No need to wait, async code is generated: just regenerate it. (Do not manually edit it.) Follow [contributing](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md) instructions. [Async generation](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md#regenerate-async-code) is an option of the [build menu](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md#the-build-menu).

mishaurubkov:
DONE. Generated Async method with tests. See #98b1918 commit.

src/NHibernate.Test/NHSpecificTest/GH1496/AuditEventListener.cs
```diff
@@ -0,0 +1,69 @@
...
+			ModifiedItems.Clear();
+			isActive = true;
+		}
+
+		public void Stop()
+		{
+			isActive = false;
+		}
+
+		public async Task OnPostUpdateAsync(PostUpdateEvent @event, CancellationToken cancellationToken)
```
fredericDelaporte: Normally there is a no need to write this async method too. The async generator will do it for you too. Of course this means that at first your class will lack some interface method implementations and the project will not compile till you run the async generator, but this is the preferred way.

src/NHibernate.Test/NHSpecificTest/GH1496/Entities.cs
```diff
@@ -0,0 +1,129 @@
...
+
+		public Person(int id, string name, Address address)
+		{
+			Id = id;
+			Name = name;
+			Address = address;
+		}
+	}
+
+	// Entity with ManyToOne type (Contact) with comnposite (ContactId) Id.
```
fredericDelaporte: Typo on `comnposite`.

src/NHibernate.Test/NHSpecificTest/GH1496/Entities.cs
```diff
@@ -0,0 +1,129 @@
...
+
+	[Serializable]
+	public class Contact
+	{
+		public virtual ContactIdentifier ContactIdentifier { get; set; }
+		public virtual string Phone { get; set; }
+	}
+
+
+	public class ContactIdentifier : IEquatable<ContactIdentifier>
```
fredericDelaporte: Shouldn't this be marked serializable too?

src/NHibernate.Test/NHSpecificTest/GH1496/Entities.cs
```diff
@@ -0,0 +1,129 @@
...
+		{
+			return !Equals(left, right);
+		}
+
+
+
+	}
+
+
+
```
fredericDelaporte: Please clean the file from such consecutive blank lines.

mishaurubkov:
I did cleanup and committed the changes:
1) OnPostUpdateAsync was created using generator
2) Removed spaces
3) I didn't really need [Serializable] so I removed it.

PR #1505, opened 2017-12-27
Chaining scopes with ODBC is bugged: disabling the test.

fredericDelaporte:
~~It enlists in a disposed transaction~~ Its transaction enlistment checks the previous transaction which may be disposed, causing a failure.

This test is flaky with ODBC. It fails sometime on dual core VM (which AppVeyor is, and I believe to have seen it on TeamCity when it was an Azure BS2, but I am not finding a build showing it).
Example failure here:
https://ci.appveyor.com/project/fredericDelaporte/nhibernate-core/build/5.1.0.46/job/kwp2bah4pxf8evuk/tests

### DIFF
--- a/src/NHibernate.Test/Async/SystemTransactions/SystemTransactionFixture.cs
+++ b/src/NHibernate.Test/Async/SystemTransactions/SystemTransactionFixture.cs
@@ -14,6 +14,7 @@ using System.Linq;
 using System.Threading;
 using System.Transactions;
 using NHibernate.Cfg;
+using NHibernate.Driver;
 using NHibernate.Engine;
 using NHibernate.Linq;
 using NHibernate.Test.TransactionTest;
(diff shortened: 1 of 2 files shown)


fredericDelaporte:
I will add some complements in comments. I remember something like ODBC going always distributed (this test is not supposed to be distributed), and I think the trouble arise due to the test being distributed, its completion delayed, while the "dodge transaction completion" put in distributed tests is not there since here we are in a case not supposed to be distributed. So I will check all this.

fredericDelaporte:
Here lies the ODBC bug (decompiled with Rider):

```c#
public override void EnlistTransaction(Transaction transaction)
{
  OdbcConnection.ExecutePermission.Demand();
  Bid.Trace("<prov.DbConnectionHelper.EnlistTransaction|RES|TRAN> %d#, Connection enlisting in a transaction.\n", this.ObjectID);
  Transaction enlistedTransaction = this.InnerConnection.EnlistedTransaction;
  if (enlistedTransaction != (Transaction) null)
  {
    if (enlistedTransaction.Equals((object) transaction))
      return;
    if (enlistedTransaction.TransactionInformation.Status == System.Transactions.TransactionStatus.Active)
      throw ADP.TransactionPresent();
  }
  this.InnerConnection.EnlistTransaction(transaction);
  GC.KeepAlive((object) this);
}
```

First I confirm ODBC (at least with SQL-Server) always causes scopes to go distributed. Then, the code above in `OdbcConnection` is bugged: it checks the previously enlisted transaction without guarding against it being disposed. But accessing `enlistedTransaction.TransactionInformation` on a disposed transaction throws.

PR #1507, opened 2017-12-27
NH-3119 - fix test not supporting optimization

fredericDelaporte:
It was relying on the stack trace, which is not reliable when calls can be inlined.


fredericDelaporte:
This is one of these special options, not working as the others. I do not think it can be configured by code in another way. I will check again later.

PR #1508, opened 2017-12-27
AppVeyor

fredericDelaporte:
I have enabled AppVeyor on NHibernate, but only for branches having the Yaml file.

This PR relies on three commits I have separately PR-ed (#1505, #1506, #1507). (Now merged and rebased.)

This AppVeyor testes all databases tested on TeamCity excepted Oracle, which is not available on AppVeyor images, and for which scripting an installation is a bit challenging.

It also generates the build artifacts within its release job.

Build times are generally shorter than on TeamCity, but with greater variability. Sometimes a job takes more than double the time than usual. The way AppVeyor works, each job uses a clean VM created just for the job, so it will not have issues like those sometimes occurring on the TeamCity build agent due to things like Windows Update.

Currently all the work is scripted in the Yaml file, but it could be split a bit if wished by putting powershell scripts in dedicated files.

We could moreover add a 'beta release' job which would publish pre-release NuGet from merges on master, allowing easier early testing.

TeamCity only supplies the software: the build agent has to be provided. Switching to AppVeyor would avoid having to supply it.

About the Oracle case, the first difficulty is to get an installer for it on the build machine. It should be downloaded from somewhere, but Oracle, as far as I know, does not provide a public accessible URI for such a download.

@hazzik has suggested me three options:
 * > Get this URL as a secure AppVeyor variable

    Meaning putting an installer "somewhere", without disclosing publicly its url thanks to AppVeyor encryption. Using the AppVeyor cache feature, we could then even suppress it once it has been cached for all jobs of the matrix (cache is isolated by matrix jobs).
 * > Use docker image for Oracle (it will require running docker for ubuntu-on-windows)

    Surely cleaner than the first option. And I guess it would also solve the second likely issue, getting Oracle installed by scripts. Now about Docker, my knowledge is currently limited to knowing that it exists. Well, AppVeyor also seems to have some built-in [support](https://www.appveyor.com/docs/build-environment/#docker) for it, included in the image used in this PR.
 * > Or we can leave Oracle on the TeamCity.

    Why not, it would at least allow to reduce its "size", and maybe wait for a native support by AppVeyor.

So this PR currently takes the third path.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,335 @@
+version: 5.1.0.{build}
+image: Visual Studio 2017
+configuration:
+    - Debug
+    - Release
+environment:
+    global:
+        targetFx: net461
+    matrix:
+        - db: Firebird
+        - db: MySQL
+        - db: Odbc
+        - db: PostgreSQL
+        - db: SQLite
+        - db: SqlServerCe
+        - db: SqlServer2008
+        - db: SqlServer2012
+
+matrix:
+    exclude:
+        - configuration: Release
+          db: Firebird
+        - configuration: Release
+          db: MySQL
+        - configuration: Release
+          db: Odbc
+        - configuration: Release
+          db: PostgreSQL
+        - configuration: Release
+          db: SQLite
+        - configuration: Release
+          db: SqlServerCe
+        - configuration: Release
+          db: SqlServer2008
+
+init:
+    # Required for having windows endlines in sources zip
+    - git config --global core.autocrlf true
+
+cache:
+    - C:\firebird\Firebird-3.0.2.32703-0_x64.zip
+
+before_build:
+    - ps: >-
+        Invoke-Command -ScriptBlock {
+            # Package sources if this is a release build.
+            If ($env:CONFIGURATION -ne 'Release') {
+                return
+            }
+
+            $srcPackage = (Join-Path $env:APPVEYOR_BUILD_FOLDER 'NHibernate-5.0.3-src')
+            mkdir $srcPackage
+            cp (Join-Path $env:APPVEYOR_BUILD_FOLDER 'build-common') $srcPackage -recurse
+            cp (Join-Path $env:APPVEYOR_BUILD_FOLDER 'lib') $srcPackage -recurse
+            cp (Join-Path $env:APPVEYOR_BUILD_FOLDER 'src') $srcPackage -recurse
+            cp (Join-Path $env:APPVEYOR_BUILD_FOLDER "*.build") $srcPackage
+            cp (Join-Path $env:APPVEYOR_BUILD_FOLDER 'LICENSE.txt') $srcPackage
+            cp (Join-Path $env:APPVEYOR_BUILD_FOLDER 'releasenotes.txt') $srcPackage
+            cp (Join-Path $env:APPVEYOR_BUILD_FOLDER 'README.md') $srcPackage
+        }
+
+    - which msbuild.exe
+    - nuget restore src/NHibernate.sln
+
+# Disable automated build, use build script instead
+build: off
+build_script:
+    - ps: >-
+        Invoke-Command -ScriptBlock {
+            $nugetPath = (Join-Path $env:APPVEYOR_BUILD_FOLDER 'nuget')
+            mkdir $nugetPath
+            dotnet msbuild (Join-Path $env:APPVEYOR_BUILD_FOLDER 'src\NHibernate.sln') /verbosity:minimal `
+                '/p:Platform="Any CPU"' /p:GeneratePackageOnBuild=True /p:IncludeSymbols=True `
+                /p:IncludeSource=True "/p:PackageOutputPath=$nugetPath"
+        }
+
+before_test:
+    - ps: >-
+        Invoke-Command -ScriptBlock {
+            $configDir = (Join-Path $env:APPVEYOR_BUILD_FOLDER 'current-test-configuration')
+
+            #Configuration matrix
+            $allSettings = @{
+                'Firebird' = @{
+                    'connection.connection_string' = 'DataSource=localhost;Database=nhibernate;User ID=SYSDBA;Password=masterkey;MaxPoolSize=200;';
+                    'connection.driver_class' = 'NHibernate.Driver.FirebirdClientDriver';
+                    'dialect' = 'NHibernate.Dialect.FirebirdDialect'
+                };
+                'MySQL' = @{
+                    'connection.connection_string' = 'Server=127.0.0.1;Uid=root;Pwd=Password12!;Database=nhibernate;Old Guids=True;';
+                    'connection.driver_class' = 'NHibernate.Driver.MySqlDataDriver';
+                    'dialect' = 'NHibernate.Dialect.MySQL5Dialect'
+                };
+                'Odbc' = @{
+                    # The OdbcDriver inherits SupportsMultipleOpenReaders=true from DriverBase, which requires Mars_Connection=yes for SQL Server.
+                    'connection.connection_string' = 'Server=(local)\SQL2017;Uid=sa;Pwd=Password12!;Database=nhibernateOdbc;Driver={SQL Server Native Client 11.0};Mars_Connection=yes;';
+                    'connection.driver_class' = 'NHibernate.Driver.OdbcDriver';
+                    'odbc.explicit_datetime_scale' = '3';
+                    <# We need to use a dialect that avoids mapping DbType.Time to TIME on MSSQL. On modern SQL Server
+                        this becomes TIME(7). Later, such values cannot be read back over ODBC. The
+                        error we get is "System.ArgumentException : Unknown SQL type - SS_TIME_EX.". I don't know for certain
+                        why this occurs, but MS docs do say that for a client "compiled using a version of SQL Server Native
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -0,0 +1,335 @@
...
+        - db: MySQL
+        - db: Odbc
+        - db: PostgreSQL
+        - db: SQLite
+        - db: SqlServerCe
+        - db: SqlServer2008
+        - db: SqlServer2012
+
+matrix:
+    exclude:
```
fredericDelaporte: Debug configuration is tested with all databases, release is done only with SQL 2012.

appveyor.yml
```diff
@@ -0,0 +1,335 @@
...
+            }
+            $config.Save($configFile)
+        }
+
+    - ps: >-
+        Invoke-Command -ScriptBlock {
+            switch ($env:db) {
+                'Firebird' {
+                    # Install Firebird
+                    # Firebird zip is cached, check if we have it, otherwise download it
```
fredericDelaporte: There is choco package for Firebird too. I have not tested it, and it currently has a failing status. This manual install is not complicated, so I have not estimated it was trying the choco package.

appveyor.yml
```diff
@@ -0,0 +1,335 @@
...
+                    # Declare nhibernate db
+                    Add-Content -Path '.\databases.conf' -Value "`r`nnhibernate = C:\firebird\Data\nhibernate.fdb"
+                    # Minimal db settings
+                    Add-Content -Path '.\firebird.conf' -Value "`r`nAuthServer = Srp`r`nAuthClient = Srp`r`nUserManager = Srp`r`nWireCrypt = Enabled"
+                    # Create SYSDBA account
+                    New-Item SYSDBA.sql -Type File
+                    Add-Content -Path '.\SYSDBA.sql' -Value "create user SYSDBA password 'masterkey';`r`ncommit;`r`nquit;"
+                    .\isql -user sysdba employee -input SYSDBA.sql
+                    # Start Firebird
+                    .\firebird -a
```
fredericDelaporte: Not even installed as a service, just directly run.

appveyor.yml
```diff
@@ -0,0 +1,335 @@
+version: 5.1.0.{build}
+image: Visual Studio 2017
```
fredericDelaporte: Required for having the `dotnet` command.

appveyor.yml
```diff
@@ -0,0 +1,335 @@
...
+                    iex "& $cmd"
+                }
+                'Odbc' { Start-Service 'MSSQL$SQL2017' }
+                'PostgreSQL' {
+                    # Enable prepared transactions
+                    Add-Content -Path 'C:\Program Files\PostgreSQL\10\data\postgresql.conf' -Value "`r`nmax_prepared_transactions = 100"
+                    Start-Service 'postgresql-x64-10'
+                }
+                'SqlServer2008' { Start-Service 'MSSQL$SQL2017' }
+                'SqlServer2012' { Start-Service 'MSSQL$SQL2017' }
```
fredericDelaporte: The `Visual Studio 2017` image does not have SQL 2008 and SQL 2012, it has [2014, 2016 and 2017](https://www.appveyor.com/docs/build-environment/#databases). So I have picked one of them.

appveyor.yml
```diff
@@ -0,0 +1,335 @@
...
+            for ($i=0; $i -lt $saxonClassPath.length; $i++) {
+                $saxonClassPath[$i] = (Join-Path $refFolder $saxonClassPath[$i])
+            }
+            $saxonClassPath = $saxonClassPath -Join ';'
+            $masterFile = (Join-Path $refFolder 'master.xml')
+            #Html
+            $html = (Join-Path $refPackage 'html')
+            mkdir $html
+            cd $html
+            java -classpath $saxonClassPath com.icl.saxon.StyleSheet $masterFile (Join-Path $refFolder 'styles/html_chunk.xsl')
```
fredericDelaporte: These java commands does also yield non-error messages in the error output, which is a bit annoying when watching the log output on AppVeyor.

ngbrown:
When I developed an AppVeyor build, I did it with the intent to also be able to run in a Linux docker container so the builds and tests for .Net core could run on Linux with Circle-CI (see fd78581b50ecb7463806863ceea0077d5cd780b2).  The same would apply for testing under macOS.

I used [Cake](https://github.com/cake-build/cake) so I could use only the .NET Core SDK and didn't need Mono configured.

With no new releases of NAnt, does it still work well with newer versions of Mono?

fredericDelaporte:
> With no new releases of NAnt, does it still work well with newer versions of Mono?

No idea. But NAnt is not used by the AppVeyor build here. All steps have been coded as scripts in appveyor.yaml (mostly powershell scripts), without any dependency on some other scripts in NHibernate repository.

I have not written that AppVeyor build for .Net Core, but as a way to migrate away from TeamCity.

hazzik:
The idea of CI is that it uses the same tools and means as you would use on your machine. The ultimate goal was to switch away from NAnt to MSBuild. 

I'm strongly against introducing new unnecessary tools like Cake or PowerShell.

fredericDelaporte:
> The idea of CI is that it uses the same tools and means as you would use on your machine.

I do not have this view. I never use the same tools because one is a fully automated process while the other is a process with a human in the loop. On my machine, I build and test from an IDE most of the time, not from command line. I am mostly never building doc or packages from my machine, because that is the build server job. Powershells script can still be run on my machine if I need to debug them locally.

> Otherwise we will need to rewrite scripts for different CI's

That will always be the case. They will always have some variations requiring some dedicated scripting, like having to start the services we need, installing additional software, adjusting configuration to the way the build agent is setup, ...

> PowerShell (for the reasons of duplication)

All the "duplicated as ps script" NAnt stuff could be removed, and replaced by the powershell scripts with some more configuration variables. (Off course, only once TeamCity is dropped.)

But what should be considered duplicated? The teamcity NAnt tasks for tests are not really "machine agnostic", doing a bunch of assumption on the way databases are installed. Most of them cannot run on my machine due to this. We have the build menu for running tests with the "current test database" configuration instead, which could already be considered as a kind of duplicate of the teamcity NAnt tasks for tests.

hazzik:
>(Off course, only once TeamCity is dropped.)

i don't understand how this is related.

gliljas:
I agree with @fredericDelaporte, and I think that aiming for MSBuild only is wrong. There's a reason why so many builds are moving away from declarative "scripts". 

Cake is great. It has quite a few rough edges, but in general I find it to be much more convenient than Powershell.

fredericDelaporte:
> >(Off course, only once TeamCity is dropped.)
>
> I don't understand how this is related.

It is related because I am not considering migrating the TeamCity build away from NAnt if we consider moving away from TeamCity.

About Cake vs PowerShell, I never tried Cake, so I cannot judge of its advantages. For what we have to do, doing it "only" with PowerShell does not seem over complicated to me (335 lines in the Yaml, which includes all the PS scripts currently). And PowerShell for sure has more backing than Cake, so I am not convinced it is worth introducing Cake here.
Gunnar, maybe you could PR a Cake version on my AppVeyor branch if you wish, to show how things would be changed.

Happy new year to you all by the way.

ngbrown:
@fredericDelaporte This isn't really PowerShell that can be used or tested anywhere else, but PowerShell code embedded in a `.yml` file.  If it is meant to replace the NAnt builds, then there should be `.ps1` files that can be ran and tested from outside of the CI server.

Microsoft has released PowerShell to both Linux and macOS, so it is an option, but again the `.ps1` files need to exist, and using something like the [psake](https://github.com/psake/psake) modules would make for a better build experience.

Cake is an alternative to having to install PowerShell and use psake.  It works on all platforms that `dotnet` runs on and can be used in both CI and on developer machines.  Of course, even with Cake, there are differences between CI environments, but that is confined to only dealing with the differences instead of having the entirety of the build in a CI specific format.

If you are open to the option, I could expand my current Cake file to do everything that you are trying to do, but for now, it builds and tests on AppVeyor (Windows), Circle-CI (Linux and theoretically macOS), and the command-line.  Building documentation wasn't going to be possible on Linux/macOS anyways:  https://github.com/nhibernate/nhibernate-core/commit/fd78581b50ecb7463806863ceea0077d5cd780b2#diff-4d3634747586b964df9f7712bc3ec669

As for using msbuild for providing a single unifying build pipeline for all required tasks on all systems; maybe it's possible, but it certainly seems very obtuse to do.  Even Microsoft, who should know the most in how to build entirely with msbuild, uses a lot of `.ps1` files for builds: https://github.com/Microsoft/ChakraCore/tree/master/Build/scripts and https://github.com/aspnet/EntityFrameworkCore/blob/dev/run.ps1

fredericDelaporte:
This PR is an attempt at moving away from TeamCity, for no more having to provide a build agent. It is not meant to supply a general build solution.

Although in my opinion the current NHibernate NAnt builds are also far from supplying a general build solution too, it is closer to it, having some parts re-used by the `ShowBuildMenu.bat`. I was not considering this partial sharing of build script as an important asset. But it seems it is, looking at the comments here.

So in this regard this PR is wrong. It cannot simply move away from TeamCity. First a truly general build logic has to be put in place, very likely as a dedicated PR not changing the CI platform at the same time.

I am not much knowledgeable in build systems overall, and moreover, I am not very interested in investing time investigating solutions like Cake, psake or whatever. So it looks like this subject will have to wait for someone else to propose a PR for such a general build solution first.

PR #1509, opened 2017-12-29
Add missing NHibernateLogLevel.Info in example web project

ngbrown:
For logging, Serilog wasn't handling the `null` argument case correctly and there was a missing `NHibernateLogLevel.Info` in example web project

### DIFF
--- a/src/NHibernate.Example.Web/Infrastructure/NHibernateToMicrosoftLogger.cs
+++ b/src/NHibernate.Example.Web/Infrastructure/NHibernateToMicrosoftLogger.cs
@@ -18,6 +18,7 @@ namespace NHibernate.Example.Web.Infrastructure
 		{
 			{ NHibernateLogLevel.Trace, LogLevel.Trace },
 			{ NHibernateLogLevel.Debug, LogLevel.Debug },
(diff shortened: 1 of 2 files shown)


src/NHibernate/Logging.extensions.cs
```diff
@@ -20,7 +20,7 @@ public static void Fatal(this INHibernateLogger logger, Exception exception, str
 
 		public static void Fatal(this INHibernateLogger logger, Exception exception, string message)
 		{
-			logger.Log(NHibernateLogLevel.Fatal, new NHibernateLogValues(message, null), exception);
+			logger.Log(NHibernateLogLevel.Fatal, new NHibernateLogValues(message, Array.Empty<object>()), exception);
```
fredericDelaporte: Couldn't this be handled in the struct constructor, which is already doing this for its `message` argument?

ngbrown: Good point.  Changed and rebased.

fredericDelaporte:
> I believe this is Serilog's problem and shall be fixed in the adapter.

Not wrong, but as the fix is so lightweight, as you say:

> Who cares

PR #1511, opened 2018-01-02
Test Unicode string.

fredericDelaporte:
The tests do not check Unicode string, here is a basic test for this.
I expect it to fail for Firebird and Oracle. I will add test fix commits (and async regen) afterward.

### DIFF
--- a/src/NHibernate.Config.Templates/FireBird.cfg.xml
+++ b/src/NHibernate.Config.Templates/FireBird.cfg.xml
@@ -22,6 +22,7 @@ for your own use before compile tests in VisualStudio.
 			Database=nhibernate;
 			User ID=SYSDBA;Password=masterkey;
 			MaxPoolSize=200;
+			charset=utf8;
 		</property>
 		<property name="show_sql">false</property>
 		<property name="dialect">NHibernate.Dialect.FirebirdDialect</property>
--- a/src/NHibernate.Config.Templates/Oracle-Managed.cfg.xml
+++ b/src/NHibernate.Config.Templates/Oracle-Managed.cfg.xml
@@ -14,5 +14,12 @@ for your own use before compile tests in VisualStudio.
 		<property name="show_sql">false</property>
 		<property name="dialect">NHibernate.Dialect.Oracle10gDialect</property>
 		<property name="query.substitutions">true 1, false 0, yes 'Y', no 'N'</property>
+		<!-- If your database setup use an ASCII charset, switch following property to true. -->
+		<property name="oracle.use_n_prefixed_types_for_unicode">false</property>
+		<!-- Depending on your database setup, the default cast length of 4000 may be too big.
+		     By example, if previous setting is true, NHibernate may try to use nvarchar2(4000),
+		     which will be rejected if its underlying charset is UTF16 and the database
+		     MAX_STRING_SIZE is not extended. In such case, reduce it to 2000. -->
+		<property name="query.default_cast_length"></property>
 	</session-factory>
 </hibernate-configuration>
--- a/build-common/teamcity-hibernate.cfg.xml
+++ b/build-common/teamcity-hibernate.cfg.xml
@@ -24,5 +24,7 @@
 			<property name="adonet.wrap_result_sets">false</property>
 
 			<property name="odbc.explicit_datetime_scale"></property>
+			<property name="oracle.use_n_prefixed_types_for_unicode"></property>
+			<property name="query.default_cast_length"></property>
 		</session-factory>
 	</hibernate-configuration>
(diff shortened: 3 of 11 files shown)


src/NHibernate.Config.Templates/FireBird.cfg.xml
```diff
@@ -22,6 +22,7 @@ for your own use before compile tests in VisualStudio.
 			Database=nhibernate;
 			User ID=SYSDBA;Password=masterkey;
 			MaxPoolSize=200;
+			charset=utf8;
```
fredericDelaporte: Default charset for FireBird is "None", theoretically meaning that it is the application business to encode/decode it, Firebird being supposed to stores and gives data back "as is", maybe in some binary sens. In our case, that goes wrong. I have not investigated why, I have just switched that to utf8 instead.

This requires to recreate the database, which the TeamCity build does for Firebird. For local tests, adjust your configuration files accordingly and run the TestDatabaseSetup test for fixing this.

(Of course this file here is just the default template, the actual fix for TeamCity is the similar change done in teamcity.build file.)

src/NHibernate.Config.Templates/Oracle-Managed.cfg.xml
```diff
@@ -14,5 +14,6 @@ for your own use before compile tests in VisualStudio.
 		<property name="show_sql">false</property>
 		<property name="dialect">NHibernate.Dialect.Oracle10gDialect</property>
 		<property name="query.substitutions">true 1, false 0, yes 'Y', no 'N'</property>
+		<property name="oracle.use_n_prefixed_types_for_unicode">false</property>
```
fredericDelaporte: For helping those installing Oracle with an non-Unicode default charset, I have added this parameter in the sample file. So they will just need to switch it to true, as explained in the instruction file in the same folder.

fredericDelaporte:
@hazzik, the remaining Oracle failure is a new issue. A cast to `nvarchar2(4000)` fails with ORA-00910, meaning the test has tried to oversize the string type. I think that when Oracle was recently re-installed on TeamCIty, it has lost a setting: [extended string support](https://dbasolved.com/2013/06/26/change-varchar2-to-32k-12c-edition/).

I am going to enable it.

A bit of background for explaining why I think this:

When [NH-4062](https://nhibernate.jira.com/browse/NH-4062) (Properly handle Oracle Unicode support dual model) was done, the TeamCity build was not properly changed for correctly setting `oracle.use_n_prefixed_types_for_unicode`. The TeamCity NHibernate agent needs it since its Oracle is configured "the old default way", with an ASCII charset. The lack of this parameter results in all string types using `varchar2` instead of `nvarchar2`, which on TeamCity's Oracle current setup, means using an non Unicode string type. Thus the failure of the new test I have added here. But since no previous were actually requiring Unicode, this a gone unoticed, causing the remaining failing test here to use a `varchar2(4000)`.

Before NH-4062, NHibernate Oracle dialect was assuming `nvarchar2` has to be always used for Unicode string, which is wrong with modern setup and not recommended. (See by example [here](https://dba.stackexchange.com/a/114590/28061).) So before NH-4062, the failing test here was already using a `nvarchar2(4000)`, but without triggering the ORA-00910. This is only possible under two cases: either the `NLS_NCHAR_CHARACTERSET` was not the default UTF16 but UTF8 instead, or `MAX_STRING_SIZE` was set to `EXTENDED`.

That is why I think the recent re-installation of Oracle on the build agent has lost a setting.

So we can fix the remaining failing test here by re-configuring Oracle on TeamCity.

But by the way, that also means the Oracle dialects maximal sizing of string is not very good: depending on how the database is actually configured, they may be too large or too small. Since there is no fixed setting for them, I guess it should be left to the user to define them properly if the NHibernate defaults causes issues. The only way for this is to currently derive a custom dialect from Oracle and override `RegisterCharacterTypeMappings`.

fredericDelaporte:
> That is why I think the recent re-installation of Oracle on the build agent has lost a setting.

Well, bad conclusion. Another change has also occurred: #709, changing the default cast length from 255 to 4000. That is the cause of this test failure, once gone back to `nvarchar2` instead of `varchar2`.

So the simplest is to reduce `DefaultCastLength` to 2000 for Oracle tests. The configuration of `MAX_STRING_SIZE` is quite cumbersome.

fredericDelaporte:
Rebased, some more comments added, Oracle config templates completed. I have done it in such a way that the builds have been run on the first commit, for demonstrating the tests are not properly setup for Unicode support.

hazzik:
LGTM. Please resolve the conflict and merge.

fredericDelaporte:
Rebased.

PR #1514, opened 2018-01-05
Fix exceptions serialization

hazzik:
This is a partial rework of #1469 in a non-breaking way concerning only exceptions.

### DIFF
--- a/src/NHibernate/PropertyNotFoundException.cs
+++ b/src/NHibernate/PropertyNotFoundException.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Runtime.Serialization;
+using System.Security;
+using NHibernate.Util;
 
 namespace NHibernate
 {
@@ -9,10 +11,6 @@ namespace NHibernate
 	[Serializable]
 	public class PropertyNotFoundException : MappingException
 	{
-		private readonly System.Type targetType;
-		private readonly string propertyName;
-		private readonly string accessorType;
-
 		/// <summary>
 		/// Initializes a new instance of the <see cref="PropertyNotFoundException" /> class,
 		/// used when a property get/set accessor is missing.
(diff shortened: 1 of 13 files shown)


src/NHibernate/PropertyNotFoundException.cs
```diff
@@ -65,23 +59,41 @@ public PropertyNotFoundException(string propertyName, string fieldName, System.T
...
+				else if (entry.Name == "AccessorType")
+				{
+					AccessorType = entry.Value?.ToString();
+				}
+				else if (entry.Name == "TargetType")
+				{
+					var typeName = entry.Value?.ToString();
+					if (!string.IsNullOrEmpty(typeName))
+						TargetType = System.Type.GetType(typeName, true);
+				}
```
fredericDelaporte: When having many values, a `switch` seems more appropriate.
```c#
switch (entry.Name)
{
	case "PropertyName":
		PropertyName = entry.Value?.ToString();
		break;
	case "AccessorType":
		AccessorType= entry.Value?.ToString();
		break;
	case "TargetType":
		var typeName = entry.Value?.ToString();
		if (!string.IsNullOrEmpty(typeName))
			TargetType = System.Type.GetType(typeName, true);
		break;
}
```
(`DuplicateMappingException` and `LazyInitializationException` have two values, others only one.)

src/NHibernate/SchemaValidationException.cs
```diff
@@ -16,7 +18,7 @@ public SchemaValidationException(string msg, IList<string> validationErrors) : b
 
 		protected SchemaValidationException(SerializationInfo info, StreamingContext context) : base(info, context)
 		{
-			ValidationErrors =
+			ValidationErrors = 
 				(ReadOnlyCollection<string>) info.GetValue("ValidationErrors", typeof(ReadOnlyCollection<string>));
```
fredericDelaporte: Undue added space.

src/NHibernate/PropertyNotFoundException.cs
hazzik: I did consider this option and even written it this way, but for some reasons decided against it.

src/NHibernate/SchemaValidationException.cs
hazzik: Force-pushed with revert

PR #1517, opened 2018-01-06
Scoped table aliases to session factory configuration

### LINKED ISSUE #1417: Table counter for aliases should be stable

bahusoid:
Make table aliases generated for same mapping but from different session factories the same:
1) To make sure that the same queries are executed for the same database from different apps (I described this scenario [here](https://github.com/nhibernate/nhibernate-core/pull/1503#issuecomment-355638540))

2) To allow further code improvements (we can for instance start interning aliases/pre-generated queries for entities, as this data now should be the same for same mapping)

3) Fixes #1417.
~~Though tableCounter is still rise endlessly - but its value is not used for mapped tables stored in `Configuration.tables`.~~
It turns out that tableCounter is not really needed outside configuration (created table classes don't use UniqueInteger)  - so it's been removed.

### DIFF
--- a/src/NHibernate/Mapping/Table.cs
+++ b/src/NHibernate/Mapping/Table.cs
@@ -30,7 +30,7 @@ namespace NHibernate.Mapping
 		private readonly LinkedHashMap<string, Column> columns = new LinkedHashMap<string, Column>();
 		private readonly Dictionary<ForeignKeyKey, ForeignKey> foreignKeys = new Dictionary<ForeignKeyKey, ForeignKey>();
 		private readonly Dictionary<string, Index> indexes = new Dictionary<string, Index>();
-		private int uniqueInteger;
+		private int? uniqueInteger;
 		private readonly Dictionary<string, UniqueKey> uniqueKeys = new Dictionary<string, UniqueKey>();
 		private string catalog;
 		private string comment;
@@ -189,7 +189,7 @@ namespace NHibernate.Mapping
 		/// <value>The unique number of the Table.</value>
 		public int UniqueInteger
 		{
-			get { return uniqueInteger; }
+			get { return uniqueInteger ?? throw new InvalidOperationException(nameof(UniqueInteger) + " has not been supplied"); }
 			internal set { uniqueInteger = value; }
 		}
 
--- a/src/NHibernate/Mapping/Table.cs
+++ b/src/NHibernate/Mapping/Table.cs
@@ -33,7 +33,7 @@ namespace NHibernate.Mapping
 		private readonly LinkedHashMap<string, Column> columns = new LinkedHashMap<string, Column>();
 		private readonly Dictionary<ForeignKeyKey, ForeignKey> foreignKeys = new Dictionary<ForeignKeyKey, ForeignKey>();
 		private readonly Dictionary<string, Index> indexes = new Dictionary<string, Index>();
-		private readonly int uniqueInteger;
+		private int uniqueInteger;
 		private readonly Dictionary<string, UniqueKey> uniqueKeys = new Dictionary<string, UniqueKey>();
 		private string catalog;
 		private string comment;
@@ -194,6 +194,7 @@ namespace NHibernate.Mapping
 		public int UniqueInteger
 		{
 			get { return uniqueInteger; }
+			internal set { uniqueInteger = value; }
 		}
 
 		/// <summary>
--- a/src/NHibernate/Mapping/Table.cs
+++ b/src/NHibernate/Mapping/Table.cs
@@ -26,9 +26,6 @@ namespace NHibernate.Mapping
 	[Serializable]
 	public class Table : IRelationalModel
 	{
-		[ThreadStatic]
-		private static int tableCounter;
-
 		private readonly List<string> checkConstraints = new List<string>();
 		private readonly LinkedHashMap<string, Column> columns = new LinkedHashMap<string, Column>();
 		private readonly Dictionary<ForeignKeyKey, ForeignKey> foreignKeys = new Dictionary<ForeignKeyKey, ForeignKey>();
@@ -52,7 +49,6 @@ namespace NHibernate.Mapping
 		/// </summary>
 		public Table()
 		{
-			uniqueInteger = tableCounter++;
 		}
 
 		public Table(string name) : this()
--- a/src/NHibernate/Mapping/Table.cs
+++ b/src/NHibernate/Mapping/Table.cs
@@ -185,12 +185,13 @@ namespace NHibernate.Mapping
 
 		/// <summary>
 		/// Gets the unique number of the Table.
+		/// Used for SQL alias generation
 		/// </summary>
 		/// <value>The unique number of the Table.</value>
 		public int UniqueInteger
 		{
 			get { return uniqueInteger ?? throw new InvalidOperationException(nameof(UniqueInteger) + " has not been supplied"); }
-			internal set { uniqueInteger = value; }
+			set { uniqueInteger = value; }
 		}
 
 		/// <summary>
--- a/teamcity.build
+++ b/teamcity.build
@@ -124,7 +124,7 @@
 	<target name="setup-teamcity-oracle32">
 		<property name="nhibernate.connection.driver_class"			value="NHibernate.Driver.OracleDataClientDriver" />
 		<property name="nhibernate.dialect"							value="NHibernate.Dialect.Oracle10gDialect" />
-		<property name="nhibernate.connection.connection_string"	value="User ID=nhibernate;Password=nhibernate;Data Source=(DESCRIPTION = (ADDRESS = (PROTOCOL = TCP)(HOST = localhost)(PORT = 1521)) (CONNECT_DATA = (SERVER = DEDICATED) (SERVICE_NAME = XE)))" />
+		<property name="nhibernate.connection.connection_string"	value="User ID=nhibernate;Password=nhibernate;Metadata Pooling=false;Self Tuning=false;Data Source=(DESCRIPTION = (ADDRESS = (PROTOCOL = TCP)(HOST = localhost)(PORT = 1521)) (CONNECT_DATA = (SERVER = DEDICATED) (SERVICE_NAME = XE)))" />
 		<!-- Teamcity Oracle test database is configured with a non-Unicode encoding -->
 		<property name="nhibernate.oracle.use_n_prefixed_types_for_unicode" value="true" />
 		<copy todir="${bin.dir}">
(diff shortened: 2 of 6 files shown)


bahusoid:
>Keep uniqueInteger readonly.
>Add a constructor taking uniqueInteger as an argument.
It was my first thought. But it's not very convenient to supply index beforehand. Plus this constructor now needs to be also exposed in DenormalizedTable. That's why I made this "shortcut". Though if you think it's a right way to do it - I will do it.

>Remove tableCounter.
Wouldn't it be a breaking change? It's a public class and that changes behaviour of default constructor. It seems it's better to keep tableCounter and keep incrementing it to make obsolete constructor behaviour intact and remove it later along with ctor.

fredericDelaporte:
> Wouldn't it be a breaking change? It's a public class and that changes behaviour of default constructor. It seems it's better to keep tableCounter and keep incrementing it to make obsolete constructor behaviour intact and remove it later along with ctor.

Removing a private member is not a breaking change. Changing the uniqueness of values of a public property is a gray area, but this PR anyway already does this, since previously any new `Table` instance would have an unique value per creation thread, while with this PR it may reuse the value of another table from another configuration.
But keeping it as obsolete (both, the constructor and the member) is also valid, so as you wish.

> But it's not very convenient to supply index beforehand.

There are only three cases where this index is used in NHibernate.csproj: `AddTable`, `AddDenormalizedTable` and `PrepareTemporaryTables`.

I do not see any difficulties neither for `AddTable` nor for `AddDenormalizedTable`. It looks to me as a simple re-ordering of operations in these methods.

The third case will require more work, adding a new overload to `PrepareTemporaryTables` taking an index, obsoleting the old `PrepareTemporaryTables`, exposing the `Configuration.tables` count as an internal property as written previously, ... But that does not look overwhelming to me.

The other usages of `Table` inside NHibernate.csproj are for schema generation. These usages do not use `UniqueIndex` (which is only used for `GetAlias`), they only use `Table` as a kind of name parsing helper.

> Plus this constructor now needs to be also exposed in DenormalizedTable.

Yes, indeed, and the previous one obsoleted too. But that is a small inconvenience I think, allowing to have a consistent `UniqueInteger` computing, instead of mixing your new way and the old way.

There are also many testes in NHibernate.Test.csproj which will be impacted. I have not checked them, but I except them to either not really need that counter, or to be obsoleted by the way we change the counter logic, or to be not hard to fix for supplying a counter valid for the test.

bahusoid:
I'm looking into `PersistentClass.PrepareTemporaryTables`:
https://github.com/nhibernate/nhibernate-core/blob/d372ee3133aeae8aa8e6d8f28877e1d735e71ded/src/NHibernate/Mapping/PersistentClass.cs#L1121-L1134

And from what I see aliases here are irrelevant. Table class is not saved anywhere and only two strings are saved:
`temporaryIdTableName` which is generated using only table name
'temporaryIdTableDDL' which is a create temp table string SQL query. Something like:
```sql
create table #[TableName] ([PrimaryKeyID] INT not null)
```
So aliases are also not needed.

So it seems we don't really need to do anything for temp tables.

fredericDelaporte:
Indeed, I have forgotten to thoroughly check this case. So there, why not using the constructor with table name overload, and forget about the index? It should be the simplest way to do the change.

bahusoid:
@fredericDelaporte Some strange exceptions from Oracle. Don't you happen to know that it's something that might be not related to my changes? ) 
https://teamcity.jetbrains.com/viewLog.html?buildId=1282404&buildTypeId=bt1099#testNameId2417795984423602997

fredericDelaporte:
I have seen them, but I do not see how they could relate to changes here. I can relaunch the build to check if they occur again. Or if you change this PR implementation, we can just wait for the new builds which will be run.

src/NHibernate/Mapping/Table.cs
```diff
@@ -194,6 +190,7 @@ public string Schema
 		public int UniqueInteger
 		{
 			get { return uniqueInteger; }
+			internal set { uniqueInteger = value; }
```
fredericDelaporte: Since this member now depends on being externally supplied (be it by setter or constructor), maybe should we guard against it being used without having been set.
I would do it by switching the backing field to nullable then adjusting the getter to something like `get => uniqueInteger ?? throw new InvalidOperation("UniqueInteger has not been supplied");`. (If using expression bodied getter, change also the setter for consistency.)

bahusoid: @fredericDelaporte Good idea. Will do it. And before making any changes in ctor... Are you sure about it? ) I see no troubles with this change but I just don't see why to fight to keep `uniqueInteger` readonly. Considering that it's the only read-only field in this class.

bahusoid:
>It really looks like changing this counter causes issues to Oracle. Now we have to find out why.
Yep. What's more interesting  - from what I see exception is thrown not from test. It's thrown from test clean up code:
   at NHibernate.Test.TestCase.CheckDatabaseWasCleaned() in ..
   at NHibernate.Test.TestCase.TearDown() in ...

src/NHibernate/Mapping/Table.cs
fredericDelaporte: Forget about the constructor. I mentioned it just to tell that if this member value was supplied through a constructor, I would still consider doing this check in the getter.

fredericDelaporte:
Running NH-1362 test in isolation succeeds. Running it along with NH-1301 test fails. It looks like Oracle caches returned types for a query then fails re-running the same query when the underlying table definition has changed.

Clearing the connection pool avoids the trouble. So we have to use a trick similar to the one put in place for Firebird, see `TestCase.DropSchema`. In the case of Oracle, the clear should likely be put after the schema has been dropped.

src/NHibernate.Test/TestCase.cs
```diff
@@ -313,6 +315,15 @@ protected virtual DebugSessionFactory BuildSessionFactory()
 		private void Cleanup()
 		{
 			Sfi?.Close();
+
+			// Clear connection pool for Oracle to avoid problem that was manifested with https://github.com/nhibernate/nhibernate-core/pull/1517:
+			// As it seems Oracle can cache returned types for query for given connection.
+			// So exception can be thrown if two tests execute same query but with different types in result (like for Entity.Id int and Entity.Id Guid)
+			if (Dialect is Oracle8iDialect)
+			{
+				OracleConnection.ClearAllPools();
```
fredericDelaporte: This will work only if the Oracle driver is the managed one. But two other Oracle drivers are usable. (Currently their specific TeamCity builds are not enabled.)

I have made a PR on your branch for handling them all: bahusoid/nhibernate-core#1.

bahusoid: Thanks. I've adjusted it per @hazzik comments

fredericDelaporte:
Now it succeeds but the [Oracle build](https://teamcity.jetbrains.com/viewLog.html?buildId=1288117&buildTypeId=bt1099&tab=buildResultsDiv) time seems to have raised from twenty minutes to fifty... It is too near to the one hour timeout.

For checking it twice, I have relaunched another [Oracle build](https://teamcity.jetbrains.com/viewLog.html?buildId=1288269&tab=buildResultsDiv&buildTypeId=bt1099) for the latest commit of this PR.

If it takes fifty minutes again, we will have to devise another solution.

 1. Moving the pool clear into `DropSchema`, at its end?
   It would avoid clearing the pool on tests which have been ignored before even creating the schema, on which case there is no reason anything could have been cached with connections. But I do not expect this to gain a lot of time.
 2. Putting the pool clear in the `AppliesTo(ISessionFactoryImplementor factory)` of tests we see failing?
   It will allow clearing the pool far less, but it will not handle new tests, and it may be a bit iterative process: clearing the pool allowing caching new query schema causing other following tests to fail.
 3. ~~A try catch in the `CheckDatabaseWasCleaned`, catching the ORA-00932, clearing the pool in such case and trying again (with some logic for avoiding an endless retry loop)?~~
   It would still be "generic", able of handling new test encountering the trouble, while avoiding clearing the pool for each fixture. But this ORA-00932 trouble could also occur in the test itself as shown by NH-1716 failure [here](https://teamcity.jetbrains.com/viewLog.html?buildId=1285333&buildTypeId=bt1099), not only in its cleanup checking. So this solution will not work.
 4. Or maybe a combination of 3. then 2. for remaining failing tests.
   It will clear the less possible while reducing cases needing to be handled specifically.

Or other idea?

fredericDelaporte:
The second Oracle run has taken fifty minutes again, we need another solution.

teamcity.build
```diff
@@ -116,15 +116,15 @@
...
+		<property name="nhibernate.connection.connection_string"	value="User ID=nhibernate;Password=nhibernate;Data Source=XE;Metadata Pooling=false;Self Tuning=false;" />
 		<property name="NHibernate.Test.IgnoreFail" value="true" />
 		<property name="teamcity.last.result" value="${root.dir}/lib/teamcity/oracle/NHibernate.Test.last-results.xml" />
 	</target>
 
 	<target name="setup-teamcity-oracle32">
 		<property name="nhibernate.connection.driver_class"			value="NHibernate.Driver.OracleDataClientDriver" />
 		<property name="nhibernate.dialect"							value="NHibernate.Dialect.Oracle10gDialect" />
-		<property name="nhibernate.connection.connection_string"	value="User ID=nhibernate;Password=nhibernate;Data Source=(DESCRIPTION = (ADDRESS = (PROTOCOL = TCP)(HOST = localhost)(PORT = 1521)) (CONNECT_DATA = (SERVER = DEDICATED) (SERVICE_NAME = XE)))" />
+		<property name="nhibernate.connection.connection_string"	value="User ID=nhibernate;Password=nhibernate;Metadata Pooling=false;Self Tuning=false;Data Source=(DESCRIPTION = (ADDRESS = (PROTOCOL = TCP)(HOST = localhost)(PORT = 1521)) (CONNECT_DATA = (SERVER = DEDICATED) (SERVICE_NAME = XE)))" />
```
fredericDelaporte: Good finding with these parameters, it works way better than clearing the pool. Can you please update `lib\teamcity\oracle\oracle_installation.txt` too? It should tell about these parameters for helping developers having the tests pass on their box. (But it cannot be put in the config template, since this one is also distributed with NHibernate as a config sample, and that disabling these parameters makes sens only for the tests, not for a normal application.)

It could be adding toward the end of `oracle_installation.txt`, just above its last line I think:
```
Adjust the connection string for the tests:
The tests involve creating and dropping many tables, sometimes with the same names but different data
types. This does not play well with Oracle meta data pooling, which needs to be disabled.
Add into your ODAC connection string:
Metadata Pooling=false;Self Tuning=false;
```

fredericDelaporte:
I have pushed the cleanup and my suggestion about Oracle instructions.

bahusoid:
Thanks for you changes. I've just come back to start working on them :)

bahusoid:
I see there is still some leftovers like (dotSettings file). I will handle it

fredericDelaporte:
I got trapped by the "end of file normalization" due to editor setting. You will need to revert the end file last line with an editor which does not honor `.editorconfig`.

bahusoid:
@fredericDelaporte  Done

fredericDelaporte:
> Let me play a little bit with stable table aliases

What do you mean by this? Have you some additional changes to do here, or can this be merged?

bahusoid:
@fredericDelaporte This PR is ready. I meant play with interning aliases and related stuff. It's all about another PR 1503 where I made this comment.

PR #1519, opened 2018-01-10
Reuse SchemaExport in CreateSchema/DropSchema in tests

bahusoid:


src/NHibernate.Test/TestCase.cs
```diff
@@ -48,6 +49,8 @@ protected virtual string MappingsAssembly
 			get { return "NHibernate.DomainModel"; }
 		}
 
+		protected SchemaExport SchemaExport => _schemaExport ?? (_schemaExport = new SchemaExport(cfg));
```
fredericDelaporte:
```
NHSpecificTest\NH2195\SQLiteMultiCriteriaTest.cs(35,18): warning CS0108: 'SQLiteMultiCriteriaTest.SchemaExport(Configuration)' hides inherited member 'TestCase.SchemaExport'. Use the new keyword if hiding was intended.
```

Release build will fail. Maybe is it time to enable warning as error in the test project debug config too.

bahusoid:
Can I also push here some small modification in `SchemaExport` class to avoid unnecessary calls for `SchemaExport.Initialize` for each `dropSQL`/`createSQL` statement (private Execute method doesn't need to call Initialize)?

fredericDelaporte:
So it would be just removing this [line](https://github.com/nhibernate/nhibernate-core/blob/5.0.3/src/NHibernate/Tool/hbm2ddl/SchemaExport.cs#L170), because this method is called only in one other method which already call `Initialize`.

Still a bit unrelated, but well why not.

bahusoid:
@fredericDelaporte Yeah. Also renamed it to ExecuteInitialized to differentiate it from other public Execute methods where Initialize is called.

PR #1521, opened 2018-01-10
Enable warning as error for all projects and configurations

fredericDelaporte:


build-common/NHibernate.props
```diff
@@ -21,5 +21,8 @@
     <PackageLicenseUrl>https://raw.githubusercontent.com/nhibernate/nhibernate-core/master/LICENSE.txt</PackageLicenseUrl>
     <RepositoryUrl>https://github.com/nhibernate/nhibernate-core.git</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
+
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+    <TreatSpecificWarningsAsErrors />
```
fredericDelaporte: Putting it here has a small side effect in Rider IDE: instead of signaling directly warnings as errors (red underlined in text editor), it signals them as warning. But building still fails in case of warnings.

PR #1523, opened 2018-01-11
.NET Core 2.0 and .NET Standard 2.0 support

hazzik:
These are minimal changes required to make a .NET Core 2.0 and .NET Standard 2.0 builds.

There are no test infrastructure yet, but all tests pass locally on windows dotnet core.

Summary of changes:

1. Assume that serialization of System.Type and co is a foreign concept to .NET Core 2.0 so the users required to write serialization surrogates. We might provide these in the future if we want to. 
2. SqlClient, Odbc, Oledb drivers are converted to ReflectionBasedDriver to avoid the extra dependencies.
3. There are some problems loading App.config when test run from VS/dotnet test. This is a known issue in [NUnit adapter](https://github.com/nunit/nunit3-vs-adapter/issues/356)/[dotnet](https://github.com/dotnet/corefx/issues/22101). We can avoid this issue by switching to NUnitLite to run .NET Core 2.0 tests.
4. CallSessionContext uses a static AsyncLocal field to mimic the CallContext behavior.
5. .NET Core 2.0 DriverBase indicates that System.Transaction is not supported (.NET Standard that it is supported).


hazzik:
Build is failing because NAnt is trying to tell MSBuild to build the project into the same folder.

src/NHibernate.Test/CfgTest/ConfigurationSchemaFixture.cs
```diff
@@ -27,6 +28,8 @@ public void InvalidConfig()
 		[Test]
 		public void FromAppConfigTest()
 		{
+			Assume.That(Assembly.GetEntryAssembly()?.GetName().Name != "testhost");
```
fredericDelaporte: This is used at many place. A common helper for testing this could be nice (and its naming would be more explicit about what case it detects than this code). (I guess this is about the app.config issue in tests.)

src/NHibernate.Test/Async/TransformTests/AliasToBeanResultTransformerFixture.cs
```diff
@@ -257,6 +257,8 @@ public void ToPropertiesInsensitivelyDuplicated_WithoutAnyProjectionsAsync()
 		[Test]
 		public async Task SerializationAsync()
 		{
+			Assume.That(typeof(System.Type).IsSerializable);
```
fredericDelaporte: Since it is put at many places, refactoring this into a common helper may help detect later all tests which needs to be re-enabled if #1425 gets merged. (If the code above is always identical, it could also be found again by a simple search, but still a common helper will be slightly cleaner.)

src/NHibernate.Test/NHSpecificTest/NH2985/Model.cs
```diff
@@ -24,7 +24,7 @@ public class WebImage
 		public WebImage()
 		{
 
-			OldPosition = -1; //di base un immagine � nuova.
+			OldPosition = -1; //di base un immagine � nuova.
```
fredericDelaporte: ascii/unicode corruption?

src/NHibernate/Driver/DriverBase.cs
```diff
@@ -313,9 +313,15 @@ public DbParameter GenerateOutputParameter(DbCommand command)
 
 		public virtual bool RequiresTimeSpanForTime => false;
 
+#if NETCOREAPP2_0
+		public virtual bool SupportsSystemTransactions => false;
```
fredericDelaporte: System transactions are officially in .Net Core 2.0 according to [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.transactions.transactionscope).
But due to the modular nature of .Net Core, it seems its support is still lacking from data providers as stated [here](https://github.com/dotnet/corefx/issues/2949#issuecomment-316480951) or [here](https://github.com/dotnet/corefx/issues/2949#issuecomment-326523019).

Is that the reason for disabling `System.Transactions` globally? It looks like it could be still supported within some data providers.

src/NHibernate/Linq/Functions/StringGenerator.cs
```diff
@@ -242,12 +242,17 @@ public class TrimGenerator : BaseHqlGeneratorForMethod
...
+			{
+				ReflectHelper.GetMethodDefinition<string>(s => s.Trim()),
+				ReflectHelper.GetMethodDefinition<string>(s => s.Trim('a')),
+				ReflectHelper.GetMethodDefinition<string>(s => s.Trim('a', 'a')),
+				ReflectHelper.GetMethodDefinition<string>(s => s.TrimStart()),
+				ReflectHelper.GetMethodDefinition<string>(s => s.TrimStart('a')),
+				ReflectHelper.GetMethodDefinition<string>(s => s.TrimStart('a', 'a')),
+				ReflectHelper.GetMethodDefinition<string>(s => s.TrimEnd()),
+				ReflectHelper.GetMethodDefinition<string>(s => s.TrimEnd('a')),
+				ReflectHelper.GetMethodDefinition<string>(s => s.TrimEnd('a', 'a'))
```
fredericDelaporte: It looks like this file changes could be useful even without a .NetStandard/.NetCore version, or am I wrong? They may deserve a dedicated PR.

hazzik: No. These are new methods added in .NET Core 2.0 (the optimization which accepts one character), but they are not in full .NET. These changes, however will work in full .NET.

LightCZ:
What is the difference between this pull request and #633 ? As i understood this one, this should be the minimal required changes but not finished and #633 have some major breaking changes also. Is that correct?

hazzik:
@LightCZ pretty much. This one is a minimal changes. The other one are with breaking changes. Both of them lack the CI changes so builds fail.

hazzik:
hahaha, no. Check the actual commit. It seems that something is redirecting console output so the timer did not write to it.

appveyor.yml
```diff
@@ -0,0 +1,67 @@
...
+  allow_failures:
+  - DB: MySQL
+build: off
+before_test:
+- ps: |-
+    switch ($env:DB) {
+      'Firebird' {
+        $FireBirdPath = 'C:\firebird'
+        # Install Firebird
+        # Firebird zip is cached, check if we have it, otherwise download it
```
fredericDelaporte: For this to be true, appveyor file must include:
```
cache:
    - C:\firebird\firebird.zip
```
As [here](https://github.com/nhibernate/nhibernate-core/pull/1508/commits/47a11ac4a6181edf025a6691d8a7f843a9e76d6d#diff-180360612c6b8c4ed830919bbb4dd459R40).

And it will actually be taken into account only once merged. On a PR, no caching is done by AppVeyor. (Which is better, this avoids having a bad PR tinkering with cached artifacts.)

hazzik: I had troubles installing firebird on appveyor so disabled the cache. It turns out that the service start did not work in pwsh (powershell core) - the pull hanged after that, so I had to switch to ps.

src/NHibernate.Test/CfgTest/ConfigurationSchemaFixture.cs
hazzik: Done

appveyor.yml
hazzik: I've removed the FB cache check and mentions.

src/NHibernate.Test/Async/TransformTests/AliasToBeanResultTransformerFixture.cs
hazzik: Done

hazzik:
Current state of the PR for the .NET Core 2.0 build:
- SqlServer on Windows & Linux: green
- Postgres on Windows & Linux: green
- Firebird on Windows & Linux: green
- MySQL on 
    - Windows - 4 tests are failing (same tests fail on Full .NET Framework on Windows with 6.10.x provider, 2 of them due to the bug in MySql data adapter);
    - Linux - 8 tests are failing (same 4 as on windows + different ones)

Still TODO:
- [ ] Build cross-platform NuGet package on release
- [ ] Consider adding MacOS X tests
- [ ] Consider adding Oracle tests
- [ ] Remove log4net hack (move log4net configuration from App.config to log4net.config). There are 51 tests failing, but it works on my machine.
- [ ] Refine distributed transaction support. Currently it is disabled for every driver by default.

Not in the scope for this PR:
- Full migration of build system from NAnt
- Serialization support
- Split "driver" packages

ngbrown:
If PSake builds are going to be used, then there should be a separate pull request for them and AppVeyor.  The Linux/macOS specific integration builds (Travis CI, etc.) can remain part of this pull request.

By using NUnit Lite to work around picking up the `app.config`, doesn't that preclude tests working through the GUI test runners when targeting .NET Core?  Selecting individual tests or groups of tests to run in the IDE is very helpful while developing.

fredericDelaporte:
> Refine distributed transaction support. Currently it is disabled for every driver by default.

For every target or only .Net Core/Standard?

And by "distributed transaction support" is it really meant only distributed transactions, meaning scopes are supported provided they are not distributed?
I guess instead it concerns the whole System.Transactions support, meaning no transaction scopes support at all, be they distributed or not distributed. (I think this because many core data providers do not support system transactions at all currently, including the SQL Server one. System.Transactions has been added to .Net Core but without adding its support into data providers... Moreover, support of distributed transactions seems out of scope for .Net Core / Standard, since System.Transactions, when distributed, rely on MSDTC, which is a Windows only technology.)

hazzik:
What I meant is the `SupportsSystemTransactions` property which is set to `false` on `DriverBase` (only in .NET Core) instead of individual drivers. However this property is used _only_ in tests.


>If PSake builds are going to be used, then there should be a separate pull request for that and AppVeyor. The Linux/macOS specific integration builds (Travis CI, etc.) can remain part of this pull request.

AppVeyor tests only .NET Core. .NET Core has to be tested on Windows, as it is completely different platform to Full .NET Framework. So AppVeyor has to stay here.

PSake only provides support for testing .NET Core. Trying to build this on NAnt would be extremely difficult as it does no work on mono anymore. I found that in PS it is easier to work with XML. I could do this in msbuild, but given that the work has been already performed by @fredericDelaporte I just reused it. The difference between bare PowerShell and PSake would be in under 10 lines of code: 2 lines to change function declaration + few lines for tasks orchestration.

> By using NUnit Lite to work around picking up the app.config, doesn't that preclude tests working through the GUI test runners when targeting .NET Core? Selecting individual tests or groups of tests to run in the IDE is very helpful while developing.

The workaround to run under VS/ReSharper/VStest is still there, but activated only when “testhost” is detected.

The NUnitLite has been chosen to run tests as it does not require this workaround, so tests test the behavior closer to how the NHibernate would be run in production.

ngbrown:
> AppVeyor tests only .NET Core. .NET Core has to be tested on Windows, as it is completely different platform to Full .NET Framework. So AppVeyor has to stay here.

AppVeyor can have a configuration that tests full .NET Framework and this would be helpful to cross-check the CI platform.  With at least one database and testing the full .NET Framework, PSake and AppVeyor can have their own pull request.

src/NHibernate/Driver/DriverBase.cs
hazzik: These properties are used only in tests, so I intend to leave them as is.

hazzik:
> Build cross-platform NuGet package on release

NAnt scrips have been fixed to produce the artefacts identical to the artefacts produced before.

> Consider adding Oracle tests

Considered not to do.

>  Consider adding MacOS X tests

I gave it a quick [unsuccessful]  try and decided not to invest more time. There are following pitfalls:

- Travis CI's MacOS does not like bash syntax I've used in the config file
- The environment natively supports only PostgreSQL
- The environment does not support docker, so no SQL Server

> Remove log4net hack

Considered leaving it as is for now. Maybe remove it in the future.

> Refine distributed transaction support ("SupportXXX" properties of the drivers)

These properties are used only in tests, so decided leave them as is.

fredericDelaporte:
The AppVeyor build does not seem to start. Maybe because it has an issue with the numbering reset you seem to have attempted. By the way, maybe we should enable the `Pull Requests do not increment build number` setting.

hazzik:
AppVeyor does not start because of some seems to be common internal issue of AppVeyor. You can check the build in my branch https://ci.appveyor.com/project/hazzik/nhibernate-core

hazzik:
I did not reset numbers, I deleted the project and created it again in the hope it would work after it stuck in the queue. But it did not help.

fredericDelaporte:
Looking [here](https://help.appveyor.com/discussions/problems/5675-all-builds-stuck-at-queued-status), it seems we have to leave them a message for them to unblock the queue.

hazzik:
I did. Builds are working now.

src/NHibernate.TestDatabaseSetup/TestDatabaseSetup.cs
```diff
@@ -19,15 +21,17 @@ public class DatabaseSetup
...
-				{"NHibernate.Driver.OdbcDriver", SetupSqlServerOdbc},
 				{"NHibernate.Driver.FirebirdClientDriver", SetupFirebird},
-				{"NHibernate.Driver.SQLite20Driver", SetupSQLite},
 				{"NHibernate.Driver.NpgsqlDriver", SetupNpgsql},
 				{"NHibernate.Driver.OracleDataClientDriver", SetupOracle},
 				{"NHibernate.Driver.MySqlDataDriver", SetupMySql},
 				{"NHibernate.Driver.OracleClientDriver", SetupOracle},
 				{"NHibernate.Driver.OracleManagedDataClientDriver", SetupOracle},
+#if NET461
+				{"NHibernate.Driver.OdbcDriver", SetupSqlServerOdbc},
```
fredericDelaporte: Odbc is supported, and I have tested its setup locally with Core target, it works.

src/NHibernate/Cfg/Environment.cs
```diff
@@ -309,57 +330,26 @@ static Environment()
 			}
 		}
 
-		private static void LoadGlobalPropertiesFromAppConfig()
+		public static IHibernateConfiguration HibernateConfiguration { get; set; }
```
fredericDelaporte: Shouldn't the setter be private?

src/NHibernate/Context/CallSessionContext.cs
```diff
@@ -1,20 +1,28 @@
...
 	/// </summary>
+#pragma warning restore CS1574
 	[Serializable]
 	public class CallSessionContext : MapBasedSessionContext
 	{
+		#if NET461
 		private const string SessionFactoryMapKey = "NHibernate.Context.CallSessionContext.SessionFactoryMapKey";
+		#else
+		private static readonly AsyncLocal<IDictionary> SessionFactoryMap = new AsyncLocal<IDictionary>();
+		#endif
```
fredericDelaporte: I looks strange to me to fallback to async local. Shouldn't it be considered an error to try using call context with .Net Core?

src/NHibernate/Context/CallSessionContext.cs
```diff
@@ -1,20 +1,28 @@
...
+#if NET461
+using System.Runtime.Remoting.Messaging;
+#endif
 
 namespace NHibernate.Context
 {
+#pragma warning disable CS1574
 	/// <summary>
 	/// Provides a <see cref="ISessionFactory.GetCurrentSession()">current session</see>
 	/// for each <see cref="System.Runtime.Remoting.Messaging.CallContext"/>.
```
fredericDelaporte: Maybe add ` (Uses <c>AsyncLocal</c> instead if run under .Net Core.)`.

src/NHibernate/Linq/Functions/StringGenerator.cs
```diff
@@ -260,13 +265,9 @@ public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject,
...
 
-			string trimChars = "";
-			if (method.GetParameters().Length > 0)
-				foreach (char c in (char[])((ConstantExpression)arguments[0]).Value)
-					trimChars += c;
-
+			var trimChars = ExtractTrimChars(arguments);
 
-			if (trimChars == "")
+			if (trimChars == string.Empty)
```
fredericDelaporte: I would rather use `IsNullOrEmpty` here (or since it is never null, eventually `Length == 0`). Comparing string for testing emptiness is not usual.

src/NHibernate/Linq/Functions/StringGenerator.cs
```diff
@@ -275,6 +276,20 @@ public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject,
...
 			}
 		}
+
+		private static string ExtractTrimChars(IReadOnlyList<Expression> arguments)
+		{
+			if (arguments.Count > 0)
+			{
+				var constantExpression = (ConstantExpression) arguments[0];
+				if (constantExpression.Value is char c)
+					return c.ToString();
```
fredericDelaporte: This code path is an addition compared to previous code. Is this an unrelated bug fix?

src/NHibernate/Cfg/Environment.cs
hazzik: Made internal with private setter.

src/NHibernate/Context/CallSessionContext.cs
hazzik: Done

src/NHibernate/Context/CallSessionContext.cs
hazzik: I think it's better to provide feature-parity. The behaviour will remain consistent with Full .NET.

src/NHibernate/Linq/Functions/StringGenerator.cs
hazzik: Done

src/NHibernate.Test/MappingTest/NonReflectiveBinderFixture.cs
```diff
@@ -194,5 +195,6 @@ public void XmlSerialization()
 				Assert.That(mapping, Is.XmlSerializable);
 			}
 		}
+#endif
```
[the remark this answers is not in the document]
hazzik: Done

src/NHibernate.Test/NHAssert.cs
```diff
@@ -53,7 +53,11 @@ public static void IsSerializable(object obj)
 
 		public static void IsSerializable(object obj, string message, params object[] args)
 		{
+#if NET461
 			Assert.That(obj, Is.BinarySerializable, message, args);
+#else
+			//TODO:
+#endif
```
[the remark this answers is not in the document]
hazzik: Implemented the check (will push shortly).

hazzik:
@fredericDelaporte I've fixed most of the concerns, could you please take another look?

src/NHibernate.Test/DynamicProxyTests/LazyFieldInterceptorTests.cs
```diff
@@ -26,6 +26,7 @@ public void LazyFieldInterceptorMarkedAsSerializable()
 			Assert.That(typeof(DefaultDynamicLazyFieldInterceptor), Has.Attribute<SerializableAttribute>());
 		}
 
+#if NET461
 		[Test]
 		public void LazyFieldInterceptorIsBinarySerializable()
```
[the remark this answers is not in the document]
hazzik: Fixed

fredericDelaporte:
Yes but it may take a while, I am not at home and currently spend most of my (awaken) time outside.

src/NHibernate.Test/Async/NHSpecificTest/NH3583/AutoFlushFixture.cs
```diff
@@ -61,9 +61,12 @@ protected override void OnTearDown()
 				Assert.That(result.Count, Is.EqualTo(1));
 			}
 		}
+
 		[Test]
 		public async Task ShouldAutoFlushWhenInDistributedTransactionAsync()
 		{
+			Assume.That(Sfi.ConnectionProvider.Driver.SupportsSystemTransactions);
```
fredericDelaporte: There is a discrepancy here with the sync version, which gets changed on async regen.

.travis.yml
```diff
@@ -0,0 +1,49 @@
...
+  - postgresql
+  - docker
+env:
+  - DB=SqlServer2008 CONNECTION_STRING="Server=localhost;initial catalog=nhibernate;User Id=sa;Password=P@ssw0rd;"
+  - DB=PostgreSQL CONNECTION_STRING="Host=localhost;Port=5432;Username=postgres;Database=nhibernate;Enlist=true;"
+  - DB=Firebird 
+  - DB=MySQL CONNECTION_STRING="Server=127.0.0.1;Uid=root;Database=nhibernate;Old Guids=True;"   
+matrix:
+  allow_failures:
+    - env: DB=MySQL CONNECTION_STRING="Server=127.0.0.1;Uid=root;Database=nhibernate;Old Guids=True;"
```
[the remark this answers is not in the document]
hazzik: This bug: https://bugs.mysql.com/?id=89639&edit=2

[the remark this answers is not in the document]
hazzik: > It should ignore only known failing tests.

I don’t have time nor desire to port the test result comparator at the moment. We would need to extend it to support different tests results for different platform-framework combination. I think it’s fair complex task and needs to be done (if at all) in a different pull request.

src/NHibernate.Test/TestsContext.cs
```diff
@@ -0,0 +1,66 @@
...
+#if NETCOREAPP2_0
+	[SetUpFixture]
+#endif
+	public class TestsContext
+	{
+		public static bool ExecutingWithVsTest { get; } =
+			System.Reflection.Assembly.GetEntryAssembly()?.GetName().Name == "testhost";
+
+		public static void AssumeSystemTypeIsSerializable() =>
+			Assume.That(typeof(System.Type).IsSerializable, Is.True);
```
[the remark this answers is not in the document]
hazzik: Ok, I’ll change this to `Ignore` instead.

appveyor.yml
```diff
@@ -0,0 +1,60 @@
...
+      'PostgreSQL' {
+        # Enable prepared transactions
+        Add-Content -Path 'C:\Program Files\PostgreSQL\10\data\postgresql.conf' -Value "`r`nmax_prepared_transactions = 100"
+        Start-Service 'postgresql-x64-10'
+      }
+      'SqlServer2008' { Start-Service 'MSSQL$SQL2017' }
+      'SqlServer2012' { Start-Service 'MSSQL$SQL2017' }
+    }
+test_script:
+- cmd: powershell -noprofile -command "& ./build.ps1 -TaskList Set-Configuration,Test -properties @{\"Database\" = \"%DB%\";\"ConnectionString\"=\"%CONNECTION_STRING%\"}"
```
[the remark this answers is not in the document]
hazzik: Travis does not support test results at all. It has an ability to upload artifacts to a third party storage provider (AWS).

src/NHibernate.Test/Async/NHSpecificTest/NH3583/AutoFlushFixture.cs
hazzik: Fixed

appveyor.yml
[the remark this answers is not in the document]
hazzik: I've implemented the upload test results to AppVeyor. However they are still unusable. As AppVeyor just hangs when you try to open "Tests" tab.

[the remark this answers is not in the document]
ngbrown: If `dotnet test` is used then the real-time reporting of test results to AppVeyor can use `Appveyor.TestLogger` (<https://github.com/Faizan2304/LoggerExtensions>).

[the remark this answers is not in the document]
fredericDelaporte: > I've implemented the upload test results to AppVeyor. However they are still unusable. As AppVeyor just hangs when you try to open "Tests" tab (at least on my machine).

Or maybe at least with Chrome. I am used to AppVeyor causing Chrome to hang. But Firefox (at least since its "Quantum" version) seems to better support AppVeyor pages when they have a lot of content. Currently I have opened the test tab with Chrome and Firefox, Firefox did it faster but Chrome finally did it too.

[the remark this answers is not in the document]
fredericDelaporte: > If dotnet test is used then the real-time reporting of test results to AppVeyor can use Appveyor.TestLogger (https://github.com/Faizan2304/LoggerExtensions).

Would not this require tweaking the platform agnostic build.ps1 script for AppVeyor? I guessed that adding an AppVeyor specific script afterward is then more compliant with Alexander's goals than handling some AppVeyor specific case for test reporting inside the general build/test script.

.travis.yml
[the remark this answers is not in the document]
fredericDelaporte: The result comparer should not be used again ideally. I was thinking of additional ignore rules for those failing test. Something like `if !NETFX \n if (MySql) ignore` could do it, but that is maybe not very satisfactory since it seems more related to MySql data driver than to the fact of not running under NETFX.

So well, ok for letting this as is for now.

src/NHibernate.Test/SystemTransactions/TransactionNotificationFixture.cs
```diff
@@ -164,11 +164,12 @@ public void ShouldNotifyAfterDistributedTransaction(bool doCommit)
 			Assert.That(interceptor.afterTransactionCompletionCalled, Is.EqualTo(2));
 		}
 
-
 		[Description("NH2128")]
 		[Theory]
 		public void ShouldNotifyAfterDistributedTransactionWithOwnConnection(bool doCommit)
 		{
+			Assume.That(Sfi.ConnectionProvider.Driver.SupportsSystemTransactions, Is.True);
```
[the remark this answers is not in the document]
fredericDelaporte: We still have in AppVeyor log near its end:
```
3) Failed : NHibernate.Test.SystemTransactions.TransactionNotificationFixture.ShouldNotifyAfterDistributedTransactionWithOwnConnection
All test cases were inconclusive
4) Failed : NHibernate.Test.SystemTransactions.TransactionWithoutConnectionFromPrepareNotificationFixture.ShouldNotifyAfterDistributedTransactionWithOwnConnection
All test cases were inconclusive
```

But those tests are not reported as errors in the tests tabs. So in fact they could stay this way, using `Assume`, without actually causing the test suit to fail. Firebird/SQL Server/PostgreSQL builds allow to ascertain this: their log also reports those two tests as failed, without causing the build to fail.

But it is a bit confusing to see them succeeded while the log say:
```
Test Run Summary
  Overall result: Failed
```

appveyor.yml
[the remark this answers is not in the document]
hazzik: `dotnet test` is broken due to the bug with `GetEntryAssembly`.

src/NHibernate.Test/SystemTransactions/TransactionNotificationFixture.cs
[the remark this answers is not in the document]
hazzik: Ok, ignored this test.

stevenlauwers22:
Awesome! Looking forward to the release of 5.1!

src/NHibernate/NHibernate.csproj
```diff
@@ -41,12 +45,26 @@
...
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
     <Reference Include="System.ServiceModel" />
     <Reference Include="System.Transactions" />
     <Reference Include="System.Configuration" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
```
ngbrown: It seems that building the core library for netcoreapp2.0 is unnecessary.

Is leaving off the `System.Reflection.Emit` package references worth it for the increased bulk in the NuGet package?

cloviscoli:
Any forecast when we will have a release?

stephan-simondshealth-com-au:
1) I'm also hoping this will make it into a release soon. IMHO supporting a newer version of .NET Core (Even if its not fully) would justify a release. But I'm probably biased :-)

2) In the mean time, are there overnight NuGet package that I can pull?
(For instance: Is there an AppVeyor project feed?) Thanks !

fredericDelaporte:
> I'm also hoping this will make it into a release soon. IMHO supporting a newer version of .NET Core (Even if its not fully) would justify a release. But I'm probably biased :-)

So do I, but some pending PR deserve being merged in 5.1. Of course at some point those taking too long to finalize may be postpone for a later version.

> In the mean time, are there overnight NuGet package that I can pull?

The CI builds include a "NHibernate (Release Package)" TeamCity build, having in its artifact NuGet packages, the binary release zip and more. Please note that they are still versioned according to latest release: we increment the version only when actually performing a release.

You can take them from master builds or from a specific PR, as you wish. Access the latest master builds from the [commits page](https://github.com/nhibernate/nhibernate-core/commits/master). Clicking on the green check (GitHub desktop site version, likely not available on mobile version) should show you the builds list. Then navigate on "details" link of the release build (like [this one](https://teamcity.jetbrains.com/viewLog.html?buildId=1340090&buildTypeId=NHibernate_NHibernateReleasePackage)). TeamCity will ask you to login if you are not already. An anonymous logon option is available. Then go on the artifacts tab ([here by example](https://teamcity.jetbrains.com/viewLog.html?buildId=1340090&buildTypeId=NHibernate_NHibernateReleasePackage&tab=artifacts)) and download what you need.

fredericDelaporte:
About:

> - MySQL on 
>    - Windows - 4 tests are failing (same tests fail on Full .NET Framework on Windows with 6.10.x provider, 2 of them due to the bug in MySql data adapter);

Has written later by Alexander, "the bug in MySql data adapter" is [here](https://bugs.mysql.com/?id=89639).

Now the two others are also due to a bug in MySql data adapter 6.10, now reported [here](https://bugs.mysql.com/bug.php?id=90086). (It is closely related to another [one](https://bugs.mysql.com/bug.php?id=89159) and may be considered a duplicate.)

This other bug is due to MySql.Data 6.10 command disposal closing now any open data reader found on the connection, be it related to the command or not. So a workaround would be to close previously opened command before opening a data reader with a new command. But in the case of the failing test, this would need introducing more coupling, because the current sequence of events implies a `JoinedEnumerable` of `EnumerableImpl` holding `NHybridDataReader`, which are for all of them but the last read into memory by the batcher at a place where their matching command is not known (and anyway, that is not the batcher responsibility to close commands here). The `JoinedEnumerable` disposes of its current `EnumerableImpl` as soon as it finishes iterating over it and before starting iterating the next one. This dispose causes the current `EnumerableImpl` to dispose of the command it is bound to, and when it is a MySql command, this command then dispose any reader found on the connection, resulting in closing the reader of the last `EnumerableImpl` in the `JoinedEnumerable`.

So another workaround could be to change `JoinedEnumerable` for ceasing disposing an `EnumerableImpl` when it has exhausted it, but instead disposing all of them when it has exhausted all of them. But well, I would rather wait for MySql Connector to get fixed.

bgrainger:
> I would rather wait for MySql Connector to get fixed.

This may be a long wait...

I have created a competing OSS library, https://github.com/mysql-net/MySqlConnector, that is a drop-in replacement for Oracle's Connector/NET (aka MySql.Data). (One of the motivating factors in creating this alternative was to fix longstanding Connector/NET bugs that weren't getting addressed. It also supported .NET Standard before Connector/NET did.)

I understand that many NHibernate users are using Connector/NET, so it makes sense to test with the library your users are using; however, if you'd like a higher-quality MySQL ADO.NET library then I suggest testing with MySqlConnector (either in addition to or instead of Connector/NET). (As the primary author of MySqlConnector, I'd be happy to assist with any integration questions.)

Issue #1530, opened 2018-01-12
Merging a bidirectonal list creates unecessary UPDATE statement


RogerKratz:
Having following mapping...

```
<class name="Parent">
	<id name="Id" type="guid">
	  <generator class="guid.comb" />
	</id>
	<list name="Children" cascade="all-delete-orphan" inverse="true">
		<key column="Parent" not-null="true"/>
		<list-index column="OrderIndex"  />
		<one-to-many class="Child"/>
	</list>
	<property name="Name" />
</class>

<class name="Child">
	<id name="Id" type="guid">
	  <generator class="guid.comb" />
	</id>
	<many-to-one name="Parent" class="Parent" />
	<property name="OrderIndex" access="readonly"/>
	<property name="Name" />	
 </class>
```

...will cause INSERT + UPDATE when adding a new child and doing session.Merge(parent).

Will add a PR soon.

PR #1531, opened 2018-01-12
Test for Merging a bidirectonal list creates unecessary UPDATE statement

RogerKratz:
Test for #1530.


RogerKratz:
Could get my tests green without any code changes, just changed the mapping. However - it was far from obvious how to get this right... Still think "normal" biref should work. Anyhow, to make this use case work, I had to change the mapping as follows...

* collection owns FK (inverse=false)
* Removing OrderIndex on child
* Set update/insert=false on child's parent mapping
* Set parent's list's key/parent to update=false <-- This took me some time to find out

Don't know if...
* this PR/issue should be deleted
* docs should be updated with above info
* this PR with old, "normal" mapping should work so the bug/issue is still valid

fredericDelaporte:
> Still think "normal" biref should work.

Well it works if I have understood all this well, but in a sub-optimal manner.

> - docs should be updated with above info

Why not, [doc](http://nhibernate.info/doc/nhibernate-reference/index.html) sources are in doc\reference\modules, if you see a proper way of documenting this.

> - this PR with old, "normal" mapping should work so the bug/issue is still valid

Ideally it should work in an optimal way (a single insert rather than insert then update), but if this implies too complicated code for achieving a single insert, I will not consider it is worth it. Especially since mapping with a `list` is not a common place thing I think, because in my experience it causes more issues than it resolves. (I have always ended up handling my indexes manually for setting their value, while using an `order-by` in my collection mapping for sorting according to them.)

hazzik:
Not sure what to do with this tests :(

RogerKratz:
Just delete them in their present form. 
What would be really good if they were green using "normal" bidirectional mapping. So maybe keep them as failing test with "normal" mapping instead?

fredericDelaporte:
Thinking about it again, should a list mapped as inverse even be considered a "normal" thing? Handling indexes looks more as the responsibility of the list. At least, the list is always concerned by the indexes, while children may ignore them completely.

RogerKratz:
Sorry, didn't mean it was a "normal thing to do", but rather "the normal way of mapping a biref collection is using inverse=true".
That being said, feel free to just drop this PR.

fredericDelaporte:
Closing this PR should be done if #1530 is closed too without being done. I do not think it should be closed.

So I have updated this PR for including again the failing case, along with the mapping not having the issue. The failing case is marked as known bug for allowing merging this.

I have also moved the files to a folder named `GH1530`, as it is the preferred naming under `NHSpecificTest`.

PR #1532, opened 2018-01-15
Fix ExpressionKeyVisitor to produce unique keys for anon types

### LINKED ISSUE #1526: ExpressionKeyVisitor does not produce unique keys for anonymous types coming from different assemblies
The `Linq.Visitors.ExpressionKeyVisitor.Visit()` method does not guarantee a unique key generation for expressions that are based on compiler generated anonymous types of same shape, but coming from different assemblies.

Consider this code:
```csharp
session.Query<MyEntity>().Select(e => new { e.Name }).ToList();
```
Upon compilation, the compiler will generate an anonymous type into the assembly where the query code is located. The type will be named something like (simplified) ``AnonymousTypeX`1[[System.String]]`` (where _X_ is a sequential number within the assembly) and will be placed in global namespace (that is, will have no namespace at all). It will also be `internal` to that assembly.

Now, if the same query is also placed in code in different assembly within same app, the compiler will generate same-shaped anonymous type in that other assembly. If it happens that the sequential _X_ in that assembly aligns with the _X_ in the first assembly, both anonymous types will have the same resulting names. When inspected via `System.Type.FullName` they will be presented as exactly the same, because they both don't have any namespace to differentiate them. (N.B., this can also happen with user defined, same-named internal types that are in separate assemblies, but are declared in the same namespace).

Unfortunately, the `ExpressionKeyVisitor` uses `.FullName` when processing generic methods in its `.VisitMethod()`, like:
https://github.com/nhibernate/nhibernate-core/blob/49b998d2afc17de37337b2830e52a53bfd94ed42/src/NHibernate/Linq/Visitors/ExpressionKeyVisitor.cs#L235-L244
This would produce (simplified) ``ToList[AnonymousType42`1[[System.String]]`` (given _X_ = 42) for both anon types. In effect this will cause the generated query plan cache keys to be the same for two essentially different types, and will rightfully blow in runtime.

The quick solution is to use `.AssemblyQualifiedName` here, which will properly distinguish both types (downside being slightly higher memory usage for cache keys, because of longer type identifiers).

I'm able to PR this change, along with tests - albeit a bit unorthodox ones, as they need to simulate internal anonymous types from two assemblies, but they do work.

mkasperski:
Fixes #1526


hazzik:
I think this fix needs to be applied to all types in the ExpressionKeyVisitor and not only to `MethodCall`s

mkasperski:
I was gonna suggest (after fixing this case) even wider approach, in which any occurence of `FullName` is inspected to see whether its usage is based on the same false assumption demonstrated in this PR ("equal `FullName` implies equal type"). I'm not that experienced in NH codebase though to analyze each case, so with this PR I just wanted to "plant a seed of suspicion" and present a setup in which such cases could possibly be tested.

As for more cases inside `ExpressionKeyVisitor`, I did notice other visiting methods that use type metadata, I just didn't have a proof that they are vulnerable or even possible to trigger (an unary expression on anonymous type?). Let me clean up this PR after current review and then I'll see if I can find more linq queries that still produce colliding cache keys. If you have suggestions for them already, that would be very helpful.

src/NHibernate.Test/NHSpecificTest/GH1526/Fixture.cs
```diff
@@ -0,0 +1,23 @@
...
+	public class Fixture
+	{
+		[Test]
+		public void ShouldCreateDifferentKeyForAnonymousTypesFromDifferentAssemblies()
+		{
+			var exp1 = AnonymousTypeExpressionProviderFromNHibernateTestAssembly.GetExpression();
+			var exp2 = AnonymousTypeExpressionProviderFromNHibernateDomainModelAssembly.GetExpression();
+
+			var key1 = ExpressionKeyVisitor.Visit(exp1, null);
+			var key2 = ExpressionKeyVisitor.Visit(exp2, null);
```
fredericDelaporte: Better provide a non-null second argument(`new Dictionary<ConstantExpression, NamedParameter>()`). Although it lacks some pre-condition checks, `ExpressionKeyVisitor` does not actually support receiving it as `null`, and will fail on some expressions if it was `null`.

mkasperski: I've followed with your remark in the new commit. However, I believe the real change should be in `ExpressionKeyVisitor` itself, to either a.) make the second arg optional, or b.) default to `new Dictionary<...>` upon receiving `null` in `Visit`/`ctor`.

a.) would be cleaner, although would be a breaking change, since the type is public (does it have to be public, though?)
b.) would be the opposite: non-breaking, but leaving the method's api somewhat dirty

The method is called more times with `null` than with real dictionary, so some kind of cleanup there seems due. Separate PR, though, I reckon.

fredericDelaporte: > The method is called more times with null than with real dictionary,

Indeed. Not a good thing since it may then throw a null ref exception.

I am not for b., this just hides the trouble. Well, a. maybe too. I would rather instead, if some change is done (but it is not the subject of this PR), throw an `InvalidOperationException` in `VisitConstant` if `_constantToParameterMap` is `null`.

fredericDelaporte: #1536 added for this subject.

PR #1533, opened 2018-01-15
Add query support for the static methods of System.Decimal

### LINKED ISSUE #831: NH-3515 - Support for Decimal.Round, Decimal.Ceiling, Decimal.Floor and other static methods of Decimal class
**Alexander Zaytsev** created an [issue](https://nhibernate.jira.com/browse/NH-3515) — <time datetime="2013-08-21T20:45:50.763Z" title="2013-08-21T20:45:50.763Z">21st August 2013, 20:45:50</time>:
> Full list of methods available here: http://msdn.microsoft.com/ru-ru/library/System.Decimal_methods.aspx
>

weelink:
Fixes #831

- [x] Firebird
- [x] MySql
- [x] Oracle Managed 32
- [x] PostgreSQL
- [x] SQLite
- [x] SQL Server
- [x] SQL Server Compact


weelink:
Since this is my first contribution and the change was too easy, could someone inspect this PR more than usual?

src/NHibernate/Linq/Functions/DecimalGenerator.cs
```diff
@@ -0,0 +1,43 @@
...
+				ReflectHelper.GetMethodDefinition(() => decimal.Divide(default(decimal), default(decimal))),
+				ReflectHelper.GetMethodDefinition(() => decimal.Equals(default(decimal), default(decimal))),
+				ReflectHelper.GetMethodDefinition(() => decimal.Floor(default(decimal))),
+				ReflectHelper.GetMethodDefinition(() => decimal.Multiply(default(decimal), default(decimal))),
+				ReflectHelper.GetMethodDefinition(() => decimal.Negate(default(decimal))),
+				ReflectHelper.GetMethodDefinition(() => decimal.Remainder(default(decimal), default(decimal))),
+				ReflectHelper.GetMethodDefinition(() => decimal.Round(default(decimal))),
+				ReflectHelper.GetMethodDefinition(() => decimal.Round(default(decimal), default(int))),
+				ReflectHelper.GetMethodDefinition(() => decimal.Round(default(decimal), default(MidpointRounding))),
+				ReflectHelper.GetMethodDefinition(() => decimal.Round(default(decimal), default(int), default(MidpointRounding))),
```
fredericDelaporte: Is `MidpointRounding` argument actually supported? It looks likely to me it is in fact ignored. In such case, better do as the `Math` generator do: do not support overloads with this argument.

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		[Test]
+		public void CanHandleSubtract()
+		{
+			CanHandle(e => decimal.Subtract(e.Value, 1m) > 1m);
+			CanHandle(e => decimal.Subtract(2m, e.Value) > 1m);
+		}
+
+		private void CanHandle(Func<Entity, bool> predicate)
```
hazzik: Shouldn't this be `Expression<Func<Entity, bool>>`?

weelink: It should. I'll try to make it work with an Expression

src/NHibernate/Linq/Functions/DecimalGenerator.cs
weelink: Probably not. I'll remove this option. It worked in the tests because I was passing a Func<> and not an Expression<>.

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,152 @@
...
+		public void CanHandleCompare()
+		{
+			CanHandle(e => decimal.Compare(e.EntityValue, 1.5m) < 1);
+			CanHandle(e => decimal.Compare(1.0m, e.EntityValue) < 1);
+		}
+
+		[Test]
+		public void CanHandleDivide()
+		{
+			CanHandle(e => decimal.Divide(e.EntityValue, 1.25m) < 1);
```
fredericDelaporte: Some databases (Firebird) do not support divide/multiply on decimal. So a new property on `TestDialect` is needed, like `SupportsMultiplicationDecimal`, `true` by default, and overridden to `false` in `FirebirdTestDialect`.

Then divide and multiply tests should `Assume.That(TestDialect.SupportsMultiplicationDecimal, Is.True);`

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,152 @@
...
+
+				session.Flush();
+				transaction.Commit();
+			}
+		}
+
+		[Test]
+		public void CanHandleAdd()
+		{
+			CanHandle(e => decimal.Add(e.EntityValue, 2) > 3.0m);
```
fredericDelaporte: Its failure on Firebird seems to me due to some Firebird limitation. It looks like the operation on the column is enough to cause Firebird to lose track of the data type.
Maybe a workaround could be to add a cast, but I think we should just ignore that Firebird limitation with something like:
```c#
Assume.That(TestDialect.SupportsNonDataBoundCondition, Is.True);
```

Granted, here the condition is data bound, but the addition (or subtraction) on the data causes the same trouble. And I lack an idea for a new `TestDialect` property name for such a case.

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,152 @@
...
+			CanHandle(e => decimal.Add(2, e.EntityValue) > 3.0m);
+		}
+
+		[Test]
+		public void CanHandleCeiling()
+		{
+			CanHandle(e => decimal.Ceiling(e.EntityValue) > 1.0m);
```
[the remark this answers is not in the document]
weelink: I noticed this as well and wanted to create a new bug for it, but you beat me to it

weelink:
I need to investigate the problems with SQLite a bit more. The results of executing the queries in the test are not consistent with what I see if I run the queries myself in a SQLite client. E.g. 'ceiling' is not supported if I run the query in the client, but I get no exception in the test. And the `CanHandleAdd`-test, that is doing a simple `where (EntityValue + x) > y`, returns no result.

There might be something strange going on.

fredericDelaporte:
If some function are not supported and has to be ignored, something like 
```c#
Assume.That(Dialect.Functions.ContainsKey("ceiling"), Is.True, Dialect + " doesn't support ceiling function.");
```
should take care of it. Of course this would also ignore the trouble for Oracle which only registers `ceil`. But since this is not the subject of this PR to fix registered functions, it is ok. (The SQL ANSI standard supports both `ceil` and `ceiling`. Many db implements both (MySql, Postgres, Firebird, ...), some implements only one (Oracle, SQL Server).)

weelink:
It seems that NHibernate and/or SQLite are in a disagreement about how to interpret the return type of the methods. The following does not work:

    session.Query<Entity>()
        .Where(e => decimal.Add(e.EntityValue, 2) > 3.0m); // Results in an empty list, but several are expected

Adding an explicit `Convert.ToDecimal` does return the correct results:

    session.Query<Entity>()
        .Where(e => Convert.ToDecimal(decimal.Add(e.EntityValue, 2)) > 3.0m) // Results in the expected list

This will be translated into `cast(entity0_.EntityValue+@p0 as NUMERIC)`. This is not necessary if I execute the query manually in a client, so I don't think the problem is in SQLite.

@fredericDelaporte Do you have a possible direction for a solution? I'm not familiar with the codebase enough to figure it out on my own.

fredericDelaporte:
> This is not necessary if I execute the query manually in a client, so I don't think the problem is in SQLite.

Executing the query manually is probably done without using `System.Data.SQLite`, so the trouble could be there too.

Another cause may be a kind of parameter type mismatch with the resulting type from addition. So this may be dodged by using `.MappedAs(NHibernateUtil.Double)` or some other type on `3.0m`. Or maybe by changing the decimal type registration in `SQLiteDialect`. It is currently `numeric`, but you could try with `real` which is the actual type used by SQLite. (Well, `numeric` can also result into `integer` type being used, if the value is indeed an integer.)
(The SQLite typing system looks quite messy to me. It as only five "in-memory" types (see [data types](https://www.sqlite.org/datatype3.html), but can store them on disk with many more data-types, for which I am not finding an official documentation.)

By the way, have you tried the query with an explicitly decimal literal? `decimal.Add(e.EntityValue, 2m)`

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,166 @@
...
+		}
+
+		[Test]
+		public void CanHandleSubtract()
+		{
+			CanHandle(e => decimal.Subtract(e.EntityValue, 1m) > 1m);
+			CanHandle(e => decimal.Subtract(2m, e.EntityValue) > 1m);
+		}
+
+		private void CanHandle(Expression<Func<Entity, bool>> predicate)
```
fredericDelaporte: A more complete test would be to not only test filtering (what does this `CanHandle`, which could be renamed `CanFilter`) but also test selecting.

Can you add a `CanSelect` test ?

weelink: Good idea. Will add it shortly

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,166 @@
...
+			using (ITransaction transaction = session.BeginTransaction())
+			{
+				session.Delete("from System.Object");
+
+				session.Flush();
+				transaction.Commit();
+			}
+		}
+
+		private void IgnoreIfNotSupported(string function)
```
fredericDelaporte: With the merge of #1535 in master, this helper should be removed for using `AssumeFunctionSupported` (inherited from `TestCase`) instead.

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
weelink: I've added a `CanSelect` test.

It turns out that the return types are different. The return type of the `mod` function is `Int32`, but `decimal.Remainder` returns a `decimal`. The `IQueryTranslator` correctly returns a collection of type `int`, and then it crashes because it wants to add the (`int`) results to the `decimal` collection in `NHibernate.Engine.Query.HQLQueryPlan` line 149.

To make the test pass for SQL Server, I registered a new `mod` function to the dialects. But it is a temporary hack that I'll remove later when I figured out how deep the rabbit hole goes.

weelink: From what I understand from the documentation of SQLite, it doesn't have a type system like other databases do. It determines the type of the value by looking at the value, not the column definition. It ignores the declared type.

Changing the parameter to a decimal explicitly (i.e. `3m`) doesn't help, but  `.MappedAs(NHibernateUtil.Double)` solves the problem.

weelink:
I've added a `CanSelect` test.

It turns out that the return types are different. The return type of the `mod` function is `Int32`, but `decimal.Remainder` returns a `decimal`. The `IQueryTranslator` correctly returns a collection of type `int`, and then it crashes because it wants to add the (`int`) results to the `decimal` collection in `NHibernate.Engine.Query.HQLQueryPlan` line 149.

To make the test pass for SQL Server, I registered a new `mod` function to the dialects. But it is a temporary hack that I'll remove later when I figured out how deep the rabbit hole goes.

From what I understand from the documentation of SQLite, it doesn't have a type system like other databases do. It determines the type of the value by looking at the value, not the column definition. It ignores the declared type.

Changing the parameter to a decimal explicitly (i.e. `3m`) doesn't help, but  `.MappedAs(NHibernateUtil.Double)` solves the problem.

fredericDelaporte:
> It turns out that the return types are different. The return type of the `mod` function is `Int32`, but `decimal.Remainder` returns a `decimal`.

So here we have a semantic mismatch to investigate. In .Net runtime, executing `decimal.Remainder(5.2m, 5.1m)` yields `0.1m`. So using a function yielding only `int` to emulate this cannot yield the same result.

Should we accept such a mismatch? Well, if no database implements a decimal remainder yielding non integer results, it would be moot to not accept it. SQL Server seems to yield only integer results by example.

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		private void CanSelect(Expression<Func<Entity, decimal>> predicate, IEnumerable<decimal> expected)
+		{
+			using (ISession session = OpenSession())
+			using (session.BeginTransaction())
+			{
+				IEnumerable<decimal> inSession = session.Query<Entity>().Select(predicate).ToList();
+
+				Assert.That(inSession.OrderBy(x => x), Is.EqualTo(expected.OrderBy(x => x)).Within(0.001M));
```
fredericDelaporte: I would rather avoid the order by on result values, which may hide a trouble (granted, unlikely). Instead, the `inSession` query could be completed by an `.OrderBy(e => e.EntityValue)` before the `Select`, which should ensure the db results have been computed on a list having the same ordering than the in-memory one.

src/NHibernate/Linq/Functions/DecimalGenerator.cs
```diff
@@ -0,0 +1,71 @@
...
+					return treeBuilder.Equality(expressions[0], expressions[1]);
+				case "negate":
+					return treeBuilder.Negate(expressions[0]);
+				case "compare":
+					return treeBuilder.MethodCall("sign", treeBuilder.Subtract(expressions[0], expressions[1]));
+				case "multiply":
+					return treeBuilder.Multiply(expressions[0], expressions[1]);
+				case "remainder":
+					HqlMethodCall mod = treeBuilder.MethodCall("mod", expressions[0], expressions[1]);
+					return treeBuilder.Cast(mod, typeof(decimal));
```
fredericDelaporte: Clearly a better solution. Now is it needed to have the cast in the resulting SQL query? If no, you could use a `TransparentCast` instead, which is just a hint for the HQL engine, and does not go into the generated SQL query. (Well, for SQLite, it does still perform the cast in SQL too, because that db has special needs.)

weelink:
I don't like it that `TransparentCast` is necessary twice here. It feels like a hack: a generic solution for a specific problem. You would like it to be clear somehow that we need this because a completely different part of the code will fail for a particular database.

Of course a comment can be added with the explaination, but I consider that plan B. I prefer a better solution. I will need to think about this.

fredericDelaporte:
The PostgreSQL `round` issue has its roots in its registration, in `PostgreSQLDialect`. It registers `round` as a function yielding always `Double`, thus the issue with decimal. Other dialects do not specify a type, which causes HQL to consider it yields out the same type that the one it has as input.

PostgreSQL seems to be a special case, since its `round` registration is not a `StandardSQLFunction` but a templated one. `SQLFunctionTemplate` does not take an overload without a type specification. I guess that is why the coder have decided to put `Double` here. But it can still accept having no type specified, by supplying `null` instead. I think this should be the correct fix.

(About why PostgreSQL `round` needs to be templated: see its [documentation](https://www.postgresql.org/docs/10/static/functions-math.html). Its two arguments `round` only accepts decimal input, so the template is here for converting non-decimal like floats to decimal.)

hazzik:
>The PostgreSQL round issue has its roots in its registration, in PostgreSQLDialect. It registers round as a function yielding always Double, thus the issue with decimal. 

This is a bug. Acording to the docs:

round(dp or numeric)	
(same as input)
round to nearest integer


round(v numeric, s int)
numeric
round to s decimal places

fredericDelaporte:
> This is a bug.

Yes, it should be fixed within another PR. But it can still be tested here.

fredericDelaporte:
> Yes, it should be fixed within another PR.

Done: #1562.

src/NHibernate/Linq/Functions/DecimalGenerator.cs
```diff
@@ -0,0 +1,72 @@
...
+					return treeBuilder.TransparentCast(mod, typeof(decimal));
+				case "round":
+					HqlExpression numberOfDecimals = (arguments.Count == 2) ? expressions[1] : treeBuilder.Constant(0);
+					HqlMethodCall round = treeBuilder.MethodCall("round", expressions[0], numberOfDecimals);
+					return treeBuilder.TransparentCast(round, typeof(decimal));
+			}
+
+			if (arguments.Count == 2)
+			{
+				return treeBuilder.MethodCall(function, expressions[0], expressions[1]);
```
hazzik: There is no functions which accepts 2 arguments anymore. Only functions left after the switch above is "ceiling" and "floor", both accept single argument.

src/NHibernate/Linq/Functions/DecimalGenerator.cs
```diff
@@ -0,0 +1,72 @@
...
+			};
+		}
+
+		public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject, ReadOnlyCollection<Expression> arguments, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor)
+		{
+			string function = method.Name.ToLowerInvariant();
+
+			HqlExpression[] expressions = arguments.Select(x => visitor.Visit(x).AsExpression()).ToArray();
+
+			switch (function)
```
hazzik: It seems that it would be better to split the class into different builders.

fredericDelaporte:
#1562 is now merged. Updating this branch should allow removing the `TransaparentCast` put for `round` case.

src/NHibernate/Linq/Functions/DecimalGenerator.cs
weelink: Correct. After everything works I need to clean it all up.

src/NHibernate/Linq/Functions/DecimalGenerator.cs
```diff
@@ -0,0 +1,188 @@
...
+		{
+			SupportedMethods = new[]
+			{
+				ReflectHelper.GetMethodDefinition(() => decimal.Divide(default(decimal), default(decimal)))
+			};
+		}
+
+		public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject, ReadOnlyCollection<Expression> arguments, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor)
+		{
+			return treeBuilder.TransparentCast(treeBuilder.Divide(visitor.Visit(arguments[0]).AsExpression(), visitor.Visit(arguments[1]).AsExpression()), typeof(decimal));
```
fredericDelaporte: Putting `Cast` here fix Oracle failure. So long for avoiding having it in SQL for the other DBs which do not need it, provided they support having the cast.

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,236 @@
...
+
+			CanSelect(predicate, inMemory);
+		}
+
+		private void CanSelect(Expression<Func<Entity, decimal>> predicate, IEnumerable<decimal> expected)
+		{
+			using (ISession session = OpenSession())
+			using (session.BeginTransaction())
+			{
+				IEnumerable<decimal> inSession = session.Query<Entity>().Select(predicate).ToList();
```
fredericDelaporte: For the `Assert.Multiple` to be more effective, this line would need to be change to something like:
```c#
IEnumerable<decimal> inSession = null;
Assert.That(() => inSession = session.Query<Entity>().Select(predicate).ToList(), Throws.Nothing);
```

(And use `inSession?.OrderBy(x => x)` below, or as I was suggesting previously, change the test to:
```c#

IEnumerable<decimal> inSession = null;
Assert.That(() => inSession = session.Query<Entity>().OrderBy(e => e.EntityValue).Select(predicate).ToList(), Throws.Nothing);

Assert.That(inSession, Is.EqualTo(expected).Within(0.001M));
```
)

weelink: I'll change this soon. It took a month, but I think I'm nearly there.

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,237 @@
...
+
+		private void CanSelect(Expression<Func<Entity, decimal>> predicate, IEnumerable<decimal> expected)
+		{
+			using (ISession session = OpenSession())
+			using (session.BeginTransaction())
+			{
+				IEnumerable<decimal> inSession = null;
+				Assert.That(() => inSession = session.Query<Entity>().OrderBy(e => e.EntityValue).Select(predicate).ToList(), Throws.Nothing);
+
+				Assert.That(inSession, Is.EqualTo(expected.OrderBy(x => x)).Within(0.001M));
```
fredericDelaporte: The order by of `expected.OrderBy(x => x)` must be removed. It is ordering by computed value instead of preserving the ordering by original value the in memory source has.

src/NHibernate.Test/NHSpecificTest/GH0831/FixtureByCode.cs
```diff
@@ -0,0 +1,237 @@
...
+				IEnumerable<Entity> inMemory = entities.Where(predicate.Compile()).ToList();
+				IEnumerable<Entity> inSession = session.Query<Entity>().Where(predicate).ToList();
+
+				CollectionAssert.AreEquivalent(inMemory, inSession);
+			}
+		}
+
+		private void CanSelect(Expression<Func<Entity, decimal>> predicate)
+		{
+			IEnumerable<decimal> inMemory = entities.Select(predicate.Compile()).ToList();
```
fredericDelaporte: If you wish to keep on explicitly ordering the in-memory case, add the ordering here:
`entities.OrderBy(e => e.EntityValue).Select(predicate.Compile()).ToList()`.

weelink: No. I would like to get rid of it entirely. It is noise. It should be replaced with an assertion that uses a tolerance when comparing values and where the order does not matter. As far as I can find, NUnit doesn't have this. It can either assert with a tolerance, or where the order doesn't matter, not both.

I noticed that `NHibernate.Test` already has a `ObjectAssertion`, `NHAssert` and `SubclassAssert`. I can add one for collections.

hazzik:
Added one minor adjustment.

weelink:
Thanks for all the help

hazzik:
@fredericDelaporte if my changes are OK. I'll merge the PR in a couple of hours.

weelink:
I completely forgot about Decimal.Truncate. I'll try to make it work.

weelink:
The SQL Server definition can be solved by using the `VarArgsSQLFunction`. Not very happy with this solution, but I'm not seeing a different possibility.
We could consider adding something that can handle these situations better. But that can be done in a separate PR.

hazzik:
> The SQL Server definition can be solved by using the VarArgsSQLFunction. Not very happy with this solution, but I'm not seeing a different possibility.

`SQLFunctionTempate` works better in that situation.

src/NHibernate/Dialect/SybaseSQLAnywhere10Dialect.cs
```diff
@@ -343,6 +343,8 @@ protected virtual void RegisterMiscellaneousFunctions()
 			RegisterFunction("transactsql", new StandardSQLFunction("transactsql", NHibernateUtil.String));
 			RegisterFunction("varexists", new StandardSQLFunction("varexists", NHibernateUtil.Int32));
 			RegisterFunction("watcomsql", new StandardSQLFunction("watcomsql", NHibernateUtil.String));
+			RegisterFunction("truncnum", new StandardSafeSQLFunction("truncnum", 2));
+			RegisterFunction("truncate", new StandardSafeSQLFunction("truncnum", 2));
```
fredericDelaporte: I do not think that we should register specific function names for being available in HQL when using their specific dialect, since HQL should be database agnostic. `RegisterFunction("truncnum"` should have been added.

src/NHibernate/Dialect/SybaseSQLAnywhere10Dialect.cs
```diff
@@ -343,6 +343,8 @@ protected virtual void RegisterMiscellaneousFunctions()
 			RegisterFunction("transactsql", new StandardSQLFunction("transactsql", NHibernateUtil.String));
 			RegisterFunction("varexists", new StandardSQLFunction("varexists", NHibernateUtil.Int32));
 			RegisterFunction("watcomsql", new StandardSQLFunction("watcomsql", NHibernateUtil.String));
+			RegisterFunction("truncnum", new StandardSafeSQLFunction("truncnum", 2));
```
fredericDelaporte: I do not think that we should register specific function names for being available in HQL when using their specific dialect, since HQL should be database agnostic. `RegisterFunction("truncnum"` should not have been added.

hazzik: I think opposite. Moreover, this is how it always was: dialect specific functions + aliases to the more common name.

fredericDelaporte: "Always" widely depends on dialect. Some do register specific names for some of their functions in addition to the generic one, others do not. (By example SQL-Server registers only generic names for `getdate`, `len`, `datepart`, `charindex`, same for Postgres and `now`, ...)
It looks to me as something quite accidental.

This practice of registering specific names does not promote portability of HQL, while an ORM is usually meant to be database agnostic. That is why I think this is a bad thing, causing me to prefer not adding new specific names when we have a generic one. (Removing them after a release would of course be a breaking change for those using them.)

PR #1534, opened 2018-01-16
Fixed entity name retrieval for EntityProjection

bahusoid:
I've just realized that entity name should be retrieved from criteria instead of using `_entityType.FullName`. Otherwise queries with explicitly provided entity names won't work. I believe that existing tests in `EntityProjectionsTest` cover this change.


fredericDelaporte:
> Otherwise queries with explicitly provided entity names won't work.

Having a test for them would then be great.

PR #1537, opened 2018-01-17
Avoid unnecessary persister lookup in Loader

bahusoid:
Additional persister lookup is really required only for subclasses (`rootPersister.HasSubclasses && rootPersister.EntityName != instanceClass`, also see `GetInstanceClass`  where `instanceClass` is retrieved) 

But let's simply check if given entity name comes directly from provided persister.

### DIFF
--- a/src/NHibernate/Async/Loader/Loader.cs
+++ b/src/NHibernate/Async/Loader/Loader.cs
@@ -706,7 +706,9 @@ namespace NHibernate.Loader
 			object id = key.Identifier;
 
 			// Get the persister for the _subclass_
-			ILoadable persister = (ILoadable)Factory.GetEntityPersister(instanceClass);
+			ILoadable persister = ReferenceEquals(instanceClass, rootPersister.EntityName)
+				? rootPersister
+				: (ILoadable) Factory.GetEntityPersister(instanceClass);
 
 			if (Log.IsDebugEnabled())
 			{
--- a/src/NHibernate/Async/Loader/Loader.cs
+++ b/src/NHibernate/Async/Loader/Loader.cs
@@ -706,7 +706,7 @@ namespace NHibernate.Loader
 			object id = key.Identifier;
 
 			// Get the persister for the _subclass_
-			ILoadable persister = ReferenceEquals(instanceClass, rootPersister.EntityName)
+			ILoadable persister = instanceClass == rootPersister.EntityName
 				? rootPersister
 				: (ILoadable) Factory.GetEntityPersister(instanceClass);
 
--- a/src/NHibernate/Loader/Loader.cs
+++ b/src/NHibernate/Loader/Loader.cs
@@ -1062,7 +1062,9 @@ namespace NHibernate.Loader
 			object id = key.Identifier;
 
 			// Get the persister for the _subclass_
-			ILoadable persister = (ILoadable)Factory.GetEntityPersister(instanceClass);
+			ILoadable persister = ReferenceEquals(instanceClass, rootPersister.EntityName)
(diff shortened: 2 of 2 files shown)


src/NHibernate/Async/Loader/Loader.cs
```diff
@@ -706,7 +706,9 @@ protected virtual Task<object[]> GetResultRowAsync(Object[] row, DbDataReader rs
 			object id = key.Identifier;
 
 			// Get the persister for the _subclass_
-			ILoadable persister = (ILoadable)Factory.GetEntityPersister(instanceClass);
+			ILoadable persister = ReferenceEquals(instanceClass, rootPersister.EntityName)
```
hazzik: Are you sure that should be ReferenceEquals?

bahusoid: I think it's pretty safe and most effective.

I've verified that the same instance is retrieved when subclass lookup `persister.GetSubclassForDiscriminatorValue` is performed for base subclass discriminator value. And from what I see all entity names  come from `PersistentClass.EntityName'

Maybe I should add an additional comment to `GetInstanceClass`. Something like: if entity name is dynamically calculated and might be equal to persister.EntityName please return perster.EntityName to avoid additional persister lookup in `GetResultRow`.

Or just let me know if you think it's better to use full string comparison here :)

fredericDelaporte: Personally I am hard to convince about doing such micro-optimisation. Here we do not actually care about the string having the same reference but about them being the same according to default equality.
Default string equality is already quite optimized. I do not think it is worth trying to be "smarter", at the risk of causing more inefficiency with some later code change, even if "protected" by a comment asking for coding in a special way due to this micro-optimization here.

bahusoid: Ok (though I don't think this code can cause more inefficiency  than before)  What about comparer? :)  Ordinal? Default (culture specific)?

I see that `SessionFactoryImpl.entityPersisters` is created with default culture specific comparer. Is it really right? I think it should be either Ordinal or Invariant (I'm voting for  Ordinal for entity names)

bahusoid: Oups.. For some reasons I thought that string.Equals do by default culture specific comparison.. Sorry for all the fuss :)

hazzik:
@bahusoid put of curiosity: are there any measurable benefits of doing this change?

bahusoid:
Well this lookup was executed for each row and for each entity in row. So the benefit is really configuration/data specific. It really should be noticeable if you load big amount of data and have many persisters in configuration and no subclasses in query. 

Actually I've noticed another place to improve here:
https://github.com/nhibernate/nhibernate-core/blob/bea25bf14ceca698807784d90e6e12f667fd8053/src/NHibernate/Loader/Loader.cs#L1079-L1082
For subclass persister column aliases are recalculated for each row - `EntityAliases[i].GetSuffixedPropertyAliases(persister)`. We can cache them by `i`, `entityName`.
Do you agree? I can do it as part of this PR.

bahusoid:
> It really should be noticeable
I've made some tests... And no - it won't be visibly noticeable in  most cases. But on 100,000 iterations I clearly see a difference 1ms  vs 5-7 ms  :) Still - little bit here and little bit there and it will make a difference on the whole :)

fredericDelaporte:
So you may be interested in having a look into #476, it is also a "perf PR", stale for some time.

bahusoid:
Well interface changes.. No I'm not into it..

PR #1540, opened 2018-01-18
 Fix constant value in function cached in LINQ query

### LINKED ISSUE #1330: NH-3673 - Closure variable values locked in from expressions in NHibernate LINQ provider
> 
> Reproduction case:
> 
> ```csharp
> class Repro
> {
>     {
>     }
>     public object Run()
>     {
>         {
>                 .ToList();
>         }
>         return result;
>     }
> }
> ```
> 
> 
---
>

fredericDelaporte:
* Fixes #1330 (NH-3673)
 * Allows a workaround for #1363 (NH-2500)

~~Includes #1539 commit.~~ (Now merged and this PR is rebased.)

### DIFF
--- a/src/NHibernate/Linq/Visitors/ExpressionKeyVisitor.cs
+++ b/src/NHibernate/Linq/Visitors/ExpressionKeyVisitor.cs
@@ -163,20 +163,21 @@ namespace NHibernate.Linq.Visitors
 		{
 			var old = insideSelectClause;
 
-			switch (expression.Method.Name)
-			{
-				case "First":
-				case "FirstOrDefault":
-				case "Single":
-				case "SingleOrDefault":
-				case "Select":
-				case "GroupBy":
-					insideSelectClause = true;
-					break;
-				default:
-					insideSelectClause = false;
-					break;
-			}
+			if (expression.Method.DeclaringType?.Namespace == "System.Linq")
+				switch (expression.Method.Name)
+				{
+					case "First":
+					case "FirstOrDefault":
+					case "Single":
+					case "SingleOrDefault":
+					case "Select":
+					case "GroupBy":
+						insideSelectClause = true;
+						break;
+					default:
+						insideSelectClause = false;
+						break;
+				}
 
 			Visit(expression.Object);
 			_string.Append('.');
--- a/src/NHibernate/Linq/Visitors/ExpressionKeyVisitor.cs
+++ b/src/NHibernate/Linq/Visitors/ExpressionKeyVisitor.cs
@@ -109,19 +109,23 @@ namespace NHibernate.Linq.Visitors
 				{
 					_string.Append("NULL");
 				}
+				else if (expression.Value is IQueryable)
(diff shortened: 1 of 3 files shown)


src/NHibernate/Linq/Visitors/ExpressionKeyVisitor.cs
```diff
@@ -163,20 +163,21 @@ protected override Expression VisitMethodCall(MethodCallExpression expression)
...
-				case "SingleOrDefault":
-				case "Select":
-				case "GroupBy":
-					insideSelectClause = true;
-					break;
-				default:
-					insideSelectClause = false;
-					break;
-			}
+			if (expression.Method.DeclaringType?.Namespace == "System.Linq")
```
hazzik: It should have `esle` branch.

fredericDelaporte: For putting what in it? All the point of that `if` is to not alter `insideSelectClause`, considering that only System.Linq calls may represent an exit out of select clause.

fredericDelaporte:
> Can you add a test case to verify that assumption that?

Unless extending the LINQ query model with new custom clause extensions (things having a purpose similar to `Queryable.Where`), which Linq-2-NH does not do, I do not see what could prove this wrong.

Anyway, I have added a test on key generation with constants in a where.

fredericDelaporte:
If #1542 gets merged before this one, this one would be obsoleted and in need of being closed.

src/NHibernate/Linq/Visitors/ExpressionKeyVisitor.cs
hazzik: To put `insideSelectClause = false;` as it is in the `default` branch of switch statement.

hazzik: Ok, I've checked that my suggestion does not work.

hazzik:
I have removed the `.ToString` overload requirements. This works by adding `GetHashCode` to the key.

src/NHibernate/Linq/Visitors/ExpressionKeyVisitor.cs
```diff
@@ -109,19 +109,23 @@ protected override Expression VisitConstant(ConstantExpression expression)
...
-					else
+					_string.Append(expression.Value);
+				}
+				else if (expression.Value is IEnumerable enumerable && !(enumerable is string))
+				{
+					_string.Append("{");
+					foreach (var value in enumerable)
 					{
-						_string.Append(expression.Value);
+						_string.Append(value).Append("#").Append(value.GetHashCode()).Append(",");
```
fredericDelaporte: I have a hard time conceiving a valid query having objects instances instead of primitive types in a list/array. Are you sure this is useful to change the enumerable case too?

fredericDelaporte:
Obsoleted by #1544.

PR #1541, opened 2018-01-19
Cache subclass entity aliases in Loader

bahusoid:
Avoid subclass entity aliases recalculation for each row.

### DIFF
--- a/src/NHibernate/Loader/Loader.cs
+++ b/src/NHibernate/Loader/Loader.cs
@@ -60,6 +60,11 @@ namespace NHibernate.Loader
 		/// </summary>
 		private bool? _canUseLimits;
 
+		/// <summary>
+		/// Caches subclass entity aliases for given persister index in <see cref="EntityPersisters"/>  and subclass entity name
+		/// </summary>
+		private readonly Dictionary<Tuple<int, string>, string[][]> _subclassEntityAliasesMap = new Dictionary<Tuple<int, string>, string[][]>();
+			
 		protected Loader(ISessionFactoryImplementor factory)
 		{
 			_factory = factory;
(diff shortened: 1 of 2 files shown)


src/NHibernate/Loader/Loader.cs
```diff
@@ -60,6 +60,11 @@ public abstract partial class Loader
 		/// </summary>
 		private bool? _canUseLimits;
 
+		/// <summary>
+		/// Caches subclass entity aliases for given persister index in <see cref="EntityPersisters"/>  and subclass entity name
+		/// </summary>
+		private readonly Dictionary<Tuple<int, string>, string[][]> _subclassEntityAliasesMap = new Dictionary<Tuple<int, string>, string[][]>();
```
fredericDelaporte: This assumes a loader cannot be re-used for a different query. It seems to be true, but it is not easy to ascertain (or I have missed something).

gliljas: Good point. It seems to be true, but needs more investigation.

bahusoid: @fredericDelaporte I don't think it's possible.

https://github.com/nhibernate/nhibernate-core/blob/083a9280f95a2b1dfc2ecf07cc414bc064e56a1a/src/NHibernate/Loader/Loader.cs#L1692-L1696

Comment suggests that EntityAliases and other caches are calculated once. Plus there is already `_columnNameCache` cache which looks pretty query specific

bahusoid: Ok. I've checked all the existing loaders - all initializations are done in ctor once. It's not really practical to reuse Loader - you need to recalculate everything. So don't know what else to add.

Again comment for `PostInstantiate` states that you *must* it call after instantiation and *will* cache select clause suffixes. Doesn't sound reusable at all.

PR #1542, opened 2018-01-19
Avoid re-using query plan embedding different constant values.

### LINKED ISSUE #866: NH-2658 - Allow the IHqlGeneratorForMethod to decide how to build a query cache key
**Daniel Guenter** created an [issue](https://nhibernate.jira.com/browse/NH-2658) — <time datetime="2011-04-20T09:27:32.916Z" title="2011-04-20T09:27:32.916Z">20th April 2011, 9:27:32</time>:
> The linq provider caches a query assuming that ALL constants will be turned into HQL parameters. Sometimes this is not the case. A linq extension that selects a concrete entity property based on a constant value will never pass that value into HQL.
> 
---
**Daniel Guenter** added a comment — <time datetime="2011-04-20T09:40:52.837Z" title="2011-04-20T09:40:52.837Z">20th April 2011, 9:40:52</time>:
> You will have to import the files into the project.
> 
---
**Daniel Guenter** added a comment — <time datetime="2011-04-20T16:31:26.541Z" title="2011-04-20T16:31:26.541Z">20th April 2011, 16:31:26</time>:
> I have a patch (HG) which solves this issue.
> 
> https://bitbucket.org/nicaog/nhibernate-patches/changeset/8ea3eb358e61
> 
> However it does introduce a new concept into the system. IHqlGenerator(s) are allowed to participate in the initial building of the expression tree and key through the use of a SpecialMemberVisitor and a new interface method called Visit.  This allows for expression replacing and parameter manipulation before the expression key is finalized. 
> 
> This solution may be a bit more intrusive than I had hoped it would be. However it does allow for more advanced and intelligent hql generators.
> 
---
**Daniel Guenter** added a comment — <time datetime="2011-08-29T20:44:55.326Z" title="2011-08-29T20:44:55.326Z">29th August 2011, 20:44:55</time>:
> I've updated my solution for this issue to include a NhNonParameterExpression that is used to circumvent ConstantExpression(s) and prevent the value from being cached.
> 
> This is a systemic problem with NHibernate's linq provider in dealing with any programmatic values that deal with column names.
> 
> If there is any interest in this solution, I will go through the work of generating a patch that works against SVN, otherwise you can see the changes at
> https://bitbucket.org/nicaog/nhibernate-patches/changeset/0d88bb5862a2
> 
---
**kkozmic** added a comment — <time datetime="2012-02-08T10:38:38.983Z" title="2012-02-08T10:38:38.983Z">8th February 2012, 10:38:38</time>:
> Any progress on this issue? I wasted 2h today chasing an issue that was caused by this. How likely is this going to get fixed in vNext?
> 
---
**Oskar Berggren** added a comment — <time datetime="2012-02-08T10:49:30.319Z" title="2012-02-08T10:49:30.319Z">8th February 2012, 10:49:30</time>:
> Doesn't this sound a bit like NH-2500?
> 
---
**Oskar Berggren** added a comment — <time datetime="2012-02-08T17:15:45.787Z" title="2012-02-08T17:15:45.787Z">8th February 2012, 17:15:45</time>:
> Linking NH-2658 and NH-2500. Not exactly the same, but sort of complementary.
> 
---
**Tamizhvendan S** added a comment — <time datetime="2014-07-02T12:32:14.768Z" title="2014-07-02T12:32:14.768Z">2nd July 2014, 12:32:14</time>:
> Any Solutions for this issue ?
> 
> Can we add the patch provided by Daniel, Guenter ?
> 
---
**Sÿl** added a comment — <time datetime="2015-04-22T09:55:30.978Z" title="2015-04-22T09:55:30.978Z">22nd April 2015, 9:55:30</time>:
> I agree this issue becomes a real problem when optimisations are required.
> I reviewed the patch from <~danielg> and it seems quite ok, why this long to react on it?
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2015-07-09T04:21:52.138Z" title="2015-07-09T04:21:52.138Z">9th July 2015, 4:21:52</time>:
> Dynamic component part has been fixed as NH-2664
>

### LINKED ISSUE #1330: NH-3673 - Closure variable values locked in from expressions in NHibernate LINQ provider

fredericDelaporte:
* Fixes #1330 (NH-3673)
 * Obsoletes #866 (NH-2658)
 * Fixes most of #1363 (a workaround is still needed for one case; NH-2500)

As said in #1363, I find the `insideSelectClause` trick a bit hackish, and I was willing to investigate patches from #1363 and #866 corresponding JIRAs.
The one from #1363 was quite interesting, but is missing cases where the constant is not parameterized while not being in select clause (handled by a post transformer), as in #866. (Cases which are also missed by the `insideSelectClause` trick.)
Patch from #866 just addresses custom LINQ extensions needing to signal a constant as non-parameterized. So it was fixing just a very small subset of the trouble, while requiring many changes, both in NHibernate and on user side.

Here I propose yet another way: detect than some constants have not been used as parameters, and refine adequately the plan caching. This must be a two steps process since such constants are only known once the plan has been generated.
So at that point, if such constants are detected, the query is flagged as refined, it has a new refined key, and it is cached for that additional key, along with being cached with its "original" key.
On cache hit with a non-refined key, the key is refined thanks to plan's query informations, and checked against plan's query again. If not matching, it is looked-up again, but this time with the key already refined.

It allows to remove the `insideSelectClause` trick, and solves most of the cases. Only the "cached object instance" issue remains, but with a workaround.

If this PR is merged before #1540, #1540 would then be obsoleted and no more needed.


src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -58,15 +58,38 @@ public IQueryExpressionPlan GetHQLQueryPlan(IQueryExpression queryExpression, bo
...
 					log.Debug("unable to locate HQL query plan in cache; generating ({0})", queryExpression.Key);
 				}
+
+				var refinableQuery = queryExpression as IRefinableKeyQueryExpression;
+				var wasAlreadyRefined = refinableQuery?.RefinedKey;
+
 				plan = new QueryExpressionPlan(queryExpression, shallow, enabledFilters, factory);
 				planCache.Put(key, plan);
+
+				if (wasAlreadyRefined == false && refinableQuery.RefinedKey)
```
fredericDelaporte: Here, instead of caching with original and refined key, we could take quite a different path: cache nothing. So the `RefinedKey` would be renamed `NotCacheable`, it would be checked right after plan generation, and if `true`, the plan would not be put in cache. It would have several advantages:

- Simpler.
- Would solve last #1363 trouble without the workaround.
- Would avoid retaining captured object instances.
- Would avoid bloating the cache with as many entries as constant values used with queries having non-parameterized constants.

In #1363, some users were even asking for a way to fully disable the query plan cache, writing they were not feeling a performance hit for not having it. Here it would disable it only for queries embedding constants which are not translated to HQL parameters, so it will be even less likely to have a negative impact.

fredericDelaporte:
Replaced by #1544.

PR #1543, opened 2018-01-19
Various string manipulation optimizations

bahusoid:
I've started with 'StringHelper.Root' refactoring to avoid double search. But somehow ended up 
with bunch of other string related changes :)

1) StringHelper.Root refactored to get rid of double search;
2) string.IndexOf - use ordinal comparison where applicable;
3) No need to convert full string to lowercase for StartsWith comparison;
4) Avoid SqlString.ToString conversion;
5) Optimized and generalized HqlVariablePrefix usage;

It's better to review per commit to better understand the scope of changes.

### DIFF
--- a/src/NHibernate/Engine/Query/CallableParser.cs
+++ b/src/NHibernate/Engine/Query/CallableParser.cs
@@ -37,13 +37,20 @@ namespace NHibernate.Engine.Query
 
 			callableDetail.FunctionName = functionMatch.Groups[1].Value;
 
-			callableDetail.HasReturn = indexOfCall > 0 &&
-										sqlString.IndexOf('?') > 0 &&
-										sqlString.IndexOf('=') > 0 &&
-										sqlString.IndexOf('?') < indexOfCall &&
-										sqlString.IndexOf('=') < indexOfCall;
+			callableDetail.HasReturn = HasReturnParameter(sqlString, indexOfCall);
 
 			return callableDetail;
+		}
+
+		internal static bool HasReturnParameter(string sqlString, int indexOfCall)
+		{
+			int indexOfQuestionMark;
+			int indexOfEqual;
+			return indexOfCall > 0 &&
+					(indexOfQuestionMark = sqlString.IndexOf('?')) > 0 &&
+					(indexOfEqual = sqlString.IndexOf('=')) > 0 &&
+					indexOfQuestionMark < indexOfCall &&
+					indexOfEqual < indexOfCall;
 		}
 	}
 }
--- a/src/NHibernate/Engine/Query/CallableParser.cs
+++ b/src/NHibernate/Engine/Query/CallableParser.cs
@@ -21,9 +21,9 @@ namespace NHibernate.Engine.Query
 		{
 			Detail callableDetail = new Detail();
 
-			callableDetail.IsCallable = sqlString.IndexOf("{") == 0 &&
-										sqlString.IndexOf("}") == (sqlString.Length - 1) &&
-										sqlString.IndexOf("call") > 0;
+			callableDetail.IsCallable = sqlString.IndexOf('{') == 0 &&
+										sqlString.IndexOf('}') == (sqlString.Length - 1) &&
+										sqlString.IndexOf("call", StringComparison.Ordinal) > 0;
 
 			if (!callableDetail.IsCallable)
 				return callableDetail;
@@ -35,11 +35,11 @@ namespace NHibernate.Engine.Query
 
 			callableDetail.FunctionName = functionMatch.Groups[1].Value;
 
-			callableDetail.HasReturn = sqlString.IndexOf("call") > 0 &&
-										sqlString.IndexOf("?") > 0 &&
-										sqlString.IndexOf("=") > 0 &&
-										sqlString.IndexOf("?") < sqlString.IndexOf("call") &&
-										sqlString.IndexOf("=") < sqlString.IndexOf("call");
+			callableDetail.HasReturn = sqlString.IndexOf("call", StringComparison.Ordinal) > 0 &&
+										sqlString.IndexOf('?') > 0 &&
+										sqlString.IndexOf('=') > 0 &&
+										sqlString.IndexOf('?') < sqlString.IndexOf("call", StringComparison.Ordinal) &&
+										sqlString.IndexOf('=') < sqlString.IndexOf("call", StringComparison.Ordinal);
 
 			return callableDetail;
 		}
--- a/src/NHibernate/Engine/Query/CallableParser.cs
+++ b/src/NHibernate/Engine/Query/CallableParser.cs
@@ -21,9 +21,11 @@ namespace NHibernate.Engine.Query
 		{
 			Detail callableDetail = new Detail();
 
-			callableDetail.IsCallable = sqlString.IndexOf('{') == 0 &&
-										sqlString.IndexOf('}') == (sqlString.Length - 1) &&
-										sqlString.IndexOf("call", StringComparison.Ordinal) > 0;
+			int indexOfCall = -1;
+			callableDetail.IsCallable = sqlString.Length > 5 && // to be able to check sqlString[0] we at least need to make sure that string has at least 1 character. The simplest case all other conditions are true is "{call}" which is 6 characters, so check it.
+										sqlString[0] == '{' &&
+										sqlString[sqlString.Length - 1] == '}' &&
+										(indexOfCall = sqlString.IndexOf("call", StringComparison.Ordinal)) > 0;
 
 			if (!callableDetail.IsCallable)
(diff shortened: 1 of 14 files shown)


src/NHibernate/Hql/Ast/ANTLR/Util/JoinProcessor.cs
```diff
@@ -168,12 +168,12 @@ private static SqlString ProcessFromFragment(SqlString frag, JoinSequence join)
 
 		private static bool HasDynamicFilterParam(SqlString sqlFragment)
 		{
-			return sqlFragment.IndexOfCaseInsensitive(ParserHelper.HqlVariablePrefix) < 0;
```
bahusoid: Actually I don't understand this code. Why "has dynamic filter param" is true when HqlVariablePrefix is missing?

src/NHibernate/Engine/Query/CallableParser.cs
```diff
@@ -21,9 +21,9 @@ public static Detail Parse(string sqlString)
 		{
 			Detail callableDetail = new Detail();
 
-			callableDetail.IsCallable = sqlString.IndexOf("{") == 0 &&
-										sqlString.IndexOf("}") == (sqlString.Length - 1) &&
-										sqlString.IndexOf("call") > 0;
+			callableDetail.IsCallable = sqlString.IndexOf('{') == 0 &&
+										sqlString.IndexOf('}') == (sqlString.Length - 1) &&
+										sqlString.IndexOf("call", StringComparison.Ordinal) > 0;
```
hazzik: In the negative case (most of the times) it's more performant to check `sqlString[0]` == '}' && sqlString[sqlString.Length - 1] == '}'`

hazzik: Eg, the full check shall be:

```
sqlString.Length > 5 && // to be able to check sqlString[0] we at least need to make sure that string has at least 1 character. The simplest case all other conditions are true is "{call}" which is 6 characters, so check it.
sqlString[0] == '}' &&
sqlString[sqlString.Length - 1] == '}' &&
sqlString.IndexOf("call", StringComparison.Ordinal) > 0
```

src/NHibernate/Engine/Query/CallableParser.cs
```diff
@@ -35,13 +35,13 @@ public static Detail Parse(string sqlString)
...
-			callableDetail.HasReturn = sqlString.IndexOf("call") > 0 &&
-										sqlString.IndexOf("?") > 0 &&
-										sqlString.IndexOf("=") > 0 &&
-										sqlString.IndexOf("?") < sqlString.IndexOf("call") &&
-										sqlString.IndexOf("=") < sqlString.IndexOf("call");
+			callableDetail.HasReturn = sqlString.IndexOf("call", StringComparison.Ordinal) > 0 &&
+										sqlString.IndexOf('?') > 0 &&
+										sqlString.IndexOf('=') > 0 &&
+										sqlString.IndexOf('?') < sqlString.IndexOf("call", StringComparison.Ordinal) &&
+										sqlString.IndexOf('=') < sqlString.IndexOf("call", StringComparison.Ordinal);
```
hazzik: We are talking about performance in this PR, right? So `sqlString.IndexOf("call", StringComparison.Ordinal)` needs to be cached in a variable.

src/NHibernate/Engine/Query/ParameterParser.cs
```diff
@@ -42,11 +42,11 @@ private ParameterParser()
...
 			// TODO: WTF? "CALL"... it may work for ORACLE but what about others RDBMS ? (by FM)
-			bool hasMainOutputParameter = sqlString.IndexOf("call") > 0 &&
-										  sqlString.IndexOf("?") > 0 &&
-										  sqlString.IndexOf("=") > 0 &&
-										  sqlString.IndexOf("?") < sqlString.IndexOf("call") &&
-										  sqlString.IndexOf("=") < sqlString.IndexOf("call");
+			bool hasMainOutputParameter = sqlString.IndexOf("call", StringComparison.Ordinal) > 0 &&
+										  sqlString.IndexOf('?') > 0 &&
+										  sqlString.IndexOf('=') > 0 &&
+										  sqlString.IndexOf('?') < sqlString.IndexOf("call", StringComparison.Ordinal) &&
```
hazzik: Same

src/NHibernate/Engine/Query/CallableParser.cs
```diff
@@ -35,13 +37,13 @@ public static Detail Parse(string sqlString)
...
-			callableDetail.HasReturn = sqlString.IndexOf("call") > 0 &&
-										sqlString.IndexOf("?") > 0 &&
-										sqlString.IndexOf("=") > 0 &&
-										sqlString.IndexOf("?") < sqlString.IndexOf("call") &&
-										sqlString.IndexOf("=") < sqlString.IndexOf("call");
+			callableDetail.HasReturn = indexOfCall > 0 &&
+										sqlString.IndexOf('?') > 0 &&
+										sqlString.IndexOf('=') > 0 &&
+										sqlString.IndexOf('?') < indexOfCall &&
+										sqlString.IndexOf('=') < indexOfCall;
```
fredericDelaporte: Maybe even go as far as:
```c#
var indexOfQuestionMark = -1;
var indexOfEqual = -1;
callableDetail.HasReturn =
	indexOfCall > 0 &&
		(indexOfQuestionMark = sqlString.IndexOf('?')) > 0 &&
		(indexOfEqual = sqlString.IndexOf('=')) > 0 &&
		indexOfQuestionMark < indexOfCall &&
		indexOfEqual < indexOfCall;

```

src/NHibernate/Engine/Query/ParameterParser.cs
```diff
@@ -42,11 +42,12 @@ private ParameterParser()
...
-										  sqlString.IndexOf("?") > 0 &&
-										  sqlString.IndexOf("=") > 0 &&
-										  sqlString.IndexOf("?") < sqlString.IndexOf("call") &&
-										  sqlString.IndexOf("=") < sqlString.IndexOf("call");
+			var indexOfCall = sqlString.IndexOf("call", StringComparison.Ordinal);
+			bool hasMainOutputParameter = indexOfCall > 0 &&
+										  sqlString.IndexOf('?') > 0 &&
+										  sqlString.IndexOf('=') > 0 &&
+										  sqlString.IndexOf('?') < indexOfCall &&
+										  sqlString.IndexOf('=') < indexOfCall;
```
fredericDelaporte: Same as previous.

src/NHibernate/SqlCommand/SqlString.cs
```diff
@@ -404,6 +404,11 @@ public int IndexOfCaseInsensitive(string text)
 			return IndexOf(text, 0, _length, StringComparison.InvariantCultureIgnoreCase);
```
fredericDelaporte: It appears to me this one should be switched to `StringComparison.OrdinalIgnoreCase`. Invariant does some Unicode character equivalency logic, which is most likely to cause "false positives" from a query standpoint. Moreover case insensitive ordinal is more performant than invariant. And this method is used either on ASCII characters or on query aliases.

But overall they are more than hundred cases were invariant is used in the code base. It looks to me most of them if not all should be switched to ordinal. Maybe worth a dedicated PR.

(By the way if changing this case here, the `text` parameter comment looks obsolete, its last part should be removed.)

bahusoid: >It appears to me this one should be switched to StringComparison.OrdinalIgnoreCase. 

Yeah. But it's public method. I'm afraid to change its behaviour :) 

> It looks to me most of them if not all should be switched to ordinal. 
Agreed (but not as part of this PR)

src/NHibernate/Engine/Query/CallableParser.cs
bahusoid: Ok. Will do

src/NHibernate/SqlCommand/SqlString.cs
fredericDelaporte: > Yeah. But it's public method. I'm afraid to change its behaviour :)

Fixing a behavior which is a bug is OK. I do not think the invariant behavior discrepancies with ordinal can be anything else than bugs for this case, but well, maybe I am wrong. Something that is equal with invariant but not equal with ordinal looks very likely to me to be considered not equal by the database query parser as well, in which case we would have a bug. Still this can be the subject of another PR, switching all invariant which should be to ordinal. I also think the change of most of them could be considered as also fixing bug possibilities.

@hazzik, what is your view on this subject? Do you think it is ok to switch from invariant to ordinal in a minor release, considering it is also a bug fix, or would you rather have such change in a major release?

PR #1545, opened 2018-01-20
Support to join not associated entities in Criteria (aka Entity Join)

bahusoid:
With this PR it's possible to join entities with not mapped association in QueryOver via `JoinEntityQueryOver`/`JoinEntityAlias `methods:
```C#
Entity root = null;
NotRelatedEntity notRelatedEntity = null;

root = session.QueryOver(() => root)
//Join on not mapped entity
.JoinEntityQueryOver(() => notRelatedEntity, () => notRelatedEntity.Id == root.NotRelatedEntityId)
//Also preload child for NotRelatedEntity
.JoinQueryOver(nre => nre.SomeChild)
.Take(1)
.SingleOrDefault();

//Obtain from session already preloaded notRelatedEntity along with notRelatedEntity.SomeChild
notRelatedEntity = session.Get<NotRelatedEntity>(root.NotRelatedEntityId);
```
See other options in [tests](https://github.com/bahusoid/nhibernate-core/blob/customJoin/src/NHibernate.Test/Criteria/EntityJoinCriteriaTest.cs)

"Entity Join" name comes from Hibernate (where it's implemented for hql only):
http://in.relation.to/2016/02/10/hibernate-orm-510-final-release/


src/NHibernate/Criterion/QueryOver.cs
```diff
@@ -280,7 +284,8 @@ IQueryOver<TRoot> IQueryOver<TRoot>.ReadOnly()
 	/// Implementation of the <see cref="IQueryOver&lt;TRoot, TSubType&gt;"/> interface
 	/// </summary>
 	[Serializable]
-	public class QueryOver<TRoot,TSubType> : QueryOver<TRoot>, IQueryOver<TRoot,TSubType>
+	public class QueryOver<TRoot,TSubType> : QueryOver<TRoot>, IQueryOver<TRoot,TSubType>,
+		ISupportEntityJoinQueryOver<TRoot>
```
fredericDelaporte: I have not seen the reason for implementing this interface on this class rather than on `QueryOver<TRoot>`. Why this choice?

src/NHibernate/Criterion/ISupportEntityJoinQueryOver.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Linq.Expressions;
+using NHibernate.SqlCommand;
+
+namespace NHibernate.Criterion
+{
+	//TODO: Make interface more flexible for changes (maybe it should take only 2 params alias + EntityJoinConfig)
+	public interface ISupportEntityJoinQueryOver
```
fredericDelaporte: I would add a " 6.0 TODO: merge with IQueryOver". Is there a reason for not aiming at this? The same apply to below interface, but of course with `IQueryOver<TRoot>`.

src/NHibernate/Impl/ISupportEntityJoinCriteria.cs
```diff
@@ -0,0 +1,11 @@
+using NHibernate.Criterion;
+using NHibernate.SqlCommand;
+
+namespace NHibernate.Impl
+{
+	public interface ISupportEntityJoinCriteria
```
fredericDelaporte: I would add a " 6.0 TODO: merge with ICriteria". Is there a reason for not aiming at this?

src/NHibernate/EntityJoinExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+using NHibernate.Criterion;
+using NHibernate.Impl;
+using NHibernate.SqlCommand;
+
+namespace NHibernate
+{
+	public static class EntityJoinExtensions
+	{
+		// 6.0 TODO: merge into 'IQueryOver<TType, TSubType>
+		public static TThis JoinEntityAlias<TThis, TAlias>(this TThis queryOver, Expression<Func<TAlias>> alias, JoinType joinType, ICriterion withClause, string entityName = null) where TThis : IQueryOver
```
fredericDelaporte: So that is somewhat the todo I am asking for on interface.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
```diff
@@ -408,15 +439,26 @@ private void CreateCriteriaCollectionPersisters()
...
+			if (tokens.Length == 0)
+				return last;
+			
+			IPropertyMapping lastEntity = rootPersister;
+			int i = 0;
+			if (entityJoins.TryGetValue(tokens[0], out var entityJoinInfo))
+			{
+				last = entityJoinInfo.Persister;
+				lastEntity = (IPropertyMapping) last;
+				++i;
```
fredericDelaporte: `i++;` is more common place, and we do not need pre-increment here.

src/NHibernate/Loader/OuterJoinableAssociation.cs
```diff
@@ -34,7 +34,12 @@ public sealed class OuterJoinableAssociation
 			rhsColumns = JoinHelper.GetRHSColumnNames(joinableType, factory);
 			on = new SqlString(joinableType.GetOnCondition(rhsAlias, factory, enabledFilters));
 			if (SqlStringHelper.IsNotEmpty(withClause))
-				on = on.Append(" and ( ").Append(withClause).Append(" )");
+			{
+				on = on.Count == 0
+					? on.Append(withClause)
+					: on.Append(" and ( ").Append(withClause).Append(" )");
```
fredericDelaporte: For me, that is an undue change. It looks to me the way it is generated, `on` is either empty or starts with `" and "`. Testing on `Count` here changes behavior. Better not do that and keep previous code.

src/NHibernate/SqlCommand/ANSIJoinFragment.cs
```diff
@@ -39,6 +40,12 @@ public override void AddJoin(string tableName, string alias, string[] fkColumns,
 
 			_fromFragment.Add(joinString + tableName + ' ' + alias + " on ");
 
+			if (fkColumns.Length == 0)
+			{
+				_fromFragment.Add(on);
+				return;
```
fredericDelaporte: Here, an additional adjustment should be done. `on` is usually generated with a leading `" and "`, which should be removed before adding to `_fromFragment`.

src/NHibernate/SqlCommand/JoinFragment.cs
```diff
@@ -62,7 +62,7 @@ protected bool AddCondition(SqlStringBuilder buffer, SqlString on)
 		{
 			if (SqlStringHelper.IsNotEmpty(on))
 			{
-				if (!on.StartsWithCaseInsensitive(" and"))
+				if (buffer.Count > 0 && !on.StartsWithCaseInsensitive(" and"))
```
fredericDelaporte: I doubt this change is legit. Was it causing a failure not having this change done? It looks to me code expects condition clauses to always potentially starts with an " and", removing it if needed.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
bahusoid: No, of course It's not about optimization :) it's about habit - I just use `++i;` since C++ and it's more common for me.
I have no objections to change it but curious why do you think that `i++` is more common for C#? Is it because VS/Resharper expands `for` with `i++`?

bahusoid: P.S. I don't have much time this week but I will try to address all the questions/concerns next week.

src/NHibernate/Criterion/QueryOver.cs
bahusoid: Yeah I've just put it next to existing `JoinAlias`/`JoinQueryOver` methods. Looks more natural to have them in the same place.

src/NHibernate/SqlCommand/JoinFragment.cs
bahusoid: The intention here is the same as in `ANSIJoinFragment.AddJoin` - avoid bare `withClause` to be added with " and" prefix to SQL. But yeah - maybe I should handle it in each dialect explicitly.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
fredericDelaporte: > Is it because VS/Resharper expands `for` with `i++`?

No. Just because I have far more seen post-increments than pre-increments in C#, and I do not think I am alone. Alexander had made the same demand, switching away from `++i` to `i++` on a previous PR. The opener argued for performance as far a I remember, and it was discarded as not relevant in C#.

fredericDelaporte: This was [here](https://github.com/nhibernate/nhibernate-core/pull/367#discussion_r19515494).

src/NHibernate/SqlCommand/JoinFragment.cs
bahusoid: You are right - I've removed this change. I've struggled with this " and" prefix and thought this change is required for Oracle.

bahusoid:
@fredericDelaporte It seems I've addressed all your requests.

I also checked how to make `Fetch` work with entity joins. And I think I found a bug. Interface  for `Fetch` is defined as:
https://github.com/nhibernate/nhibernate-core/blob/8d2b1231ed8f0beef1d1f6a6e3b75180cd0dfe7b/src/NHibernate/IQueryOver.cs#L369

`Expression<Func<TRoot, object>> path` -  type for root criteria is provided (which is `TRoot`) in expression parameter . But in implementation fetch mode is applied to `UnderlyingCriteria`(which is `TSubType` based) instead of `RootCriteria`:
https://github.com/nhibernate/nhibernate-core/blob/8d2b1231ed8f0beef1d1f6a6e3b75180cd0dfe7b/src/NHibernate/Criterion/Lambda/QueryOverFetchBuilder.cs#L44

So basically fetch doesn't work after `JoinQueryOver` call.

I think it's actually an `interface` bug - as this `TRoot` restriction is artificial and not really necessary. So it really should be:
```
IQueryOverFetchBuilder<TRoot,TSubType> Fetch(Expression<Func<**TSubType**, object>> path);
```
Any ideas how to resolve it properly? Maybe in 5.x we should made a more straightforward fix - use `RootCriteria` in fetch builder and in 6.x change interface to use `Expression<Func<TSubType, object>> path`.

fredericDelaporte:
> I also checked how to make `Fetch` work with entity joins. And I think I found a bug.

I think it will need to be addressed as another PR, that is why I approve this one as is.

> Any ideas how to resolve it properly? Maybe in 5.x we should made a more straightforward fix - use `RootCriteria` in fetch builder and in 6.x change interface to use `Expression<Func<TSubType, object>> path`.

Well, as usual, for a minor, we cannot change anything in an interface. Does it work to define a suitable extension, that would handle only NHibernate QueryOver internal implementation, while obsoleting the interface method? That should be the preferred way, with some "6.0 TODO" for merging it back into the interface.

bahusoid:
I've made last small adjustments:
1) Fixed exception message. Don't know how `nameof(T)` got there and why it's even allowed :)
2) Added one more extension to Criteria. Just to make things even for CreateEntityCriteria/CreateEntityAlias

fredericDelaporte:
> It seems this PR has fixed another issue btw.

Which is?

hazzik:
I was not near the PC when wrote this. It should be something along NH-3506 and NH-2029

fredericDelaporte:
I have cherry-picked tests for #1366 from #268, and tested them on this PR and on its base commit: no changes, fails the same way, with the same resulting queries (catched inside SQL-Profiler). So it does not look like this PR is fixing anything additional. (NH-2029 was already fixed in 5.0.)

fredericDelaporte:
So I intend to merge this soon.

fredericDelaporte:
There is something forgotten: documenting the new feature in the reference documentation. I think this new feature deserve a mention into [criteria](https://github.com/nhibernate/nhibernate-core/blob/master/doc/reference/modules/query_criteria.xml) and [queryover](https://github.com/nhibernate/nhibernate-core/blob/master/doc/reference/modules/query_queryover.xml) chapters.

PR #1555, opened 2018-01-23
Catch practices: avoid losing catched exception information.

fredericDelaporte:
All catches have been reviewed for avoiding losing original exception information:

 * When another exception is raised, it should embed the original one as an inner exception.
 * Otherwise when re-thrown, it should be done without altering its original stack-trace.
   Preferred way is of course to use "throw;" in the catch clause without specifying the exception, but otherwise ReflectHelper provide a way to preserve the stack-trace.
 * Otherwise it should be transmitted to the logger.

Some cases still do not follow this pattern for avoiding being too noisy, like "Try" utilities, failure in 
closing an already failed object, some "implemented by exception" features, ...

### DIFF
--- a/src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
@@ -362,8 +362,6 @@ namespace NHibernate.Hql.Ast.ANTLR
 			}
 			catch ( RecognitionException e ) 
 			{
-				// we do not actually propogate ANTLRExceptions as a cause, so
-				// log it here for diagnostic purposes
 				if ( log.IsInfoEnabled() ) 
 				{
 					log.Info(e, "converted antlr.RecognitionException");
--- a/src/NHibernate/Hql/Ast/ANTLR/Tree/FromElement.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/Tree/FromElement.cs
@@ -549,7 +549,7 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 					// ignore it; the incoming property could not be found so we
 					// cannot be sure what to do here.  At the very least, the
 					// safest is to simply not apply any dereference toggling...
-					Log.Info(ex, "Unable to find property {0}, no dereference will be handled for it.", propertyName);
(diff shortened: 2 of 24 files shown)


src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -362,8 +362,6 @@ private void DoCompile(IDictionary<string, string> replacements, bool shallow, S
 			}
 			catch ( RecognitionException e ) 
 			{
-				// we do not actually propogate ANTLRExceptions as a cause, so
-				// log it here for diagnostic purposes
```
fredericDelaporte: Many cases do propagate them as inner exception, so removing the comment, they are now all propagated.
But I let the log in place, as this could be considered as a (very minor) breaking change to remove logs.

src/NHibernate/Hql/Ast/ANTLR/Tree/FromElement.cs
```diff
@@ -544,11 +544,12 @@ public void HandlePropertyBeingDereferenced(IType propertySource, string propert
 						_dereferencedBySuperclassProperty = true;
 					}
 				}
-				catch (QueryException)
+				catch (QueryException ex)
 				{
 					// ignore it; the incoming property could not be found so we
 					// cannot be sure what to do here.  At the very least, the
 					// safest is to simply not apply any dereference toggling...
+					Log.Info(ex, "Unable to find property {0}, no dereference will be handled for it.", propertyName);
```
hazzik: This shall be at Debug level. Info is too high.

fredericDelaporte: This is for matching the [other log](https://github.com/nhibernate/nhibernate-core/pull/1555/files/9a4acd4d05683ed39a6b59398d8c27485c1dead0#diff-13423b9d5cda494b63c32e6413dd7f58R535). Not very important, so well:
 - change both to Debug?
 - just this one?
 - keep them both at Info?

hazzik: Just this one, thanks

fredericDelaporte: Done.

PR #1558, opened 2018-01-26
Improved collection batch fetching

gliljas:
Implemented as it was done in https://github.com/hibernate/hibernate-orm/pull/337. HHH-1775


src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -125,7 +132,44 @@ public void AddBatchLoadableEntityKey(EntityKey key)
...
+		public void AddBatchLoadableCollection(IPersistentCollection collection, CollectionEntry ce)
+		{
+			var persister = ce.LoadedPersister;
+
+			if (!batchLoadableCollections.TryGetValue(persister.Role, out var map))
+			{
+				map = new LinkedHashMap<CollectionEntry, IPersistentCollection>();
+				batchLoadableCollections.Add(persister.Role, map);
+			}
+			map[ce]=collection;
```
fredericDelaporte: Some formatting is lacking here.

src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -5,31 +5,32 @@
...
-		/// Even though this is a map, we only use the keys.  A map was chosen in
-		/// order to utilize a <see cref="LinkedHashMap{K, V}" /> to maintain sequencing
-		/// as well as uniqueness.
+		/// A Map structure is used to segment the keys by entity type since loading can only be done for a particular entity
+		/// type at a time.
 		/// </remarks>
-		private readonly IDictionary<EntityKey, object> batchLoadableEntityKeys = new LinkedHashMap<EntityKey, object>(8);
-
+		private readonly IDictionary<string, LinkedHashSet<EntityKey>> batchLoadableEntityKeys = new Dictionary<string, LinkedHashSet<EntityKey>>(8);
+
```
fredericDelaporte: Undue whitespace addition.

src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -143,23 +187,42 @@ public object[] GetCollectionBatch(ICollectionPersister collectionPersister, obj
...
-			// this only works because collection entries are kept in a sequenced
-			// map by persistence context (maybe we should do like entities and
-			// keep a separate sequences set...)
-			foreach (DictionaryEntry me in context.CollectionEntries)
+			if (batchLoadableCollections.TryGetValue(collectionPersister.Role, out var map))
 			{
-				CollectionEntry ce = (CollectionEntry) me.Value;
-				IPersistentCollection collection = (IPersistentCollection) me.Key;
-				if (!collection.WasInitialized && ce.LoadedPersister == collectionPersister)
+				foreach (KeyValuePair<CollectionEntry,IPersistentCollection> me in map)
```
fredericDelaporte: Missing space after comma.

src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -143,23 +187,42 @@ public object[] GetCollectionBatch(ICollectionPersister collectionPersister, obj
...
+						continue;
+					}
+
+					if (collection.WasInitialized)
+					{
+						// should never happen
+						if (log.IsWarnEnabled())
+						{
+							log.Warn("Encountered initialized collection in BatchFetchQueue, this should not happen.");
+						}
```
fredericDelaporte: I personally tend to minimize such code by eliminating the "isEnabled" guard for simple logs, since the logger itself is supposed to test this. I keep the guard only if the log involve more than calling the logger, like having some string computing before calling it.
The "should never happen" comment is also a bit redundant with the log message. So I would tend to only have one line here instead of five, the one doing the log.

src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -202,9 +266,9 @@ public object[] GetEntityBatch(IEntityPersister persister,object id,int batchSiz
 			int end = -1;
 			bool checkForEnd = false;
 
-			foreach (EntityKey key in batchLoadableEntityKeys.Keys)
+			if (batchLoadableEntityKeys.TryGetValue(persister.EntityName,out var set))//TODO: this needn't exclude subclasses...
```
fredericDelaporte: Missing space after comma. And I would rather have the comment on the line above instead of at end of line. I view most end of line comments as bad practice, as they can lack visibility.

src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -202,9 +266,9 @@ public object[] GetEntityBatch(IEntityPersister persister,object id,int batchSiz
...
 			{
-				if (key.EntityName.Equals(persister.EntityName))
+				foreach (var key in set)
 				{
 					//TODO: this needn't exclude subclasses...
```
fredericDelaporte: This comments now looks redundant.

src/NHibernate/Event/Default/EvictVisitor.cs
```diff
@@ -53,6 +53,10 @@ private void EvictCollection(IPersistentCollection collection)
 			Session.PersistenceContext.CollectionEntries.Remove(collection);
 			if (log.IsDebugEnabled())
 				log.Debug("evicting collection: {0}", MessageHelper.CollectionInfoString(ce.LoadedPersister, collection, ce.LoadedKey, Session));
+			if (ce.LoadedPersister?.BatchSize > 1)
```
fredericDelaporte: Why some code test `> 0` and other `> 1`? Shouldn't it be `> 1` everywhere?

hazzik:
I've removed breaking changes. Please review.

src/NHibernate/Engine/CollectionEntry.cs
```diff
@@ -302,6 +302,10 @@ public void PostInitialize(IPersistentCollection collection)
 		{
 			snapshot = LoadedPersister.IsMutable ? collection.GetSnapshot(LoadedPersister) : null;
 			collection.SetSnapshot(loadedKey, role, snapshot);
+			if (LoadedPersister.GetBatchSize() > 1)
+			{
+				collection.GetCurrentSession().PersistenceContext.BatchFetchQueue.RemoveBatchLoadableCollection(this);
```
fredericDelaporte: It is no more a binary breaking change but it is still a breaking one, isn't it? Because if anyone supplies its own `IPersistentCollection` not deriving from `AbstractPersistentCollection`, it will now throw here.

src/NHibernate/Engine/StatefulPersistenceContext.cs
```diff
@@ -835,6 +835,10 @@ public void AddUninitializedCollection(ICollectionPersister persister, IPersiste
 		{
 			CollectionEntry ce = new CollectionEntry(collection, persister, id, flushing);
 			AddCollection(collection, ce, id);
+			if (persister.GetBatchSize() > 1)
```
fredericDelaporte: Same here but for `ICollectionPersister`.

src/NHibernate/Engine/CollectionEntry.cs
hazzik: What if returning 0/1 on get batch size if this is not supported one?

src/NHibernate/Persister/Collection/ICollectionPersister.cs
```diff
@@ -283,4 +284,21 @@ public partial interface ICollectionPersister
...
+		/// </summary>
+		//6.0 TODO: Merge into ICollectionPersister and convert to a property.
+		public static int GetBatchSize(this ICollectionPersister persister)
+		{
+			if (persister is AbstractCollectionPersister acp)
+			{
+				return acp.GetBatchSize();
+			}
+
+			throw new InvalidOperationException("Only persisters of AbstractCollectionPersister are supported.");
```
hazzik: @fredericDelaporte I think returning 1 here would fix this possible breaking change, wouldn't it?

hazzik:
Ok, now it does not have any breaking changes.

gliljas:
@hazzik Nice!

src/NHibernate/Engine/CollectionEntry.cs
```diff
@@ -302,9 +302,20 @@ public void PostInitialize(IPersistentCollection collection)
 		{
```
fredericDelaporte: The single argument `PostInitialize` should be obsoleted. Or is there any reason to keep it, beside avoiding a breaking change? It seems now only called from the two arguments `PostInitialize`.

hazzik: Done

fredericDelaporte:
As for #1560, I consider this one, if squashed, should be squashed by Alexander for slightly better attribution, since he has significantly contributed to it.

PR #1560, opened 2018-01-26
Fix using a loaded IPersistentCollection in a Query

### LINKED ISSUE #1556: Linq query with "Contains" on persistent collection fails
Code like
```                    var parent = session.Get<ParentEntity>(1);

                    session.Query<GrandChildEntity>()
                        .Where(e => parent.Children.Contains(e.Parent))
                        .FirstOrDefault();
```
worked in v4, but fails in v5.
Self-contained test: https://github.com/v-kabanov/Nh5Regression/blob/master/ContainsSubquery.cs

gliljas:
Fixes #1556 

The problem was that the IPersistentCollection now implement IQueryable, which the PartialEvaluatingExpressionVisitor always rewrites to its inner expression. ~I could find no way to intercept this mechanism, except to clone the entire class from source and rewrite it according to our needs. Relinq is licensed under Apache 2.0, so that should not be an issue.~


src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -1,15 +1,110 @@
...
+using NHibernate.Collection;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Parsing;
-using Remotion.Linq.Parsing.ExpressionVisitors;
 using Remotion.Linq.Parsing.ExpressionVisitors.TreeEvaluation;
 
 namespace NHibernate.Linq.Visitors
 {
+	//Modified version of PartialEvaluatingExpressionVisitor from Relinq (https://github.com/re-motion/Relinq)
+	//Copyright (c) rubicon IT GmbH, www.rubicon.eu
```
hazzik: According to https://tldrlegal.com/license/apache-license-2.0-(apache-2.0) we must include the license. So, I think it shall be included on the top of the file.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -1,15 +1,110 @@
...
+	//
+	//Used under the Apache Software License 2.0
 	internal class NhPartialEvaluatingExpressionVisitor : RelinqExpressionVisitor, IPartialEvaluationExceptionExpressionVisitor
 	{
+		public static Expression EvaluateIndependentSubtrees(Expression expression)
+		{
+			return EvaluateIndependentSubtrees(expression, new NhEvaluatableExpressionFilter());
+		}
+		/// <summary>
```
fredericDelaporte: Better have a blank line between methods.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
fredericDelaporte: I think lines differing from original implementation should be signaled by comments, for easing seeing what is done differently in NHibernate.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -1,15 +1,110 @@
...
+		private Expression EvaluateSubtree(Expression subtree)
+		{
+			if (subtree == null) throw new ArgumentNullException(nameof(subtree));
+
+			if (subtree.NodeType == ExpressionType.Constant)
+			{
+				var constantExpression = (ConstantExpression) subtree;
+				var valueAsIQueryable = constantExpression.Value as IQueryable;
+				if (valueAsIQueryable != null && !IsEvaluatedQueryable(constantExpression.Value) && valueAsIQueryable.Expression != constantExpression)
+					return valueAsIQueryable.Expression;
```
fredericDelaporte: I guess this is here that code differs. It seems to create a bunch of test failures, but I have not tried analyzing them yet.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -35,6 +124,8 @@ public Expression VisitPartialEvaluationException(PartialEvaluationExceptionExpr
 		}
 	}
 
+
+
```
fredericDelaporte: Undue additional blank lines.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
gliljas: I'll have a look.

gliljas:
Interesting test errors on a couple of DB types. Investigating.

hazzik:
The simplest fix is to extend filter to mark this type of collection as not enumerable.

hazzik:
An interesting side effect of this change is that operations on `IPersistenceCollection` + `IQueryable` are executed on server side (embedded as a sub-query) as opposite being included as a parameter list.

gliljas:
@hazzik That's a bad side effect.

hazzik:
I’ll double check, but I’m pretty sure that your original changes had the same side effect.

The code change is trivial, so it was easier to check-in, because to verify my assumption I had to make this change.

hazzik:
Actually, I was wrong (it seems that I'm blind). The both cases produce exactly the same queries. They do not inline the query, but rather use the `in (@p1, ...)` expression.

hazzik:
@gliljas, actually, I was wrong (it seems that I'm blind). The both cases produce exactly the same query. They do not inline the collection into the query, but rather use the `in (@p1, ...)` expression.

fredericDelaporte:
For proper attribution it would be better if this PR is squashed/merged by Alexander. If I squash/merge it myself only Gunnar and me would appear on the master commit.

hazzik:
> For proper attribution it would be better if this PR is squashed/merged by Alexander. If I squash/merge it myself only Gunnar and me would appear on the master commit.

Apparently GitHub supports [co-authors](https://help.github.com/articles/creating-a-commit-with-multiple-authors/) for commits.

PR #1562, opened 2018-01-30
Fix round registration

fredericDelaporte:
* PostgreSQL:
   * It was yielding double on decimal columns.
   * It was failing on no-second-argument round.
 * SQL Server and SQL Server Compact Edition:
   * They were not supporting single argument round.


fredericDelaporte:
I have fixed a typo in a comment ('two argument" without a "s").

PR #1567, opened 2018-01-31
Avoid Trim().Length as empty check and ToLowerInvariant() in string comparison

bahusoid:
And made another round of setting ordinal string comparison


src/NHibernate/Util/StringHelper.cs
```diff
@@ -365,8 +365,8 @@ internal static bool IsNotRoot(string qualifiedName, out string root)
 		/// </returns>
 		public static bool BooleanValue(string value)
 		{
-			string trimmed = value.Trim().ToLowerInvariant();
-			return trimmed.Equals("true") || trimmed.Equals("t");
+			string trimmed = value.Trim();
```
fredericDelaporte: This file has an old ascii corruption, at line 162: `fa�ade`, for `façade`, which is a fault by the way, since the comment is not in french. May you please fix that to `facade`? This thing cause some IDE (Rider) to alter the file at each opening, which is a nuisance, but not worth a dedicated PR.

src/NHibernate/Engine/CascadeStyle.cs
```diff
@@ -78,7 +78,7 @@ public virtual bool HasOrphanDelete
 
 		#region  Static helper methods
 
-		private static readonly Dictionary<string, CascadeStyle> Styles = new Dictionary<string, CascadeStyle>();
+		private static readonly Dictionary<string, CascadeStyle> Styles = new Dictionary<string, CascadeStyle>(StringComparer.OrdinalIgnoreCase);
```
fredericDelaporte: This looks to me as an undue change. Cascade styles are not case insensitive, even if currently we would not have names conflict by going insensitive.

fredericDelaporte:
There are also a bunch of `ToUpperInvariant` in Hql.g which should be converted to case insensitive comparison.

bahusoid:
>In SchemaExport and SchemaUpdate, we have for only one test:
I didn't touch this `autoKeyWordsImport` because it's not string with string comparison. It's string compared with object 'Hbm2DDLKeyWords' with implicit to string conversion.

src/NHibernate/Engine/CascadeStyle.cs
bahusoid: Are you sure? Change is based on the following line:

https://github.com/nhibernate/nhibernate-core/blob/4e84dbc4da655c9f63923282fca3e4a511234638/src/NHibernate/Mapping/Property.cs#L114

fredericDelaporte: Well, so the internal usage of this public class is case insensitive. It is then unlikely it could be a breaking change for anyone to render the class case insensitive, but not strictly impossible either.

Well, keep your change in place.

src/NHibernate/Hql/Ast/ANTLR/CollectionProperties.cs
```diff
@@ -11,34 +12,39 @@ namespace NHibernate.Hql.Ast.ANTLR
 	internal static class CollectionProperties
 	{
 
-		public static Dictionary<string, string> HQL_COLLECTION_PROPERTIES;
-
-		private static readonly string COLLECTION_INDEX_LOWER = CollectionPropertyNames.Index.ToLowerInvariant();
+		public static readonly Dictionary<string, string> HQL_COLLECTION_PROPERTIES = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
hazzik: Usage of this collection looks like just a normal HashSet<string>(). As the class is internal, maybe it's better to replace it.

src/NHibernate/SqlCommand/SqlString.cs
```diff
@@ -652,6 +652,18 @@ public SqlString SubstringStartingWithLast(string text)
 			return lastIndex >= 0 ? Substring(lastIndex) : Empty;
 		}
 
+		/// <summary>
+		/// Returns true if content is empty or white space characters only
+		/// </summary>
+		public bool IsEmptyOrWhitespace()
```
hazzik: Probably better to align with a `String` class: make it static; add support for `null`; call it `IsNullOrWhitespace`.

bahusoid: I'm not sure it would give any benefits (at least from dev usability point of view). I tend to check instance methods for emptiness first (and sadly static methods are not even suggested). So I  would find this method quicker.

bahusoid: But let majority rule :) @fredericDelaporte  Any preferences?

fredericDelaporte: Looking at it again, I see a third option: put that into `SqlStringHelper`. It has already an `IsEmpty` method (which is actually an `IsNullOrEmpty`), so it would be more consistent.

About discoverability, this can be improved by making it an extension method.

About aligning with `string` class, we would then have to move `IsEmpty` in `SqlString` too for consistency (and rename it by the way, of course).

bahusoid: Some instance method is still required as I need to obtain data from `GetTrimmedIndexes`.
So I think I can add extension method  `IsNullOrWhitespace`  to 'SqlStringHelper' that will do null check + call `IsEmptyOrWhitespace` . Does it make any sense?

bahusoid: From what I see instance method `IsEmptyOrWhitspace` needs to exists. So the main question does it improve anything adding additional wrappers over it.

hazzik: Leave it as is.

fredericDelaporte: I am not for having an `IsEmptyOrWhitespace` on instance along an `IsNullOrWhitespace` on helper. Better make `GetTrimmedIndexes` internal.

fredericDelaporte: > Leave it as is.

Simplest option. Agreed.

PR #1568, opened 2018-02-01
Fix for update with outer join in PostgreSQL

### LINKED ISSUE #1565: For update with outer join fails with PostgreSQL
When building SQL in `OuterJoinLoader`, `Dialect.SupportsOuterJoinForUpdate` is not being respected. This causes an exception with Postgres because Postgres will not accept blank lock hint with query containing an outer join such as this query: 
```sql
SELECT entity0_.Id as id1_0_0_, entity0_1_.Name as name2_1_0_
  FROM Entities entity0_
  left outer join EntityName entity0_1_ on entity0_.Id=entity0_1_.EntityId
  WHERE entity0_.Id=? for update
```

If we change `SqlSelectBuilder` to use lock hint `for update of entity0_` instead of `for update`, the query can be executed.

I will create a pull request in next few days.

Doan-Van-Tuan:
Add alias in "for update" clause to entity loading query containing outer join if database (e.g Postgres) does not support locking with outer join.

Fixes #1565.


Doan-Van-Tuan:
Hi, please help me with the test failure on TeamCity. I'm not sure how to make it only runs with Postgres.

bahusoid:
I think your `Configure` override needs to be removed. It seems it forces `Postgres` dialect on all configurations.

hazzik:
@Doan-Van-Tuan, @bahusoid is right.

fredericDelaporte:
And also, please reformat the .cs test files added in this PR. There is a mix of tabs and spaces, NHibernate uses tabs.

Doan-Van-Tuan:
Thank you. I'm not familiar with NHibernate test configurations and conventions.

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
```diff
@@ -181,9 +187,10 @@ public SqlSelectBuilder SetHavingClause(SqlString havingSqlString)
 			return this;
 		}
 
-		public SqlSelectBuilder SetLockMode(LockMode lockMode)
+		public SqlSelectBuilder SetLockMode(LockMode lockMode, string alias)
```
fredericDelaporte: This is a binary breaking change. Changing a public method must be done in two steps: obsolete the method, add an overload. The `alias` parameter could be a bit more explicit about what it is, such as `fromTableAlias`. Since the for update can accept a list of tables, it could even be something like `nonOuterAliases`.

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
```diff
@@ -40,6 +42,7 @@ public SqlSelectBuilder SetComment(string comment)
 		public SqlSelectBuilder SetFromClause(string fromClause)
 		{
 			this.fromClause = fromClause;
+			hasOuterJoin = hasOuterJoin || fromClause.ToLowerInvariant().Contains("left outer join");
```
fredericDelaporte: Better use [this code](https://github.com/nhibernate/nhibernate-core/pull/1567/files#diff-92318c5b35fd56f9c1354dca5d86c4bbR693) for a case insensitive `Contains`. It has been merge just now, so update your branch for having this `StringHelper.ContainsCaseInsensitive` method.

The test should be moved into `GetForUpdateString`, and done only if the dialect does not support for update on outer joins. It would avoid doing such a string scan for all other cases.

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
```diff
@@ -110,6 +113,9 @@ public SqlSelectBuilder SetOuterJoins(SqlString outerJoinsAfterFrom, SqlString o
 			}
 
 			this.outerJoinsAfterWhere = tmpOuterJoinsAfterWhere;
+
+			hasOuterJoin = hasOuterJoin || 
+			               outerJoinsAfterFrom != null && outerJoinsAfterFrom.ToString().ToLowerInvariant().Contains("outer join");
```
fredericDelaporte: The test should be moved into `GetForUpdateString`, and done only if the dialect does not support for update on outer joins. It would avoid doing such a string scan for all other cases.

By the way, is it really required here to test the `outerJoinsAfterFrom` contains an outer join? Shouldn't it be enough to test the string is not null or just whitespaces?

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
```diff
@@ -291,6 +298,19 @@ public SqlString ToSqlString()
 			return sqlBuilder.ToSqlString();
 		}
 
+		private string GetForUpdateString()
+		{
+			if (!Dialect.SupportsOuterJoinForUpdate && hasOuterJoin)
+			{
+				if (Equals(lockMode, LockMode.Upgrade))
+					return Dialect.GetForUpdateString(aliasToLock);
```
fredericDelaporte: We still have a trouble here: databases not supporting `for update` on outer join and not supporting `for update` aliases. For such database the code here will keep a `for update` clause, and it will still fail.

So either a new property needs to be added for knowing if the dialect supports lock aliases (currently, if it does not, it yields the `for update` without aliases), and used here, or the for update should just yield here an empty string, removing the need for transmitting the alias by the way.

src/NHibernate.Test/NHSpecificTest/GH1565/PostgresOuterJoinLockTest.cs
```diff
@@ -0,0 +1,62 @@
...
+			base.OnTearDown();
+			using (var session = OpenSession())
+			{
+				session.CreateSQLQuery("delete from MainEntity").ExecuteUpdate();
+			}
+		}
+
+		protected override bool AppliesTo(Dialect.Dialect dialect)
+		{
+			return dialect is PostgreSQLDialect;
```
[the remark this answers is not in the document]
Doan-Van-Tuan: AppliesTo() was removed.

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
Doan-Van-Tuan: Changed to `StringHelper.ContainsCaseInsensitive`; and moved the check for outer join into `GetForUpdateString()`.

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
Doan-Van-Tuan: Is there any DBMS that matches? I'd like to leave this logic for a later date when we need it.

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
```diff
@@ -300,17 +303,23 @@ public SqlString ToSqlString()
...
+				return Dialect.GetForUpdateNowaitString(mainTableAlias);
 
 			return Dialect.GetForUpdateString(lockMode);
-		}
 
+			bool HasOuterJoin()
+			{
+				return
+					StringHelper.ContainsCaseInsensitive(fromClause, "outer join") ||
+					!string.IsNullOrWhiteSpace(outerJoinsAfterFrom?.ToString());
```
bahusoid: It's better to swap those conditions and if whitespace check is really required use `SqlString.IsEmptyOrWhitespace`
https://github.com/nhibernate/nhibernate-core/blob/f8b1bc46f92c7a8cd503d3450f73f6b33ad4e8fb/src/NHibernate/SqlCommand/SqlString.cs#L658

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
fredericDelaporte: Maybe not, but see my update on #1565, it is definitely incorrect to systematically supply a table alias here. Some DB do take indeed a column list here, not a table alias. The correct code handling both cases is in `ForUpdateFragment`.

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
hazzik: @Doan-Van-Tuan could you please incorporate this suggestion?

Doan-Van-Tuan: I'll resume work on this fix in next few days. It was our Lunar New Year.

hazzik:
@fredericDelaporte could you please review this to include in 5.1. It seems only minor request is missing. I'm going to adjust it myself.

src/NHibernate/SqlCommand/SqlSelectBuilder.cs
```diff
@@ -291,6 +301,43 @@ public SqlString ToSqlString()
...
 
+		private string GetForUpdateString()
+		{
+			if (!Dialect.SupportsOuterJoinForUpdate && HasOuterJoin())
+			{
+				var isUpgrade = Equals(lockMode, LockMode.Upgrade);
+				var isUpgradeNoWait = !isUpgrade && (
+					Equals(lockMode, LockMode.UpgradeNoWait) || Equals(lockMode, LockMode.Force));
+				if (!isUpgrade && !isUpgradeNoWait)
+					return string.Empty;
```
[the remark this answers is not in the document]
Doan-Van-Tuan: Thank you all for picking up and finish the work on this. I'm on a tight schedule and could not find the peace of mind to finalize it.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -526,8 +526,16 @@ public virtual string ForUpdateString
 			get { return " for update"; }
 		}
 
-		/// <summary> Is <tt>FOR UPDATE OF</tt> syntax supported? </summary>
-		/// <value> True if the database supports <tt>FOR UPDATE OF</tt> syntax; false otherwise. </value>
+		/// <summary>Is <c>FOR UPDATE OF</c> syntax supported?</summary>
+		/// <value><see langword="true"/> if the database supports <c>FOR UPDATE OF</c> syntax; <see langword="false"/> otherwise. </value>
+		public virtual bool ForUpdateOf
```
hazzik: `SupportsForUpdateOf`

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -526,8 +526,16 @@ public virtual string ForUpdateString
...
+			// By default, just check ForUpdateOfColumns. ForUpdateOf needs to be overriden only for dialect suppoting
+			// "For Update Of" on table aliases.
+			=> ForUpdateOfColumns;
+
+		/// <summary>Is <c>FOR UPDATE OF</c> syntax expecting columns?</summary>
+		/// <value><see langword="true"/> if the database expects a column list with <c>FOR UPDATE OF</c> syntax,
+		/// <see langword="false"/> if it expects table alias instead or do not support <c>FOR UPDATE OF</c> syntax.</value>
 		public virtual bool ForUpdateOfColumns
```
hazzik: `true`/`false` semantic does not make sense, at least with the current name of the property. Let's rename it to `RequiresColumnsForUpdateOf`, or similar if you know a better name.

fredericDelaporte: `UseColumnsWithForUpdateOf`?

PR #1588, opened 2018-03-01
Add a generic batcher for insert/update/delete statements, usable with PostgreSQL and others

maca88:
The batcher is working by concatenating multiple statements with a semicolon. In order to avoid having duplicate parameter names the `IDriver.GenerateCommand` method is used to generate a new command from the current one, where the passed `SqlString` is altered so that the parameter names will start from the latest parameter name of the batch command.

Update:
As the batcher does not use any PostgeSQL specifics and neither uses any reflection to do its job, it could be renamed to something more generic if desired. The batcher works on SQL Server (tested on 2012), so it can be used where `SqlClientBatchingBatcher` is not available with the limitation of [2100 parameters](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/ddd407e2-a20e-4340-bdf8-c4d48db72589/spexecutesql-limitation-?forum=transactsql).


hazzik:
> The batcher works on SQL Server (tested on 2012), so it can be used where SqlClientBatchingBatcher is not available with the limitation of 2100 parameters.

Can you add automatic flush if the limit is going to be reached?

fredericDelaporte:
> As the batcher does not use any PostgeSQL specifics and neither uses any reflection to do its job, it could be renamed to something more generic if desired.

This sounds as a good idea. According to [wikipedia](https://en.wikipedia.org/wiki/SQL#Syntax), the semi-colon is in the SQL standard (quite lame this standard is not public and forces to find "second hand" sources), so why not naming it `GenericBatchingBatcher`. (Drop the `Client`, it looks to me as a remnant of `SqlClient` naming, used by the SQL-Server client code.) Since at least some databases allow to redefine the statement terminaison character to something else than a semi-colon, maybe this character should be configurable. But well, it would be just a small nice to have.

But for this PR I think we should then not go more far than using it for PostgreSQL. Further enhancements for using it with other DB should in my opinion be done with another PR.

> Can you add automatic flush if the limit is going to be reached?

That 2100 limit is likely specific to SQL-Server, so it would have to be something applied only when this batcher is used for SQL-Server. I think it is a good idea, but for a complementary PR aiming at re-using this generic batcher for SQL-Server.

maca88:
> Can you add automatic flush if the limit is going to be reached?

Done with a test added.

> so why not naming it `GenericBatchingBatcher`

Renamed.

> Since at least some databases allow to redefine the statement terminaison character to something else than a semi-colon, maybe this character should be configurable

I added the character to be configurable on the generic batcher factory.

> But for this PR I think we should then not go more far than using it for PostgreSQL

As I did the proposed changes we could edit the title of this PR or create a new PR and close this one.

maca88:
Unfortunately `FirebirdSql.Data.FirebirdClient`, `Oracle.ManagedDataAccess` and `System.Data.SqlServerCe` clients do not support concatenating commands with a semicolon.
I did some research to see if there is something that we could do but I didn't found anything that would work correctly.
With `FirebirdClient`, the only way to execute multiple statements is via [EXECUTE BLOCK](https://firebirdsql.org/refdocs/langrefupd20-execblock.html) command, but the problem is that it is not possible to pass parameters as far as I tested.

With `Oracle.ManagedDataAccess` I was able to execute multiple statements by wrapping them into a [BEGIN END block](https://stackoverflow.com/a/20248824), but the problem is that the returned value of the modified rows is always `-1`.

With `SqlServerCe` I tried to add a `GO` command after each command as proposed [here](https://stackoverflow.com/a/6971516), but I didn't worked.

I've removed those three from tests and added a list of unsupported clients in the summary of the batcher.

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
```diff
@@ -0,0 +1,245 @@
...
+			: base(connectionManager, interceptor)
+		{
+			BatchSize = Factory.Settings.AdoBatchSize;
+			StatementTerminator = statementTerminator;
+			_currentBatch = new BatchingCommandSet(this);
+			// On Sql Server there is a limit of 2100 parameters, but 2 are reserved for sp_executesql, so
+			// we are able to use up to 2098 parameters. When sp_prepexec is used for preparing and executing
+			// statements then one more parameter is used. Set the max number of parameters to 2097 in order
+			// to ensure that we will never exceed the limit.
+			if (Factory.Dialect is MsSql2000Dialect)
```
hazzik: Could this be added to Dialect itself? Eg `public virtual int? MaxNumberOfParameters {get;}`?

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
```diff
@@ -176,9 +176,11 @@ public void Append(DbParameterCollection parameters)
 						_batcher.CurrentCommandParameterTypes);
 					UpdateCommandParameters(command, parameters);
 					_batchCommand.CommandText += $"{_batcher.StatementTerminator}{command.CommandText}";
-					foreach (DbParameter parameter in command.Parameters)
+					while (command.Parameters.Count > 0)
 					{
-						_batchCommand.Parameters.Add(CopyParameter(_batchCommand, parameter));
+						var pram = command.Parameters[0];
+						command.Parameters.RemoveAt(0);
```
bahusoid: Removing fist element is relatively slow for lists (as array is recopied on each call). It would be much more effective to remove last element instead. As It seems order of parameters is not important.

https://dotnetfiddle.net/ZKRkkb (it shows 4x difference, but on my machine in Release mode it's closer to 10x)

hazzik: Unfortunately, order of parameters _is_ important for some dialects

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
maca88: Added.

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
maca88: @bahusoid thanks for pointing it out! I redesigned the logic in order to call `GenerateCommand` only once and removed the code for removing parameters. After that I tested the difference and I was shocked when I saw the results. The new approach is 10x faster than the old one, [here](https://github.com/maca88/nhibernate-core/commit/7656d846b97b70e85237ce0eee1b1d37793dc1be) is how I setup the test. 
Old BatchingCommandSet took avg: `4959ms`
New BatchingCommandSet took avg: `435ms`

src/NHibernate/Driver/DriverBase.cs
```diff
@@ -178,13 +178,23 @@ protected virtual SqlStringFormatter GetSqlStringFormatter()
...
-		private void SetCommandParameters(DbCommand cmd, SqlType[] sqlTypes)
+		private void SetCommandParameters(DbCommand cmd, SqlType[] sqlTypes, Parameter[] parameters)
 		{
-			for (int i = 0; i < sqlTypes.Length; i++)
+			var usedParameters = new HashSet<int>();
+			for (var i = 0; i < sqlTypes.Length; i++)
 			{
-				string paramName = ToParameterName(i);
+				var parameter = parameters.Length > i ? parameters[i] : null;
+				var position = parameter?.ParameterPosition ?? i;
```
fredericDelaporte: I am suspicious about this change but I have not yet had the time to analyze and understand its rationals. It is not obvious why `sqlTypes` ordering is supposed to match `parameters` one. (Maybe a comment at line 136 could explain it.)

I do not have yet seen either in which case two parameters may declare the same position (could it be three by the way?), in which case parameter positions may be different of `sqlTypes` indexes, ...

maca88: From what I've seen by debugging, `parameters` were always ordered the same way as `sqlTypes` though there were some exceptions. For example when running [this](https://github.com/nhibernate/nhibernate-core/blob/1919480bff54c563902bdefabf462c26e4e45611/src/NHibernate.Test/NHSpecificTest/NH3634/FixtureByCode.cs#L92) test the number of `parameters` is two and `sqlTypes` is three, as the last parameter is `null`. There is [another special case](https://github.com/nhibernate/nhibernate-core/blob/ad4c2ef101cbf8ba798220973f4f78dd795e0896/src/NHibernate.Test/NHSpecificTest/NH1693/Fixture.cs#L6) that have duplicated parameters, where `ParameterPosition` is the same.
As I recently modified the batcher in order that only one command is created, this change is not needed anymore, so I will revert it in order to prevent having an unexpected breaking change.

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
```diff
@@ -0,0 +1,235 @@
...
+		private readonly int? _maxNumberOfParameters;
+		private readonly BatchingCommandSet _currentBatch;
+		private int _totalExpectedRowsAffected;
+		private StringBuilder _currentBatchCommandsLog;
+
+		public GenericBatchingBatcher(ConnectionManager connectionManager, IInterceptor interceptor, char statementTerminator)
+			: base(connectionManager, interceptor)
+		{
+			BatchSize = Factory.Settings.AdoBatchSize;
+			StatementTerminator = statementTerminator;
```
fredericDelaporte: It is possible to handle this as a setting, read through `Factory.Settings`. But it would have been better to have access to the configuration properties dictionary rather than adding a new property to `Settings` for such a specialized parameter. Unfortunately it seems not available.

So maybe is it better to keep current implementation as it is currently.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2446,6 +2446,11 @@ public virtual string LowercaseFunction
 		/// </summary>
 		public virtual int MaxAliasLength => 18;
 
+		/// <summary>
+		/// The maximum number of parameters allowed in a query.
+		/// </summary>
+		public virtual int? MaxNumberOfParameters { get; } = null;
```
fredericDelaporte: Why not, similarly to `MsSql2000Dialect`, write it:
```c#
public virtual int? MaxNumberOfParameters => null;
```
?

src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
```diff
@@ -0,0 +1,155 @@
...
+			{
+				s.SetBatchSize(5);
+				for (var i = 0; i < 20; i++)
+				{
+					s.Save(new VerySimple { Id = 1 + i, Name = $"Fabio{i}", Weight = 1.45 + i });
+				}
+				tx.Commit();
+
+				var log = sqlLog.GetWholeLog();
+				Assert.That(4, Is.EqualTo(FindAllOccurrences(log, "Batch commands:")));
```
fredericDelaporte: Looks a bit inverted, I would instead have written this:
```c#
Assert.That(FindAllOccurrences(log, "Batch commands:"), Is.EqualTo(4));
```

src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
```diff
@@ -0,0 +1,155 @@
...
+			using (var s = Sfi.OpenSession())
+			using (var tx = s.BeginTransaction())
+			{
+				for (var i = 0; i < totalRecords; i++)
+				{
+					s.Save(new VerySimple {Id = 1 + i, Name = $"Fabio{i}", Weight = 1.45 + i});
+				}
+				tx.Commit();
+			}
+			Assert.That(Sfi.Statistics.PrepareStatementCount, Is.EqualTo(1));
```
fredericDelaporte: This does not seem to check anything. Changing the batch size to 1 in `Configure` still allow the test to succeed this assert. That seems to be an old false assumption that this property would allow to get the number of command executed.

fredericDelaporte:
> With SqlServerCe I tried to add a GO command after each command as proposed here, but I didn't worked.

Indeed it can not, because the quetsion is about "SQL Compact 4.0 Toolbox", which is a client tool (the question title is misleading about this). The `GO` statement is a client command, I have never seen it understood "server" side.

src/NHibernate/Dialect/Dialect.cs
maca88: My bad, now is corrected.

src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
maca88: Fixed.

src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
maca88: I noticed that. The problem is that the only way to have this corrected (at least what I found) is to use `SqlLogSpy` as in the `BatchSizeTest` method, which I didn't use as it would flood the logs. Is there any other alternative?

fredericDelaporte: Unfortunately I do not know another way in current NHibernate state. And yes we also have a test log trouble, this log already being to much flooded by the tests, causing CI platforms log checking to be a bit painful. (Well, TeamCity is better than others on this subject.)

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
hazzik: Should come from the dialect.

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
```diff
@@ -0,0 +1,235 @@
...
+		{
+			var batchUpdate = CurrentCommand;
+			if (_maxNumberOfParameters.HasValue && 
+				_currentBatch.CountOfParameters + CurrentCommand.Parameters.Count > _maxNumberOfParameters)
+			{
+				ExecuteBatchWithTiming(batchUpdate);
+			}
+			_totalExpectedRowsAffected += expectation.ExpectedRowCount;
+			Driver.AdjustCommand(batchUpdate);
+			string lineWithParameters = null;
```
hazzik: Can logging (L59-74) please be moved to a separate method?

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
fredericDelaporte: This thing has come from a suggestion of mine. I was stating `;` is the standard, followed by all databases I have checked, but with some allowing it to be changed to something else. So I had written:

> Since at least some databases allow to redefine the statement terminaison character to something else than a semi-colon, maybe this character should be configurable. But well, it would be just a small nice to have.

Putting it in the dialect sounds indeed better. An overridable property in `BaseDialect` should be enough, those needing an other character would then be able of having it with a custom dialect. (If the need is frequent users could then ask for a configuration property, since the dialects are configurable.)

maca88: Moved into `BaseDialect` and used it also in `BasicResultSetsCommand`, when batching select statements.

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
maca88: Done.

PR #1592, opened 2018-03-02
Merge >net Core into MI 

tker:
Merge .Net Core


fredericDelaporte:
It looks to me you are trying to merge MI into NHibernate instead.

PR #1596, opened 2018-03-07
GH1594: AsyncLocal leak in SystemTransactionContext

### LINKED ISSUE #1594: AsyncLocal leak in SystemTransactionContext
The `SystemTransactionContext` uses following field: `private readonly AsyncLocal<bool> _bypassLock = new AsyncLocal<bool>();` By design, each creation of the 'SystemTransactionContext' (specifically the '_bypassLock' field) adds a new item into `ExecutionContext._localValues` dictionary and each assignment to `_bypassLock.Value` creates a copy of `ExecutionContext._localValues` dictionary. 

So if you run many transactions in one execution context (it can be a background job thread or main thread in a console application) the `ExecutionContext._localValues` dictionary gets bigger and bigger which causes 
1. Memory leak issues
2. Performance issues (as assigning to the '_bypassLock.Value' creates a copy of dictionary with a lot of items).

I have checked the code and at first glance it seems to me that the _bypassLock field could be made static as each thread gets its own copy of the execution context. I have created a unit test and a proposed fix (see [pull request 1596](https://github.com/nhibernate/nhibernate-core/pull/1596)). 

However I understand that the unit test is not 100% correct as it's using reflection to get the .NET system private dictionary (`ExecutionContext._localValues`) which might change in future versions of .NET, but I don't have an idea how to write the test more correctly.

Please check the proposed solution and / or comment the issue.

pecanw:
Fixes #1594.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/GH1594/ExecutionContextExtensions.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH1594/ExecutionContextExtensions.cs
@@ -1,0 +1,15 @@
+﻿using System.Collections;
+using System.Threading;
+
+namespace NHibernate.Test.NHSpecificTest.GH1594
+{
+	public static class ExecutionContextExtensions
+	{
+		public static int LocalValuesCount(this ExecutionContext c)
+		{
+			var f = typeof(ExecutionContext).GetField("_localValues", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
+			var d = (IDictionary) f.GetValue(c);
+			return d?.Count ?? 0;
+		}
+	}
+}
--- a/src/NHibernate.Test/NHSpecificTest/GH1594/ExecutionContextExtensions.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH1594/ExecutionContextExtensions.cs
@@ -7,7 +7,12 @@ namespace NHibernate.Test.NHSpecificTest.GH1594
 	{
 		public static int LocalValuesCount(this ExecutionContext c)
 		{
-			var f = typeof(ExecutionContext).GetField("_localValues", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
+#if NETSTANDARD1_3 || NETSTANDARD1_6 || NETSATNDARD2_0 || NETCOREAPP2_0
+			const string localValuesFieldName = "m_localValues";
+#else			
+			const string localValuesFieldName = "_localValues";
+#endif
+			var f = typeof(ExecutionContext).GetField(localValuesFieldName, System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
 			var d = (IDictionary) f.GetValue(c);
 			return d?.Count ?? 0;
 		}
--- a/src/NHibernate.Test/NHSpecificTest/GH1594/ExecutionContextExtensions.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH1594/ExecutionContextExtensions.cs
@@ -7,10 +7,10 @@ namespace NHibernate.Test.NHSpecificTest.GH1594
 	{
 		public static int LocalValuesCount(this ExecutionContext c)
 		{
-#if NETSTANDARD1_3 || NETSTANDARD1_6 || NETSATNDARD2_0 || NETCOREAPP2_0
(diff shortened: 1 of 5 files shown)


src/NHibernate.Test/NHSpecificTest/GH1594/ExecutionContextExtensions.cs
```diff
@@ -7,7 +7,12 @@ public static class ExecutionContextExtensions
 	{
 		public static int LocalValuesCount(this ExecutionContext c)
 		{
-			var f = typeof(ExecutionContext).GetField("_localValues", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
+#if NETSTANDARD1_3 || NETSTANDARD1_6 || NETSATNDARD2_0 || NETCOREAPP2_0
+			const string localValuesFieldName = "m_localValues";
```
fredericDelaporte: We support only 2.0. And we have defined a NETFX variable you can use instead, of course with opposite semantic.

src/NHibernate.Test/NHSpecificTest/GH1594/Fixture.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System.Linq;
+using System.Threading;
+using System.Transactions;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.GH1594
+{
+	[TestFixture]
+	public class Fixture : BugTestCase
```
fredericDelaporte: Since it uses system transaction, the fixture must be disabled if the environment does not support them. Add:

```c#
protected override bool AppliesTo(ISessionFactoryImplementor factory) =>
	factory.ConnectionProvider.Driver.SupportsSystemTransactions;
```

(Since currently this is false for all .Net Core / .Standard drivers, the private field fix is in fact not really needed, but it does not harm to have it.)

hazzik:
As I understand when one talks about memory leaks what they mean is that the objects survive the GC. The test does not consider any GC at all at the moment. 

We have discovered previously (for other matter) that `AsyncLocal<>` has heavy memory footprint, but in majority of the cases the memory becomes reclaimed after the GC.

src/NHibernate.Test/NHSpecificTest/GH1594/Fixture.cs
pecanw: Done, thanks for describing.

src/NHibernate.Test/NHSpecificTest/GH1594/ExecutionContextExtensions.cs
pecanw: Changed to use NETFX.

pecanw:
As for the @hazzik's comment: in this case every time when `AsyncLocal<bool> _bypassLock` is assigned a value the current `ExecutionContext._localValues` dictionary gets copied into a new one with a new / updated value of that `_bypassLock` object instance. The GC can collect the original dictionary, but the new copy (with the additional item in case of the first assignment to the `_bypassLock` instance) is referenced in the `ExecutionContext` and thus cannot be released by the GC.

fredericDelaporte:
This is true in both case, when the async local is static or not. But when it is not static, the current context dictionary references all past instances created and having with their value affected on the same thread, and grows endlessly till thread end.

> The test does not consider any GC at all at the moment.

GC can not change anything in the test, since the created asynclocal-s stay all at any point in time referenced in a dictionary, itself referenced in a instance referenced as a thread static (without weak references anywhere).

pecanw:
I have a question not related to this problem - I have the latest version of Resharper installed on my PC and it auto-updates the NHibernate.sln.DotSettings. I have unstaged it not to include it in the commit, but it is quite annoying to keep it it still in mind. May be other contributors have similar problem. Can you tell me if I
- should keep it it is
- include the change in the pull request
- create a new branch + pull request for this specific think
- just add a new item to NH backlog to let someone do it?

fredericDelaporte:
Yes we know, and there is the same trouble with Rider. It does that at almost each update, which is quite a nuisance. Sometimes we let it get updated with an unrelated PR.

hazzik:
>GC can not change anything in the test

It can demonstrate and prove that the claims are true.

fredericDelaporte:
For the record, async re-generation was forgotten is this PR and has been fixed with #1608. (See [contributing](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md) if you need more information on async generation.)

PR #1597, opened 2018-03-07
Add support for single-argument truncate to dialects that do not support it natively

fredericDelaporte:
* Support single parameter truncate, like round.
 * Add PostgreSQL truncate
 * Fix Sybase Anywhere round which was not supporting single parameter calls either.

This is a follow up to #1533, which had made some minimal adjustments on truncate for supporting its specific needs. Its subject was not truncate, so complete fix had to be done in another PR.


src/NHibernate/Dialect/PostgreSQLDialect.cs
```diff
@@ -69,7 +69,9 @@ public PostgreSQLDialect()
 			RegisterFunction("mod", new SQLFunctionTemplate(NHibernateUtil.Int32, "((?1) % (?2))"));
 
 			RegisterFunction("sign", new StandardSQLFunction("sign", NHibernateUtil.Int32));
-			RegisterFunction("round", new RoundFunction());
+			RegisterFunction("round", new RoundFunction(false));
+			RegisterFunction("truncate", new RoundFunction(true));
+			RegisterFunction("trunc", new RoundFunction(true));
```
fredericDelaporte: As written [here](https://github.com/nhibernate/nhibernate-core/pull/1533/files#r172564339), I consider supporting dialect specific names (`trunc` here) as a bad practice. But since there is some disagreement about this, I will not fight against it and so I am still adding it here.

src/NHibernate/Dialect/SybaseSQLAnywhere10Dialect.cs
```diff
@@ -140,12 +140,13 @@ protected virtual void RegisterMathFunctions()
 			RegisterFunction("radians", new StandardSQLFunction("radians", NHibernateUtil.Double));
 			RegisterFunction("rand", new StandardSQLFunction("rand", NHibernateUtil.Double));
 			RegisterFunction("remainder", new StandardSQLFunction("remainder"));
-			RegisterFunction("round", new StandardSQLFunction("round"));
+			RegisterFunction("round", new RoundEmulatingSingleParameterFunction("round"));
```
fredericDelaporte: Tested with an Anywhere 17, using [this custom driver](https://github.com/fredericDelaporte/nhibernate-core/blob/Anywhere17/src/NHibernate/Driver/SapSQLAnywhere17Driver.cs).

src/NHibernate/Dialect/SybaseSQLAnywhere10Dialect.cs
```diff
@@ -140,12 +140,13 @@ protected virtual void RegisterMathFunctions()
...
 			RegisterFunction("sign", new StandardSQLFunction("sign", NHibernateUtil.Int32));
 			RegisterFunction("sin", new StandardSQLFunction("sin", NHibernateUtil.Double));
 			RegisterFunction("sqrt", new StandardSQLFunction("sqrt", NHibernateUtil.Double));
 			RegisterFunction("tan", new StandardSQLFunction("tan", NHibernateUtil.Double));
-			RegisterFunction("truncate", new StandardSQLFunction("truncate"));
```
fredericDelaporte: Removed, it was an invalid registration.

src/NHibernate/Dialect/SybaseSQLAnywhere10Dialect.cs
```diff
@@ -343,8 +344,6 @@ protected virtual void RegisterMiscellaneousFunctions()
 			RegisterFunction("transactsql", new StandardSQLFunction("transactsql", NHibernateUtil.String));
 			RegisterFunction("varexists", new StandardSQLFunction("varexists", NHibernateUtil.Int32));
 			RegisterFunction("watcomsql", new StandardSQLFunction("watcomsql", NHibernateUtil.String));
-			RegisterFunction("truncnum", new StandardSafeSQLFunction("truncnum", 2));
-			RegisterFunction("truncate", new StandardSafeSQLFunction("truncnum", 2));
```
fredericDelaporte: Moved to `RegisterMathFunctions` (and adjusted).

hazzik:
I would not say that this is a bugfix. It's rather an improvement.

src/NHibernate/Dialect/Function/RoundFunction.cs
```diff
@@ -12,9 +12,20 @@ namespace NHibernate.Dialect.Function
...
+		private readonly ISQLFunction _singleParamRound;
+		private readonly ISQLFunction _round;
+		private readonly string _name;
 
-		private static readonly ISQLFunction Round = new StandardSQLFunction("round");
+		/// <summary>
+		/// Constructs a <c>RoundEmulatingSingleParameterFunction</c>.
+		/// </summary>
+		/// <param name="name">The SQL name of the round function to call.</param>
+		public RoundEmulatingSingleParameterFunction(string name)
```
hazzik: Class is public, so you need to add a default constructor.

fredericDelaporte: For avoiding a breaking change? This class has not yet been published.

src/NHibernate/Dialect/Function/RoundFunction.cs
```diff
@@ -24,9 +35,9 @@ public class RoundEmulatingSingleParameterFunction : ISQLFunction
 
 		public SqlString Render(IList args, ISessionFactoryImplementor factory)
 		{
```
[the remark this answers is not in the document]
fredericDelaporte: Either I do not get what you are thinking about, or it will be slightly more CPU lightweight but with a lot more code, rewriting a `SqlString` generation just for this case. For the later I do not think it is worth it.

```c#
var buf = new SqlStringBuilder();
buf.Add(_name)
   .Add("(");
for (int i = 0; i < args.Count; i++)
{
	object arg = args[i];
	if (arg is Parameter || arg is SqlString)
	{
		buf.AddObject(arg);
	}
	else
	{
		buf.Add(arg.ToString());
	}
	if (i < (args.Count - 1)) buf.Add(", ");
}

if (args.Count == 1)
	buf.Add(", 0");
return buf.Add(")").ToSqlString();
```

And this is an almost duplicate of `StandardSQLFunction`.

src/NHibernate/Dialect/Function/RoundFunction.cs
hazzik: Ok. Fine then.

src/NHibernate/Dialect/Function/RoundFunction.cs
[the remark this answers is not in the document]
hazzik: Why would you need a loop here? You have 2 arguments. 

```
return new SqlString(
  _name,
  "(",
  args[0],
  ", ",
  args.Count > 1 ? args[1] : "0",
  ")"
)
```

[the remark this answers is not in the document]
fredericDelaporte: This would be breaking for dialects supporting more than two arguments (SQL-Server `round` case).

hazzik:
Sent you a PR https://github.com/fredericDelaporte/nhibernate-core/pull/3

PR #1599, opened 2018-03-08
Full control of entities fetching in Criteria

bahusoid:
**Old behavior**: ManyToOne entities joined via ``JoinAlias/JoinQueryOver/Join...`` are always fetched.
OneToMany collections are only fetched when they are joined via LeftOuterJoin, but element entities are fetched anyway.
And you have no control over it.

**With this PR:**
You can explicitly control fetching behavior via setting `SelectMode` enum for criteria association paths. So with:
**1) `SelectMode.Fetch`** you can fetch current criteria/alias association path (if persister supports it). So now it's possible to fetch collections with inner or whatever join you want:
```C#
var list = session.QueryOver<EntityComplex>()
		.Fetch(SelectMode.Fetch, ec => ec.Child1)
		.JoinQueryOver(ec => ec.ChildrenList, JoinType.InnerJoin)
		//now we can fetch inner joined collection
		.Fetch(SelectMode.Fetch, childrenList => childrenList)
		.TransformUsing(Transformers.DistinctRootEntity)
		.List();
```
[link to test](https://github.com/bahusoid/nhibernate-core/blob/c62ba07cc140f60d5a232572e89f16ab2b1f0bcc/src/NHibernate.Test/Criteria/SelectModeTest/SelectModeTest.cs#L53-L59)

**2) `SelectMode.JoinOnly`** you can skip current criteria/alias association path to be fetched (so it won't be included in SELECT statement):
```C#
root = session.QueryOver(() => root)
		//Child1 is required solely for filtering, no need to be fetched, so skip it from select statement
		.JoinQueryOver(r => r.Child1, JoinType.InnerJoin)
		.Fetch(SelectMode.JoinOnly, child1 => child1)
		.Take(1)
		.SingleOrDefault();
```
[link to test](https://github.com/bahusoid/nhibernate-core/blob/c62ba07cc140f60d5a232572e89f16ab2b1f0bcc/src/NHibernate.Test/Criteria/SelectModeTest/SelectModeTest.cs#L32-L37)
**Note**: JoinOnly for root entity is not implemented (It's not trivial task and it's too much additional code change noise, so maybe later with another PR). But you can use `SelectMode.ChildFetch` instead (see below).

**3) `SelectMode.FetchLazyProperties`** you can fetch entities with lazy properties initialized:
```C#
var root = session.QueryOver<EntityComplex>()
			.Fetch(SelectMode.FetchLazyProperties, ec => ec)
			.Where(ec => ec.LazyProp != null)
			.Take(1)
			.SingleOrDefault();
```
[link to test](https://github.com/bahusoid/nhibernate-core/blob/c62ba07cc140f60d5a232572e89f16ab2b1f0bcc/src/NHibernate.Test/Criteria/SelectModeTest/SelectModeTest.cs#L109-L113)

**4) `SelectMode.Undefined`** - Basically no-op for queries:
```C#
//SelectMode.Undefined is no op - fetching is controlled by mapping:
//SameTypeChild won't be loaded, and ChildrenList collection won't be fetched due to InnerJoin
var list = session.QueryOver<EntityComplex>()
		.Fetch(SelectMode.Undefined, ec => ec.SameTypeChild)
		.JoinQueryOver(ec => ec.ChildrenList, JoinType.InnerJoin)
		.Fetch(SelectMode.Undefined, childrenList => childrenList)
		.TransformUsing(Transformers.DistinctRootEntity)
		.List();

//So it's a full equivalent of the following query (without SelectMode)
session.QueryOver<EntityComplex>()
		.JoinQueryOver(ec => ec.ChildrenList, JoinType.InnerJoin)
		.TransformUsing(Transformers.DistinctRootEntity);
```
[link to test](https://github.com/bahusoid/nhibernate-core/blob/c62ba07cc140f60d5a232572e89f16ab2b1f0bcc/src/NHibernate.Test/Criteria/SelectModeTest/SelectModeTest.cs#L80-L90)

**5) `SelectMode.ChildFetch`** - Experimental feature. The main idea is to allow effective multiple collections fetching in one database roundtrip. In this mode only entity Identifier columns are added to SELECT statement. Retrieved identifier is used only to obtain object from session and use it further for child entities fetching. If object is not found in session it's returned as null. Example of usage:
```C#
EntityComplex root = null;
var futureRoot = session.QueryOver(() => root)
			.Where(r => r.Id == _parentEntityComplexId)
			.FutureValue();
session
	.QueryOver(() => root)
	//Only ID is added to SELECT statement for root so it's index scan only
	.Fetch(SelectMode.ChildFetch, ec => ec)
	.Fetch(SelectMode.Fetch, ec => ec.ChildrenList)
	.Where(r => r.Id == _parentEntityComplexId)
	.Future();
session
	.QueryOver(() => root)
	.Fetch(SelectMode.ChildFetch, ec => ec)
	.Fetch(SelectMode.Fetch, ec => ec.ChildrenListEmpty)
	.Where(r => r.Id == _parentEntityComplexId)
	.Future();
root = futureRoot.Value;
```
[link to test](https://github.com/bahusoid/nhibernate-core/blob/c62ba07cc140f60d5a232572e89f16ab2b1f0bcc/src/NHibernate.Test/Criteria/SelectModeTest/SelectModeTest.cs#L131-L158)
Combining JoinOnly, ChildFetch and Fetch allows to do some crazy multilevel collections fetching in one database roundtrip:
```C#
EntityComplex root = null;
var rootFuture = session
			.QueryOver(() => root)
			.Future();

session
	.QueryOver(() => root)
	.Fetch(SelectMode.ChildFetch, () => root)
	.Fetch(SelectMode.Fetch, () => root.ChildrenList)
	.Future();
session
	.QueryOver(() => root)
	.Fetch(SelectMode.ChildFetch, () => root, () => root.ChildrenList)
	.Fetch(SelectMode.Fetch, () => root.ChildrenList[0].Children)
	.Future();
session
	.QueryOver(() => root)
	.Fetch(SelectMode.JoinOnly, () => root.ChildrenList)
	.Fetch(SelectMode.ChildFetch,() => root, () => root.ChildrenList[0].Children)
	.Fetch(SelectMode.Fetch, () => root.ChildrenList[0].Children[0].Children)
	.Future();
root = rootFuture.ToList().First(r => r.Id == _parentEntityComplexId);
```
[link to test](https://github.com/bahusoid/nhibernate-core/blob/c62ba07cc140f60d5a232572e89f16ab2b1f0bcc/src/NHibernate.Test/Criteria/SelectModeTest/SelectModeTest.cs#L247-L270)
So here `root.ChildrenList.Children.Children ` - 3 level deep collections are fetched in 1 db roundtrip.

**6) `SelectMode.Skip'**- skips join (and so fetching) from query. Useful only to disable eager mapping association.

[Functionality](https://github.com/bahusoid/nhibernate-core/blob/c62ba07cc140f60d5a232572e89f16ab2b1f0bcc/src/NHibernate/SelectModeExtensions.cs) is exposed via `Fetch` extension methods on `IQueryOver` interface taking current criteria  paths or aliased criteria paths as parameters:
```C#
queryOver
.Fetch(SelectMode.Fetch, curEntity => curEntity, curEntity => curEntity.ChildProp1, curEntity => curEntity.ChildProp2, ... )
.Fetch(SelectMode.JoinOnly, () => criteriaAlias, () => criteriaAlias.ChildProp1, () => criteriaAlis2.ChildProp1, ...)
.Fetch(SelectMode.FetchLazyProperties, ...)
```


src/NHibernate/ISupportSelectModeCriteria.cs
```diff
@@ -0,0 +1,16 @@
...
+	// 6.0 TODO: merge into 'ICriteria'.
+	public interface ISupportSelectModeCriteria
+	{
+		/// <summary>
+		/// Applies <paramref name="selectMode"/> for criteria with given <paramref name="alias"/> and given <paramref name="associationPath"/>
+		/// </summary>
+		/// <param name="selectMode"></param>
+		/// <param name="associationPath">Criteria association path. If empty root entity for given criteria is used.</param>
+		/// <param name="alias">Criteria alias. If empty current <see cref="ICriteria"/> criteria is used.</param>
+		void SetSelectMode(SelectMode selectMode, string associationPath, string alias);
```
hazzik: Shouldn't it return `ICriteria`?

bahusoid: Yeah. It should. Fixed.

fredericDelaporte:
The point 3 is NH-2888 (#972).

fredericDelaporte:
There is something forgotten: documenting the new feature in the reference documentation. I think this new feature deserve a mention into [criteria](https://github.com/nhibernate/nhibernate-core/blob/master/doc/reference/modules/query_criteria.xml) and [queryover](https://github.com/nhibernate/nhibernate-core/blob/master/doc/reference/modules/query_queryover.xml) chapters.

(Ok you have been notified of this already twice on two other already merged PR. If updating the doc here, do not do it for the other PR at the same time, because the other PRs should be documented in a 5.1.1 PR if possible, while this PR here would be merged in 5.2.)

hazzik:
Why the decision has been made to put this as a separate method? For me it's counterintuitive. Could it be just an additional parameter on existing methods (JoinQueryOver, etc.)?

bahusoid:
Because functionality is not bounded to (Join...) or any other existing methods. SelectMode can be applied for root query (or any other current criteria):
```
//fetch lazy properties for root and for root.ChildEntity
session.QueryOver<Entity>()
.With(SelectMode.FetchLazyProperties, 
   e => e,
   e => e.ChildEntity)...

//fetch lazy properties for current criteria  only - ChildEntity
session.QueryOver<Entity>()
.JoinQueryOver(e => e.ChildEntity)
.With(SelectMode.FetchLazyProperties, childEntity => childEntity)

```
But it also can be applied only to particular list of aliases (so select mode can be set in the end after query is written).
```
session.QueryOver<Entity>(()=>root)
.JoinQueryOver(e => e.ChildEntity)
.With(SelectMode.FetchLazyProperties, () => root)...
```
It's more like existing `Fetch` method but with more control. May be instead of `With` it can be exposed as `Fetch` override?

hazzik:
Approved in principle. However I'm still not convinced with the method name: I think that this feature would be hard to discover.

> It's more like existing Fetch method but with alias support and more control. May be instead of With it can be exposed as Fetch override?

Can you show some examples of what you mean?

bahusoid:
>I think that this feature would be hard to discover.

If you really need it - you will find it :) But if you have in mind a better name - just let me know.

@hazzik
Not sure which part of quoted text should I explain. But I meant that `SelectMode.Fetch` can fully replace existing `Fetch` method and adds additional functionality.
Code  below
```C#
query
.Fetch(p => p.Child1).Eager
.Fetch(p => p.Child2).Eager
```
can be rewritten as
```C#
query
.With(SelectMode.Fetch, p => p.Child1, p => p.Child2)
```
Plus it supports fetching by aliases (not supported by Fetch and criteria SetFetchMode methods);
```C#
NotAssociatedEntity notAssociatedAlias = null;
ChildEntity aliasForList = null;
query.JoinEntityAlias(() => notAssociatedAlias, ... )
.JoinAlias(p => p.ChildrenList, () => childList);
.With(SelectMode.Fetch, () => notAssociatedAlias.Child, () => childList)
```

So SelecMode currently is the only option for fetching inner joined collections and entity joins children. Plus all other SelectMode options explained in [pull request](https://github.com/nhibernate/nhibernate-core/pull/1599#issue-173824586). 

And instead of `With`  as an option I suggested to name it `Fetch`. So after rename it would be:
```C#
query
.Fetch(SelectMode.Fetch, p => p.Child1, p => p.Child2)
```
and
```C#
query.JoinEntityAlias(() => notAssociatedAlias, ... )
.JoinAlias(p => p.ChildrenList, () => childList);
.Fetch(SelectMode.Fetch, () => notAssociatedAlias.Child, () => childList)
```
But if you agree `Fetch` is a better name I need to check it doesn't break anything (like compilation issues due to ambiguous method resolution due to existing Fetch methods)

bahusoid:
rebased + fixed CriteriaImpl.Clone

src/NHibernate/Loader/SelectMode.cs
```diff
@@ -0,0 +1,32 @@
...
+		/// Fetch lazy properties.
+		/// </summary>
+		FetchLazyProperties,
+
+		/// <summary>
+		/// Only Identifier columns are added to select statement.
+		/// Use it for fetching child objects for already loaded entities.
+		/// Note: Missing entities in session will be returned as null
+		/// </summary>
+		ChildFetch,
```
fredericDelaporte: > Note: Missing entities in session will be returned as null

What does that mean? That using this mode could result in persisted entities to be returned as null if they were not already loaded in the session? If yes, this is a very unusual behavior for NHibernate. It should yield a proxy instead. (Or throw an error, or trigger a load, whatever, but guarantee no inconsistent data is returned.)

fredericDelaporte: > You should be ready for null if you are doing right join for instance.

Yes, but that does not look to me as a right join case here.

There are not any cases in NHibernate where a query result correctness depends on having already loaded these entities in the session. Yielding `null` in the query results instead of an entity (or proxy of an entity) because this entity was not already loaded in the session is not acceptable in my opinion.

fredericDelaporte: It appears to me that yielding `null` will moreover wreck child collection loading, defeating the purpose of the feature. See `Loader.ReadCollectionElement`, which is called on the row array. It may end up calling `EntityType.ResolveIdentifier` which has an early exit bypassing instantiating the element if the owner is null.

hazzik:
>About the naming, I am for changing With to Fetch, but also obsolete the previous Fetch, since the new one cover all its use cases.

💯% agree.

I would also request to change parameter positions:
```csharp
Fetch(path, SelectMode)
```

bahusoid:
@fredericDelaporte Wow. Thanks for you changes. If it works it would really be better approach. For some reasons I thought that another entity loading in loader will lead to multiple open  data readers that might not be supported by database. I've added test to ensure that this scenario works.

fredericDelaporte:
> For some reasons I thought that another entity loading in loader will lead to multiple open data readers that might not be supported by database.

Indeed, but this is already handled by NHibernate. See `IDriver.SupportsMultipleOpenReaders` and the `NDataReader`/`NHybridDataReader` which allow to workaround this limitation.

src/NHibernate/Loader/Loader.cs
```diff
@@ -937,6 +947,17 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
 					obj = session.GetEntityUsingInterceptor(key);
 					var alreadyLoaded = obj != null;
 					var persister = persisters[i];
+					if (IsChildFetchEntity(i))
+					{
+						if (!alreadyLoaded && mustLoadMissingEntity)
+						{
+							// Missing in session while its data has not been selected: fallback on immediate load
+							obj = session.ImmediateLoad(key.EntityName, key.Identifier);
```
fredericDelaporte: I have first attempted to do an `InternalLoad` here instead, without accounting for `returnProxies`/`mustLoadMissingEntity`, and without changing anything to `GetRow`caller. But this was (as expected) potentially yielding a proxy, which `GetRow` caller was not supporting: it was forcing the proxy initialization with itself as implementation, resulting in stack overflows when attempting to access the entity state.

For this reason, when proxy are supported, better yield `null` here and let `GetRow` caller handle returning a proxy.

bahusoid:
Regarding 
`Fetch(path, SelectMode)`
Yeah with name `Fetch` path is more logical as first param. But I think it's convenient to have path as params parameter - so you don't need to write repetitive code with many Fetch calls.

How about FetchWith(SelectMode, path1, path2,...)?

src/NHibernate/Loader/Loader.cs
```diff
@@ -937,6 +947,17 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
+						}
+						rowResults[i] = obj;
+						continue;
```
fredericDelaporte: Now completely skipping `InstanceAlreadyLoaded` and `CacheByUniqueKey` for child-fetch cases.
`InstanceAlreadyLoaded` handle the locking: should we lock a thing for which we only have selected its identifier, and which the query was designed to not fetch?
Likewise, handling caching for an entity which the query is not meant to actually load should not be done.

src/NHibernate/Loader/Loader.cs
```diff
@@ -372,13 +372,23 @@ internal static EntityKey GetOptionalObjectKey(QueryParameters queryParameters,
...
 					{
-						// Force the proxy to resolve itself
-						((INHibernateProxy)proxy).HibernateLazyInitializer.SetImplementation(entity);
-						row[i] = proxy;
+						object proxy = session.PersistenceContext.ProxyFor(persisters[i], keys[i], entity);
+
+						if (entity != proxy)
+						{
+							// Force the proxy to resolve itself
+							((INHibernateProxy) proxy).HibernateLazyInitializer.SetImplementation(entity);
```
fredericDelaporte: I was tempted to remove that line, because `ProxyFor` implementation already handles it. But well, that is not the subject of this PR.

fredericDelaporte:
I am for keeping the `params` and so for keeping the current ordering of `Fetch` parameters. Renaming it to `FetchWith` sounds not valuable to me: more wordy for little semantic gain.

bahusoid:
>I meant that SelectMode.Fetch can fully replace existing Fetch method
>but also obsolete the previous Fetch, since the new one covers all its use cases.

Actually I've realized `SelectMode` covers only eager fetching cases. The following case is not supported by SelectMode (just because I didn't think of this case):
`queryOver.Fetch(p => p.EagerList).Lazy;`

And I'm not sure that we should drop/obsolete existing `Fetch` method. It operates with `FetchMode` - enum that if I'm not mistaken is used in mappings for defining default "fetching strategy" and it's not obsolete.  So it's good to have method that allows to override it (as it might be also extended somehow in the future).

So 
1) Are you sure that existing `Fetch` should be obsoleted?
2) Should I add `SelectMode.Lazy` as equivalent of `FetchMode.Lazy`?

src/NHibernate/Impl/CriteriaImpl.cs
```diff
@@ -15,12 +16,13 @@ namespace NHibernate.Impl
 	/// Implementation of the <see cref="ICriteria"/> interface
 	/// </summary>
 	[Serializable]
-	public partial class CriteriaImpl : ICriteria, ISupportEntityJoinCriteria
+	public partial class CriteriaImpl : ICriteria, ISupportEntityJoinCriteria, ISupportSelectModeCriteria
 	{
 		private readonly System.Type persistentClass;
 		private readonly List<CriterionEntry> criteria = new List<CriterionEntry>();
 		private readonly List<OrderEntry> orderEntries = new List<OrderEntry>(10);
 		private readonly Dictionary<string, FetchMode> fetchModes = new Dictionary<string, FetchMode>();
```
fredericDelaporte: This should likely be removed, in favor of storing only select modes, with the obsoleted `Fetch` being translated to the new one for keeping it working.

src/NHibernate/ISupportSelectModeCriteria.cs
```diff
@@ -0,0 +1,18 @@
...
+	{
+		/// <summary>
+		/// Applies <paramref name="selectMode"/> for the criteria with the given <paramref name="alias"/> and the
+		/// given <paramref name="associationPath"/>.
+		/// </summary>
+		/// <param name="selectMode">The select mode to apply.</param>
+		/// <param name="associationPath">The criteria association path. If empty, the root entity for the given
+		/// criteria is used.</param>
+		/// <param name="alias">The criteria alias. If empty, the current <see cref="ICriteria"/> criteria is used.</param>
+		ICriteria SetSelectMode(SelectMode selectMode, string associationPath, string alias);
```
fredericDelaporte: Either rename to `Fetch` too or rename the `ICriteria` extension to `SetSelectMode`. Otherwise, once merged, the `ICriteria` API would expose two ways of doing exactly the same thing, which would be confusing for the user.

fredericDelaporte:
2. Yes.
1. If 2 is done, what will be missing?

> It operates with `FetchMode` - enum that if I'm not mistaken is used in mappings for defining default "fetching strategy" and it's not obsolete.

Yes but that is, in the QueryOver case, an internal implementation detail already quite hidden away from user. And the select mode does also operate the `FetchMode` almost the same way, isn't it? Keeping the two looks to me like we could ask for contradictory things, with the old `Fetch` currently having priority over the new (`CriteriaImp.GetFetchMode` code by example). This could lead to unexpected results for the user.

In the criteria case, I think `SetFetchMode` should be likewise obsoleted, especially since it exposes the whole `FetchMode` enumeration while only `Default`, `Join` and `Lazy` are actually supported, the other values being handled as `Lazy`. (See `CriteriaJoinWalker.GetJoinType`.)

bahusoid:
So it seems I've adressed all requested changes.

Some notes:
I've named `FetchMode.Lazy` and `FetchMode.Select` equivalent as `SelectMode.SkipJoin`.
I found that `Lazy` name is confusing. For instance for the following query:
```C#
s.QueryOver<Entity>().JoinQueryOver(e => e.Child).Fetch(SelectMode.Lazy, child => child)
```

I would expect inner join present in query - I just don't want it to be loaded. But in fact the way `FetchMode.Lazy` is implemented - it simply skips join from query. So `JoinQueryOver` logic is completely lost.
With `SkipJoin` name I think it's more clear.

I've added shortcuts for SelectMode.Fetch
```C#
criteria.Fetch("Child");

queryOver.Fetch(c => c.Child);
```
But QueryOver version is hidden by obsolete `Fetch` method in 5.2. So it will be exposed only in 6.0 with deletion of obsolete method.
And now shortcut version is available only for multiple properties:
```C#
queryOver.Fetch(c => c.Child, c => c.Child2 );
```

src/NHibernate/Impl/CriteriaImpl.cs
```diff
@@ -379,10 +377,28 @@ public ICriteria AddOrder(Order ordering)
 
 		public ICriteria SetFetchMode(string associationPath, FetchMode mode)
```
fredericDelaporte: This should be obsoleted too.

bahusoid:
>Looking at the list causes me to wonder, couldn't "Do not fetch, fetch children instead" and "Do not fetch, just join" be combined in a single SelectMode? 

But why? From implementation standpoint  "Do not fetch, just join" is simple change. Why to merge it with much more complex scenario? And you can't skip that way non-proxified objects:
```C#
//collection elements won't be loaded to session
s.QueryOver<Parent>().JoinQueryOver(l => l.ChildListNonProxified).Fetch(SelectMode.Skip);

//collection elements will be loaded to session with separate query for each object
s.QueryOver<Parent>().JoinQueryOver(l => l.ChildListNonProxified).Fetch(SelectMode.ChildFetch);
```

Regarding `JoinOnly` and `Skip` - OK. I will rename enums. It might indeed avoid possible confusions.

Regarding `AsMapped`. My only concern is multilevel fetch:
```C#
s.QueryOver<Parent>().Fetch( SelectMode.AsMapped, p => p.LazyChild, p => p.LazyChild.EagerSubChild);
```
Do you think it's clear that both `LazyChild` and `EagerSubChild` won't be fetched. IMHO `Undefined` in this context is better choice. As `AsMapped` might be confused with the following expectations:
```C#
s.QueryOver<Parent>()
.Fetch( SelectMode.ChildFetch, p => p.LazyChild)
.Fetch( SelectMode.Fetch, p => p.LazyChild.EagerSubChild);
```

Klausrex:
The new error message in IQueryOver.Fetch "Use Fetch(SelectMode mode, Expression<Func<TSubType, object>> path) instead" does not help a lot. A Fetch method with two parameters is not part of the interface IQueryOver so it could not been called.
Is this method just missing in the interface?

bahusoid:
It's not part of interface to avoid breaking changes (will be in 6.0) -  it's an extension method (in namespace NHibernate). See [here](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/SelectModeExtensions.cs)

Klausrex:
Ah thanks a lot!

lcetinapta:
Does this work on Lazy properties too? Because for mapping like this:
`    <property name="file_stream" type="BinaryBlob" length="2147483647" lazy="true" />`
and query like this:
`_session.QueryOver<StorageItem>().Fetch(SelectMode.FetchLazyProperties, x => x.file_stream).ListAsync()`
It does not select the lazy property.

bahusoid:
`SelectMode.FetchLazyProperties` should be applied to entity and not property
`_session.QueryOver<StorageItem>().Fetch(SelectMode.FetchLazyProperties, x => x)`

lcetinapta:
> `SelectMode.FetchLazyProperties` should be applied to entity and not property
> `_session.QueryOver<StorageItem>().Fetch(SelectMode.FetchLazyProperties, x => x)`

I see, that worked. Thanks.

PR #1600, opened 2018-03-08
Set MySqlClientBatchingBatcher as a default batcher for MySqlDataDriver

### LINKED ISSUE #1604: MySql batcher disables db exception translation
When an exception occurs in the database provider layer, NHibernate has a mechanism for translating it to a common format, the `GenericADOException`.

This mechanism is disabled when the MySql batcher is used, because this batcher catches the provider exception and encapsulates them in a `HibernateException`, which is ignored by NHibernate translation mechanism.

### LINKED ISSUE #1605: MySql batcher may attempt initiating a new batch without closing open reader first.
The MySql batcher misses a `CheckReaders` call when it initiate a new batch, causing it to fail when a datareader is already opened (which is possible through the use of `Enumerable` method on queries).

This is showcased on [this build](https://teamcity.jetbrains.com/viewLog.html?buildId=1343519&buildTypeId=bt868), errors on Legacy fixtures. (This build has the MySql batcher enabled, while this batcher is currently by default not enabled in NHibernate tests.)

fredericDelaporte:
Setting MySql batcher as the default batcher seems to have been forgotten when it was added. (At least no one answered my [question about this on its ticket](https://nhibernate.jira.com/browse/NH-2778?focusedCommentId=54804&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-54804).)

Also fix #1604 (no translation of data provider exception) and fix #1605 (may not close pending data reader before initiating a batch).

### DIFF
--- a/src/NHibernate/Driver/MySqlDataDriver.cs
+++ b/src/NHibernate/Driver/MySqlDataDriver.cs
@@ -1,4 +1,5 @@
 using System;
+using NHibernate.AdoNet;
 
 namespace NHibernate.Driver
 {
@@ -16,7 +17,7 @@ namespace NHibernate.Driver
 	/// for any updates and/or documentation regarding MySQL.
 	/// </para>
 	/// </remarks>
-	public class MySqlDataDriver : ReflectionBasedDriver
+	public class MySqlDataDriver : ReflectionBasedDriver, IEmbeddedBatcherFactoryProvider
 	{
 		/// <summary>
 		/// Initializes a new instance of the <see cref="MySqlDataDriver"/> class.
@@ -95,5 +96,7 @@ namespace NHibernate.Driver
 		// https://dev.mysql.com/doc/refman/5.7/en/datetime.html
 		/// <inheritdoc />
 		public override DateTime MinDate => new DateTime(1000, 1, 1);
+
+		System.Type IEmbeddedBatcherFactoryProvider.BatcherFactoryClass => typeof(MySqlClientBatchingBatcherFactory);
 	}
 }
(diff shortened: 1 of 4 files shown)


src/NHibernate/Driver/MySqlDataDriver.cs
```diff
@@ -95,5 +96,7 @@ public override bool SupportsMultipleQueries
 		// https://dev.mysql.com/doc/refman/5.7/en/datetime.html
 		/// <inheritdoc />
 		public override DateTime MinDate => new DateTime(1000, 1, 1);
+
+		System.Type IEmbeddedBatcherFactoryProvider.BatcherFactoryClass => typeof(MySqlClientBatchingBatcherFactory);
```
hazzik: This shall be guarded by `#if NETFX`

hazzik: Shall this be guarded by `#if NETFX`?

hazzik: Seems not. It's ok then.

fredericDelaporte:
It seems the MySql batcher causes issues not under .Net Core but under .Net Framework!

fredericDelaporte:
The MySql batcher is encapsulating all exceptions occurring in `ExecuteNonQuery` into a `HibernateException`. This prevents the exception translation mechanism to take place, which is a bug.

I am fixing it inside this PR because the bug is showcased by current tests only if the batcher is defined as the defaut for MySql.

fredericDelaporte:
> The six others are more problematic. They look like the errors we ignore in .Net Core builds. And we have also more errors in those builds now.

The MySql batcher misses a `CheckReader` call causing it to try initiating a new batch while a reader is still open on the connection, which MySql does not support. This is another bug of the batcher: #1605. I am willing to fix it likewise in this PR, which is now a finalization of [NH-2778](https://nhibernate.jira.com/browse/NH-2778) which has introduced this batcher.

fredericDelaporte:
This PR is back to 4 failures for the AppVeyor MySql build and 8 for the Travis one.

PR #1602, opened 2018-03-09
Preserve original snapshot mode.

fredericDelaporte:
As shown [here](https://ci.appveyor.com/project/nhibernate/nhibernate-core/build/5.1.0.104/job/bu3hoy20acramkyh/tests) or [here](https://ci.appveyor.com/project/nhibernate/nhibernate-core/build/5.1.0.105/job/gk3wn75135s1j248/tests), there are some tests which fail sometimes unexpectedly with an error stating `Snapshot isolation transaction failed accessing database 'nhibernate' because snapshot isolation is not allowed in this database.`.

There is one fixture which may disable it, because its setup always enables it and its teardown always disables it, as if assuming the feature can only be off by default. This PR fix it for having this fixture no more touching it if it is already enabled.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH1553/MsSQL/SnapshotIsolationUpdateConflictTest.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH1553/MsSQL/SnapshotIsolationUpdateConflictTest.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Data;
 using NHibernate.Cfg;
 using NHibernate.Dialect;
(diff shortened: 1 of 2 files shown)


src/NHibernate.Test/NHSpecificTest/NH1553/MsSQL/SnapshotIsolationUpdateConflictTest.cs
```diff
@@ -134,22 +136,42 @@ protected override bool AppliesTo(ISessionFactoryImplementor factory)
...
+					_isSnapshotIsolationAlreadyAllowed =
+						StringComparer.OrdinalIgnoreCase.Equals(command.ExecuteScalar() as string, "on");
+				}
+			}
+		}
+
 		private void SetAllowSnapshotIsolation(bool on)
 		{
 			using (ISession session = OpenSession())
 			{
```
hazzik: Can you remove curly braces from nested usings?

fredericDelaporte: Done.

PR #1607, opened 2018-03-12
Raise maximal decimal precision from 28 to 29

### LINKED ISSUE #1606: NHibernate 5 precision maximum on decimal vs. NHibernate 4 (MsSQL)
Hi,

I am facing an issue with the precision on an `System.Decimal` property.

I map a properties of `System.Decmial` (using FluentNHibernate) to `decimal(38,5)` in NHibernate 4.x, which works well.
Now I do the same thing with NHibernate 5.0.3, but I only get a maximum of `decimal(28,5)`.

(I use the ExportSchema to generate to data base an SQL Server 2016.)

Now trying to store a valid System.Decimal of the value `792281625142643375935439.50335m` (== `decimal.MaxValue / 10000m`) within the data base and I get an SqlException: `Arithmetic overflow error converting numeric to data type numeric.
The statement has been terminated.`

So in this test one has a valid `System.Decimal` with 29 significant digits and it would work with `decimal(29,5)`, but I can not set that precision either.

29 to 38 significant digits have the same space requirements within the data base (17 bytes, [decimal and numeric (Transact-SQL)](https://docs.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql)) so I use the maximum of 38.

I think it changed in "NH-4088 - Dialect.GetCastTypeName is buggy (#709)".

fredericDelaporte:
* Adjust default registrations of decimal and currency accross dialects for consistency.
 * Cease applying the decimal limit to Oracle double.
 * Fixes #1606

Details:

 * MySQL prior to version 5: it seems it was stored as string so with a huge precision, bumping 19 to 29.
 * SQLite: it does not handle precision/scale, it stores decimal in the same binary format than double. (Thus the `HasBrokenDecimal` set to true for SQLite, since it does not have a true decimal handling.) What should be changed then?
 * Oracle: double raised to 40 by the way, because it stores them as decimal but read them as .Net double, so it should not be limited to .Net decimal capabilities.
 * SybaseAsa9: default was 18,5 instead of 19,5.
 * SybaseAse15: default was 18,0 instead of 19,5. Set as 23,5 for avoiding a possible breaking change for those storing 10^18 numbers with the previous default.
 * Also standardizes Currency registration.

### DIFF
--- a/src/NHibernate.Test/DriverTest/SqlClientDriverFixture.cs
+++ b/src/NHibernate.Test/DriverTest/SqlClientDriverFixture.cs
@@ -161,7 +161,7 @@ namespace NHibernate.Test.DriverTest
 		[Test]
 		public void DefaultPrecisionScale()
 		{
-			const byte defaultPrecision = 28;
+			const byte defaultPrecision = 29;
 			const byte defaultScale = 10;
 			var driver = Sfi.ConnectionProvider.Driver;
 			try
--- a/src/NHibernate/Cfg/Environment.cs
+++ b/src/NHibernate/Cfg/Environment.cs
@@ -226,7 +226,7 @@ namespace NHibernate.Cfg
 
 		/// <summary>
 		/// Set the default precision used in casting when the target type is decimal and
-		/// does not specify it. <c>28</c> by default, automatically trimmed down according to dialect type registration.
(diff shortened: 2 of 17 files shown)


hazzik:
Need to update DriverTest.

PR #1610, opened 2018-03-13
Move MsSql constants from driver to dialect.

ngbrown:
This is a split off from #626 just incase I don't get that fully rebased in time for v5.1.  These changes are still valuable, but easy to review.

Also:
* Fix description of DB2400Driver.
* Driver code formatting tweaks.


hazzik:
> Move MsSql constants from driver to dialect

I don't understand what is the point of doing this?

fredericDelaporte:
It is required for externalizing the driver, which should be done for allowing people to not rely on reflection under .Net Core/Standard if they wish. (Unless it is considered as something to be supplied externally, but I am not agree with such a view. Supplying drivers with explicit dependencies for most common databases should be done by an ORM in my opinion.)

PR #1611, opened 2018-03-13
 Fix IFilter.SetParameterList not supporting HashSet<T> arguments

### LINKED ISSUE #1585: Hashset unsupported by SetParameterList
Hello,
I got a `GenericADOException` when enabling a filter in the session with `SetParameterList("FilterA", HashSet<Guid>)`.

When I query the session with a hql statement I got the `GenericADOException` and the
inner exception said `"System.Collections.Generic.HashSet``1[System.Guid]" cannot convert to type "System.String"`.

The reason seems to be that `HashSet<T>` doesn't implement the `ICollection` interface.

fredericDelaporte:
Fixes #1585

### DIFF
--- a/src/NHibernate.Test/FilterTest/DynamicFilterTest.cs
+++ b/src/NHibernate.Test/FilterTest/DynamicFilterTest.cs
@@ -1,13 +1,10 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using log4net;
 using NHibernate.Cache;
 using NHibernate.Cache.Entry;
 using NHibernate.Criterion;
-using NHibernate.Engine;
-using NHibernate.Persister.Collection;
 using NHibernate.Transform;
 using NUnit.Framework;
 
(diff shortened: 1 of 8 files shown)


src/NHibernate.Test/FilterTest/DynamicFilterTest.cs
```diff
@@ -354,6 +354,35 @@ public void InStyleFilterParameter()
...
+			using (var session = OpenSession())
+			{
+				Assert.Multiple(
+					() =>
+					{
+						session.EnableFilter("regionlist")
+						       .SetParameterList("regions", new HashSet<string> { "LA", "APAC" });
+
+						log.Debug("Performing query of Salespersons");
+						var salespersons = session.CreateQuery("from Salesperson").List();
```
fredericDelaporte: Before the fix, this case does not throw, but convert the hashset to a single string it uses as single argument in the `in` clause, causing it to fail retrieving expected results.

src/NHibernate.Test/FilterTest/DynamicFilterTest.cs
```diff
@@ -354,6 +354,35 @@ public void InStyleFilterParameter()
...
+						Assert.That(salespersons.Count, Is.EqualTo(1), "Incorrect salesperson count");
+
+						session.EnableFilter("guidlist")
+						       .SetParameterList("guids", new HashSet<Guid> { testData.Product1Guid, testData.Product2Guid });
+
+						log.Debug("Performing query of Products");
+						var products = session.CreateQuery("from Product").List();
```
fredericDelaporte: Before the fix, this case throws an invalid cast exception, which seems dependent on the dataprovider.

fredericDelaporte:
`TypedValue` is not involved in the cases I have reproduced, but it likely represents an issue for `FilterKey` equality. I could add a commit in this PR for it.

src/NHibernate/Engine/Query/NativeSQLQueryPlan.cs
```diff
@@ -151,19 +151,18 @@ private SqlString ExpandDynamicFilterParameters(SqlString sqlString, ICollection
...
 
 						// Add query chunk
 						string typeBindFragment = string.Join(", ", Enumerable.Repeat("?", parameterColumnSpan).ToArray());
 						string bindFragment;
-						if (collectionValue != null && !type.ReturnedClass.IsArray)
+						if (listValue != null && !type.ReturnedClass.IsArray)
 						{
-							collectionSpan = collectionValue.Count;
-							bindFragment = string.Join(", ", Enumerable.Repeat(typeBindFragment, collectionValue.Count).ToArray());
+							collectionSpan = listValue.Cast<object>().Count();
```
bahusoid: Wouldn't it result in full enumeration and unnecessary boxing for all struct elements?
If we still need to deal with IEnumerable interface I think it's better to add internal Count() extension:
```C#
internal static int Count(this IEnumerable source)
{
	if (source is ICollection col)
		return col.Count;

	int count = 0;
	var e = source.GetEnumerator();
	//in case it's generic disposable enumerator
	using (e as IDisposable)
	{
		while (e.MoveNext())
		{
			count++;
		}
	}
	return count;
}
```

fredericDelaporte: Yes you are right. But I was unwilling to add this to the recently obsoleted `EnumerableExtensions` utility.

bahusoid: It's obsoleted to be public. I think it's still can survive as internal helper.

fredericDelaporte: Commit amended.

fredericDelaporte:
>  I could add a commit in this PR for it.

Instead done as another PR (#1611) build over this one.

src/NHibernate/Impl/FilterImpl.cs
```diff
@@ -16,6 +17,7 @@ public class FilterImpl : IFilter
 		private FilterDefinition definition;
 
 		private readonly IDictionary<string, object> parameters = new Dictionary<string, object>();
+		private readonly IDictionary<string, IEnumerable> _parameterLists = new Dictionary<string, IEnumerable>();
```
hazzik: Why non generic `IEnumerable`?

src/NHibernate/Engine/Query/NativeSQLQueryPlan.cs
hazzik: >Wouldn't it result in full enumeration and unnecessary boxing for all struct elements?

No it would not.

src/NHibernate/Util/EnumerableExt.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using System.Collections;
+
+namespace NHibernate.Util
+{
+	internal static class EnumerableExt
+	{
+		internal static int Count(this IEnumerable source)
```
hazzik: This method takes precedence over standard `Enumerable.Count<T>` in some cases.

src/NHibernate/Impl/FilterImpl.cs
fredericDelaporte: Because it is not possible to put anything more specific.

src/NHibernate/Util/EnumerableExt.cs
fredericDelaporte: Fixed (forced push).

src/NHibernate/Engine/Query/NativeSQLQueryPlan.cs
bahusoid: >Not in every case.
Could you give an example? I'm looking at Cast and Count implementation and don't see any shortcuts.

hazzik: >Could you give an example?

#### Cast

`IEnumerable<T>` is co-variant. `Cast` has [a shortcut for it](https://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,1023):
```csharp
public static IEnumerable<TResult> Cast<TResult>(this IEnumerable source) {
  IEnumerable<TResult> typedSource = source as IEnumerable<TResult>;
  if (typedSource != null) return typedSource;
  if (source == null) throw Error.ArgumentNull("source");
  return CastIterator<TResult>(source);
}
```
If collection is generic, cast will chose this path. `CastIterator` would be involved only if collection is not generic.

In our case we first cast `ICollection<T>` to `IEnumerable` then do `.Cast<T>` which will do the upcast to `IEnumerable<T>` without involving `CastIterator`.

#### Count

Count [has shortcuts also](https://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,1305):

```
public static int Count<TSource>(this IEnumerable<TSource> source) {
  if (source == null) throw Error.ArgumentNull("source");
  ICollection<TSource> collectionoft = source as ICollection<TSource>;
  if (collectionoft != null) return collectionoft.Count;
  ICollection collection = source as ICollection;
  if (collection != null) return collection.Count;
  int count = 0;
  using (IEnumerator<TSource> e = source.GetEnumerator()) {
    checked {
      while (e.MoveNext()) count++;
    }
  }
  return count;
}
```

Because we did not perform `CastIterator` it will just call `.Count` on our collection.

hazzik: It seems I misread your comment. Indeed for *structs* it will always result in a cast.

src/NHibernate/Impl/FilterImpl.cs
hazzik: `System.Array` is possible. But it will require to "materialize" the incoming collection.

fredericDelaporte:
5.1.1 or 5.2? (I think this non-regression bug fix is simple enough for 5.1.1.)

hazzik:
I've made a PR: https://github.com/fredericDelaporte/nhibernate-core/pull/4

src/NHibernate/Impl/FilterImpl.cs
```diff
@@ -102,12 +102,12 @@ public object GetParameter(string name)
 		}
 
 		/// <summary>
-		/// Get a span of collection parameter by name. <see langword="null" /> if the parameter is not a collectino or
-		/// there is no such parameter exist.
+		/// Get the span of a value list parameter by name. <see langword="null" /> if the parameter is not a value list
+		/// or if there is no such parameter.
```
fredericDelaporte: I have changed "collection" for "value list" for consistency with the xml comment of `SetParameterList` above.

fredericDelaporte:
I intend to squash the last three commits in one with Alexander as co-author.

fredericDelaporte:
Now that we have started merging 5.2 stuff in master, I am keen to putting only regressions in 5.1.x, in order to limit the likelihood of having conflicts when merging it back to master.

fredericDelaporte:
Squashed and 5.1.x rebased, waiting builds. (Well, maybe not TeamCity which has a two days long queue.)

PR #1612, opened 2018-03-13
 Fix TypedValue not always using adequate comparer with SetParameterList

fredericDelaporte:
`TypedValue` may use the default comparer instead of the list comparer when a list parameter is supplied with a `HashSet`. This will then result in undue cache miss.

~~This PR is build over #1611 (two first commit currently appearing here).~~


src/NHibernate.Test/CacheTest/FilterKeyFixture.cs
```diff
@@ -8,105 +9,129 @@ namespace NHibernate.Test.CacheTest
...
-			fk = new FilterKey(filterName, f.Parameters, f.FilterDefinition.ParameterTypes);
-			Assert.That(fk.ToString(), Is.EqualTo("FilterKey[DescriptionEqualAndValueGT{'pDesc'='something', 'pValue'='10'}]"));
+			fk = new FilterKey(f);
+			Assert.That(fk.ToString(), Is.EqualTo("FilterKey[DescriptionEqualAndValueGT{'pDesc'='something', 'pValue'='10'}]"), "Value");
 		}
 
 		[Test]
 		public void Equality()
 		{
-			// Equality is aware only by parameters names not values
```
fredericDelaporte: Wrong, but not contradicted by the test because the test was bugged.

src/NHibernate.Test/CacheTest/FilterKeyFixture.cs
```diff
@@ -8,105 +9,129 @@ namespace NHibernate.Test.CacheTest
...
 		{
 			const string filterName = "DescriptionLike";
 			var f = new FilterImpl(Sfi.GetFilterDefinition(filterName));
 			f.SetParameter("pLike", "so%");
-			fk = new FilterKey(filterName, f.Parameters, f.FilterDefinition.ParameterTypes);
+			fk = new FilterKey(f);
 
 			var f1 = new FilterImpl(Sfi.GetFilterDefinition(filterName));
-			f1.SetParameter("pLike", "%ing");
-			fk1 = new FilterKey(filterName, f.Parameters, f.FilterDefinition.ParameterTypes);
```
fredericDelaporte: Here lies the test bug, affecting all tests: the second key was construct with the first filter values...

src/NHibernate.Test/CacheTest/FilterKeyFixture.cs
```diff
@@ -8,105 +9,129 @@ namespace NHibernate.Test.CacheTest
...
+		private void FilterValueInToCompare(out FilterKey fk, out FilterKey fk1, bool sameValue)
+		{
+			const string filterName = "ValueIn";
+			var f = new FilterImpl(Sfi.GetFilterDefinition(filterName));
+			f.SetParameterList("pIn", new HashSet<int> { 10, 11 });
+			fk = new FilterKey(f);
 
 			var f1 = new FilterImpl(Sfi.GetFilterDefinition(filterName));
-			f1.SetParameter("pDesc", "something").SetParameter("pValue", 11);
-			fk1 = new FilterKey(filterName, f.Parameters, f.FilterDefinition.ParameterTypes);
```
fredericDelaporte: Same test bug than in `FilterDescLikeToCompare`, fixed too.

src/NHibernate.Test/CacheTest/FilterKeyFixture.cs
```diff
@@ -8,105 +9,129 @@ namespace NHibernate.Test.CacheTest
...
 		}
 
 		[Test]
 		public void Equality()
 		{
-			// Equality is aware only by parameters names not values
-			FilterKey fk, fk1;
-			FilterDescLikeToCompare(out fk, out fk1);
-			Assert.That(fk, Is.EqualTo(fk1));
+			FilterDescLikeToCompare(out var fk, out var fk1, true);
```
fredericDelaporte: They can be equal only with same value. The `NotEquality` test is completed with an assert showcasing that a different value causes inequality.

src/NHibernate.Test/CacheTest/FilterKeyFixture.cs
```diff
@@ -8,105 +9,129 @@ namespace NHibernate.Test.CacheTest
...
+			Assert.That(fk, Is.Not.EqualTo(fvk), "fk & fvk");
+			Assert.That(fk1, Is.Not.EqualTo(fvk1), "fk1 & fvk1");
+			Assert.That(fvk, Is.Not.EqualTo(fik), "fvk & fik");
+			Assert.That(fvk1, Is.Not.EqualTo(fik1), "fvk1 & fik1");
 		}
 
 		[Test]
 		public void HashCode()
 		{
-			// HashCode is aware only by parameters names not values (should work as Equal)
```
fredericDelaporte: Unfortunately true for the current hashcode implementation, which is bad for performances. But in fact it should ideally not be equal, since actually filter with different parameters values are not equals.

src/NHibernate.Test/CacheTest/QueryKeyFixture.cs
```diff
@@ -14,95 +14,91 @@ public class QueryKeyFixture : TestCase
...
-			var fk =  new FilterKey(filterName, f.Parameters, f.FilterDefinition.ParameterTypes);
+			var fk =  new FilterKey(f);
 			ISet<FilterKey> fks = new HashSet<FilterKey> { fk };
 			qk = new QueryKey(Sfi, SqlAll, new QueryParameters(), fks, null);
 
 			var f1 = new FilterImpl(Sfi.GetFilterDefinition(filterName));
-			f1.SetParameter("pLike", "%ing");
-			var fk1 = new FilterKey(filterName, f.Parameters, f.FilterDefinition.ParameterTypes);
+			f1.SetParameter("pLike", sameValue ? "so%" : "%ing");
+			var fk1 = new FilterKey(f1);
```
fredericDelaporte: Same test bug as in `FilterKeyFixture`.

hazzik:
Needs to be rewritten as 5.1.2 is merged to master now.

fredericDelaporte:
Done.

src/NHibernate/Cache/FilterKey.cs
```diff
@@ -13,6 +13,8 @@ public class FilterKey
 		private readonly string _filterName;
 		private readonly Dictionary<string, TypedValue> _filterParameters = new Dictionary<string, TypedValue>();
 
+		// Since v5.1
```
hazzik: 5.2

src/NHibernate/Engine/TypedValue.cs
```diff
@@ -90,15 +95,15 @@ private int GetHashCode(IType type, ICollection values)
...
 					return true;
 
 				if (x == null || y == null)
 					return false;
 
-				if (x.Count != y.Count)
+				if (x is ICollection xCol && y is ICollection yCol && xCol.Count != yCol.Count)
 					return false;
 
 				var ye = y.GetEnumerator();
```
hazzik: Here is a bug:
```csharp
IEnumerable<T> ToLazyEnumerable<T>(params T[] array)
{
	// ReSharper disable once LoopCanBeConvertedToQuery
	foreach (var item in array)
		yield return item;
}

var v1 = new TypedValue(NHibernateUtil.Int32, ToLazyEnumerable(3, 2, 1, 0, 0, 0), true);
var v2 = new TypedValue(NHibernateUtil.Int32, ToLazyEnumerable(3, 2, 1, 0), true);

Warn.If(v1.GetHashCode() == v2.GetHashCode(), "Equal hash codes");
Assert.AreNotEqual(v1, v2);
```

fredericDelaporte:
Rebased and hashcode tests enabled.

(All commits are to be squashed in one.)

src/NHibernate/Criterion/InExpression.cs
```diff
@@ -135,15 +135,15 @@ private List<TypedValue> GetParameterTypedValues(ICriteria criteria, ICriteriaQu
...
+						list.Add(new TypedValue(types[i], subval, false));
 					}
 				}
 
 				return list;
 			}
 			else
 			{
-				return _values.Select(v => new TypedValue(type, v)).ToList();
+				return _values.Select(v => new TypedValue(type, v, false)).ToList();
```
hazzik: Not sure about this...

fredericDelaporte: `_values` is the collection of values to be put in the `in` clause. They must be elements, not themselves lists.

src/NHibernate/Engine/TypedValue.cs
```diff
@@ -66,50 +81,60 @@ public bool Equals(TypedValue x, TypedValue y)
 				if (y == null) return false;
```
hazzik: Can we add reference equals short-cut?

hazzik: And for DefaultComparer as well

src/NHibernate/Engine/TypedValue.cs
```diff
@@ -17,15 +17,32 @@ public sealed class TypedValue
...
+		/// Construct a typed value.
+		/// </summary>
+		/// <param name="type">The type of the value (or of its elements if it is a list value)</param>
+		/// <param name="value">The value.</param>
+		/// <param name="isList"><see langword="true" /> if the value is a list value (for a parameter list),
+		/// <see langword="false" /> otherwise.</param>
+		public TypedValue(IType type, object value, bool isList)
+		{
+			if (isList && value != null && !(value is IEnumerable))
+				throw new ArgumentException($"{nameof(value)} must be an {nameof(IEnumerable)} when {nameof(isList)} is true", nameof(value));
```
fredericDelaporte: In case some mistake is done later, better check `value` is at least `IEnumerable` when `isList` is true. (The list comparator will consider as always equals values which are both not `IEnumerable`.)

src/NHibernate/Engine/TypedValue.cs
```diff
@@ -63,53 +80,66 @@ public class ParameterListComparer : IEqualityComparer<TypedValue>
 		{
 			public bool Equals(TypedValue x, TypedValue y)
 			{
-				if (y == null) return false;
+				if (ReferenceEquals(x, y))
+					return true;
```
fredericDelaporte: Shortcut added, although I do not expect it to be much useful, since its main usage is in query cache key comparison. Unless a not originally cached query instance is re-executed a second time (which is likely a code smell), typed values will be different instances.

src/NHibernate/Engine/TypedValue.cs
```diff
@@ -63,53 +80,66 @@ public class ParameterListComparer : IEqualityComparer<TypedValue>
...
+				if (x == null || y == null)
+					return false;
```
fredericDelaporte: With its current usage, `x` is never supposed to be `null`. But as the comparer class is publicly accessible, better still check this.

PR #1618, opened 2018-03-17
Fix NuGet push script

fredericDelaporte:
The nuget push bat is generating obsolete command lines which are no more accepted by NuGet.
The release procedure is lacking a point and has a typo.

### DIFF
--- a/default.build
+++ b/default.build
@@ -319,10 +319,21 @@
 			<in>
 				<items>
 					<include name="${nuget.nupackages.dir}/*.nupkg"/>
+					<exclude name="${nuget.nupackages.dir}/*.symbols.nupkg"/>
 				</items>
 			</in>
 			<do>
-				<echo message="nuget push ${filename} ${environment::newline()}" file="${nuget.nupackages.pushbatfile}" append="true"/>
+				<echo message="nuget push -source https://nuget.org/ ${path::get-file-name(filename)} ${environment::newline()}" file="${nuget.nupackages.pushbatfile}" append="true"/>
+			</do>
+		</foreach>
+		<foreach item="File" property="filename">
+			<in>
+				<items>
+					<include name="${nuget.nupackages.dir}/*.symbols.nupkg"/>
+				</items>
+			</in>
+			<do>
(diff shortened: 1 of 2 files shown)


default.build
```diff
@@ -319,10 +319,21 @@
 			<in>
 				<items>
 					<include name="${nuget.nupackages.dir}/*.nupkg"/>
+					<exclude name="${nuget.nupackages.dir}/*.symbols.nupkg"/>
 				</items>
 			</in>
 			<do>
-				<echo message="nuget push ${filename} ${environment::newline()}" file="${nuget.nupackages.pushbatfile}" append="true"/>
+				<echo message="nuget push -source https://api.nuget.org/v3/index.json ${path::get-file-name(filename)} ${environment::newline()}" file="${nuget.nupackages.pushbatfile}" append="true"/>
```
fredericDelaporte: Destination ("source") is become mandatory. [Documentation](https://docs.microsoft.com/en-us/nuget/create-packages/publish-a-package) uses this v3 url, aligning on it.

Adding also `path::get-file-name` because otherwise this was emitting the full path to the package, which is specific to the build machine and will very likely not be right after having downloaded the artifacts from TeamCity.

default.build
```diff
@@ -319,10 +319,21 @@
 			<in>
 				<items>
 					<include name="${nuget.nupackages.dir}/*.nupkg"/>
+					<exclude name="${nuget.nupackages.dir}/*.symbols.nupkg"/>
```
fredericDelaporte: Due to having to specify the destination, symbols package needs now a different command line.

default.build
hazzik: As I recall: `https://nuget.org/` should be enough to push. Also, if it is just nuget.org the symbols would be pushed automatically.

fredericDelaporte: I will test this on NHibernate.Caches.

fredericDelaporte: It does not push symbols package. This feature has been removed. NuGet/Home#6113

fredericDelaporte:
Rebased and NuGet url simplified.

hazzik:
@fredericDelaporte could you please give more descriptive title to the PR?

PR #1620, opened 2018-03-18
GH1149 one-to-one related object not loaded with stateless session

### LINKED ISSUE #1149: NH-3391 - StatelessSession: one-to-one detail-object is always null
**Alex Kravchenko** created an [issue](https://nhibernate.jira.com/browse/NH-3391) — <time datetime="2013-02-11T04:53:08.151Z" title="2013-02-11T04:53:08.151Z">11th February 2013, 4:53:08</time>:
> After reading one-to-one related objects from database (within StatelessSession) detail-object is lost.
> Of course this works fine in regular Session.
> 
> //OneToOneType.Hydrate method gets related object identifier from session, but StatelessSession.GetContextEntityIdentifier(owner) always returns null, so it's impossible to read such objects properly from database
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2013-02-11T05:03:07.381Z" title="2013-02-11T05:03:07.381Z">11th February 2013, 5:03:07</time>:
> Does it work in in 3.2.0 - 3.3.3CR1 ?
> 
---
**Alex Kravchenko** added a comment — <time datetime="2013-02-11T07:42:07.651Z" title="2013-02-11T07:42:07.651Z">11th February 2013, 7:42:07</time>:
> found in 3.3.1, tested on last sources
>

seamus-hoban:
Fixes #1149

Here's my attempt to correct the above issue.
All unit tests run successfully, but I'm not sure that my change won't have other side-effects given that the fix is pretty much a crib from the stateful session, so any pointers would be most welcome.

Hopefully I've followed the contribution guidelines correctly.


fredericDelaporte:
Well, I was believing stateless sessions were not handling any kind of association, but the [documentation](http://nhibernate.info/doc/nhibernate-reference/batch.html#batch-statelesssession) only writes about collections. Granted, I have almost never used a stateless session.

Does `many-to-one` works? If no, having only `one-to-one` supported would be a bit awkward.

seamus-hoban:
I believe so - the project that I was working on where I came across the 1:1 bug was with a hierarchy that had a many to one relationship that was being loaded correctly.
I have added an object with a many-to-one relationship and a test for that to my code, should I update the pull-request with it?

fredericDelaporte:
No, no need. Ideally such a test should already exist somewhere in NHibernate tests, but I have not checked.

seamus-hoban:
I think the mappings and tests in NH1574 show retrieval of an object with a many-to-one association (SpecializedPrincipal). (StatelessTest.cs)

src/NHibernate.Test/NHSpecificTest/GH1149/Mappings.hbm.xml
```diff
@@ -0,0 +1,25 @@
...
+<hibernate-mapping xmlns="urn:nhibernate-mapping-2.2"
+										namespace="NHibernate.Test.NHSpecificTest.GH1149"
+										assembly="NHibernate.Test">
+	<class name="Company" >
+		<id name="Id">
+			<generator class="native"/>
+		</id>
+		<property name="Name" />
+
+		<one-to-one name="Address" class="Address" property-ref="Company" cascade="all-delete-orphan"/>
```
fredericDelaporte: May you add a test with an actual `one-to-one` in db? I mean, a `one-to-one` not using a `property-ref` but with one entity pk being also the pk of the other entity (so a pk which is also a fk), both side mapped as `one-to-one` with one side being `inverse`. (`PersonTrueO2O` with `AddressTrueO2O` in NH3931 tests is an example of this.)

Unless this case is already tested in NHibernate.Test and working (meaning the issue report was not really precise).

hazzik:
From my point of view the changes are correct and consistent with other places where associations are involved.

src/NHibernate.Test/NHSpecificTest/GH1149/Mappings.hbm.xml
seamus-hoban: Done, PR updated.

fredericDelaporte:
Some notes of interest (which do not infirm this PR validity overall):

I have checked the `many-to-one` case: it does not work by default, because the association is loaded as a proxy and stateless session does not allow initializing a proxy. So it fails as soon as an access to a member of the association other than the id is done.

The `one-to-one` case is special in this respect, because it cannot be proxified from the non-constrained side. (It does not know if it exists without querying its table, so it loads it instead.)  So from the non-constrained side, it is always directly loaded, instead of being proxified.

But the constrained side fails like a `many-to-one` by default.

Doing more checks, if the `many-to-one` is set as `lazy="false"`, it does work (even without this PR), so we can consider that supporting accessing a non-constrained `one-to-one` on an entity loaded from a stateless session is consistent with the `many-to-one` case, because a non-constrained `one-to-one` is non lazy by default.

src/NHibernate/Impl/StatelessSessionImpl.cs
```diff
@@ -228,7 +228,20 @@ public override void AfterTransactionCompletion(bool successful, ITransaction tx
 
 		public override object GetContextEntityIdentifier(object obj)
 		{
-			return null;
+			using (BeginProcess())
+			{
+				if (obj.IsProxy())
```
hazzik: I gave it another thought. I think the `true` branch is wrong/not possible/not needed.

hazzik:
Need async tests generation.

hazzik:
Hi @seamus-hoban. Please drop the last commit and regenerate the async tests. To regenerate async tests:

#### On windows:

1. Run ShowBuildMenu.bat 
2. Press `H` on keyboard (option "H. Generate async code (Generates files in Async sub-folders)").
3. Wait the tool to complete
4. Commit and push generated code 

#### On other platforms (mono)

1. Navigate to `./Tools` folder
2. Run `nuget install`
3. Navigate to `./src` folder
4. Run `mono ../Tools/CSharpAsyncGenerator.CommandLine.0.8.2/tools/AsyncGenerator.CommandLine.exe`
5. Wait the tool to complete
6. Commit and push generated code

seamus-hoban:
Okay, will do sorry.

fredericDelaporte:
The first error is usual (at least on my machine) and does not prevent the task to execute. I do not know about the second one, I have never seen it. Is your tooling up to date?

I have pushed the async regeneration.

fredericDelaporte:
Ok, I guess you have updated Visual Studio to its yesterday release, 15.6.3. With 15.6.2, it was still working. Now that I have updated it myself, it is broken.

```
generate-async:

     [exec] 11:37:31,760 INFO  CommandLine:0 - AsyncGenerator
     [exec]
     [exec] 11:37:32,072 INFO  AsyncCodeGenerator:0 - Generating async code started
     [exec] 11:37:32,650 INFO  AsyncCodeGenerator:0 - Opening project 'E:\Projets\nhibernate\nhibernate-core\src\NHibernate\NHibernate.csproj' started
     [exec] 11:37:33,374 FATAL CommandLine:0 - System.InvalidOperationException: One or more errors occurred while opening the project:
     [exec] Msbuild failed when processing the file 'E:\Projets\nhibernate\nhibernate-core\src\NHibernate\NHibernate.csproj' with message: MSB0001: Internal MSBuild Error: Missing resource ''
     [exec] Hint: For suppressing irrelevant errors use SuppressDiagnosticFailures option.
     [exec]    à AsyncGenerator.AsyncCodeGenerator.CheckForErrors(MSBuildWorkspace workspace, String itemType, ImmutableArray`1 supressFailuresPredicates, ILogger logger)
     [exec]    à AsyncGenerator.AsyncCodeGenerator.<OpenProject>d__8.MoveNext()
     [exec] --- Fin de la trace de la pile à partir de l'emplacement précédent au niveau duquel l'exception a été levée ---
     [exec]    à System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
     [exec]    à System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     [exec]    à AsyncGenerator.AsyncCodeGenerator.<GenerateAsync>d__0.MoveNext()
     [exec] --- Fin de la trace de la pile à partir de l'emplacement précédent au niveau duquel l'exception a été levée ---
     [exec]    à System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
     [exec]    à System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     [exec]    à AsyncGenerator.CommandLine.Program.Main(String[] args)
```

@maca88, do you have the issue too?

seamus-hoban:
Thank you all!
Yeah 15.6.3 screws up any building from within Studio, but command line build is fine, but the AsyncGenerator stuff is as per your comment above. There's a thread on the developercommunity.visualstudio where are few others are affected.

fredericDelaporte:
If you are speaking of this [thread](https://developercommunity.visualstudio.com/content/problem/209642/msb0001-internal-msbuild-error-missing-resource.html)? I think it is another (maybe related) trouble, since it has been opened before 15.6.3 release and is about project creation.

Otherwise could you link it?

I have not seen building in VS is broken too, I use primarily Rider. And well, I can still build with VS, even after having fully cleaned obj/bin folders in case it was reusing some assets compiled through Rider.

fredericDelaporte:
@hazzik, do you think this change is sufficiently mundane for going into 5.1.1? Or should it wait 5.2?

seamus-hoban:
@fredericDelaporte This thread: https://developercommunity.visualstudio.com/content/problem/218326/after-1563-publish-fails-due-to-error-projectasset.html
I can't build within studio even after deleting bin and obj folders.

hazzik:
This is kind-a a bug, so fine if it will be in 5.1.1.

maca88:
@fredericDelaporte Yes, it happened also to me after the upgrade. [The solution](https://github.com/maca88/AsyncGenerator/commit/a8d81c9a6e5080091772e65bb6a509a414423459) was to upgrade the `Microsoft.Build` package to the latest version as there is a [bug with the older MSBuild](https://github.com/Microsoft/msbuild/issues/2997#issuecomment-371196871). I released a  new version `0.10.0`, that includes the upgrade of  `Microsoft.Build`.

PR #1625, opened 2018-03-24
Output only netstandard2.0 and net461 libraries

ngbrown:
Remove netcoreapp2.0 output for the primary NHibernate library.

I think the netcoreapp2.0 output was there because of trying to build the solution to a specific framework and on Linux/macOS it would fail because NETFX wasn't available.  A better way is to build and run the test assemblies in one step, targeting the desired framework.

This will reduce the size of the resulting NuGet file and decrease the build time some.

The netstandard2.0 output wasn't being selected for testing and there are some issues with `Trim` reflection for Linq then needed to be corrected.


hazzik:
netcoreapp2.x assembly will likely come back in the future (most likely in 5.2). And there is very little benefit of removing it now.

>I think the netcoreapp2.0 output was there because of trying to build the solution to a specific framework and on Linux/macOS it would fail because NETFX wasn't available.

Not because of that. 

There are some differences between the .netstandard 2.0 and .netcoreapp2.0 assemblies: additional `string.Trim` overloads, System.Transactions, etc.

ngbrown:
Right now there are three "platforms" that NHibernate.dll assembly is targeting, however only two are tested.

> There are some differences between the .netstandard 2.0 and .netcoreapp2.0 assemblies: additional `string.Trim` overloads, System.Transactions, etc.

I took care of the `string.Trim` in my original pull request from last year; the same solution is in this pull request.

No changes needed to be made for System.Transactions for the tests to pass unchanged, so I'm not sure what that is referring to.  .netstandard 2.0 is not the end-runtime, it's a set of common API's, and the non-full framework version isn't going beyond the common API's so there's no need to also build a .netcoreapp2.0 version, when the application (or tests) will (and should be) be targeting .netcoreapp2.0.

hazzik:
@fredericDelaporte if you do not mind, I'm going to apply 313585e6e2721b38d87ac93e3714bc8b99fbc408 to the master

fredericDelaporte:
> @fredericDelaporte if you do not mind, I'm going to apply 313585e6e2721b38d87ac93e3714bc8b99fbc408 to the master

Ok.

ngbrown:
Re-based because a commit from this pull request was cherry-picked into `master`.

Also split "Use only the NETFX define not and not NETSTANDARD2_0 or NETCOREAPP2_0" into it's own commit.  No code changes from original pull request.

psake.ps1
```diff
@@ -102,7 +102,7 @@ Task Build {
     }
 }
 
-Task Test -depends Build  {
+Task Test -depends Build {
```
ngbrown: Running the "build" on Linux without Mono available meant that `dotnet461` wasn't available, so the build would fail, even though it isn't needed to run the tests when targeting `netcoreapp2.0`.  That was why building/testing in one step was the best way to success in all cases, because if `netcoreapp2.0` was available, then it would be able to build the dependencies.

I realize you modified the Travis CI scripts to include Mono, but it shouldn't be strictly necessary.

psake.ps1
```diff
@@ -102,15 +102,15 @@ Task Build {
     }
 }
 
-Task Test {
+Task Test -depends Build  {
```
ngbrown: Running the "build" on Linux without Mono available meant that `dotnet461` wasn't available, so the build would fail, even though it isn't needed to run the tests when targeting `netcoreapp2.0`.  That was why building/testing in one step was the best way to success in all cases, because if `netcoreapp2.0` was available, then it would be able to build the dependencies.

I realize you modified the Travis CI scripts to include Mono, but it shouldn't be strictly necessary.

_Sorry, put the previous comment on the wrong commit, so it made less sense._

psake.ps1
ngbrown: Running the "build" on Linux without Mono available meant that `dotnet461` wasn't available, so the build would fail, even though it isn't needed to run the tests when targeting `netcoreapp2.0`.  That was why building/testing in one step was the best way to success in all cases, because if `netcoreapp2.0` was available, then it would be able to build the dependencies.

I realize you modified the Travis CI scripts to include Mono, but it shouldn't be strictly necessary.

~_Sorry, put the previous comment on the wrong commit, so it made less sense._~ _Never-mind, it has to be on the most recent of that line change to be visible._

ngbrown: If you want the CI tasks to build for all frameworks, my proposal would be to add `Build` to the `-TaskList` in each of the CI scripts.  If the project is already built, then `dotnet run` would have minimal work to do and should not need the `--no-build` option.

Separately in #1626, the command-line `ShowBuildMenu.bat` uses the psake Test to build/run all combinations of frameworks and bitness

hazzik: >If the project is already built, then dotnet run would have minimal build work to do and should not need the --no-build option.

In theory. But in real life it does build again.

You can check this by adding `-v n` to the run command.

ngbrown:
I think this is ready for merging to master.  I don't have any problems with the changes @hazzik made.

hazzik:
.NET Core 2.0 has some methods which are not available in .NET/.NET Standard 2.0 which I want to support. See #1682 for example.

PR #1626, opened 2018-03-25
Use dotnet test for running all tests.

ngbrown:
Since the vstest runner output is lighter weight, only showing logged messages on test not passing, the tests shouldn't be encountering out of memory errors when writing the XML file.

The XML file is NUnit v3 format, so not sure if TeamCity can accept it.  TeamCity should still be getting real-time test updates through integration with `TeamCity.VSTest.TestAdapter`.

AppVeyor will also get real-time test updates through integration with `Appveyor.TestLogger`.

The command line build menu can now build and test both .net framework and .net core tests (both x86 and x64).


hazzik:
The Sql Server test on Travis-CI will fail now. And will always fail.

hazzik:
As I told.

ngbrown:
stalled build... I had this initially working against SQL server for Linux, but it started doing this and I thought it was because an updated version wasn't working so well against the Ubuntu 14.04 (not technically supported) that Circle-CI was running.

hazzik:
>the vstest runner output is lighter weight, only showing logged messages on test not passing,

This is the default behaviour of NUnitLite as well. The option to output all tests to console was chosen deliberately. For one of the reasons see below.

>the tests shouldn't be encountering out of memory errors when writing the XML file.

It will write all the tests to the xml file anyway, so this is not related in any way to the fact that the successful tests are not displayed into the console.

>TeamCity should still be getting real-time test updates through integration with TeamCity.VSTest.TestAdapter.
>AppVeyor will also get real-time test updates through integration with Appveyor.TestLogger.

TC is getting the test as you run output from NUnitLite as well. However, because tests take 10-40 minutes to run, the ability to see tests as you go is seldom needed as no one is watching tests as they run.

However, previous behaviour (when test reported to the console/log as they go) is valuable as it allows one to analyse [accidentally] interdependent tests (like the ones with System.Transaction failures).

ngbrown:
> The Sql Server test on Travis-CI will fail now (in about 5 minutes from now). And will always fail.

SQL server passes on Travis CI.  And finishes much quicker.  (14 minutes vs. 35 minutes)

> This is the default behaviour of NUnitLite as well. The option to output all tests to console was chosen deliberately.

I could not find a way to prevent console output (coming from ether `Console.WriteLine` or through log4net) on passing tests for either nunit-console or nunitlite.  Also the messages are stuffed into the XML report, which is causing intermittent out of memory errors for the x86 runs on TeamCity.

> ...allows one to analyse [accidentally]...

Accidentally isn't a good plan.

hazzik:
It is

>[accidentally] interdependent tests

Not

> analyse [accidentally]

ngbrown:
I'm just trying to help.  There are still other ways to run the tests like with nunit lite and nunit-console, as well as the test runners in Visual Studio and ReSharper.

I think this solves a few current problems with the builds and makes things more predictable for most development.

hazzik:
@fredericDelaporte if you do not mind, I'm going to apply 5747c82fec9aa3c4942093123730c813f7f45348 to the master

ngbrown:
Re-based because a commit from this pull request was cherry-picked into `master`.

Also split "Check for running from testhost on all frameworks." and "No longer post-process test results and ignore certain failures" into separate commits.  No code in pull request.

teamcity.build
```diff
@@ -136,8 +124,6 @@
 				<include name="*.dll"/>
 			</fileset>
 		</copy>
```
hazzik: @fredericDelaporte shall we remove the oracle step (and binaries) completely? It wasn't build by TeamCity for more than a year.

psake.ps1
```diff
@@ -102,15 +104,30 @@ Task Build {
     }
 }
 
-Task Test -depends Build {
+Task Test {
```
ngbrown: Since the Test task is used in the `ShowBuildMenu.bat` tool, building all frameworks when trying to test only one is excessive.  The `Build` task has been added to the task list of each of the CI scripts.

teamcity.build
fredericDelaporte: Non-built on TeamCity Oracle targets are: `oracle`, `oracle32`, `oracle64` and `oracle-managed64`. Only `oracle-managed32` is built.

Why not removing the unused targets. (`oracle32` and `oracle-managed64` are available as TeamCity built options in `ShowBuildMenu.bat`, and would have to be removed from there too, if removed from `teamcity.build`.)

About the binaries, they also serve for the `E. Add a test configuration for Oracle.` option of `ShowBuildMenu.bat`. So either this option is kept for people wishing to test with the unmanaged driver and then the binaries are to stay, or this option has to be removed too.

fredericDelaporte:
#1775 has taken another approach for reducing the memory pressure on the TeamCity build agent, and so far it seems to work.

Now the PR here may still optimize a bit more builds and tests. But as 5.2 is overdue, should we keep it in the 5.2 milestone?

PR #1627, opened 2018-03-26
Refactored session List<T> method  for Criteria

bahusoid:
Don't fallback `IList<T>` to `IList` implementation.

Motivation: 
To make further refactorings easier (non-generic List methods can be simply dropped);
Give more control in method implementation. Having actual `List<T>` class instead of `IList` allows to use its specialized methods unavailable for interface (like `AddRange` etc)


hazzik:
> Optimized

Can you provide any numbers?

bahusoid:
@hazzik Yeah  I guess "Optimized" is not quite right word here. "Refactored" is more appropriate -  I think it's better to make non-generic implementation be based on generic. I actually don't expect any noticeable performance gain (maybe only for huge lists)  - but I will check.

bahusoid:
Here's some numbers:
For [.NET Core](https://tio.run/##lVTfTyIxEH5m/4oJTyV6C@qbcCQemouJvyIk92B8KLsD10u33WuLHjH@7Tj9ISBg5EjY3XZmvu@b6UwL@63QBheLmRVqCsO5dVh1s/VVPtBSYuGEVvZzS/4TFRpRbHicCz5V2jpRbMZeCfW3m2WKV2hrXiCcleU9V1O8QzPRFVcFZi9Zo56NpSigkNxauDN6aniVNciwtBC5A6Ec3MyqMZrbyYXECpWz8B2OOuFHPEt/6zipgSctSrjmQjHrDMl6eARuprZFjh68MSBYLTH/ZYRD0oqsGfFBTwATwyk04WCLNx/pYcCkkE6z1QrsjSduwJAkhc9AeZa6YmRJhonRVbJdCet6evyH6tpnm9AxYqINMJ@xoJhOl169gJAPOFVSuDltHRz4VGIujWCk@jKT3@A/tw0bcV@XQiWJOFoXRGz9Nb3efvyF/WSX/ZPS3iqE8Rw0vZiWJYiqjtK4b65WMyLH/mEjUaHJh44bt5YjpXcmJQH9mNODhQwOQ1nWk9vBTYFgfOMBI7mt0/3JAsnxXiQDTk3KpQTmdP7e6SwdmnV5nv8PsQ8J5Cdb5OFB1UajuPzQ7DG0N@qzcCKjPjh9CJd@EUHeW39Es@DB0/mNHqJMPVPuMfZfXNfzkQ4qDqETFazn5g1h9ytFPpk9VO2sm49r7cWybIwI/QkJTRby4jc1YRhA0HSxLD3SEUQhTL8XPfw9fbvdhp6dVRU3835cDp2un7kjxGfD6xpN8mqv3OLNFk4bTuHyXNhaWz6WGG@5lMkKyNJXOpnlZpqsdMMFLLbMyQfk92jRpQENG6Gx2Kp0H2/HKCf4wAtM0T8NuplRsSUiQxdeYT04lDsmgBv8Xmqi356OC8lri@W1kNQ0SBd6aelabUJlmyuBSVmikppmZFfchtpAvsMvSifk12yxeAM):

Number of elements: 1,000,000
One by one (old implementation)
33 ms
Add range (new):
13 ms
Cast all (to.AddRange(from.Cast...):
37 ms

For [Full .NET Framework](https://dotnetfiddle.net/9WpUJi)
Number of elements: 100,000
One by one (old implementation)
8 ms
Add range (new):
8 ms
Cast all (to.AddRange(from.Cast...):
9 ms

I'm providing numbers for .NET Core as MS optimized AddRange for ICollection<T> (and I guess this optimization will be merged to full .NET framework sometime later).

.NET fiddle doesn't allow bigger numbers to make AddRange win but at least it shows it doesn't make things worse.

src/NHibernate/Util/ArrayHelper.cs
```diff
@@ -114,6 +114,13 @@ public static void AddAll<T>(IList<T> to, IList<T> from)
 				to.Add(obj);
 		}
 
+		internal static void AddAll<T>(List<T> to, IList from)
+		{
```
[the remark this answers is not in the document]
bahusoid: To have performance gain we need `from` to be ICollection\<T\>. But in our case `from` is  List\<object\>. So `Cast` will lead to worse performance then original implementation.

I've updated code to avoid second data copy by adding generic wrapper over 'ICollection'

src/NHibernate/Async/Impl/AbstractSessionImpl.cs
```diff
@@ -78,17 +80,16 @@ public abstract partial class AbstractSessionImpl : ISessionImplementor
 			}
 		}
 
+		//TODO 6.0: Make virtual
```
hazzik: You can make it virtual in 5.2

src/NHibernate/Util/CollectionHelper.cs
```diff
@@ -579,6 +579,77 @@ public IEnumerator GetEnumerator()
...
+				_col = col;
+			}
+
+			public int Count => _col.Count;
+
+			public bool IsReadOnly => true;
+
+			public bool Contains(T item)
+			{
+				foreach (var obj in _col)
```
hazzik: Shouldn't it just call `_col.Contains(item)`?

src/NHibernate/Async/Impl/AbstractSessionImpl.cs
hazzik: >Does it make sense?

Yes, it does

src/NHibernate/Util/CollectionHelper.cs
bahusoid: There is no ICollection.Contains. Though could be _col.Cast\<T\>().Contains. Should I made this change?

hazzik: No. Thats ok.

bahusoid:
>I dislike a bit the CollectionWrapper\<T\>

Well.. I can just minimize number of required memory allocations by updating list.Capacity instead of using AddRage. So no dependency on proper CopyTo implementation :) and no need in CollectionWrapper

fredericDelaporte:
I do not really get how you would implement `ArrayHelper.AddAll<T>` with your last comment. May you add a code snippet?

bahusoid:
Well. Could be something like this:
```C#
internal static void AddAll<T>(List<T> to, ICollection<IList> lists)
{
	if (lists.Count == 0)
		return;
 
	var totalCount = lists.Sum(list => list.Count);
	if (to.Capacity < totalCount)
	{
		to.Capacity = totalCount;
	}
 
	foreach (IList loadedList in lists)
	{
		AddAll(to, loadedList);
	}
}
```

And List\<T\> call would look something like:
```C#
List<IList> loadedLists = Factory
		.GetImplementors(criteria.EntityOrClassName)
		.Reverse()
		.Select(implementor => new CriteriaLoader(GetOuterJoinLoadable(implementor), Factory, criteria, implementor, EnabledFilters))
		.Select(criteriaLoader => criteriaLoader.List(this))
		.ToList();
 
ArrayHelper.AddAll(
	results,
	loadedLists);
```

fredericDelaporte:
The `CollectionWrapper` is there for `internal static void AddAll<T>(List<T> to, IList from)` implementation. The code above does not show how you would change its implementation if `CollectionWrapper` is removed. And by the way, why is it on `ArrayHelper` rather than `CollectionHelper`? It does neither take any array as argument nor yield one.

bahusoid:
I meant that `AddAll<T>(List<T> to, ICollection<IList> lists)` will be used instead of added here  `internal static void AddAll<T>(List<T> to, IList from)`. So it will call already existing (and that's why it's inside `ArrayHelper`) :
https://github.com/nhibernate/nhibernate-core/blob/92de0b749c712678fe9397e1fa514bd6d8e83d7c/src/NHibernate/Util/ArrayHelper.cs#L102

bahusoid:
To make it more clear I can just push a temp commit with changes I have in mind.

fredericDelaporte:
No need. But now I realize that this `AddAll` seems to be just an additional micro-optimization, not a change required by this PR. And by the numbers you provide, it may just save 20ms for one million of elements, which will likely take quite considerably more time to get loaded first. So that optimization looks a bit moot. Can we just stick to the old implementation, adding one by one, instead of adding many lines of code for such a micro-optimization?

src/NHibernate/Async/Impl/CriteriaImpl.cs
```diff
@@ -28,33 +28,29 @@ public partial class CriteriaImpl : ICriteria, ISupportEntityJoinCriteria
...
 		{
 			cancellationToken.ThrowIfCancellationRequested();
-			var results = new List<object>();
-			await (ListAsync(results, cancellationToken)).ConfigureAwait(false);
-			return results;
+			return (await (ListAsync<object>(cancellationToken)).ConfigureAwait(false)).ToIList();
 		}
 
 		public async Task ListAsync(IList results, CancellationToken cancellationToken = default(CancellationToken))
+		{
```
hazzik: Is this method used?

src/NHibernate/Impl/CriteriaImpl.cs
```diff
@@ -253,31 +253,27 @@ public ICriteria Add(ICriterion expression)
 
 		public IList List()
 		{
-			var results = new List<object>();
-			List(results);
-			return results;
+			return List<object>().ToIList();
 		}
 
 		public void List(IList results)
```
hazzik: Is this method used?

fredericDelaporte: Internally, still used by `CriteriaImpl.Subcriteria.List(IList results)`, since it is implemented by delegation. But indeed `ICriteria.List(IList results)` could now be obsoleted for removal in 6.0, leaving only the generic list method. (There is no `IList List()` method in this interface, contrary to `IQuery` which has `void List(IList results)` and `IList List()`. But on `ISessionImplementor`, there is still a `IList List(CriteriaImpl criteria)`, only referenced by xml comments: maybe it should be obsoleted too.)

By the way, `IQuery.List(IList results)` has also no internal usage. `IList IQuery.List()` has five internal usages, which could all be replaced by calls to `List<object>`. But it has also 908 usages in tests. A similar PR for `IQuery` would be a good thing.

fredericDelaporte: Well, there is still a `IList List()` on `ICriteria`, I have missed it the other day.

src/NHibernate.Test/NHSpecificTest/NH1393/Fixture.cs
```diff
@@ -52,7 +52,7 @@ public void CanSumProjectionOnSqlFunction()
 					s.CreateCriteria(typeof (Person)).SetProjection(
 						Projections.Sum(Projections.SqlFunction(arithmaticAddition, NHibernateUtil.GuessType(typeof (double)),
 						                                        Projections.Property("IQ"), Projections.Property("ShoeSize"))));
-				IList list = c.List();
+				var list = c.List<object>();
```
fredericDelaporte: I think, for sake of "giving right examples", we should use a more specialized type. If you agree, I will add a revert commit then I will take the time of putting those more specialized types everywhere.

src/NHibernate/Loader/Criteria/CriteriaLoader.cs
```diff
@@ -12,6 +13,28 @@
 
 namespace NHibernate.Loader.Criteria
 {
+	internal static partial class CriteriaLoaderExtensions
+	{
+		/// <summary>
+		/// Loads all loaders results to single typed list
+		/// </summary>
+		internal static List<T> LoadAllToList<T>(this IList<CriteriaLoader> loaders, ISessionImplementor session)
```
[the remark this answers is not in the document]
bahusoid: I can't use `AddRange` here because I need to add `IList` not `IList\<T\>`. And it's List<object> not List<T> underline.

src/NHibernate/ICriteria.cs
```diff
@@ -8,6 +8,14 @@
 
 namespace NHibernate
 {
+	public static partial class CriteriaExtensions
+	{
+		public static IList<object> List(this ICriteria criteria)
```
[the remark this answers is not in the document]
bahusoid: It's something I forgot to delete.

src/NHibernate.Test/NHSpecificTest/NH1393/Fixture.cs
bahusoid: Well I agree it makes sense in our `main` tests but I think it's not worth to update legacy bug related tests in NHSpecificTest folder. But I'm OK if you are willing to do it.

src/NHibernate/Loader/Criteria/CriteriaLoader.cs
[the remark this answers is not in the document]
fredericDelaporte: `.AddRange(loader.List(this).Cast<T>())` will do it. Anyway the `ArrayHelper.Add(IList, IList)` which is still used in `LoadAllToList` does that casting.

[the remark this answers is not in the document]
bahusoid: Yeah. I meant it won't preallocate. So no difference with AddAll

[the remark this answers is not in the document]
fredericDelaporte: Due to the casting, right, it would need to be written as `.AddRange(loader.List(this).Cast<T>().ToList())` for benefiting of `AddRange`, but then another list would be created in the process.

So now this change looks to me as a workaround optimization for not having the loader generic.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1737,7 +1737,8 @@ public override IList<T> List<T>(CriteriaImpl criteria)
 			}
 		}
 
-		//TODO 6.0: Remove (use base class implementation)
+		//Since 5.3
+		[Obsolete("Please use List<object> instead.")]
 		public override void List(CriteriaImpl criteria, IList results)
```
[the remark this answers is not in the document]
bahusoid: Why resolved? Did you already made those changes?

src/NHibernate/Loader/Criteria/CriteriaLoader.cs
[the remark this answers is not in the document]
bahusoid: Well it's one place that works effectievly. It can be simplified/inlined when generic version is available. I made a look sometime ago - and it's not simple change at all.

src/NHibernate/Impl/SessionImpl.cs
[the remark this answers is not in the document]
fredericDelaporte: Locally, yes. But if you push anything before, I will pull and adapt what needs to.

src/NHibernate.Test/NHSpecificTest/NH1393/Fixture.cs
fredericDelaporte: Done. I have checked them with PostgreSQL and SQL-Server. Maybe some other db may fail.

fredericDelaporte:
I am not merging this since @hazzik has self requested himself a review for this PR.

bahusoid:
@hazzik Can you please fulfill your request for review? :) Just do it per commit so you can skip "Obsolete non generic Criteria List methods (tests)" related changes

src/NHibernate.Test/Criteria/DetachedCriteriaSerializable.cs
```diff
@@ -391,13 +391,16 @@ public void ExecutableCriteria()
 			SerializeAndList<Student>(dc);
 
 			// Subquery
-			dc = DetachedCriteria.For(typeof(Student))
-				.Add(Property.ForName("StudentNumber").Eq(232L))
-				.SetProjection(Property.ForName("Name"));
+			if (TestDialect.SupportsOperatorAll)
```
hazzik: I don't get it. Why did it work for SQLite in non-generic version?

bahusoid: Because test was wrong before... Frederic've noticed it and fixed.
It was something like
```
var dc = query;
var dcs = subquery;
SerializeAndList(dc); <- subquery is not used.. so non-generic version works
```

hazzik: Ok

hazzik:
I don't really understand why we want to obsolete non generic version of `List` for criteria. I think this is perfectly valid usage.

bahusoid:
> why we want to obsolete non generic version of List for criteria

Actually that was very long time but it wasn't originally in PR and was suggested as additional refactoring. So should I restore it?

hazzik:
> should I restore it?

I don't know. I'm just trying to understand the motivation of this change.

bahusoid:
Original intention to make generic implementation do not depend on non-generic version (PR title is still about original change). To minimize and avoid possible `ToList<T>` conversions inside and to allow usage of specific `List<T>` methods.

fredericDelaporte:
I have suggested this change because I consider non generic `IList` as being a legacy, which support require additional code. So if the generic implementation does not require it anymore, I thought it would be better to obsolete that code to eventually get rid of it later.

Now of course if there is some disagreement with this, better drop it from this PR.

PR #1630, opened 2018-03-29
Remove class baseType interfaces from proxified methods

### LINKED ISSUE #1628: StackOverflowException for lazy proxied entities with explicit interface properties
In a domain model where PKs are not always of the same type, a fairly complex architecture of interfaces and classes is defined.
Said architecture provide for:

- a `IEntity` defining a `object Id { get; }` property

- a `IEntity<T> : IEntity` defining a `new T Id { get; }` property

Hence, every domain object has to implement both:

- a `object IEntity.Id { get { return Id; } }`

- and a `public virtual T Id { get; }`

To make things even more complicated, some entities has lazy fields on them and a `IPostLoadEventListener` is used.

Problem is: all this machinery functioned pretty well... up until the latest release of NHibernate.
In 5.1, for all entities with lazies, when trying to access the aforementioned `Id` property inside of the `IPostLoadEventListener`, a `StackOverflowException` is thrown.

Please note that all three elements above are needed for the bug to trigger.

fredericDelaporte:
Revert the lazy property proxy to its 5.0.x behavior.
This allows to avoid proxifying explicitly implemented interface methods, which are currently not supported by the dynamix proxy (lazy properties proxy).

Add test case adapted from gcesena test case.

Fixes #1628


hazzik:
I disagree with this change. Please make the DefaultProxyFactory the default proxy factory and run tests. Some will fail. There are specific cases when explicit interfaces are required.

hazzik:
Ok, please ignore previous comment. It's OK actually.

hazzik:
Still. Can you run whole test suite with old proxy factory for proxies?

fredericDelaporte:
> Still. Can you run whole test suite with old proxy factory for proxies?

I have done it locally with SQL-Server. Only the test checking the default factory was still the static one has failed, since I had changed the default rather than configuring the test.

PR #1633, opened 2018-04-01
Added support for batching 2nd level cache operations when loading entities and collections

maca88:
The aim of this PR is to reduce the performance drop when switching from an in-memory 2nd level cache to an external one. With the current status of this PR, only batching a read operation is implemented, the write operation is not yet implemented, which is why I put as WIP. In order to easily understand the new algorithm added in [BatchFetchQueue](https://github.com/maca88/nhibernate-core/commit/542245c7858cc9b89308d7681a7f6cd3d4ea7f46#diff-e743c8c25b00a55e7b7d6e31bd4e02f3R401) I've [made a fiddle](https://dotnetfiddle.net/vd2i0w) that shows how it works in comparison with the old one. As it can be seen from the fiddle, the old algorithm isn't always consistent when ordering the returned keys and also in certain cases it makes more cache calls than it would be needed. The new algorithm has always a consistent ordering when returning the keys and also the number of cache calls is reduced even if the cache doesn't support batching read operations. In the example you can see that the number of cache calls is reduced by half in certain cases, when batching is disabled.

Having a batchable write operation is a little more trickier when a `read-write` strategy is used. Having just a batchable put operation is not enough when the `read-write` strategy is used as we need to lock the keys prior writing in order to have consistent data in the cache. In order to have a fully batchable write when `read-write` strategy is used we need a batchable lock and unlock method. I am planning to add an interface for a batchable write cache as follows:
```
public interface IBatchableWriteCache
{
    void PutMultiple(object[] keys, object[] values);
    void LockMultiple(object[] keys);
    void UnlockMultiple(object[] keys);
}
```

I've spent some time analyzing `Redis` and `Memcached` if they would be able to implement the above methods correctly and my conclusion was that they could.

Redis:
- `PutMultiple`: can be implemented by using [MSET](https://redis.io/commands/mset) command or by creating an [atomic lua script](https://redis.io/commands/eval#atomicity-of-scripts) that would loop trough the keys and store them by using the [SET EX command](https://redis.io/commands/set) when the key should expire after a certain amount of time.
- `LockMultiple`: can be implemented by creating a lua script that would check for each key if a lock was already been made or not prior setting the lock. The lock would be successful only if the lock would be aquired for all keys. After that, the implementor would need to save the locked keys in order to release them when `UnlockMultiple` is called, to prevent releasing sombody else's locks.
- `UnlockMultiple`: can be implemented as an atomic lua script that would delete the locks that were made by the implementor by using the [DEL command](https://redis.io/commands/del).

Memcached:
- `PutMultiple`: [can be implemented](https://github.com/memcached/memcached/wiki/CommonFeatures#multiset) if the used memcached client supports [binary protocol](https://github.com/memcached/memcached/wiki/BinaryProtocolRevamped)
- `LockMultiple`: can be implemented by pipelining multiple [add commands](https://github.com/memcached/memcached/wiki/Commands#add) via the binary protocol and the lock would be obtained only if all add commands were successfuly completed. The implementor would need to save the locked keys in order to release them when `UnlockMultiple` is called, to prevent releasing sombody else's locks.
- `UnlockMultiple`: can be implemented by pipelining multiple [delete commands](https://github.com/memcached/memcached/wiki/Commands#delete) but only for keys that where locked by the implementor.

After this PR, as I am at the moment interested in `Redis` and the currently [Redis provider for NHiberante](https://github.com/TheCloudlessSky/NHibernate.Caches.Redis/blob/master/src/NHibernate.Caches.Redis/RedisCache.cs) isn't been ported to NH 5, I am willing to make one for [NHibernate Caches](https://github.com/nhibernate/NHibernate-Caches) that would support the batching operations from this PR. I've seen that @fredericDelaporte added a `Redis` implementation in [his PR](https://github.com/nhibernate/NHibernate-Caches/pull/29) by using the new AspNet Caching but unfortunately after a quick look at their [RedisCache](https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Redis/RedisCache.cs) class, it doesn't seem to be extendable.

Also, the `AsyncGenerator` has to be updated to version `0.10.0` as [local functions](https://github.com/maca88/AsyncGenerator/issues/85) are used in this PR.


fredericDelaporte:
nhibernate/NHibernate-Caches#29 integrates caches implementing the [`IDistributedCache`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.distributed.idistributedcache) interface, which is quite minimal. So yes, for using more advanced features of specific caches implementations, dedicated caches implementations will be required.

maca88:
With the current state, `ICacheConcurrencyStrategy` [Get](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Cache/ICacheConcurrencyStrategy.cs#L30) and [Put](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Cache/ICacheConcurrencyStrategy.cs#L43) methods are now batchable. Which means that batching can occur when loading an entity or a collection by its id (`Get` operation) or after loading an entity or collection from the database (`Put` operation).

In order to support batching other `ICacheConcurrencyStrategy` methods we would need two things:
- A cache batcher on the session like the [IBatcher](https://github.com/nhibernate/nhibernate-core/blob/dce4d0204fc9be446ef7d2d76f478182f18a5367/src/NHibernate/Engine/ISessionImplementor.cs#L98) in order to [flush the operations](https://github.com/nhibernate/nhibernate-core/blob/73797df2ef79ecbd018dd2f416ea68849f8655c9/src/NHibernate/Engine/ActionQueue.cs#L157) after the execution of actions in the queue
- Enable [BeforeExecutions](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Action/EntityAction.cs#L100) for `EntityAction` in order to batch the lock operation for  [EntityUpdateAction](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Action/EntityUpdateAction.cs#L74) and [EntityDeleteAction](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Action/EntityDeleteAction.cs#L61)

I am not planning to implement the other methods from `ICacheConcurrencyStrategy` in this PR in order to avoid too much changes, so I will update the title to reflect that.

fredericDelaporte:
@gliljas, I believe you had written about adding support for batching second level cache operations, although I do not find where anymore. Do you have some thoughts to share here?

src/NHibernate/Cache/IBatchableWriteCache.cs
```diff
@@ -0,0 +1,39 @@
...
+	/// The implementor should use this interface along with <see cref="ICache"/> when the
+	/// cache supports a multiple get and put operation.
+	/// </summary>
+	/// <remarks>
+	/// <threadsafety instance="true" />
+	/// <para>
+	/// All implementations <em>must</em> be threadsafe.
+	/// </para>
+	/// </remarks>
+	public partial interface IBatchableReadWriteCache : IBatchableReadCache
```
hazzik: Are there any reasons to split these interfaces?

src/NHibernate/Cache/ICacheConcurrencyStrategy.cs
```diff
@@ -139,4 +140,84 @@ public partial interface ICacheConcurrencyStrategy
...
+		/// <summary>
+		/// Attempt to retrieve multiple objects from the Cache
+		/// </summary>
+		/// <param name="cache">The cache concurrency strategy.</param>
+		/// <param name="keys">The keys (id) of the objects to get out of the Cache.</param>
+		/// <param name="txTimestamp">A timestamp prior to the transaction start time</param>
+		/// <returns>An array of cached objects or <see langword="null" /></returns>
+		/// <exception cref="CacheException"></exception>
+		//6.0 TODO: Merge into ICacheConcurrencyStrategy.
+		public static object[] GetMultiple(this ICacheConcurrencyStrategy cache, CacheKey[] keys, long txTimestamp)
```
hazzik: I would like to have `interface IMultiCacheConcurrencyStrategy : ICacheConcurrencyStrategy` interface instead.

src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -136,6 +146,45 @@ public object Get(CacheKey key, long txTimestamp)
 			}
 		}
 
+		public object[] GetMultiple(CacheKey[] keys, long txTimestamp)
```
hazzik: Maybe rename to just `Get` or to `GetMany`?

src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -171,6 +220,86 @@ public ISoftLock Lock(CacheKey key, object version)
...
 		}
 
+		/// <summary>
+		/// Do not add an item to the cache unless the current transaction
+		/// timestamp is later than the timestamp at which the item was
+		/// invalidated. (Otherwise, a stale item might be re-added if the
+		/// database is operating in repeatable read isolation mode.)
+		/// </summary>
+		/// <returns>Whether the items were actually put into the cache</returns>
+		public bool[] PutMultiple(CacheKey[] keys, object[] values, long timestamp, object[] versions, IComparer[] versionComparers,
```
hazzik: Maybe rename to just `Put` or `PutMany`

hazzik:
I would like to understand how does it marry with #1392?

src/NHibernate/Cache/IBatchableWriteCache.cs
maca88: The only reason is to support caches that have only a batchable read operation like [EnyimMemcached](https://github.com/enyim/EnyimMemcached/blob/develop/Enyim.Caching/IMemcachedClient.cs).

src/NHibernate/Cache/ReadWriteCache.cs
maca88: Why not, both are fine for me. Which would you prefer?

fredericDelaporte: I would vote for `GetMany`. It is more immediate to understand its purpose with this nam, compared to `Get` alone which would require interpreting its arguments/return value.

maca88:
If I understood correctly, the mentioned PR is about minimizing `ICache.Clear` calls when `ExecuteUpdate` is called, where this PR is about batching `ICache.Get` and `ICache.Put` operations. From the code changes I don't see any complication merging both PRs.

src/NHibernate/Cache/ReadWriteCache.cs
hazzik: `GetMany` it is.

src/NHibernate/Cache/ICacheConcurrencyStrategy.cs
maca88: I've added a `IBatchableCacheConcurrencyStrategy` interface.

src/NHibernate/Cache/ReadWriteCache.cs
maca88: Done.

src/NHibernate/Cache/ReadWriteCache.cs
maca88: Renamed to `PutMany`.

maca88:
Great! I've renamed the interfaces and added a small optimization for checking if batching is supported.

maca88:
I am currently working on the Redis cache for NHibernate that will support batching from this PR and I found out that the current locking mechanism defined by `ICache.Lock` and `ICache.Unlock` methods is not adequate for distributed locks. Locking in a distributed cache like Redis is done by adding an additional key (lock key) for each key that we want to lock where the lock key value has to be a unique value in order to know who locked the key. Currently, `ICache.Lock` does not have a return value that would represent the lock value that was used to lock the key, which would then be passed to the `ICache.Unlock` method to unlock only the keys that were locked with the given lock value. This forced me to store the locked keys with their lock values locally when the `ICache.Lock` is called and use them when `ICache.Unlock` is called, producing a non thread safe implementation of `ICache.Lock` and `ICache.Unlock`:
```
/// <summary>
/// With the current NHibernate version (5.1) the <see cref="ICache.Lock"/> method does not have a
/// return value that would indicate which lock value was used to lock the key, that would then be
/// passed to the <see cref="ICache.Unlock"/> method in order to prevent unlocking keys that were locked
/// by others. Luckily, the <see cref="ICache.Lock"/> and <see cref="ICache.Unlock"/> methods are currently
/// always called inside a lock statement which we abuse by storing the locked key when the
/// <see cref="ICache.Lock"/> is called and use it when <see cref="ICache.Unlock"/> is called. This
/// means that our <see cref="Lock"/> and <see cref="Unlock"/> are not thread safe and have to be called
/// sequentially in order prevent overriding the lock key for a given key.
/// Currently, the <see cref="ICache.Unlock"/> is always called after the <see cref="ICache.Lock"/> method
/// even if an exception occurs which we are also abusing in order to avoid having a special mechanism to 
/// clear the dictionary in order to prevent memory leaks.
/// </summary>
private readonly ConcurrentDictionary<string, string> _aquiredKeyLocks = new ConcurrentDictionary<string, string>();
```

I am planning to correct that for the `IBatchableCache` by changing the `LockMany` and `UnlockMany` methods to:

```
/// <summary>
/// Lock the objects from being changed by another thread.
/// </summary>
/// <param name="keys">The keys to lock.</param>
/// <returns>The value that was used to lock the keys.</returns>
object LockMany(object[] keys);

/// <summary>
/// Unlock the objects that were previously locked.
/// </summary>
/// <param name="keys">The keys to unlock.</param>
/// <param name="lockValue">The value that was used to lock the keys.</param>
void UnlockMany(object[] keys, object lockValue);
```

Are you ok with this?

fredericDelaporte:
Yes, better have the right signature in this interface from the start.
It would then be better to add another PR for supporting this pattern on `Lock` and `Unlock` too. (With the constraint of avoiding breaking change...)

maca88:
Supporting the same pattern for `Lock` and `Unlock` methods without breaking changes, would require to add an additional interface like `IDistributedLock`. If we add it in `5.2` what would happen in `6.0`, would the `IDistributedLock` be removed or would we remove the `ICache.Lock` and `ICache.Unlock` methods instead? If the latter is true, then probably would be better to introduce also an `IBatchableDistributedLock` in this PR.

fredericDelaporte:
The way we should deal with all the "transitional" interfaces we add for avoiding breaking changes, once we do the next major version, is not yet defined.
My view is, we need to have it as smooth as possible for upgraders which would have implemented them. Removing completely the "transitional" interfaces would leave little clues as to how to migrate previous code using them. I think we will have to obsolete them instead, for providing guidance through the obsolete message.
Then, about, the methods they define, the smoother would be to keep them and keep them functional altogether. But I think that would be too much hassle. I am for emptying these "transitional" interfaces, in order to enforce moving to the updated interfaces.

> If the latter is true, then probably would be better to introduce also an IBatchableDistributedLock in this PR.

For consistency with what we will do for `ICache.Lock` and `ICache.Unlock`? I would rather avoid adding one more "transitional" interface for a small consistency concern.

maca88:
I agree that we should avoid adding a "transitional" interface for a small change like this, but I don't see how we could correct that for `5.2` otherwise. Maybe would be better to not make any "transitional" interface and leave the `ICache.Lock` and `ICache.Unlock` as they are today and correct them in `6.0`. The implementor by taking into the consideration how the two methods are called in `ReadWriteCache` today, can still provide a thread safe implementation by storing the `lockValue` locally when the `ICache.Lock` is called and reuse it in `ICache.Unlock` method. Note that the current implementation of the two methods in the [new Redis provider](https://github.com/nhibernate/NHibernate-Caches/pull/45) are not thread safe, but I will fix that.

maca88:
If that is the case then wouldn't it be better to replace the interfaces `IBatchableCache` and `IBatchableReadOnlyCache` in this PR with abstract classes `BatchableCacheBase` and `BatchableReadOnlyCacheBase` for future consistency?

fredericDelaporte:
No, that will be the job of the PR doing the switch to `CacheBase` to remove these no more needed transitional interfaces (provided that other PR is done before these transitional interfaces get released).

src/NHibernate/Cache/CacheBatcher.cs
```diff
@@ -0,0 +1,125 @@
...
+using NHibernate.Persister.Entity;
+
+namespace NHibernate.Cache
+{
+	/// <summary>
+	/// A batcher for batching operations of <see cref="ICacheConcurrencyStrategy"/>, where the batch size is retrived
+	/// from an <see cref="IEntityPersister"/> or <see cref="ICollectionPersister"/>.
+	/// When a different persister or a different operation is added to the batch, the current batch will be executed.
+	/// </summary>
+	internal partial class CacheBatcher
```
fredericDelaporte: I am realizing a bit late this class is maybe mostly not necessary.

- It contains logic as if it could handle different kind of batches, but only put batches currently exist. This logic is likely a remnant from conception where a get batch was planned (or even done then removed).
- Why should put batches be segregated by persisters and restricted by `batch-size`?

`batch-size` is meant for get in my point of view. Limiting the batch size reduces the likeliness of loading unneeded data, since the batched gets are done on the assumption that when accessing an entity or collection, the user will later use the other pending entities or collections. This is of course not guaranteed.

But for put, instead, NHibernate will not put data that do not need to be put. So for me, limiting by batch-size here does not really make sense.

And I guess the segregation by persisters was done for handling batch-size, so if limiting by batch-size should be removed, it should be removed too.

maca88: > It contains logic as if it could handle different kind of batches, but only put batches currently exist.

This class was mentioned to batch also other operations of `ICacheConcurrencyStrategy` like `Lock` in order to support batching the `EntityUpdateAction` and `EntityDeleteAction` in the future ([related comment](https://github.com/nhibernate/nhibernate-core/pull/1633#discussion_r200837292)). I've implemented the lock batcher which would be used prior updating or deleting entities, but was not included as the scope of this PR was reduced.

> Why should put batches be segregated by persisters and restricted by batch-size

I took a similar approach as `IBatcher` has, where the "batch_size" is used also for inserts, but I agree that we could ignore the batch size for the put operation and rather having it configured by the cache provider if needed. I didn't yet perfom a test that would test if splitting in smaller batches would increase or reduce the overall performance, I will test it and let you know. 
About segregation by persister, it was done to make the batcher simpler so that we can have only one batch active at the given moment, similar to `AbstractBatcher` with the `CurrentCommand` property.

maca88: Here are the results with different batch-sizes for `PutMany` operation:

Binary serializer:
```
PutMany operation for 1000 keys with region strategy FastRegionStrategy:
Total iterations: 5
Batch size: 20
Times per iteration 75ms,66ms,74ms,69ms,71ms
Average 71ms

PutMany operation for 1000 keys with region strategy FastRegionStrategy:
Total iterations: 5
Batch size: 100
Times per iteration 42ms,39ms,44ms,42ms,46ms
Average 42,6ms

PutMany operation for 1000 keys with region strategy FastRegionStrategy:
Total iterations: 5
Batch size: 500
Times per iteration 40ms,40ms,40ms,38ms,40ms
Average 39,6ms

PutMany operation for 1000 keys with region strategy FastRegionStrategy:
Total iterations: 5
Batch size: 1000
Times per iteration 39ms,41ms,37ms,38ms,38ms
Average 38,6ms
```

Json serializer:
```
PutMany operation for 1000 keys with region strategy FastRegionStrategy:
Total iterations: 5
Batch size: 20
Times per iteration 55ms,54ms,58ms,59ms,58ms
Average 56,8ms

PutMany operation for 1000 keys with region strategy FastRegionStrategy:
Total iterations: 5
Batch size: 100
Times per iteration 33ms,29ms,34ms,32ms,31ms
Average 31,8ms

PutMany operation for 1000 keys with region strategy FastRegionStrategy:
Total iterations: 5
Batch size: 500
Times per iteration 29ms,26ms,27ms,29ms,26ms
Average 27,4ms

PutMany operation for 1000 keys with region strategy FastRegionStrategy:
Total iterations: 5
Batch size: 1000
Times per iteration 27ms,25ms,24ms,28ms,27ms
Average 26,2ms
```

The tests were run the same way as explained [here](https://github.com/nhibernate/NHibernate-Caches/pull/45#issuecomment-401609796).
Based on the result, we should remove the batch-size for the `Put` operation as you said.

fredericDelaporte: Thanks for having benchmarked this!

gokhanabatay: @fredericDelaporte I tried this with below configuration.

I have been download NHibernate master repository, then build.

EntApiDefMap.cs Fluent mapping
``Cache.ReadWrite();
``
For 233 records NHibernate Profiler shows 234 Redis Cache Access
1 for query cache the others are entity cache retrieve.
Why is not returned batched, I do not understand?
``
public EntApiDef Get(string apiCode)
        {
            return ISession.Query<EntApiDef>()
                           .Where(x => x.Code == apiCode)
                           .WithOptions(options =>
                           {
                               options.SetReadOnly(true);
                               options.SetCacheable(true);
                           }).FirstOrDefault();
        }
``

fredericDelaporte: This stuff is not released, and no second level cache supporting it either.

maca88: @gokhanabatay NHibernate Profiler will currently not tell you the actual cache calls when batching is enabled. Instead, use [Redis Monitor](https://redis.io/commands/monitor) to verify how many calls actually happened.

gokhanabatay: @maca88 
I test it with profiler too and debug the code; bathSize equals  1 so it does not fetch many just fetchs single record.
What am i doing wrong.
@fredericDelaporte  I know as you mentioned its not released, I just download Nhibernate master code and debug it.

``
var batchSize = persister.GetBatchSize();
			if (batchSize > 1 && persister.Cache.PreferMultipleGet())
			{
				// The first item in the array is the item that we want to load
				var entityBatch =
					source.PersistenceContext.BatchFetchQueue.GetEntityBatch(persister, @event.EntityId, batchSize, false);
				// Ignore null values as the retrieved batch may contains them when there are not enough
				// uninitialized entities in the queue
				var keys = new List<CacheKey>(batchSize);
``

fredericDelaporte: You have to set either a `default_batch_fetch_size` in [configuration](http://nhibernate.info/doc/nhibernate-reference/session-configuration.html#configuration-optional), or set the `batch-size` in the [mappings](http://nhibernate.info/doc/nhibernate-reference/performance.html#performance-fetching-batch).

gokhanabatay: @fredericDelaporte  I get your point now, NHibernate.Cache providers must implement CacheBase abstract class in order to support get many and put many. 
At this point it seems to be, i will use corememorycache as second level cache provider.
for data consistency i will use NH Post Events in order to sync other instances of my application.
So i will ignore sync time as a result it could have delays.

src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -181,64 +213,176 @@ public void RemoveBatchLoadableCollection(CollectionEntry ce)
...
+				else
+				{
+					collectionKeys.Add(new KeyValuePair<KeyValuePair<CollectionEntry, IPersistentCollection>, int>(me, index));
+					// Check the cache only when we have collected as many keys as are needed to fill the batch,
+					// that are after the demanded key.
+					if (!keyIndex.HasValue || index < keyIndex.Value + batchSize)
+					{
+						return false;
+					}
+					return CheckCacheAndProcessResult();
```
bahusoid: I noticed that `ProcessKey` and `CheckCacheAndProcessResult` can call each other. Maybe recursive calls are not possible due to some parameters check but it's quite unclear. I really can't understand what's going on here. So I have a question: is it really possible that `CheckCacheAndProcessResult`  call can lead to `ProcessKey` that leads again to  `CheckCacheAndProcessResult`?

If not I think we should introduce separate `ProcessKey` method without `CheckCacheAndProcessResult` (maybe return null here). It will simplify code reading - as I've got brain stack overflow while trying to understand it :)

maca88: > is it really possible that `CheckCacheAndProcessResult` call can lead to `ProcessKey` that leads again to `CheckCacheAndProcessResult`?

It is not possbile. The first `ProcessKey` call inside `CheckCacheAndProcessResult` method cannot call it because of `batchableCache` variable being `null` and the second `ProcessKey` call cannot call it due to `ignoreCache` which is set to `true`. So the PR that you've made is correct and indeed simplifies the code reading :)

PR #1637, opened 2018-04-03
Supply expected values to GetIdentitySelectString

### LINKED ISSUE #1635: IdentitySelectString implementation is inconsistent
From `NHibernate.Dialect.Dialect`:

```
/// <summary> 
/// Get the select command to use to retrieve the last generated IDENTITY
/// value for a particular table 
/// </summary>
/// <param name="tableName">The table into which the insert was done </param>
/// <param name="identityColumn">The PK column. </param>
/// <param name="type">The <see cref="DbType"/> type code. </param>
/// <returns> The appropriate select command </returns>
public virtual string GetIdentitySelectString(string identityColumn, string tableName, DbType type)
{
	return IdentitySelectString;
}
```

From the comments I would expect the first parameter to be the table name. However, the name of the first parameter suggests it should be the column name.

From `NHibernate.Persister.Entity.AbstractEntityPersister`:

```
public string IdentitySelectString
{
	get
	{
		if (identitySelectString == null)
			identitySelectString =
				Factory.Dialect.GetIdentitySelectString(GetTableName(0), GetKeyColumns(0)[0], IdentifierType.SqlTypes(Factory)[0].DbType);
		return identitySelectString;
	}
}
```
Here the table name is passed as the first parameter matching the comment on `NHibernate.Dialect.Dialect.GetIdentitySelectString`.

From `NHibernate.Persister.Collection.AbstractCollectionPersister`:

```
public string IdentitySelectString
{
	get
	{
		if (identitySelectString == null)
		{
			identitySelectString =
				Factory.Dialect.GetIdentitySelectString(IdentifierColumnName, qualifiedTableName,	IdentifierType.SqlTypes(Factory)[0].DbType);
		}
		return identitySelectString;
	}
}
```

Here the column name is passed as the first parameter matching the parameter naming on `NHibernate.Dialect.Dialect.GetIdentitySelectString`.

fredericDelaporte:
The entity persister was supplying the table name as the column
parameter and the column name as the table parameter. This has no effect
on NHibernate tests because GetIdentitySelectString does not use its
parameters by default and no built-in dialects overrides it for using
them.

Fix #1635

### DIFF
--- a/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
+++ b/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
@@ -755,8 +755,8 @@ namespace NHibernate.Persister.Entity
 			{
 				if (identitySelectString == null)
 					identitySelectString =
-						Factory.Dialect.GetIdentitySelectString(GetTableName(0), GetKeyColumns(0)[0],
-																										IdentifierType.SqlTypes(Factory)[0].DbType);
+						Factory.Dialect.GetIdentitySelectString(GetKeyColumns(0)[0], GetTableName(0),
+						                                        IdentifierType.SqlTypes(Factory)[0].DbType);
 				return identitySelectString;
 			}
 		}
--- a/src/NHibernate/Dialect/Dialect.cs
+++ b/src/NHibernate/Dialect/Dialect.cs
@@ -990,14 +990,14 @@ namespace NHibernate.Dialect
 			return insertString;
 		}
 
-		/// <summary> 
+		/// <summary>
 		/// Get the select command to use to retrieve the last generated IDENTITY
-		/// value for a particular table 
+		/// value for a particular table.
 		/// </summary>
-		/// <param name="tableName">The table into which the insert was done </param>
-		/// <param name="identityColumn">The PK column. </param>
-		/// <param name="type">The <see cref="DbType"/> type code. </param>
-		/// <returns> The appropriate select command </returns>
+		/// <param name="identityColumn">The PK column.</param>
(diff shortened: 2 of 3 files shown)


src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -755,8 +755,8 @@ public string IdentitySelectString
 			{
 				if (identitySelectString == null)
 					identitySelectString =
-						Factory.Dialect.GetIdentitySelectString(GetTableName(0), GetKeyColumns(0)[0],
-																										IdentifierType.SqlTypes(Factory)[0].DbType);
+						Factory.Dialect.GetIdentitySelectString(GetKeyColumns(0)[0], GetTableName(0),
+						                                        IdentifierType.SqlTypes(Factory)[0].DbType);
```
fredericDelaporte: Maybe some external dialects requiring the use of these parameters may be broken by this fix, if they were "compensating" for the inversion while being used for mappings where only entities are using identity columns. That is why this fix should not be put in 5.1.1 in my opinion.

hazzik:
I think switching parameter names according to how they are called could provide less breaking changes.

fredericDelaporte:
> I think switching parameter names according to how they are called could provide less breaking changes.

This will imply switching parameters in collection persister instead. But since collections saving identities are likely quite infrequent (`idbag` only I guess), it may be less breaking.

hazzik:
> This will imply switching parameters in collection persister instead. 

Didn't know about that one. So, then this change is correct.

PR #1646, opened 2018-04-09
Add a link to release notes in NuGet package

fredericDelaporte:
As seen [here](https://www.nuget.org/packages/NHibernate.AspNet.Identity/), a raw url in package release notes tag yield a link on NuGet site.

Adding a link to NHibernate release notes there would boost its discoverability.


hazzik:
Shouldn’t it be a raw version?

fredericDelaporte:
What do you mean?

hazzik:
I mean the version of the url when you click on “raw” button.

Eg.
https://raw.githubusercontent.com/nhibernate/nhibernate-core/5.1.0/releasenotes.txt

fredericDelaporte:
Well, that is maybe a matter of taste. I tend to prefer having access to more context. The non-raw version lands in GitHub interface, allowing to easily browse further the nhibernate repository, while the raw version is a dead end.

hazzik:
Fine

PR #1647, opened 2018-04-09
Fix key type mismatch in EntityUniqueKey

### LINKED ISSUE #1645: One-to-one with property-ref triggers StackOverflow Exception
When we updated our project to the latest version of NHibernate (5.1), we run into a Stackoverflow exception. After some investigation, we were able to reproduce the behavior in a small project.

The structure of the database looks as following:
`[SuperParent] <-- 1:1 --> [Parent] <-- 1:n --> [Child]`

To end up in the exception, two requirements must be given:
- Parent must have a back reference to its "super parent"
- Parent must override equals

The change which introduced this misbehavior is in the Loader class (Line 953). When getting a row, it caches the the loaded entity now. The caching uses the GetHashCode method of the entity to create a unique key. If the entity is not initialized yet, it starts to load the entity again. When loading the same entity again, the entity is still not initialized (AbstractLazyInitializer.Initialize() on line 107) and it ends up again in caching the same entity.

The repro can be found here: [https://github.com/michaelestermann/NHibernateRepro](https://github.com/michaelestermann/NHibernateRepro)

fredericDelaporte:
Fixes #1645.


src/NHibernate/Engine/EntityUniqueKey.cs
```diff
@@ -34,7 +34,7 @@ public EntityUniqueKey(string entityName, string uniqueKeyName, object semiResol
 			this.entityName = entityName;
 			this.uniqueKeyName = uniqueKeyName;
 			key = semiResolvedKey;
-			this.keyType = keyType.GetSemiResolvedType(factory);
+			this.keyType = keyType;
```
fredericDelaporte: `EntityUniqueKey` was assuming the value to be always semi-resolved (to be the identifier instead of the entity for an entity type). It was always computing the value hashcode with the semi-resolved type, which is the identifier type for an entity.
But this has changed with #1492, which caches unique keys also for already fully loaded entities. The value hashcode must then be computed with the entity type, which accounts for proxies.
Otherwise the proxy loading may get triggered while the code is already loading the proxy, which results in a stack overflow.

The type to use is now fully computed before constructing `EntityUniqueKey`.
It is constructed in only two places: in the loader, see its changes; in the `EntityType.LoadByUniqueKey`, which uses `GetIdentifierOrUniqueKeyType` which yield types having themselves as semi-resolved types. (For being extra-safe, a call to `GetSemiResolvedType` could still be added in `LoadByUniqueKey` just in case.)

Ideally `semiResolvedKey` parameter should be renamed simply `key` (no more assumptions). But renaming a public method parameter is a gray area possible breaking change (if anyone has written a call explicitly naming parameters), and I avoid doing this one in patch releases.

hazzik: >For being extra-safe, a call to GetSemiResolvedType could still be added in LoadByUniqueKey just in case.

Let's do this.

>Ideally semiResolvedKey parameter should be renamed simply key (no more assumptions).

Please add a TODO for 6.0

src/NHibernate/Async/Type/EntityType.cs
```diff
@@ -218,12 +218,19 @@ public override Task<object> ResolveIdentifierAsync(object value, ISessionImplem
 
 			//TODO: implement caching?! proxies?!
 
+			var keyType = GetIdentifierOrUniqueKeyType(factory)
+				// EntityUniqueKey was doing this on the type. I suspect this was needed only for its usage in Loader,
+				// which can work with entities as keys not yet instanciated and just represented by their identifiers.
+				// But since removing this call from EntityUniqueKey is done for a patch and that the code path here has
+				// no known bugs with this GetSemiResolvedType, moving its call here for avoiding altering this code
+				// path. See GH1645.
+				.GetSemiResolvedType(factory);
```
hazzik: Based on the implementation, this actually could happen only if the primary or unique key is entity itself. Is it event possible?

hazzik: Yes, it's possible.

Issue #1655, opened 2018-04-11
Convert documentation to Markdown


weelink:
I was trying to find some documentation on Mapping by Code, but could not find anything useful except issue #1571. I'm thinking about contributing to this part of the documentation. However the current docs are in DocBook and need to be built using NAnt (that has not been updated in years).

How about converting the documentation to a more contributor-friendly format like Markdown? It will also  make it easier to put the docs up on [readthedocs.org](https://readthedocs.org/).

weelink:
Markdown is off the table if the documentation should be published on readthedocs.org. Their format of choice is reStructuredText and convert it using [Spinx](http://www.sphinx-doc.org/en/master/).

If that is going to be the direction of the docs, then the docbook should be converted to reStructuredText (pretty easy using [Pandoc](https://pandoc.org/) and the build script for the documentation should be modified to use Sphinx. Haven't looked at that yet.

weelink:
The reason to convert the documentation to Markdown was the ease of maintenance (tool support), ease of generation (no more xml/xsl/nant/java etc), and to allow the docs to be found through the newer channels (notably readthedocs.io and/or github). Since Markdown has its limits, I have converted it to reStructuredText.

I have managed to let TeamCity build the documentation using reStructuredText.
In order to make this happen, I had to include a zipped version Python as part of the Tools. This is 30MB.
There are still a few things to do, namely fixing some formatting issues with the PDF and HTML, and I accidentally broke the generation of the API documentation (so the build fails at the moment).
It will take some time before this is fixed.

When I look at the download statistics in [SourceForge](https://sourceforge.net/projects/nhibernate/files/NHibernate/), it seems that practically no one is actually downloading the releases and even less people download the documentation. NuGet and http://nhibernate.info/ seem to be the most used option.
So I was wondering if I should continue my attempt at generating the documentation in TeamCity.

There are some options to consider:

1. Close this issue and keep the documentation in DocBook
2. Keep the documentation in reStructuredText, accept the Python bloat, fix the build and let it be published to SourceForge
3. Stop generating the documentation as part of the release, and only publish it on readthedocs.io/github/whatever

Every option has its pros and cons. It is beyond my power as a contributor to make a decision.
Can some of the owners (@fredericDelaporte, @hazzik, @someone_else) of NHibernate think about this?

hazzik:
I would personally prefer option #1.

fredericDelaporte:
[...]

PR #1657, opened 2018-04-11
WIP Convert docs to markdown

weelink:
Implements #1655

- [ ] Publish as multi-pages web documentation
- [ ] Publish as single page html
- [ ] Generate html for the zip
- [ ] Generate pdf
- [x] Update filenames and anchors to be backward compatible with current docs


weelink:
This is the first conversion from DocBook to Markdown. I have added the syntax highlighting and a TOC. I want to go through this anyway to see if there are some links missing.

There are some decisions to be made. I have completely replaced the current DocBook files. I think we should only have a single set of documentation. If these docs are good enough, the link on the nhibernate.info website should be updated.
And where are these going to be published? readthedocs.org is an obvious option, as well as here on github.

doc/reference/query_criteria.md
```diff
@@ -0,0 +1,345 @@
...
+You may specify association fetching semantics at runtime using
+`SetFetchMode()`.
+
+```csharp
+    var cats = sess.CreateCriteria<Cat>()
+        .Add( Expression.Like("Name", "Fritz%") )
+        .SetFetchMode("Mate", FetchMode.Eager)
+        .SetFetchMode("Kittens", FetchMode.Eager)
+        .List<Cat>();
+```\
```
bahusoid: Broken code example

doc/reference/query_criteria.md
```diff
@@ -0,0 +1,345 @@
...
+        .List<object[]>();
+    
+    foreach (var objs in cats)
+    {
+        Cat cat = (Cat) objs[0];
+        Cat kitten = (Cat) objs[1];
+    }
+```
+
+See [???](#queryqueryover-projectionentities) for more information.
```
bahusoid: Broken link

doc/reference/nhibernate_caches.md
```diff
@@ -0,0 +1,715 @@
+# NHibernate.Caches
+
+> **NHibernate.Caches namespace contains several second-level cache
+> providers for NHibernate..**
```
fredericDelaporte: Double dot. Should we check paragraph title? It seems like some logic has added a dot to them.

And the abstract title, `What is NHibernate.Caches?`, is missing.

doc/reference/nhibernate_caches.md
```diff
@@ -0,0 +1,715 @@
...
+> not a tree of objects but rather a map of arrays.
+> 
+> With the *session-per-request* model, a high number of sessions can
+> concurrently access the same entity without hitting the database each
+> time; hence the performance gain.
+> 
+> Depending on the chosen cache provider, the second level cache may be
+> actually shared between different session factories. If you need to
+> avoid this for some session factories, configure each of them with a
+> different `cache.region_prefix`. See [???](#configuration-optional).
```
fredericDelaporte: As reported by Roman, internal links are broken.

doc/reference/nhibernate_caches.md
```diff
@@ -0,0 +1,715 @@
...
+> 
+> Several cache providers have been contributed by NHibernate users:
+> 
+>   - `NHibernate.Caches.Prevalence`  
+>     Uses `Bamboo.Prevalence` as the cache provider. Open the file
+>     `Bamboo.Prevalence.license.txt` for more information about its
+>     license; you can also visit its
+>     [website](http://bbooprevalence.sourceforge.net/). This provider
+>     is available for the .Net Framework only. Also see
+>     [section\_title](#NHibernate.Caches.Prevalence).
```
fredericDelaporte: Other case of broken link. (Unless this is handled by markdown and will correctly output the section title.)

fredericDelaporte:
The Docbook doc is generating many format: a single page html, a multi-page html, a pdf and a chm. Is there ways to generates this different formats from the markdown?

All these formats are currently published on [nhibernate.info](http://nhibernate.info/doc/index.html) as you likely have seen. nhibernate.info is a GitHub site, that is https://github.com/nhibernate/nhibernate.github.io.

If the previous format can be generated from markdown, will the html format have the same page names and anchors for avoiding breaking existing links done from other sites like Stack Overflow? (In case we replace the current documentation on nhibernate.info.) Otherwise, we may consider publishing them elsewhere like readthedoc, add a canonical ref in the docs on nhibernate.info toward the migrated doc, and for current doc point nhibernate.info links toward readthedoc.

Otherwise, depending on which formats cannot be generated, this may be considered blocking.

Yes there are build scripts. The TeamCity release build is building the DocBook doc, through the nant task [`package`](https://github.com/nhibernate/nhibernate-core/blob/5.1.0/default.build#L285), sub task [`reference-zip`](https://github.com/nhibernate/nhibernate-core/blob/5.1.0/default.build#L196) calling [`manual-zip`](https://github.com/nhibernate/nhibernate-core/blob/5.1.0/doc/documentation.build#L37) task in `doc\documentation.build`.
This `documentation.build` does not only build the DocBook documentation, but can also build a second chm, from the xml comments. (This is not done by `manual-zip` but by some other tasks.)

doc/reference/query_criteria.md
weelink: Fixed

doc/reference/query_criteria.md
weelink: Fixed

doc/reference/nhibernate_caches.md
weelink: Fixed. I don't know where that double dot came from. I haven't encountered it elsewhere

doc/reference/nhibernate_mapping_attributes.md
```diff
@@ -0,0 +1,371 @@
+# NHibernate.Mapping.Attributes
+
+> **NHibernate.Mapping.Attributes is an add-in for
+> [NHibernate](http://nhibernate.info/) contributed by Pierre Henri
+> Kuaté (aka *KPixel*); the former implementation was made by John
+> Morris..**
```
fredericDelaporte: Another instance of double dot. I think any `<title>` inside a `<formalpara>` has been appended with a dot. But all other cases were not already having one, so they dot not end up with a double one.

The tool you have used has likely done this because DocBook do this too, but is smarter: it adds a dot only if the title was not already ending with a dot.

doc/reference/nhibernate_caches.md
```diff
@@ -0,0 +1,627 @@
...
+      - The `use_sliding_expiration` value is whether you wish to use a
+        sliding expiration or not. Defaults to `false`. Not all
+        providers support this setting, it may be ignored. Check their
+        respective documentation.
+
+  - Add `<cache usage="read-write|nonstrict-read-write|read-only"/>`
+    (just after `<class>`) in the mapping of the entities you want to
+    cache. It also works for collections (bag, list, map, set, ...).
+
+**Be careful.**
```
fredericDelaporte: It was originally without a dot (but that does not harm here).

weelink:
It is possible to use a tool like [pandoc](https://pandoc.org) for this. It can convert it to a single page html, multi page html and pdf. I haven't seen .chm yet.

According to [this StackOverflow answer](https://stackoverflow.com/questions/15560114/tool-to-generate-chm-and-pdf-documentation-from-markdown-on-windows), it should be possible to generate .chm files with [Sphinx](http://www.sphinx-doc.org/en/master/index.html).

The page names and anchors are different. Maybe there is a way to control these when github/pandoc/whatever is processing the Markdown. This needs to be looked into further.

weelink:
It seems it is possible to (somewhat) control the anchors. By adding `<a name="#whatever"></a>` in the titles, you can link to this using `[Whatever](#whatever)`. It will take some work, but I think we can make these anchors backwards compatible.

Generating html should be relatively easy. We need to change the build scripts somewhat, but I don't see problems there at the moment.

doc/reference/nhibernate_caches.md
weelink: I went through all of these and I think I have fixed all of them. I will still do a manual check.

doc/reference/nhibernate_caches.md
weelink: I went through all of these and I think I have fixed all of them. I will still do a manual check.

weelink:
I'm not an expert on readthedocs, but it seems it can generate PDF and single page html.
We could change the build process to download the latest PDF and single page html from readthedocs instead of trying to build them ourselves.

weelink:
I would like to see how the current docs look. It seems that readthedocs allows the admin to select which branch to use for generating. @fredericDelaporte : Can you create an account there and point it to this branch?

fredericDelaporte:
I had a look at readthedocs documentation. It appears markdown is not their documentation format of choice, but just a possibility against which they advice, preferring Sphinx. They point [there](http://ericholscher.com/blog/2016/mar/15/dont-use-markdown-for-technical-docs/) for motivations.

Furthermore, from [here](https://docs.readthedocs.io/en/latest/builds.html#mkdocs):

> As MkDocs doesn’t support automatic PDF generation, Read the Docs cannot create a PDF version of your documentation with the Mkdocs option.

MkDocs is their builder for markdown documentations.

So generating all our formats with readthedocs will not happen with a markdown documentation. Anyway this option has an additional drawback: it means also adding a new external dependency in NHibernate build process. The more external dependencies, the more likely the build process will experience failures due to external dependencies own failures.

I conclude from this that the most straightforward path to get this PR done is likely to target the documentation for still being hosted on GitHub, in https://github.com/nhibernate/nhibernate.github.io.

The current documentation build task needs to be adapted for generating the documentation formats from the new md files in any cases, since delivering a zip of documentation is part of the NHibernate release process anyway (thus the current TeamCity failure for the "NHibernate (Release Package)" build. (`Could not find file 'D:\BuildAgent\work\30546188361a242\doc\reference\reference.build'.`)

This also means that there may be little value into working for getting the md files deployed directly on https://github.com/nhibernate/nhibernate.github.io, because the build process will have anyway generated the html files that could then be deployed instead of the markdown.

So I think this PR should now focus on getting the documentation formats generated within the release build process. For this, I think the [`manual`](https://github.com/nhibernate/nhibernate-core/blob/5.1.0/doc/documentation.build#L27) Nant task should be changed. The Nant change part could be limited to causing that task to call another script that will do the work of generating the documentation formats into `${doc.out.dir}`, if this is more convenient than scripting it all with Nant.

By the way, once html pages are generated, this will allow to check how the doc looks.

A later PR for migrating to readthedocs could still be done, if we wish to benefit from some interesting features like automatic handling of documentation versions.

NikitaEgorov:
I use RST format in my projects!
Sphinx Doc it is great tool. Sphinx check all references

weelink:
Locally I have the documentation converted to RestructuredText. As @NikitaEgorov already mentioned, it does check for invalid references. So that is nice. The only possible problem to not switch is the build scripts. I'll take a look at that soon. Depending on how much impact it has, it might be a disappointing possibility that the docs stay in DocBook for now.

NikitaEgorov:
In our own projects, we have a docker image for generating html
``` docker
# It is image for build RST documentation
# Image includes sphinx, 'read the docs' theme, plugins for graphics
#
# VERSION 1.0.0

FROM alpine:3.5

RUN apk add --no-cache python2 py2-pip graphviz openjdk8 ttf-liberation \
    && export JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk \
    && pip install --upgrade pip \
    && pip install --trusted-host pypi.python.org sphinx \
    && pip install sphinx-rtd-theme \
    && pip install sphinx-autobuild \
    && pip install sphinxcontrib-plantuml \
    && rm -rf  /usr/share/man /tmp/* /var/cache/apk/*
```

And two scripts for build 
``` bat
@echo off
set currentPath=%cd%
set currentPath=%currentPath:\=/%

docker run --rm -m 2g -v %currentPath%:/docs *****/sphinx:latest sh -c "cd /docs && sphinx-build -n -W -d build/doctrees source build"
```
and watch

``` bat
@echo off
set currentPath=%cd%
set currentPath=%currentPath:\=/%

docker run --rm -m 2g -p 8000:8000 -v %currentPath%:/docs ***/sphinx:latest sh -c "cd /docs && sphinx-autobuild -d build/doctrees -H 0.0.0.0 -p 8000 --poll source build"
```

weelink:
Small update: the layout for the PDF is giving some problems, but it should be possible to generate a readable PDF

PR #1658, opened 2018-04-12
Add missing cache setting

### LINKED ISSUE #1650: Cannot use cache.use_sliding_expiration in hibernate.cfg.xml
When I include `<property name="cache.use_sliding_expiration">true</property>` in a hibernate.cfg.xml file, I received the following exception:

`An exception occurred parsing configuration :The 'name' attribute is invalid - The value 'cache.use_sliding_expiration' is invalid according to its datatype 'String' - The Enumeration constraint failed.`

When I checked the source, **cache.use_sliding_expiration** is not defined in nhibernate-configuration.xsd

fredericDelaporte:
Fix #1650 in 5.0.x

### DIFF
--- a/build-common/NHibernate.props
+++ b/build-common/NHibernate.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <VersionMajor Condition="'$(VersionMajor)' == ''">5</VersionMajor>
     <VersionMinor Condition="'$(VersionMinor)' == ''">0</VersionMinor>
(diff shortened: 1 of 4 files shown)


fredericDelaporte:
I intend to add the release commit here directly, is it fine for you?

fredericDelaporte:
(And I have already added some doc annotation for that setting, but not yet force pushed.)

hazzik:
Yes, I’m fine with the release notes bearing added here

PR #1659, opened 2018-04-12
Merge 5.0.4 into 5.1.x

fredericDelaporte:
Should this be squashed or merged further? It looks like merging in such case may bloat the history with many merge commits.

And do we release 5.1.2 at once, or do we wait in case some other things are to be fixed in 5.1.x?


hazzik:
You already have a merge commit here. You can just fast-forward (with cli tools).

PR #1661, opened 2018-04-13
Fix persistent collections' CopyTo implementations

### LINKED ISSUE #1660: Wrong CopyTo implementation
It seems that most ICollection.CopyTo implementations on Nhibernate are wrong.

For example at:
https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Collection/Generic/PersistentGenericBag.cs#L95

Index parameter is the index on the target array the data should be copied to. So, the correct implementation would be something like:
```
void ICollection.CopyTo(Array array, int index)
{
  for (var i = 0; i < Count; i++)
  {
    array.SetValue(this[i], index + i);
  }
}
```

I detected that after facing issues with .NET Core methods that uses CopyTo.

SelectMany fails:
```
var bags = new[]
{
    new PersistentGenericBag<string>(session, new [] {
        "A"
    }),
    new PersistentGenericBag<string>(session, new[] {
        "B"
    })
};

var items = bags.SelectMany(b => b).ToArray();
//Items should be ["A", "B"], but it is ["A", null]
```

Lists AddRange fails:
```
var list = new List<string>();
list.Add("A");

var bag = new PersistentGenericBag<string>(session, new[] {
  "B"
});

list.AddRange(bag);

//Items should be ["A", "B"], but it is ["A", null]
```

hazzik:
Fixes #1660

### DIFF
--- a/src/NHibernate.Test/CollectionTest/PersistentCollectionsFixture.cs
+++ b/src/NHibernate.Test/CollectionTest/PersistentCollectionsFixture.cs
@@ -1,0 +1,40 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using NHibernate.Collection.Generic;
+using NUnit.Framework;
+
+namespace NHibernate.Test.CollectionTest
+{
+	[TestFixture]
+	public class PersistentCollectionsFixture
+	{
+		[Test]
+		public void SelectManyWorksCorrectly()
+		{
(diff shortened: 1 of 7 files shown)


lucaslorentz:
Good to know it will be on next minor release.

@hazzik Thanks for taking care of it so quickly.

Issue #1663, opened 2018-04-16
Future queries not executed if no support for multiple queries


StormimOn:
NHibernate 5.X did change a little the Future extensions methods.
This cause a regression if there is no support for multiple queries, because in some scenarios the future queries won't be called.

For example

```c#
var query = Session.QueryOver<ParentObject>()....Future<ParentObject>();
Session.QueryOver<ParentObject>().Fetch(x => x.ChildCollection1)....Future<ParentObject>();
Session.QueryOver<ParentObject>().Fetch(x => x.ChildCollection2)....Future<ParentObject>();
...
Session.QueryOver<ParentObject>().Fetch(x => x.ChildCollectionN)....Future<ParentObject>();

var result = query.SingleOrDefault();
```

In this case the queries fetching childs collections won't be called. 
With NH 3.X the `.Future()` would react like a `.List()` so there was no problem.

fredericDelaporte:
Well, this is indeed a missing breaking change in release notes. But it has been done on purpose, see [here](https://github.com/nhibernate/nhibernate-core/pull/642#discussion_r121978225).

> This also adjusts a "detail": previous implementation was causing an immediate execution of the query in case of lack of multiple query support, contrary to what do FutureValue. Now it will be delayed too, so a discarded future query will not have its query run at all when multiple queries are not supported, it will be run only if the future query is actually enumerated.

So the code above should be changed to:

```c#
var query = Session.QueryOver<ParentObject>()....Future<ParentObject>();
var childQueries = new List<IFutureEnumerable<ParentObject>>();
childQueries.Add(Session.QueryOver<ParentObject>().Fetch(x => x.ChildCollection1)....Future<ParentObject>());
childQueries.Add(Session.QueryOver<ParentObject>().Fetch(x => x.ChildCollection2)....Future<ParentObject>());
...
childQueries.Add(Session.QueryOver<ParentObject>().Fetch(x => x.ChildCollectionN)....Future<ParentObject>());

// Using the IFutureEnumerable directly as an IEnumerable is deprecated.
var result = query.GetEnumerable().SingleOrDefault();

// When Future are actually supported by the data provider, this code will be no-op.
// Otherwise it is mandatory for getting the children futures actually executed.
foreach(var cq in childQueries)
{
    cq.GetEnumerable();
}
```

[...]

PR #1664, opened 2018-04-16
Release 5.1.2

fredericDelaporte:
And add a missing 5.0.0 possible breaking change, see #1663.

The [GitHub 5.0.0 release](https://github.com/nhibernate/nhibernate-core/releases/tag/5.0.0) is not yet updated with that missing possible breaking change, I will do that once this PR is approved. (Update: now done.)

I am already preparing this release just for not forgetting about adding that possible breaking change.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 5.1.1.{build}
+version: 5.1.2.{build}
 image: Visual Studio 2017
 environment:
   matrix:
--- a/build-common/NHibernate.props
+++ b/build-common/NHibernate.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <VersionMajor Condition="'$(VersionMajor)' == ''">5</VersionMajor>
     <VersionMinor Condition="'$(VersionMinor)' == ''">1</VersionMinor>
-    <VersionPatch Condition="'$(VersionPatch)' == ''">1</VersionPatch>
+    <VersionPatch Condition="'$(VersionPatch)' == ''">2</VersionPatch>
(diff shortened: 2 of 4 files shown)


fredericDelaporte:
Rebased and 5.1.2 release notes updated with latest change.

fredericDelaporte:
Rebased again, line for #1670 added in release notes. I think I will release this soon (maybe tomorrow).

hazzik:
Ok.

fredericDelaporte:
AppVeyor build is quite flaky. This is annoying.

Anyway, I now consider fixing #1667 in 5.1.x, if it is considered a thing to fix in NHibernate (which would be safer).

hazzik:
>Anyway, I now consider fixing #1667 in 5.1.x

Go for it.

fredericDelaporte:
Of course. That is in the release procedure anyway.

fredericDelaporte:
Notes updated and branch rebased. (Contrary to the Release5.0.6 branch, which was not up-to-date and which build must not be used for the release. I am waiting the 5.0.x release build. Anyway, we are supposed to published from base branch builds rather than from PR branch builds.)

hazzik:
Do you plan to merge 5.0.6 to 5.1.x? Or are you willing to let them diverge?

fredericDelaporte:
That would be a "no-op" merge. I think that when we have started back-porting instead of fixing in the older branch then merging to newer branches, we should let the older branch diverge. This is why I have manually ported here the 5.0.6 release notes changes.

But if you rather not have any current major branches diverging, we can still merge it back to 5.1.x before releasing 5.1.2, even if this will somewhat pollute the history. (In such case we should avoid back-porting to 5.0.x but instead fix there then merge to newer branches.)

PR #1666, opened 2018-04-17
 Handle multi-queries support in FutureBatch 

fredericDelaporte:
The previous Future implementation was letting all implementors handling the lack of multi-queries support themselves, leading to many discrepancies in behavior with a data provider which did not support
multi-queries:

 * Accessing the future value was causing the query to be executed at each access.
 * Only the accessed future was executed, not all the other defined futures. This is the cause of #1663.
 * Futures having an ExecuteOnEval were not executing it, leading to NH-3850 bugs.

Now FutureBatch handles itself the multi-queries support or lack of support.

The second commit alters NH-3850 tests for having them executed on more databases. Their failures in case of lack of multi-queries support were not seen because they were not executed with a data provider lacking this support. (I intend to squash this commit.)


src/NHibernate/Impl/FutureBatch.cs
```diff
@@ -73,21 +73,36 @@ private IList GetResults()
...
-			foreach (var query in queries)
+			if (!session.Factory.ConnectionProvider.Driver.SupportsMultipleQueries)
 			{
-				AddTo(multiApproach, query.Query, query.ResultType);
-				if (query.Future?.ExecuteOnEval != null)
-					needTransformer = true;
+				results = new List<object>();
+				foreach (var query in queries)
+				{
+					var result = List(query.Query);
```
fredericDelaporte: Here is another "discrepancy" that I have not listed in the commit message: without multi-queries support, it was executing the generic typed version. Due to current `FutureBatch` implementation, we are back to executing the non-generic one. This is nearer to what is done with multi-queries support, although I would prefer having both using generic methods. But there are much more changes to do in `FutureBatch` and multi queries if we want to replace non-generic usages, so I leave that for a later PR.

src/NHibernate/Impl/FutureBatch.cs
```diff
@@ -73,21 +73,36 @@ private IList GetResults()
...
 
-			var multiApproach = CreateMultiApproach(isCacheable, cacheRegion);
-			var needTransformer = false;
-			foreach (var query in queries)
+			if (!session.Factory.ConnectionProvider.Driver.SupportsMultipleQueries)
 			{
-				AddTo(multiApproach, query.Query, query.ResultType);
-				if (query.Future?.ExecuteOnEval != null)
-					needTransformer = true;
+				results = new List<object>();
```
hazzik: I would prefer here to use a temporary variable:

1. Collect all results
2. Assign to a "results" field.

hazzik:
Regards the second commit. How relevant the `DateTimeOffset` fields to the tests? If they are not relevant at all (or have low relevance) - please replace the type with DateTime / remove the fields.

fredericDelaporte:
I would say, `DateTimeOffset` has low relevance, granted `DateTime` is tested, which is already the case. So ok for removing it.

hazzik:
I thought I already asked this... Maybe in my dream. Can we extract DateTimeOffset tests into own fixture?

fredericDelaporte:
> If they are not relevant at all (or have low relevance) - please replace the type with DateTime / remove the fields.

This has been done in a0e9fa3. I have understood it as a remove request, not as an "extract in another fixture" request. Now I can still add another commit for putting them in an additional fixture instead of removing them.

hazzik:
> I have understood it as a remove request

Yes, this is what I asked initially. But then at some point I wanted to (and I thought that I did) ask to put these tests back as a separate fixture. Sorry for the confusion.

PR #280, opened 2014-07-05
NH-3630 - Add bitwise operation support for dialects using internal/external functions

amroel:
https://nhibernate.jira.com/browse/NH-3630
Changes to SqlGenerator and Dialect to allow injection of vendor specific bitwise functions
+
Bitwise functions for firebird

PR #1670, opened 2018-04-26
Fix bitwise functions thread safety

fredericDelaporte:
Functions instances are hold in the dialect, shared among sessions spawned from the same factory. They must be thread safe.

Implementation moved to its correct directory by the way.

### DIFF
--- a/src/NHibernate/Dialect/BitwiseNativeOperation.cs
+++ b/src/NHibernate/Dialect/BitwiseNativeOperation.cs
@@ -15,28 +15,26 @@ namespace NHibernate.Dialect
 	{
 		private readonly string _sqlOpToken;
 		private readonly bool _isNot;
-        private Queue _args;
-		private SqlStringBuilder _sqlBuffer;
 
 		/// <summary>
-		/// creates an instance using the giving token
+		/// Creates an instance using the giving token.
 		/// </summary>
 		/// <param name="sqlOpToken">
-		/// The operation token
+		/// The operation token.
 		/// </param>
 		/// <remarks>
-		/// Use this constructor only if the token DOES NOT represent a NOT-Operation
+		/// Use this constructor only if the token DOES NOT represent an unary operator.
 		/// </remarks>
 		public BitwiseNativeOperation(string sqlOpToken)
 			: this(sqlOpToken, false)
 		{
 		}
 
-        /// <summary>
-		/// creates an instance using the giving token and the flag indicating a NOT-Operation
-        /// </summary>
-        /// <param name="sqlOpToken"></param>
-        /// <param name="isNot"></param>
+		/// <summary>
+		/// Creates an instance using the giving token and the flag indicating if it is an unary operator.
+		/// </summary>
+		/// <param name="sqlOpToken">The operation token.</param>
+		/// <param name="isNot">Whether the operation is unary or not.</param>
 		public BitwiseNativeOperation(string sqlOpToken, bool isNot)
 		{
 			_sqlOpToken = sqlOpToken;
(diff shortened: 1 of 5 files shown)


src/NHibernate/Dialect/Function/BitwiseFunctionOperation.cs
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		public bool HasParenthesesIfNoArguments
+		{
+			get { return true; }
+		}
+
+		public SqlString Render(IList args, ISessionFactoryImplementor factory)
+		{
+			var filteredArgs = new Queue();
```
hazzik: Why would you need queue here?

src/NHibernate/Dialect/Function/BitwiseNativeOperation.cs
```diff
@@ -0,0 +1,87 @@
...
+
+		public bool HasParenthesesIfNoArguments
+		{
+			get { return false; }
+		}
+
+		public SqlString Render(IList args, ISessionFactoryImplementor factory)
+		{
+			var sqlBuffer = new SqlStringBuilder();
+			var arguments = new Queue(args);
```
hazzik: Again, why?

hazzik:
I understand that this is just a refactoring of a previous work. However, for me it looks incorrect. Could this be reworked? (Preferably with unit tests)

src/NHibernate/Dialect/Function/BitwiseFunctionOperation.cs
fredericDelaporte: I have done the fix in a "minimal implementation changes" way, mainly fixing the fact that putting some state without any locks in an instance shared by many threads cannot be thread safe. So I have kept the previous logic using a queue.

fredericDelaporte:
I do not call a fix a refactoring. The fact that current implementation is not thread safe seems obvious to me, as seems removing all mutable state to be a proper fix for achieving thread safety of instances. Now am I wrong considering function instances must be thread safe? Looking at the code, I do not think so.

Conceiving a test reliably demonstrating the thread safety failure while using sessions (so proving that thread safety is an issue) may be no easy task. The call to function rendering is just a very small part of the query creation processing, so having them occurring concurrently may be hard. There is also the query plan caching which may mitigate the trouble.

These functions are otherwise covered by some tests (Linq.WhereTests.BitwiseQuery/2/3, NH1192, NH1775, NH1831.)

src/NHibernate/Dialect/Function/BitwiseNativeOperation.cs
```diff
@@ -0,0 +1,87 @@
...
+
+			if (_isNot == false)
+				AddToBuffer(arguments.Dequeue(), sqlBuffer);
+
+			AddToBuffer(string.Format(" {0} ", _sqlOpToken), sqlBuffer);
```
hazzik: Should be just `sqlBuffer.Add(string.Format(" {0} ", _sqlOpToken))`, or better 

```
sqlBuffer.Add(" ").Add(_sqlOpToken).Add(" ");
```

fredericDelaporte:
Well, I did want to fix the direst trouble, thread safety, which can result in hard to debug troubles for users (rare intermittent failures). For the other shortcomings of the implementation, I have not considered them. If they cause issues, it is likely they are far easier to diagnose than a thread safety one. Or are you saying my changes introduce new issues?

This code comes from #280. @amroel, any remembering on the subject, like about:

> It is not clear for me why this class worries about parens, and `BitwiseNativeOperation` does not? Where these assumptions came from? What would happen if some parens would end up as an arguments to BitwiseNativeOperation?

src/NHibernate/Dialect/Function/BitwiseFunctionOperation.cs
```diff
@@ -0,0 +1,81 @@
...
+			foreach (var arg in args)
+			{
+				if (!IsParens(arg.ToString()))
```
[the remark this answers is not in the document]
fredericDelaporte: Thinking about it, it might be just an optimization, for removing parens which would be redundant otherwise, because the function call forcibly adds its own parens.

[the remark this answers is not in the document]
fredericDelaporte: It appears to be a hack for the BitwiseQuery3 test. The args contains four elements instead of two, the first being actually an element, the second an opening parenthesis and the fouth a closing parenthesis. Outputing them all comma-separated could only fail. For this case ignoring the parens yield a good result. Could more elaborated queries yield more complex cases requiring more subtle handling? I do not know for sure. But fiddling with the query, I cannot get such a case.

The native case does not bother with it, because parens seem to appears only as second and fourth argument, while the operator is put only after the first argument. So it get suitably placed, and parens do not creates an issue there.

That seems a bit brittle, but it is needed and it seems there are no known failures with that.

fredericDelaporte:
Thanks for the removal of the cherry-pick in 5.1.x. Now, do we simply merge this in 5.2, or should we fix the thread safety issue in previous versions?
I am more for the later, at least 5.1.x. This issue could bite anyone using a binary operator in a query in a multi-threaded application (so potentially many web applications), and that would likely cause infrequent failures, causing the application to lack reliability.

hazzik:
Let's do 1st and 4th commits (squashed) into 5.1.x (we would need to exclude PostgreSQL from some tests as I understand).

fredericDelaporte:
Base branch of PR changed, commit 1 & 4 squashed (and adjusted for ignoring PostgreSQL in bitwise xor tests), other commits dropped.

src/NHibernate/Dialect/BitwiseNativeOperation.cs
```diff
@@ -15,28 +15,26 @@ public class BitwiseNativeOperation : ISQLFunction
 	{
 		private readonly string _sqlOpToken;
 		private readonly bool _isNot;
```
fredericDelaporte: I have not kept the renaming to `_isUnary` in order to avoid a grey area breaking change (serialization) in a release version (5.1.x). Same for constructor argument name, kept as `isNot` instead of `isUnary`.

PR #1674, opened 2018-04-30
WIP - Update AsyncGenerator to v0.11.0 to fix #1672

### LINKED ISSUE #1672: Generated async methods do not correctly propagate OperationCanceledException
a number of the generated async methods will swallow OperationCanceledException and cause the resulting tasks to transition to the Faulted state instead of the Cancelled state when cancelling queries.

I have raised an issue over at https://github.com/maca88/AsyncGenerator/issues/90 but in the meantime I believe a solution is to add `catch (OperationCanceledException) { throw; } ` ahead of  `catch (Exception) ` where the catch block is catching an ancestor type of OperationCanceledException

EamonHetherton:
And regenerate the Async classes.
See #1672.

### DIFF
--- a/default.build
+++ b/default.build
@@ -50,6 +50,8 @@
 			<arg value="msbuild" />
 			<arg value="/t:Restore" />
 			<arg value="${root.dir}/src/NHibernate.sln" />
+			<arg value="/p:Platform=&quot;Any CPU&quot;" />
+			<arg value="/p:Configuration=&quot;${build.config}&quot;"/>
 		</exec> 
 	</target>
 
--- a/src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
+++ b/src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
@@ -83,36 +83,22 @@ namespace NHibernate.Test.Linq
 							" name and product availability.")]
 		public Task DLinq13Async()
 		{
-			try
-			{
-				var q =
+			var q =
 				from p in db.Products
 				select new {p.Name, Availability = p.UnitsInStock - p.UnitsOnOrder < 0 ? "Out Of Stock" : "In Stock"};
 
-				return ObjectDumper.WriteAsync(q, 1);
-			}
-			catch (Exception ex)
-			{
-				return Task.FromException<object>(ex);
-			}
+			return ObjectDumper.WriteAsync(q, 1);
 		}
 
 		[Category("SELECT/DISTINCT")]
 		[Test(Description = "This sample uses SELECT and a known type to return a sequence of employees' names.")]
 		public Task DLinq14Async()
 		{
-			try
-			{
-				IQueryable<Name> q =
+			IQueryable<Name> q =
 				from e in db.Employees
 				select new Name {FirstName = e.FirstName, LastName = e.LastName};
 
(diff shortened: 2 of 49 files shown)


fredericDelaporte:
Since you have added maca88/AsyncGenerator#96, I guess this should be considered WIP for now.

default.build
```diff
@@ -50,6 +50,8 @@
 			<arg value="msbuild" />
 			<arg value="/t:Restore" />
 			<arg value="${root.dir}/src/NHibernate.sln" />
+			<arg value="/p:Platform=&quot;Any CPU&quot;" />
+			<arg value="/p:Configuration=&quot;${build.config}&quot;"/>
```
fredericDelaporte: Why these changes?

fredericDelaporte: Check you environment variables. You likely have a configuration or platform one defined in the command line you use for generating async code. They must be unset to avoid having their values messing with MsBuild parameters.

src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
```diff
@@ -83,56 +83,35 @@ public class LinqQuerySamplesAsync : LinqTestCase
...
 				from p in db.Products
 				select new {p.Name, Availability = p.UnitsInStock - p.UnitsOnOrder < 0 ? "Out Of Stock" : "In Stock"};
 
-				return ObjectDumper.WriteAsync(q, 1);
-			}
-			catch (Exception ex)
-			{
-				return Task.FromException<object>(ex);
-			}
+			return ObjectDumper.WriteAsync(q, 1);
```
hazzik: @maca88 is there a way to turn off pass-through optimization for async code? Eg. always generate async-await?

maca88: Currently there isn't. Do you want an `async-await` in order to have this line in the stack trace or is something else that you want to achieve?

hazzik: The intention is to get clearer code for tests, or to be able to opt-out

maca88: [Will add an option](https://github.com/maca88/AsyncGenerator/issues/97) for this.

default.build
EamonHetherton: After a reboot it worked again without the specific platform and configuration switches.  However given that these are explicitly specified for the build, is it not also worthwhile doing so for the solution-restore step?  Would avoid similar unexpected issues in future.

fredericDelaporte: Granted. Still a bit unrelated to this PR and commit.

EamonHetherton: That's true

hazzik:
Fixed by #1684 + #1685

fredericDelaporte:
I think we need to keep something opened, because merging 5.1.x in master will currently only temporarily fix the trouble: the master AsyncGenerator is a 0.10 version which does not yet handle properly cancellation exceptions. When merging, I do not think we should downgrade the generator to 0.8.2.4, so the next AsyncGeneration would lose the proper exception handling.

hazzik:
> the master AsyncGenerator is a 0.10 version 

But the master is on 0.8.2.1, so this can be really closed.

PR #1682, opened 2018-05-07
Add support for System.MathF methods

hazzik:
Only in .NET Core 2.0 +


hazzik:
The only other reason is that users should be able to use a convenient function and not what's forced by some library.

PR #1684, opened 2018-05-07
Update AsyncGenerator to 0.8.2.2

hazzik:
Code regenerated to add missing OperationCanceledException where required.

Fixes #1674

### DIFF
--- a/src/NHibernate.Test/Async/NHSpecificTest/NH1553/MsSQL/SnapshotIsolationUpdateConflictTest.cs
+++ b/src/NHibernate.Test/Async/NHSpecificTest/NH1553/MsSQL/SnapshotIsolationUpdateConflictTest.cs
@@ -77,16 +77,17 @@ namespace NHibernate.Test.NHSpecificTest.NH1553.MsSQL
 		/// Tests, that NHibernate detects the update conflict and returns a StaleObjectStateException as expected.
 		/// This behaviour is part of the standard NHibernate feature set.
 		/// </summary>
+		/// <param name="cancellationToken">A cancellation token that can be used to cancel the work</param>
 		[Test]
-		public async Task UpdateConflictDetectedByNHAsync()
+		public async Task UpdateConflictDetectedByNHAsync(CancellationToken cancellationToken = default(CancellationToken))
 		{
-			Person p1 = await (LoadPersonAsync());
-			Person p2 = await (LoadPersonAsync());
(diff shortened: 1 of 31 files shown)


src/NHibernate.Test/Async/NHSpecificTest/NH1553/MsSQL/SnapshotIsolationUpdateConflictTest.cs
```diff
@@ -77,16 +77,17 @@ private void SavePerson(Person p)
 		/// Tests, that NHibernate detects the update conflict and returns a StaleObjectStateException as expected.
 		/// This behaviour is part of the standard NHibernate feature set.
 		/// </summary>
+		/// <param name="cancellationToken">A cancellation token that can be used to cancel the work</param>
 		[Test]
-		public async Task UpdateConflictDetectedByNHAsync()
+		public async Task UpdateConflictDetectedByNHAsync(CancellationToken cancellationToken = default(CancellationToken))
```
hazzik: @maca88 that's strange

maca88: Sorry, I overlooked this change. I've released version `0.8.2.3` which doesn't generate it.

fredericDelaporte: Thanks again @maca88, but these changes are still there with 0.8.2.3.

maca88: I tried to regenerate the code twice before releasing and the change was never there. Then I remembered that this issue occurred randomly depending on the order of the methods being analyzed, which was discovered and fixed when [#86](https://github.com/maca88/AsyncGenerator/issues/86) was done. I've ported [the fix](https://github.com/maca88/AsyncGenerator/commit/9e9b450bba341c30aa7384818475a2e992d9941a) and released it in `0.8.2.4`. Again, sorry for the confusion.

fredericDelaporte: No trouble, do not be sorry, it is already great you take the time to back-port things to old versions.

PR #1687, opened 2018-05-09
Update NUnit to 3.10.1

hazzik:
NUnitLite 3.10 has fixed `--nocolor` option which would be really helpful for Travis CI: Travis CI does not have a concept of a unit tests explorer so to analyse failure one needs to check the log file. But unfortunately with broken `--nocolor` option this is not always a trivial task.


hazzik:
Hahaha. The `--nocolor` option works, but one test for Postgres is failing.... Sad...

hazzik:
(to future me): TODO: mark as KnownBug and regenerate tests

PR #1688, opened 2018-05-09
Merge 5.1.2 into master

hazzik:


hazzik:
I'll just wait tests to pass and then will merge it.

src/NHibernate.Test/TestCase.cs
```diff
@@ -447,6 +447,19 @@ protected DateTime RoundForDialect(DateTime value)
...
+					"bxor",
+					new HashSet<System.Type>
+					{
+						// Could be supported like Oracle, with a template
+						typeof (SQLiteDialect),
+						// Could be supported by overriding registration with # instead of ^
+						typeof (PostgreSQLDialect),
+						typeof (PostgreSQL81Dialect),
+						typeof (PostgreSQL82Dialect),
+						typeof (PostgreSQL83Dialect)
```
fredericDelaporte: So, the 5 lines above are obsolete, in regard to current master state. Should it be change as part of the merge? Or as a later commit? (If change as a later commit, as already written somewhere else, I intend to do it while adding support for bxor in `SQLiteDialect`.)

src/NHibernate/nhibernate-configuration.xsd
```diff
@@ -92,7 +92,16 @@
 										<xs:enumeration value="cache.region_prefix" />
 										<xs:enumeration value="cache.use_minimal_puts" />
 										<xs:enumeration value="cache.default_expiration" />
-										<xs:enumeration value="query.substitutions" />
+										<xs:enumeration value="cache.use_sliding_expiration">
+									<xs:annotation>
```
fredericDelaporte: The indentation here is not correct, due to master being more deeply nested (`xs:union` added on master).

PR #1693, opened 2018-05-14
Implement SurrogateSelector

### LINKED ISSUE #1652: Provide example for a .NET Core App serializing a session.
From the release notes on .Net Core 2.0 framework compatibility:

>Binary serialization is not supported - the user shall implement serialization surrogates for System.Type,
>FieldInfo, PropertyInfo, MethodInfo, ConstructorInfo, Delegate, etc.

This issue replaces the pull request #1425 as a direction was chosen to not provide in-the-library support for session serialization on the new frameworks.  This is probably for the best as the pull-request was fairly invasive.

Application level support using serialization surrogates will be more elegant, but an example implementation, and tests, of those surrogates would certainly help.

If your scenario actually needs this support, and you want to migrate to dotnet, then this is the issue to vote on.

hazzik:
- Supports System.Type, System.Reflection.MemberInfo and System.Delegate

Closes #1652

### DIFF
--- a/src/NHibernate.Test/Legacy/FooBarTest.cs
+++ b/src/NHibernate.Test/Legacy/FooBarTest.cs
@@ -2632,8 +2632,6 @@ namespace NHibernate.Test.Legacy
 		[Test]
 		public void PersistCollections()
 		{
-			TestsContext.AssumeSystemTypeIsSerializable();
-
 			ISession s = OpenSession();
 			ITransaction txn = s.BeginTransaction();
 			IEnumerator enumer = s.CreateQuery("select count(*) from b in class Bar").Enumerable().GetEnumerator();
@@ -2831,7 +2829,20 @@ namespace NHibernate.Test.Legacy
 
 			// serialize and then deserialize the session.
 			Stream stream = new MemoryStream();
-			IFormatter formatter = new BinaryFormatter();
+#if NETFX
+			var formatter = new BinaryFormatter();
+#else
+			var selector = new SurrogateSelector();
+			selector.AddSurrogate(
+				typeof(CultureInfo),
+				new StreamingContext(StreamingContextStates.All),
+				new CultureInfoSerializationSurrogate());
+			selector.ChainSelector(new SerializationHelper.SurrogateSelector());
+			var formatter = new BinaryFormatter
+			{
+				SurrogateSelector = selector
+			};
+#endif
 			formatter.Serialize(stream, s);
 
 			s.Close();
@@ -4647,8 +4658,6 @@ namespace NHibernate.Test.Legacy
 		[Test]
 		public void ProxyArray()
 		{
-			TestsContext.AssumeSystemTypeIsSerializable();
-
 			ISession s = OpenSession();
 			GlarchProxy g = new Glarch();
 			Glarch g1 = new Glarch();
@@ -4685,7 +4694,12 @@ namespace NHibernate.Test.Legacy
 
 			// serialize the session.
 			Stream stream = new MemoryStream();
-			IFormatter formatter = new BinaryFormatter();
+			var formatter = new BinaryFormatter()
+			{
+#if !NETFX
+				SurrogateSelector = new SerializationHelper.SurrogateSelector()	
+#endif
+			};
 			formatter.Serialize(stream, s);
 
 			// close the original session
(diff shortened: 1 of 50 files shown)


src/NHibernate.Test/Legacy/FooBarTest.cs
```diff
@@ -2831,7 +2829,20 @@ public void PersistCollections()
 
 			// serialize and then deserialize the session.
 			Stream stream = new MemoryStream();
-			IFormatter formatter = new BinaryFormatter();
+#if NETFX
+			var formatter = new BinaryFormatter();
+#else
+			var selector = new SurrogateSelector();
+			selector.AddSurrogate(
```
hazzik: Somewhere deep in the `Baz` hierarchy some class has few properties of `CultureInfo` type. Unfortunately it is not marked as serializable in .NET Core / .NET Standard.

src/NHibernate/Util/SerializationHelper.cs
```diff
@@ -1,32 +1,60 @@
...
+using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Binary;
 
 namespace NHibernate.Util
 {
-	public class SerializationHelper
+	public static partial class SerializationHelper
 	{
-		private SerializationHelper()
+		private static readonly BinaryFormatter Formatter = new BinaryFormatter
```
fredericDelaporte: It cannot be stored because it is [not thread safe](https://msdn.microsoft.com/en-us/library/system.runtime.serialization.formatters.binary.binaryformatter(v=vs.110).aspx#Anchor_8). It must be created on a per usage basis (or at least put in a thread static variable).

src/NHibernate/Util/SerializationHelper.Surrogates.cs
```diff
@@ -0,0 +1,312 @@
...
+
+			ISurrogateSelector ISurrogateSelector.GetNextSelector()
+				=> _next;
+
+			ISerializationSurrogate ISurrogateSelector.GetSurrogate(
+				System.Type type,
+				StreamingContext context,
+				out ISurrogateSelector selector)
+			{
+				if (typeof(System.Type).IsAssignableFrom(type) && !SystemTypeIsSerializable)
```
fredericDelaporte: I would test the boolean first, as it is likely more lightweight than the assignable test.

src/NHibernate/Util/SerializationHelper.Surrogates.cs
```diff
@@ -0,0 +1,312 @@
...
+				catch (SerializationException)
+				{
+					return false;
+				}
+			}
+
+			private ISurrogateSelector _next;
+
+			void ISurrogateSelector.ChainSelector(ISurrogateSelector selector)
+				=> _next = selector;
```
fredericDelaporte: The interface contract is not explicit, but looking at [`System.Runtime.Serialization.SurrogateSelector` code](https://github.com/Microsoft/referencesource/blob/4.6.2/mscorlib/system/runtime/serialization/surrogateselector.cs#L81), it seems it is supposed to handle multiple chaining without losing the previous one as does the code here. (It also does extensive checking for preventing cycles.)

The logic for chaining is to set `selector` as the next one and if there was already another next, then chain it to the last next of `selector` (recursive call to GetNextSelector until finding null, then chaining to the last not null).

Rather than re-coding this it would be better if `SurrogateSelector` implementation could be reused. Otherwise maybe harden the checks here for stating our own contract, like throwing if `_next` is not null and tell to chain on `GetNextSelector` instead.

hazzik: It is possible to just inherit SurrogateSelector.

src/NHibernate/Util/SerializationHelper.Surrogates.cs
hazzik: Ok

src/NHibernate/Util/SerializationHelper.cs
hazzik: Ok

src/NHibernate/Util/SerializationHelper.Surrogates.cs
hazzik: In fact, 99.999% of the time the boolean would be true, so it does not really matter.

fredericDelaporte: Indeed, since the whole file is `!NETFX`.

PR #1694, opened 2018-05-14
Implement CollectionHelper.GetHashCode that accepts IEqualityComparer<T>

hazzik:
- Also fix CollectionHelper.GetHashCode to account for the length of the collections

This is an alternative implementation to #1613

### DIFF
--- a/src/NHibernate.Test/UtilityTest/CollectionHelperFixture.cs
+++ b/src/NHibernate.Test/UtilityTest/CollectionHelperFixture.cs
@@ -1,4 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using NHibernate.Engine;
 using NHibernate.Util;
 using NUnit.Framework;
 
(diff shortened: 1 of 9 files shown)


fredericDelaporte:
`NamedParameterComparer` remains not fully tested (not checking hashcode are better with it), but testing it would be easier done with #1612 fixes of `QueryKey`/`FilterKey` tests.

src/NHibernate.Test/UtilityTest/CollectionHelperFixture.cs
```diff
@@ -180,6 +182,69 @@ public void SetsWithoutSameCountShouldBeInequal()
...
+		public void MapsWithSameContentShouldHaveSameHashCode()
+		{
+			var d1 = new Dictionary<string, string> { { "1", "2" }, { "3", "4" } };
+			var d2 = new Dictionary<string, string> { { "1", "2" }, { "3", "4" } };
+			Assert.That(
+				CollectionHelper.GetHashCode(d1, KvpStringComparer.Instance),
+				Is.EqualTo(CollectionHelper.GetHashCode(d2, KvpStringComparer.Instance)));
+		}
+
+		// Failure of following tests is not an error from GetHashCode semantic viewpoint, but it causes it
```
hazzik: Better to replace asserts with warnings then

fredericDelaporte: The trouble with such warnings is that their failures will likely be unnoticed. A test that may just generate warnings without having any assert does no good in my opinion.

Even if the `GetHashCode` semantics allows collision, having collisions on simple and recurring cases is a bad thing and is worth a failing test case in my opinion.

hazzik: NUnit’s warnings mark tests as inconclusive.

fredericDelaporte: Yes but who checks inconclusive tests? If the build does not fail, a PR causing a regression on this point is very likely to not be rejected due to causing that regression, because it will likely goes unnoticed.

PR #1696, opened 2018-05-15
Fixed CriteriaImpl.Clone for readonly query

bahusoid:
While fixing https://github.com/nhibernate/nhibernate-core/pull/1681 I've noticed that readonly setting is also lost after Clone


fredericDelaporte:
You mean, this fix an old bug actually unrelated to #1681 but likely seen by the way? So

> Noticed that readonly setting is also lost after Clone

Should be understood as:

> I have noticed that the readonly setting is also lost after Clone. This PR fixes that loss.

bahusoid:
Yep. Updated description to make it more clear

PR #1699, opened 2018-05-16
Add ability to load types from in-memory-only assemblies

### LINKED ISSUE #961: NH-4080 - Load virtual assembly for mapping
**Trifle.pl** created an issue — <time datetime="2017-09-22T13:23:51.700Z" title="2017-09-22T13:23:51.700Z">22nd September 2017, 13:23:51</time>:
> Add functionality to load virtual assembly created in application memory for get types from loaded assemblies. Assembles need to be add before use function to add mapping to configuration.
> 
> Class "\util\ReflectHelperMemoryAssembly.cs" for store added assembly.
> Used by method "TypeFromAssembly()" in "ReflectHelper.cs".
> 
> Use new functionality by adding assembly in Configuration method "AddMemoryAssembly()".
> 
> It is similar method in Configuration "AddAssembly()", but that method don't work with virtual assembly.
> 
> 
> 
---
**Trifle.pl** added a comment — <time datetime="2017-09-22T13:48:42.919Z" title="2017-09-22T13:48:42.919Z">22nd September 2017, 13:48:42</time>:
> Link to new functionality  https://github.com/Triflepl/nhibernate-core/tree/4.1.x
>

hazzik:
GetType only works on assemblies loaded from disk. To overcome this limitation we need to supply an assembly resolver.


src/NHibernate.Test/CfgTest/LoadMemoryOnlyAssembyTest.cs
```diff
@@ -0,0 +1,107 @@
...
+						var text = "Compile error: " +
+						           string.Join(System.Environment.NewLine, compile.Errors.Cast<CompilerError>());
+
+						Assert.Fail(text);
+					}
+
+					return compile.CompiledAssembly;
+				}
+			}
+			catch (PlatformNotSupportedException e)
```
hazzik: At least on Linux/MacOS X `CompileAssemblyFromSource` throws an exception. It is possible that it is supported on .NET Core on Windows.

fredericDelaporte: For .Net Core on Windows, it fails with message telling it is not supported on that platform. (But for some reason the message reported in Rider test UI is another quite meaningless one: "Outcome 0 is not understood". Maybe a NUnit bug or a Rider one. At least the status is inconclusive.)

hazzik:
Yay! Less than 50 opened PRs!

PR #1700, opened 2018-05-17
Single place to specify TargetFrameworks

bahusoid:
It's better to have single place for build configuration;

But my main reason:
When you are on some shitty ultrabook and want to quickly disable other targets to make compilation less painful :)

### DIFF
--- a/build-common/NHibernate.props
+++ b/build-common/NHibernate.props
@@ -11,7 +11,7 @@
     <FileVersion>$(VersionPrefix).0</FileVersion>
 
     <NhAppTargetFrameworks>net461;netcoreapp2.0</NhAppTargetFrameworks>
-    <NhLibTargetFrameworks>net461;netcoreapp2.0;netstandard2.0;</NhLibTargetFrameworks>
+    <NhLibTargetFrameworks>net461;netcoreapp2.0;netstandard2.0</NhLibTargetFrameworks>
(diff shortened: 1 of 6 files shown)


build-common/NHibernate.props
```diff
@@ -10,6 +10,9 @@
     <AssemblyVersion>$(VersionMajor).$(VersionMinor).0.0</AssemblyVersion>
     <FileVersion>$(VersionPrefix).0</FileVersion>
 
+    <NhAppTargetFrameworks>net461;netcoreapp2.0</NhAppTargetFrameworks>
+    <NhLibTargetFrameworks>net461;netcoreapp2.0;netstandard2.0;</NhLibTargetFrameworks>
```
fredericDelaporte: The trailing `;` after netstandard2.0 creates a glitch in Rider (it adds a `.NETFramework, Version=v0.0` target). It should be removed.

bahusoid: Shame on Rider for this :)

Should I also add a comment in project files to make it more clear where those NhAppTargetFrameworks/NhLibTargetFrameworks are defined? Something like:
"TargetFrameworks variables are defined in NHibernate.props"

fredericDelaporte: No, there is only one include to check.

PR #1703, opened 2018-05-21
Remove dialect instantiation in AddDeserializedMapping

### LINKED ISSUE #787: NH-1934 - Cache Dialect instance instead create for each hbm
**BNReddy** created an [issue](https://nhibernate.jira.com/browse/NH-1934) — <time datetime="2009-08-17T07:26:09.257Z" title="2009-08-17T07:26:09.257Z">17th August 2009, 7:26:09</time>:
> When I am running my application what I have observed is `Dialect.GetDialect(IDictionary<string, string> props)` is called more than 100 times.
> Reason is for each `Configuration.AddXml(string xml)` call above method is called. Finally for `Configuration.BuildSessionFactory()`. These calls create many objects. My application is having 150+ hbm.xml files with complicated collections.
> 
> Is there any specific reason to create Dialect for each call for one database dialect? why can't we cache each dialect for each configuration object.
> 
> Or 
> 
> Is there any other way to reduce these calls?
> 
> One more point is my application doesn't reuiqred to add all hbm.xml mapping files at same point of time. And I am using singleton session factory.
> 
> Is there any other way to add hbm.xml file to already created session factory? 
> 
> Which will help me allot. 
> 
---
**ayenderahien** added a comment — <time datetime="2009-08-17T07:39:02.572Z" title="2009-08-17T07:39:02.572Z">17th August 2009, 7:39:02</time>:
> Do you have any perf numbers for the call to GetDialect ?
> I don't think it should be expensive.
> 
---
**Fabio Maulo** added a comment — <time datetime="2009-08-19T13:45:23.625Z" title="2009-08-19T13:45:23.625Z">19th August 2009, 13:45:23</time>:
> This is not the place for questions.
> btw the session-factory, from your point of view, is immutable.
>

hazzik:
Replaces #787 

The sole reason the dialect was created in the mapping was to generate a name for PK's in Join and PersistentClass classes.

The previous code was bothered by invoking proper rules to escape the quoting characters. But, we do not need to be concerned by the escaping rules: instead, we try to trim the quotes if we get a quote character on the edge of wanted alias sub-string.


src/NHibernate/Cfg/Mappings.cs
```diff
@@ -260,7 +309,7 @@ public void AddImport(string className, string rename)
 
 		public Table AddTable(string schema, string catalog, string name, string subselect, bool isAbstract, string schemaAction)
 		{
-			string key = subselect ?? dialect.Qualify(catalog, schema, name);
+			string key = subselect ?? BuildTableNameKey(catalog, schema, name);
```
hazzik: `key` here is an internal thing and is not exposed outside, so we are not concerning by dialect specific qualification of it (eg `catalog.schema.table` in SQL Server vs `catalog_schema_table` in SQLite and SQL Server Compact).

fredericDelaporte:
Is the "possible breaking change" the fact that some primary keys may have their name changed? Or something else I have missed?

hazzik:
There are following possible breaking changes:

1. Always null `Dialect` property in `BindMappingEventArgs`. It has moderate impact as it supposed to be a "user-facing" property.
2. Always null `Dialect` property in `Mappings` and `dialect` field in Binding classes. These has low impact as they are not a "user-facing" members.
3. The mentioned change in the PK name. It has a minuscule (or no) impact as it would affect only those who have a quoting symbols in the name of a table at precise 13th character.

fredericDelaporte:
Yes I should have thought about `Dialect` property switching to always `null`. This is not defined as a grey area breaking change. For including it in 5.2 rather than 6.0, a solution for avoiding yielding it null should be put in place. This could be done by providing a `Configuration` reference to `BindMappingEventArgs` and `Mappings` for allowing them to lazily get the dialect if their `Dialect` properties happen to be called.

hazzik:
@fredericDelaporte what about 2. (Mappings and Bindings)?

hazzik:
@fredericDelaporte what about `*Binding` classes?

fredericDelaporte:
For `*Binding` classes, that switches to always null a field visible to sub-classes, from a public non-sealed class. This is also strictly speaking a non-gray area breaking change, but it appears to me no user supplied deriving class can be provided to NHibernate for it to use them. There are no extension points for doing this. They are used as internal artifacts. Externally deriving from them has no purpose for using NHibernate.

It is then very likely than very few (if any) users may have done that. If some have done it, it could only serves purpose unrelated to NHibernate. So anyway they should have used something else as a base class. That is why I think we should accept this possible breaking change for a minor release.

(The same reasoning could be applied to `PersistentClass` indeed.)

PR #110, opened 2012-05-16
NH-3150: Improvements to Select Post Insert Generator

mickfold:
#822: Add support to Select Post Insert Generator for:
1. Natural id with more than one property.
2. IDBag.

PR #1707, opened 2018-05-24
Fix NH-3150: select id generator improvements

### LINKED ISSUE #822: NH-3150 - Select Post Insert Generator Improvements
**Michael Charalambous** created an [issue](https://nhibernate.jira.com/browse/NH-3150) — <time datetime="2012-05-16T10:53:13.026Z" title="2012-05-16T10:53:13.026Z">16th May 2012, 10:53:13</time>:
> The proposed changes adds the following functionality to the Select Post Insert Generator:
> 
> 1. Support for Natural id with more than one property.
> 2. Support for insertion into an IDBag collection.
> 
---
**Michael Charalambous** added a comment — <time datetime="2012-05-16T14:00:11.112Z" title="2012-05-16T14:00:11.112Z">16th May 2012, 14:00:11</time>:
> I have created a pull request containing the unit tests and proposed changes for the select POID
> 
> https://github.com/nhibernate/nhibernate-core/pull/110 
> 
> Mike
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2017-09-14T01:57:24.509Z" title="2017-09-14T01:57:24.509Z">14th September 2017, 1:57:24</time>:
> Move minor and trivial improvements to 5.1
>

fredericDelaporte:
Fixes #822 
Replaces #110 

The select id generator generates the id by getting it from the database with a select query filtered by an unique key. This unique key could only be a single entity property, forcing to use a component if the key has to spawn many properties. The first commit add support for keys spanning many properties without using a component.

The select id generator was not supporting idbags, although some preliminary work was already done for this. The second commit fixes this.

These two commits are based on #110 commit, rebased then split and altered on many points. (Breaking changes removed, redundant casts removed, syntax modernized, multiple properties support added to the key parameter and not only to natural-id, tests adjusted, ...)


src/NHibernate/Id/SelectGenerator.cs
```diff
@@ -72,19 +81,42 @@ public partial class SelectGeneratorDelegate : AbstractSelectingDelegate
...
-				idSelectString = persister.GetSelectByUniqueKeyString(uniqueKeyPropertyName);
-				uniqueKeyType = ((IEntityPersister) persister).GetPropertyType(uniqueKeyPropertyName);
+				if (persister is IEntityPersister entityPersister)
+				{
+					this.entityPersister = entityPersister;
+					uniqueKeyPropertyNames = DetermineNameOfPropertiesToUse(entityPersister, suppliedUniqueKeyPropertyNames);
+
+					uniqueKeyTypes = uniqueKeyPropertyNames.Select(p => entityPersister.GetPropertyType(p)).ToArray();
+				}
+				else if (persister is AbstractCollectionPersister collectionPersister)
```
hazzik: Probably should be `IQueryableCollection` instead

fredericDelaporte: Indeed, amended.

fredericDelaporte:
Rebased. (The async generator update does not change anything to this PR added tests.)

src/NHibernate/Id/SelectGenerator.cs
```diff
@@ -72,19 +81,42 @@ public partial class SelectGeneratorDelegate : AbstractSelectingDelegate
...
+					uniqueKeyPropertyNames[1] = collectionPersister.ElementColumnNames[0];
+					uniqueKeyTypes[1] = collectionPersister.ElementType;
+				}
+				else 
+				{
+					throw new IdentifierGenerationException(
+						$"Persister type {persister.GetType()} is not supported by Select Generator");
+				}
+
+				idSelectString = persister.GetSelectByUniqueKeyString(uniqueKeyPropertyNames);
```
hazzik: After a closer look, I think this all shall be hidden inside `GetSelectByUniqueKeyString`

src/NHibernate/Id/SelectGenerator.cs
```diff
@@ -72,19 +81,42 @@ public partial class SelectGeneratorDelegate : AbstractSelectingDelegate
...
+				{
+					this.entityPersister = entityPersister;
+					uniqueKeyPropertyNames = DetermineNameOfPropertiesToUse(entityPersister, suppliedUniqueKeyPropertyNames);
+
+					uniqueKeyTypes = uniqueKeyPropertyNames.Select(p => entityPersister.GetPropertyType(p)).ToArray();
+				}
+				else if (persister is IQueryableCollection collectionPersister)
+				{
+					uniqueKeyPropertyNames = new string[2];
+					uniqueKeyTypes = new IType[2];
```
hazzik: Are the `uniqueKeyPropertyNames` and `uniqueKeyTypes` set in this path used at all?

hazzik:
Overall solution looks like a hack (I'm not talking here about the workarounds to avoid breaking changes). It seems there could be a more elegant solution.

src/NHibernate/Id/IPostInsertIdentityPersister.cs
```diff
@@ -42,4 +68,23 @@ public interface IPostInsertIdentityPersister
...
 	}
-}
\ No newline at end of file
+
+	// 6.0 TODO: merge in IPostInsertIdentityPersister
+	/// <summary> 
+	/// An <see cref="IPostInsertIdentityPersister" /> that supports selecting by an unique key spanning
+	/// multiple properties.
+	/// </summary>
+	public interface IMultiPropertiesPostInsertIdentityPersister
```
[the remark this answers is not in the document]
fredericDelaporte: This is not the identity which is composite, but the key used to retrieve the id. So maybe `ICompositeKeyPostInsertIdentityPersister`.

fredericDelaporte:
That is a solution attempting minimal impact, likely because the author has not contributed much to NHibernate. I could have reworked it a bit more.

fredericDelaporte:
Rebased, and fixup commits addressing review added.

src/NHibernate/Id/SelectGenerator.cs
```diff
@@ -95,31 +71,65 @@ protected internal override SqlString SelectSQL
...
+				}
+			}
+
+			protected internal override void BindParameters(ISessionImplementor session, DbCommand ps, IBinder binder)
+			{
+				// 6.0 TODO: remove the "if" block and do the other TODO of this method
+				if (persister is IEntityPersister)
+				{
+#pragma warning disable 618
+					BindParameters(session, ps, binder.Entity);
```
fredericDelaporte: This is required to avoid a possible breaking change in the unlikely case someone uses a class overriding the obsoleted method. This causes `BindSelectByUniqueKey` of entity persisters to be currently not used.

Still more unlikely, this is maybe not enough: what if someone had implemented deriving classes correctly supporting idbag collections? But I do not see how we could handle that.

fredericDelaporte:
Rebased again. Since it is also about unique key, we could fear some conflicts with changes done for associations with property-ref. But there are none, because here this is about natural-ids, not property-ref.

hazzik:
The issue noted that this solution is a hack has not been resolved since I posted it.

fredericDelaporte:
> The issue noted that this solution is a hack has not been resolved since I posted it.

In June:

> Rebased, and fixup commits addressing review added.

I had reworked it right after your post. But now the fixups are squashed.

src/NHibernate/Id/IPostInsertIdentityPersister.cs
```diff
@@ -31,6 +104,8 @@ public interface IPostInsertIdentityPersister
 		/// column(s) to use in the select statement restriction.
 		/// </param>
 		/// <returns> The SQL select string </returns>
+		// Since 5.2
+		[Obsolete("Have the persister implement IMultiPropertiesPostInsertIdentityPersister and use its GetSelectByUniqueKeyString(string[] suppliedPropertyNames, out IType[] parameterTypes).")]
```
hazzik: The comment has to change to a new name

fredericDelaporte: Done.

(And rebased, commits directly amended.)

PR #1709, opened 2018-05-27
Lazy properties static proxy

fredericDelaporte:
This is a follow up to #1451.

The new static proxy factory was delegating the lazy properties proxy implementation to the old dynamic proxies.

With this PR, it does implement its own field interceptor, with far less reflection.

The PR also obsoletes dynamic proxies.


src/NHibernate/Proxy/FieldInterceptorProxyBuilder.cs
```diff
@@ -0,0 +1,298 @@
...
+		private static readonly MethodInfo FieldInterceptorInterceptMethod = FieldInterceptorType.GetMethod(nameof(IFieldInterceptor.Intercept));
+		private static readonly MethodInfo FieldInterceptorMarkDirtyMethod = FieldInterceptorType.GetMethod(nameof(IFieldInterceptor.MarkDirty));
+		private static readonly System.Type AbstractFieldInterceptorType = typeof(AbstractFieldInterceptor);
+		private static readonly FieldInfo AbstractFieldInterceptorInvokeImplementationField =
+			AbstractFieldInterceptorType.GetField(nameof(AbstractFieldInterceptor.InvokeImplementation));
+
+		public static TypeInfo CreateProxyType(System.Type baseType)
+		{
+			// Avoid having a suffix ending with "Proxy", for disambiguation with INHibernateProxy proxies
+			var typeName = $"{baseType.Name}ProxyForFieldInterceptor";
```
fredericDelaporte: Previous implementation was causing field interceptor proxies to be build with same type and assembly name than `INHibernateProxy`. Apparently this was not causing any issue, even when lazily loading an entity as a `INHibernateProxy` loading its target implementation as a field interceptor proxy due to the entity having lazy properties.
Still it seems to me cleaner to avoid the confusion.

src/NHibernate/Proxy/FieldInterceptorProxyBuilder.cs
```diff
@@ -0,0 +1,298 @@
...
+			var serializableConstructor = typeof(SerializableAttribute).GetConstructor(System.Type.EmptyTypes);
+			var customAttributeBuilder = new CustomAttributeBuilder(serializableConstructor, Array.Empty<object>());
+			typeBuilder.SetCustomAttribute(customAttributeBuilder);
+			// Default serialization is enough to handle serialization/de-serialization in the same process.
+			// For inter-process serialization, the proxy factory should be serialized too, for ensuring
+			// re-creation of the proxy type. The static NHibernateProxy does only that, letting the entity state
+			// be restored by lazy load. The field interceptor proxy cannot do that: it only knows how to load
+			// lazy properties, and not the other ones. (Moreover, its session is not restored on persistence
+			// context deserialization, contrary to NHibernateProxy. This causes deserialized proxies to only be
+			// usable for accessing data already loaded before serialization.)
```
fredericDelaporte: Those limitations are the same than those of dynamic proxies serialization of a field interceptor proxy.

src/NHibernate/Proxy/NHibernateProxyBuilder.cs
```diff
@@ -24,7 +23,6 @@ internal class NHibernateProxyBuilder
 		private static readonly PropertyInfo LazyInitializerIdentifierProperty = LazyInitializerType.GetProperty(nameof(ILazyInitializer.Identifier));
 		private static readonly MethodInfo LazyInitializerInitializeMethod = LazyInitializerType.GetMethod(nameof(ILazyInitializer.Initialize));
 		private static readonly MethodInfo LazyInitializerGetImplementationMethod = LazyInitializerType.GetMethod(nameof(ILazyInitializer.GetImplementation), System.Type.EmptyTypes);
-		private static readonly IProxyAssemblyBuilder ProxyAssemblyBuilder = new DefaultProxyAssemblyBuilder();
```
fredericDelaporte: I have not seen any interest in keeping the `NHibernate.Proxy.DynamicProxy.IProxyAssemblyBuilder` pattern now that it was used only by non-injectable private members. So I have moved its features to `ProxyBuilderHelper`.

src/NHibernate/Proxy/ProxyBuilderHelper.cs
```diff
@@ -0,0 +1,188 @@
...
+	{
+#if NETFX
+		private static bool _saveAssembly;
+		private static string _saveAssemblyPath;
+
+		// Called by reflection
+		internal static void EnableDynamicAssemblySaving(bool enable, string saveAssemblyPath)
+		{
+			_saveAssembly = enable;
+			_saveAssemblyPath = saveAssemblyPath;
```
fredericDelaporte: `DefaultProxyAssemblyBuilder` was enabling that with compilation directives, which was not very practical. Moreover, it was not handling the save path, which was further complicating using the feature with potentially more file system access right troubles.

src/NHibernate/Proxy/StaticProxyFactory.cs
```diff
@@ -43,9 +44,16 @@ public override INHibernateProxy GetProxy(object id, ISessionImplementor session
 
 		public override object GetFieldInterceptionProxy(object instanceToWrap)
 		{
-			var factory = new ProxyFactory();
-			var interceptor = new DefaultDynamicLazyFieldInterceptor();
-			return factory.CreateProxy(PersistentClass, interceptor, typeof(IFieldInterceptorAccessor));
+			var cacheEntry = new ProxyCacheEntry(IsClassProxy ? PersistentClass : typeof(object), System.Type.EmptyTypes);
```
fredericDelaporte: `IsClassProxy ? PersistentClass : typeof(object)` has to be reverted to `PersistentClass`: the field interception proxy does not handle being an interface proxy. (I will add a test, this is lacking.)

(Optionally, it may be attempted here to handle being an interface proxy. But the `INHibernateProxy` interface will have to be filtered out of `Interfaces`. And the support would only be partial, as there will be no way of handling interface methods, but only properties. So I do not think it should be done.)

src/NHibernate/Proxy/FieldInterceptorProxyBuilder.cs
```diff
@@ -0,0 +1,288 @@
...
+				typeof(IFieldInterceptorAccessor)
+			};
+
+			// Use the object as the base type
+			// since we're not inheriting from any class type
+			var parentType = baseType;
+			if (baseType.IsInterface)
+			{
+				parentType = typeof(object);
+				interfaces.Add(baseType);
```
fredericDelaporte: This does indeed not make sens for a field interceptor. The `baseType` can only be an interface for entities mapped as abstract. The field interceptor proxy should never be built for such case, it is built only on entity load from database. And anyway, it can not work in such case, since it uses its base implementation for storing its state.

fredericDelaporte:
I have addressed the two previous concerns, but I know look into changing even more the lazy property proxy implementation, for having it delegating its state like then lazy entity proxy. It would give two benefits: handling serialization the same way, and supporting the `proxy` mapping option, which currently is ignored by lazy property proxies.

fredericDelaporte:
>  I now look into changing even more the lazy property proxy implementation, for having it delegating its state like then lazy entity proxy.

This would require too much changes. Delegating its state cause issues to field setters and getters (from `IPropertyAccessor`), which ends up setting fields of the proxy instead of fields of the implementation. The entity proxy does not have the issue, because it is unwrapped and not directly supplied to accessors. This is handled in the persistence context, which makes a distinction between lazy entity proxies and entities. But there is no such thing with lazy properties proxies, which are handled as entities by the persistence context. Changing that seems to me a bit too much for the scope of this PR.

So I have squashed the fixups, and I consider this PR as ready for reviewing.

fredericDelaporte:
Checking again the dynamic proxy factory case, its serialization was ensuring the proxy type was rebuilt, which my implementation currently does not. Switching this PR to WIP again.

Testing this is a bit inconvenient as it requires to stop the process for unloading dynamic assemblies before deserializing. I intend to add some `Explicit` test allowing to check this manually.

fredericDelaporte:
Commits redone, for having a correct serialization from the first one. Serialization tests added.

Now this PR should really be ready.

fredericDelaporte:
About serialization, there is one "non-user-bug" case still unsupported: if the user supplies a serialization surrogate handling the entity type, it will be ignored by the proxy serialization. We do not have access to surrogates from the `GetObjectData` method. The only way to overcome this would be to copy the state of the proxy (excepted uninitialized properties) into a raw entity class instance and then to add it into `info`. Afterward copy it back to the proxy on deserialization (excepted uninitialized properties again). I do not consider it is worth the trouble.

Issue #1712, opened 2018-05-28
Support IEquatable<T> in LINQ provider


PinpointTownes:
Hey,

First, I'd like to say I'm very impressed by the effort you guys put to make NHibernate better than ever (.NET Standard 2.0 is just awesome).

I'm currently writing NHibernate-based stores for my [OpenIddict project](https://github.com/openiddict/openiddict-core) and so far, the experience has been extremely pleasant. I'm only blocked by the fact the default `EqualsGenerator` doesn't support the `IEquatable<T>` interface, that I use in my NHibernate LINQ queries.

Consider the following model and mapping ([a simplified copy of what's used in ASP.NET Core Identity 1.x/2.x, that uses exactly the same pattern](https://github.com/aspnet/Identity/blob/dev/src/Stores/IdentityUser.cs#L41)):

```csharp
public class MyIdentityUser<TKey> where TKey : IEquatable<TKey>
{
    public virtual TKey Id { get; set; }
}

public class MyIdentityUserMapping<TKey> : ClassMapping<MyIdentityUser<TKey>>
    where TKey : IEquatable<TKey>
{
    public MyIdentityUserMapping()
    {
        Id(user => user.Id);
    }
}
```

Thanks to the generic `TKey` identifier, you can write framework code that uses equality checks without forcing a specific type at compile time (e.g string, int, long, Guid, etc.):

```csharp
public Task<IdentityUser<TKey>> GetUserAsync(TKey id)
    where TKey : IEquatable<TKey>
{
    return (from user in _session.Query<IdentityUser<TKey>>()
            where user.Id.Equals(id)
            select user).FirstOrDefaultAsync();
}
```

Since `EqualsGenerator` doesn't support this interface, all you get when trying to execute this method is an exception:

```
{System.NotSupportedException: Boolean Equals(System.Guid)
   at NHibernate.Linq.Visitors.HqlGeneratorExpressionVisitor.VisitMethodCallExpression(MethodCallExpression expression)
   at NHibernate.Linq.Visitors.HqlGeneratorExpressionVisitor.VisitExpression(Expression expression)
   at NHibernate.Linq.Visitors.HqlGeneratorExpressionVisitor.Visit(Expression expression, VisitorParameters parameters)
   at NHibernate.Linq.Visitors.QueryModelVisitor.VisitWhereClause(WhereClause whereClause, QueryModel queryModel, Int32 index)
[...]

PR #1714, opened 2018-05-28
Support IEquatable<T> in the LINQ provider

PinpointTownes:
Fixes https://github.com/nhibernate/nhibernate-core/issues/1712.

/cc @fredericDelaporte


src/NHibernate.Test/NHSpecificTest/GH1712/Fixture.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+	[TestFixture]
+	public class Fixture : BugTestCase
+	{
+		protected override void OnSetUp()
+		{
+			using (var session = OpenSession())
+			using (var tx = session.BeginTransaction())
+			{
+				session.Save("GenericEntityWithGuid", new GenericEntity<Guid>(),
```
PinpointTownes: @fredericDelaporte let me know if you want me to add the other key types (bytes, numeric types, time/date types, etc.)

fredericDelaporte:
>> Since NHibernate handles entity equality and enum equality, I think supporting any type should be done.
>
>Only for limited types which are known to be translatable to SQL.

But equals is supported by entity types. (Enums do not implement `IEquatable<>` so they are indeed ruled out.) This choice forbid supporting `IEquatable<T>` for entity types, while there is no harm in trying to convert to HQL a not supported equals: that will just replace a LINQ failure by an HQL one.

src/NHibernate.Test/NHSpecificTest/GH1712/Fixture.cs
```diff
@@ -0,0 +1,55 @@
...
+
+		[Test]
+		public void PerformingAQueryUsingIEquatableShouldNotThrowEquals()
+		{
+			using (var session = Sfi.OpenSession())
+			{
+				GenericEntity<TId> Query<TId>(string name, TId id) where TId : IEquatable<TId>
+					=> (from e in session.Query<GenericEntity<TId>>(name)
+						where e.Id.Equals(id)
+						select e).FirstOrDefault();
```
fredericDelaporte: The async generator currently fails handling the local function conversion. Better use a regular class function.

@maca88, it sometime generates nothing, other times generates invalid code: the async counterpart is found in `NHibernate.Linq` namespace which is not used in the sync file, and the async file is lacking the `using`.

It is not lacking when a regular class function is used. The trouble occurs with 0.11.1 version too.

PinpointTownes:
> But equals is supported by entity types. (Enums do not implement IEquatable<> so they are indeed ruled out.) This choice forbid supporting IEquatable<T> for entity types, while there is no harm in trying to convert to HQL a not supported equals: that will just replace a LINQ failure by an HQL one.

I'm fine trying your option - even if it seems more complicated to implement. But I want to make sure you guys agree on what the solution should be (so I don't spend time implementing something that will be eventually discarded :sweat_smile:).

src/NHibernate/Linq/Functions/EqualsGenerator.cs
```diff
@@ -39,7 +39,25 @@ public EqualsGenerator()
...
+					ReflectHelper.GetMethodDefinition<IEquatable<uint>>(x => x.Equals(default(uint))),
+					ReflectHelper.GetMethodDefinition<IEquatable<long>>(x => x.Equals(default(long))),
+					ReflectHelper.GetMethodDefinition<IEquatable<ulong>>(x => x.Equals(default(ulong))),
+					ReflectHelper.GetMethodDefinition<IEquatable<float>>(x => x.Equals(default(float))),
+					ReflectHelper.GetMethodDefinition<IEquatable<double>>(x => x.Equals(default(double))),
+					ReflectHelper.GetMethodDefinition<IEquatable<decimal>>(x => x.Equals(default(decimal))),
+					ReflectHelper.GetMethodDefinition<IEquatable<Guid>>(x => x.Equals(default(Guid))),
+					ReflectHelper.GetMethodDefinition<IEquatable<DateTime>>(x => x.Equals(default(DateTime))),
+					ReflectHelper.GetMethodDefinition<IEquatable<DateTimeOffset>>(x => x.Equals(default(DateTimeOffset))),
+					ReflectHelper.GetMethodDefinition<IEquatable<bool>>(x => x.Equals(default(bool)))
```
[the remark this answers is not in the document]
PinpointTownes: I didn't add it because there was no entry for the "regular" equals in this list. I can certainly add it if you want.

fredericDelaporte:
I have made PinpointTownes/nhibernate-core#1, with commits addressing my review, but also demonstrating entity equality support.

fredericDelaporte:
I have also looked into:

> I am also thinking about whether it could be possible to still handle this case with a single addition to `SupportedMethods` or not. Like registering instead the method from the generic type definition, then adding after this [line](https://github.com/nhibernate/nhibernate-core/blob/5.1.2/src/NHibernate/Linq/Functions/DefaultLinqToHqlGeneratorsRegistry.cs#L84) a check on the `DeclaringType` for resolving the method on the generic type definition and checking again the dictionary.

And gave it up: resolving the method from the generic type definition in a robust manner does not look achievable, as there is no way to know which types from the original method should be switched for the declaring type generic type parameter. (By example, with `bool IEquatable<bool>.Equals(bool)`, the return type must stay `bool` while the argument type must be generic.)

Other subject, three tests were failing since the merging of my PR here. I had removed by mistake the equals generator registration. (I was willing to move it some lines above but finally decided to leave it in its current place, but forgot to cancel the cut.)

src/NHibernate.Test/NHSpecificTest/GH1712/Fixture.cs
maca88: This is really odd, I've run the generator 5 times (version `0.8.2.4`) on my computer and the local function was always converted correctly and the `using` statement was always there. Also, the `WhereEquatableEqualAsync` was never generated in those 5 times. The only thing that I found is that the namespace `NHibernate.Linq` was never added for which I filled an [issue](https://github.com/maca88/AsyncGenerator/issues/101).

I also tried with version `0.11.1` and I got the same results as with `0.8.2.4`.

src/NHibernate.Test/NHSpecificTest/GH1712/Entity.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace NHibernate.Test.NHSpecificTest.GH1712
+{
+	class GenericEntity<TId> : IEquatable<GenericEntity<TId>> where TId : IEquatable<TId>
+	{
+		public virtual TId Id { get; set; }
+		public virtual bool Equals(GenericEntity<TId> other)
+		{
+			throw new NotImplementedException();
```
hazzik: Doesn't this proof my point that we shall not try to add support for the generic `IEquatable<>`?

fredericDelaporte: I do not follow you. What proof do you see there? Is it the fact than someone may implement something else than the equality that will be applied on by NHibernate on the IDs, and so have a discrepancy between linq-to-object and linq-to-nhibernate? And then? That is also the case for `object.Equals`, that is also the case for the `==` operator. If that is your proof, then we should also remove support of `==` on entities in the Linq provider. And many more similar cases...

src/NHibernate.Test/NHSpecificTest/GH1712/Fixture.cs
maca88: Oh my bad, I don't know why I thought that the `using` was referred to the `using` statement used for the session, but after rereading your comment it is clear that it is referred to the `using` for the namespace. 
With the version `0.8.2.5` the missing `using` statement will now be generated.

src/NHibernate.Test/NHSpecificTest/GH1712/Entity.cs
hazzik: @fredericDelaporte what is the current behavior if there is no `IEquatable<GenericEntity<TId>>` eg if you are using `object.Equals(object)` method?

fredericDelaporte: Currently not supported, because, like for `TimeSpan`, this overload registration is lacking. Adding it (`ReflectHelper.GetMethodDefinition<object>(x => x.Equals(default(object))),`) will fix this. Indeed it should be added too.

hazzik: > Currently not supported

No, it's actually supported. Re-Linq/Linq handles this and converts to Expression.Eqaul

src/NHibernate.Test/NHSpecificTest/GH1712/Fixture.cs
hazzik: @maca88 can you push missing tags to the repo?

src/NHibernate/Linq/Functions/EqualsGenerator.cs
```diff
@@ -42,6 +42,9 @@ public EqualsGenerator()
 					ReflectHelper.GetMethodDefinition<TimeSpan>(x => x.Equals(x)),
 					ReflectHelper.GetMethodDefinition<bool>(x => x.Equals(default(bool))),
 
+					ReflectHelper.GetMethodDefinition<object>(x => x.Equals(default(object))),
```
hazzik: This is already handled by Linq itself.

hazzik: I've double-checked and it turns it does not work.

hazzik:
@fredericDelaporte I'm strongly against adding the runtime generator for `IEquatable<T>`, and  `object.Equals(object)`/`object.Equals(object, object)`.

The reason for that is because it brings the discrepancy to the core methods and potentially unexpected results. Also, as this is easily extensible by the users, so the people willing to have this behaviour could implement it themselves. But I believe this shall not be in the core.

src/NHibernate/Linq/Functions/EqualsGenerator.cs
fredericDelaporte: Remove it, and `QueryWithDefaultEqualsShouldNotThrow` fails [here](https://github.com/nhibernate/nhibernate-core/blob/5.1.2/src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs#L514). So it does appears to be translated to another expression.

fredericDelaporte:
I still disagree with your point of view, but I do not consider these cases support to be worth debating more. I am going to revert them.

src/NHibernate.Test/NHSpecificTest/GH1712/Fixture.cs
maca88: I've added the missing tags and pushed them.

PinpointTownes:
Thanks for merging this PR. Do you have, by any chance, a rough estimate of the 5.2 release date?

fredericDelaporte:
Not really. By example for 5.1, the fact .Net Core support was added in it has pushed for finishing some other important things then releasing it. For 5.2, we do not have currently anything comparable, and 5.1 is still relatively recent, 5.2 does not contain yet many changes. There was 6 months between 5.0 and 5.1, maybe we will settle on this release pace.

PR #1716, opened 2018-05-30
Release 5.1.3

fredericDelaporte:
Preparing the release.

I intend to wait until 5.1.2 is at least one month old (so until June, 10th), or until we have some critical fix to do.


bahusoid:
Should we also include https://github.com/nhibernate/nhibernate-core/pull/1706? It's a 5.1 bug for #1471. If yes - how? via separate PR?

PR #1718, opened 2018-05-31
Combine Future queries for ICriteria and IQuery

### LINKED ISSUE #752: NH-3541 - Future queries of Criteria API/QueryOver are batched separately from other query methods
**Massimiliano Alberti** created an issue — <time datetime="2013-09-28T08:09:10.819Z" title="2013-09-28T08:09:10.819Z">28th September 2013, 8:09:10</time>:
> I was interested in batching together (through the use of .Future<>()) a CreateSQLCriteria (so a native SQL query) with a QueryOver query, but I've discovered that the two are batched separately (so two distinct commands are sent to the SQL). I've done some tests and it seems there are two queues:
> 
> * HQL, Native SQL, LINQ queue
> * Criteria API/QueryOver queue
> 
> I've built an nunit to show it (I have done all the 10 possible X-Y tests, I trust that the X-X tests (HQL+HQL) will work, but I've only tested .Future(), not .FutureValue(), and without parameters in the query)
>

### LINKED ISSUE #865: NH-2428 - Session.MultiCriteria and FlushMode.Auto inside transaction
**Carlos Alberto Costa Beppler** created an [issue](http://nhibernate.jira.com/browse/NH-2428) — <time datetime="2010-11-25T09:25:57.865Z" title="2010-11-25T09:25:57.865Z">25th November 2010, 9:25:57</time>:
> When using Session.CreateMultiCriteria to execute batches of queries on an session witch FlushMode is Auto, the session is not flushing the pending changes to database.
> 
> This causes the same problem when using Future and FutureValue when used with ICriteria or IQueryOver because they use the MultiCriteria to do their jobs.
> 
> The problem can be reproduced with the following test case that inserts some entities an then try to query them using those APIs:
> ```csharp
>     public void OutrosSalvarRecuperarNaMesmaSessao() {
>       var original = new List<Aplicativo>() {
>         new Aplicativo("Meu Aplicativo 1", "Aplicativo de Teste 1"),
>         new Aplicativo("Meu Aplicativo 2", "Aplicativo de Teste 2"),
>         new Aplicativo("Meu Aplicativo 3", "Aplicativo de Teste 3"),
>         new Aplicativo("Meu Aplicativo 4", "Aplicativo de Teste 4")
>       };
>       using (var session = SessionFactory.OpenSession()) {
>         session.FlushMode = FlushMode.Auto;
>         using (var transaction = session.BeginTransaction(IsolationLevel.ReadCommitted)) {
>           foreach (var item in original)
>             session.Save(item);
> 
>           // works
>           //var query = QueryOver.Of<Aplicativo>().WhereRestrictionOn(a => a.Nome).IsIn(original.Select(o => o.Nome).ToArray());
>           //var fromDb = query.GetExecutableQueryOver(session).List();
> 
>           // error => returns 0 items 
>           var query = QueryOver.Of<Aplicativo>().WhereRestrictionOn(a => a.Nome).IsIn(original.Select(o => o.Nome).ToArray());
>           var fromDb = query.GetExecutableQueryOver(session).Future().ToList();
> 
>           // works
>           //var query = DetachedCriteria.For<Aplicativo>().Add(Restrictions.InG("Nome", original.Select(o => o.Nome).ToArray()));
>           //var fromDb = query.GetExecutableCriteria(session).List<Aplicativo>();
> 
>           // error => returns 0 items 
>           //var query = DetachedCriteria.For<Aplicativo>().Add(Restrictions.InG("Nome", original.Select(o => o.Nome).ToArray()));
>           //var fromDb = query.GetExecutableCriteria(session).Future<Aplicativo>().ToList();
> 
>           // works
>           //var query = session.CreateQuery("from Aplicativo a where a.Nome in (:nomes)").SetParameterList("nomes", original.Select(o => o.Nome).ToArray());
>           //var fromDb = query.List<Aplicativo>();
> 
>           // works
>           //var query = session.CreateQuery("from Aplicativo a where a.Nome in (:nomes)").SetParameterList("nomes", original.Select(o => o.Nome).ToArray());
>           //var fromDb = query.Future<Aplicativo>().ToList();
> 
>           Assert.AreEqual(original.Count, fromDb.Count);
>         }
>       }
> ```
> 
> The thests are been run using .NET Framework 4 and SQLite data provider.
>

### LINKED ISSUE #1293: NH-3350 - Duplicate records using Future()
**Boris Drajer** created an [issue](https://nhibernate.jira.com/browse/NH-3350) — <time datetime="2012-12-08T12:31:52.180Z" title="2012-12-08T12:31:52.180Z">8th December 2012, 12:31:52</time>:
> .Net 3.5 SP1, SQL Server 2005 dialect, with 3.3.2 GA source
> 
> Join-fetching a collection using Future() produces duplicate records if the collection is affected by multiple queries. The problem seems to be that all multiple queries are executed at once and the multi query just keeps adding all the records from different queries into the same collection.
> 
> For example, if a statement like this -
> 
> session.CreateQuery("from Person p left join fetch p.Children where pp.Id = 1").Future<Person>();
> 
> - is executed three times, after the query is run the p.Children collection will contain triple the number of child records. Using List() instead of Future() loads the child records properly.
> 
> The problem seems to depend on the child record having a many-to-one relation: remove this and everything seems to work correctly.
> 
> I've attached a test case: add these two files to the NHibernate.Test\NHSpecificTest\Futures folder and run the supplied test to see the issue.
>

bahusoid:
The intention is to implement batcher that can contain both `ICriteria` and `IQuery` based queries. And make Future implementation based on this batcher. See examples in [tests](https://github.com/bahusoid/nhibernate-core/blob/futureMerge2/src/NHibernate.Test/Futures/QueryBatchFixture.cs#L20-L73)

New batcher interface [IQueryBatch](https://github.com/bahusoid/nhibernate-core/blob/futureMerge2/src/NHibernate/Multi/IQueryBatch.cs) (implementation [QueryBatch](https://github.com/bahusoid/nhibernate-core/blob/futureMerge2/src/NHibernate/Multi/QueryBatch.cs) ) allows only to Add queries  wrapped in to [IQueryBatchItem](https://github.com/bahusoid/nhibernate-core/blob/futureMerge2/src/NHibernate/Multi/IQueryBatchItem.cs) interface to batch and Execute batch.

See implementation for [Criteria](https://github.com/bahusoid/nhibernate-core/blob/futureMerge2/src/NHibernate/Multi/CriteriaBatchItem.cs),  [IQuery ](https://github.com/bahusoid/nhibernate-core/blob/futureMerge2/src/NHibernate/Multi/QueryBatchItem.cs) and [Linq](https://github.com/bahusoid/nhibernate-core/blob/futureMerge2/src/NHibernate/Multi/LinqBatchItem.cs). Most of functionality is shared in [QueryBatchItemBase](https://github.com/bahusoid/nhibernate-core/blob/futureMerge2/src/NHibernate/Multi/QueryBatchItemBase.cs) .

Caching is implemented per query (it's up to IQueryBatchItem implementation to cache or not to cache). So it's possible to cache queries with different settings (cache region etc), combine cached and non-cached queries. And batcher doesn't need to know all internal details about query cache (so it might be different cache for different IQueryBatchItem implementations).

Fixes #752

As "side effects" of this re-implementation of futures:
- Supports stateless session
- Fixes #865
- Fixes #1293
- Fixes partially #840 (multi-criteria/query not fixed)
- Fixes partially #1344 (multi-criteria not fixed)


hazzik:
> I want you to confirm that you are OK in principle.

I'm OK.

fredericDelaporte:
I am also ok on the principle, provided this is not overly complex to achieve. (I have not looked yet at how you do that.)

fredericDelaporte:
> **Help** I didn't check in async code as for some reasons async generator now generates async methods for parts that I didn't change. For instance it now generates `ICriteria.FutureAsync` and `ICriteria.FutureValueAsync`.  Can you point me what might be the problem?

The "trouble" lies in having changed these parts actually. `ICriteria.Future` and `FutureValue` have both their `CriteriaImpl` implementation calling a new method, `session.GetFutureMultiBatch().AddAsValue<T>(this)` (or `AddAsEnumerable`), which implementations at some point call something having an async counterpart (`GetTranslators` in that case). So async methods are generated all the way up.

`GetTranslators` is async because it triggers an auto-flush. Obviously this method was not meant to be called for creating non-resolved futures, since auto-flushing is a non-sens if the user has the opportunity to change entity states before the queries actually executes. Auto-flushing must be handled at a point where the user will not have the opportunity to change any entity state before the query actually executes itself.

bahusoid:
@fredericDelaporte Thanks for clarification. Actually `GetTranslators` quite unexpected place for flushing.

bahusoid:
@hazzik , @fredericDelaporte  Could you please review my changes in current state.

I got rid of auto-flushing in `AbstractQueryImpl.GetTranslators`. Really strange and unexpected place - retrieving query metadata should not lead to flush. And why to do check for each query in batch - it should be done once per batch execution. I also fixed https://github.com/nhibernate/nhibernate-core/issues/865

I've enabled batcher for stateless session. Seems to be working [fine](https://github.com/nhibernate/nhibernate-core/pull/1718/files#diff-996208dcb9fb6a7d88cb2f2c2377708aR169).

Should I obsolete existing future [batcher ](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Impl/FutureBatch.cs) with all related classes?

hazzik:
@bahusoid I just wanted to suggest to make additions as a separate commits so it would be easier to see what's changed.

fredericDelaporte:
> I got rid of auto-flushing in `AbstractQueryImpl.GetTranslators`. Really strange and unexpected place - retrieving query metadata should not lead to flush. And why to do check for each query in batch - it should be done once per batch execution. I also fixed #865

A separated commit for separated subjects would be nice too. Moving the auto-flush out of translators generation could be done as a prerequisite, eventually merged as a separated PR. If the #865 fix is a consequence of this move, or a trivial additional change requiring this move, it is fine to have it in. Otherwise better also have a separated commit (or even PR if it is an independent change).

Enabling batcher for stateless session would also better have its own commit. (And if it does not depend on the new batcher, it is better to have it in a separated PR.)

> Should I obsolete existing future [batcher ](https://github.com/nhibernate/nhibernate-core/blob/5.1.2/src/NHibernate/Impl/FutureBatch.cs) with all related classes?

I do not see much value in having the choice between separated future batches per query APIs and unified one, so I am for obsoleting the previous batcher and related classes, and removing the `IsUnifiedFuture` setting.

fredericDelaporte:
> I do not see much value in having the choice between separated future batches per query APIs and unified one, so I am for obsoleting the previous batcher and related classes, and removing the `IsUnifiedFuture` setting.

In fact `IsUnifiedFuture` setting has to be kept but maybe in a different form. Since `GetFutureMultiBatch()` will throw if the session is not of the expected type, it is a breaking change. So the `IsUnifiedFuture` should be replaced by some probing of the session capabilities.

bahusoid:
>In fact IsUnifiedFuture setting has to be kept

Actually I don't think we need it.

We expose old future as:
https://github.com/nhibernate/nhibernate-core/blob/2345bb47aa0f6a19ccd56230b284abed6002c9af/src/NHibernate/Engine/ISessionImplementor.cs#L329-L331

Where both `FutureCriteriaBatch` and `FutureQueryBatch` hard-wired to `SessionImpl' class via protected base ctor:
https://github.com/nhibernate/nhibernate-core/blob/2345bb47aa0f6a19ccd56230b284abed6002c9af/src/NHibernate/Impl/FutureBatch.cs#L26-L29

And that's why old future is disabled for stateless session.

So we shouldn't bother to fallback to old code - old code can work only with `SessionImpl`.

bahusoid:
> I just wanted to suggest to make additions as a separate commits 
>A separated commit for separated subjects would be nice too

Ok. I will rewrite history in few commits. All changes are specific for this feature (and multi criteria fix is trivial after required for this change refactoring).


>Enabling batcher for stateless session would also better have its own commit.

As already said in previous comment - it's supported only by new batcher.

src/NHibernate/Multi/IMultiAnyQuery.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Data.Common;
+using NHibernate.Engine;
+using NHibernate.SqlCommand;
+
+namespace NHibernate
+{
+	/// <summary>
+	/// Interface for wrapping query to be batchable by <see cref="IMultiAnyQueryBatch"/>
+	/// </summary>
+	public interface IMultiAnyQuery<TResult> : IMultiAnyQuery
```
fredericDelaporte: Maybe rename it `IAnyQuery`. It is not itself multiple if I got it right. It is just there to supports query batch.

src/NHibernate/Multi/IMultiAnyQueryBatch.cs
```diff
@@ -0,0 +1,24 @@
+namespace NHibernate
+{
+	/// <summary>
+	/// Universal query batcher
+	/// </summary>
+	public partial interface IMultiAnyQueryBatch
```
fredericDelaporte: Naming should be simplified IMO, `IAnyQueryBatch` should be enough. Well, even `IQueryBatch` should be enough indeed, provided the previous batchers are obsoleted.

src/NHibernate/Multi/MultiAnyCriteriaQuery.cs
```diff
@@ -0,0 +1,55 @@
+using System.Collections.Generic;
+using NHibernate.Impl;
+using NHibernate.Loader.Criteria;
+using NHibernate.Persister.Entity;
+
+namespace NHibernate
+{
+	public partial class MultiAnyCriteriaQuery<T> : MultiAnyQueryBase<T>, IMultiAnyQuery<T>
```
fredericDelaporte: As previously said, `Multi` should be removed IMO.

src/NHibernate/Multi/MultiAnyLinqQuery.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using NHibernate.Linq;
+using NHibernate.Util;
+using Remotion.Linq.Parsing.ExpressionVisitors;
+
+namespace NHibernate
+{
+	public partial class MultiAnyLinqQuery<T> : MultiAnyQuery<T>
```
fredericDelaporte: Same as previous comment.

src/NHibernate/Multi/MultiAnyLinqQuery.cs
```diff
@@ -0,0 +1,110 @@
...
+	{
+		private static Delegate _postExecuteTransformer;
+		private static NhLinqExpression _linqEx;
+
+		public MultiAnyLinqQuery(IQueryable query):base(GetQuery(query))
```
fredericDelaporte: Non standard formatting, I guess a reformat would alter this file. Better do it before merging.

src/NHibernate/Multi/MultiAnyCriteriaQuery.cs
fredericDelaporte: Why `IMultiAnyQuery<T>` again, since it is already brought by `MultiAnyQueryBase<T>`?

src/NHibernate/Multi/MultiAnyQueryBase.cs
```diff
@@ -0,0 +1,217 @@
...
+using NHibernate.SqlCommand;
+using NHibernate.Transform;
+using NHibernate.Util;
+
+namespace NHibernate
+{
+	/// <summary>
+	/// Base class for both ICriteria and IQuery queries
+	/// </summary>
+	public abstract partial class MultiAnyQueryBase<TResult> : IMultiAnyQuery, IMultiAnyQuery<TResult>
```
fredericDelaporte: `IMultiAnyQuery` already comes with `IMultiAnyQuery<TResult>`, reduce redundancy. An same remark than previously for naming.

src/NHibernate/Multi/MultiAnyQuery.cs
```diff
@@ -0,0 +1,42 @@
+using System.Collections.Generic;
+using System.Linq;
+using NHibernate.Engine;
+using NHibernate.Impl;
+
+namespace NHibernate
+{
+	public partial class MultiAnyQuery<TResult> : MultiAnyQueryBase<TResult>, IMultiAnyQuery<TResult>
```
fredericDelaporte: Same remarks as previously on naming and interfaces.

src/NHibernate/Multi/MultiAnyQueryBase.cs
```diff
@@ -0,0 +1,217 @@
...
+		public virtual void Init(ISessionImplementor session)
+		{
+			Session = session;
+
+			_queryInfos = GetQueryLoadInfo();
+
+			var count = _queryInfos.Count;
+			NewArray(count, out _hydratedObjects);
+			NewArray(count, out _subselectResultKeys);
+			NewArray(count, out _loaderResults);
```
fredericDelaporte: Why not just writing regular code here?

```c#
_hydratedObjects = new List<object>[count];
_subselectResultKeys = new List<EntityKey[]>[count];
_loaderResults = new IList[count];
```

Adding a general purpose `NewArray` in that class just for this need here looks as a code smell to me. Either remove it or put that as an internal helper on a class like `ArrayHelper`.

src/NHibernate/Multi/MultiAnyQueryBase.cs
```diff
@@ -0,0 +1,217 @@
...
+							_subselectResultKeys[index].Add(keys);
+							keys = new EntityKey[entitySpan]; //can't reuse in this case
+						}
+
+						tmpResults.Add(o);
+					}
+					return rowCount;
+				};
+
+				_loaderResults[index] = tmpResults;
```
fredericDelaporte: Putting this after the yield leaves the state a bit inconsistent between each `MoveNext` when enumerating. Better move that before the yield.

src/NHibernate/Multi/MultiAnyQueryBatch.cs
```diff
@@ -0,0 +1,146 @@
...
+using NHibernate.Engine;
+using NHibernate.Exceptions;
+using NHibernate.Impl;
+
+namespace NHibernate
+{
+	/// <summary>
+	/// Universal query batcher
+	/// </summary>
+	public partial class MultiAnyQueryBatch : IMultiAnyQueryBatch
```
fredericDelaporte: Again, same remark for naming than on the interface.

bahusoid:
I've split commit with some squashed clean up in tests.

Regarding naming for `IMultiAnyQueryBatch', 'IMultiAnyQuery' and related classes. We should choose new best pair :)
'IQueryBatch' and `IAnyQuery'
or
'IAnyQueryBatch' and `IAnyQuery'
or
`IQueryBatch` and `IQueryBatchItem`

hazzik:
+ 3️⃣

src/NHibernate/Linq/DefaultQueryProvider.cs
```diff
@@ -15,13 +15,20 @@ namespace NHibernate.Linq
...
 		IFutureValue<TResult> ExecuteFutureValue<TResult>(Expression expression);
 		void SetResultTransformerAndAdditionalCriteria(IQuery query, NhLinqExpression nhExpression, IDictionary<string, Tuple<object, IType>> parameters);
 		int ExecuteDml<T>(QueryMode queryMode, Expression expression);
 		Task<TResult> ExecuteAsync<TResult>(Expression expression, CancellationToken cancellationToken);
 	}
 
+	public interface INhQueryProviderSupportMultiBatch
+	{
+		IQuery GetPreparedQuery(Expression exrpession, out NhLinqExpression expression);
+		IMultiAnyQueryBatch GetFutureMultiBatch();
```
bahusoid: I'm thinking to expose `ISessionImplementor Session` instead. And get batch from session. Any objections?

src/NHibernate/Async/Impl/AbstractQueryImpl.cs
```diff
@@ -56,7 +56,5 @@ public abstract partial class AbstractQueryImpl : IQuery
 		}
 
 		#endregion
-
-		protected internal abstract Task<IEnumerable<ITranslator>> GetTranslatorsAsync(ISessionImplementor sessionImplementor, QueryParameters queryParameters, CancellationToken cancellationToken);
```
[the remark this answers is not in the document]
bahusoid: Hmm... Adding to `AbstractQueryImpl`
```C#
// Since v5.2
[Obsolete("This method has no usages and will be removed in a future version")]
protected internal abstract Task<IEnumerable<ITranslator>> GetTranslatorsAsync(ISessionImplementor sessionImplementor, QueryParameters queryParameters, CancellationToken cancellationToken);
```
Leads async generator to generate non obsolete overrides in  `AbstractQueryImpl2` and `SqlQueryImpl` (even if manual obsolete override is added)
Should it be considered as async generator bug? Or did I miss something?

fredericDelaporte:
>  I also fixed #865

But it seems there is not test for it. Could some test be added?

fredericDelaporte:
From #1733:

> MultiCriteria and MultiQuery are separate batchers. If you need MulitCriteria or MultiQuery - you still need to fix it.

Will it be a big change to make them use the same query batcher than futures, allowing this way to only keep one non-obsoleted batcher? And it would also reduce the amount of code in which we have bugs to fix!

bahusoid:
>Will it be a big change to make them use the same query batcher than futures, allowing this way to only keep one non-obsoleted batcher? And it would also reduce the amount of code in which we have bugs to fix!

I was going to look in to it when new batcher is ready. But I suspect - yes. Are you OK with using new batcher cache approach in MultiCriteria\MultiQuery?

fredericDelaporte:
Better have a single way of tackling the same subject everywhere, but let see that in another PR.

src/NHibernate/Linq/LinqExtensionMethods.cs
```diff
@@ -2419,14 +2425,15 @@ public static IFutureEnumerable<TSource> ToFuture<TSource>(this IQueryable<TSour
 		/// <exception cref="T:System.NotSupportedException"><paramref name="source" /> <see cref="IQueryable.Provider"/> is not a <see cref="INhQueryProvider"/>.</exception>
 		public static IFutureValue<TSource> ToFutureValue<TSource>(this IQueryable<TSource> source)
 		{
-			if (FutureSettings.IsUnifiedFuture)
+			if(source.Provider is INhQueryProviderSupportMultiBatch batchProvider)
```
fredericDelaporte: Missing space after if.

src/NHibernate/Linq/LinqExtensionMethods.cs
```diff
@@ -2404,8 +2404,14 @@ public static Task<List<TSource>> ToListAsync<TSource>(this IQueryable<TSource>
 		/// <exception cref="T:System.NotSupportedException"><paramref name="source" /> <see cref="IQueryable.Provider"/> is not a <see cref="INhQueryProvider"/>.</exception>
 		public static IFutureEnumerable<TSource> ToFuture<TSource>(this IQueryable<TSource> source)
 		{
+			if(source.Provider is INhQueryProviderSupportMultiBatch batchProvider)
```
fredericDelaporte: Missing space after if.

src/NHibernate/Linq/DefaultQueryProvider.cs
fredericDelaporte: I do not really see a benefit in exposing the whole session from the Linq provider. I would rather keep just exposing the future batch. Or even cease exposing it and add methods for adding new Linq query to it, a bit like the old interface was doing.

fredericDelaporte:
> Regarding naming for `IMultiAnyQueryBatch`, `IMultiAnyQuery` and related classes. We should choose new best pair :)
>1. `IQueryBatch` and `IAnyQuery`
>2. `IAnyQueryBatch` and `IAnyQuery`
>3. `IQueryBatch` and `IQueryBatchItem`
>4. Your suggestion

The third option looks better to me. `IAnyQuery` is tightly bound to the query batch but its naming is not conveying this fact.

src/NHibernate/Linq/DefaultQueryProvider.cs
bahusoid: The benefit is to make further modifications simpler. If session was available and old future was using it - we wouldn't need to introduce new interface. It was only internal modifications in extension methods. Is Session really internal implementation detail for provider? I guess it can be abstracted one more layer down - but it looks to me as very rare breaking change (and such modification will be breaking in many more places).

fredericDelaporte: Ok, go for it.

src/NHibernate/Async/Impl/AbstractQueryImpl.cs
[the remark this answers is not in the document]
fredericDelaporte: For this case I think adding back manually the method in `AbstractQueryImpl2` non generated partial is the way to go.

bahusoid:
Made a look at existing Future issues:

1. https://github.com/nhibernate/nhibernate-core/issues/1293  - NH-3350 - Duplicate records using Future()
To fix it `Loader.InitializeEntitiesAndCollections` needs to be called right after query is loaded.
Going to fix it in new batcher.

2. https://github.com/nhibernate/nhibernate-core/issues/840 - NH-3835 - Future/MultiCriteria 2nd level caching
Fixed for Future.

3. https://github.com/nhibernate/nhibernate-core/issues/1290 - NH-3334 - 'collection is not associated with any session' upon refreshing objects from QueryOver<>().Future<>()
To fix it `Session.PersistenceContext.InitializeNonLazyCollections();` needs to be called (in QueryBatchItemBase.PostProcess or maybe once per batch)
Not sure if it's a safe change. As before - all eager mappings were loaded lazly by Future. Wasn't it intentional behavior?

4. https://github.com/nhibernate/nhibernate-core/issues/1249 - NH-2901 - Multiple .Fetch() calls within Future query results in collection that does not persist deletes 
Unclear ticket. He either tries to fetch multiple bags in one query which throws exception (maybe missing delete due to wrong exception handling?)
Or tries to fetch not mapped collection - which is no op and doesn't affect behavior in any way (double checked it).

5. https://github.com/nhibernate/nhibernate-core/issues/752 - NH-3541 - Multiple Queries of Criteria API/QueryOver are batched separately from other query methods
Fixed

fredericDelaporte:
We need to avoid broadening the scope of this PR. Trivial additional fixes could be ok, But for #1290 at least, I think it needs to be considered separately.

src/NHibernate/Async/Impl/AbstractQueryImpl.cs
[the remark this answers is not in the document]
bahusoid: No. It's not recognized by async generator and duplicate method is added.

bahusoid:
@fredericDelaporte  It seems I've addressed all the fixable requested changes.

Pending issues that  I can't fix now:
Add back obsolete GetTranslators (https://github.com/nhibernate/nhibernate-core/pull/1718/files#r192739208)

Let me know if I miss anything.

I also noticed that there is `Loader.PutResultInQueryCacheAsync` which now not used anywhere. It seems cache handling code needs to be rewritten somehow.

src/NHibernate/Loader/Loader.cs
```diff
@@ -1638,48 +1638,77 @@ protected IList List(ISessionImplementor session, QueryParameters queryParameter
...
 		}
 
-		private IList ListUsingQueryCache(ISessionImplementor session, QueryParameters queryParameters, ISet<string> querySpaces)
+		protected internal IList GetResultsIfCacheable(
+			ISessionImplementor session,
+			QueryParameters queryParameters,
+			out IQueryCache queryCache,
+			out QueryKey key,
+			ISet<string> querySpaces,
+			Func<IResultTransformer, IList> loadListFunc)
```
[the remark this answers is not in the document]
bahusoid: Take a look at my latest refactoring:
https://github.com/nhibernate/nhibernate-core/pull/1718/commits/a2a50078f9fdd74bcd05d4bcb0e26078fa65d4f9

Yeah such signature can't be used in async code - I got rid of it.

src/NHibernate/Multi/FutureExtensions.cs
```diff
@@ -0,0 +1,272 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NHibernate
+{
+	public static partial class FutureExtensions
```
fredericDelaporte: In fact, these are `IQueryBatch` extensions. If `IQueryBatch` should be usable as a replacement for multi-queries and multi-criteria, these extensions only meant for futures will be quite polluting. Do we need them to be public and/or on `IQueryBatch`?

For me, the "futures yielding" extensions here only have internal usages.

I think public extensions here should instead yield back the `batch` for allowing chaining in user code while hiding the `IQueryBatchItem` encapsulation required for adding a query to the batch.

The "futures yielding" extensions should be made internal or moved to a `FutureHelper` class not extending `IQueryBatch` in order to avoid bloating its interface.

src/NHibernate.Test/Futures/QuerytBatchFixture.cs
```diff
@@ -0,0 +1,466 @@
...
+		}
+
+		[Test]
+		public void OnAfterLoad()
+		{
+			using (var session = OpenSession())
+			{
+				var batch = NewBatch(session);
+				IList<EntityComplex> results = null;
+				batch.AddOnAfterLoad(session.Query<EntityComplex>().WithOptions(o => o.SetCacheable(true)), list => results = list);
```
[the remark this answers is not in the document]
bahusoid: >Or do you consider it could already be directly used as a multi-criteria/multi-query alternative

Yes I think it should be available as separate batcher. I see it as `IFuture...` interfaces as a modern NHibernate suggested approach for operating with batch results. So those extensions are not hard-wired to `session.FutureBatch`. It's also a suggested methods to use with directly created batches -  It's better to have the same approach everywhere. But if user has some custom batch implementations and so he doesn't use IFuture interfaces - I've added OnAfterLoad event to simplify integration.

I'm not sure if we can simply obsolete old multi interfaces and implementations - it could require a considerable amount of work to rewrite existing code to use new batcher. As quite different approach is used for retrieving results. So it might be a showstopper to upgrade for many projects.

fredericDelaporte:
About commits of this PR, there is two competing needs:

1. Once reviewing has started, updates are better made as additional commits, in order to ease the re-reviewing process.
2. A PR which happens to fix more than one issue should have, if possible, one commit per issue.

The first point is easy and does not hinder merging, provided we can just "squash and merge", process during which the commit message can be rewritten properly. (About how we merge, see this [comment](https://github.com/nhibernate/nhibernate-core/pull/582#issuecomment-290379985).)

The second point causes more work, and hinder merging by removing the possibility of just "squash-merging" all commits. It forces to do a final rebase for cleaning up the commit list and allowing merging them without squashing. That is why the general rule is to avoid having broad PRs, but instead target at fixing only one issue by PR.
For easing combining 1. with 2., put as first message line of a commit to be squashed with another the `fixup! ` prefix followed by the first line of the target commit message. On the next lines you can put its actual message. When doing a `git rebase -i ...`, such fixup-prefixed commits will be automatically moved at the right position as fixups.

About commit messages, from [here](https://github.com/nhibernate/NHibernate-Caches/pull/26#issuecomment-374751162):

> Could you please follow [this convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153)_*_ and this [guidelines](https://chris.beams.io/posts/git-commit/) when writing commit messages? 
>
>What it means:
>
>```txt
>Add a .Net Core memory cache #no full stop at the end
>
>* Upgrade log4net to 2.0.8 in all projects
>* Upgrade AsyncGenerator and regenerate all projects
>* Handle compilation with multiple target frameworks
>* Handle Core tests
>* Upgrade async generator
>
>Fixes #25 #this goes to the footer
>```
>---
>_*_ except scopes - we do not need them.

src/NHibernate.Test/Futures/QuerytBatchFixture.cs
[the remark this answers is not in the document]
fredericDelaporte: Keeping two different implementations for achieving the same functionality is a burden, and it may also confuse users about what they should use.

From my point of view, futures are an additional abstraction layer on top of multi-queries. It makes sens to let users use directly multi-queries/query-batch if they wish too.

But having two implementations of multi-queries/query-batch is not a thing which should be kept in my opinion.

> As quite different approach is used for retrieving results.

That is a point to fix before obsoleting the previous implementation.

But since this PR is already quite broad, we would have better to handle obsoleting the previous implementation in a dedicated PR.

bahusoid:
I'm not sure I'm following what should be done regarding commits.
Do you want me to squash all unimportant commits to "feature" related commits?
Or do you want me to squash all current commits in a single commit with list of tasks I've done?

I can squash all WIP commits and make history look something like:
1. Implement universal query batch

2. Integrate  universal  query batch in Future
Fixes #752
Fixes #1293
Fixes partially #840 (multi-criteria/query not fixed)
Fixes partially #1344 (multi-criteria not fixed)

3. Fix NH-2428 - Session.MultiCriteria and FlushMode.Auto inside transaction
Fixes #865

4. Tests for NH-3864 - Cachable Criteria with aliased join and NH-3334 - 'collection is not associated with any session' upon refreshing objects from QueryOver<>().Future<>()

Or did you have something else in mind?

fredericDelaporte:
If @hazzik is done with reviewing the latest commit, your proposition sounds good to me.

I would reword 3. as

> Fix auto-flush for multi queries and multi criteria
>
> Fixes #865 

Putting the Jira number was an old practice for Jira integration, we do not need it anymore.

And 4.:

> Add tests for future with aliased join and mapped eager fetch
>
> See #1344 & #1290

hazzik:
I did not check anything yet. So, as you wish.

bahusoid:
Before squashing...
1) Should I keep all files in Multi folder? Maybe I should move them to folder with old implementation (Impl folder for classes and root for interfaces)

2) Should I internalize all query batch classes (QueryBatch, CriteriaBatchItem, ... ) ?

3) Maybe you have in mind some better naming to suggest for `IQueryBatchItem` interface methods:
`PostProcess` -> `OnAfterBatchLoad`?
`ExecuteNonBatchable` -> ?
`GetProcessResultSetActions` -> ?


P.S. Another unrelated to this PR but related to Future issue. I've just noticed that delayed `IEnumerable` interface for `IFutureEnumerable` is obsoleted and suggested replacement is to use "non delayed" GetEnumerable call:
 https://github.com/nhibernate/nhibernate-core/blob/dc71c7df5c22d82991bc1b009ecdfe2ec9902b45/src/NHibernate/IFutureEnumerable.cs#L10-L12

IMHO from this description it's not clear that suggested refactoring changes behavior:
```C#
public IEnumerable<TBase> GetLazyList()
{
return session.Query<SubParent1>().Future<TBase>().Concat(session.Query<SubParent2>().Future())
}
```
So lazy method above become non lazy after suggested refactoring. And the query above can't be re-written in lazy way without wrapping this code in some custom user lazy enumerator.
So I think that we still should provide a way for retrieving lazy enumerable.

fredericDelaporte:
> delayed `IEnumerable` interface for `IFutureEnumerable` is obsoleted and suggested replacement is to use "non delayed" GetEnumerable call

Discuss this on #1665.

fredericDelaporte:
>Before squashing...
>1) Should I keep all files in Multi folder? Maybe I should move them to folder with old implementation (Impl folder for classes and root for interfaces)

It seems fine to me to keep them in this `Multi` folder.

>2) Should I internalize all query batch classes (QueryBatch, CriteriaBatchItem, ... ) ?

No, currently only utility classes are internalized, for ceasing providing utilities while that is not the purpose of NHibernate. I do not think this should be done for implementation classes.

>3) Maybe you have in mind some better naming to suggest for `IQueryBatchItem` interface methods:
`PostProcess` -> `OnAfterBatchLoad`?
`ExecuteNonBatchable` -> ?
`GetProcessResultSetActions` -> ?

As somewhat said on the extension class [here](https://github.com/nhibernate/nhibernate-core/pull/1718#discussion_r193510649), I am not currently convinced by the API shape. Using directly the query batcher looks cumbersome. How would look [these examples](http://nhibernate.info/doc/nhibernate-reference/performance.html#performance-multi-query) if they were rewritten with the query batch? I think you could find better namings by looking at that. And it should avoid exposing the future specific extension methods when manipulating a query batch.

bahusoid:
Those methods are not to be used by user. They are meant only for batch execution.

Well the following query:
```C#
IMultiQuery multiQuery =
	s.CreateMultiQuery()
	.Add(
		s.CreateQuery("from Item i where i.Id > ?")
		.SetInt32(0, 50).SetFirstResult(10))
	.Add(
		s.CreateQuery("select count(*) from Item i where i.Id > ?")
		.SetInt32(0, 50));

IList results = multiQuery.List();
IList items = (IList) results[0];
long count = (long) ((IList) results[1])[0];
```

I would rewrite with Future as follows:
```C#
var batch = new QueryBatch(s.GetSessionImplementation());
IFutureList<Item> futureItems = batch.AddAsList<Item>(
	s.CreateQuery("from Item i where i.Id > ?")
	.SetInt32(0, 50).SetFirstResult(10));

IFutureValue<long> futureCount = batch.AddAsValue<long>(
	s.CreateQuery("select count(*) from Item i where i.Id > ?")
	.SetInt32(0, 50));

IList<Item> items = futureItems.Value;
long count = futureCount.Value;
```

With AddOnAfterLoad:
```C#
IList<EntityComplex> items = null;
long count = 0;

var batch = new QueryBatch(s.GetSessionImplementation());

batch.AddOnAfterLoad<EntityComplex>(
	s.CreateQuery("from Item i where i.Id > ?")
	.SetInt32(0, 50).SetFirstResult(10),
	r => items = r);

batch.AddOnAfterLoad<long>(
	s.CreateQuery("select count(*) from Item i where i.Id > ?")
	.SetInt32(0, 50),
	r => count = r.FirstOrDefault());

batch.Execute();

//process initialized items and count values
```

Or to be close to initial code as:
```C#
List<object> results = new List<object>();

var batch = new QueryBatch(s.GetSessionImplementation());

batch.AddOnAfterLoad<EntityComplex>(
	s.CreateQuery("from Item i where i.Id > ?")
	.SetInt32(0, 50).SetFirstResult(10),
	r => results.Add(r));

batch.AddOnAfterLoad<long>(
	s.CreateQuery("select count(*) from Item i where i.Id > ?")
	.SetInt32(0, 50),
	r => results.Add(r));

batch.Execute();

IList items = (IList)results[0];
long count = (long)((IList)results[1])[0];
```
As I already said why can't Future be default way for retrieving values (no matter separate batch or session future batch)? And AddOnAfterLoad is added just for user customization if they don't like Future for some reasons.

fredericDelaporte:
> As I already said why can't Future be default way for retrieving values (no matter separate batch or session future batch)? And AddOnAfterLoad is added just for user customization if they don't like Future for some reasons.

I have answered:

> From my point of view, futures are an additional abstraction layer on top of multi-queries. It makes sens to let users use directly multi-queries/query-batch if they wish too.

Futures hide away the query batch. User may want to define its own batches of queries and get results from them, instead of having to keep separated references to each batched query (or to closure captured variables).

bahusoid:
>User may want to define its own batches of queries and get results from them

You are referring for old implementation that requires to use not typed classes and explicit typing.
New implementation avoids it. I don't think that new batcher should support this scenario. If someone wants it  - he can write own proxy class that will contain closured list that will be populated via `AddOnAfterLoad`.

fredericDelaporte:
What does actually bother me is:

> But having two implementations of multi-queries/query-batch is not a thing which should be kept in my opinion.

This is not a good move. So either we obsolete (maybe in an upcoming PR) the previous multi-query/criteria API without providing an equivalent API on the query batch because "that is not modern programming", or the query batch should expose an equivalent API, not bloated by futures specific concerns.
So in other words, obsolete multi-query/criteria for futures, or obsolete multi-query/criteria for query batch but changed for offering a similar API.

bahusoid:
Can we keep old implementations feature frozen? If someone suffers from multi criteria/query bug - he should use new batcher. But let current implementations live for time being (at least for few upcoming versions). Like add 7.0 TODO Make obsolete :).

I actually have no objections regarding:
>obsolete the previous multi-query/criteria API without providing an equivalent API on the query batch because "that is not modern programming"

But I can try to make existing MultiCriteria and MultiQuery classes as wrappers around QueryBatch (the way I already described with closured _results list provided to OnAfterLoad). But behavior will be changed for cache (I think of it  as an improvement but who knows how it's used). And I need to check all the options - maybe some additional changes need to be done in IQueryBatchItem implementations (at least something needs to be done to support SetForceCacheRefresh methods). Though frankly speaking this task burdens me.

src/NHibernate/Multi/LinqBatchItem.cs
```diff
@@ -0,0 +1,110 @@
...
+using NHibernate.Linq;
+using NHibernate.Util;
+using Remotion.Linq.Parsing.ExpressionVisitors;
+
+namespace NHibernate
+{
+	public partial class LinqBatchItem<T> : QueryBatchItem<T>
+	{
+		private static Delegate _postExecuteTransformer;
+		private static NhLinqExpression _linqEx;
```
fredericDelaporte: These members are specific to the query and cannot be static. They must be instance variables.

bahusoid: ouch..

src/NHibernate/Async/Impl/AbstractQueryImpl.cs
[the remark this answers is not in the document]
maca88: With the version `0.8.2.7` both issues were corrected, so you can use one of the two mentioned approaches.

bahusoid:
@fredericDelaporte Let me know if you are OK with my latest changes and I will start squashing.

src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -1048,5 +1050,8 @@ public override string ToString()
 		}
 
 		protected internal abstract IEnumerable<ITranslator> GetTranslators(ISessionImplementor sessionImplementor, QueryParameters queryParameters);
-	}
+
+		// Since v5.2
+		[Obsolete("This method has no usages and will be removed in a future version")]
+		protected internal abstract Task<IEnumerable<ITranslator>> GetTranslatorsAsync(ISessionImplementor sessionImplementor, QueryParameters queryParameters, CancellationToken cancellationToken);	}
```
fredericDelaporte: A mishap has occurred on the class closing brace. It is now on the method line instead of having its own.

src/NHibernate/Multi/IQueryBatchItem.cs
```diff
@@ -48,7 +54,7 @@ public partial interface IQueryBatchItem
 		void ProcessResults();
 
 		/// <summary>
-		/// Immediate query execution in case dialect is non batchable
+		/// Immediate query execution in case dialect is non batched
```
fredericDelaporte: I would have not change this comment. The query execution is not batched, due to the dialect being non-batchable. Or since that is not the dialect which is batched anyway, we could also write `in case the dialect does not support batches.`. (This is not a request for change, this method is not meant to be directly used by the user anyway.)

src/NHibernate/Multi/IQueryBatchItem.cs
```diff
@@ -0,0 +1,66 @@
...
+		/// Returns loaded typed results by query.
+		/// Must be called only after <see cref="IQueryBatch.Execute"/>.
+		/// </summary>
+		IList<TResult> GetResults();
+		
+		/// <summary>
+		/// Event is executed after results are loaded by batch.
+		/// Loaded results are provided in action parameter.
+		/// </summary>
+		event Action<IList<TResult>> OnAfterLoad;
```
fredericDelaporte: This look more like a callback than like an `event`. An `event` should respect general `event` [guidelines](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-publish-events-that-conform-to-net-framework-guidelines), as does `BeforeBindMapping` and `AfterBindMapping` in NHibernate.

Either made it a simple callback (then maybe renamed `AfterLoadCallback`), or a full blown event.

If kept as an event, it should be renamed `AfterLoad` by the way. The `On` prefix is normally reserved for the class method responsible of raising the event if it has been affected, like `OnBeforeBindMapping` and `OnAfterBindMapping` in NHibernate.

I wonder if the naming `AfterLoad` should be changed for avoiding any confusion with the `ISession.Load`, with `ILoadEventListener`/`IPostLoadEventListener`/`IPreLoadEventListener` (which are not about a .Net `event` in spite of their name; this is some Hibernate Java legacy), and with the `ILifeCycle.OnLoad`. Maybe `AfterExecution` or `Executed`, but well, I have not set up my mind on this subject.

About the choice between callback or event, it sounds more natural to me to change it to a simple callback. Then why not re-adding additional `Add` `IQueryBatch` extension overloads for each query interface, like `IQueryBatch Add<TResult>(this IQueryBatch batch, IQueryOver query, Action<IList<TResult>> afterLoad)`. (Doing that with an `event` would look quite unnatural to me.)

bahusoid: I'm not sure I understand the problem about event vs callback. I meant it as simple callback. But it was marked as event in interface just for protection. So caller code can't trigger that callback.

fredericDelaporte: Then replace it by a setter function if this is not meant to be an `event` and wants to protect against someone calling it from outside. (I do not really see the point of that protection by the way. If the user wants to do moot things, good for him. Anyway he is the one having supplied the callback in the first place, he can still hold on it and call it directly if he wishes.)

bahusoid: >Then replace it by a setter function

I don't see a problem being it event or delegate. Let it be whatever you think is best - so I've already made changes you requested.

fredericDelaporte: > I don't see a problem being it event or delegate.

Me neither, provided it is an actual event respecting .Net guidelines.

> But it was marked as event in interface just for protection.

In my view, this is a debatable practice for a public facing library. Best practices of the language should be respected, for the sake of less surprise principle. Hi-jacking a language concept for benefiting of one of its side feature while not actually implementing the concept should be reserved to private code (if not avoided at all).
(Of course since NHibernate comes from Hibernate, many Java practices potentially contradictory with .Net ones have been left. But still, better avoid adding more unconventional practices.)

src/NHibernate/Loader/Loader.cs
```diff
@@ -1638,48 +1638,73 @@ protected IList List(ISessionImplementor session, QueryParameters queryParameter
...
 		private IList ListIgnoreQueryCache(ISessionImplementor session, QueryParameters queryParameters)
 		{
 			return GetResultList(DoList(session, queryParameters), queryParameters.ResultTransformer);
 		}
 
-		private IList ListUsingQueryCache(ISessionImplementor session, QueryParameters queryParameters, ISet<string> querySpaces)
+		internal void ProcessCachedResults(
+			ISessionImplementor session,
+			QueryParameters queryParameters,
+			ISet<string> querySpaces,
```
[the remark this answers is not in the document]
fredericDelaporte: The more I look into it, the more I think it just need to be reverted. DRY (don't repeat yourself) is lower on the priority list than KISS (keep it stupid simple).

src/NHibernate/Loader/Loader.cs
```diff
@@ -1637,48 +1637,73 @@ protected IList List(ISessionImplementor session, QueryParameters queryParameter
...
+				});
 
+			//means cache is disabled
+			if (cachedResults == null)
+				return null;
+
+			return GetResultList(cachedResults, queryParameters.ResultTransformer);
+		}
+
+		internal void TransformCachedResults(QueryParameters queryParameters, CacheableResultTransformer transformer, ref IList result)
```
fredericDelaporte: `ref` and `out` parameters are to be avoided whenever possible. Having a `void` return type with a `ref` calls for refactoring the `ref` out of the method signature.

fredericDelaporte:
From my point of view, yes. But since the last commit is some refactoring I have made, I was expecting a feedback from @bahusoid, before approving.

I am already thinking of a follow-up PR which would obsolete the current multi-query/criteria for the query batcher introduced here. I do not think it should be done directly here, because this PR is already quite big, and its main subject was futures.

bahusoid:
@fredericDelaporte I have no objections with your latest changes. And I also think it's ready.

fredericDelaporte:
Rebased and squashed.

src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -0,0 +1,231 @@
...
+				if (_subselectResultKeys[i] != null)
+				{
+					queryInfo.Loader.CreateSubselects(_subselectResultKeys[i], queryInfo.Parameters, Session);
+				}
+
+				// Handle cache if cacheable.
+				if (queryInfo.Cache != null)
+				{
+					queryInfo.Loader.PutResultInQueryCache(Session, queryInfo.Parameters, queryInfo.Cache, queryInfo.CacheKey, _loaderResults[i]);
+				}
```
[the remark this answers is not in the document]
bahusoid: > The if condition is not right for this currently

It seems this `if` condition can be left in place and deal only with putting in cache. And for `TransformCacheableResults` we can use `queryInfo.CacheKey.ResultTransformer` (basically old code can be copy pasted here with slight adjustments as separate check)

src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -0,0 +1,231 @@
...
+							keys = new EntityKey[entitySpan]; //can't reuse in this case
+						}
+
+						tmpResults.Add(o);
+					}
+					_loaderResults[index] = tmpResults;
+
+					if (index == _queryInfos.Count - 1)
+					{
+						InitializeEntitiesAndCollections(reader, hydratedObjects);
```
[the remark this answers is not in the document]
bahusoid: It's still can stay here - just be placed outside `yield return' and handle early exits. As I already said I didn't want to leak entities loading related methods to base batch interfaces (and now if I understand correctly it's required anyway for "batchable cache")

src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -0,0 +1,231 @@
...
+				var index = i;
+				yield return reader =>
+				{
+					if (advanceSelection)
+					{
+						Loader.Loader.Advance(reader, selection);
+					}
+					if (queryParameters.HasAutoDiscoverScalarTypes)
+					{
+						loader.AutoDiscoverTypes(reader, queryParameters, null);
```
fredericDelaporte: I was forgetting this one spotted in #1783, which should also be included in a PR for fixing other trouble highlighted by my previous review.

src/NHibernate/Multi/QueryBatchItemBase.cs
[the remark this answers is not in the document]
fredericDelaporte: We do not have the `reader` outside of the yield. The `reader` is a yielded lambda parameter. Storing it in some local variable belonging to the outer scope would be quite a code smell.

The simplest fix seems to me switching to a regular method doing the processing directly in the query item, instead of yielding a callback for letting the batcher finish the work.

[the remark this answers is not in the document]
fredericDelaporte: See in #1789 how I have fixed it. No additional method. Instead, I have change the one yielding those handlers for a method directly doing the job.

Issue #1719, opened 2018-05-31
PropertyAccessException: Non-static method requires a target with one-to-one relations


fubar-coder:
It seems that there is a problem with one-to-one relations or my mapping.

Exception:

```text
Exception occurred getter of TestNhBug.FileEntry.Id
NHibernate.PropertyAccessException: Exception occurred getter of TestNhBug.FileEntry.Id ---> System.Reflection.TargetException: Non-static method requires a target.
   at System.Reflection.RuntimeMethodInfo.CheckConsistency(Object target)
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
   at NHibernate.Properties.BasicPropertyAccessor.BasicGetter.Get(Object target)
   --- End of inner exception stack trace ---
   at NHibernate.Properties.BasicPropertyAccessor.BasicGetter.Get(Object target)
   at NHibernate.Tuple.Entity.AbstractEntityTuplizer.GetIdentifier(Object entity)
   at NHibernate.Persister.Entity.AbstractEntityPersister.IsTransient(Object entity, ISessionImplementor session)
   at NHibernate.Engine.ForeignKeys.IsTransientFast(String entityName, Object entity, ISessionImplementor session)
   at NHibernate.Engine.ForeignKeys.IsTransientSlow(String entityName, Object entity, ISessionImplementor session)
   at NHibernate.Event.Default.DefaultDeleteEventListener.OnDelete(DeleteEvent event, ISet`1 transientEntities)
   at NHibernate.Impl.SessionImpl.FireDelete(DeleteEvent event, ISet`1 transientEntities)
   at NHibernate.Impl.SessionImpl.Delete(String entityName, Object child, Boolean isCascadeDeleteEnabled, ISet`1 transientEntities)
   at NHibernate.Engine.Cascade.CascadeOn(IEntityPersister persister, Object parent, Object anything)
   at NHibernate.Event.Default.AbstractFlushingEventListener.CascadeOnFlush(IEventSource session, IEntityPersister persister, Object key, Object anything)
[...]

PR #1720, opened 2018-05-31
Fix cascade + one-to-one failing on null no-proxy association

fubar-coder:
A unit test failure is expected, demonstrating #1719.


fredericDelaporte:
I have pushed a minimal fix, see commit details.

fredericDelaporte:
Something causes the test to fail for Oracle, going to check locally.

PR #1727, opened 2018-06-03
Fix a null-ref exception with no-proxy one-to-one

fredericDelaporte:
This null reference exception was swallowed and was only bloating logs.

The call with a `null` command originates from here:
https://github.com/nhibernate/nhibernate-core/blob/ee91396495a5e0874704f22c24ccde7412dafea0/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs#L1279-L1307

As `CloseReader` already does a silent early exit in case of `null`, I have chosen to fix this similarly in `CloseCommand`.

### DIFF
--- a/src/NHibernate/AdoNet/AbstractBatcher.cs
+++ b/src/NHibernate/AdoNet/AbstractBatcher.cs
@@ -316,6 +316,8 @@ namespace NHibernate.AdoNet
 
 		private void CloseCommand(DbCommand cmd)
 		{
+			if (cmd == null)
+				return;
 			try
 			{
 				// no equiv to the java code in here


src/NHibernate/AdoNet/AbstractBatcher.cs
```diff
@@ -316,6 +316,8 @@ public virtual void CloseCommands()
 
 		private void CloseCommand(DbCommand cmd)
 		{
+			if (cmd == null)
+				return;
```
hazzik: Shouldn't it be `cmd?.Dispose();`?

fredericDelaporte: The net effect of the null ref exception was to discard all the method remaining code. Doing `cmd?.Dispose();` will have side effects that were not done previously in case of `null`.
If going that way, then I would also change the abstract entity persister for having it no more calling close with null. And I think all current calls to close should be checked for similar cases, in order to avoid an unwanted side effect.

hazzik: Yes, I did not notice the return.

PR #1728, opened 2018-06-03
Generate a correct proxy for interfaces

fredericDelaporte:
Static proxies generated for interfaces were incorrectly attempting to call the base "interface implementation", causing a PeVerify failure (but no runtime failure, the invalid code being on a never executed path). And they were all sharing the same type name.

(This flaw is also present in the dynamic proxy, but probably since its debut. Being now obsoleted, I do not intend to fix it.)

### DIFF
--- a/src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs
+++ b/src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs
@@ -108,6 +108,34 @@ namespace NHibernate.Test.StaticProxyTest
 		}
 
 		[Test]
+		public void VerifyProxyForClassWithAdditionalInterface()
+		{
+			var factory = new StaticProxyFactory();
+			factory.PostInstantiate(
+				typeof(PublicInterfaceTestClass).FullName,
+				typeof(PublicInterfaceTestClass),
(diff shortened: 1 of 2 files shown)


hazzik:
What was the error?

hazzik:
What was the error?

Is it even possible to get null `__lazyInitializer`?

hazzik:
Ok, found the answer to my second question here:

```
// When deriving from the entity class, the entity class constructor may trigger
// virtual calls accessing the proxy state before its own constructor has a chance
// to initialize it. So although lazyInitializer is never supplied as null to the
// proxy constructor, we must guard nonetheless against it being null during base
// constructor call.
```

fredericDelaporte:
When I attempted re-implementing the lazy property proxy as a delegating proxy, I had to debug that trouble. Some tests where triggering it. I have not checked if removing that from the entity proxy would also cause some tests to start failing for the same reason. But since lazy entity proxies are used in many more tests than lazy property proxies, I am confident it would.

fredericDelaporte:
> What was the error?

The error fixed by this PR is invalid IL (failing PeVerify) generated for proxies on interface. These proxies were still usable, because the invalid code path was never executed. But better generate correct proxies.

PR #1731, opened 2018-06-04
Query cache always missed in session having altered the entities

### LINKED ISSUE #1730: Query cache always missed in session having altered the entities
Refer to https://hibernate.atlassian.net/browse/HHH-5210

When using a query which is marked as cacheable, NHibernate caches both the query with the session timestamp, and its last invalidation timestamp (due to an update or save). When retrieving a query from the cache, it first compares the timestamp of the cached query with the last invalidation timestamp (`UpdateTimestampsCache.IsUpToDate()`):

```c#
if ((long) lastUpdate >= timestamp)
{
	return false;
}
```

The problem is that the timestamp for the current session, used for caching query, is always lower than invalidation's timestamp if the entities were altered in the session, even prior to caching the query and in a dedicated transaction. This causes the result of the method to be always `false`, discarding the cache results and hitting the database forever with that session.

igitur:
Fixes #1730 

This is my first fix to NHibernate. Please let me know what else is required on this PR. Ideally, I'd like this patch to be included on the 4.1 branch. Anything I can do to make that easier?


hazzik:
@fredericDelaporte 5.1.3 or 5.2?

hazzik:
Also, I think that async tests are missing.

igitur:
Ah, yes, I'll add async tests.

Comments on the possibility of backporting to 4.1?

hazzik:
4.1 unlikely. What’s holding you from updating to 5.x?

igitur:
Stuck on net452 for a while.

hazzik:
@fredericDelaporte odd semi-unrelated question (but I was thinking about it for a while): .net 4.5 support in 5.2, yeah/nah? 

I did initial analysis: the only thing is missing is `AsyncLocal<>`. It has 2 usages: 
- `AsyncLocalSessionContext` - this has to go into the void;
- `AdoNetWithSystemTransactionFactory._bypassLock` - this can be replaced with `CallContext.LogicalGetData`/`CallContext.LogicalSetData` surrounded to compilation directive.

A rationale for that: to get a better user acquisition on a newer version.

igitur:
Async test added.

fredericDelaporte:
> Async test added.

Hum, have you missed a `git add`?

hazzik:
It seems it's manual addition

hazzik:
https://twitter.com/AlfredoCambera/status/1002966376468770816

fredericDelaporte:
> It seems it's manual addition

In other words @igitur, have you used the async generator like explained [here](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md#regenerate-async-code)?

igitur:
> In other words @igitur, have you used the async generator like explained here?

Now I have.

src/NHibernate.Test/NHSpecificTest/GH1730/Fixture.cs
```diff
@@ -0,0 +1,68 @@
...
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.GH1730
+{
+	[TestFixture]
+	public class Fixture : BugTestCase
+	{
+		protected override void Configure(Configuration configuration)
+		{
+			configuration.SetProperty("show_sql", "true");
```
hazzik: is this required?

src/NHibernate.Test/NHSpecificTest/GH1730/Fixture.cs
```diff
@@ -0,0 +1,68 @@
...
+			configuration.SetProperty("generate_statistics", "true");
```
hazzik: Please use a property from Cfg.Environment

src/NHibernate.Test/NHSpecificTest/GH1730/Fixture.cs
igitur: Errr, no.

src/NHibernate.Test/NHSpecificTest/GH1730/Fixture.cs
igitur: Ok, done.

fredericDelaporte:
> 5.1.3 or 5.2?

I am more for 5.2. Especially since the opener would need this down to 4.1, which is quite unlikely we will do for a trouble which is not a serious regression. Moreover if we do support Fx4.5 in 5.2, that may suit him as I understand his need for having it in 4.1.

igitur:
If you do plan to support net45 in 5.2, I can definitely wait until then.

fredericDelaporte:
Still on the unrelated subject, `TransactionScopeAsyncFlowOption` seems to be there only from 4.5.1, so that will be net451 rather than net45. #ifdef-ing that would create bugs with async-await.

hazzik:
Tried to do .NET 4.5 and found more issues: `Task.FromCancelled(...)`, `Task.FromException(...)`, `Task.CompletedTask`, and `Array.Empty<T>()` are missing. The array is easy to fix. Task is not so - it requires changes to AsyncGenerator. So, let's close the off-topic for now.

fredericDelaporte:
So well, we are back to "where should we put that fix"... I am still do not think we should backport it to 4.1. Of course @igitur could rebase this PR on 5.1.x for inclusion in 5.1.3 then he could create backport PRs cherry-picking the change in 5.0.x and 4.1.x, reducing the work on the team side.
But that would be still publishing two additional releases for a non-regression bug, which is time consuming. And this could incite more people to ask for more backports, which I would rather avoid.

igitur:
Don't worry about the 4.1 request. I can use this as motivation to upgrade our apps to net461+. 

Yes, I can rebase 5.1.x if @hazzik is OK with that.

igitur:
Rebased on `5.1.x`

hazzik:
Thanks @igitur

PR #1733, opened 2018-06-05
[WIP] Cacheable Multicriteria/Future'd query with aliased join throw exception

igitur:
Original Jira post at https://nhibernate.jira.com/browse/NH-3864
Subsequent issue filed at #1344 

I'm adding the unit tests that I found first. I don't have a fix yet, but I'm willing to have a bash at it.


fredericDelaporte:
Better take the habit of generating async code within each commit.

igitur:
@fredericDelaporte I did and it failed. Parking that part until I finalise the tests.

bahusoid:
If description is correct:
>The bug actually lies is in MultipleQueriesCacheAssembler Disassemble method.

I believe with new batcher https://github.com/nhibernate/nhibernate-core/pull/1718 Future is not affected by this issue (as `MultipleQueriesCacheAssembler` is not used - cache is created per each query)

fredericDelaporte:
Maybe the regen fails due to maca88/AsyncGenerator#70. You would have to try with methods instead of delegate parameters.

@bahusoid, interesting indeed. But of course better checking that with appropriate tests, so having them thanks to this PR ist still a good thing. Maybe you could then cherry-pick them in #1718 for ascertaining it does also fix #1344.

igitur:
@bahusoid That's good news. I ran the tests against your branch and 3 of the 5 tests still fail. I can push a PR to your repo to show you what I mean?

There still a problem with `QueryLoader.GetResultList` at the `var row = (Object[]) results[i];` line. `System.InvalidCastException : Unable to cast object of type 'NHibernate.Test.NHSpecificTest.NH3864.Person' to type 'System.Object[]'.`

And the `CacheableMulticriteria_QueryOverWithAliasedJoinQueryOver` test still uses `MultipleQueriesCacheAssembler` (even on your branch).

bahusoid:
New batcher is used only with Future methods.

MultiCriteria and MultiQuery are separate batchers.  If you need MulitCriteria or MultiQuery - you still need to fix it.

igitur:
> Is #383 the same?

Not sure if this question is directed at me. I'm not comfortable enough to answer it convincingly.

hazzik:
@igitur its to whomever capable to answer this question.

igitur:
@maca88 I see you've been commenting on related PRs. Just, FYI, my goal for this PR was to enable caching when using `.Include()` methods from your `NHibernate.Extensions` project.

fredericDelaporte:
> There still a problem with `QueryLoader.GetResultList` at the `var row = (Object[]) results[i];` line. `System.InvalidCastException : Unable to cast object of type 'NHibernate.Test.NHSpecificTest.NH3864.Person' to type 'System.Object[]'.`

This is a bug in query batch implementation indeed. It should be fixed with #1789.

> And the `CacheableMulticriteria_QueryOverWithAliasedJoinQueryOver` test still uses `MultipleQueriesCacheAssembler` (even on your branch).

This would change with #1783, by obsoleting Multi-Criteria/Query APIs in favor of using the query batcher. Those tests should be converted for using the query batcher.

> Is #383 the same?

They could have the same root cause.

fredericDelaporte:
With #1783, the flawed class is now obsoleted.

PR #1742, opened 2018-06-10
Add new DB2CoreDriver to use with IBM.Data.DB2.Core provider

hazzik:
7dfb32cadc5ce06fd8246b2dcf8b3150

### DIFF
--- a/src/NHibernate/Driver/DB2CoreDriver.cs
+++ b/src/NHibernate/Driver/DB2CoreDriver.cs
@@ -1,0 +1,12 @@
+namespace NHibernate.Driver
+{
+	/// <summary>
+	/// A NHibernate Driver for using the IBM.Data.DB2.Core DataProvider.
+	/// </summary>
+	public class DB2CoreDriver : DB2DriverBase
+	{		
+		public DB2CoreDriver() : base("IBM.Data.DB2.Core")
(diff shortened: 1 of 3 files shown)


src/NHibernate/Dialect/DB2Dialect.cs
```diff
@@ -128,8 +114,6 @@ public DB2Dialect()
 
 			// DB2 does not support ANSI substring syntax.
 			RegisterFunction("substring", new SQLFunctionTemplate(NHibernateUtil.String, "substring(?1, ?2, ?3)"));
-
-			DefaultProperties[Environment.ConnectionDriver] = "NHibernate.Driver.DB2Driver";
```
fredericDelaporte: Removing this is a breaking change for all apps which were using the dialect without specifying the driver. Better keep-it, eventually guarded by a compilation directive for not specifying an invalid driver under .Net Core and a maybe invalid one under .Net Standard.

hazzik: Reverted

fredericDelaporte:
HMM3. Your screenshot is from that game, of course a bit tweaked.

hazzik:
Get it now.

PR #1744, opened 2018-06-11
Reconnect lazy property proxy on deserialization

fredericDelaporte:
After deserializing a session holding entities with uninitialized lazy properties, the lazy properties cannot be accessed because their proxy does no more have a session.

This is an old issue, independent of the used proxy factory. (Provided it is considered that proxies should not serialize their session, in order to avoid serializing the whole session when just serializing one entity.)


src/NHibernate/Engine/StatefulPersistenceContext.cs
```diff
@@ -1474,6 +1475,15 @@ void IDeserializationCallback.OnDeserialization(object sender)
 					throw new InvalidOperationException(me.Message, me);
 				}
 			}
+
+			// Reconnect the lazy property proxies
+			foreach (var p in entitiesByKey)
+			{
+				if (p.Value is IFieldInterceptorAccessor lazyPropertyProxy && lazyPropertyProxy.FieldInterceptor != null)
```
fredericDelaporte: `AbstractEntityPersister.AfterReassociate` does also this kind of things, but it is meant for detached entities and also handle dirtiness, which should left untouched in the session deserialization case.

src/NHibernate.Test/StaticProxyTest/StaticProxyFixture.cs
```diff
@@ -0,0 +1,477 @@
...
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				var se = new SimpleEntity
+				{
+					Name = "1",
+					Text = "Text1"
+				};
+				s.Save(se);
+				_idSimpleEntity1 = se.Id;
```
hazzik: Better to write `_idSimpleEntity1 = (Guid)s.Save(se);`. You can declare the `_idSimpleEntity1` field as object and drop casting as the concrete type of the key does not matter.

fredericDelaporte: This looks to me as quite a matter of taste. I rather have two lines of code rather than a hard-coded cast. Granted, the test case does not need the exact typing of the id.

PR #1745, opened 2018-06-12
Remove obsoleted hibernate configuration prefix

fredericDelaporte:
The documentation was still prefixing some configuration properties
with the "hibernate." prefix, which has been removed since NH 2.0.0.GA.

### DIFF
--- a/doc/reference/modules/toolset_guide.xml
+++ b/doc/reference/modules/toolset_guide.xml
@@ -66,7 +66,7 @@
         
         <para>
             You <emphasis>must</emphasis> specify a SQL <literal>Dialect</literal> via the 
-            <literal>hibernate.dialect</literal> property when using this tool.
+            <literal>dialect</literal> configuration property when using this tool.
         </para>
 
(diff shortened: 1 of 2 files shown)


doc/reference/modules/toolset_guide.xml
```diff
@@ -262,23 +262,23 @@ new SchemaExport(cfg).Create(false, true);]]></programlisting>
...
                     <tbody>
                     <row>
-                        <entry><literal>hibernate.connection.driver_class</literal></entry>
+                        <entry><literal>connection.driver_class</literal></entry>
                         <entry>jdbc driver class</entry>
                     </row>
                     <row>
-                        <entry><literal>hibernate.connection.url</literal></entry>
+                        <entry><literal>connection.url</literal></entry>
                         <entry>jdbc url</entry>
```
fredericDelaporte: This part of the documentation is commented out, so I have just removed the prefix without fixing other Hibernate specific discrepancies.

PR #1747, opened 2018-06-14
NH-3823 - Fix extra lazy PersistentGenericSet losing pending items on initialization

### LINKED ISSUE #1339: NH-3823 - Initialization of Set with Lazy=Extra causes pending additions to disappear
**Martin Francisco** created an [issue](https://nhibernate.jira.com/browse/NH-3823) — <time datetime="2015-09-03T01:07:45.351Z" title="2015-09-03T01:07:45.351Z">3rd September 2015, 1:07:45</time>:
> See attached patch for test case.
> 
> It looks like adding an item to an uninitialized set causes the item to be queued (NHibernate.Collection.Generic.PersistentSet line 324).  Later when the set gets initialized the pending queued operations never get performed.  It looks like this normally happens in NHibernate.Collection.AbstractPersistentCollection during EndRead which calls AfterInitialize.  However PersistentSet overrides EndRead and does not call base.
> 

* [ExtraLazyFixture.patch.zip](https://github.com/nhibernate/nhibernate-core/files/2101908/ExtraLazyFixture.patch.zip)

hazzik:
Fixes #1339


hazzik:
Rebased

src/NHibernate.Test/Async/NHSpecificTest/NH1316/Fixture.cs
```diff
@@ -9,10 +9,8 @@
 
 
 using System;
-using System.Data.Common;
-using NUnit.Framework;
 using NHibernate.Dialect;
-
+using NUnit.Framework;
```
fredericDelaporte: On my first approval I have not actually verified this change is legit. I guessed this is a "fix by the way" of a previous PR lacking a small bit of async regen tests. Is that it, or some unexpected change instead?

hazzik: Yes it is.

fredericDelaporte:
Sorry, bad reflex, I have clicked that button before realizing this was not my repository.

hazzik:
I don’t follow

fredericDelaporte:
I have deleted the branch of this PR because since you have added me to your repository, it is available. Then I realized it was not a branch of my repository, and restored it.

hazzik:
Ok

PR #1750, opened 2018-06-14
Avoid more cases releasing a connection from transaction completion

fredericDelaporte:
This is a follow up to [NH-4011](https://nhibernate.jira.com/browse/NH-4011), about system transactions integration.
When a session is disposed inside a scope while not having disabled flush from system transaction completion (`transaction.use_connection_on_system_prepare`), its connection is not released until transaction completion. Releasing connection from there was determined as source of random crashes (mainly connection pool corruption), at least with SQL-Server. Thus the introduction of the setting.
It appears Npgsql may also have such troubles (see npgsql/npgsql#1985).

This change allows to remove most remaining cases of connection closing from transaction completion, by always closing it from the session disposal, provided it may not be used by the flush from transaction
completion.

### DIFF
--- a/src/NHibernate/AdoNet/ConnectionManager.cs
+++ b/src/NHibernate/AdoNet/ConnectionManager.cs
@@ -484,6 +484,21 @@ namespace NHibernate.AdoNet
 			_connection.EnlistTransaction(transaction);
 		}
 
+		/// <summary>
+		/// If the connection may not be used by a system transaction completion, release it.
+		/// </summary>
+		public void ReleaseConnectionNotNeededForSystemTransaction()
+		{
+			// The current connection will not be used if the dialect supports concurrent writting connections
+			// in the same transaction and if the connection is not user supplied. In such case it uses a dedicated
+			// connection to perform the flush from. (See BeginProcessingFromSystemTransaction.)
+			if (_connection != null && _ownConnection &&
+				Factory.Dialect.SupportsConcurrentWritingConnectionsInSameTransaction)
+			{
+				CloseConnection();
+			}
+		}
+
 		public IDisposable BeginProcessingFromSystemTransaction(bool allowConnectionUsage)
 		{
 			var needSwapping = _ownConnection && allowConnectionUsage &&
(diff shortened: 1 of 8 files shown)


src/NHibernate/AdoNet/ConnectionManager.cs
```diff
@@ -484,6 +484,21 @@ public void EnlistIfRequired(System.Transactions.Transaction transaction)
 			_connection.EnlistTransaction(transaction);
 		}
 
+		/// <summary>
+		/// If the connection may not be used by a system transaction completion, release it.
+		/// </summary>
+		public void ReleaseConnectionNotNeededForSystemTransaction()
```
fredericDelaporte: Feel free to suggest a better naming!

fredericDelaporte:
I had tested this with PostgreSQL only. But it appears SqlClient does not properly support `EnlistmentOptions.EnlistDuringPrepareRequired` in distributed rollback cases: when a connection has participated in the transaction, been closed, and that the enlisting prepare phase re-uses one with the same connection string while another resource manager force a rollback (`PreparingEnlistment.ForceRollback`), the connection used in the enlisting prepare phase gets corrupted, and subsequent attempt (Teardown) to use a connection with same connection string fails too.

I will add a commit to do an exception just for SqlClient, for checking the other databases behavior, but I think overall I will end up rejecting this change.

fredericDelaporte:
So excepted SQL-Server, this change is supported by other tested databases and may avoid some promotion to distributed cases. Maybe the Mono could benefit of it, but we would have to detect if we are under Mono in the `SupportsClosingConnectionInsideScopeWithEnlistDuringPrepare` of the Sql Server driver, for keeping it `true` in that case.

src/NHibernate/Driver/IDriver.cs
```diff
@@ -143,6 +143,12 @@ public interface IDriver
 		/// </summary>
 		bool SupportsEnlistmentWhenAutoEnlistmentIsDisabled { get; }
 
+		/// <summary>
+		/// Does this driver connections support being closed during a scope then re-acquired during prepare, including
+		/// in case of concurrent rollback?
+		/// </summary>
+		bool SupportsClosingConnectionInsideScopeWithEnlistDuringPrepare { get; }
```
hazzik: Interface change....

fredericDelaporte: Moreover... Anyway, since even the one responsible for how MSDTC+.Net works together, does not get it right when we close a connection before completion then re-open one in prepare phase, better forget about this.

PR #1755, opened 2018-06-17
Fix deletion of component with null property

### LINKED ISSUE #1170: NH-3646 - Incorrect query when items removed from a collection of components contain null values
**CBP** created an [issue](https://nhibernate.jira.com/browse/NH-3646) — <time datetime="2014-07-31T02:28:27.390Z" title="2014-07-31T02:28:27.390Z">31st July 2014, 2:28:27</time>:
> I have an entity that has a collection of components mapped.
> 
> The Fluent NH mapping like this:
> 
> 
> ```
> public sealed class FooDbMap : ClassMap<Foo>
> {
> 	public FooDbMap()
> 	{
>            ... etc
> 
> 		HasMany(x => x.Bars)
> 			.Component(part =>
> 			           	{
> 			           		part.References(x => x.Qux, "QuxId").LazyLoad().Nullable();
> 			           		part.Map(x => x.Baz).Nullable();
> 			           	})
> 			.AsSet()
> 			.KeyColumn("FooId")
> 			.LazyLoad()
> 			.Table("sfFooBars");
>         }
> }
> ```
> 
> 
> The component class looks like this:
> 
> 
> ```
> public class Bar
> {
> 	public virtual Qux Qux { get; set; }
> 	public virtual string Baz { get; set; }
> 
> 	public bool Equals(Bar other)
> 	{
> 		if (ReferenceEquals(null, other)) return false;
> 		if (ReferenceEquals(this, other)) return true;
> 		return Equals(other.Qux, Qux) && Equals(other.Baz, Baz);
> 	}
> 
> 	public override int GetHashCode()
> 	{
> 		unchecked
> 		{
> 			int result = (Qux != null ? Qux.GetHashCode() : 0);
> 			result = (result*397) ^ (Baz != null ? Baz.GetHashCode() : 0);
> 			return result;
> 		}
> 	}
> }
> ```
> 
> 
> 
> The following test demonstrates the problem:
> 
> 
> ```
> <Test>
> public void Test()
> {
>     var foo = new Foo();
> 	
> 	foo.AddBar(new Bar 
> 		{
> 			Qux = null,
> 			Baz = "asdf"
> 		});
> 		
> 	CurrentSession.Save(foo);
> 	CurrentSession.Flush();
> 	CurrentSession.Clear();
> 	
> 	var reloadedFoo = CurrentSession.Load<Foo>();
> 	
> 	reloadedFoo.RemoveBar(reloadedFoo.Bars.First());
> 	
> 	CurrentSession.Flush();
> }
> ```
> 
> 
> This test fails to remove the bar.
> The reason is because the generated SQL is incorrect:
> 
> 
> ```
> exec sp_executesql N'DELETE FROM Foo WHERE FooId = @p0 AND Baz = @p1 AND Qux = @p2',N'@p0 uniqueidentifier,@p1 varchar(8000),@p2 uniqueidentifier',@p0='494464F4-6B1A-455F-8786-A37900C9CE36',@p1='asdf',@p2=NULL
> ```
> 
> 
> The correct SQL should check if Qux is null using the syntax "Qux is null", rather than "Qux = @p2".
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2014-10-12T10:49:27.701Z" title="2014-10-12T10:49:27.701Z">12th October 2014, 10:49:27</time>:
> A test case would be appreciated
>

fredericDelaporte:
Fixes #1170 - Incorrect query when items removed from a collection
of components contain null values

[NH-3634](https://nhibernate.jira.com/browse/NH-3634) has added support of `null` valued component properties in select queries comparisons, but they are indeed still not properly supported since this still fails when deleting.


fredericDelaporte:
I do not intend to merge this soon, in case someone takes the time to add the fix here before.

fredericDelaporte:
All commits should be squashed.

The last commit may be dropped if preferred: it could be a possible breaking change for external users of the impacted utility methods, if they use it on array of a type implementing `IEquatable` without overriding `object.Equals` or with an `object.Equals` override having a different semantic than the one from `IEquatable`. The previous implementation was always delegating equality to `object.Equals`.

src/NHibernate.Test/NHSpecificTest/GH1170/Fixture.cs
```diff
@@ -0,0 +1,90 @@
...
+namespace NHibernate.Test.NHSpecificTest.GH1170
+{
+	[TestFixture]
+	public class Fixture : BugTestCase
+	{
+		// Only the set case is tested, because other cases were not affected:
+		// - bags delete everything first.
+		// - indexed collections use their index, which is currently not mappable as a composite index with nullable
+		//   column. All index columns are forced to not-nullable by mapping implementation. When using a formula in
+		//   index, they use the element, but its columns are also forced to not-nullable.
```
fredericDelaporte: This "forced to not-nullable" thing seems to me a bit undue, as this impacts only ddl generation, forbidding to test such cases (unless tweaking the generated schema afterward), but not forbidding users not using schema generation to map such things, and insert null. Then these cases will have the delete issue too, and also an issue on update. But no one seems to ask for nullable support on those cases, so I have not handled them.

src/NHibernate.Test/NHSpecificTest/GH1170/Model.cs
```diff
@@ -0,0 +1,37 @@
...
+	{
+		public virtual bool SomeBool { get; set; }
+		public virtual string SomeString { get; set; }
+
+		protected bool Equals(ChildComponent other)
+		{
+			return SomeBool == other.SomeBool && string.Equals(SomeString, other.SomeString);
+		}
+
+		public override bool Equals(object obj)
```
fredericDelaporte: Not overriding it causes the set to always delete all its elements, causing the update test to be even more moot. (It is anyway somewhat moot, has with set, updates are handled by delete/insert.)

src/NHibernate/Persister/Collection/AbstractCollectionPersister.cs
```diff
@@ -468,7 +471,10 @@ public AbstractCollectionPersister(Mapping.Collection collection, ICacheConcurre
 								   ?? ExecuteUpdateResultCheckStyle.DetermineDefault(collection.CustomSQLUpdate, updateCallable);
 			}
 
+			// 6.0 TODO: call GenerateDeleteRowString(null); instead.
+#pragma warning disable 618
 			sqlDeleteRowString = GenerateDeleteRowString();
```
fredericDelaporte: Not replaced yet for avoiding a possible breaking change for those having overridden it.

fredericDelaporte:
Rebased for also updating the documentation. (Fixup commit already squashed.)

src/NHibernate/Util/ArrayHelper.cs
```diff
@@ -179,45 +179,12 @@ public static int CountTrue(bool[] array)
 
 		public static bool ArrayEquals<T>(T[] a, T[] b)
```
hazzik: Turns out this magic already exists in the .NET Fx:

`Array` implements `IStructuralEquatable` interface:

 var a = new[] {1};
            var b = new[] {1};
            IStructuralEquatable aa = a;

            var @equals = aa.Equals(b, EqualityComparer<int>.Default);

src/NHibernate.Test/NHSpecificTest/GH1170/Model.cs
hazzik: I think this behaviour needs to be changed (in a separate PR).

fredericDelaporte: Do you mean, if the component reference has not changed, update it instead of deleting/inserting it? But the persistence context loaded state does not keep track of component references and so, does not allow to check for that condition.
Or do you mean we should somewhat "recycle" components to delete by updating them with data of components to add in order to replace as most as possible deletes/inserts by single updates? That would cause actual removals combined with adds to be replaced by updates, which could be quite unexpected for users, and breaking for custom auditing logic (like trigger based auditing).

src/NHibernate/Util/ArrayHelper.cs
fredericDelaporte: But that magic works by comparing everything as objects as it uses the non generic `IEqualityComparer` interface, causing value types to be boxed.
The `ArrayEquals(byte[] a, byte[] b)` overload would be likely less efficient if re-implemented with this.
The way I have re-implemented this (with the last commit) also allows `ArrayEquals<T>` to avoid boxing.

src/NHibernate.Test/NHSpecificTest/GH1170/Model.cs
hazzik: I'm talking about this behaviour: "Not overriding it causes the set to always delete all its elements" it shouldn't be like this.

fredericDelaporte: From [here](http://nhibernate.info/doc/nhibernate-reference/components.html#components-incollections):

> Note: if you define an `ISet` of composite elements, it is very important to implement `Equals()` and `GetHashCode()` correctly. 

It is in fact an error to map a set of component without a proper equal semantic. The set cannot enforce its semantic in such case: better use a bag instead.

Of course the side effect this error has about flushes, causing the whole set to be purged and recreated, could still be fixed. But that would be fixing a case resulting anyway from a bad mapping/implementation on the user side. Why not if this can be done in a simple way with very few impacts on normal uses, but otherwise, I do not think it should be fixed.

Fixing it will likely require to extract the set content as component values arrays, then searching for each array of values if we can find the same array in the loaded state. This looks to me as a non-negligible overhead, compared to just using the set `Contains` method. Moreover it would then ignore equals implementations having another semantic, which will be a possible breaking change.

israelaece:
I'm using the NH 5.3.5 and I continue dealing with this issue (removing itens from `<set>` collection), but, in my case, the `composite-element` has three properties, two of them are `many-to-one`.

Note that my `DocumentPrefix` property is `not-null="true"`. When I set this property to `not-null="false"` the problem goes away, but in terms of my business domain, this property is not null. Anyway, I can live with this behavior; I'm just reporting the issue to listen your opinion if this is by design or not. Thanks a lot!

    <set name="Itens"
         cascade="all-delete-orphan"
         lazy="true"
         access="field.camelcase">
      <key column="ParentId"/>
      <composite-element class="Avalista">
        <property name="DocumentPrefix"
                  type="AnsiString"
                  not-null="true">
          <column name="DocumentPrefix"
                  not-null="true"
                  sql-type="varchar(11)" />
        </property>
        <many-to-one name="Person"
                     column="PersonId"
                     not-null="false"
                     lazy="false"
                     class="Person" />
        <many-to-one name="Company"
                     column="CompanyId"
                     not-null="false"
                     lazy="false"
                     class="Company" />
      </composite-element>
    </set>

fredericDelaporte:
It is better open a new issue for this, with a more complete example. (With which data does it fails, ...)

PR #1756, opened 2018-06-18
Fix unsaved-value for assigned identifiers

fredericDelaporte:
Revert [NH-1914](https://nhibernate.jira.com/browse/NH-1914) (partially revert 3ccc965)

This change was "fixing" an invalid mapping by breaking `unsaved-value`. Instead of removing an undue `unsaved-value` from the test case mapping, it has caused it to be ignored for assigned identifiers when the identifier value is not matching `unsaved-value`.

(The bug report was associated with NH 2.0.1.GA, where `undefined` unsaved value were not available for ids. So removing the `unsaved-value` with this NH version would not have solved the trouble. What is quite unfortunate is the fact the "fix" hack has been done in NH 2.1.1.GA, while `undefined` unsaved value for ids were added in NH 2.1.0.GA (without a matching Jira, in 82545c5, treating three other subjects), which was already released at the time of the report. So the complete right answer to the user should have been "upgrade to 2.1.0.GA and remove this undue `unsaved-value` you have on your id", instead of corrupting the `unsaved-value` feature in 2.1.1.GA.)


src/NHibernate.Test/NHSpecificTest/NH1914/Mappings.hbm.xml
```diff
@@ -1,54 +0,0 @@
...
-			<key column="ParentID"/>
-			<index column="Indexer" type="Int32"/>
-			<one-to-many class="HLUT"/>
-		</list>
-	</class>
-
-	<class name="HLUT" table="HLUT">
-
-		<!--NotWorking with below id generation.-->
-		<id name="Identifier" unsaved-value="0">
```
fredericDelaporte: Here was the actual bug of the user. Specifying that `0` is an unsaved value means that all other are to be considered as saved value, thus the behavior he was complaining about. Removing completely the `unsaved-value="0"` causes the test to succeed without the "fix", and that is the actual fix which should have been applied.

(Not specifying unsaved value with a assigned generator cause it to be handled as `undefined`, which causes NHibernate to ignore the id value for determining the entity state.)

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -3795,34 +3795,24 @@ public virtual void AfterReassociate(object entity, ISessionImplementor session)
...
 
-            // check the id unsaved-value
-            // We do this first so we don't have to hydrate the version property if the id property already gives us the info we need (NH-3505).
-            bool? result2 = entityMetamodel.IdentifierProperty.UnsavedValue.IsUnsaved(id);
-            if (result2.HasValue)
-            {
-                if (IdentifierGenerator is Assigned)
-                {
-                    // if using assigned identifier, we can only make assumptions
-                    // if the value is a known unsaved-value
```
fredericDelaporte: `if` block removed, just keeping the `else` body: `return result2;`.

(The whole block was space indented since NH-3505, thus the larger change.)

hazzik:
Can you copy the comment from #1354 over here, so it makes it clear why this change is required.

src/NHibernate.Test/Async/Extralazy/ExtraLazyFixture.cs
```diff
@@ -1,4 +1,4 @@
-//------------------------------------------------------------------------------
+﻿//------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
hazzik: UTF8 BOM

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
hazzik: What if we move this block after the version check?

fredericDelaporte:
Yes, of course. Done.

src/NHibernate.Test/Async/Extralazy/ExtraLazyFixture.cs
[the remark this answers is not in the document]
fredericDelaporte: But why has it been changed in dfbcdc9 in the first place?

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
fredericDelaporte: This?
```c#
// check the id unsaved-value
// We do this first so we don't have to hydrate the version property if the id property already gives us the info we need (NH-3505).
bool? result2 = entityMetamodel.IdentifierProperty.UnsavedValue.IsUnsaved(id);
if (result2.HasValue && !(IdentifierGenerator is Assigned))
{
	return result2;
}

// check the version unsaved-value, if appropriate
if (IsVersioned)
{
	object version = GetVersion(entity);
	bool? result = entityMetamodel.VersionProperty.UnsavedValue.IsUnsaved(version);
	if (result.HasValue)
	{
		return result;
	}
}

// if using assigned identifier, we can only make assumptions
// if the value is a known unsaved-value
if (result2 == true)
	return true;
```

That would still break `unsaved-value`. There is no reason at all for this block, excepted not understanding/knowing about `unsaved-value`. If we do not want any assumption to be done for assigned identifiers, we just need to have `unsaved-value="undefined"`, which is the default for assigned identifier. Specifying `unsaved-value="0"` in the mapping then complaining about NHibernate treating all entities with a non-zero value as already existing, is nonsense.

src/NHibernate.Test/Async/Extralazy/ExtraLazyFixture.cs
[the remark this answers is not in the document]
hazzik: Who knows?

hazzik:
> Done.

?

fredericDelaporte:
I have updated the first comment.

Issue #1671, opened 2018-04-27
Decouple IObjectsFactory from BytecodeProvider


hazzik:
Objects factory is used to create majority of the objects. 

To be able to provide users a better experience (eg dependency injection) it would be nice to make ObjectsFactory configurable independently of Bytecode provider.

PR #1758, opened 2018-06-19
Decouple configuration of IObjectsFactory from BytecodeProvider

maca88:
Here is a simple solution to configure `IObjectsFactory` separately from `IBytecodeProvider`, [related issue](https://github.com/nhibernate/nhibernate-core/issues/1671). I know that having it configured statically is not the best solution, but the problem is that today, `IObjectsFactory` is used by many static methods that are public. If we want this in `5.2`, this is the only possible solution that I see.
In addition, the `IObjectsFactory` has an overload with the `nonPublic` parameter which is not possible to implement with most of the IoC containers. The overload is currently used in only two places ([AliasToBeanResultTransformer](https://github.com/nhibernate/nhibernate-core/blob/bced0cd2389fe931d885327b630e1ee9fbed4b75/src/NHibernate/Transform/AliasToBeanResultTransformer.cs#L88) and [PocoInstantiator](https://github.com/nhibernate/nhibernate-core/blob/a058d38eee7d23488a377248b903e1b90ef2e554/src/NHibernate/Tuple/PocoInstantiator.cs#L106)) and all tests passes when the second argument is removed. Wouldn't it be better to obsolete that overload and let the implementor decide whether to use a non public constructor? 

Even the overload with the `ctorArgs` is not possible to implement with some IoC containers as it is [considered as a code smell](https://www.cuttingedge.it/blogs/steven/pivot/entry.php?id=99). This overload is also used only two times, one time in [SessionFactoryImpl](https://github.com/nhibernate/nhibernate-core/blob/1024b71aec3b5e75c80e0ad6515943d6aee4adba/src/NHibernate/Impl/SessionFactoryImpl.cs#L1286) which is not that easy to change because of the circular dependency between `ICurrentSessionContext` and `ISessionFactory` and the second in [SchemaExport](https://github.com/nhibernate/nhibernate-core/blob/f8b1bc46f92c7a8cd503d3450f73f6b33ad4e8fb/src/NHibernate/Tool/hbm2ddl/SchemaExport.cs#L207) which is quite strange as the [ScriptSplitter](https://github.com/nhibernate/nhibernate-core/blob/ad4c2ef101cbf8ba798220973f4f78dd795e0896/src/NHibernate/Tool/hbm2ddl/ScriptSplitter.cs#L9) does not have any virtual members.

### DIFF
--- a/src/NHibernate/Cfg/Environment.cs
+++ b/src/NHibernate/Cfg/Environment.cs
@@ -382,6 +382,16 @@ namespace NHibernate.Cfg
 		}
 
 		/// <summary>
+		/// NHibernate's object instantiator.
+		/// </summary>
+		/// <remarks>
+		/// This should only be set before a configuration object
+		/// is created, otherwise the change may not take effect.
+		/// For entities see <see cref="IReflectionOptimizer"/> and its implementations.
+		/// </remarks>
+		public static IObjectsFactory ObjectsFactory { get; set; } = new ActivatorObjectsFactory();
+
+		/// <summary>
 		/// Whether to enable the use of reflection optimizer
 		/// </summary>
 		/// <remarks>
--- a/src/NHibernate.Test/NHSpecificTest/GH1547/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH1547/Fixture.cs
@@ -132,7 +132,7 @@ namespace NHibernate.Test.NHSpecificTest.GH1547
 
 		public DriverForSubstitutedCommand()
 		{
-			_driverImplementation = (IDriver) Cfg.Environment.BytecodeProvider.ObjectsFactory.CreateInstance(DriverClass);
+			_driverImplementation = (IDriver) Cfg.Environment.ObjectsFactory.CreateInstance(DriverClass);
 		}
 
 		DbCommand IDriver.GenerateCommand(CommandType type, SqlString sqlString, SqlType[] parameterTypes)
--- a/src/NHibernate.Test/TypesTest/AbstractDateTimeTypeFixture.cs
+++ b/src/NHibernate.Test/TypesTest/AbstractDateTimeTypeFixture.cs
@@ -533,7 +533,7 @@ namespace NHibernate.Test.TypesTest
 
 		public ClientDriverWithParamsStats()
 		{
-			_driverImplementation = (IDriver) Cfg.Environment.BytecodeProvider.ObjectsFactory.CreateInstance(DriverClass);
+			_driverImplementation = (IDriver) Cfg.Environment.ObjectsFactory.CreateInstance(DriverClass);
 		}
 
 		private static void Inc<T>(T type, IDictionary<T, int> dic)
--- a/src/NHibernate/Async/Tool/hbm2ddl/SchemaExport.cs
+++ b/src/NHibernate/Async/Tool/hbm2ddl/SchemaExport.cs
@@ -187,7 +187,7 @@ namespace NHibernate.Tool.hbm2ddl
 			cancellationToken.ThrowIfCancellationRequested();
 			if (dialect.SupportsSqlBatches)
 			{
-				var objFactory = Environment.BytecodeProvider.ObjectsFactory;
+				var objFactory = Environment.ObjectsFactory;
 				ScriptSplitter splitter = (ScriptSplitter)objFactory.CreateInstance(typeof(ScriptSplitter), sql);
 
 				foreach (string stmt in splitter)
--- a/src/NHibernate/Async/Tool/hbm2ddl/SchemaUpdate.cs
+++ b/src/NHibernate/Async/Tool/hbm2ddl/SchemaUpdate.cs
@@ -78,7 +78,7 @@ namespace NHibernate.Tool.hbm2ddl
 						{
 							cfg.SetNamingStrategy(
 								(INamingStrategy)
-								Environment.BytecodeProvider.ObjectsFactory.CreateInstance(ReflectHelper.ClassForName(args[i].Substring(9))));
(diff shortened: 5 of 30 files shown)


hazzik:
When I wrote #1671 I had in mind exact this changes, so I just mark it as fixing #1671.

hazzik:
Actually, we need to add a configuration option to be able to configure the objects factory by xml.

fredericDelaporte:
> Actually, we need to add a configuration option to be able to configure the objects factory by xml.

So, need to add something working the same way than the `bytecode.provider` setting.

> In addition, the `IObjectsFactory` has an overload with the `nonPublic` parameter which is not possible to implement with most of the IoC containers. The overload is currently used in only two places ([AliasToBeanResultTransformer](https://github.com/nhibernate/nhibernate-core/blob/bced0cd2389fe931d885327b630e1ee9fbed4b75/src/NHibernate/Transform/AliasToBeanResultTransformer.cs#L88) and [PocoInstantiator](https://github.com/nhibernate/nhibernate-core/blob/a058d38eee7d23488a377248b903e1b90ef2e554/src/NHibernate/Tuple/PocoInstantiator.cs#L106)) and all tests passes when the second argument is removed. Wouldn't it be better to obsolete that overload and let the implementor decide whether to use a non public constructor?

First, better handle changes for this subject (if any) in another PR I think.

Then, the `AliasToBeanResultTransformer` is a special case. It uses the objects factory only when the type to create is a value type. When it is a class, it manually resolves its parameter-less constructor (public or not). This was maybe done as an optimization, for avoiding resolving the class constructor at each instantiation of a bean. But of course this means the most common case of instantiation does not use the objects factory. Then why not do a small possible breaking change by ceasing using the objects factory for structs too, by inlining its default implementation there?
I think a bean is what is called a "newable" in your [link](https://www.cuttingedge.it/blogs/steven/pivot/entry.php?id=97) you have provided on nhibernate/NHibernate-Caches#45. That is not a thing to consider for dependency injection. (A bean is just a container for a query result element data, when we want to project the data to something else than entities while still having strong typing.)
The `PocoInstantiator` is very similar: it uses the objects factory only when the type is a value type (instead of testing `IsClass` it tests `IsValueType`). And likewise, entities should be considered "newables". So there too, why not dropping the usage of the objects factory too?
(The poco instantiator is used to instantiate entities.)

So in fact, should the objects factory usage be restricted to things for which dependency injection could make sens? I think yes, since there is only the two above cases which use it for "newables", and only when these "newable" are structs.

> Even the overload with the `ctorArgs` is not possible to implement with some IoC containers as it is [considered as a code smell](https://www.cuttingedge.it/blogs/steven/pivot/entry.php?id=99). This overload is also used only two times, one time in [SessionFactoryImpl](https://github.com/nhibernate/nhibernate-core/blob/1024b71aec3b5e75c80e0ad6515943d6aee4adba/src/NHibernate/Impl/SessionFactoryImpl.cs#L1286) which is not that easy to change because of the circular dependency between `ICurrentSessionContext` and `ISessionFactory` and the second in [SchemaExport](https://github.com/nhibernate/nhibernate-core/blob/f8b1bc46f92c7a8cd503d3450f73f6b33ad4e8fb/src/NHibernate/Tool/hbm2ddl/SchemaExport.cs#L207) which is quite strange as the [ScriptSplitter](https://github.com/nhibernate/nhibernate-core/blob/ad4c2ef101cbf8ba798220973f4f78dd795e0896/src/NHibernate/Tool/hbm2ddl/ScriptSplitter.cs#L9) does not have any virtual members.

Again, better address this in another PR. But I think we could obsolete this overload.

The `SessionFactoryImpl` is used to instantiate `ICurrentSessionContext` while providing it with the session factory. We could obsolete that way of building the session context by adding a transitional interface for session context to implement, and allowing to set once the session factory. (Of course ensuring "once" would be implementors responsibility.)

For the `ScriptSplitter`, the coder has surely overlooked what the objects factory is meant for by coding this nonsense. I think we can replace that call by a direct `new` call here. (By the way if there was an interface instead for the script splitter (or at the very least overridable methods), allowing it to be actually injected with a custom implementation, I think that instantiating it with the script to split is a bad design. It should instead have a `Split` method taking the script to handle.)

maca88:
> So, need to add something working the same way than the bytecode.provider setting.

I've added it in the same way as the `bytecode.provider` setting with the exception of adding a property in the `IHibernateConfiguration` interface in order to avoid breaking changes.

I agree with what you said @fredericDelaporte, I have the same thinking about how and when the `IObjectsFactory` should be used.

fredericDelaporte:
> I've added it in the same way as the `bytecode.provider` setting with the exception of adding a property in the `IHibernateConfiguration` interface in order to avoid breaking changes.

Can you add a `// 6.0 TODO add to IHibernateConfiguration` somewhere for this to be done someday?

src/NHibernate/Cfg/Environment.cs
```diff
@@ -381,6 +391,16 @@ public static IBytecodeProvider BytecodeProvider
...
 
+		/// <summary>
+		/// NHibernate's object instantiator.
+		/// </summary>
+		/// <remarks>
+		/// This should only be set before a configuration object
+		/// is created, otherwise the change may not take effect.
+		/// For entities see <see cref="IReflectionOptimizer"/> and its implementations.
+		/// </remarks>
+		public static IObjectsFactory ObjectsFactory { get; set; } = new ActivatorObjectsFactory();
```
fredericDelaporte: To avoid a possible breaking change, we should default to using `BytecodeProvider.ObjectsFactory` here. But for completeness, we should also use the objects factory to create the byte-code provider...

I may add a commit sorting this out.

fredericDelaporte:
I have done maca88/nhibernate-core#2.

fredericDelaporte:
There are definitely some more work to do with current objects factory usages, but anyway I was thinking dedicated PR were better for these related subjects.

PR #1760, opened 2018-06-19
Support formula on one-to-many map-key

fredericDelaporte:
Fix a null reference exception in OneToManyPersister

Spotted while testing #1759


src/NHibernate/Persister/Collection/OneToManyPersister.cs
```diff
@@ -76,7 +76,7 @@ protected override SqlCommandInfo GenerateDeleteString()
 				update.SetJoin(ownerPersister.TableName, KeyColumnNames, KeyType, JoinColumnNames, ownerPersister.GetPropertyColumnNames(CollectionType.LHSPropertyName));
 			}
 
-			if (HasIndex)
+			if (HasIndex && !indexContainsFormula)
```
fredericDelaporte: This is from:
https://github.com/fredericDelaporte/nhibernate-core/blob/ee91396495a5e0874704f22c24ccde7412dafea0/src/NHibernate/Persister/Collection/OneToManyPersister.cs#L134

fredericDelaporte: Yeah sorry for letting my french spelling leaking.

PR #1763, opened 2018-06-20
Support multiple formulas on one-to-one

fredericDelaporte:
Hbm supports multiple formulas on one-to-one, but mapping by code does not allow specifying them on one-to-one.
(It does neither on any other elements, but #1759 will fix that for the other elements by adding a `ColumnsAndFormulas` method.)

### DIFF
--- a/src/NHibernate/Mapping/ByCode/IOneToOneMapper.cs
+++ b/src/NHibernate/Mapping/ByCode/IOneToOneMapper.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Linq.Expressions;
 using System.Reflection;
+using NHibernate.Mapping.ByCode.Impl;
+using NHibernate.Util;
 
 namespace NHibernate.Mapping.ByCode
 {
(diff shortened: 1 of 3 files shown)


src/NHibernate/Mapping/ByCode/IOneToOneMapper.cs
```diff
@@ -19,4 +21,19 @@ public interface IOneToOneMapper<T> : IOneToOneMapper
 	{
 		void PropertyReference<TProperty>(Expression<Func<T, TProperty>> reference);
 	}
-}
\ No newline at end of file
+
+	// 6.0 TODO: move method into IOneToOneMapper
+	public static class OneToOneMapperExtension
```
hazzik: `Extensions` please

hazzik:
Why only `one-to-one`? Others lack of these methods too.

fredericDelaporte:
With #1759, others will have `ColumnsAndFormulas` for doing this. Adding `Formulas` would be just some added convenience, while there is already quite many methods on these interfaces. So I have not considered adding it on others.

PR #1764, opened 2018-06-21
Fix configuration schema forbidding custom bytecode provider

fredericDelaporte:


src/NHibernate.Test/CfgTest/ConfigurationSchemaFixture.cs
```diff
@@ -35,6 +35,23 @@ public void FromAppConfigTest()
 			Assert.IsTrue(hc.UseReflectionOptimizer);
 			Assert.AreEqual("NHibernate.Test", hc.SessionFactory.Name);
 		}
+ 
+		[Test]
+		public void ByteCodeProvider()
```
hazzik: The test one above will fail.

hazzik: I don't understand...

fredericDelaporte: The first test of the fixture still succeed because the configuration is invalid by lack of `session-factory` node. That is not a schema validation error, but a hard-coded check in the parsing.

hazzik: I understood that later, but did not report it here. Can you rename that test to something more meaningful than `InvalidConfig`?

src/NHibernate.Test/CfgTest/ConfigurationSchemaFixture.cs
```diff
@@ -13,7 +11,7 @@ namespace NHibernate.Test.CfgTest
 	public class ConfigurationSchemaFixture
 	{
 		[Test]
-		public void InvalidConfig()
+		public void ThrowIfSessionFactoryIsMissing()
```
hazzik: Strictly speaking `<session-factory />` element is not required. It is required only when config is loaded not from app.config I guess.

fredericDelaporte: Yes, but the name is already long enough I find. `ThrowIfSessionFactoryIsMissingWhenNotFromAppConfig`?

hazzik: Yes, something like that. Or, better: `SessionFactoryIsRequiredWhenConfigurationIsLoadedNotFromAppConfig`

PR #1767, opened 2018-06-21
Allow generic dictionaries for dynamic entities

fredericDelaporte:
Follow-up to #755 

Generic dictionaries could already be used for creating dynamic entities, but not for working with dynamic entities loaded from the database.

Possible breaking change: querying a dynamic entity as a `Hashtable` instead of an `IDictionary` is no more supported.

### DIFF
--- a/src/NHibernate/Tuple/DynamicEntityInstantiator.cs
+++ b/src/NHibernate/Tuple/DynamicEntityInstantiator.cs
@@ -1,0 +1,64 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using NHibernate.Mapping;
+
+namespace NHibernate.Tuple
+{
+	[Serializable]
+	public class DynamicEntityInstantiator : IInstantiator
+	{
+		public const string Key = "$type$";
+
+		private readonly string _entityName;
+		private readonly HashSet<string> _isInstanceEntityNames = new HashSet<string>();
+
+		public DynamicEntityInstantiator(PersistentClass mappingInfo)
+		{
+			_entityName = mappingInfo.EntityName;
+			_isInstanceEntityNames.Add(_entityName);
+			if (mappingInfo.HasSubclasses)
+			{
+				foreach (var subclassInfo in mappingInfo.SubclassClosureIterator)
+					_isInstanceEntityNames.Add(subclassInfo.EntityName);
+			}
+		}
+
+		protected virtual IDictionary<string, object> GenerateMap()
+		{
+			return new Dictionary<string, object>();
+		}
+
+		#region IInstantiator Members
+
+		public object Instantiate(object id)
+		{
+			return Instantiate();
+		}
+
+		public object Instantiate()
+		{
+			var map = GenerateMap();
+			if (_entityName != null)
+			{
+				map[Key] = _entityName;
+			}
+
+			return map;
+		}
+
+		public bool IsInstance(object obj)
+		{
+			if (!(obj is IDictionary<string, object> that))
+				return false;
+			if (_entityName == null)
+				return true;
+
+			var type = (string) that[Key];
+			return type == null || _isInstanceEntityNames.Contains(type);
+
+		}
+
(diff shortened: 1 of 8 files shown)


src/NHibernate/Tuple/DynamicMapInstantiator.cs
```diff
@@ -50,7 +50,7 @@ public object Instantiate()
 
 		protected virtual IDictionary GenerateMap()
 		{
-			return new Hashtable();
+			return new Dictionary<string, object>();
```
hazzik: I think it's better to create a new type `DynamicEntityInstantiator`.

fredericDelaporte: Why not.

src/NHibernate/Tuple/DynamicEntityInstantiator.cs
```diff
@@ -0,0 +1,64 @@
...
+
+		public object Instantiate(object id)
+		{
+			return Instantiate();
+		}
+
+		public object Instantiate()
+		{
+			var map = GenerateMap();
+			if (_entityName != null)
```
hazzik: This is always `true`.

src/NHibernate/Tuple/DynamicEntityInstantiator.cs
```diff
@@ -0,0 +1,64 @@
...
+			}
+
+			return map;
+		}
+
+		public bool IsInstance(object obj)
+		{
+			if (!(obj is IDictionary<string, object> that))
+				return false;
+			if (_entityName == null)
```
hazzik: This is always `false`.

src/NHibernate/Tuple/DynamicEntityInstantiator.cs
```diff
@@ -0,0 +1,64 @@
...
+				return true;
+
+			var type = (string) that[Key];
```
hazzik: That could fail. There is a difference between `IDictionary` and `IDictionary<,>`.

```csharp
var dictionary = new Dictionary<string, object>();
IDictionary a = dictionary;
var x = a["SomeThing"]; // x is null
IDictionary<string, object> b = dictionary;
var y = b["SomeThing"]; // KeyNotFoundException
```

src/NHibernate/Tuple/DynamicEntityInstantiator.cs
```diff
@@ -0,0 +1,64 @@
...
+			return type == null || _isInstanceEntityNames.Contains(type);
```
hazzik: This condition is likely to be incorrect. In my mind: if `type == null` then this dictionary is not instance of the entity type.

src/NHibernate/Tuple/DynamicEntityInstantiator.cs
fredericDelaporte: All NHibernate paths always assign a non-null `EntityName` to `PersistentClass`, but `PersistentClass` does not guarantee it will never be `null`. Now for the purpose here, we do not need a special handling of `null`. If `EntityName` is `null`, the old code would have added it to `_isInstanceEntityNames` nonetheless.

src/NHibernate/Tuple/DynamicEntityInstantiator.cs
fredericDelaporte: Yes my bad, forgot about it.

src/NHibernate/Tuple/DynamicEntityInstantiator.cs
fredericDelaporte: Agreed.

hazzik:
I just realised that dynamic entities are not queryable by Linq.

fredericDelaporte:
This change was done for criteria and HQL, as showcased by the added tests. Does the lack of Linq support have any impact of what should have been done?

hazzik:
No, it’s all fine.

Issue #1766, opened 2018-06-21
Index column gets ignored for inverse list


fubar-coder:
NHibernate creates the INSERT statements, but "forgets" to update the index column. I'll provide a PR for a unit test shortly.

fredericDelaporte:
As seen on #1768, a collection mapped as `inverse` does not maintain anything: neither what it contains, nor the index of what it contains. Both are to be handled on the non-inverse side. For a many-to-one, it may be better to map it with `update="false"` and `insert="false"`, which would allow to keep the collection as non `inverse`.

PR #1761, opened 2018-06-19
Update mapping documentation

fredericDelaporte:
Many mapping properties available in NHibernate are lacking proper
documentation. Document them. (Mainly done by synchronizing with
https://docs.jboss.org/hibernate/orm/3.5/reference/en-US/html/mapping.html
and subsequent chapters of the Hibernate documentation.)

Also fixes inaccuracies and misleading statements.

PR #1768, opened 2018-06-21
Unit test for issue #1766

fubar-coder:
This is a simplified version of NHibernate.Test.ListIndex.ListIndexFixture that uses an inverse list.

This unit test shows that an inverse list causes missing update statements to set the index column.


src/NHibernate.Test/Async/NHSpecificTest/GH1766/Fixture.cs
```diff
@@ -0,0 +1,80 @@
...
+					Items = new List<B>
+					{
+						new B(),
+						new B(),
+					},
+				};
+
+				foreach (var child in parent.Items)
+				{
+					child.A = parent;
```
fredericDelaporte: The index has to be set here, like the parent. When a list is mapped as inverse, no update occurs from it. It may at most cascade the update/save operation if cascade is enabled, but it will not set the appropriate parent and index.

fubar-coder: Ah, ok. So it was a misunderstanding on my side. Would it be possible to update/clarify the documentation for this use case and/or write a unit test that shows this behavior? I did expect the index column to always be set/updated, because it's an integral part of the list.

fredericDelaporte: There is also:

> The non-inverse side is used to save the in-memory representation to the database.

Anyway, a consequent update is also ongoing, awaiting review (see #1761). It notably adds some more examples and guidance, including some about lists.

I do not think we need to add more.

PR #1770, opened 2018-06-23
Make obsolete abstract virtual

fredericDelaporte:
I have realized than obsoleting an abstract member is not very nice to implementors: it forces them to override an obsolete member. Once dropped, their code will no more compile, while the purpose of `Obsolete` is to allow consumers to make required changes for next version and avoid errors due to upgrading.

Instead, we should switch such abstract member to virtual.

This is a kind of follow-up to #1599, since the only obsoleted abstract I have found was introduced there. (I have thought of this subject because I was starting to obsolete abstract members for another subject.)

There is somewhat a similar trouble with obsoleted interface members, but there is nothing we can do for them, and furthermore the implementor can implements them in a way that will not break compilation once they are dropped, contrary to abstract members.


bahusoid:
My only concern is method "discoverability" for new implementors. Yes - you are forced to implement obsolete method but you also see a proper new method to override in obsolete message. With this method virtual it's not clear what should be overridden in the first place.

PR #1771, opened 2018-06-24
Obsolete unused "xml" type methods

fredericDelaporte:
This is a follow up to #579 which has removed most of the xml entity mode partial implementation. NHibernate types were still having methods documented as meant to parse from xml or convert to xml values. Theses methods:

- Were mostly unused.
- Were never handling any xml encoding/decoding, contradicting their documentation.
- When used, are used on plain text string (not xml encoded) or used for getting plain text string.

This PR obsolete unused cases. The remaining used cases are of two kinds: implementation of `IVersionType`, or usage by "opportunity". The former cases are not obsoleted. The later cases are obsoleted.

(This subject was triggered by the discussion on #1765.)


src/NHibernate/Type/IVersionType.cs
```diff
@@ -28,6 +28,16 @@ public partial interface IVersionType : IType
 		/// </summary>
 		IComparer Comparator { get; }
 
+		// 6.0 TODO: rename "xml" parameter as "value": it is not a xml string. The fact it generally comes from a xml
+		// attribute value is irrelevant to the method behavior.
+		/// <summary>
```
fredericDelaporte: The value is de-serialized by .Net, it not a xml encoded value.

src/NHibernate/Type/IVersionType.cs
```diff
@@ -28,6 +28,16 @@ public partial interface IVersionType : IType
...
+		/// Parse the string representation of a value to convert it to the .NET object.
+		/// </summary>
+		/// <param name="xml">A string representation.</param>
+		/// <returns>The value.</returns>
+		/// <remarks>Notably meant for parsing <c>unsave-value</c> mapping attribute value. Contrary to what could
+		/// be expected due to its current name, <paramref name="xml"/> must be a plain string, not a xml encoded
+		/// string.</remarks>
 		object FromStringValue(string xml);
```
fredericDelaporte: This method uses the same name than `NullableType.FromStringValue`, which has no more usages and is obsoleted. This causes `IVersionType` types to override an obsoleted method with a non-obsoleted one.

It has the same purpose than `IIdentifierType.StringToObject` indeed. It would have been better to have the same name.

src/NHibernate/Type/IIdentifierType.cs
```diff
@@ -5,17 +5,21 @@ namespace NHibernate.Type
 	/// </summary>
 	public interface IIdentifierType : IType
 	{
+		// 6.0 TODO: rename "xml" parameter as "value": it is not a xml string. The fact it generally comes from a xml
+		// attribute value is irrelevant to the method behavior.
```
fredericDelaporte: The value is de-serialized by .Net, it not a xml encoded value.

src/NHibernate/Type/NullableType.cs
```diff
@@ -82,38 +82,40 @@ protected NullableType(SqlType sqlType)
...
 		/// <inheritdoc />
-		/// <remarks>
-		/// <para>
-		/// This implementation forwards the call to <see cref="ToString(object)"/> if the parameter 
-		/// value is not null.
-		/// </para>
-		/// <para>
-		/// It has been "sealed" because the Types inheriting from <see cref="NullableType"/>
-		/// do not need and should not override this method.  All of their implementation
-		/// should be in <see cref="ToString(object)"/>.
```
fredericDelaporte: So a thing meant to log plain text is implemented by calling a thing documented as meant to generate xml strings...

src/NHibernate/UserTypes/IEnhancedUserType.cs
```diff
@@ -7,9 +9,14 @@ namespace NHibernate.UserTypes
...
-		/// in an XML document.
+		/// Parse a string representation of this value.
 		/// </summary>
+		// 6.0 TODO: rename "FromString(string value)".
+		// Since 5.2
+		[Obsolete("This method was not used for parsing xml strings, but instead used for parsing already de-encoded " +
+			"strings originating from xml (so indeed, plain text strings). It will be renamed \"FromString\" in a " +
+			"future version. Implement a \"object FromString(string value)\" without any xml de-coding, called by your" +
+			"\"object FromXMLString(string xml)\", in order to get ready for the rename.")]
 		object FromXMLString(string xml);
```
fredericDelaporte: The supplied `xml` string is never xml encoded, contrary to what we could expect with such naming and documentation. So I have considered a name change is required there. I do not see a better pattern for renaming an interface method than the one showcased here.

src/NHibernate/UserTypes/IEnhancedUserType.cs
```diff
@@ -18,9 +25,13 @@ public interface IEnhancedUserType : IUserType
...
-		/// Return a string representation of this value, as it
-		/// should appear in an XML document
+		/// Return a string representation of this value. It does not need to be xml encoded.
 		/// </summary>
+		// 6.0 TODO: rename "ToString".
+		// Since 5.2
+		[Obsolete("This method was not used in a xml context, but instead just used for logs. It will be renamed " +
+			"\"ToString\" in a future version. Implement a \"string ToString(object value)\", called by your " +
+			"\"string ToXMLString(object value)\", in order to get ready for the rename.")]
 		string ToXMLString(object value);
```
fredericDelaporte: This one is less prone to incite users to code bugs: since it is used only for logs, it will not be a serious issue if the user xml-encode the value. But for consistency, renaming too.

fredericDelaporte:
Rebased for solving conflicts with #1765.

PR #1773, opened 2018-06-24
Obsolete unused version related methods of SByteType

fredericDelaporte:
`SByteType` seems to have implemented `IVersionType` previously, but that is
no more the case.


hazzik:
Need more descriptive  PR name.

PR #1775, opened 2018-06-26
Start/Stop required db-service for TeamCity

fredericDelaporte:
Since the TeamCity build agent has occasional memory issues, mainly due to some database services taking more and more memory as they accumulate up-time (and run testes), I propose to start and stop the required database service for each build.

This change is not enabled by default in order to avoid disrupting other builds for now. It will be enabled by adding a `-D:skip.db-service=false` parameter to the Nant step on the build template.

If approved, I think it needs to be done in 5.1.x then merged to master, in order to not hinder any 5.1.x patch. It will have to be back-ported to previous branches if patches are required for them.

### DIFF
--- a/teamcity.build
+++ b/teamcity.build
@@ -4,6 +4,7 @@
 	<property name="root.dir" value="." />
 	<property name="config.teamcity" value="default"/>
 	<property name="current-test-configuration.dir" value="${root.dir}/current-test-configuration" /> 
+	<property name="skip.db-service" value="true" overwrite="false" />
 
 	<include buildfile="${root.dir}/default.build" />
 
@@ -13,7 +14,15 @@
 
 	<property name="build.number" value="${CCNetLabel}" if="${property::exists('CCNetLabel')}" />
 
-	<target name="clean-configure-test" depends="cleanall init copy-teamcity-configuration binaries test verify-test-results binaries-zip" />
+	<target name="clean-configure-test" depends="cleanall init copy-teamcity-configuration binaries start-db-service test stop-db-service verify-test-results binaries-zip" />
+
+	<target name="start-db-service" unless="${skip.db-service or not property::exists('db-service')}">
+		<servicecontroller action="Start" service="${db-service}" timeout="120000" />
+	</target>
+
+	<target name="stop-db-service" unless="${skip.db-service or not property::exists('db-service')}">
+		<servicecontroller action="Stop" service="${db-service}" timeout="120000" />
+	</target>
 
 	<target name="copy-teamcity-configuration">
 		<copy file="build-common/teamcity-hibernate.cfg.xml" tofile="${current-test-configuration.dir}/hibernate.cfg.xml" />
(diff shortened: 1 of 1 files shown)


teamcity.build
```diff
@@ -13,7 +14,15 @@
 
 	<property name="build.number" value="${CCNetLabel}" if="${property::exists('CCNetLabel')}" />
 
-	<target name="clean-configure-test" depends="cleanall init copy-teamcity-configuration binaries test verify-test-results binaries-zip" />
+	<target name="clean-configure-test" depends="cleanall init copy-teamcity-configuration binaries start-db-service test stop-db-service verify-test-results binaries-zip" />
```
fredericDelaporte: This way of doing things will cause the service to stay running if the test step fails. It is not ideal, but it should not cause much issues, provided some other PR succeeds and still ends up stopping the service. (Starting an already started service does not fail.)

I have looked for using the `trycatch` Nant node, but have found out it was requiring many additional changes, so I have not considered it worth of being done:
- Needs upgrading `Nant` to latest version (0.92, we use 0.91)
- Needs rewriting the task as
```xml
<target name="clean-configure-test">
  <trycatch>
    <try>
      <!-- Use a sub-target doing it all instead of putting test prerequisites in
             depends attribute the calling just test target here, otherwise build is
             run twice -->
      <call target="do-clean-configure-test" />
    </try>
    <finally>
      <!-- Use failonerror="false" in order to avoid hiding a failure from the try -->
      <call target="stop-db-service" failonerror="false" />
    </finally>
  </trycatch>
</target>

<target name="do-clean-configure-test" depends="cleanall init copy-teamcity-configuration binaries start-db-service test verify-test-results binaries-zip" />
```
- Needs switching `<xmlpoke>` tasks to `failonerror="false"` due to nant/nant#44 fixed in 0.92, combined with the `test` target calling it on a file missing most of the nodes targeted by the `<xmlpoke>`.

teamcity.build
```diff
@@ -4,6 +4,7 @@
 	<property name="root.dir" value="." />
 	<property name="config.teamcity" value="default"/>
 	<property name="current-test-configuration.dir" value="${root.dir}/current-test-configuration" /> 
+	<property name="skip.db-service" value="true" overwrite="false" />
```
fredericDelaporte: Another way of enabling the "Start/Stop" change would be to directly change the default value here to `false`.
But that looks less convenient to me. Having to enable this in the build template instead would allow to wait for having this change merged in most branches first. Then enabling it would not cause any of those branches to fail building.

While on the contrary, enabling it by source code change will cause it to make all other branches failing until they get the source code change.

Changing its default value here would likely call for more changes in `ShowBuildMenu.bat` too: personally I do not have the same services name than the TeamCity build agent. So for the TeamCity builds to run through the menu, `ShowBuildMenu.bat` would have to provide this property as `true`.

teamcity.build
```diff
@@ -13,7 +14,15 @@
 
 	<property name="build.number" value="${CCNetLabel}" if="${property::exists('CCNetLabel')}" />
 
-	<target name="clean-configure-test" depends="cleanall init copy-teamcity-configuration binaries test verify-test-results binaries-zip" />
+	<target name="clean-configure-test" depends="cleanall init copy-teamcity-configuration binaries start-db-service test stop-db-service verify-test-results binaries-zip" />
+
+	<target name="start-db-service" unless="${skip.db-service or not property::exists('db-service')}">
+		<servicecontroller action="Start" service="${db-service}" timeout="120000" />
```
fredericDelaporte: On my local setup, Oracle takes up to one minute to start. So I have put a timeout of two minutes. (The default is five seconds.)

This of course means slowing down the builds a bit, especially for databases slow to start and stop.

We may dodge this by keeping a short timeout along with using `failonerror="false"` to ignore the timeout, and moving the service start before the build in order to rely on the build time (a bit more than 3 minutes on the TeamCity agent) for having the database ready. Similarly we could do a short ignored timeout that on the "stop service" target too, relying on the delay between two builds for having the service actually stopped before an attempt to restart it is done. (The source copy step maybe gives enough delay.)

But that could give flaky failures, so I think it is better to just wait.

fredericDelaporte:
Merged to master through client command line.

PR #1777, opened 2018-06-27
Replace ICache interface by a CacheBase class

fredericDelaporte:
In order to ease adding cache features, and provide default implementations for them, use an abstract base class instead of an interface.

This is similar to `DbConnection`, `DbCommand` and the like pattern.

It allows to reduce the needs for "transitional interfaces" and simplifies implementing new features as it allows to abstract away whether the implementation is obsolete or not, actually supports the new features or not.


src/NHibernate/Cache/CacheBase.cs
```diff
@@ -0,0 +1,386 @@
...
+		/// collections from the cache instead of only getting the proxy or collection which initialization
+		/// is asked for. If this cache implementation does not benefit from batching together get operations,
+		/// this may result in a performance loss.
+		/// </para>
+		/// <para>
+		/// When <see langword="false" />, NHibernate will still call <see cref="GetMany" /> when it has many
+		/// gets to perform. Its <see cref="CacheBase" /> default implementation is adequate for this case.
+		/// </para>
+		/// </remarks>
+		public abstract bool PreferMultipleGet { get; }
```
fredericDelaporte: When the cache does not benefit from multiple gets, NHibernate should not attempt on a single get to pro-actively get up to `batch-size` entries from the cache. Otherwise it would incurs an overhead without any benefit. This flag allows to tell NHibernate if it should pro-actively batch single gets from cache. When doing so, NHibernate may load unneeded data from the cache (the same way `batch-size` can cause NHibernate to load unneeded data from the database), so it is important to avoid doing this for caches not benefiting of multiple gets.

A cache implementation supporting multiple gets could also choose to let users configure this, if the user does not wish the `batch-size` mapping attribute to be applied to fetches from cache.

We do not need a similar flag for put or lock cases, since in these cases NHibernate do no guesses about which items are to be handled. It actually have to cache/lock them all.

src/NHibernate/Cache/CacheFactory.cs
```diff
@@ -67,16 +67,14 @@ public static class CacheFactory
 						"cache usage attribute should be read-write, read-only, nonstrict-read-write, or transactional");
 			}
 
-			ICache impl;
 			try
 			{
-				impl = settings.CacheProvider.BuildCache(name, properties);
+				ccs.Cache = settings.CacheProvider.BuildCache(name, properties);
```
fredericDelaporte: This property setter has no logic, doing the assignment outside of the `try` was just complicating the code for no reasons.

src/NHibernate/Cache/CacheFactory.cs
```diff
@@ -67,16 +67,14 @@ public static class CacheFactory
...
 			}
 			catch (CacheException e)
```
fredericDelaporte: Anyway I do not know of any cache implementation throwing this exception... Those from NHibernate-Caches never thrown it.

src/NHibernate/Cache/HashtableCacheProvider.cs
```diff
@@ -10,7 +11,14 @@ public class HashtableCacheProvider : ICacheProvider
...
 
-		public ICache BuildCache(string regionName, IDictionary<string, string> properties)
+		// Since 5.2
+		[Obsolete]
+		ICache ICacheProvider.BuildCache(string regionName, IDictionary<string, string> properties)
+		{
+			return BuildCache(regionName, properties);
+		}
+
+		public CacheBase BuildCache(string regionName, IDictionary<string, string> properties)
```
fredericDelaporte: This implementation of `BuildCache` allows to get the provider almost ready for the switch from `ICache` to `CacheBase`. It will only have to drop the explicit implementation.

src/NHibernate/Cache/IBatchableCache.cs
```diff
@@ -1,41 +0,0 @@
...
-	/// The implementor should use this interface along with <see cref="ICache"/> when the
-	/// cache supports a multiple get and put operation.
-	/// </summary>
-	/// <remarks>
-	/// <threadsafety instance="true" />
-	/// <para>
-	/// All implementations <em>must</em> be threadsafe.
-	/// </para>
-	/// </remarks>
-	public partial interface IBatchableCache : IBatchableReadOnlyCache
```
fredericDelaporte: Directly added to `CacheBase` instead. (And we can drop this interface as it is not yet released.)

hazzik:
This needs to be fixed across the change-set

>// 6.0 TODO: type as BaseCache instead

src/NHibernate/Cache/ICacheConcurrencyStrategy.cs
```diff
@@ -134,13 +135,17 @@ public partial interface ICacheConcurrencyStrategy
 		/// </summary>
 		string RegionName { get; }
 
+		// 6.0 TODO: type as BaseCache instead
+#pragma warning disable 618
```
fredericDelaporte: Implementors will be warned by the `ICache` obsolete message about that, but they will not be able to actually prepare for it. I do not see a way to avoid this. The `HashtableCacheProvider.BuildCache` implementation pattern commented previously is not usable here, as we may have a cache not deriving from `CacheBase` here.

src/NHibernate/Cache/ICacheConcurrencyStrategy.cs
```diff
@@ -153,11 +158,11 @@ internal static partial class CacheConcurrencyStrategyExtensions
...
 		{
-			if (!(cache is IBatchableCacheConcurrencyStrategy batchableCache))
-			{
-				throw new InvalidOperationException($"Cache concurrency strategy {cache.GetType()} does not support batching");
-			}
-			return batchableCache.GetMany(keys, timestamp);
+			// PreferMultipleGet yields false if !IBatchableCacheConcurrencyStrategy, no GetMany call should be done
+			// in such case.
+			return ReflectHelper
+				.CastOrThrow<IBatchableCacheConcurrencyStrategy>(cache, "batching")
```
fredericDelaporte: This `GetMany` is currently called only in case of lazy batch fetches (using `batch-size`). The cases where NHibernate calls `GetMany` even if `PreferMultipleGet` is false do not use this `GetMany`.

src/NHibernate/Cache/ICacheConcurrencyStrategy.cs
```diff
@@ -175,23 +180,37 @@ public static object[] GetMany(this ICacheConcurrencyStrategy cache, CacheKey[]
...
+		internal static CacheBase GetCacheBase(this ICacheConcurrencyStrategy cache)
 		{
-			// ReSharper disable once SuspiciousTypeConversion.Global
-			return cache.Cache is IBatchableCache && cache is IBatchableCacheConcurrencyStrategy;
+			if (cache is IBatchableCacheConcurrencyStrategy batchableCache)
+				return batchableCache.CacheBase;
+			var concreteCache = cache.Cache;
+			if (concreteCache == null)
+				return null;
+			return concreteCache as CacheBase ?? new ObsoleteCacheWrapper(concreteCache);
```
fredericDelaporte: This allows a simplification, by always yielding a `CacheBase` provided there is a cache, even if it does not implement itself `CacheBase`.
In most cases it will yield the wrapper held by the cache concurrency, avoiding the instantiation here. (The instantiation will occur only if the user uses a custom cache concurrency, which looks very unlikely to me, since the mapping does not allow specifying a custom implementation.)

src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -201,21 +198,20 @@ public ISoftLock Lock(CacheKey key, object version)
 					log.Debug("Invalidating: {0}", key);
 				}
 
+				var lockValue = _cacheBase.Lock(key);
 				try
 				{
-					cache.Lock(key);
```
fredericDelaporte: The `finally` is done for unlocking, but there is nothing to unlock if the locking has failed. This line should have been before the `try`, even without having the `lockValue` to handle. All similar cases have been changed likewise.

src/NHibernate/Cache/IBatchableCacheConcurrencyStrategy.cs
```diff
@@ -19,29 +19,34 @@ namespace NHibernate.Cache
...
+		/// <param name="versionComparers">The comparers to be used to compare version numbers.</param>
 		/// <param name="minimalPuts">Indicates that the cache should avoid a put if the item is already cached.</param>
-		/// <returns><see langword="true" /> if the objects were successfully cached.</returns>
+		/// <returns>An array of boolean indicating if each item was successfully cached.</returns>
 		/// <exception cref="CacheException"></exception>
 		bool[] PutMany(CacheKey[] keys, object[] values, long timestamp, object[] versions, IComparer[] versionComparers,
 								  bool[] minimalPuts);
+
+		// 6.0 TODO: remove for using ICacheConcurrencyStrategy.Cache re-typed CacheBase instead
+		CacheBase CacheBase { get; }
```
fredericDelaporte: We can change this interface as it is not yet released.

This property is somewhat obsolete as soon as added. But it should not bother many implementors, since mapping does not allow to specify custom cache concurrency implementations.

fredericDelaporte:
>The following comment needs to be fixed across the change-set (It should be CacheBase instead of BaseCache)
>
>>// 6.0 TODO: type as BaseCache instead

Done, commit reworded, amended and forced push (since that was the only change, I think it was not deserving a separated fix commit).

src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -506,10 +505,10 @@ bool ProcessKey(EntityKey key, bool ignoreCache = false)
...
 
-					if (!checkCache || !IsCached(key, persister))
-					{
-						ids[i++] = key.Identifier;
-					}
+					// No need to check "!checkCache || !IsCached(key, persister)": "batchableCache == null"
+					// already means there is no cache, so IsCached can only yield false. (This method is now
+					// removed.)
+					ids[i++] = key.Identifier;
 				}
```
maca88: The same correction should also be done in `GetCollectionBatch` method as it has the same logic.

fredericDelaporte: Fixed.

src/NHibernate/Cache/CacheBase.cs
hazzik: I think this needs to be `virtual` and yield `false` by default.

src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -420,7 +418,8 @@ internal object[] GetEntityBatch(IEntityPersister persister, object id, int batc
 			// List of entity keys that haven't been checked for their existance in the cache. Besides the entity key,
 			// the index where the key was found is also stored in order to correctly order the returning keys.
 			var entityKeys = new List<KeyValuePair<EntityKey, int>>(batchSize);
-			var batchableCache = persister.Cache?.Cache as IBatchableReadOnlyCache;
+			// If there is a cache, obsolete or not, batchableCache will not be null.
+			var batchableCache = persister.Cache?.GetCacheBase();
```
hazzik: In line 238 you use as CacheBase. But here is GetCacheBase? Is it intentional?

src/NHibernate/Cache/CacheBase.cs
fredericDelaporte: I have chosen to have it abstract for avoiding it to be overlooked by implementors. Why not defaulting to `false`, but then I will surely add some notes in `GetMany` xml documentation for hinting at caring about this property.

src/NHibernate/Engine/BatchFetchQueue.cs
fredericDelaporte: No, line 238 is an error and causes the last fixup to introduce a potential bug causing cache to be ignored in some cases, thanks for having spotted it.

fredericDelaporte:
Rebased and squashed.

src/NHibernate/Cache/IBatchableCacheConcurrencyStrategy.cs
hazzik: I think this needs to be renamed to just `Cache`.

src/NHibernate/Cache/ReadOnlyCache.cs
```diff
@@ -11,34 +11,39 @@ namespace NHibernate.Cache
...
 		/// </summary>
 		public string RegionName
 		{
-			get { return cache.RegionName; }
+			get { return Cache.RegionName; }
 		}
 
+		// 6.0 TODO: type as CacheBase instead and switch to autoproperty
+#pragma warning disable 618
 		public ICache Cache
```
hazzik: I thin this needs to be made explicit.

src/NHibernate/Cache/IBatchableCacheConcurrencyStrategy.cs
hazzik: This way it would allow to make less changes.

src/NHibernate/Cache/ReadOnlyCache.cs
```diff
@@ -11,34 +11,39 @@ namespace NHibernate.Cache
...
+#pragma warning restore 618
 		{
 			get { return cache; }
 			set
 			{
 				cache = value;
-				// ReSharper disable once SuspiciousTypeConversion.Global
-				_batchableReadOnlyCache = value as IBatchableReadOnlyCache;
-				_batchableCache = value as IBatchableCache;
+				_batchableCache = value as CacheBase ?? new ObsoleteCacheWrapper(value);
```
hazzik: Is there any value in having 2 fields to hold a cache and a wrapper?

fredericDelaporte: It avoids to get through the wrapper for cases not requiring it, but it is likely to get inlined by compilation optimizations. So not really.

src/NHibernate/Cache/ReadOnlyCache.cs
fredericDelaporte: It would be a binary breaking change. Setting an `ReadOnlyCache.Cache` with an `ICache` not deriving from `CacheBase` would no more compile.

src/NHibernate/Cache/IBatchableCacheConcurrencyStrategy.cs
fredericDelaporte: And made explicit for now, since we cannot do that to the former property without introducing a binary breaking change.

src/NHibernate/Cache/ReadOnlyCache.cs
hazzik: So, are we talking about a binary breaking change (drop-in replacement) or a source breaking change (when it does not compile anymore)? I don't think it will create a binary breaking change.

fredericDelaporte: It is a binary breaking change for anyone using the concrete type, which is public. Of course no one is supposed to use that concrete type directly. But for ensuring this it should have been internal.
Unfortunately it is not, so that is still a binary breaking change. It will go unnoticed by most users, but maybe not by some.

hazzik:
One note: the `XXXBase` pattern is not found anywhere in the code base. Instead NH is tend to use Java borrowed pattern of "AbstractXXX", not sure what I do prefer though.

fredericDelaporte:
> One note: the `XXXBase` pattern is not found anywhere in the code base. Instead NH is tend to use Java borrowed pattern of `AbstractXXX`, not sure what I do prefer though.

83 matches for `class Abstract` in NHibernate project, 24 for `class [a-zA-Z0-9_]+Base[^a-zA-Z0-9_]`. So the Java pattern clearly dominates, but the .Net one is already there for around 20% of cases.

For the cache case, I dislike a bit both patterns indeed. I thought of naming it just `Cache`. After all `DbCommand` does not include `Base` in its name either. But the `Db` ones are well known classes. The purpose of a `Cache` class would be less obvious. So I have kept the `Base` suffix.

fredericDelaporte:
Squashed and rebased for resolving conflicts with #1788.

fredericDelaporte:
Rebased, and one additional adjustment suggested by @hazzik has been made.

PR #1778, opened 2018-06-28
Allow dynamic in dynamic component

hazzik:
Sorry for the tautology


fredericDelaporte:
This also allows following test, added to `NHibernate.Test.EntityModeTest.Map.Basic.DynamicClassFixture`, to succeed:

```c#
[Test]
public void ShouldWorkWithLinq()
{
	TestLazyDynamicClass(
		s => (IDictionary<string, object>) s.Query<dynamic>("ProductLine").OrderBy("Description").Single(),
		s => s.Query<dynamic>("Model").ToList().Cast<IDictionary<string, object>>().ToList());
}
```

Maybe the `DynamicComponent` could even be used in the dynamic entity instantiator for avoiding casting back to `IDictionary`. I am going to check that, but then it would need another name.

src/NHibernate/Util/DynamicDictionaryWrapper.cs
```diff
@@ -0,0 +1,215 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Dynamic;
+using System.Runtime.Serialization;
+
+namespace NHibernate.Util
+{
+	[Serializable]
+	public sealed class DynamicComponent : DynamicObject,
```
fredericDelaporte: File should be renamed.

fredericDelaporte:
> Maybe the DynamicComponent could even be used in the dynamic entity instantiator for avoiding casting back to IDictionary. I am going to check that, but then it would need another name.

It works.

About the name, no idea for a better name (since `DynamicObject` is already taken, unless we do not care for the potential confusion with the inherited one).

Now it does not fully works, `MapProxy` needs to be adapted too for proxies to support being used as dynamics too.

fredericDelaporte:
Commit added for updating the documentation.

hazzik:
@fredericDelaporte I think the dynamic entities need to go into a separate PR. The dynamic entities have different scope and level of support (for example you cannot query Dynamic Entity by Linq except of simple `Query<dynamic>("Entity").ToList()`)

fredericDelaporte:
Check again my examples. `Query<IDictionary>("Entity").OrderBy("Name") .ToList()` works, so why `Query<IDictionary>("Entity").Where("Name = @0", "Alexander") .ToList()` would not? The changes you have made for components work for them too. Anyway dynamic-component and dynamic-entities are two part of the same need, mapping things dynamically. I do not really agree it should be considered different subjects.

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -48,7 +48,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="log4net" Version="2.0.8" />
-    <PackageReference Include="System.Linq.Dynamic.Core" Version="1.0.8" />
+    <PackageReference Include="System.Linq.Dynamic.Core" Version="1.0.8.11" />
```
hazzik: The 1.0.8 version did not work on mono/.NET Core due to this bug: https://github.com/StefH/System.Linq.Dynamic.Core/issues/157. So updated to latest.

hazzik:
I'm talking about "static" Linq (the similar detection of dictionary is required for dynamic entities as `VisitorUtil.IsDynamicComponentDictionaryGetter`).

fredericDelaporte:
You mean, for handling the case of navigating associations? I have not checked that case, granted.

hazzik:
>Check again my examples. `Query<IDictionary>("Entity").OrderBy("Name") .ToList()` works

But in your example it is `Query<dynamic>("Entity").OrderBy("Name") .ToList()` (the `dynamic` instead of `IDictionary`).

> why Query<IDictionary>("Entity").Where("Name = @0", "Alexander").ToList() would not?

Because of either: a) it will fail to find a property "Name" on `IDictionary`; b) when changed to `.Where("[\"Name\"] = @0", "Alexander")` it would fail because there is no conversion of indexer to property (see my previous comment).

hazzik:
Ok, I've checked this:

```csharp
[Test]
public void ShouldWorkWithLinqAndGenerics()
{
	TestLazyDynamicClass(
		s => s.Query<IDictionary<string, object>>("ProductLine").OrderBy("Description").Single(),
		s => s.Query<IDictionary<string, object>>("Model").ToList().ToList());
}
```

It fails:
```
System.InvalidOperationException : Operation is not valid due to the current state of the object.
  at NHibernate.Hql.Ast.ANTLR.Tree.IdentNode.ResolveIndex (NHibernate.Hql.Ast.ANTLR.Tree.IASTNode parent) [0x000d4] in /Users/hazzik/Projects/nhibernate-core/src/NHibernate/Hql/Ast/ANTLR/Tree/IdentNode.cs:109 
```

Turns out the dynamic Linq is clever enough to convert "Name" to dictionary indexer.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -206,6 +222,18 @@ private HqlTreeNode VisitNhNominated(NhNominatedExpression nhNominatedExpression
 
 		private HqlTreeNode VisitInvocationExpression(InvocationExpression expression)
 		{
+			//This is an ugly workaround for dynamic expressions.
```
hazzik: In .NET Core/Mono the `DynamicExpression` is "reduced" to a CallSite delegate call, but on full .NET it does not happen.

hazzik:
Regardless of the views on relations with `dynamic-entity` vs `dynamic-component` can we agree to move with the different PRs for them?

hazzik:
I wonder. The System.Linq.Dynamic supports only `GetMemberBinder`. Do we need to carry about other binder types?

fredericDelaporte:
I am not knowledgeable enough in that area for telling. Would it allow to better support querying a component sub-collection? (I mean, if a dynamic component owns a collection, currently trying to apply to it an `.Any` condition will fail, the dynamic expression engine being not able to recognize it as a collection.)

fredericDelaporte:
> I wonder. The System.Linq.Dynamic.Core supports only `GetMemberBinder`. Do we need to care about other binder types?

I have taken a look at it. The other types are `GetIndexBinder`, `SetMemberBinder` and `SetIndexBinder`, right?

Supporting `GetIndexBinder` should allow to support dictionary syntax in dynamic Linq. Why not but I do not see much value in it. The dictionary syntax is quite ugly. Maybe we should even obsolete it, now that dynamic is supported.

About the setters, maybe it could make sens to support them for NHibernate linq `update` and `insert` extensions. But since they already supports anonymous objects, I do not think we really need to do that.

hazzik:
>I have taken a look at it. The other types are GetIndexBinder, SetMemberBinder and SetIndexBinder, right?

Yes, these are the ones which potentially interest us. Others are not so interested.

> The later could maybe allow to support the `Any` case.

Again, this is not supported by the "dynamic" provider. 

Also, I could not find any provider which does support/use `System.Linq.Expressions.DynamicExpression` "in the wild" (except the System.Linq.Dynamic.Core which has very limited use of it and only emits `GetMemberBinder`).

We could write tests which construct the desired expressions and add support for them, but do we wont to do this?

fredericDelaporte:
Without concrete use cases, no.

hazzik:
Just force-pushed to be inline with the master. No changes.

hazzik:
Related #947

fredericDelaporte:
I have locally checked the rebase, no trouble, no conflict to resolve. It also seems the two commits of this PR could be directly merged (with merge commit) without a rebase. Maybe were you thinking about #1452, on which I have the same question, and that I consider merging very soon.

hazzik:
Rebased with fix.

hazzik:
I intend to merge it as soon as test pass, please do not merge anything to master before this PR is merged.

fredericDelaporte:
It seems to me this new feature of mapping a dynamic component as a C# `dynamic` is not available for Mapping by code. Am I wrong?

fredericDelaporte:
I am. `dynamic` is anyway not supported by lambdas used by By-Code. It can only be mapped using the property name, and this one works.

PR #1781, opened 2018-07-03
Clean-up IObjectsFactory usages

fredericDelaporte:
`IObjectsFactory` is meant for instantiating NHibernate dependencies, but it was used in some cases for instantiating value types. Some of its methods are also hard or impossible to implement with many dependency injection frameworks.

Follow up to #1758 

Possible breaking change:
- Users providing some custom logic for instantiating value types will now need to supply their own result transformer if they were using `AliasToBeanResultTransformer` with value types, or their own entity tuplizer if they were using value types as entities.
- Users providing some custom logic for instantiating their custom session contexts will have to implement `ICurrentSessionContextWithFactory` and add a parameterless public constructor to their custom context, and move their custom instantiation logic from `IObjectsFactory.CreateInstance(Type, object[])` to `IObjectsFactory.CreateInstance(Type)`.

### DIFF
--- a/src/NHibernate/Context/ICurrentSessionContext.cs
+++ b/src/NHibernate/Context/ICurrentSessionContext.cs
@@ -1,3 +1,4 @@
+using NHibernate.Bytecode;
 using NHibernate.Engine;
 
 namespace NHibernate.Context
@@ -11,7 +12,8 @@ namespace NHibernate.Context
 	/// Implementations should adhere to the following:
 	/// <list type="bullet">
 	/// <item><description>contain a constructor accepting a single argument of type
-	/// <see cref="ISessionFactoryImplementor" /></description></item>
+	/// <see cref="ISessionFactoryImplementor" />, or implement
+	/// <see cref="ICurrentSessionContextWithFactory"/></description></item>
 	/// <item><description>should be thread safe</description></item>
 	/// <item><description>should be fully serializable</description></item>
 	/// </list>
@@ -42,9 +44,10 @@ namespace NHibernate.Context
 		ISession CurrentSession();
 	}
 
-	// 6.0 TODO: merge into ICurrentSessionContext
 	/// <summary>
-	/// Transitional interface for <see cref="ICurrentSessionContext"/>.
+	/// An <see cref="ICurrentSessionContext"/> allowing to set its session factory. Implementing
+	/// this interface allows the <see cref="IObjectsFactory"/> to be used for instantiating the
+	/// session context.
 	/// </summary>
 	public interface ICurrentSessionContextWithFactory : ICurrentSessionContext
 	{
--- a/src/NHibernate/Context/ICurrentSessionContext.cs
+++ b/src/NHibernate/Context/ICurrentSessionContext.cs
@@ -1,4 +1,3 @@
-using System;
 using NHibernate.Engine;
 
 namespace NHibernate.Context
@@ -41,5 +40,18 @@ namespace NHibernate.Context
 		/// <exception cref="HibernateException">Typically indicates an issue
 		/// locating or creating the current session.</exception>
 		ISession CurrentSession();
+	}
+
+	// 6.0 TODO: merge into ICurrentSessionContext
+	/// <summary>
+	/// Transitional interface for <see cref="ICurrentSessionContext"/>.
+	/// </summary>
+	public interface ICurrentSessionContextWithFactory : ICurrentSessionContext
+	{
+		/// <summary>
+		/// Sets the factory. This method should be called once after creating the context.
+		/// </summary>
+		/// <param name="factory">The factory.</param>
+		void SetFactory(ISessionFactoryImplementor factory);
 	}
 }
--- a/src/NHibernate/Context/ICurrentSessionContext.cs
+++ b/src/NHibernate/Context/ICurrentSessionContext.cs
@@ -13,7 +13,7 @@ namespace NHibernate.Context
 	/// <list type="bullet">
 	/// <item><description>contain a constructor accepting a single argument of type
 	/// <see cref="ISessionFactoryImplementor" />, or implement
-	/// <see cref="ICurrentSessionContextWithFactory"/></description></item>
+	/// <see cref="ISessionFactoryAwareCurrentSessionContext"/></description></item>
 	/// <item><description>should be thread safe</description></item>
 	/// <item><description>should be fully serializable</description></item>
 	/// </list>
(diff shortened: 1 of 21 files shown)


src/NHibernate/Tool/hbm2ddl/SchemaExport.cs
```diff
@@ -203,10 +203,7 @@ private void ExecuteSql(DbCommand cmd, string sql)
 		{
 			if (dialect.SupportsSqlBatches)
 			{
-				var objFactory = Environment.ObjectsFactory;
-				ScriptSplitter splitter = (ScriptSplitter)objFactory.CreateInstance(typeof(ScriptSplitter), sql);
-
-				foreach (string stmt in splitter)
+				foreach (var stmt in new ScriptSplitter(sql))
```
fredericDelaporte: As discussed on #1758, the `ScriptSplitter` does not expose anything overridable, and so it cannot benefit from being instantiated by the object factory: it cannot supply a custom implementation.

src/NHibernate/Transform/AliasToBeanResultTransformer.cs
```diff
@@ -85,7 +85,7 @@ public override object TransformTuple(object[] tuple, String[] aliases)
 			{
 				result = _resultClass.IsClass
 							? _beanConstructor.Invoke(null)
-							: Cfg.Environment.ObjectsFactory.CreateInstance(_resultClass, true);
+							: Activator.CreateInstance(_resultClass, true);
```
fredericDelaporte: As discussed in #1758, this usage is not a dependency but a DTO instantiation. If the user wants some custom behavior for instantiating value types, he now would have to provide its own transformer.

src/NHibernate/Tuple/PocoInstantiator.cs
```diff
@@ -103,7 +103,7 @@ private object GetInstance()
 			}
 			if (mappedClass.IsValueType)
 			{
-				return Cfg.Environment.ObjectsFactory.CreateInstance(mappedClass, true);
+				return Activator.CreateInstance(mappedClass, true);
```
fredericDelaporte: As discussed in #1758, this usage is not a dependency but an entity instantiation. If the user wants some custom behavior for instantiating value types, he now would have to provide its own tuplizer.

hazzik: I don't think that the `true` here is required.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1261,35 +1261,65 @@ private void Init()
 
 		private ICurrentSessionContext BuildCurrentSessionContext()
 		{
```
hazzik: I would not change this method and would not change the standard contexts' constructors. Also I would make the `ICurrentSessionContextWithFactory` interface optional.

src/NHibernate/Tuple/PocoInstantiator.cs
hazzik: Probably in some partial trust or similar environments.

src/NHibernate/Impl/SessionFactoryImpl.cs
fredericDelaporte: If this method is not changed then `ICurrentSessionContextWithFactory` would not have any usage and should not be kept.

So, do you mean you would not change just the direct instantiation cases?

About making `ICurrentSessionContextWithFactory` optional, then the obsoleted line:
```c#
(ICurrentSessionContext) Environment.ObjectsFactory.CreateInstance(implClass, this);
```

Would have to be replaced by:
```c#

**Comment by fredericDelaporte** [comment]
Rebased and fixup added for cleaning-up session context case in another way. It adds a possible breaking change, see first comment update.

File: src/NHibernate/Tuple/PocoInstantiator.cs (hunk 3 above)
Comment (fredericDelaporte) [reply src/NHibernate/Tuple/PocoInstantiator.cs]: I have just preferred to be quite conservative, inlining the default previous implementation "as is".

File: src/NHibernate/Context/ICurrentSessionContext.cs [hunk 5]
```diff
@@ -42,4 +43,18 @@ public interface ICurrentSessionContext
 		/// locating or creating the current session.</exception>
 		ISession CurrentSession();
 	}
+
+	/// <summary>
+	/// An <see cref="ICurrentSessionContext"/> allowing to set its session factory. Implementing
+	/// this interface allows the <see cref="IObjectsFactory"/> to be used for instantiating the
+	/// session context.
+	/// </summary>
+	public interface ICurrentSessionContextWithFactory : ICurrentSessionContext
```
hazzik: Please rename to `ISessionFactoryAwareCurrentSessionContext`.

hazzik:
Shall we go further (in other PR) and replace `IObjectsFactory` with `IServiceProvider`?

fredericDelaporte:
> Shall we go further (in other PR) and replace `IObjectsFactory` with `IServiceProvider`?

Yes, we should.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1281,9 +1281,15 @@ private ICurrentSessionContext BuildCurrentSessionContext()
...
-					(ICurrentSessionContext)Environment.ObjectsFactory.CreateInstance(implClass, new object[] { this });
+				var implClass = ReflectHelper.ClassForName(impl);
+				if (!typeof(ISessionFactoryAwareCurrentSessionContext).IsAssignableFrom(implClass))
+				{
+					return (ICurrentSessionContext) Activator.CreateInstance(implClass, this);
+				}
+
+				var context = (ISessionFactoryAwareCurrentSessionContext) Environment.ObjectsFactory.CreateInstance(implClass);
+				context.SetFactory(this);
+				return context;
```
hazzik: I was thinking a lot about this chunk of code, but could not come up with a great solution, unfortunately.

So, I came up with this:

```csharp

var implClass = ReflectHelper.ClassForName(impl);
var constructor = implClass.GetConstructor(new [] { typeof(ISessionFactoryImplementor) });
ICurrentSessionContext context;
if (constructor != null)
{
  context = (ICurrentSessionContext) constructor.Invoke(new object[] { this });
}
else
{
  context = (ICurrentSessionContext) Environment.ObjectsFactory.CreateInstance(implClass);
}
if (context is ISessionFactoryAwareCurrentSessionContext sessionFactoryAwareContext)
{
  sessionFactoryAwareContext.SetFactory(this);
}
```

fredericDelaporte: Ok, changed for this. (And rebased.)

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1281,9 +1281,21 @@ private ICurrentSessionContext BuildCurrentSessionContext()
...
+					context = (ICurrentSessionContext) constructor.Invoke(new object[] { this });
+				}
+				else
+				{
+					context = (ICurrentSessionContext) Environment.ObjectsFactory.CreateInstance(implClass);
+				}
+				if (context is ISessionFactoryAwareCurrentSessionContext sessionFactoryAwareContext)
+				{
+					sessionFactoryAwareContext.SetFactory(this);
+				}
```
hazzik: You've forgot to add `return context;`

PR #1783, opened 2018-07-05
Obsolete MultiQuery and MultiCriteria

### LINKED ISSUE #840: NH-3835 - Future/MultiCriteria 2nd level caching
**Fabrizio Ferrari** created an [issue](https://nhibernate.jira.com/browse/NH-3835) — <time datetime="2015-10-22T07:38:27.150Z" title="2015-10-22T07:38:27.150Z">22nd October 2015, 7:38:27</time>:
> When using Future/MultiCriteria the queries issued to the db are cached as a single query. So, if later you need only some of them the cache has no effect. I don't know how if it implies big changes, but it would be great to have the queries cached separately, otherwise the only option is not to use Futures.
>

### LINKED ISSUE #1344: NH-3864 - Cacheable Multicriteria/Future'd query with aliased join throw exception
**Pawel Owerko** created an [issue](https://nhibernate.jira.com/browse/NH-3864) — <time datetime="2016-04-25T10:03:34.893Z" title="2016-04-25T10:03:34.893Z">25th April 2016, 10:03:34</time>:
> Short scenario:
> 
> ```csharp
> session.QueryOver<Person>()
> .JoinAlias(x => x.Children, () => childAlias);
> 
> var multiCriteria = s.CreateMultiCriteria();
> multiCriteria.Add("myQuery", query);
> multiCriteria.SetCacheable(true);
> ```
> 
> 
> Execution:
> 
> ```csharp
> multiCriteria.List()
> ```
> 
> 
> Will throw exception:
> >System.InvalidCastException : Unable to cast object of type 'namespace.Person' to type 'System.Object[]'.
> 
> Same for 
> ```csharp
>   .JoinQueryOver(x => x.Children, () => childAlias);
> ```
> 
> 
> The bug actually lies is in MultipleQueriesCacheAssembler Disassemble method.
> 
---
**Pawel Owerko** added a comment — <time datetime="2016-04-25T10:16:59.554Z" title="2016-04-25T10:16:59.554Z">25th April 2016, 10:16:59</time>:
> Commit with failing tests:
> https://github.com/itmagination/nhibernate-core/commit/0ef5918660cabb062913125bab667f59f4be0277
> 
>

fredericDelaporte:
The `QueryBatch` can replace them both, and does not suffer of some of their bugs.

This is a follow-up to #1718 (Combine Future queries for ICriteria and IQuery), which has added the `QueryBatch`.

At first I was considering adding to it the `SetParameter` capabilities of `MultiQuery`, so I was considering better to do that in another PR. But finally, I think we can drop that feature, as it does not make sense for most querying API handled: Criteria, QueryOver and Linq. Only HQL and SQL queries can benefit of it.

Closes #840 by obsoleting it.
Closes #1344 by obsoleting it.


fredericDelaporte:
Rebased and amended for accounting `TestCase` changes.

fredericDelaporte:
Rebased for resolving conflicts with #1789.

src/NHibernate.Test/NHSpecificTest/NH2195/SQLiteMultiCriteriaTest.cs
```diff
@@ -144,5 +144,63 @@ public void MultiCriteriaQueriesWithStringsShouldExecuteCorrectly()
 				Assert.AreEqual(1, list.Count);
 			}
 		}
+
+		[Test]
+		public void QueryBatchWithIntsShouldExecuteCorrectly()
+		{
+			var driver = Sfi.ConnectionProvider.Driver;
+			if (!driver.SupportsMultipleQueries)
+				Assert.Ignore("Driver {0} does not support multi-queries", driver.GetType().FullName);
```
bahusoid: Are you sure we want to skip such tests (that are not checking that single query is executed for batch but some data conditions retrieval)? As query batch doesn't throw maybe it's good to test that it also works adequeatly when `driver.SupportsMultipleQueries == false`.

src/NHibernate.Test/QueryTest/QueryBatchFixture.cs
```diff
@@ -0,0 +1,1003 @@
...
+
+				t.Commit();
+			}
+		}
+
+		/// <summary>
+		/// Get the inner Hashtable from the IQueryCache.Cache
+		/// </summary>
+		/// <returns></returns>
+		public static Hashtable GetHashTableUsedAsQueryCache(ISessionFactoryImplementor factory)
```
bahusoid: Could you please clarify the reasons for this hackish method? It seems all usages are equivalent to `factory.QueryCache.Clear()`

src/NHibernate.Test/NHSpecificTest/NH2195/SQLiteMultiCriteriaTest.cs
fredericDelaporte: I have port existing multi-query/criteria tests without changing their conditions. Many of them would fail with drivers not supporting multiple queries, because they would see more queries executed than expected.

src/NHibernate.Test/QueryTest/QueryBatchFixture.cs
fredericDelaporte: No I cannot. That is just the way it was done for multi-criteria/query.

fredericDelaporte: In fact reading the code, I can: it is not used only for clearing, but also for checking directly that it gets filled with items.

bahusoid: Isn't it something that can be checked with "Sfi.Statistics.QueryCachePutCount"?

fredericDelaporte: It could, assuming there are no bugs in stats. Here with this test it is really certain things are in the cache.

bahusoid: But that's something that you are already using for other tests -  `CanGetMultiCriteriaFromSecondLevelCache`  for example.
So the whole usage seems to be analog of clearing statistics and checking put counts. IMHO we should get rid of this method for good.

fredericDelaporte: Yes I know it is done for other tests. About tests I do not think consistency in the way things are tested is good. Instead, testing with different ways help catching more cases.

bahusoid: I see your point. But for me it looks like introducing dirty hack for no particular reasons (to me checking count of items in internal cache implementation doesn't add any value to existing tests - you can't say for sure that those items are stored properly and can be retrieved by the same query reexecution).

doc/reference/modules/performance.xml
```diff
@@ -1216,131 +1216,194 @@ sessionFactory.EvictCollection("Eg.Cat.Kittens");]]></programlisting>
...
         </para>
 
+        <programlisting><![CDATA[using NHibernate.Multi;
+
+...
+
+var queries = s.CreateQueryBatch()
+    .Add("queryOverList", s.QueryOver<Item>().Where(i => i.Category == "Food"))
+    .Add<long>("sqlCount",
+        s.CreateSQLQuery("select count(*) as count from Item i where i.Category = :cat")
```
bahusoid: As this "as count" can cause issues on some dialects maybe we shouldn't use it documentation.

src/NHibernate.Test/NHSpecificTest/NH2201/Fixture.cs
```diff
@@ -54,5 +55,25 @@ public void CanUseMutliCriteriaAndFetchSelect()
 				Console.WriteLine("*** end");
 			}
 		}
+
+		[Test]
+		public void CanUseQueryBatchAndFetchSelect()
+		{
+			using (var s = OpenSession())
+			{
+				Console.WriteLine("*** start");
```
bahusoid: I think it's better to remove these "*** star*" and "*** end" logging in ported tests.

src/NHibernate.Test/Stats/StatsFixture.cs
```diff
@@ -243,16 +244,40 @@ public void IncrementQueryExecutionCount_WhenExplicitQueryIsExecuted()
...
+#pragma warning restore 618
 				}
 				Assert.AreEqual(1, stats.QueryExecutionCount);
 
 				stats.Clear();
 				using (var s = OpenSession())
 				{
+#pragma warning disable 618
 					var r = s.CreateMultiCriteria().Add(DetachedCriteria.For<Country>()).Add(DetachedCriteria.For<Continent>()).List();
+#pragma warning restore 618
```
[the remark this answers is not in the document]
bahusoid: To simplify removing obsoleted things later I think it's better scope this disable warning to  the whole block that should be remove.

igitur:
Excuse me if this is the wrong place to ask. With the new query batch method, is it possible to get the number of queries from the statistics? Previously, I could use `SessionFactory.Statistics.Queries[0]` which would return e.g. `5 queries (MultiQuery)`. Is there something similar for query batch?

fredericDelaporte:
This was similar up to #1788, where it has been changed "by the way" following [this discussion](https://github.com/nhibernate/nhibernate-core/pull/1788#discussion_r201948948). It will now log the actual sql executed instead of just the count of queries, aligning with single queries behavior.

PR #1786, opened 2018-07-06
Document future results

fredericDelaporte:
They were not documented, excepted for the Linq case.


doc/reference/modules/performance.xml
```diff
@@ -1342,4 +1342,67 @@ ICollection<Policy> policies = CollectionHelper.ToArray<Policy>(results[1]);]]><
...
+
+        <para>
+            Queries can be converted to future results instead of being directly executed. Future
+            results are not evaluated till one gets executed. At that point, all defined future
+            results are evaluated in one single round-trip to the database.
+        </para>
+
+        <para>
+            Future results are an alternative to using <xref linkend="performance-multi-query"/>.
+            They avoid the need to explicitly regroup queries, but they also hides which queries will
```
hazzik: “hide”

PR #1788, opened 2018-07-09
Implement multiple get and put for query cache and query batch

fredericDelaporte:
This is a follow up to @maca88  #1633 (batching 2nd level cache operations) and @bahusoid #1718 (query batch). It aims at combining both: enabling batching of 2nd level cache operations for the query batch.


src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -106,60 +143,250 @@ public IList Get(QueryKey key, ICacheAssembler[] returnTypes, bool isNaturalKeyL
...
+
+				var queryParams = queryParameters[i];
+				if (queryParams.IsReadOnlyInitialized)
+					persistenceContext.DefaultReadOnly = queryParams.ReadOnly;
+				else
+					queryParams.ReadOnly = persistenceContext.DefaultReadOnly;
+
+				try
+				{
+					results[i] = GetResultFromCacheable(key, returnTypes[i], queryParams.NaturalKeyLookup, session, cacheable);
```
fredericDelaporte: Here we are in a loop on each query for handling one query result per iteration. But `GetResultFromCacheable` may also trigger cache gets, in case the query result contains some `ManyToOneType` to `Assemble` by example.
In such case, an immediate load is triggered. It will cause entity cache gets, which will be batched through the `batch-size` mechanism, since #1633: the result is looped for calling `BeforeAssemble` before being looped a second time for actually assembling. `BeforeAssemble` does schedule batching of the entities, with the lazy load batching mechanism.
So there is still some batching of those entity cache gets too, but it is not well suited to the query cache get case (including single query gets):
- It may get other entities unrelated to the queries being loaded, potentially with an inadequate readonly status on the persistence context. This trouble is not new, it could already occurs previously on single query get, through batch loading from database when the entity was not cached (either by user choice or due to a cache miss for whatever reason). But it has somewhat be worsen by #1633, since it may now occur without a cache miss due to entity cache get batching using the `batch-size` mechanism.
- It is restricted to `batch-size` entities, as set on the entity mapping. In the query case, it would be better to batch the gets of all the entities matching the query.
- It is segregated by entity persister, while again in the query case, it would  be better to batch them all together.
- It is segregated by query, while in the multiple query get case, it could be batched across queries sharing the same readonly status.

For handling all that, a dedicated cache batching mechanism would be required, allowing to batch the entity cache gets of the queries together without batching with them other pending gets. It should also take care of the readonly status of the persistence context.  The last point (the only one specific to multi-query get) would moreover require to handle the `BeforeAssemble` calls in another dedicated loop on queries in order to batch them together before doing another loop for actually assembling all the queries.

That may be an improvement for later. I have decided not to do it here to avoid expanding a bit too much the scope of this PR.

src/NHibernate/Cache/UpdateTimestampsCache.cs
```diff
@@ -18,7 +19,8 @@ public partial class UpdateTimestampsCache
 	{
 		private static readonly INHibernateLogger log = NHibernateLogger.For(typeof(UpdateTimestampsCache));
 		private ICache updateTimestamps;
-		private readonly IBatchableReadOnlyCache _batchUpdateTimestamps;
+		private readonly IBatchableReadOnlyCache _batchReadOnlyUpdateTimestamps;
+		private readonly IBatchableCache _batchUpdateTimestamps;
```
fredericDelaporte: Like the query cache, it would be simpler with #1777.

src/NHibernate/Cache/UpdateTimestampsCache.cs
```diff
@@ -72,45 +72,107 @@ public virtual void Invalidate(IReadOnlyCollection<string> spaces)
...
 		}
 
 		[MethodImpl(MethodImplOptions.Synchronized)]
 		public virtual bool IsUpToDate(ISet<string> spaces, long timestamp /* H2.1 has Long here */)
 		{
-			if (_batchUpdateTimestamps != null)
+			if (_batchReadOnlyUpdateTimestamps != null)
 			{
+				if (spaces.Count == 0)
+					return true;
```
fredericDelaporte: Avoid an undue cache `GetMany` hit: NHibernate may ask up-to-dateness for no spaces at all. With single operations, this is resolved to an empty loop, so no actual call to the cache is issued. With multi-operations, a multi-get call was issued on the cache, with an empty list of keys: better avoid that call.

src/NHibernate/Cache/UpdateTimestampsCache.cs
```diff
@@ -72,45 +72,107 @@ public virtual void Invalidate(IReadOnlyCollection<string> spaces)
...
+		}
+
+		private void SetSpacesTimestamp(IReadOnlyCollection<string> spaces, long ts)
+		{
+			if (_batchUpdateTimestamps != null)
 			{
-				log.Debug("Invalidating space [{0}]", space);
-				updateTimestamps.Put(space, ts);
+				if (spaces.Count == 0)
+					return;
```
fredericDelaporte: Avoid an undue cache `PutMany` hit: NHibernate may ask to "invalidate" no spaces at all. With single operations, this is resolved to an empty loop, so no actual call to the cache is issued. With multi-operations, a multi-put call would be issued on the cache, with an empty list of keys: better avoid that call.

src/NHibernate/Impl/MultipleQueriesCacheAssembler.cs
```diff
@@ -77,7 +77,7 @@ public object Assemble(object cached, ISessionImplementor session, object owner)
 			if (!queryParameters.ForceCacheRefresh)
 			{
 				IList list =
-					queryCache.Get(key, new ICacheAssembler[] {this}, queryParameters.NaturalKeyLookup, querySpaces, session);
+					queryCache.Get(key, queryParameters, new ICacheAssembler[] {this}, querySpaces, session);
```
fredericDelaporte: Multi-criteria/query do not handle readonly globally either (at least their `queryParameters` readonly setting is not set anywhere), so calling the new `Get` will not introduce a handling which was not here previously.

src/NHibernate/Loader/Loader.cs
```diff
@@ -599,7 +601,9 @@ private IEnumerable<SubselectFetch> CreateSubselects(IList<EntityKey[]> keys, Qu
 			}
 		}
 
-		internal void InitializeEntitiesAndCollections(IList hydratedObjects, object resultSetId, ISessionImplementor session, bool readOnly)
+		internal void InitializeEntitiesAndCollections(
+			IList hydratedObjects, object resultSetId, ISessionImplementor session, bool readOnly,
+			CacheBatcher cacheBatcher)
```
fredericDelaporte: For minimizing changes I have initially attempted to give a default `null` value to the added `cacheBatcher` parameter, but maca88/AsyncGenerator#115 forbids this by generating an invalid async counterpart in such case.

This change allows batching cache puts across queries in the case of query batch. Due to current `CacheBatcher` implementation, it will be limited to `batch-size` (according to entities mapping) and segregated by persister, which I do not consider very suitable, as discussed [here](https://github.com/nhibernate/nhibernate-core/pull/1633#discussion_r200837292).

Here too, for avoiding broadening the scope of the PR, I will leave this for a later improvement.

src/NHibernate/Loader/Loader.cs
```diff
@@ -1708,60 +1716,54 @@ private CacheableResultTransformer CreateCacheableResultTransformer(QueryParamet
...
+			ISessionImplementor session, QueryParameters queryParameters, ISet<string> querySpaces,
+			IQueryCache queryCache, QueryKey key)
+		{
+			if (!CanGetFromCache(session, queryParameters))
+				return null;
 
-				if (queryParameters.IsReadOnlyInitialized)
-					persistenceContext.DefaultReadOnly = queryParameters.ReadOnly;
-				else
-					queryParameters.ReadOnly = persistenceContext.DefaultReadOnly;
```
fredericDelaporte: This handling has been moved on the standard query cache side, as written in previous comments.

src/NHibernate/Loader/Loader.cs
```diff
@@ -1708,60 +1716,54 @@ private CacheableResultTransformer CreateCacheableResultTransformer(QueryParamet
...
+			var result = queryCache.Get(
+				key, queryParameters, 
+				queryParameters.HasAutoDiscoverScalarTypes
+					? null
+					: key.ResultTransformer.GetCachedResultTypes(ResultTypes),
+				querySpaces, session);
 
-				try
+			if (_factory.Statistics.IsStatisticsEnabled)
+			{
```
fredericDelaporte: I was tempted to move the statistics handling on the query cache side too, but that is not done for concurrency strategy caches either. So for consistency, I have kept it here.

src/NHibernate/Multi/QueryBatch.cs
```diff
@@ -44,18 +47,17 @@ public void Execute()
 			{
 				Init();
 
-				if (!Session.Factory.ConnectionProvider.Driver.SupportsMultipleQueries)
+				using (Session.BeginProcess())
```
fredericDelaporte: Unrelated slight improvement here: better have the `BeginProcess` around the lack of multi-queries support case too, in order to avoid starting processing on each query.

src/NHibernate/Multi/QueryBatch.cs
```diff
@@ -44,18 +47,17 @@ public void Execute()
...
+						}
+						return;
 					}
-					return;
-				}
-
-				using (Session.BeginProcess())
-				{
-					DoExecute();
+					ExecuteBatched();
```
fredericDelaporte: Unrelated slight improvement here: renamed for consistency with `ExecuteNonBatched`.

src/NHibernate/Multi/QueryBatch.cs
```diff
@@ -120,53 +122,51 @@ private void Init()
...
-
-			var querySpaces = new HashSet<string>(_queries.SelectMany(t => t.GetQuerySpaces()));
-			if (resultSetsCommand.HasQueries)
+			var querySpaces = new HashSet<string>(
+				_queries.SelectMany(q => q.QueryInformation).SelectMany(qi => qi.QuerySpaces));
+			if (querySpaces.Count > 0)
 			{
+				// The auto-flush must be handled before obtaining the commands, because an auto-flush
+				// may have to invalidate cached data which otherwise would cause a command to be skipped.
 				Session.AutoFlushIfRequired(querySpaces);
```
fredericDelaporte: Somewhat unrelated bug fix: the auto-flush was previously done after obtaining results from cache (through `CombineQueries`), which is an error and may cause queries to use outdated cached result. The auto-flush must be done before any cache hit.

src/NHibernate/Multi/QueryBatch.cs
```diff
@@ -120,53 +122,51 @@ private void Init()
...
 
-			bool statsEnabled = Session.Factory.Statistics.IsStatisticsEnabled;
-			Stopwatch stopWatch = null;
-			if (statsEnabled)
-			{
-				stopWatch = new Stopwatch();
-				stopWatch.Start();
-			}
+			var resultSetsCommand = Session.Factory.ConnectionProvider.Driver.GetResultSetsCommand(Session);
+			CombineQueries(resultSetsCommand, GetCachedResults());
```
fredericDelaporte: I could have put `GetCachedResults()` in `CombineQueries`, but I have decided to surface it here as a counterpart to `PutCacheableResults()` later in this method, and to make it more obvious there is a cache get operation here.

src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -149,61 +172,63 @@ public IEnumerable<ISqlCommand> GetCommands()
 
 						tmpResults.Add(o);
 					}
-					_loaderResults[index] = tmpResults;
-
-					if (index == _queryInfos.Count - 1)
-					{
-						InitializeEntitiesAndCollections(reader, hydratedObjects);
-					}
```
fredericDelaporte: Somewhat unrelated bug fix by the way here: this `InitializeEntitiesAndCollections(reader, hydratedObjects)` call would be missing if the last result happen to be cached (and not a previous result). But for handling cache put batching I had to move it as a distinct call from the query batcher, fixing this trouble here by the way.

bahusoid:
I dislike `IQueryInformation` interface and even more dislike how it ended up in base `IQueryBatchItem` interface - with all entity loading specifics (like `Loader` and `QueryParameters` classes) and along with some temp values  (like untyped `IList Result` (yayk!)). Why all this information is required outside of `QueryBatchItemBase`? 

In `QueryBatch` query cache was disconnected from `QueryBatchItem` - and I think we should keep it that way.
Won't it be more cleaner if we introduce some way in `QueryBatchItem` to retrieve common context from  `IQueryBatch`. Something like:
```C#
	public interface IExecutableContext
	{
		void Execute();
	}

	public class QueryBatchContext : IExecutableContext
	{
		Dictionary<System.Type, IExecutableContext> _dic = new Dictionary<System.Type, IExecutableContext>();

		public void Execute()
		{
			foreach (var v in _dic.Values)
				v.Execute();
		}
		
		//retrieves the same instance of TContext for all IQueryBatchItem inside the same IQueryBatch
		public TContext GetOrCreate<TContext>(Func<TContext> createNew) where TContext : IExecutableContext
		{
			if(_dic.TryGetValue(typeof(TContext), out var val))
			{
				return (TContext) val;
			}

			var newValue = createNew();
			_dic[typeof(TContext)] = newValue;
			return newValue;
		}
	}
```	
And extend `IQueryBatchItem`:
`void Init(ISessionImplementor session, `**IQueryBatchContext context**`);	`

And somewhere in `QueryBatchItemBase` operate with common `CacheBatcher`.
```C#
		batchContext = _context.GetOrCreate<QueryBatchItemContext>(()=> new QueryBatchItemContext(){CacheBatcher = new CacheBatcher()});
		batchContext.CacheBatcher...
```
Just a quick thoughts for further discussion.

fredericDelaporte:
> I dislike `IQueryInformation` interface and even more dislike how it ended up in base `IQueryBatchItem` interface - with all entity loading specifics (like `Loader` and `QueryParameters` classes) and along with some temp values  (like untyped `IList Result` (yayk!)). Why all this information is required outside of `QueryBatchItemBase`? 

Because they are required for processing caching, and we must process caching in the query batcher for batching it. Batching is the responsibility of the batcher, right? That is like commands extracted out of queries: they are needed in the batcher for being batched in it. Likewise, we need the cache keys of each query, along with all required information for processing caching which includes the `QueryParameters` and data and logic carried by the `Loader`. (We could eventually hide this one with additional properties and methods in `IQueryInformation`, which would give access to that data and logic from the `Loader`. But I do not think it would be worth it.)
Of course for batching puts, we also need the results of each queries, and we need to know which results have to be put. It is an `IList` because, #892 is still far from being done and requires a lot of breaking changes. We cannot switch them easily to generic lists.

I consider `IQueryBatchItem` is an implementation dependent interface, like `ISessionImplementor`. It is not meant to be used directly by users. Indeed, users do not need to care for query batch items, they are abstracted away by the `IQueryBatch` extension methods. `IQueryBatchItem` is there to provide the query batch with the data and processing it requires to batch queries. So exposing additional data and processing for batching caching just looks natural to me.

> In `QueryBatch` query cache was disconnected from `QueryBatchItem` - and I think we should keep it that way.

Not really. The `QueryBatchItem` was having logic tied to the query cache, even resolving it in some method. Granted, from the query batch perspective it was hidden away, likely favoring the auto-flush bug I have commented on in this PR (which fixes it by the way). But as written above, for batching caching, the batcher has to handle it.

> Won't it be more cleaner if we introduce some way in `QueryBatchItem` to retrieve common context from  `IQueryBatch`.

I do not really see how the context idea could handle all cache batching cases without adding a bit too much complexity. But feel free to propose changes through by example a PR on this PR branch.

bahusoid:
>It is not meant to be used directly by users.

`IQueryBatchItem` was meant to be easily extended by user with some custom user implemented queries   (not necessarily even read queries) - not to be bounded to Loader and all other IQuery/ICriteria related stuff and not bound to query cache  (to me it looks like a pretty complex and internal concept).

And now all the `QueryBatchItemBase` implementation details are leaked in main `IQueryBatchItem` interface and in to QueryBatch class to implement this batch caching. So this implementation looks wrong to me. And I think you solution is already "adding a bit too much complexity".

Unfortunatly I don't have access to Visual Studio right now and have limited coding capacity (so I won't be able to provide any live code examples in the following month). So right now all I can do is whining and complaining :)

fredericDelaporte:
Then maybe the `GetQuerySpaces` has to be put back in place, and the `QueryInformation` has to be replaced by an optional `CachingInformation`. It would allow implementors to just ignore it if they wish while still allowing others to handle it for enabling batching of cache. (Eventually it could even be changed to a class like `CacheBatcher` updated by this PR, exposing no public members and so clearly reserved for internal usage.)

> So this implementation looks wrong to me. And I think you solution is already "adding a bit too much complexity".

Indeed, batching queries is complex. Any simplification is welcome. But the context idea may obsfuscate what implementors should do with it.

bahusoid:
>Then maybe the GetQuerySpaces has to be put back in place, and the QueryInformation has to be replaced by an optional CachingInformation. 

Yeah that sounds like better solution to me. It's indeed should be exposed as an optional for implementor property/method that should return details only if you want to deal with "batchable query cache" (maybe name should also be more explicit with "batchable" in it).

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -106,60 +143,250 @@ public IList Get(QueryKey key, ICacheAssembler[] returnTypes, bool isNaturalKeyL
...
 			}
+			catch (UnresolvableObjectException ex)
+			{
+				if (isNaturalKeyLookup)
+				{
+					//TODO: not really completely correct, since
+					//      the UnresolvableObjectException could occur while resolving
+					//      associations, leaving the PC in an inconsistent state
+					Log.Debug(ex, "could not reassemble cached result set");
+					// TODO: add and handle a RemoveMany?
```
maca88: I forgot about `RemoveMany` when doing #1633 as it was not needed there, but it is [already implemented](https://github.com/nhibernate/NHibernate-Caches/pull/45/files#diff-bd639a7bcffa3fc62ab405a2ab4a63d5R353) in nhibernate/NHibernate-Caches#45, so why not add it if it can be used here.

fredericDelaporte: I have put a question mark because this case is almost an error case, and it does not look likely to me it would benefit from batching. It should not happen frequently, and even less frequently would it benefit from a `Many` handling, since the first encountered one short-circuit the result-set processing: it could be batched only in multi-queries case, if many queries in it encounter a "no more resolvable entity reference" case.

So in fact I should change that comment, I do not think it would be worth it to handle a many case here.

src/NHibernate/Multi/IQueryBatchItem.cs
```diff
@@ -32,43 +32,57 @@ public interface IQueryBatchItem<TResult> : IQueryBatchItem
...
+		IEnumerable<string> GetQuerySpaces();
 		
 		/// <summary>
-		/// Gets the commands to execute for getting the not-already cached results of this query.
+		/// Get the commands to execute for getting the not-already cached results of this query.
 		/// </summary>
 		/// <param name="cachedResults">The cached results.</param>
 		/// <returns>The commands for obtaining the results not already cached.</returns>
-		IEnumerable<ISqlCommand> GetCommands(IDictionary<IQueryInformation, IList> cachedResults);
+		IEnumerable<ISqlCommand> GetCommands(IDictionary<ICachingInformation, IList> cachedResults);
```
bahusoid: It seems this change is not really required. Why can't we set all required information right in to `ICachingInformation`? 
You can add something like `ICachingInfromation.SetCachedResults` and call it right in `BatchItem.GetCachedResults`. So no additional lookup and less impact on base interface  for this specific feature. 

We can also provide reference to common `CacheBatcher`  via `ICachingInformation` - so it seems we also can get rid  of 'InitializeEntitiesAndCollections'. Thus making cache implementation exposed only via single property.

src/NHibernate/Multi/IQueryBatchItem.cs
```diff
@@ -30,37 +32,60 @@ public interface IQueryBatchItem<TResult> : IQueryBatchItem
...
+		IEnumerable<string> GetQuerySpaces();
 		
 		/// <summary>
-		/// Returns commands generated by query
+		/// Get the commands to execute for getting the not-already cached results of this query.
 		/// </summary>
-		IEnumerable<ISqlCommand> GetCommands();
+		/// <param name="cachedResults">The cached results.</param>
+		/// <returns>The commands for obtaining the results not already cached.</returns>
+		IEnumerable<ISqlCommand> GetCommands(IDictionary<ICachingInformation, IList> cachedResults);
```
bahusoid: It seems this change is not really required. Why can't we set all required information right in to `ICachingInformation`? 
You can add something like `ICachingInfromation.SetCachedResults` and call it right in `BatchItem.GetCachedResults`. So no additional lookup and less impact on base interface  for this specific feature. 

We can also provide reference to common `CacheBatcher`  via `ICachingInformation` - so it seems we also can get rid  of 'InitializeEntitiesAndCollections'. Thus making cache implementation exposed only via single property.

fredericDelaporte: Removing `IQueryBatchItem.InitializeEntitiesAndCollections` is not fully done, as it requires some refactoring done in #1789 for being done without re-introducing a bug.

src/NHibernate/Loader/Loader.cs
maca88: >  maca88/AsyncGenerator#115 forbids this by generating an invalid async counterpart in such case

Fixed in version `0.8.2.9`.

fredericDelaporte:
Squashed and rebased for resolving conflicts with #1789.

src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -17,36 +18,104 @@ public abstract partial class QueryBatchItemBase<TResult> : IQueryBatchItem<TRes
...
 
-		protected abstract List<QueryLoadInfo> GetQueryLoadInfo();
+		protected abstract List<QueryInfo> GetQueryInformation(ISessionImplementor session);
+
+		/// <inheritdoc />
+		public IEnumerable<ICachingInformation> CachingInformation
+		{
+			get
+			{
+				ThrowIfNotInitialized();
```
fredericDelaporte: This one would have been better moved in #1789, but I forgot it.

src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -243,8 +294,15 @@ private void InitializeEntitiesAndCollections(DbDataReader reader, List<object>[
 				if (queryInfo.IsResultFromCache)
 					continue;
 				queryInfo.Loader.InitializeEntitiesAndCollections(
-					hydratedObjects[i], reader, Session, Session.PersistenceContext.DefaultReadOnly);
+					hydratedObjects[i], reader, Session, queryInfo.Parameters.IsReadOnly(Session),
```
fredericDelaporte: This change would also have been better moved in #1789. The old code was from multi-query/criteria, which was not knowing for which queries they were initializing and so could not handle their readonly flag.

src/NHibernate/Loader/Loader.cs
```diff
@@ -1708,60 +1714,54 @@ private CacheableResultTransformer CreateCacheableResultTransformer(QueryParamet
 				queryParameters.HasAutoDiscoverScalarTypes, SqlString);
 		}
 
-		internal IList GetResultFromQueryCache(ISessionImplementor session, QueryParameters queryParameters,
-											  ISet<string> querySpaces, IQueryCache queryCache,
-											  QueryKey key)
+		internal bool CanGetFromCache(ISessionImplementor session, QueryParameters queryParameters)
```
bahusoid: This method doesn't handle per query `SetCacheMode` calls. Which actually now is used in tests (as no op but might lead to confusion that it actually can be used in Future):

https://github.com/nhibernate/nhibernate-core/blob/7ce79eda3a184e3225fcf6fa4ed3c15fc27dc0ab/src/NHibernate.Test/Futures/QueryBatchFixture.cs#L354-L355

Maybe this query cache mode needs to be moved to `QueryParameters` and handled similarly as `QueryParameters.ReadOnly` flag?

So we should move this code to `QueryParameters.CanGetFromCache(session)` method. And for parity introduce `QueryParameters.CanPutToCache(session)`?

src/NHibernate/Multi/QueryBatch.cs
```diff
@@ -195,5 +203,110 @@ protected void ExecuteBatched()
 					stopWatch.Elapsed);
```
bahusoid: I've basically copy pasted this statistic call from multi query/criteria implementation. But it seems it's not correct usage - as it expects query string as parameter and use this query string as dictionary key later. So it seems `resultSetsCommand.Sql` or some other unique batch identifier needs to be added here to `$"{_queries.Count} queries"` part.

src/NHibernate/Multi/ICachingInformation.cs
```diff
@@ -0,0 +1,76 @@
...
+	{
+		/// <summary>
+		/// The query parameters.
+		/// </summary>
+		QueryParameters Parameters { get; }
+
+		/// <summary>
+		/// The query result.
+		/// </summary>
+		IList Result { get; }
```
bahusoid: May be it should be exposed as `GetResultsToCache` or some other name that suggests that implementor should return values that's going to be put to cache.

src/NHibernate/Loader/Loader.cs
fredericDelaporte: Indeed, the queries individual cache mode has never been handled for future and multi-query/criteria, but we could. I do not think it should be done here, but better as another PR.

src/NHibernate/Multi/ICachingInformation.cs
fredericDelaporte: If doing so, it should then hold neither results retrieved from cache nor results after cacheable transformation which should not be put (the put must be done on the untransformed result).

src/NHibernate/Loader/Loader.cs
fredericDelaporte: I think it will not work, so better stick to YAGNI, not doing this in advance.

`CacheMode` needs to be temporarily switched on the session instead, because many cache interactions triggered deeply in the loading implementation have no access to query parameters.

src/NHibernate/Multi/QueryBatch.cs
fredericDelaporte: Something like `StringHelper.CollectionToString(_queries)` should do it, provided a suitable `ToString` override is added on query items. I would rather do that on the PR for handling cache mode.

src/NHibernate/Loader/Loader.cs
bahusoid: For this PR I'm just saying that it's better to have consistent way for readonly check for query and cache mode check for query. So from this point - what will not work?
We use `queryParameters.IsReadOnly(session)` for readonly check. So why introduce new logically similair method in `Loader` instead? It can be still marked internal here if you think this "cache mode per query" won't work.

src/NHibernate/Multi/QueryBatch.cs
bahusoid: Well... Using `resultSetsCommand.Sql` sounds as much simpler solution. Considering that caching changes number of really executed queris per batch

src/NHibernate/Loader/Loader.cs
fredericDelaporte: These cases are not comparable in my opinion:
- `IsReadonly` is a bit messy, originating from times where nullable were not avaliable. It is also there to hide the complexity of replacing nullability by an additional boolean.
- `CacheMode` is not a boolean but a flag enumeration value. It conveys more state than a boolean. `CanGetFromCache` is just a small part of it, because by example it tells nothing about what should be done with entities assembled from their id cached in query result, whether they should themselves be put into the cache or not if they were missing in it when re-assembled.
  So it would be just a partial part of the information we actually need to handle properly cache mode, and we will have to put more somehow somewhere. It then looks likely to me that `CanGetFromCache` will just be redundant and not needed.

src/NHibernate/Multi/QueryBatch.cs
fredericDelaporte: Putting SQL here is not an issue. But are this stats meant to cover only db hits or also cache hits?

Looking at their other usages, it is only db hits, so then better use `resultSetsCommand.Sql` than `_queries` without filtering those retrieved from cache. Note that we have the `QueryIdentifier`, meant to be provided to these stats, available in `CachingInformation`. But since these `CachingInformation` are optional, we cannot use that.

src/NHibernate/Loader/Loader.cs
bahusoid: I'm not talking about how data needs to be stored. But rather how it should be used in code.
Calling `queryParameters.IsReadOnly(session)` for checking query read-only flag looks pretty clean to me. How messy its implementation is irrelevant.
And as we need now method  `CanGetFromCache` I think place and synatx should be the same. But let's not spend more time on it. If you are not convinced - let's keep it as is for time being.

fredericDelaporte:
I validated it, so it was also my responsibility.

src/NHibernate/Loader/Loader.cs
fredericDelaporte: You were right with the need for `CanGetFromCache` and `CanPutToCache`, but I still think it was better to do it in a dedicated PR.

PR #1789, opened 2018-07-10
Fix shortcomings of QueryBatcher initial implementation

fredericDelaporte:
* AutoFlush is handled too late, causing stale data to be yielded
 * Cacheable result transformation is not fully done
 * Auto discovery of types fails
 * InitializeEntitiesAndCollections may be wrongly skipped or called
 * DoExecute should be renamed ExecuteBatched
 * BeginProcess should span the whole processing

 Follow up to #1718. See its reviews for more details on the above shortcomings.


fredericDelaporte:
#1788 already contains all these changes, but they should have their own PR I think.

I will rebase #1788 once this gets merged.

src/NHibernate/Multi/QueryBatch.cs
```diff
@@ -161,11 +166,7 @@ protected void DoExecute()
 					{
 						foreach (var multiSource in _queries)
 						{
-							foreach (var resultSetHandler in multiSource.GetResultSetHandler())
-							{
-								rowCount += resultSetHandler(reader);
-								reader.NextResult();
```
bahusoid: I was trying to avoid code duplication in `IQueryBatchItem`  implementations  and from this point -  `reader.NextResult` call should be responsibility of `IQueryBatch' implemention. And I've made [suggestion](https://github.com/nhibernate/nhibernate-core/pull/1718#discussion_r201482522) that should allow to keep old interface/behavior intact.
But it seems your refactoring fixed one more issue - no async counterpart was generated for old `GetResultSetHandler` method.
So this implementation might be a better option as it's suppored by async generator.

src/NHibernate.Test/Futures/QueryBatchFixture.cs
```diff
@@ -266,6 +266,116 @@ public void FutureForEagerMappedCollection()
...
 
+		[Test]
+		public void AutoDiscoverWorksWithFuture()
+		{
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				var future =
+					s
+						.CreateSQLQuery("select count(*) as count from EntitySimpleChild where Name like :pattern")
```
hazzik: Firebird does not like `as count` part.

src/NHibernate/Multi/QueryBatch.cs
fredericDelaporte: Well, about `NextResult`, I understand it could be considered the responsibility of the batcher. But at the same time, that is the queries who know how many result sets they have to consume. So there is no clear cut on that subject from my point of view.

hazzik:
From the release management perspective. I'm not sure if we need to display this issue in the release log. Thoughts?

fredericDelaporte:
I am agree. (But I am quite sure there are a bunch of similar cases, fixing changes before they were released.)

So flagging as fix.

PR #1790, opened 2018-07-11
Fix SQLite/SqlCE validation failure on tables with a schema

### LINKED ISSUE #1609: Schema validation using SQLite/SqlCe and a specific schema fails
Schema validation (`NHibernateSchemaValidator.Validate()`) using SQLite (or SQL Server CE) and a specific schema fails for a generated/exported schema (`NHibernateSchemaExport.Create(false, true)`).

**Example which works:**
schema name: [n/a]
entity name: "TestEntity"

exported result:
MsSql: "dbo.TestEntity"
SQLite: "TestEntity"

No `NHibernate.SchemaValidationException`. 

**Example which fails:**
schema name: "Test"
entity name: "TestEntity"

exported result:
MsSql: "Test.TestEntity"
SQLite: "Test_TestEntity"

`NHibernate.SchemaValidationException` for SQLite (MsSql works fine):

```
Message
-------
Schema validation failed: see list of validation errors

Source
------
NHibernate

TargetSite
----------
Void ValidateSchema(NHibernate.Dialect.Dialect, NHibernate.Tool.hbm2ddl.IDatabaseMetadata)

StackTrace
----------
   bei NHibernate.Cfg.Configuration.ValidateSchema(Dialect dialect, IDatabaseMetadata databaseMetadata)
   bei NHibernate.Tool.hbm2ddl.SchemaValidator.Validate()
   ...

HResult (System.Int32)
----------------------
-2146233088

ValidationErrors (System.Collections.ObjectModel.ReadOnlyCollection`1[System.String])
-------------------------------------------------------------------------------------
0: System.String: Missing table: TestEntity
```

Somehow the specified schema gets ignored doing the validation.

Can so far be reproduced with NHibernate 4.x to 5.0.3 (never tried that with earlier versions).

fredericDelaporte:
Fixes #1609 

See [the issue comments](https://github.com/nhibernate/nhibernate-core/issues/1609#issuecomment-404120657), that is hacks all the way for fixing a hack from 2008. But reverting this ten years old hack is likely to be a big breaking change fro SQLite users, so well, I think it is still better to fix it that way.


src/NHibernate.Test/Tools/hbm2ddl/SchemaValidator/SchemaValidateTableWithSchemaFixture.cs
```diff
@@ -0,0 +1,102 @@
...
+
+		protected override string[] Mappings => new[] { "Tools.hbm2ddl.SchemaValidator.VersionWithSchema.hbm.xml" };
+
+		protected override bool AppliesTo(Dialect.Dialect dialect)
+		{
+			switch (Dialect)
+			{
+				case MsSql2000Dialect _:
+				case PostgreSQLDialect _:
+				case SQLiteDialect _:
```
hazzik: SQL Ce uses the same pattern as SQLiteDialect

fredericDelaporte: Indeed, done two years later in e79f47e. Now I have checked `Dialect.Qualify` overrides, no other database does that, although Firebird and MySql do not support schema either.

(Firebird may be fixed by overriding `Qualify` for ignoring catalog and schema; I think it is better than introducing the hack in it. MySql should ideally be fixed for ignoring schema; but this may be a breaking change for people "benefiting" of the current behavior handling schema like a catalog (provided catalog is not supplied at the same time). Anyway both cases would have to be handled in another PRs.)

PR #1796, opened 2018-07-13
Support CacheMode in QueryBatch

fredericDelaporte:
And add a test for ReadOnly

### DIFF
--- a/src/NHibernate.Test/Async/Futures/QueryBatchFixture.cs
+++ b/src/NHibernate.Test/Async/Futures/QueryBatchFixture.cs
@@ -388,6 +388,116 @@ namespace NHibernate.Test.Futures
 			}
 		}
 
+		[Test]
+		public async Task ReadOnlyWorksWithFutureAsync()
+		{
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				var futureSimples =
+					s
+						.CreateQuery("from EntitySimpleChild")
+						.SetReadOnly(true)
+						.Future<EntitySimpleChild>();
+				var futureSubselect =
+					s
+						.CreateQuery("from EntitySubselectChild")
+						.Future<EntitySubselectChild>();
+
+				var simples = (await (futureSimples.GetEnumerableAsync())).ToList();
+				Assert.That(simples, Has.Count.GreaterThan(0));
+				foreach (var entity in simples)
+				{
+					Assert.That(s.IsReadOnly(entity), Is.True, entity.Name);
+				}
+
+				var subselect = (await (futureSubselect.GetEnumerableAsync())).ToList();
+				Assert.That(subselect, Has.Count.GreaterThan(0));
+				foreach (var entity in subselect)
+				{
+					Assert.That(s.IsReadOnly(entity), Is.False, entity.Name);
+				}
+
+				await (t.CommitAsync());
+			}
+		}
+
+		[Test]
+		public async Task CacheModeWorksWithFutureAsync()
+		{
+			Sfi.Statistics.IsStatisticsEnabled = true;
+
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				s
+					.CreateQuery("from EntitySimpleChild")
+					.SetCacheable(true)
+					.SetCacheMode(CacheMode.Get)
+					.Future<EntitySimpleChild>();
+				s
+					.CreateQuery("from EntityComplex")
+					.SetCacheable(true)
+					.SetCacheMode(CacheMode.Put)
+					.Future<EntityComplex>();
+				await (s
+					.CreateQuery("from EntitySubselectChild")
+					.SetCacheable(true)
+					.Future<EntitySubselectChild>()
+					.GetEnumerableAsync());
+				Assert.That(Sfi.Statistics.QueryCachePutCount, Is.EqualTo(2), "Future put");
+
+				await (t.CommitAsync());
+			}
+
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				Sfi.Statistics.Clear();
+				await (s
+					.CreateQuery("from EntitySimpleChild")
+					.SetCacheable(true)
+					.ListAsync<EntitySimpleChild>());
+				Assert.That(Sfi.Statistics.QueryCacheHitCount, Is.EqualTo(0), "EntitySimpleChild query hit");
+
+				Sfi.Statistics.Clear();
+				await (s
+					.CreateQuery("from EntityComplex")
+					.SetCacheable(true)
+					.ListAsync<EntityComplex>());
+				Assert.That(Sfi.Statistics.QueryCacheHitCount, Is.EqualTo(1), "EntityComplex query hit");
+
+				Sfi.Statistics.Clear();
+				await (s
+					.CreateQuery("from EntitySubselectChild")
+					.SetCacheable(true)
+					.ListAsync<EntitySubselectChild>());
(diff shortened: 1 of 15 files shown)


src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -171,80 +177,96 @@ public int ProcessResultsSet(DbDataReader reader)
...
+				// the backing field
+				_cacheMode != cacheModeOrig)
 			{
-				var queryInfo = _queryInfos[i];
-				var loader = queryInfo.Loader;
-				var queryParameters = queryInfo.Parameters;
-
-				//Skip processing for items already loaded from cache
-				if (queryInfo.IsResultFromCache)
+				Session.CacheMode = _cacheMode.Value;
```
bahusoid: As this code is used more than once and it's not really one-liner I would wrap it in `IDisposable` class. Something like:
```C#
class SessionContext : IDisposable
{
	private readonly ISession _session;
	private readonly CacheMode? _cacheModeToReturn;

	public SessionContext(ISession session, CacheMode? cacheMode)
	{
		if (cacheMode.HasValue && cacheMode != session.CacheMode)
		{
			_cacheModeToReturn = session.CacheMode;
			_session.CacheMode = cacheMode.Value;
		}
	}

	public void Dispose()
	{
		if(_cacheModeToReturn.HasValue)
			_session.CacheMode = _cacheModeToReturn.Value;
	}
}
```
So usage in code would be more concise:
```C#
using(new SessionContext(Session, _cacheMode)
...
```

I think  this class can also be made private in `QueryParameters` and exposed only via method. Something like: 
```C#
using(queryParams.FillSessionContext(session))
```
So it can be extended easily with filling some other context fields if added. But not sure if it's really usefull.

src/NHibernate/Engine/QueryParameters.cs
```diff
@@ -201,19 +202,31 @@ public void ValidateParameters()
...
-			            ProcessedSqlParameters = ProcessedSqlParameters != null ? ProcessedSqlParameters.ToList() : null
-			           };
+			var copy = new QueryParameters(
+				PositionalParameterTypes, PositionalParameterValues, NamedParameters, LockModes, selection,
+				IsReadOnlyInitialized, readOnly, Cacheable, CacheRegion, Comment, CollectionKeys, OptionalObject,
+				OptionalEntityName, OptionalId, ResultTransformer)
+			{
+				ProcessedSql = ProcessedSql,
+				ProcessedSqlParameters = ProcessedSqlParameters != null ? ProcessedSqlParameters.ToList() : null,
+				CacheMode = CacheMode
```
bahusoid: IMHO it's better to end with "," where allowed.
So it's less change noise when new initialization property is added. And sometimes it even allows to skip merge conflicts.

src/NHibernate/Async/Multi/QueryBatchItemBase.cs
```diff
@@ -35,80 +35,96 @@ public abstract partial class QueryBatchItemBase<TResult> : IQueryBatchItem<TRes
 			var dialect = Session.Factory.Dialect;
 			var hydratedObjects = new List<object>[_queryInfos.Count];
 
-			var rowCount = 0;
-			for (var i = 0; i < _queryInfos.Count; i++)
+			var cacheModeOrig = Session.CacheMode;
+			if (_cacheMode.HasValue &&
```
hazzik: Maybe move this code into a disposable object pattern?

src/NHibernate/Multi/QueryBatchItemBase.cs
hazzik: Agree.

src/NHibernate/Engine/QueryParameters.cs
fredericDelaporte: It does not look to me as an usual practice in NHibernate current code base, and while I understand the rationals of it, I still dislike it a bit. So I do not tend to follow this practice.

src/NHibernate/Engine/ISessionImplementor.cs
```diff
@@ -48,6 +48,41 @@ internal static void AutoFlushIfRequired(this ISessionImplementor implementor, I
...
+		/// <summary>
+		/// Switch the session current cache mode.
+		/// </summary>
+		/// <param name="session">The session for which the cache mode has to be switched.</param>
+		/// <param name="cacheMode">The desired cache mode. <see langword="null" /> for not actually switching.</param>
+		/// <returns><see langword="null" /> if no switch is required, otherwise an <see cref="IDisposable"/> which
+		/// dispose will set the session cache mode back to its original value.</returns>
+		internal static IDisposable SwitchCacheMode(this ISessionImplementor session, CacheMode? cacheMode)
+		{
+			if (session == null || !cacheMode.HasValue || cacheMode == session.CacheMode)
```
bahusoid: `session == null` check looks unnecessary in internal method. Unless it's a live scenario I think it's actually better to simply allow it to throw NRE on `session.CacheMode` access.

src/NHibernate.Test/Async/Futures/QueryBatchFixture.cs
```diff
@@ -388,6 +388,116 @@ public void CacheableCriteriaWithAliasedJoinFutureAsync()
...
+					.SetCacheable(true)
+					.ListAsync<EntitySimpleChild>());
+				Assert.That(Sfi.Statistics.QueryCacheHitCount, Is.EqualTo(0), "EntitySimpleChild query hit");
+
+				Sfi.Statistics.Clear();
+				await (s
+					.CreateQuery("from EntityComplex")
+					.SetCacheable(true)
+					.ListAsync<EntityComplex>());
+				Assert.That(Sfi.Statistics.QueryCacheHitCount, Is.EqualTo(1), "EntityComplex query hit");
```
bahusoid: Did you mean "query cache miss"? "query hit" is a bit unclear error message considering that in previous check above it's also used but for different condition.

src/NHibernate/Engine/ISessionImplementor.cs
fredericDelaporte: If it has to change I would instead throw an `ArgumentNullException`. I do not want to code NRE cases. Anyway that part will go away once merged into `ISessionImplementor`.

src/NHibernate.Test/Async/Futures/QueryBatchFixture.cs
fredericDelaporte: I generally use the label for identifying which assert has failed. That is the assert on query hit. I do not think this label has to describe what is the failure. That would be already interpreting it, and sometimes it could be off the actual cause.

src/NHibernate/Engine/ISessionImplementor.cs
bahusoid: Ah. Ok. I've missed the fact that you are actually planning to include this method in interface.

bahusoid: But then.. this `session == null` check only makes things worse. If there is such case in code - it will be silently ignored now. But when merged to interface - NRE will happen.

fredericDelaporte: That is true, but it is done also on `BeginContext` and  `BeginProcess` right above.

bahusoid: And is it really required to go in interface? It seems to me that it can leave as internal static extension. It doesn't depend on any specific implementation. So why to clutter interface with something that can be implemented on top of existing members?

fredericDelaporte: It is a possibility.

PR #1802, opened 2018-07-16
Generate stable sql objects names

### LINKED ISSUE #1769: Table mapping for UniqueColumn uses unstable GetHashCode() method
The snippet below presents the way how `Table` class generates `UniqueColumn` name. This value is used in several code paths, like `CreateForeignKey` and/or generating an unique column name itself. 

https://github.com/nhibernate/nhibernate-core/blob/c4fb3dc25f90d2370ca26eaaf4e5cb07bddce431/src/NHibernate/Mapping/Table.cs#L859-L872

`string.GetHashCode` is not a stable hashing function, according to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.string.gethashcode?view=netcore-2.1#System_String_GetHashCode)

> The hash code itself is not guaranteed to be stable. Hash codes for identical strings can differ across versions of the .NET Framework and across platforms (such as 32-bit and 64-bit) for a single version of the .NET Framework. In some cases, they can even differ by application domain. This implies two subsequent runs of the same program may return different hash codes. 

Creating the script then, can have a different outcome when run in two different environments, like `.NET Core 2.1` and `net461` or even across 32/64 bits.

fredericDelaporte:
Fix #1769.

Possible breaking change:

- Auto-generated constraint names will not be the same than the ones generated with previous NHibernate versions under .Net Framework. (Under .Net Core those names were anyway changing at each run.)

### DIFF
--- a/src/NHibernate/Mapping/Constraint.cs
+++ b/src/NHibernate/Mapping/Constraint.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using NHibernate.Engine;
 using NHibernate.Util;
@@ -61,14 +62,9 @@ namespace NHibernate.Mapping
 
 			// Ensure a consistent ordering of columns, regardless of the order
 			// they were bound.
-			// Clone the list, as sometimes a set of order-dependent Column
-			// bindings are given.
-			var alphabeticalColumns = new List<Column>(columns);
-			alphabeticalColumns.Sort(ColumnComparator.Instance);
-			foreach (var column in alphabeticalColumns)
+			foreach (var column in columns.OrderBy(c => c.CanonicalName))
 			{
-				var columnName = column == null ? "" : column.Name;
-				sb.Append("column`").Append(columnName).Append("`");
+				sb.Append("column`").Append(column.CanonicalName).Append("`");
 			}
 			// Hash the generated name for avoiding collisions with user choosen names.
 			// This is not 100% reliable, as hashing may still have a chance of generating
@@ -78,15 +74,6 @@ namespace NHibernate.Mapping
 			var name = prefix + Hasher.HashToString(sb.ToString());
 
 			return name;
-		}
-
-		private class ColumnComparator : IComparer<Column>
-		{
-			public static readonly ColumnComparator Instance = new ColumnComparator();
-
-			public int Compare(Column col1, Column col2) {
-				return StringComparer.Ordinal.Compare(col1?.Name, col2?.Name);
-			}
 		}
 
 		/// <summary>
--- a/src/NHibernate/Mapping/Constraint.cs
+++ b/src/NHibernate/Mapping/Constraint.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Security.Cryptography;
 using System.Text;
 using NHibernate.Engine;
 using NHibernate.Util;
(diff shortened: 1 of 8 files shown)


src/NHibernate/Mapping/Constraint.cs
```diff
@@ -38,6 +39,142 @@ public IEnumerable<Column> ColumnIterator
...
+			alphabeticalColumns.Sort(ColumnComparator.Instance);
+			foreach (var column in alphabeticalColumns)
+			{
+				var columnName = column == null ? "" : column.Name;
+				sb.Append("column`").Append(columnName).Append("`");
+			}
+			// Hash the generated name for avoiding collisions with user choosen names.
+			// This is not 100% reliable, as hashing may still have a chance of generating
+			// collisions.
+			var name = prefix + HashName(sb.ToString());
```
fredericDelaporte: I quite dislike the idea of "uniqueness through hashcode", since it is fundamentally flawed: collisions may occur, even if very unlikely. But I have no better idea here.

src/NHibernate/Mapping/Constraint.cs
```diff
@@ -38,6 +39,142 @@ public IEnumerable<Column> ColumnIterator
...
+
+			// Hibernate uses an algorithm yielding names shorter than 30 characters. But we cannot
+			// use it (see HashName). And also we have DB limited to even less (Informix)...
+			if (name.Length > 30)
+			{
+				// This, of course, increases the collision risk.
+				name = name.Substring(0, 30);
```
fredericDelaporte: Here I was tempted to use `Dialect.MaxAliasLength`. But that would require to partially undo #1703... Maybe instead should we add some setting for the max length here. After all we are not generating an alias. And a setting would be quite less heavy to resolve than the dialect.

src/NHibernate/Mapping/Constraint.cs
```diff
@@ -38,6 +39,142 @@ public IEnumerable<Column> ColumnIterator
...
+		/// <summary>
+		/// Hash a constraint name. Convert the hash digest to base 32
+		/// (full alphanumeric) for shortening the hash string representation
+		/// while keeping it suitable for db names.
+		/// </summary>
+		/// <param name="name">The name to be hashed.</param>
+		/// <returns>The hased name.</returns>
+		private static string HashName(string name)
+		{
+			// Hibernate uses MD5, but with .Net this would throw on FIPS enabled machine.
```
fredericDelaporte: This thing is from my point of view a very unfortunate design decision of the .Net Framework: if FIPS is enabled on the machine, any usage of non FIPS compliant algorithm is forbidden. But FIPS does not forbid them, it just forbids to use them for handling a security related task. And here we are not handling a security related task. But there is no way to tell it for allowing using MD5 here even on FIPS enabled machine. See [this Microsoft blog](https://blogs.technet.microsoft.com/secguide/2014/04/07/why-were-not-recommending-fips-mode-anymore/) (may be very long to load) for more information.

So if we want to use MD5 here without causing NHIbernate to be unusable on a FIPS enabled machine, we would have to use a custom implementation, like [this one](https://rosettacode.org/wiki/MD5/Implementation#C.23).

src/NHibernate/Mapping/Constraint.cs
```diff
@@ -38,6 +39,142 @@ public IEnumerable<Column> ColumnIterator
...
+			// As a consequence generated names will be quite longer.
+			using (var hasher = SHA256.Create())
```
fredericDelaporte: This one should pass on a FIPS enabled machine, `SHA256.Create` being documented as choosing a FIPS compliant one if needed. At first I was willing to use `HashAlgorithm.Create` instead (which default to SHA-1, a bit shorter), but it throws `PlatformNotSupported` under Core.

src/NHibernate/Mapping/Constraint.cs
```diff
@@ -38,6 +39,142 @@ public IEnumerable<Column> ColumnIterator
...
+			{
+				var hash = hasher.ComputeHash(Encoding.UTF8.GetBytes(name));
+				// Converting to base 32 for shortening the name.
+				// Hibernate uses base 35, but we do not have a native implementation
+				// in .Net, and base 32 is easier to implement.
+				return ToBase32String(hash);
```
fredericDelaporte: Hibernate creates a `BigInteger` from the byte array (.Net can do that too) then asks its base 35 representation: .Net does not have this natively, and implementing it was a bit heavy in my opinion.

hazzik:
I think using a cryptographic hash for this trivial task is an overkill. I was researching using lightweight non-cryptographic hashes to perform the operation, like [MurmurHash2](https://en.wikipedia.org/wiki/MurmurHash) and [xxHash](https://github.com/Cyan4973/xxHash). xx claims to be the fastest. They both generate 32 bit integer as a result and provide 10/10 results on SMHasher. I think it should be enough for us.

fredericDelaporte:
For avoiding adding a new dependency, we could take [xxHash](https://github.com/noricube/xxHashSharp/blob/master/xxHashSharp/xxHash.cs) version from Seok-Ju Yun by taking just the part of the source we need. The speed claim is about C version, so it may not be faster than [MurMurHash2](http://landman-code.blogspot.com/2009/02/c-superfasthash-and-murmurhash2.html) unsafe or hackish versions from David Landmann (which we could also incorporate as source code).

But for the usage it will have, mapping compilation and only on some foreign keys and unique constraints, I do not think it will make much differences.

hazzik:
I benchmarked the implementations:

``` ini

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i7-2670QM CPU 2.20GHz (Sandy Bridge), 1 CPU, 8 logical and 4 physical cores
Frequency=2143565 Hz, Resolution=466.5126 ns, Timer=TSC
.NET Core SDK=2.1.302
  [Host] : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT  [AttachedDebugger]
  Clr    : .NET Framework 4.7 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3131.0
  Core   : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT


```
|                         Method |  Job | Runtime |     N |        Mean |      Error |     StdDev | Rank |
|------------------------------- |----- |-------- |------ |------------:|-----------:|-----------:|-----:|
|             **_MurmurHash2Simple** |  **Clr** |     **Clr** |  **1000** |  **1,969.5 ns** |  **15.228 ns** |  **14.244 ns** |    **7** |
| _MurmurHash2InlineBitConverter |  Clr |     Clr |  1000 |  1,272.0 ns |   6.131 ns |   5.735 ns |    5 |
|             _MurmurHash2Unsafe |  Clr |     Clr |  1000 |    390.6 ns |   2.917 ns |   2.729 ns |    2 |
|                        _xxHash |  Clr |     Clr |  1000 |  2,832.5 ns |  26.088 ns |  24.402 ns |    8 |
|             _MurmurHash2Simple | Core |    Core |  1000 |    814.0 ns |   3.734 ns |   3.492 ns |    3 |
| _MurmurHash2InlineBitConverter | Core |    Core |  1000 |  1,266.1 ns |   7.694 ns |   7.197 ns |    4 |
|             _MurmurHash2Unsafe | Core |    Core |  1000 |    387.6 ns |   2.836 ns |   2.653 ns |    1 |
|                        _xxHash | Core |    Core |  1000 |  1,693.8 ns |  11.788 ns |   9.203 ns |    6 |
|             **_MurmurHash2Simple** |  **Clr** |     **Clr** | **10000** | **19,655.9 ns** | **133.815 ns** | **118.623 ns** |   **13** |
| _MurmurHash2InlineBitConverter |  Clr |     Clr | 10000 | 12,530.1 ns |  82.511 ns |  77.181 ns |   11 |
|             _MurmurHash2Unsafe |  Clr |     Clr | 10000 |  3,800.7 ns |  17.604 ns |  15.605 ns |    9 |
|                        _xxHash |  Clr |     Clr | 10000 | 28,462.9 ns | 270.660 ns | 239.933 ns |   14 |
|             _MurmurHash2Simple | Core |    Core | 10000 |  8,165.6 ns | 243.565 ns | 239.213 ns |   10 |
| _MurmurHash2InlineBitConverter | Core |    Core | 10000 | 12,558.2 ns |  88.049 ns |  82.361 ns |   11 |
|             _MurmurHash2Unsafe | Core |    Core | 10000 |  3,815.4 ns |  63.321 ns |  59.231 ns |    9 |
|                        _xxHash | Core |    Core | 10000 | 16,889.6 ns | 112.653 ns |  99.864 ns |   12 |

hazzik:
Let's stick with MurMurHash2Simple

src/NHibernate/Mapping/Constraint.cs
```diff
@@ -38,6 +38,57 @@ public IEnumerable<Column> ColumnIterator
...
+
+			return name;
+		}
+
+		private class ColumnComparator : IComparer<Column>
+		{
+			public static readonly ColumnComparator Instance = new ColumnComparator();
+
+			public int Compare(Column col1, Column col2) {
+				return StringComparer.Ordinal.Compare(col1?.Name, col2?.Name);
```
hazzik: This should compare on `Column.CanonicalName`. The `Name` of a Column class is case-sensitive when quoted and case-insensitive when not quoted.

src/NHibernate/Mapping/Constraint.cs
```diff
@@ -38,6 +38,57 @@ public IEnumerable<Column> ColumnIterator
...
+			// Ensure a consistent ordering of columns, regardless of the order
+			// they were bound.
+			// Clone the list, as sometimes a set of order-dependent Column
+			// bindings are given.
+			var alphabeticalColumns = new List<Column>(columns);
+			alphabeticalColumns.Sort(ColumnComparator.Instance);
+			foreach (var column in alphabeticalColumns)
+			{
+				var columnName = column == null ? "" : column.Name;
+				sb.Append("column`").Append(columnName).Append("`");
```
hazzik: Again, here should be CanonicalName prob.

src/NHibernate/Mapping/Constraint.cs
fredericDelaporte: No, it depends on database. SQL Server 2017 keeps respecting the collation of the database (in fact, of its metadata tables), quoted or not.
It may also be case sensitive when unquoted, not only when quoted. And this varies from databases to databases.

Quoted identifier being case sensitive is an Oracle feature. So I think it is better taking the name "as is", relying on the user to map its objects in a consistent way.

bahusoid:
Maybe we should include in generated key some truncated real parts like table and/or column names to minimize collisions risks on large databases and make generated keys more "user friendly"
Something like "FK_TABLEX_3B355A0C"

hazzik:
User has an ability to specify a name, so I would not bother

PR #1808, opened 2018-07-18
Support mixed formulas and columns

fredericDelaporte:
Many mapping elements allow to define their columns or formulas by nested elements, but most of them do not support having both nested columns and formulas.

Follow-up to [NH-2117](https://nhibernate.jira.com/browse/NH-2117), which has added this support to  one-to-many.

Why not qualifying this as a bug by the way, since nothing tells anywhere columns and formulas are not to be mixed.
A workaround could be in some cases to use formulas for columns too, but that can cause troubles to the DDL generation, as it cannot detect columns needing to be created when they are mapped as formulas.

This issue was spotted while checking [this example](https://docs.jboss.org/hibernate/orm/3.5/reference/en-US/html/example-mappings.html#example-mappings-composite-key-manytomany).

This PR is #1759 rebased and squashed without the mapping-by-code part, since this part has some debate on its method naming, causing it to be blocked.

Possible breaking change:
- Mappings mixing column elements and formula elements were taking into account only the formula elements. They will now take into account all elements.
- Mappings mixing column elements and/or formula elements with a column attribute or a formula attribute were silently ignoring the attribute. They will now throw.
- Mappings mixing a column attribute and a formula attribute were silently doing some best effort logic, either considering this as a two columns mapping, the second one being the formula (most cases), or only taking into account the formula (case of the `<element>` mapping). They will now throw.


src/NHibernate/Cfg/MappingSchema/HbmElement.cs
```diff
@@ -71,5 +71,24 @@ public HbmType Type
...
+		/// Columns and Formulas, in declared order
+		/// </summary>
+		[XmlIgnore]
+		public IEnumerable<object> ColumnsAndFormulas
+		{
+			get
+			{
+				if (Items != null)
+					return Items;
+				// Avoid a possible breaking change for mapping having left a column attribute along with a formula one.
```
hazzik: Is this logic deliberate? In `HbmManyToOne` currently is 
```csharp
if (Items != null)
	return Items;

return AsColumns()
	.Concat(AsFormulas().Cast<object>());
```
(if inline `Columns` and `Formulas` fields)

fredericDelaporte: Yes, and this was a possible breaking change at the time it was coded. Since it was not hard to avoid it and keep the previous behavior for `element` and other cases, I have done it that way.
This means keeping in place an inconsistency in behavior between `many-to-one` and others.

If instead we accept doing a possible breaking change here (for somewhat erroneous mappings anyway: it is not documented what happens when specifying both attributes; in my mind they ideally should have been mutually exclusive; people wanting to declare a column and a formula should use tags instead of attributes), we have three possibilities:

 - Align them all on the `many-to-one` behavior. This behavior causes a tag declaring both a `column` attribute and a `formula` attribute to be considered as spanning two columns, the first being a regular table column and the second being a formula. Why not, but this looks a bit arbitrary, nothing intuitive in knowing which one would be considered as the first or second.
- Align them all on this PR `element` behavior. This behavior causes a tag declaring both a `column` attribute and a `formula` attribute to be considered as declaring only a formula, ignoring the column. Why not, but likewise this looks a bit arbitrary, nothing intuitive in knowing which one would be taken or ignored.
- "Break" them all by throwing an explicit `MappingException` when both attributes are specified. To be consistent, we should then throw also if one of them is specified along with sub-tags. (Currently this case ignores the attributes.)

hazzik: Probably mapping exception is preferred option.

src/NHibernate.Test/NHSpecificTest/NH1007/Mappings.hbm.xml
```diff
@@ -6,18 +6,15 @@
 		<property name="Name" />
 	</class>
 
-	<!-- 
-		Mapping to demonstrate that the 'column' attribute takes precedence over a 'column' element - and, for consistency,
-		the same behaviour applies to the generator attribute/element
```
fredericDelaporte: Wrong for `column`: that was the element which was taking precedence.
([NH-1007](https://nhibernate.jira.com/browse/NH-1007) is about adding the `generator` attribute as an alternate to specifying the element, when the user just needs to specify the generator name.)

src/NHibernate.Test/NHSpecificTest/NH1007/Mappings.hbm.xml
```diff
@@ -6,18 +6,15 @@
...
+		Mapping to demonstrate that the 'generator' attribute takes precedence over a 'generator' element
 	-->
 	<class name="Employer2">
 		<id name="Id" generator="guid">
 			<generator class="invalid_generator_class" />
 		</id>
-		
-		<property name="Name" column="Name">
-			<column name="InvalidColumnName" />
-		</property>
```
fredericDelaporte: It was demonstrating nothing, since `InvalidColumnName` is a valid column name. And it was actually the name used in database.

hazzik:
What's the precedence of elements and tags in other cases (eg. generators, etc.)?

fredericDelaporte:
As implied by [this comment](https://github.com/nhibernate/nhibernate-core/pull/1808#discussion_r204439389), in the case of `generator`, the attribute takes precedence over the element. (It looks to me as if the implementor of NH-1007 was mistakenly believing this was the priority used for columns, although it seems column elements has always taken precedence over attributes.)

For `subselect`, which support has been added quite later ([NH-2602](https://nhibernate.jira.com/browse/NH-2602)), the attribute also takes precedence over the element.

As already written, for `column` (and for `formula`), the element takes precedence over the attribute, and this seems to be the case since the first versions of NHibernate.

hazzik:
Ok, can we investigate Hibernate then?

fredericDelaporte:
Hibernate does not support the `generator` attribute, it only supports the element.

For `subselect`, the attribute also takes precedence over the element.

For `column` and `formula`, that is even yet another rule: the `formula` attribute takes precedence over all, then `column` and `formula` elements are considered (so, these elements are taken into account if there is no `formula` attribute), and last the `column` attribute is used (so, it is used if there is neither a `formula` attribute nor `column` of `formula` elements).

But maybe there is some validity checks done before parsing that I have not spotted.

src/NHibernate/Cfg/MappingSchema/HbmManyToMany.cs
```diff
@@ -83,14 +83,16 @@ public IEnumerable<object> ColumnsAndFormulas
...
 			{
-				if (Items != null)
-					return Items;
-				// Avoid a possible breaking change for mapping having left a column attribute along with a formula one.
-				// This is a mapping error, but previous implementation was silently ignoring the column attribute in
-				// such case.
-				if (!string.IsNullOrEmpty(formula))
-					return AsFormulas();
-				return AsColumns();
+				if (Items != null && (!string.IsNullOrEmpty(column) || !string.IsNullOrEmpty(formula)))
```
hazzik: Not sure if this is possible, but probably need to check `Items` for `null` or empty.

fredericDelaporte: It seems unlikely the xml deserialization will change its behavior, but after all some framework implementations may differ there. And mapping-by-code may also differ someday (it is not the case currently), so for robustness, it would be better to change this.

src/NHibernate/Cfg/MappingSchema/HbmProperty.cs
```diff
@@ -104,5 +104,25 @@ public HbmType Type
...
+		/// </summary>
+		[XmlIgnore]
+		public IEnumerable<object> ColumnsAndFormulas
+		{
+			get
+			{
+				if (Items != null && (!string.IsNullOrEmpty(column) || !string.IsNullOrEmpty(formula)))
+					throw new MappingException(
+						$"On {Name} property: specifying columns or formulas with both attributes and " +
+						$"sub-elements is invalid. Please use only sub-elements, or only one of them as attribute");
```
hazzik: `$` is not necesary on the second line

hazzik:
There is some cleanup required regards using of  interpolated strings (`$`)

hazzik:
I’ll double check, but in principle - yes. I wanted to check something regarding the interfaces implemented (I think some of the classes are missing some interfaces, but I don’t remember the details)

fredericDelaporte:
Maybe you are thinking about one case which allows a nested `formula` or `column` element, but which indeed does not allow multiple ones, reason why I have not changed it. (I do not remember which case it was either, I just remember having seen it when I have checked all possible cases requiring this change.)

fredericDelaporte:
I have rechecked this, and spotted no forgotten case.

So,

PR #1803, opened 2018-07-18
Fix HQL and LINQ query by a type on <any/> with meta-types other than int

hazzik:
Fixes #1774

PR #1809, opened 2018-07-18
Update the <any /> mapping documentation

fredericDelaporte:
This documentation is obsolete since NH 2.0.

### DIFF
--- a/doc/reference/modules/basic_mapping.xml
+++ b/doc/reference/modules/basic_mapping.xml
@@ -3353,24 +3353,25 @@
                 example, for audit logs, user session data, etc.
             </para>
 
-            <programlisting><![CDATA[<any name="AnyEntity" id-type="Int64" meta-type="Eg.Custom.Class2TablenameType">
-    <column name="table_name"/>
+            <programlisting><![CDATA[<any name="AnyEntity" id-type="long" meta-type="class">
+    <column name="type"/>
     <column name="id"/>
 </any>]]></programlisting>
 
             <para>
                 The <literal>meta-type</literal> attribute allows the application to specify a custom type that maps
                 database column values to persistent classes that have identifier properties of the type specified by
-                <literal>id-type</literal>. If the meta-type returns instances of <literal>System.Type</literal>,
-                nothing else is required. On the other hand, if it is a basic type like <literal>String</literal> or
-                <literal>Char</literal>, you must specify the mapping from values to classes.
+                <literal>id-type</literal>. If the meta-type is <literal>class</literal>, nothing else is required.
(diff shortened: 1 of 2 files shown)


doc/reference/modules/basic_mapping.xml
```diff
@@ -3353,24 +3353,25 @@
                 example, for audit logs, user session data, etc.
             </para>
 
-            <programlisting><![CDATA[<any name="AnyEntity" id-type="Int64" meta-type="Eg.Custom.Class2TablenameType">
-    <column name="table_name"/>
+            <programlisting><![CDATA[<any name="AnyEntity" id-type="long" meta-type="class">
+    <column name="type"/>
```
fredericDelaporte: This does not have to be a table name.

doc/reference/modules/basic_mapping.xml
```diff
@@ -3353,24 +3353,25 @@
...
     <column name="id"/>
 </any>]]></programlisting>
 
             <para>
                 The <literal>meta-type</literal> attribute allows the application to specify a custom type that maps
                 database column values to persistent classes that have identifier properties of the type specified by
-                <literal>id-type</literal>. If the meta-type returns instances of <literal>System.Type</literal>,
-                nothing else is required. On the other hand, if it is a basic type like <literal>String</literal> or
-                <literal>Char</literal>, you must specify the mapping from values to classes.
+                <literal>id-type</literal>. If the meta-type is <literal>class</literal>, nothing else is required.
```
fredericDelaporte: Support of `TypeType` is gone since NH 2.0, as reported in [NH-1403](https://nhibernate.jira.com/browse/NH-1403). This has been fixed by "hijacking" the `TypeType` alias `class` for instead yielding `ClassMetaType` when `class` is used in the `meta-type` attribute of an `any` mapping (or `...-to-any`).
(This can be confusing for those who know `class` is `TypeType` alias, because `TypeType` persists the assembly qualified type name whereas `ClassMetaType` persists the full-name instead (no assembly in it).)

doc/reference/modules/basic_mapping.xml
```diff
@@ -3353,24 +3353,25 @@
...
+                The class full name will be persisted in the database as the type of the associated entity.
+                On the other hand, if it is a basic type like <literal>string</literal> or
+                <literal>char</literal>, you must specify the mapping from values to classes.
```
fredericDelaporte: This is not actually true: `string` and any string like types may work without meta-values. In such case they work somewhat similarly to `class`, but without a dedicated NHibernate `MetaType`. That causes issues, like those not fixed in #1803. So I have decide to keep that sentence.

If we fix this, this part would have to be updated.

PR #1811, opened 2018-07-19
Fix Informix Boolean Type

### LINKED ISSUE #1016: NH-3007 - Informix dialect generates incorrect boolean constants
**Alexander Bolotov** created an [issue](https://nhibernate.jira.com/browse/NH-3007) — <time datetime="2012-01-11T17:08:31.574Z" title="2012-01-11T17:08:31.574Z">11th January 2012, 17:08:31</time>:
> In NHibernate 3.2.0 GA, the InformixDialect (and its descendants as well) generates the SQL code that the Informix Dynamic Server 11.50 considers incorrect. The HQL query:
> 
> IQuery query = session.CreateQuery("from User u where u.IsLocked = true");
> 
> The generated SQL code:
> 
> select user0*.Id as Id0_, user0_.Name as Name0_, user0_.IsLocked as IsLocked0_ from User user0_ where user0*.IsLocked=t
> 
> The problem here is that the final "t" (or "f" if you put "false" in the HQL query) must be quoted according to Informix documentation (see <http://publib.boulder.ibm.com/infocenter/idshelp/v10/index.jsp?topic=/com.ibm.ddi.doc/ddi64.htm>) I am using Informix 11.50 and the native driver from Client SDK 3.50.TC9; cannot test it against other Informix versions, so I am not sure whether it this issue is 11.50-specific or applies to other versions as well.
> 
> 
---
**Alexander Bolotov** added a comment — <time datetime="2012-01-11T17:22:12.758Z" title="2012-01-11T17:22:12.758Z">11th January 2012, 17:22:12</time>:
> The unit test to reproduce the issue
>

hazzik:
Fixes #1016


src/NHibernate.Test/DialectTest/InformixDialectFixture.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+	[TestFixture]
+	public class InformixDialectFixture
+	{
+		private readonly InformixDialect _dialect = new InformixDialect();
+
+		[Test]
+		public void ToBooleanValueStringTrue()
+		{
+			Assert.That(_dialect.ToBooleanValueString(false), Is.EqualTo("'f'"));
```
fredericDelaporte: Shouldn't it be testing `true` here?

hazzik: Yes, it should. Fixed. My shift is over - I'm going to sleep.

Issue #1774, opened 2018-06-25
HQL and LINQ query by the type on <any/> with meta-type "string" fails


mishaurubkov:
HQL and LINQ query by **the type** on `<any/>` with `meta-type="string"` fails with `System.Data.SqlClient.SqlException : Invalid column name`. Looks like **the type** makes into SQL query text without single quotes.

To replicate the issue, follow the steps:

1. Map using `<any/>` with `meta-type="string"`:

>   <class name="ToyBox">
    <id name="Id">
      <generator class="native"/>
    </id>
    <property name="Name" />
    <any name="Shape" id-type="int" meta-type="string">
      <meta-value value="CircleType" class="Circle"/>
      <meta-value value="SquareType" class="Square"/>
      <column name="ShapeType" />
      <column name="ShapeId" />
    </any>
  </class>

2. Try to query **by type** using HQL:

`  var boxes = 	s.CreateQuery("from ToyBox t where t.Shape.class = Square").List<ToyBox>();`

or LINQ:

```
var boxes =
	(from t in s.Query<ToyBox>()
		where t.Shape is Square
		select t).ToList();
```

3. The Query will fail witch exception:

> NHibernate.Exceptions.GenericADOException : could not execute query
[ select toybox0_.Id as id1_0_, toybox0_.Name as name2_0_, toybox0_.ShapeType as shapetype3_0_, toybox0_.ShapeId as shapeid4_0_ from ToyBox toybox0_ where toybox0_.ShapeType=SquareType ]
[SQL: select toybox0_.Id as id1_0_, toybox0_.Name as name2_0_, toybox0_.ShapeType as shapetype3_0_, toybox0_.ShapeId as shapeid4_0_ from ToyBox toybox0_ where toybox0_.ShapeType=SquareType]
  ----> System.Data.SqlClient.SqlException : Invalid column name 'SquareType'.
   at NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters, IResultTransformer forcedResultTransformer)
   at NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters)
   at NHibernate.Loader.Loader.ListIgnoreQueryCache(ISessionImplementor session, QueryParameters queryParameters)
   at NHibernate.Loader.Loader.List(ISessionImplementor session, QueryParameters queryParameters, ISet`1 querySpaces)
[...]

PR #1812, opened 2018-07-19
Fix the <any/> handling without meta-values

fredericDelaporte:
When not specifying meta-values, while not using meta-type `class`, the `<any/>` type mapping is directly using the mapped `meta-type` for handling reading and writing the meta-value.

This causes issues like #1774. Moreover, if the type is not string based, the `AnyType` will fail reading it and writing into it will very likely fail too (due to cast to/from string).

When mapping an `any` without meta-values, the class full name is directly used as the meta-value, which is indeed the behavior of `ClassMetaType`. ~~So the fix is to use `ClassMetaType` in that case too, but with the mapped meta-type instead of the default `StringType`.~~

As pointed by Alexander, there is no need for a separated `ClassMetaType`: it should be merged into `MetaType` instead, allowing it to behave like `ClassMetaType` when no value map is provided.

### DIFF
--- a/src/NHibernate/Mapping/Any.cs
+++ b/src/NHibernate/Mapping/Any.cs
@@ -19,6 +19,7 @@ namespace NHibernate.Mapping
 
 		public Any(Table table) : base(table)
 		{
+			_type = GetLazyType();
 		}
 
 		/// <summary>
@@ -30,29 +31,45 @@ namespace NHibernate.Mapping
 			set { identifierTypeName = value; }
 		}
 
-		private IType type;
-		public override IType Type
+		private Lazy<IType> GetLazyType()
 		{
-			get
-			{
-				if (type == null)
+			return new Lazy<IType>(
+				() =>
 				{
-					type =
-						new AnyType(
-							metaValues == null
-							? ("class".Equals(metaTypeName) ? new ClassMetaType(): TypeFactory.HeuristicType(metaTypeName))
-								: new MetaType(metaValues, TypeFactory.HeuristicType(metaTypeName)),
-							TypeFactory.HeuristicType(identifierTypeName));
-				}
-				return type;
-			}
+					var identifierType = TypeFactory.HeuristicType(identifierTypeName);
+					if (identifierType == null)
+						throw new MappingException($"Identifier type '{identifierTypeName}' is invalid");
+
+					var hasMetaValues = metaValues != null && metaValues.Count > 0;
+
+					var metaType = !hasMetaValues && "class" == metaTypeName
+						? NHibernateUtil.String
+						: TypeFactory.HeuristicType(metaTypeName);
+
+					if (metaType == null)
+						throw new MappingException($"Meta type '{metaTypeName}' is invalid");
+
+					if (!hasMetaValues && metaType.ReturnedClass != typeof(string))
+						throw new MappingException(
+							$"Using a non-string meta type ('{metaTypeName}') without providing meta values is invalid");
+
+					return new AnyType(
+						new MetaType(metaValues, metaType),
+						identifierType);
+				});
 		}
+
+		// Types may be used by many threads, we must use a thread safe lazy initialization.
+		// (Or we should build this class in another way, but this would likely be a breaking change.)
+		private Lazy<IType> _type;
+
+		public override IType Type => _type.Value;
 
 		public void ResetCachedType()
 		{
 			// this is required if the user is programatically modifying the Any instance
 			// and need to reset the cached type instance;
-			type = null;
+			_type = GetLazyType();
 		}
 
 		public override void SetTypeUsingReflection(string className, string propertyName, string access)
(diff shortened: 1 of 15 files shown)


src/NHibernate/Mapping/Any.cs
```diff
@@ -30,29 +31,47 @@ public virtual string IdentifierTypeName
...
+						hasMetaValues
+							? new MetaType(metaValues, TypeFactory.HeuristicType(metaTypeName))
+							: (IType) new ClassMetaType(metaType),
+						identifierType);
+				});
 		}
 
+		// Types may be used by many threads, we must use a thread safe lazy initialization.
+		// (Or we should build this class in another way, but this would likely be a breaking change.)
+		private Lazy<IType> _type;
```
fredericDelaporte: Unrelated fix by the way. If done separately, it would have to wait for the merge here (or cause conflicts).

hazzik:
I was thinking about going opposite way: drop ClassMetaType and add support for it into MetaType directly - if meta values map is empty - than does not do the mapping.

hazzik:
Also, I think MetaType should have ReturnedClass of System.Type. And handle get/set similar to TypeType. In other words - I vastly disagree with the original change which brought us the ClassMetaType.

fredericDelaporte:
> I was thinking about going opposite way: drop ClassMetaType and add support for it into MetaType directly - if meta values map is empty - than does not do the mapping.

Why not.

> Also, I think MetaType should have ReturnedClass of System.Type. And handle get/set similar to TypeType.

On write it should not cause issues, but on read, we would need to resolve the type. With meta-values, it should be doable, but without them, we may have troubles due to the lack of assembly in the currently persisted type name. (And changing what is persisted would be a breaking change for all current data mapped with this.)

fredericDelaporte:
>> I was thinking about going opposite way: drop ClassMetaType and add support for it into MetaType directly - if meta values map is empty - than does not do the mapping.
>
>Why not.

This was done as a fixup commit in July.

>> Also, I think MetaType should have ReturnedClass of System.Type. And handle get/set similar to TypeType.
>
>On write it should not cause issues, but on read, we would need to resolve the type. With meta-values, it should be doable, but without them, we may have troubles due to the lack of assembly in the currently persisted type name. (And changing what is persisted would be a breaking change for all current data mapped with this.)

So it would be a breaking change to do these additional changes and I think we should not do them.

This PR also removes a public interface. But it is not a breaking change, because this interface is not yet released.

So,

src/NHibernate/Type/MetaType.cs
```diff
@@ -42,24 +57,29 @@ public override System.Type ReturnedClass
...
 
-		public override void NullSafeSet(DbCommand st,object value,int index,ISessionImplementor session)
+		public override void NullSafeSet(DbCommand st, object value, int index, ISessionImplementor session)
 		{
-			baseType.NullSafeSet(st, value == null ? null : keys[(string)value], index, session);
+			_baseType.NullSafeSet(
+				st,
+				value == null
+					? null
+					: _keys?[(string) value]?? value,
```
hazzik: Can `_keys[(string) value]` (no question mark) return null? I think there is a latent bug here.

fredericDelaporte: That would require the `values` `IDictionary<object, string>` supplied to constructor to contain a `null` key entry, which this interface is not supposed to support.

Conversely, we could think about the same question for ` _values[raw]`. But it cannot contains `null` values either, because otherwise building the reversed `_keys` dictionary in constructor would fail.

fredericDelaporte: My previous comment is valid for valid keys and values only. If an invalid key or value is supplied, it will yield null.

The previous code would have yield null in such case, while the new code will yield the invalid value (or invalid key for the Get case).

Supplying an invalid key is easy: just alter the data in the database. For this case, the code behavior change is a possible breaking change. An invalid key value in database with previous code is translated to a null association (for an any association). With the new code, it will cast the invalid key to string (and potentially already fail here) then use it as an entity name (and fail if it is not one).

Supplying an invalid value should also be easy: associate the any association with an entity not having a corresponding mapping in the association. There too, the new code could be consider as a possible breaking change: previously, it would have set the association as `null` in the database, ignoring the "unknown" entity. With the new code, it will attempt to save the entity name as being the meta-type key, and will fail if the underlying type is not compatible with a string.

The old behavior does not look very suitable to me: silently ignoring invalid values may be convenient in some cases, but it also causes errors to be harder to diagnose. Still that is an unrelated change.

fredericDelaporte: Well no, I am wrong with the second comment. `IDictionary<object, string>` does not yield `null` in case of missing key but throws, so my first comment holds true for both valid and invalid key or values.

fredericDelaporte:
Rebased, everything squashed with some comments added in code about the case requiring a bit too much thoughts for checking its validity.

PR #1818, opened 2018-08-04
Handle DbDataReaders that do not support GetSchemaTable

hazzik:
Some ADO.NET providers do not support `DbDataReader.GetSchemaTable` method which makes these providers completely unusable with NHibernate. This issue is relevant to the ADO.NET providers which target .NET Standard prior 2.0.

These issues were found when I was testing IBM.Data.DB2.Core, but they are not exclusive to this provider. The problem lies in the fact that [`DbDataReader.GetSchemaTable`](https://referencesource.microsoft.com/#System.Data/System/Data/Common/DbDataReader.cs,c89b8cb58d0d8a5c) throws `NotSupportedException`. As I understand this method was not in .NET Standard <2.0, and was brought there. But because one is not allowed to add `abstract` methods they added a virtual which throws `NotSupportedException`.

### DIFF
--- a/src/NHibernate/Driver/NDataReader.cs
+++ b/src/NHibernate/Driver/NDataReader.cs
@@ -488,6 +488,7 @@ namespace NHibernate.Driver
 			private int colCount = 0;
 
 			private DataTable schemaTable;
+			private NotSupportedException schemaTableNotSupportedException;
 
 			// key = field name
 			// index = field index
@@ -508,9 +509,11 @@ namespace NHibernate.Driver
 			/// </param>
 			internal static NResult Create(DbDataReader reader, bool isMidstream)
 			{
+				var schemaTable = SafeGetSchemaTable(reader, out var exception);
 				var result = new NResult
 				{
-					schemaTable = reader.GetSchemaTable()
+					schemaTable = schemaTable,
+					schemaTableNotSupportedException = exception
 				};
 
 				List<object[]> recordsList = new List<object[]>();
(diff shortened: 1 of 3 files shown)


src/NHibernate/Loader/Loader.cs
```diff
@@ -1424,7 +1424,7 @@ private ColumnNameCache RetreiveColumnNameToIndexCache(DbDataReader rs)
 			if (_columnNameCache == null)
 			{
 				Log.Debug("Building columnName->columnIndex cache");
-				_columnNameCache = new ColumnNameCache(rs.GetSchemaTable().Rows.Count);
+				_columnNameCache = new ColumnNameCache(rs.FieldCount);
```
hazzik: Here GetSchemaTable is completely unnecessary so replace with `FieldCount` which does the same as previous code. This is the main culprit of the issue.

src/NHibernate/Driver/NDataReader.cs
```diff
@@ -549,6 +552,21 @@ internal static NResult Create(DbDataReader reader, bool isMidstream)
 				return result;
 			}
```
hazzik: Here we want to delay throwing NotSupportedException until when someone actually calls `GetSchemaTable`.

hazzik:
The issues were found when I was doing DB2 fixes, however these fixes are not exclusive to DB2, so I decided to make a separate PR.

src/NHibernate/Driver/NDataReader.cs
```diff
@@ -549,6 +552,21 @@ internal static NResult Create(DbDataReader reader, bool isMidstream)
...
+			private static DataTable SafeGetSchemaTable(IDataReader reader, out NotSupportedException exception)
+			{
+				exception = null;
+				try
+				{
+					return reader.GetSchemaTable();
+				}
+				catch (NotSupportedException e)
+				{
+					exception = e;
```
fredericDelaporte: Adding a `ReflectHelper.PreserveStackTrace(e);` here would be better in my opinion. This way, if re-thrown later, the stack trace will not be overridden. So this would reduce the likeliness of users asking here about this lack of support.

PR #1820, opened 2018-08-06
Move WcfOperationSessionContext into a separate project

mattzink:
NHibernate currently references the System.ServiceModel.Primitives package, which along with it's own dependencies, adds ~4 MB to a published .NET Core application. The reference is only used to provide WcfOperationSessionContext. This PR creates a new project and NuGet package, NHibernate.Wcf, and moves WcfOperationSessionContext and the System.ServiceModel.Primitives reference accordingly, allowing users to chose when to opt-in to the additional dependencies.

### DIFF
--- a/src/NHibernate.Wcf/Context/WcfOperationSessionContext.cs
+++ b/src/NHibernate.Wcf/Context/WcfOperationSessionContext.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections;
 using System.ServiceModel;
 
--- a/src/NHibernate.Wcf/Context/WcfOperationSessionContext.cs
+++ b/src/NHibernate.Wcf/Context/WcfOperationSessionContext.cs
@@ -1,4 +1,5 @@
 using System.Collections;
+using System.Runtime.CompilerServices;
 using System.ServiceModel;
 
 using NHibernate.Engine;
@@ -9,6 +10,7 @@ namespace NHibernate.Context
 	/// Provides a <see cref="ISessionFactory.GetCurrentSession()">current session</see>
 	/// for the current OperationContext in WCF. Works only during the lifetime of a WCF operation.
 	/// </summary>
+	[TypeForwardedFrom("NHibernate, Version=5.1.0.0, Culture=neutral, PublicKeyToken=aa95f207798dfdb4")]
 	public class WcfOperationSessionContext : MapBasedSessionContext
 	{
 		public WcfOperationSessionContext(ISessionFactoryImplementor factory) : base(factory) {}
--- a/src/NHibernate.Wcf/NHibernate.Wcf.csproj
+++ b/src/NHibernate.Wcf/NHibernate.Wcf.csproj
@@ -1,0 +1,53 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="../../build-common/NHibernate.props" />
+
+  <PropertyGroup>
+    <RootNamespace>NHibernate</RootNamespace>
+    <Description>WCF session context support for NHibernate.</Description>
+    <TargetFrameworks>$(NhLibTargetFrameworks)</TargetFrameworks>
+    <NoWarn>$(NoWarn);3001;3002;3003;3005;1591;419</NoWarn>
+    <SignAssembly>True</SignAssembly>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <AssemblyOriginatorKeyFile>..\NHibernate.snk</AssemblyOriginatorKeyFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+
+    <PackageDescription>NHibernate is a mature, open source object-relational mapper for the .NET framework. It is actively developed, fully featured and used in thousands of successful projects.</PackageDescription>
+    <PackageTags>ORM; O/RM; DataBase; DAL; ObjectRelationalMapping; NHibernate; ADO.Net; Core; WCF</PackageTags>
+  </PropertyGroup>
+  
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net461'">
+    <DefineConstants>NETFX;$(DefineConstants)</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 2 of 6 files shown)


mattzink:
See https://groups.google.com/d/topic/nhibernate-development/r1tpUTii80s/discussion for related discussion

hazzik:
Did not look into much details yet.

Probably need to play with TypeForwardedTo/TypeForwardedFrom from [attributes](https://docs.microsoft.com/en-us/dotnet/framework/app-domains/type-forwarding-in-the-common-language-runtime). Also, need to be able to support users who (for some reasons - most likely FNH) used a full context name instead of an alias.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1276,7 +1276,13 @@ private ICurrentSessionContext BuildCurrentSessionContext()
...
 					return new WebSessionContext(this);
 				case "wcf_operation":
-					return new WcfOperationSessionContext(this);
+					// The WCF session context used to be included in this assembly,
+					// but now is in an optional external assembly, so to maintain
+					// compatibility we map the well-known name to a assembly qualified
+					// type name so it can be loaded via reflection below.
+					var assemblyName = typeof(SessionFactoryImpl).Assembly.FullName.Replace("NHibernate,", "NHibernate.Wcf,");
+					impl = "NHibernate.Context.WcfOperationSessionContext, " + assemblyName;
+					break;
```
hazzik: Below (int try-catch) we should also handle `"NHibernate.Context.WcfOperationSessionContext, NHibernate, xxxx"

mattzink: Good idea. Fixed.

fredericDelaporte:
@Tasteful, this code dates back to 186eb7e, almost twelve years ago. We will not change it in this PR, as it is a change unrelated to this old behavior. Better open an issue or propose a PR if you wish this current error handling to be changed.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1275,12 +1275,21 @@ private ICurrentSessionContext BuildCurrentSessionContext()
...
+				// The WcfOperationSessionContext type that was previously included in this
+				// assembly is now in an optional external assembly, so to maintain
+				// compatibility we map the old alias and type names to the updated type name.
+				const string wcfTypeName = "NHibernate.Context.WcfOperationSessionContext";
+				if (impl.Equals("wcf_operation", StringComparison.Ordinal) ||
+					impl.StartsWith(wcfTypeName, StringComparison.OrdinalIgnoreCase))
+				{
+					var wcfAssemblyName = typeof(SessionFactoryImpl).Assembly.FullName.Replace("NHibernate,", "NHibernate.Wcf,");
+					impl = string.Concat(wcfTypeName, ", ", wcfAssemblyName);
+				}
```
fredericDelaporte: We likely need a dedicated error handling for this case, at least for hinting at referencing the new library. Maybe should we add something like `isWcfOperationContext = true` here then test in the catch for this condition in order to add in the error message something like `(Is a reference to NHibernate.Wcf assembly missing?)`.

src/NHibernate.Wcf/Context/WcfOperationSessionContext.cs
```diff
@@ -10,6 +10,7 @@ namespace NHibernate.Context
 	/// Provides a <see cref="ISessionFactory.GetCurrentSession()">current session</see>
 	/// for the current OperationContext in WCF. Works only during the lifetime of a WCF operation.
 	/// </summary>
+	[TypeForwardedFrom("NHibernate, Version=5.1.0.0, Culture=neutral, PublicKeyToken=aa95f207798dfdb4")]
```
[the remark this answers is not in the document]
shatl: Should be a build step to update this

src/NHibernate.Wcf/NHibernate.Wcf.csproj
```diff
@@ -0,0 +1,53 @@
...
+    <RootNamespace>NHibernate</RootNamespace>
+    <Description>WCF session context support for NHibernate.</Description>
+    <TargetFrameworks>$(NhLibTargetFrameworks)</TargetFrameworks>
+    <NoWarn>$(NoWarn);3001;3002;3003;3005;1591;419</NoWarn>
+    <SignAssembly>True</SignAssembly>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <AssemblyOriginatorKeyFile>..\NHibernate.snk</AssemblyOriginatorKeyFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+
+    <PackageDescription>NHibernate is a mature, open source object-relational mapper for the .NET framework. It is actively developed, fully featured and used in thousands of successful projects.</PackageDescription>
```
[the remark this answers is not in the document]
hazzik: Yes, it should

fredericDelaporte:
I have implemented a test in NHibernate.Test (not yet pushed anywhere), and discovered by doing so that WCF server is not available under .Net Core. It seems it is only available under .Net Framework.

See dotnet/wcf#1200.

As I have only little experience with WCF, I wonder, does it make sens to provide a `WcfOperation` session context under .Net Standard or .Net Core in such case? Can this session context be used on client side? From my understanding of operation context, I expect no. It should be null on client side, causing the context to unusable. And I have tested it, in my test it is null on client side.

(My test is a bit contrived: the setup launch a server exposing a method actually testing the session context, the test itself use a client to connect to the self-launched server and run its method. All that in the same process. In the test itself, the operation context is null. In the server method, it is available.)

So then, there is no point to provide it under .Net Core. Under .Net Standard, it would allow feature parity for those using the .Net Standard library in a .Net Framework application. But is it worth it?

Personally, I vote for making the `WcfOperationSessionContext` a .Net Framework only thing. (`#ifdef NETFX` it)

@hazzik, any thoughts?

hazzik:
I tend to agree.

hazzik:
Superseded by #1842

hazzik:
Please create a separate repository for NHibernate.Wcf if needed.

PR #1830, opened 2018-08-22
Add Linux support to debug

lillo42:
Hi, 
In this PR I add linux support to debug.

I've tested on Ubuntu with IDE Jetbrains Rider, I tried to test on Fedora but I had some problem, the test neve begin. Ubuntu it's OK.

- Add ShowBuildMenu.sh
- Modify BuildTool
- Modify NHibernate.props
- Modify ShowBuildMenu.bat

I create a new file ShowBuildMenu.sh , this file is similar to ShowBuildMenu.bat, but with less function, today with this file is possible create and activated a new Database, build and execute test

I recreated BuildTool to .NET Core and change ShowBuildMenu.bat to use this project, now to debug on Windows is necessary .NET Core installed  

I change NHibernate.props to add and change NhAppTargetFrameworks and NhLibTargetFrameworks


build-common/NHibernate.props
```diff
@@ -10,8 +10,10 @@
     <AssemblyVersion>$(VersionMajor).$(VersionMinor).0.0</AssemblyVersion>
     <FileVersion>$(VersionPrefix).0</FileVersion>
 
-    <NhAppTargetFrameworks>net461;netcoreapp2.0</NhAppTargetFrameworks>
-    <NhLibTargetFrameworks>net461;netcoreapp2.0;netstandard2.0</NhLibTargetFrameworks>
+    <NhAppTargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net461;netcoreapp2.0</NhAppTargetFrameworks>
```
hazzik: What is the point of this change? I can debug net461 without problems on my non windows laptop.

lillo42: Really? Every time I try to build Nhiberte fail, says I don’t have .Net Framework 4.6.1

Have you installed Mono on your laptop? I didn’t try to install mono

lillo42: To debug NHibernate on Linux , Should be necessary install mono?

src/NHibernate/NHibernate.csproj
```diff
@@ -52,15 +52,11 @@
...
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
-    <PackageReference Include="System.ServiceModel.Primitives" Version="4.4.0" />
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.4.1" />
-    <PackageReference Include="System.Security.Permissions" Version="4.4.1" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="System.ServiceModel.Primitives" Version="4.4.0" />
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.4.1" />
-    <PackageReference Include="System.Security.Permissions" Version="4.4.1" />
```
fredericDelaporte: Changes to this files are undue. But merging master in this PR branch should remove them.

ShowBuildMenu.sh
```diff
@@ -0,0 +1,271 @@
...
+	echo  "A. (Step 1) Set up a new test configuration for a particular database."
+	echo  "B. (Step 2) Activate a test configuration."
+	echo  "C. (Step 3) Run tests."
+	echo  "."
+	echo  "--- BUILD ---"
+	echo  "D. Build NHibernate (Debug)"
+	echo  "E. Build NHibernate (Release)"
+	echo  "."
+	echo  "--- Exit ---"
+	echo  "X. Make the beautiful build menu go away."
```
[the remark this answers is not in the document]
hazzik: > Unfortunately it seems not.
@hazzik, are you able to run the generator under anything else than Windows?

Yes, I do it all the time.

.editorconfig
```diff
@@ -31,3 +31,6 @@ indent_size = 2
 [*.cshtml]
 indent_style = space
 indent_size = 4
+
+[*.sh]
+end_of_line = LF
```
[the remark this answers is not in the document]
hazzik: Yes, this needs to be in the .gitattributes, but not in for of text=auto (but rather an explicit)

ShowBuildMenu.sh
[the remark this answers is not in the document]
hazzik: I'll try on the latest version.

ShowBuildMenu.sh
```diff
@@ -0,0 +1,221 @@
...
+	LIB_FILES2=""
+	testSetupGeneric
+}
+
+testSetupMenu() {
+	echo "A. Add a test configuration for SQL Server."
+	echo "B. Add a test configuration for Firebird."
+	echo "C. Add a test configuration for SQLite."
+	echo "D. Add a test configuration for PostgreSQL."
+	echo "E. Add a test configuration for SQL Server Compact."
```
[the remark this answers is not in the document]
lillo42: I made a search if is possible run SQL Server CE on Linux, it isn't possible run SQL Server CE:
[Stack Overflow](https://stackoverflow.com/questions/2644464/can-i-use-sql-server-compact-edition-ce-on-mono)

ShowBuildMenu.sh
```diff
@@ -0,0 +1,221 @@
...
+	then
+		echo "SQL Server"
+		testSetupSqlServer
+		mainMenu
+	fi
+}
+
+testRun(){
+	eval "dotnet test ./src/NHibernate.Test/NHibernate.Test.csproj" -f netcoreapp2.0
+	eval "dotnet test ./src/NHibernate.Test.VisualBasic/NHibernate.Test.VisualBasic.vbproj" -f netcoreapp2.0
```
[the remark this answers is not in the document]
lillo42: Why? Isn't better run the tests with netcoreapp2.0 and net461?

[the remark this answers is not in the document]
fredericDelaporte: Mono has issues running all the NHibernate tests. Many of them always fail with Mono.

fredericDelaporte:
I have indeed installed mono-complete, I have not tried with just mono-devel.

ShowBuildMenu.sh
```diff
@@ -0,0 +1,207 @@
...
+	testSetupGeneric
+}
+
+testSetupMenu() {
+	echo "A. Add a test configuration for SQL Server."
+	echo "B. Add a test configuration for Firebird."
+	echo "C. Add a test configuration for SQLite."
+	echo "D. Add a test configuration for PostgreSQL."
+	echo "F. Add a test configuration for MySql."
+	echo "."
```
fredericDelaporte: The prompt is still prompting for `E`, which will now just fully exit the menu. Better switch MySQL to `E` and renumber test 6 and 5 to 5 and 4.
(Contributing do not mention letters of sub-menus. It mentions only those of the main menu. So for sub-menu, no need to keep them in sync with the .bat IMO.)

ShowBuildMenu.bat
```diff
@@ -2,14 +2,15 @@
 pushd %~dp0
 
 set NANT="%~dp0Tools\nant\bin\NAnt.exe" -t:net-4.0
-set BUILD_TOOL_PATH=%~dp0Tools\BuildTool\BuildTool.csproj
-set BUILD_TOOL_SLN_PATH=%~dp0Tools\BuildTool\BuildTool.sln
-set BUILDTOOL=dotnet run -p %BUILD_TOOL_PATH% -c Release --no-build
+set BUILD_TOOL_PATH=%~dp0Tools\BuildTool\bin\BuildTool.dll
+set BUILDTOOL=dotnet %BUILD_TOOL_PATH%
```
fredericDelaporte: Be it under Linux or Windows, I was noticing a lag in the build menu for displaying the letter choices prompt. And there was also a startup lag due to relaunching the build.

There was also ome differences in the way the build tool was handled between Linux and Windows, without me seeing why they would need to be differently handled.

So I have changed the logic for building only if the binary is not already there, and for running directly the binary (`dotnet pathOfBinary`) instead of running the project from sources (`dotnet run pathToSrc`).

This eliminates the lag in the build menu both under Linux and Windows.

For simplicity, I have redirected the build output directly in the bin folder instead of letting it go into `Release\netcoreapp2.0`.

fredericDelaporte:
I have added the async generator option.
I am still unable to get it run "out of the box" on my setup, but maybe someone else could see what is wrong, if anything, with the way I am launching it.
I have tried with versions 0.8.2.8, 0.8.2.9, 0.8.2.10 and 0.13.0, all do fail with the same error, being unable to load [SQLitePCLRaw.core](https://www.nuget.org/packages/SQLitePCLRaw.core/) dependency, seemingly coming from [Microsoft.CodeAnalysis.Workspaces.Common](https://www.nuget.org/packages/Microsoft.CodeAnalysis.Workspaces.Common), although the generator is supposed to depend on 2.7.0 version of Microsoft.CodeAnalysis.Workspaces.Common while this SQLite dependency seems to be there only since 2.9.2 version.

But when inspecting the 2.7.0 assembly with ILSpy, it actually already depends on it. Manually adding the assembly in the async generator folder solves the trouble.

So it seems this thing is a somewhat "rogue" dependency causing issues only on some setup (Mono version maybe?), and using latest Microsoft.CodeAnalysis.Workspaces.Common versions should solve it since it now explicitly depends on it.

For now I have added a hack for including it in the async generator directory.

Tools/.gitignore
```diff
@@ -2,4 +2,5 @@ nuget.exe
 NUnit.*
 vswhere.*
 CSharpAsyncGenerator.CommandLine.*
-gitreleasemanager.*
\ No newline at end of file
+gitreleasemanager.*
+SQLitePCLRaw.core.*
```
fredericDelaporte: Required for the Linux/Mono hack I have used to get async generator running.

ShowBuildMenu.sh
```diff
@@ -0,0 +1,261 @@
...
+	fi
+
+	cd Tools
+
+	if [ ! -f nuget.exe ]
+	then
+		wget https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
+	fi
+
+	async_generator_path="CSharpAsyncGenerator.CommandLine.$(cat packages.config | grep id=\"CSharpAsyncGenerator.CommandLine | cut -d\" -f4)/tools"
```
fredericDelaporte: Hack for parsing the async generator version out of the packages.config file:
 - Output the content of packages.config
 - Filter out any line not containing `id=\"CSharpAsyncGenerator.CommandLine`
 - Split the remaining line(s) by `"` and take the fourth string resulting of the split

If you know a better way, of course propose it.
I need it to build the path to the async generator version we have to use.

CONTRIBUTING.md
```diff
@@ -20,7 +20,7 @@ The main GitHub repository is at [https://github.com/nhibernate/nhibernate-core]
 
 ## The Build Menu
 
-**ShowBuildMenu.bat** will be your friend throughout this journey. He's easiest to work with if you make his box bigger by following these steps: 
+**ShowBuildMenu.bat** will be your friend throughout this journey. A .sh version exists for Linux. He's easiest to work with if you make his box bigger by following these steps:
```
hazzik: Why `ShowBuildMenu.bat` is referenced as "He"? Shouldn't it be "It"?

fredericDelaporte: Yes.

Issue #1828, opened 2018-08-21
Lazy loading of collections with wrong case in key - different behaviour using Get vs Query?


nicklascarlsson:
I'm receiving a request from another system where the key of an entity has the **wrong case** 
(e.g. Id = 'ABC', where it should be Id = 'Abc').

Fetching the entity from our database works fine, but regarding lazy loading of related collections I run into problems.

Seems like I get different behaviour when using 
**Session.Get<T>('ABC')**, vs 
**Session.Query<T>(x => x.Id == 'ABC').SingleOrDefault()**

When trying to fetch the related collection with Get (mapped as a Bag) into an IList, the collection isn't loaded. I can see in NHibernate Profiler that the sql query is run and 5 records are found, but the list remains empty.

When using Query, the sql query is run and the list is populated though.

I've check in earlier releases that we use (3.3.3.4001, 4.1.1.4000 and 5.1.1) and found the same behaviour.
Anyone have experienced the same behaviour, is there a logical explanation to this perhaps?

Regards
// Nicklas

fredericDelaporte:
It seems to me this trouble is already reported somewhere, but I am not finding it...

hazzik:
Try JIRA

fredericDelaporte:
I had searched there too.

nicklascarlsson:
After some searching I found this:
[https://nhibernate.jira.com/browse/NH-3833](url)
Is this the one you are thinking about @fredericDelaporte ?

The difference, I feel, is that I get different behaviour between Get and Query.
And in the latter, it actually works regardless of case.. :)

fredericDelaporte:
Indeed, it is not the same but related. The suggested solution in this Jira should work here too.

I guess that when using `Get` (and likely `Load` too), the id gets hydrated with the supplied argument, instead of being read from the database. The database being case insensitive in your case, it returns an entity, but probably hydrated by NHibernate with the wrongly cased identifier, causing later NHibernate to discard collection children having the right casing as foreign key, because comparison of string in .Net are case sensitive by default.

String types in NHibernate use the default comparison. So for solving this mismatch between comparisons in database and comparison in NHibernate, you need to map your identifier with a custom type having a case insensitive equality.

The query case works because it hydrates the identifier from the database, not from the wrongly cased supplied argument.

[...]

PR #1833, opened 2018-08-28
Parameterize string type comparer

fredericDelaporte:
In order to allow users handling cases where a string foreign key is case insensitive, allow setting the comparer for string types.

This allows built-in support for use cases like #1828 and [NH-3833](https://nhibernate.jira.com/browse/NH-3833).

This is an unlikely breaking change, because it causes string types to parse parameters if any is supplied. Anyone having mapped a string type with parameters matching the one parsed with this PR may see its mapping change behavior.


src/NHibernate/Type/NullableType.cs
```diff
@@ -332,11 +332,6 @@ public override bool[] ToColumnNullness(object value, IMapping mapping)
 			return value == null ? ArrayHelper.False : ArrayHelper.True;
 		}
 
-		public override bool IsEqual(object x, object y)
-		{
-			return Equals(x, y);
-		}
```
fredericDelaporte: This override has the exact same implementation as the overridden method, so removing. Moreover, overriding `IsEqual` without overriding `GetHashcode` is normally a serious bug. (But there, since the override was just redundant, it was not causing troubles.)

src/NHibernate/Type/TypeFactory.cs
```diff
@@ -386,6 +386,22 @@ private static TypeClassification GetTypeClassification(string typeName)
...
+		/// Int32, or even NHibernate.Type.DecimalType, NHibernate.Type.DecimalType(19,0),
+		/// NHibernate.Type.Int32Type, then return an instance of NHibernate.Type.IType
+		/// </summary>
+		/// <param name="name">The name of the type.</param>
+		/// <param name="parameters">The parameters for the type, if any.</param>
+		/// <returns>The instance of the IType that the string represents.</returns>
+		/// <remarks>
+		/// This method will return null if the name is not found in the basicNameMap.
+		/// </remarks>
+		public static IType Basic(string name, IDictionary<string, string> parameters)
```
fredericDelaporte: Support of parameters was lacking in case of built-in types. I was not expecting this, since `DateType` is parameterized... Its parameter could then only work by mapping the type with its .Net fullname instead of using its alias.

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -1,13 +1,42 @@
...
+		/// The default string comparer for string equality and hashcodes. <see langword="null" /> for
+		/// using .Net default equality and hashcode computation.
+		/// </summary>
+		public static StringComparer DefaultComparer { get; set; }
+
+		/// <summary>
+		/// The string comparer of this instance of string type, for string equality and hashcodes.
+		/// <see langword="null" /> for using <see cref="DefaultComparer"/>.
+		/// </summary>
+		protected StringComparer Comparer { get; set; }
```
fredericDelaporte: I have considered exposing this publicly, for allowing user to set the `Comparer` by code on a specific type instance. But since the type factory get type methods may yield cached instances, it would be a trap.

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -83,5 +122,78 @@ public string ObjectToSQLString(object value, Dialect.Dialect dialect)
...
+
+		/// <summary>
+		/// Determines whether the specified <see cref="Object"/> is equal to this
+		/// <see cref="AbstractStringType"/>.
+		/// </summary>
+		/// <param name="obj">The <see cref="Object"/> to compare with this <c>AbstractStringType</c>.</param>
+		/// <returns><see langword="true"/> if the SqlType, Name and Comparer properties are the same.</returns>
+		public override bool Equals(object obj)
+		{
+			if (!base.Equals(obj))
```
hazzik: Shouldn't it be opposite?

```csharp 
if (base.Equals(obj)) return true;
```

hazzik: Em... but you are saying exactly what I’m asking about. And your code does the opposite.

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -1,13 +1,42 @@
...
+		/// </summary>
+		/// <remarks>Default comparison is ordinal.</remarks>
+		public const string ComparerCultureParameterName = "ComparerCulture";
+
+		/// <summary>
+		/// The case sensitivity parameter name. Value should be a boolean, <c>false</c> meaning
+		/// case insensitive.
+		/// </summary>
+		/// <remarks>Default comparison is case sensitive.</remarks>
+		public const string CaseSensitiveParameterName = "CaseSensitive";
```
hazzik: `CaseSensitive` is a bad parameter because its default value is `true` which contradicts the default value of `Boolean`, so it can be confusing.

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -34,6 +63,16 @@ public override object Get(DbDataReader rs, string name, ISessionImplementor ses
 			return Convert.ToString(rs[name]);
 		}
 
+		public override bool IsEqual(object x, object y)
+		{
+			return (Comparer ?? DefaultComparer)?.Equals(x, y) ?? base.IsEqual(x, y);
```
hazzik: `null` comparer is the same as `Ordinal`, so it does not make sense to piling up null propagation and condition operators. It should be simplified to just `return Comparer.Equals(x, y)` (see notes on `Comparer` and `DefaultComparer` properties for context).

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -34,6 +63,16 @@ public override object Get(DbDataReader rs, string name, ISessionImplementor ses
...
+		}
+
+		public override int GetHashCode(object x)
+		{
+			return (Comparer ?? DefaultComparer)?.GetHashCode(x) ?? base.GetHashCode(x);
```
hazzik: Same as [`IsEqual`](https://github.com/nhibernate/nhibernate-core/pull/1833/files#r217241027).

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -1,13 +1,42 @@
...
+		/// case insensitive.
+		/// </summary>
+		/// <remarks>Default comparison is case sensitive.</remarks>
+		public const string CaseSensitiveParameterName = "CaseSensitive";
+
+		/// <summary>
+		/// The default string comparer for string equality and hashcodes. <see langword="null" /> for
+		/// using .Net default equality and hashcode computation.
+		/// </summary>
+		public static StringComparer DefaultComparer { get; set; }
```
hazzik: Lets make `DefaultComparer` not nullable and default to `StringComparer.Ordinal`, eg.:

```
public static StringComparer DefaultComparer
{
	get => _defaultComparer;
	set => _defaultComparer = value ?? throw new ArgumentNullException(nameof(value));
}
private static StringComparer _defaultComparer = StringComparer.Ordinal;
```

src/NHibernate/Type/AbstractStringType.cs
hazzik: The `Comparer` should be copy on write:
```csharp
protected StringComparer Comparer
{
	get => _comparer ?? _defaultComparer;
	set => _comparer = value;
}
private StringComparer _comparer;
```

src/NHibernate/Type/TypeFactory.cs
```diff
@@ -446,6 +471,9 @@ public static IType Basic(string name)
 				// doesn't have built into it.
 				return null;
 			}
+
+			InjectParameters(returnType, parameters);
```
hazzik: This should not be here as it might corrupt the shared type.

src/NHibernate/Type/AbstractStringType.cs
fredericDelaporte: And it does not play nice with the behavior of `boolean.TryParse`, if the culture is set.

fredericDelaporte:
Thanks for the PR.

bahusoid:
Guys could you please clarify how to use this feature in mapping by code? Case when you don't want to change behavior globally. I would expect to see some helper method like `NhibernateUtil.StringType(comparer, culture)`

fredericDelaporte:
It seems mapping by code does also directly support parameterized types. It should be something like `.Type<StringType>(new {IgnoreCase = true})` by example.

bahusoid:
@hazzik  Why do you hate mapping by code? :) 

IMHO this `IParameterizedType` is solely required for initialization from text paramters (like XML mapping). And in mapping by code we should use "full .NET power" and initialize objects directly to avoid unecessary parameters -> value conversion. So in my ideal world I would like to see something like:
```C#
mapper.Type(NHibernateUtil.StringType(StringComparer.OrdinalIgnoreCase));
```

I also noticed that `IIdMapper` interface doesn't provide method suggested by @fredericDelaporte. So I would say parameterized types are not fully supported . And this feature is actually useful for IDs.

hazzik:
>@hazzik Why do you hate mapping by code? :)

This is just a way to perceive it.

> I also noticed that IIdMapper interface doesn't provide method suggested by @fredericDelaporte. So I would say parameterized types are not fully supported .

This is why I wrote my comment — it is not supported everywhere, but there are work arounds (like one I suggested).

fredericDelaporte:
>IMHO this `IParameterizedType` is solely required for initialization from text paramters (like XML mapping). And in mapping by code we should use "full .NET power" and initialize objects directly to avoid unecessary parameters -> value conversion. So in my ideal world I would like to see something like:
>```C#
>mapper.Type(NHibernateUtil.StringType(StringComparer.OrdinalIgnoreCase));
>```

I do not view it that way. `IParameterizedType` is there to allow types to have arbitrary parameters. When looking at the code, it was clearly initially meant for custom types, allowing them to have whatever parameters they need without requiring code changes in NHibernate. Its support was also partially added to built-in types, with lacking pieces in their case, that this PR adds because it cannot work without them.

I would rather avoid adding strongly typed helpers, because that is adding new method signatures which may require changes in case we have new parameters to add. That is not due to a limitation of XML mapping: we can also strongly type things in it, through the XSD. There is no reason for having something strongly typed in mapping by code and not in the XML mapping. But that is more work and more code to maintain.

It is not yet proven this feature will be widely used. It has spurred out of only two issue reports, one there since 2015 and not having gathered many votes. For me it was still looking as a fundamental mismatch in some cases between DB behavior and NHibernate behavior, that is why I have considered it was deserving a built-in solution. But I do not expect it to have so many usages that it deserves more than a minimal built-in support. My stance there is "lets it get released and see how it is received".

> I also noticed that `IIdMapper` interface doesn't provide method suggested by @fredericDelaporte. So I would say parameterized types are not fully supported . And this feature is actually useful for IDs.

That is indeed its main use case. Not having it for other cases has minor consequences. (It may allow to avoid some string value updates when only the string case changes, but for me it is even undesirable in most cases.)
This is a lacking piece for mapping by code which should probably be addressed in its own PR.

bahusoid:
> This is a lacking piece for mapping by code which should probably be addressed in its own PR.
OK. 

> I do not view it that way. IParameterizedType is there to allow types to have arbitrary parameters. When looking at the code, it was clearly initially meant for custom types, allowing them to have whatever parameters they need without requiring code changes in NHibernate.

Well that would also be possible if custom types and typed parameters are exposed so user can instantiate them directly in code:
```C#
mapper.Type(new StringType{Comparer = StringComparer.OrdinalIgnoreCase});
```
I found inability to provide strongly typed parameters frustrating. Why `StringType/AnsiString` consturctors are internal? I would understand it if the whole type is internal and exposed only as some base type. So please consider to make them public (or at leaset internal protected) in another PR.

fredericDelaporte:
> I found inability to provide strongly typed parameters frustrating. Why `StringType/AnsiString` constructors are internal? I would understand it if the whole type is internal and exposed only as some base type. So please consider to make them public (or at least internal protected) in another PR.

I guess built-in types  (at least for some) were made that way to encourage the use of the type factory, which does cache them.
This caching is the reason why I have set the `Comparer` property as protected: otherwise users may get trapped by getting the type from the type factory then altering its comparer without realizing they are changing it for many other property mappings.
Types are also currently designed as immutable objects, at least from the user perspective. The `IParameterizedType` is not supposed to be called directly by the users, but only by NHibernate when instantiating the type.

PR #1837, opened 2018-09-04
Have more robust log4net loading

### LINKED ISSUE #1836: Cannot create configuration due to log4net loading failure
I'm trying to upgrade NH from 4.0.0 til 5.1.3. I have installed the proper Nuget packages and have not yet changed any existing code. This is the first (and right now, only) line of NHibernate-related code:

`var config = new NHibernate.Cfg.Configuration();`

After this, I plan to do different things to complete the configuration. However, the code fails at this line and I am unable to do any configuration.

I cloned the NH repo so I could debug the problem, and I cannot get past the line

`properties = Environment.Properties;`

in the `Reset()` method of the `NHibernate.Cfg.Configuration` class.

All projects in my solution target .NET 4.6.1. A thing to note is that my web projects are SharePoint (2016) projects, and the assemblies are therefore deployed to the GAC. I have confirmed that all the assemblies have been deployed correctly and that there are no versioning issues (only one version of each NHibernate-related assembly is deployed to the GAC). And just to be clear, everything worked without issue on the same setup with NH 4.0.0 (if I uninstall 5.1.3 and go back to 4.0.0, then everything works just fine again).

I have spent the last couple of days on this upgrade and I am getting no closer to a solution. I hope you can help me out.

The complete stack trace for the exception is:

```
System.TypeInitializationException: The type initializer for 'NHibernate.Cfg.Environment' threw an exception.
---> System.TypeInitializationException: The type initializer for 'NHibernate.NHibernateLogger' threw an exception.
---> NHibernate.InstantiationException: Unable to instantiate: NHibernate.Log4NetLoggerFactoryNHibernate.Log4NetLoggerFactory
---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
---> System.TypeInitializationException: The type initializer for 'NHibernate.Log4NetLoggerFactory' threw an exception.
---> System.NullReferenceException: Object reference not set to an instance of an object.
   at NHibernate.Log4NetLoggerFactory.GetGetLoggerByNameMethodCall()
   at NHibernate.Log4NetLoggerFactory..cctor()
   at NHibernate.Log4NetLoggerFactory..ctor()
   at System.RuntimeTypeHandle.CreateInstance(RuntimeType type, Boolean publicOnly, Boolean noCheck, Boolean& canBeCached, RuntimeMethodHandleInternal& ctor, Boolean& bNeedSecurityCheck)
   at System.RuntimeType.CreateInstanceSlow(Boolean publicOnly, Boolean skipCheckThis, Boolean fillCache, StackCrawlMark& stackMark)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
   at System.Activator.CreateInstance(Type type)
   at NHibernate.NHibernateLogger.GetLoggerFactory(String nhibernateLoggerClass)
   at NHibernate.NHibernateLogger.GetLoggerFactory(String nhibernateLoggerClass)
   at NHibernate.NHibernateLogger..cctor()
   at NHibernate.NHibernateLogger.For(Type type)
   at NHibernate.Cfg.Environment..cctor()
   at NHibernate.Cfg.Configuration.Reset()
   at MyCompany.MyProject.CreateNHibernateConfiguration()
```

fredericDelaporte:
In some circumstances, it appears the log4net loading mechanism may fail with null reference exceptions, which does not help understanding what is the trouble. (See #1836.)

Moreover, log4net loading is triggered after some probing logic which is in no way involved in how it is actually loaded, creating a possibility the probing considers it loadable while it is not.

Fixes #1836


src/NHibernate/Logging.cs
```diff
@@ -117,13 +117,8 @@ private static string GetNhibernateLoggerClass()
...
-				string relativeSearchPath = AppDomain.CurrentDomain.RelativeSearchPath;
-				string binPath = relativeSearchPath == null ? baseDir : Path.Combine(baseDir, relativeSearchPath);
-				string log4NetDllPath = binPath == null ? "log4net.dll" : Path.Combine(binPath, "log4net.dll");
-
-				if (File.Exists(log4NetDllPath) || AppDomain.CurrentDomain.GetAssemblies().Any(a => a.GetName().Name == "log4net"))
+				// look for log4net
+				if (System.Type.GetType("log4net.LogManager, log4net") != null)
 				{
 					nhibernateLoggerClass = typeof(Log4NetLoggerFactory).AssemblyQualifiedName;
 				}
```
fredericDelaporte: I am tempted to put a swallowing `try catch` there, but in case of failure we would have no way of knowing what happens excepted debugging. (No way to log anything in such a `catch` here, since that is the logging system which fails...)

src/NHibernate/Logging.log4net.cs
```diff
@@ -12,12 +12,13 @@ namespace NHibernate
...
-		private static readonly System.Type LogManagerType = System.Type.GetType("log4net.LogManager, log4net");
+		private static readonly System.Type LogManagerType;
 		private static readonly Func<Assembly, string, object> GetLoggerByNameDelegate;
 		private static readonly Func<System.Type, object> GetLoggerByTypeDelegate;
 
 		static Log4NetLoggerFactory()
 		{
+			LogManagerType = System.Type.GetType("log4net.LogManager, log4net", true);
 			GetLoggerByNameDelegate = GetGetLoggerByNameMethodCall();
 			GetLoggerByTypeDelegate = GetGetLoggerMethodCall<System.Type>();
```
fredericDelaporte: The method calls following the get type are using the type without checking if it was found, so better not allow `GetType` to yield `null`.

src/NHibernate/Logging.log4net.cs
```diff
@@ -92,29 +96,31 @@ public class Log4NetLogger : IInternalLogger
...
-			DebugFormatDelegate = DelegateHelper.BuildAction<string, object[]>(ILogType, "DebugFormat");
-
-			InfoDelegate = DelegateHelper.BuildAction<object>(ILogType, "Info");
-			InfoExceptionDelegate = DelegateHelper.BuildAction<object, Exception>(ILogType, "Info");
-			InfoFormatDelegate = DelegateHelper.BuildAction<string, object[]>(ILogType, "InfoFormat");
-
-			WarnDelegate = DelegateHelper.BuildAction<object>(ILogType, "Warn");
-			WarnExceptionDelegate = DelegateHelper.BuildAction<object, Exception>(ILogType, "Warn");
-			WarnFormatDelegate = DelegateHelper.BuildAction<string, object[]>(ILogType, "WarnFormat");
+			var iLogType = System.Type.GetType("log4net.ILog, log4net", true);
```
fredericDelaporte: Same here. Moreover, there was no need (anymore?) for having the type stored as a field.

src/NHibernate/Logging.log4net.cs
```diff
@@ -12,14 +13,56 @@ namespace NHibernate
...
+		{
+			if (LogManagerType == null)
+				throw new TypeLoadException("Could not load LogManager type", LogManagerTypeLoadException);
+		}
+
+		// Code mostly adapted from ReflectHelper.TypeFromAssembly, which cannot be called directly due
+		// to depending on the logger.
+		private static System.Type GetLogManagerType()
+		{
+			var typeName = new AssemblyQualifiedTypeName("log4net.LogManager", "log4net");
```
hazzik: `typeName` here is completely useless.

src/NHibernate/Logging.log4net.cs
```diff
@@ -12,14 +13,56 @@ namespace NHibernate
...
+		internal static readonly Exception LogManagerTypeLoadException;
+		private static readonly System.Type LogManagerType;
 		private static readonly Func<Assembly, string, object> GetLoggerByNameDelegate;
 		private static readonly Func<System.Type, object> GetLoggerByTypeDelegate;
 
 		static Log4NetLoggerFactory()
 		{
-			GetLoggerByNameDelegate = GetGetLoggerByNameMethodCall();
-			GetLoggerByTypeDelegate = GetGetLoggerMethodCall<System.Type>();
+			try
```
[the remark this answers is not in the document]
fredericDelaporte: It may help diagnosing log4net loading failure. By default they will be silent, but someone having troubles getting log4net logging could try to explicitly enable it with the code `LoggerProvider.SetLoggersFactory(new Log4NetLoggerFactory());` by example, and then the exception will be raised.

hazzik:
Wouldn't it be enough to just `Assembly.Load("log4net")`? I don't have a windows machine handy to check the scenario with log4net being in the GAC (my laptop has died of old age).

fredericDelaporte:
The method is not even [documented](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.load?#System_Reflection_Assembly_Load_System_String_) for usage with short names. With the GAC tendency to reject short names unless they are already loaded in the application domain, I would not rely on this.

Since the code from `ReflectionHelper` is the one used for loading reflection based drivers, I think it is quite proven it works well in many situation, and so I would rather avoid deviating from it /simplifying it.

hazzik:
> With the GAC tendency to reject short names unless they are already loaded in the application domain

But this is anyway what happens here. Isn't it?

fredericDelaporte:
It happens with the `Type.GetType` call, but then the `ReflectHelper` logic fallback on searching among already loaded assemblies any loosely matching one for loading the type from it, and I guess that is the part doing the trick for the case reported here. The last fallback, attempting to explicitly load the assembly with a short name, is likely to not be useful in the case reported here. But maybe it still helps in some other cases.
Once again here, I am a bit conservative, preferring to stick to code which works since years without known complains, rather than trying to simplify it without being sure it won't harm any case.

hazzik:
>code which works since years without known complains

The "logic fallback on searching among already loaded assemblies any loosely matching one for loading the type from it" has been added a few month back in 5.2

fredericDelaporte:
By yourself, granted, but the `Assembly.Load` logic was already there, and that is the one we are currently discussing, aren't we? There is no comment to tell in which case it may help, but does it harms to keep it? Less code is always better, provided we are sure removing it will not break some cases. There, I am not sure.

wensaint:
@fredericDelaporte could you send a fixed release version, i failed to build the dll.
thanks,
my email is wensaint@qq.com

fredericDelaporte:
Expands the checks of this PR, find the "NHibernate (Release Package)" one and click on "Details". Teamcity may ask you for a login but an anonymous option is normally available. Then browse to the artifacts tab of the build and take what you need.

But the current release build was not having artifacts, I do not know why. I have relaunched it, wait for it to complete if it is not finished yet.

fredericDelaporte:
There was some discussion in this PR about whether the loading code could be simplified or not. I was, and still am, for keeping this code as near as possible to the loading logic of `ReflectHelper` (which we cannot use directly due to otherwise creating a circular reference trouble), even if some of it is maybe not useful in all circumstances.
There are just too many possible cases (in GAC or not in GAC, netfx or netcoreapp or mono, directly referenced by the application or just sitting somewhere in a loadable path or in the GAC, strong signed application or not, running in special hosts like Sharepoint, ...) for being sure to have checked them all.

So,

hazzik:
As I still disagree with this change - does it count as "continuous disagreement"?

fredericDelaporte:
Of course, but I would like to better understand what cause you to still disagree. I believe to have explained why this change was the way it was, where you have expressed concerns. May you tell what is wrong with those explanations?

src/NHibernate/Logging.log4net.cs
```diff
@@ -12,16 +13,48 @@ namespace NHibernate
...
+		{
+			if (LogManagerType == null)
+				throw new TypeLoadException("Cannot find log4net.LogManager type");
+		}
+
+		// Code adapted from ReflectHelper.TypeFromAssembly, which cannot be called directly due
+		// to depending on the logger.
+		private static System.Type GetLogManagerType()
+		{
+			var typeName = new AssemblyQualifiedTypeName("log4net.LogManager", "log4net");
```
hazzik:
```suggestion
			var typeName = "log4net.LogManager, log4net";
```

src/NHibernate/Logging.log4net.cs
```diff
@@ -12,16 +13,48 @@ namespace NHibernate
...
+			// Try to get the type from an already loaded assembly
+			var type = System.Type.GetType(typeName.ToString());
+			if (type != null)
+				return type;
+
+			// Load type from an already loaded assembly, or yield null
+			return System.Type.GetType(
+				typeName.ToString(),
```
hazzik: Together with the previous suggestion
```suggestion
				typeName,
```

src/NHibernate/Logging.log4net.cs
```diff
@@ -12,16 +13,48 @@ namespace NHibernate
...
+				throw new TypeLoadException("Cannot find log4net.LogManager type");
+		}
+
+		// Code adapted from ReflectHelper.TypeFromAssembly, which cannot be called directly due
+		// to depending on the logger.
+		private static System.Type GetLogManagerType()
+		{
+			var typeName = new AssemblyQualifiedTypeName("log4net.LogManager", "log4net");
+			// Try to get the type from an already loaded assembly
+			var type = System.Type.GetType(typeName.ToString());
```
hazzik: Together with the previous suggestion
```suggestion
			var type = System.Type.GetType(typeName);
```

PR #1840, opened 2018-09-05
Version steps System.ServiceModel.Primitives avoiding known vulnerability

rheone:
See issue #1839

Modifies NHibernate to use  System.ServiceModel.Primitives >=4.4.3

NHibernate, for both .NETCoreApp 2.0 and .NETStandard 2.0, presently have a dependency on System.ServiceModel.Primitives >= 4.4.0

System.ServiceModel.Primitives 4.4.0 - 4.4.2 have known vulnerabilities. 

The Microsoft advisories are available at:
 - [CVE-2018-0764](https://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2018-0764)
 - [CVE-2018-0786](https://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2018-0786)

[Microsoft .NET Flaws Let Remote Users Deny Service and Bypass Certificate Security Restrictions
](http://www.securitytracker.com/id/1040152)

Issue #1839, opened 2018-09-04
NHibernate 5.1.3 relies on version of System.ServiceModel.Primitives with a known vulnerabilities


rheone:
NHibernate 5.1.3 for both .NETCoreApp 2.0 and .NETStandard 2.0  has a dependency on System.ServiceModel.Primitives >= 4.4.0

System.ServiceModel.Primitives 4.4.0 has known vulnerabilities. The dependency on System.ServiceModel.Primitives  should have its version bumped to a non vulnerable release

The Microsoft advisories are available at:
 - [CVE-2018-0764](https://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2018-0764)
 - [CVE-2018-0786](https://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2018-0786)

[Microsoft .NET Flaws Let Remote Users Deny Service and Bypass Certificate Security Restrictions
](http://www.securitytracker.com/id/1040152)

fredericDelaporte:
NHibernate does not open itself any WCF communication channel and as such is not concerned by these vulnerabilities.

When NHibernate is used in WCF application, this is the responsibility of the application to use a non-vulnerable WCF dependency, which NHibernate will not prevent.

Anyway, the bit of code taking this dependency, the `WcfOperationContext`, may be moved out of NHibernate with #1820.

rheone:
I understand what your are saying, and I get that NHibernate doesn't deal with WFC Communication (and neither does my application that relies on NHibernate), but is it really the stance of NHibernate to introduce code with known vulnerabilities into peoples projects? If so, is it documented anywhere of the known vulnerable libraries that are introduced to a project by including NHibernate?

oskarb:
rheone,
Perhaps there is something I'm not aware of, but I fail to follow your reasoning. NHibernate merely states that it needs at least version 4.4.0 to be able to perform its function. If you add NHibernate to a project, surely NuGet would pull the latest available version of System.ServiceModel.Primitives? If you already have 4.4.0, then it's a matter of keeping up to date on NuGet.

A vulnerable version of the package you mention is not delivered as part of NHibernate but directly from NuGet, so it's not really NHibernate's responsibilty to enforce or ensure that this package is updated, is it? If it is, should NHibernate also refuse to run unless you have installed the latest monthly security fix from Windows Update? What I'm saying is, why is this particular package the responsibility of NHibernate and not the other dependencies?

[...]

PR #1842, opened 2018-09-07
Remove WcfOperationSessionContext from .Net Core and .Net Standard

fredericDelaporte:
Currently, WCF Server support is a .Net Framework only feature.
On .Net Core, it does not seem it will be implemented any time soon, see dotnet/wcf#1200.

The `WcfOperationSessionContext` is meant to be used as a session context during server side calls, so providing it for .Net Core has currently no usages.

Moreover, providing it for .Net Core or .Net Standard requires additional dependencies under .Net Core and .Net Standard, which creates some concerns (see #1820 & #1839).

The .Net Standard use case for it seems very tiny, it could be currently useful only if a .Net Framework WCF server application was using the .Net Standard distribution of NHibernate instead of the .Net Framework one. So better not provide it for .Net Standard too.

Replaces #1820 and #1840 (unless, for the later, if this PR cannot be merged in 5.2).

### DIFF
--- a/src/NHibernate/Context/WcfOperationSessionContext.cs
+++ b/src/NHibernate/Context/WcfOperationSessionContext.cs
@@ -2,7 +2,7 @@
 // a WCF OperationContext for it. Since it adds additional heavy dependencies, it has been
 // considered not desirable to provide it for .Net Standard. (It could be useful in case some
 // WCF server becames available in another frameworks or if a .Net Framework application
-// consumes the .Net standard distribution of NHibernate instead of the .Net Framework one)
+// consumes the .Net Standard distribution of NHibernate instead of the .Net Framework one.)
 // See https://github.com/dotnet/wcf/issues/1200 and #1842
 #if NETFX
 using System.Collections;
@@ -65,17 +65,17 @@ using NHibernate.Engine;
 namespace NHibernate.Context
 {
 	/// <summary>
-	/// Obsolete class not usable in the .Net Core and .Net Standard distributions of NHibernate. Use the
+	/// Obsolete class not usable with the current framework. Use the
 	/// .Net Framework distribution of NHibernate if you need it. See
 	/// https://github.com/nhibernate/nhibernate-core/issues/1842
 	/// </summary>
-	[Obsolete("Not supported in the .Net Core and .Net Standard distributions of NHibernate", true)]
+	[Obsolete("Not supported in this platform", true)]
 	public class WcfOperationSessionContext : MapBasedSessionContext
 	{
 		public WcfOperationSessionContext(ISessionFactoryImplementor factory) : base(factory)
 		{
-			throw new NotSupportedException(
-				"WcfOperationSessionContext is currently supported only by the .Net Framework distribution of NHibernate");
+			throw new PlatformNotSupportedException(
+				"WcfOperationSessionContext is not supported for the current framework");
 		}
 
 		protected override IDictionary GetMap()
--- a/src/NHibernate/Context/WcfOperationSessionContext.cs
+++ b/src/NHibernate/Context/WcfOperationSessionContext.cs
@@ -69,7 +69,7 @@ namespace NHibernate.Context
 	/// .Net Framework distribution of NHibernate if you need it. See
 	/// https://github.com/nhibernate/nhibernate-core/issues/1842
 	/// </summary>
-	[Obsolete("Not supported in the .Net Core and .Net Standard distributions of NHibernate")]
+	[Obsolete("Not supported in the .Net Core and .Net Standard distributions of NHibernate", true)]
 	public class WcfOperationSessionContext : MapBasedSessionContext
 	{
 		public WcfOperationSessionContext(ISessionFactoryImplementor factory) : base(factory)
--- a/src/NHibernate/Context/WcfOperationSessionContext.cs
+++ b/src/NHibernate/Context/WcfOperationSessionContext.cs
@@ -1,4 +1,10 @@
-using System;
+// There is no support of WCF Server under .Net Core, so it makes little sense to provide
+// a WCF OperationContext for it. Since it adds additional heavy dependencies, it has been
+// considered not desirable to provide it for .Net Standard. (It could be useful in case some
+// WCF server becames available in another frameworks or if a .Net Framework application
+// consumes the .Net standard distribution of NHibernate instead of the .Net Framework one)
+// See https://github.com/dotnet/wcf/issues/1200 and #1842
+#if NETFX
 using System.Collections;
 using System.ServiceModel;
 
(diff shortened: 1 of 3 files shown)


src/NHibernate/Context/WcfOperationSessionContext.cs
```diff
@@ -50,3 +56,36 @@ public class WcfStateExtension : IExtension<OperationContext>
...
+		protected override IDictionary GetMap()
+		{
+			return null;
+		}
+
+		protected override void SetMap(IDictionary value)
+		{
+		}
+	}
+}
```
fredericDelaporte: The `#else` part defining a dummy `WcfOperationSessionContext` is unfortunately not enough to avoid a binary breaking change for .Net Core and .Net Standard: there is also the `WcfStateExtension` class which was unfortunately public. But it cannot be put back in that `#else`, because it implements the `IExtension<OperationContext>` interface from `System.ServiceModel`.

May this change still be acceptable for a 5.2? Or is it anyway a no-go, even if no one should have referenced that class, especially in a .Net Core application?

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1276,7 +1276,18 @@ private ICurrentSessionContext BuildCurrentSessionContext()
...
 					return new WcfOperationSessionContext(this);
+#else
+					// There is no support of WCF Server under .Net Core, so it makes little sense to provide
+					// a WCF OperationContext for it. Since it adds additional heavy dependencies, it has been
+					// considered not desirable to provide it for .Net Standard. (It could be useful in case some
+					// WCF server becames available in another frameworks or if a .Net Framework application
+					// consumes the .Net standard distribution of NHibernate instead of the .Net Framework one)
+					// See https://github.com/dotnet/wcf/issues/1200 and #1842
+					throw new NotSupportedException(
+						"WcfOperationSessionContext is currently supported only by the .Net Framework distribution of NHibernate");
```
fredericDelaporte: This `#if #else` added here is currently not strictly needed, since the `WcfOperationSessionContext` constructor under .Net Core or .Net Standard will do the same throw. But since the class is obsoleted under those targets, it avoids referencing an obsolete class.

src/NHibernate/Context/WcfOperationSessionContext.cs
```diff
@@ -50,3 +56,36 @@ public class WcfStateExtension : IExtension<OperationContext>
 		public void Detach(OperationContext owner) { }
 	}
```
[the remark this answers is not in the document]
fredericDelaporte: We cannot, since it implements an interface from `System.ServiceModel`. Not implementing it would be no less binary breaking, since removing an interface is binary breaking. Moreover the interface's members use other `System.ServiceModel` types and could not be left in the class either.

src/NHibernate/Context/WcfOperationSessionContext.cs
```diff
@@ -50,3 +56,36 @@ public class WcfStateExtension : IExtension<OperationContext>
...
+	/// .Net Framework distribution of NHibernate if you need it. See
+	/// https://github.com/nhibernate/nhibernate-core/issues/1842
+	/// </summary>
+	[Obsolete("Not supported in the .Net Core and .Net Standard distributions of NHibernate", true)]
+	public class WcfOperationSessionContext : MapBasedSessionContext
+	{
+		public WcfOperationSessionContext(ISessionFactoryImplementor factory) : base(factory)
+		{
+			throw new NotSupportedException(
+				"WcfOperationSessionContext is currently supported only by the .Net Framework distribution of NHibernate");
```
hazzik: Need to change wording to emphasise that this is not supported in the current framework. The fact that it is only supported in full .NET Framework is not relevant.

src/NHibernate/Impl/SessionFactoryImpl.cs
hazzik: Same as in the constructor.

src/NHibernate/Context/WcfOperationSessionContext.cs
hazzik: It should not, but I'm ok with that.

src/NHibernate/Context/WcfOperationSessionContext.cs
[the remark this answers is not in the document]
hazzik: In theory we can do this, but I would not want to.

src/NHibernate/Context/WcfOperationSessionContext.cs
```diff
@@ -50,3 +56,36 @@ public class WcfStateExtension : IExtension<OperationContext>
...
+	/// Obsolete class not usable in the .Net Core and .Net Standard distributions of NHibernate. Use the
+	/// .Net Framework distribution of NHibernate if you need it. See
+	/// https://github.com/nhibernate/nhibernate-core/issues/1842
+	/// </summary>
+	[Obsolete("Not supported in the .Net Core and .Net Standard distributions of NHibernate", true)]
+	public class WcfOperationSessionContext : MapBasedSessionContext
+	{
+		public WcfOperationSessionContext(ISessionFactoryImplementor factory) : base(factory)
+		{
+			throw new NotSupportedException(
```
hazzik: Should be `PlatformNotSupportedException`

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1276,7 +1276,18 @@ private ICurrentSessionContext BuildCurrentSessionContext()
...
+#if NETFX
 					return new WcfOperationSessionContext(this);
+#else
+					// There is no support of WCF Server under .Net Core, so it makes little sense to provide
+					// a WCF OperationContext for it. Since it adds additional heavy dependencies, it has been
+					// considered not desirable to provide it for .Net Standard. (It could be useful in case some
+					// WCF server becames available in another frameworks or if a .Net Framework application
+					// consumes the .Net standard distribution of NHibernate instead of the .Net Framework one)
+					// See https://github.com/dotnet/wcf/issues/1200 and #1842
+					throw new NotSupportedException(
```
hazzik: Should be `PlatformNotSupportedException`

src/NHibernate/Context/WcfOperationSessionContext.cs
```diff
@@ -50,3 +56,36 @@ public class WcfStateExtension : IExtension<OperationContext>
...
+using NHibernate.Engine;
+
+namespace NHibernate.Context
+{
+	/// <summary>
+	/// Obsolete class not usable in the .Net Core and .Net Standard distributions of NHibernate. Use the
+	/// .Net Framework distribution of NHibernate if you need it. See
+	/// https://github.com/nhibernate/nhibernate-core/issues/1842
+	/// </summary>
+	[Obsolete("Not supported in the .Net Core and .Net Standard distributions of NHibernate", true)]
```
hazzik: Make it generic "Not supported in this platform"

src/NHibernate/Context/WcfOperationSessionContext.cs
fredericDelaporte: The thing is, current framework can be .Net Framework, while using the .Net Standard binaries. In such case stating it is not supported by the current framework while the lack of support actually comes from the binaries would be confusing.
Granted, that is a corner case. Normally by using NuGet no-one should ends up using the .Net Standard binaries under .Net Framework, and currently NHibernate .Net Standard binaries are not provided anywhere else than on NuGet.

PR #1844, opened 2018-09-10
Adjusting string parameter sizes for MSSQL

gliljas:
Fix for #1300. Replaces #480 

Additional logic could be applied in the length calculation, to prevent query plan cache misses.


src/NHibernate.Test/NHSpecificTest/NH3403/Fixture.cs
```diff
@@ -0,0 +1,183 @@
...
+			using (ISession session = OpenSession())
+			using (ITransaction transaction = session.BeginTransaction())
+			{
+				var e1 = new Entity { Name = "Al", AnsiName = "Al" };
+				session.Save(e1);
+				transaction.Commit();
+				Assert.AreEqual(SqlDbType.NVarChar, Driver.LastCommandParameters.First().SqlDbType);
+				Assert.AreEqual(3, Driver.LastCommandParameters.First().Size);
+				Assert.AreEqual(SqlDbType.VarChar, Driver.LastCommandParameters.Last().SqlDbType);
+				Assert.AreEqual(3, Driver.LastCommandParameters.Last().Size);
```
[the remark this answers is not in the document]
gliljas: True. I reused my old tests from the previous PR. Will fix.

src/NHibernate/Driver/SqlClientDriver.cs
```diff
@@ -17,9 +18,9 @@ namespace NHibernate.Driver
 	/// </summary>
 	public class SqlClientDriver
 #if NETFX
-		: DriverBase, IEmbeddedBatcherFactoryProvider
+		: DriverBase, IEmbeddedBatcherFactoryProvider, IParameterAdjuster
 #else
-		: ReflectionBasedDriver, IEmbeddedBatcherFactoryProvider
+		: ReflectionBasedDriver, IEmbeddedBatcherFactoryProvider, IParameterAdjuster
```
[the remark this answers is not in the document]
gliljas: Good idea.

gliljas:
I'll add some more tests (e.g using MappedAs, edge lengths etc.)

src/NHibernate/Driver/SqlClientDriver.cs
```diff
@@ -18,18 +18,19 @@ namespace NHibernate.Driver
...
 	{
 		// Since v5.1
 		[Obsolete("Use MsSql2000Dialect.MaxSizeForAnsiClob")]
 		public const int MaxSizeForAnsiClob = 2147483647; // int.MaxValue
-		// Since v5.1
+														  // Since v5.1
 		[Obsolete("Use MsSql2000Dialect.MaxSizeForClob")]
 		public const int MaxSizeForClob = 1073741823; // int.MaxValue / 2
-		// Since v5.1
+													  // Since v5.1
```
fredericDelaporte: Strange spacing glitch here

gliljas: Huh? Strange indeed!

gliljas:
Hmmm...what do you say @fredericDelaporte? What behavior should we aim for when a value exceeds the max size. Going above e.g. MaxSizeForLengthLimitedString is mostly not a problem (but it may be for SQL Server 2000)

fredericDelaporte:
About SQL Server 2000, well, its support has completely ended in 2013. So, I am for not caring about an edge case for this product. (Limited string parameters which value size happens to get above the the max limited string size.)

So, when a value exceeds the max mapped size, I think setting the parameter size to the value size is a good choice.
Of course, it will still cause plan cache misses if the application use many value lengths above mapped size. But going to `MaxSizeForLengthLimitedString` instead of the value size looks overkill to me. I think it is an edge case which does not deserve a fix. And moreover at least for Linq, the developer can handle it by specifying its own length with `MappedAs`.

In fact I also question the validity of setting the size to `MaxSizeForClob` for a size above `MaxSizeForLengthLimitedString`: normally, the parameter size for `nvarchar(max)`/`varchar(max)`/`ntext`/`text` should be set to `-1`. Unfortunately, in the documentation, I can only find [an example of this](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/modifying-large-value-max-data#example-2), not an explicit statement.
(Moreover, see [here](https://github.com/nhibernate/nhibernate-core/pull/710#issuecomment-335149234), Sql Server CE tends to allocate the memory for the specified parameter size. Fortunately it does no more use the Sql Server driver logic currently.)
This blob size subject is unrelated to the PR here, so it should likely not be handled here.

Side note: the async texts needs a regen. This is why they fail currently.

gliljas:
I think I'm done with this one. Can rebase-squash if so desired.

fredericDelaporte:
Other dialect drivers do not set parameter sizes, so the tests would fail. Unless the size assert are changed for being done only when the driver is the sql server one.

hazzik:
> Other dialect drivers do not set parameter sizes,

At least ODBC driver does so.

fredericDelaporte:
No, it does no more since [NH-4083](https://nhibernate.jira.com/browse/NH-4083) (v5.0, #703, 3be6668).

src/NHibernate/AdoNet/IParameterAdjuster.cs
```diff
@@ -0,0 +1,11 @@
+using System.Data.Common;
+using NHibernate.SqlTypes;
+using NHibernate.Type;
+
+namespace NHibernate.AdoNet
+{
+	internal interface IParameterAdjuster
```
fredericDelaporte: Why is it `internal`? An external driver implementor may need it.

gliljas: It was a prototype, so I kept it from "the public". I wouldn't mind making it public.

fredericDelaporte:
Since it was conflicting with the latest merge on master, I have rebased it. And I have added a commit for switching the interface to public.

gliljas:
Thanks!

PR #1846, opened 2018-09-12
Remove dependency on System.Security.Permissions package for .NET Standard and .NET Core

hazzik:
System.Security.Permissions package is used to check that `ReflectionPermission` is granted to perform a small optimization for `DynamicMethod`, however in .NET Standard and .NET Core permissions are not really supported and `SecurityManager.IsGranted` always returns `false`.

Also for .NET Framework the obsoleted `SecurityManager.IsGranted` was replaced with a preferred method of checking permissions.

### DIFF
--- a/src/NHibernate/Bytecode/Lightweight/ReflectionOptimizer.cs
+++ b/src/NHibernate/Bytecode/Lightweight/ReflectionOptimizer.cs
@@ -101,11 +101,20 @@ namespace NHibernate.Bytecode.Lightweight
 
 		protected DynamicMethod CreateDynamicMethod(System.Type returnType, System.Type[] argumentTypes)
 		{
-			System.Type owner = mappedType.IsInterface ? typeof (object) : mappedType;
-#pragma warning disable 612,618
-			bool canSkipChecks = SecurityManager.IsGranted(new ReflectionPermission(ReflectionPermissionFlag.MemberAccess));
-#pragma warning restore 612,618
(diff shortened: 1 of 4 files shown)


src/NHibernate/Driver/OdbcDriver.cs
```diff
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
-using System.Data.Odbc;
```
fredericDelaporte: Unrelated, but why not.

src/NHibernate/Bytecode/Lightweight/ReflectionOptimizer.cs
```diff
@@ -101,13 +101,22 @@ protected virtual void ThrowExceptionForNoDefaultCtor(System.Type type)
...
 		}
 
+		private static bool CanSkipVisibilityChecks()
+		{
+#if NETFX
+			var permissionSet = new PermissionSet(PermissionState.None);
+			permissionSet.AddPermission(new ReflectionPermission(ReflectionPermissionFlag.MemberAccess));
+			return permissionSet.IsSubsetOf(AppDomain.CurrentDomain.PermissionSet);
+#else
+			return false;
```
fredericDelaporte: As far as I understand it, under .Net Core, we should always have the permission. But with .Net Standard, hard to know.

PR #1848, opened 2018-09-14
Add in ByCode support of all type mappings on Id

fredericDelaporte:
Mapping by code does not currently allow to set a parameterized type on ids.
This is an undue limitation, which may cause issues to users of custom types, and may also hinder usage of #1844 with mapping by code.

Having checked the xsd, this was the only case lacking full type specification possibilities in by-code. The `type` element is used on `element`, `property`, `key-property` and `id`. By code was already allowing to define parameters on the three former.

### DIFF
--- a/src/NHibernate/Mapping/ByCode/IIdMapper.cs
+++ b/src/NHibernate/Mapping/ByCode/IIdMapper.cs
@@ -1,19 +1,44 @@
 using System;
+using NHibernate.Mapping.ByCode.Impl;
 using NHibernate.Type;
+using NHibernate.Util;
 
 namespace NHibernate.Mapping.ByCode
 {
+	// 6.0 TODO: move into IIdMapper,
+	// and probably add an ITypeMapper for mutualizing the four Type methods with IElementMapper and IPropertyMapper
+	// (Note that there is no IKeyPropertyMapper to be concerned with, the KeyPropertyMapper use IPropertyMapper
+	// directly instead.)
+	public static class IdMapperExtensions
+	{
+		public static void Type<TPersistentType>(this IIdMapper idMapper) where TPersistentType : IIdentifierType
+		{
+			ReflectHelper
+				.CastOrThrow<IdMapper>(idMapper, "generic Type method")
+				.Type<TPersistentType>();
+		}
+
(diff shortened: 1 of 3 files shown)


src/NHibernate/Mapping/ByCode/IIdMapper.cs
```diff
@@ -1,20 +1,45 @@
...
 	{
 		void Generator(IGeneratorDef generator);
 		void Generator(IGeneratorDef generator, Action<IGeneratorMapper> generatorMapping);
 
 		void Type(IIdentifierType persistentType);
-		//void Type<TPersistentType>() where TPersistentType : IIdentifierType;
-		//void Type<TPersistentType>(object parameters) where TPersistentType : IIdentifierType;
-		//void Type(System.System.Type persistentType, object parameters);
-		//void Column(Action<IColumnMapper> columnMapper);
-		//void Columns(params Action<IColumnMapper>[] columnMapper);
```
fredericDelaporte: The column methods are not lacking, they are here through `IColumnsMapper`.

src/NHibernate/Mapping/ByCode/IIdMapper.cs
```diff
@@ -1,20 +1,45 @@
...
+		}
+
+		public static void Type<TPersistentType>(this IIdMapper idMapper, object parameters) where TPersistentType : IIdentifierType
+		{
+			ReflectHelper
+				.CastOrThrow<IdMapper>(idMapper, "generic parameterized Type method")
+				.Type<TPersistentType>(parameters);
+		}
+
+		public static void Type(this IIdMapper idMapper, System.Type persistentType, object parameters)
```
bahusoid: Isn't it the only method that needs to be added to interface? I think it's better to   keep other methods as extensions and inline them.

src/NHibernate/Mapping/ByCode/Impl/IdMapper.cs
```diff
@@ -66,6 +67,38 @@ public void Type(IIdentifierType persistentType)
 			}
 		}
 
+		public void Type(System.Type persistentType, object parameters)
```
hazzik: In a(!) future I would like to have a method which accepts `IDictionary<string, object>` for parameters.

PR #1850, opened 2018-09-17
Add HqlTreeBuilder.From overload that accepts HqlJoin

### LINKED ISSUE #1838: Cannot add HqlJoin to HqlFrom
The function `HqlTreeBuilder.From` only accepts `HqlRange`, but not `HqlJoin` in addition to `HqlRange`, but it seems that JOINs must be part of the FROM AST.

hazzik:
Fixes #1838

### DIFF
--- a/src/NHibernate/Hql/Ast/HqlTreeBuilder.cs
+++ b/src/NHibernate/Hql/Ast/HqlTreeBuilder.cs
@@ -74,6 +74,11 @@ namespace NHibernate.Hql.Ast
 			return new HqlSelectFrom(_factory, @from);
 		}
 
+		public HqlFrom From(HqlJoin join)
+		{
+			return new HqlFrom(_factory, join);
+		}
+
 		public HqlFrom From(HqlRange range)
 		{
 			return new HqlFrom(_factory, range);


fubar-coder:
I guess that there was some kind of misunderstanding, but this is the code I used in our project:

```csharp
HqlTreeNode source = null; // The source
var sourceAlias = treeBuilder.Alias("sourceAlias"); // The source alias
var sourceAliasIdent = treeBuilder.Ident("sourceAlias"); // The source alias as identifier
var refProperty = "MyRefProp"; // The referenced property in the source object
var refAlias = treeBuilder.Alias("refAlias"); // The alias for the joined reference property
var from = treeBuilder.From(treeBuilder.Range(source, sourceAlias));
var join = treeBuilder.Join(treeBuilder.Dot(sourceAliasIdent, treeBuilder.Ident(refProperty)), refAlias);
_addChild.Invoke(from, new object[] { join });
```

With `_addChild` being defined as:

```csharp
private static readonly MethodInfo _addChild = typeof(HqlTreeNode).GetMethod(
        "AddChild",
        BindingFlags.Instance | BindingFlags.NonPublic);
```

That means, that the fix should be something like this:

```csharp
public HqlFrom From(HqlRange range, params HqlJoin[] joins)
{
    return new HqlFrom(_factory, joins);
}
```

Not sure if this compiles, but we should be able to add 1+ joins for a "from" token.

hazzik:
Ok...

fubar-coder:
I'm generating this HQL AST from within a `BaseHqlGeneratorForMethod` implementation and I don't seem to have the HQL tree available there.

This is the code I used to look at the AST for JOINs:

```csharp
var sfi = (ISessionFactoryImplementor) session.SessionFactory;
var expr = new HqlParseEngine("from entity sourceAlias join sourceAlias.MyRefProp refAlias", false, sfi);
var node = expr.Parse();
System.Diagnostics.Debug.WriteLine(node.ToStringTree());
```

PR #1851, opened 2018-09-17
Add HqlTreeBuilder.From overload which accepts HqlRange and a collection of HqlJoin

### LINKED ISSUE #1838: Cannot add HqlJoin to HqlFrom

hazzik:
Fixes #1838 

Hopefully this time is correct:)

/cc @fubar-coder


hazzik:
Thats fine. Thanks

PR #1852, opened 2018-09-19
Avoid initializing proxies when ordering inserts

### LINKED ISSUE #1338: NH-3806 - Saving entities with proxy associations leads to fetching associated entities
**Serhiy Shekhovtsov** created an [issue](https://nhibernate.jira.com/browse/NH-3806) — <time datetime="2015-07-06T09:26:51.808Z" title="2015-07-06T09:26:51.808Z">6th July 2015, 9:26:51</time>:
> During batch saving the method IsProcessedAfterAllAssociatedEntities is called and it calls GetHashCode of the associated Entity. At this moment DefaultLazyInitializer intercepts the call and fires ImmediateLoad for the associated proxy entity, which leads to entity fetching.
> More details [here](http://stackoverflow.com/questions/30981735/nhibernate-batch-saving-entities-with-proxy-associations).
> 
---
**Serhiy Shekhovtsov** added a comment — <time datetime="2015-07-07T12:03:44.438Z" title="2015-07-07T12:03:44.438Z">7th July 2015, 12:03:44</time>:
> As I figured out this is not related to batching and can be reproduced with a [simplest example](http://stackoverflow.com/questions/31267984/nhibernate-insert-entity-with-proxy-associations-without-fetching).
> 
---
**Serhiy Shekhovtsov** added a comment — <time datetime="2015-07-08T09:23:15.432Z" title="2015-07-08T09:23:15.432Z">8th July 2015, 9:23:15</time>:
> Complete and very basic sample can be downloaded [here](https://drive.google.com/file/d/0B6H7K08_eEWiUHBjRExMUXVWSFU/view?usp=sharing).
> 

---

Summarizing it:

When using entities which do override the `GetHashCode` and `Equals` implementation, and use an identifier generator which does not trigger an immediate flush, saving many entities associated to uninitialized proxies causes the proxies to be loaded.
This reduces the usefulness of `ISession.Load`.

fredericDelaporte:
Fix #1338

### DIFF
--- a/src/NHibernate/Async/Action/IExecutable.cs
+++ b/src/NHibernate/Async/Action/IExecutable.cs
@@ -8,6 +8,7 @@
 //------------------------------------------------------------------------------
 
 
+using System;
 using NHibernate.Engine;
 
 namespace NHibernate.Action
--- a/src/NHibernate.Test/Async/Insertordering/AnimalModel/Fixture.cs
+++ b/src/NHibernate.Test/Async/Insertordering/AnimalModel/Fixture.cs
@@ -102,5 +102,66 @@ namespace NHibernate.Test.Insertordering.AnimalModel
 				Assert.DoesNotThrowAsync(() => { return tran.CommitAsync(); });
 			}
 		}
+
+		// #1338
+		[Test]
+		public async System.Threading.Tasks.Task InsertShouldNotInitializeManyToOneProxyAsync()
+		{
+			var person = new Person {  Name = "AnimalOwner" };
+			using (var s = OpenSession())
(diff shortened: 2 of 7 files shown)


src/NHibernate/Async/Action/IExecutable.cs
```diff
@@ -8,6 +8,7 @@
 //------------------------------------------------------------------------------
 
 
+using System;
```
fredericDelaporte: The Nhibernate.dll async changes in this PR are all due to a missing async regeneration in #1452. Fortunately, they are just noisy changes without any practical impact. So I think we can let them go here.

PR #1853, opened 2018-09-19
Reduce proxy initialization cases on GetHashCode/Equals calls

fredericDelaporte:
#1338 has lead me to think about a possible optimization for reducing the number of cases triggering a proxy initialization.

It may not be worth it but well, I'd like to have your viewpoint.

Since the addition of auto-properties in .Net, most entity classes should no more use explicitly declared backing fields. In such case, their method implementations have to use the properties for accessing the state.
So in such case, the proxy could let method implementations be un-intercepted. Any property accessed by the method implementation will still yield the correct value or set the expected state, since the property access will still be intercepted by the proxy if needed.

In this PR, I have changed only the `GetHashCode` and `Equals` case. It is frequent that these methods are overridden for working on the id property, as usually advised. The id property can be accessed without initializing the proxy, so these methods would benefit most of not being intercepted.

This would allow adding an entity to a `set` without actually loading the entity. Currently, adding to a set a proxied entity overriding `GetHashCode` and `Equals` forces its initialization. (This may be a reason why frequently bags or lists are used instead of sets, although it is probably more due to a lack of reflection about their respective semantic.)

Maybe there is a wicked case though: polymorphic association loading a proxy of the base class. In such case, the concrete class could have an override using a field, which without this change would still be called. Well, if this case is possible, we already have the trouble with a base class not overriding `GetHashCode` and `Equals` while the concrete class would be overriding them.


src/NHibernate/Proxy/NHibernateProxyBuilder.cs
```diff
@@ -37,6 +39,22 @@ public NHibernateProxyBuilder(MethodInfo getIdentifierMethod, MethodInfo setIden
 
 		public TypeInfo CreateProxyType(System.Type baseType, IReadOnlyCollection<System.Type> baseInterfaces)
 		{
+			if (_overridesEquals)
+			{
```
hazzik: Can this please go to a method?

hazzik:
It seems like this has a big performance impact.

hazzik:
We could extend this logic to regular methods as well.

fredericDelaporte:
> It seems like this has a big performance impact.

I was expecting a performance impact on proxy generation. But not to the point of causing all builds to timeout... Not worth it in such case.

But looking at the logs, it seems the builds have all frozen on the same test: `NHibernate.Test.NHSpecificTest.GH1645.Fixture.SOEOnLoad`. I will have a look.

> We could extend this logic to regular methods as well.

This could be a performance impact on proxy usage: a method using many properties would have each property call intercepted by the proxy logic. While with current implementation, only the method call is intercepted, but after that it runs on a non-proxied instance, without additional interceptions for each property call done by the method.
With static proxy implementation, "proxy interception" means a call to `__lazyInitializer.GetImplementation()` calling itself `Initialize()` which for an initialized proxy calls directly `CheckTargetState()` which does some trivial tests. So that should have a moderate impact, but maybe could it be still noticeable for methods doing a lot of property calls.

Of course this impact applies to `GetHashCode` and `Equals` too. But as written in this PR description, I expect these methods to be the ones benefiting the most of the possibility of running without initializing the proxy.

Otherwise we could also reduce this potential impact by still intercepting method calls, but in a different way when there are no fields: delegate them to base implementation if the proxy is not yet initialized, otherwise call them on the inner non-proxied instance.

hazzik:
>This could be a performance impact on proxy usage: a method using many properties would have each property call intercepted by the proxy logic. While with current implementation, only the method call is intercepted, but after that it runs on a non-proxied instance, without additional interceptions for each property call done by the method.

Properties _are_ methods. We do intercept all properties ATM. And with this approach we could intercept _only_ properties.

hazzik:
Ok, it seems I misread your comment. I now got the point.

fredericDelaporte:
No harm, let just delete them. (Yes I know that under the hood properties are methods, but for the purpose of this point of discussion it was not much relevant.)

hazzik:
@fredericDelaporte have you seen my PR? There I've moved the "has fields" calculation to the very early stage of creating AbstractProxyFactory.

fredericDelaporte:
I forgot, a class can have private properties too. This PR should still intercept `Equals` and `GetHashCode` if there are private properties. That is one more thing to check.

src/NHibernate.Test.VisualBasic/NHibernate.Test.VisualBasic.vbproj
```diff
@@ -29,7 +29,9 @@
     <None Include="..\NHibernate.Test\App.config" Link="App.config" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
     <PackageReference Include="Microsoft.VisualBasic" Version="10.2.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.10.0" />
```
fredericDelaporte: Not having directly these references in the VB project was not allowing Rider to run the unit tests.

src/NHibernate/Proxy/AbstractProxyFactory.cs
```diff
@@ -18,7 +19,7 @@ public abstract class AbstractProxyFactory: IProxyFactory
 		protected virtual MethodInfo GetIdentifierMethod { get; private set; }
 		protected virtual MethodInfo SetIdentifierMethod { get; private set; }
 		protected virtual IAbstractComponentType ComponentIdType { get; private set; }
-		protected virtual bool OverridesEquals { get; set; }
+		protected virtual bool InterceptsEquals { get; set; }
```
hazzik: You cannot just rename the field.

src/NHibernate/Util/TypeExtensions.cs
```diff
@@ -48,5 +53,60 @@ internal static System.Type UnwrapIfNullable(this System.Type type)
...
+					}
+
+					analyzing = analyzing.BaseType;
+				}
+			}
+		}
+
+		internal static IEnumerable<PropertyInfo> GetPropertiesOfHierarchy(this System.Type type)
+		{
+			if (!type.IsInterface)
```
hazzik: a) you do not need to traverse hierarchy here as `GetProperties` is capable of returning all properties of hierarchy.
b) while interface could not have fields they could have properties. So in a method above this is a simple optimization while here it prevents a valid case.

src/NHibernate/Proxy/AbstractProxyFactory.cs
fredericDelaporte: Yes of course, I forgot about this being a breaking change.

src/NHibernate/Util/TypeExtensions.cs
fredericDelaporte: According to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.type.getproperties?view=netframework-4.7.2#System_Type_GetProperties_System_Reflection_BindingFlags_):

> Only protected and internal properties on base classes are returned; private properties on base classes are not returned.

The point of the change is specifically to detect private ones. Of course, private properties are not an usual thing. Usually fields are used instead. This is likely why in most cases, it is not cared for getting private properties.

fredericDelaporte:
At this point, I do not think it is worth it. It will still not help reduce proxy initializations in a number of cases. Notably when `IEquatable` is implemented, it will usually mean having an `object.Equals` override calling a class specific `Equals`. With this PR, it would still initialize the proxy even if not using any other state than the id.

PR #1854, opened 2018-09-20
Add SQL Anywhere 17 support

pettersoft:
Added a SQL Anywhere 17 driver and SQL Anywhere 17 dialect.


src/NHibernate/Dialect/SybaseSQLAnywhere17Dialect.cs
```diff
@@ -0,0 +1,107 @@
...
+	///		<term>connection.driver_class</term>
+	///		<description><see cref="NHibernate.Driver.SybaseSQLAnywhere17Driver" /></description>
+	///	</item>
+	///	<item>
+	///		<term>prepare_sql</term>
+	///		<description><see langword="false" /></description>
+	///	</item>
+	/// </list>
+	/// </remarks>
+	public class SybaseSQLAnywhere17Dialect : SybaseSQLAnywhere12Dialect
```
fredericDelaporte: It seems to me the name Sybase is phased out. Shouldn't it be named `SapSQLAnywhere17Dialect`?

src/NHibernate/Dialect/SybaseSQLAnywhere17Dialect.cs
```diff
@@ -0,0 +1,107 @@
...
+	{
+		public SybaseSQLAnywhere17Dialect()
+			: base()
+		{
+			DefaultProperties[Environment.ConnectionDriver] = "NHibernate.Driver.SybaseSQLAnywhere17Driver";
+			RegisterDateTimeTypeMappings();
+			RegisterKeywords();
+		}
+
+		new protected void RegisterKeywords()
```
fredericDelaporte: `SybaseSQLAnywhere12Dialect` implementation is not what we do usually. Maybe that is the right time to fix it: switch these methods (`RegisterKeywords` and `RegisterDateTimeTypeMappings`) to `override`, calling their base implementation, and remove them from constructor call, both in `SybaseSQLAnywhere12Dialect` and here. `SybaseSQLAnywhere10Dialect` own constructor will call them, once switched to overrides.

src/NHibernate/Driver/SybaseSQLAnywhere17Driver.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace NHibernate.Driver
+{
+	public class SybaseSQLAnywhere17Driver : ReflectionBasedDriver
```
fredericDelaporte: As for the dialect, it should likely be renamed `SapSQLAnywhere17Driver`.

src/NHibernate/Dialect/SybaseSQLAnywhere17Dialect.cs
pettersoft: Yes, that's correct! It should be renamed to Sap instead.

src/NHibernate/Dialect/SapSQLAnywhere17Dialect.cs
```diff
@@ -20,35 +20,30 @@ namespace NHibernate.Dialect
...
-
-		new protected void RegisterKeywords()
+	
+		protected override void RegisterKeywords()
 		{
-			RegisterKeyword("near");
-			RegisterKeyword("limit");
-			RegisterKeyword("offset");
-			RegisterKeyword("datetimeoffset");
+			base.RegisterKeywords();
```
fredericDelaporte: So, indeed these keywords were already registered in the base dialect. Better remove completely the override then.

src/NHibernate/Dialect/SapSQLAnywhere17Dialect.cs
```diff
@@ -20,35 +20,30 @@ namespace NHibernate.Dialect
...
 		}
 
-		new void RegisterDateTimeTypeMappings()
+		protected override void RegisterDateTimeTypeMappings()
 		{
-			RegisterColumnType(DbType.DateTimeOffset, "DATETIMEOFFSET");
+			base.RegisterDateTimeTypeMappings();
```
fredericDelaporte: Same here.

fredericDelaporte:
Numerous tests are failings, and `HqlIsThreadSafe_UsingThreads` seems to freeze. Granted, previous Anywhere dialects were likely not fixed for all tests either. But now we rather fix tests or at least ignore those due to known limitations of the target database. (This was done for latest dialect additions, see #1662 by example.)

I may directly push here some commits for fixing these tests failures, but of course feel free to fix some of them too.

src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
```diff
@@ -29,7 +29,8 @@ protected override bool AppliesTo(Dialect.Dialect dialect)
 			return !(dialect is FirebirdDialect) &&
 			       !(dialect is Oracle8iDialect) &&
 			       !(dialect is MsSqlCeDialect) &&
-			       !(dialect is HanaDialectBase);
+			       !(dialect is HanaDialectBase) &&
+			       !(dialect is SybaseSQLAnywhere10Dialect);
```
fredericDelaporte: This batcher is not supported by SQL Anywhere. It seems it does not supports to have multiple inserts in the same query, as described [here](https://stackoverflow.com/a/4400607/1178314).

src/NHibernate.Test/Component/Basic/ComponentWithUniqueConstraintTests.cs
```diff
@@ -21,8 +21,8 @@ protected override HbmMapping GetMappings()
 
 			mapper.Component<Person>(comp =>
 			{
-				comp.Property(p => p.Name);
-				comp.Property(p => p.Dob);
+				comp.Property(p => p.Name, m => m.NotNullable(true));
+				comp.Property(p => p.Dob, m => m.NotNullable(true));
 				comp.Unique(true); // hbm2ddl: Generate a unique constraint in the database
```
fredericDelaporte: This test does not need those columns to be nullable, so for avoiding troubles with database not supporting null in unique constraint, setting the columns as not nullable.

src/NHibernate/Dialect/SybaseSQLAnywhere10Dialect.cs
```diff
@@ -192,6 +194,11 @@ protected virtual void RegisterAggregationFunctions()
 
 		protected virtual void RegisterBitFunctions()
 		{
+			// SQL Anywhere does not respect usual priorities with the bitwise not. Unfortunately the HQL parser
+			// furthermore remove "undue" parenthesis according to usual rules. As the bitwise not should have maximal
+			// priority, we can work around this by using a template forcing parenthesis around it.
+			RegisterFunction("bnot", new VarArgsSQLFunction(NHibernateUtil.Int64, "(~", "", ")"));
```
fredericDelaporte: This is not a breaking change, because the function argument will by default be all that follows the `~`. But thanks to this change, if parenthesis are used to explicit the priority, they will be taken into account for supplying the "function" argument, instead of being silently removed by the parser.

src/NHibernate/Dialect/SybaseSQLAnywhere10Dialect.cs
```diff
@@ -556,6 +565,11 @@ public override SqlString GetLimitString(SqlString sql, SqlString offset, SqlStr
 
 			if (insertionPoint > 0)
 			{
+				if (limit == null && offset == null)
+					throw new ArgumentException("Cannot limit with neither a limit nor an offset");
+				if (limit == null)
+					throw new NotSupportedException($"Dialect {this} does not support setting an offset without a limit");
```
fredericDelaporte: This case was previously blowing with a null reference exception. Overriden in the 17 version since SQL Anywhere 12.0.1 (but not 12.0.0) has a syntax allowing the support of this case.

src/NHibernate/SqlCommand/SqlStringBuilder.cs
```diff
@@ -158,10 +158,9 @@ public SqlStringBuilder AddObject(object part)
 		/// </summary>
 		/// <param name="sqlString">The SqlString to add to this SqlStringBuilder</param>
 		/// <returns>This SqlStringBuilder</returns>
-		/// <remarks>This calls the overloaded Add(sqlString, null, null, null, false)</remarks>
 		public SqlStringBuilder Add(SqlString sqlString)
 		{
-			sqlString.Visit(AddingVisitor);
+			sqlString?.Visit(AddingVisitor);
```
fredericDelaporte: Additional change for avoiding blowing with a null reference exception when the `sqlString` is null. The usual behavior for string builder being to do nothing on null, doing the same here.

hazzik:
@fredericDelaporte can we setup an automatic build for it?

fredericDelaporte:
On the TeamCity agent, it should be doable. On AppVeyor or Travis, I do no think there is a reasonably lightweight way to set it up on each build.

fredericDelaporte:
> Update: it is maybe not possible for TeamCity too: there is no NuGet package as far as I know for the data provider, and I have not checked yet if its license could allow us to put it in NHibernate lib folder in Git.

There is [a Nuget package](https://www.nuget.org/packages/Sap.Data.SQLAnywhere/), but it is very likely to be unofficial. Looking at the license files available on my local setup, no redistribution is allowed for all components, including client ones. So I do not think we can legally put the data provider in the NHibernate lib folder. (The license allows a single installation for the developer and 25 internal deployments for testing purpose.)

hazzik:
Have you considered installing into GAC? We can put a readme file to /lib/teamcity/sqlAnywhere

fredericDelaporte:
It does, and it is installed in GAC by the SQL Anywhere installer. Nothing to tell in lib/teamcity indeed.

fredericDelaporte:
I appears a hack similar to the one done on Firebird parameters in its driver could solves the bad type inference done by the data-provider on projected columns. This troubles seems to occurs only when the projection uses parameter values. And adding a cast around the parameter to explicit its type in the SQL solves the trouble. (At least in the tests [here](https://gist.github.com/fredericDelaporte/5a7362b9373b5be983208e8898931d1d).)

But I do not think it is worth it. Such a hack implies parsing with reg-ex each generated query, and tweaking it with "word replacements". It is brittle and may breaks some query cases not covered by NHibernate tests.

fredericDelaporte:
I have setup a SAP SQL Anywhere TeamCity build. Its first almost successful build can be seen in 609f033 checks.

22 tests have failed. With c3597a3, only 14 failing tests should remain. Their fixes are in #1855 and #1859. Either I change them for being ignored, or this PR has to wait for these other PR merges.

The Anywhere build is paused, and I am launching it manually. Enabling it would likely cause all other branches builds to fail, unless they get updated with 0545b45 changes.

fredericDelaporte:
Rebased for resolving conflicts with latest merges, and test suit fixes squashed together.

fredericDelaporte:
It does still depend on #1855 and #1859. I should have switched it to WIP.

fredericDelaporte:
Rebased, Anywhere TeamCity build queued.

fredericDelaporte:
Another fix is required, still due to the lack of support of `null` values in an unique constraint, see #1900.

This time, I have also locally applied this fix to the AnyWhere branch, and no tests are failing any-more for SQL Anywhere, with #1900 cherry-picked.

fredericDelaporte:
About enabling the build on TeamCity, we should likely wait for having active PR updated first.

fredericDelaporte:
@hazzik, are you agree for enabling the SAP SQL Anywhere build for all builds?

I think it is time to have all PR targeting master at least up-to-date with 5.2.

It will cause a failure for PR targeting older branches though. I do not see how to avoid that.

PR #1856, opened 2018-09-23
Fix Guid.ToString()

### LINKED ISSUE #1151: NH-3426 - Wrong result when converting Guid to string
**Oskar Berggren** created an [issue](https://nhibernate.jira.com/browse/NH-3426) — <time datetime="2013-03-18T09:06:30.236Z" title="2013-03-18T09:06:30.236Z">18th March 2013, 9:06:30</time>:
> Having a sqlite database where a column contains guids stored as blobs.
> 
> In the select clause of a linq query, ToString() is applied on the Guid property of the model object. NHibernate will forward this to the SQL as
>```sql
>      select
>         cast(Id as char) as col_2_0_ 
> ```
> Unfortunately, the resulting strings won't contain the intended standard string representations of GUID, but instead the raw bytes directly interpreted as characters, which is something completely different.
>

lillo42:
Fixes #1151


src/NHibernate.Test/NHSpecificTest/NH3426/Fixture.cs
```diff
@@ -0,0 +1,68 @@
...
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.NH3426
+{
+	[TestFixture]
+	public class Fixture : TestCaseMappingByCode
+	{
+		protected override bool AppliesTo(Dialect.Dialect dialect)
+		{
+			return dialect is SQLiteDialect;
```
hazzik: Need to apply to all dialects

hazzik:
Current failure on firebird is due provider not storing GUID correctly: http://tracker.firebirdsql.org/browse/DNET-509 (fixed in 6.0+)

hazzik:
I think this has gone too far. @fredericDelaporte any thoughts on this topic? Shall we support such a case or shall we throw `NotSupportedException`? Shall we refine `ToString` support?

hazzik:
SybaseASA9Dialect most likely need to have an implementation similar to MySQL or Oracle.

lillo42:
@hazzik Is there any Sybase I can use to test my query?

hazzik:
I don't know. @fredericDelaporte might have one, but I'm not sure.

fredericDelaporte:
> Is there any Sybase I can use to test my query?

You can download and test SAP SQL Anywhere 17 (formerly Sybase SQL Anywhere, formerly Sybase ASA, formerly SQL Anywhere (yes they have abandoned then retaken that name), formerly Watcom SQL, Watcom having been bought by Sybase) for free from [here](https://learn.sapdigital.com/SQLA-Trial-Registration-Page.html), no time limitation, no registration key required. (SAP has bought Sybase and is phasing out the Sybase name.)
This is just an executable, not a service. Their .Net dataprovider is able of starting it automatically on first connection attempt. That is still a shared database but it is usable as an embedded one.
But you will need the driver from #1854, it is not currently merged.

In fact there are two Sybase database systems.
There is also SAP Adaptive Server Enterprise (ASE), trial available only for 90 days by default. This one is the "historical" Sybase SQL Server database, a time co-developed with Microsoft before they parted ways.

I consider that `SybaseASA9Dialect` is an obsolete thing, superseded by `SybaseSQLAnywhere` dialects (and maybe some time later by the `SapSQLAnywhere` dialect from #1854).
Better work on `SybaseSQLAnywhere` dialects rather than on the ASA one. Likely you can still test the ASA dialect with SAP SQL Anywhere 17, but again you will need the driver from #1854.

For the ASE ones, it is quite less convenient to test since free trials of ASE are time limited.

> I think this has gone too far. @fredericDelaporte any thoughts on this topic? Shall we support such a case or shall we throw `NotSupportedException`? Shall we refine `ToString` support? 

I have seen the test case, and I wonder, what is the use case for converting a GUID to a string representation on the db side rather than on the application side?
In principle, why not doing that, provided that is not a complex thing to support. But apparently it is complex to support, and even to test since the behavior of databases about Guid may vary according to the underlying system endianness.
So unless I am missing something, why not just advise to not convert Guid to string in projections, but instead get them as Guid and convert them by code? It looks far more reliable.

Throwing `NotSupportedException` would allow to state our stance explicitly, but if done whatever the dialect, that could be a serious breaking change for those using it with databases on which it currently does not have issues. Better do it only for dialects having issues.

It seems currently fixed for all tested databases. But for some of them, the fix looks brittle to me, and likely to be sensitive to system endianness. It would not be an issue for databases locked on one operating system family, but that is not the case of Oracle or MySQL by example. So for these databases, maybe better throw a not supported exception. (Is-it possible to "unregister" a function?)

src/NHibernate/Dialect/MsSql2000Dialect.cs
```diff
@@ -352,6 +352,7 @@ protected virtual void RegisterFunctions()
 			// Casting to CHAR (without specified length) truncates to 30 characters. 
 			// A longer version would be safer, but 50 is enough to prevent errors when casting uniqueidentifer to a string representation (NH-2858)
 			RegisterFunction("str", new SQLFunctionTemplate(NHibernateUtil.String, "cast(?1 as nvarchar(50))"));
```
fredericDelaporte: Off-topic side-note: I have missed that hard-coded `50` length limitation in a cast. It should use the default cast length instead. Worth a PR?
(Of course for the `strguid` case, using a hard-coded length make more sense.)

src/NHibernate/Dialect/MySQL55Dialect.cs
```diff
@@ -8,6 +9,8 @@ public class MySQL55Dialect : MySQL5Dialect
 		public MySQL55Dialect()
 		{
 			RegisterColumnType(DbType.Guid, "CHAR(36)");
+
+			RegisterFunction("strguid", new SQLFunctionTemplate(NHibernateUtil.String, "?1"));
```
fredericDelaporte: This overrides the rather complex fix in `MySQL5Dialect`. Currently the CI test with `MySQL5Dialect`. Is this override tested somewhere? Locally?

src/NHibernate/Dialect/MySQL5Dialect.cs
```diff
@@ -22,6 +23,8 @@ public MySQL5Dialect()
 			RegisterCastType(DbType.Double, "DECIMAL(19,5)");
 			RegisterCastType(DbType.Single, "DECIMAL(19,5)");
 			RegisterCastType(DbType.Guid, "BINARY(16)");
+
+			RegisterFunction("strguid", new SQLFunctionTemplate(NHibernateUtil.String, "concat(hex(reverse(substr(?1, 1, 4))), '-', hex(reverse(substring(?1, 5, 2))), '-', hex(reverse(substr(?1, 7, 2))), '-', hex(substr(?1, 9, 2)), '-', hex(substr(?1, 11)))"));
```
fredericDelaporte: So works both under Linux and Windows. But endianness is processor dependent. I guess both run under x86/amd64 systems. Anyone able of testing this with a PowerPC maybe?

src/NHibernate/Dialect/MySQLDialect.cs
```diff
@@ -107,6 +107,8 @@ public MySQLDialect()
 			//special:
 			RegisterColumnType(DbType.Guid, "VARCHAR(40)");
 
+			RegisterFunction("strguid", new SQLFunctionTemplate(NHibernateUtil.String, "?1"));
```
fredericDelaporte: Does this override make sense? Has it-been tested?

src/NHibernate/Dialect/Oracle8iDialect.cs
```diff
@@ -300,6 +300,7 @@ protected virtual void RegisterFunctions()
 			RegisterFunction("next_day", new StandardSQLFunction("next_day", NHibernateUtil.Date));
 
 			RegisterFunction("str", new StandardSQLFunction("to_char", NHibernateUtil.String));
+			RegisterFunction("strguid", new SQLFunctionTemplate(NHibernateUtil.String, "substr(rawtohex(?1), 7, 2) || substr(rawtohex(?1), 5, 2) || substr(rawtohex(?1), 3, 2) || substr(rawtohex(?1), 1, 2) || '-' || substr(rawtohex(?1), 11, 2) || substr(rawtohex(?1), 9, 2) || '-' || substr(rawtohex(?1), 15, 2) || substr(rawtohex(?1), 13, 2) || '-' || substr(rawtohex(?1), 17, 4) || '-' || substr(rawtohex(?1), 21) "));
```
fredericDelaporte: Same here, I wonder if this would work on a big-endian processor, if Oracle is available on any.

src/NHibernate/Dialect/PostgreSQL82Dialect.cs
```diff
@@ -16,6 +17,7 @@ public class PostgreSQL82Dialect : PostgreSQL81Dialect
 		public PostgreSQL82Dialect()
 		{
 			RegisterColumnType(DbType.Guid, "uuid");
+			RegisterFunction("strguid", new SQLFunctionTemplate(NHibernateUtil.String, "cast(?1 as varchar(36))"));
```
fredericDelaporte: Why having this override specific to 8.2 version? Do really previous versions need to use `nvarchar` for an alphanumeric string?
I would expect this function to work for `PostgreSQLDialect`, and if yes, it would be better to use it there.

src/NHibernate/Dialect/MySQL55Dialect.cs
hazzik: No, it isn't. But in an excuse I would say that in this dialect GUIDs are stored as CHAR(36) in a normal guid format.

src/NHibernate/Dialect/PostgreSQL82Dialect.cs
lillo42: Yes, I made this decision where dialect declare GUID.

If I try to use GUID with  `PostgreSQLDialect` or `PostgreSQL81Dialect.` Dialect, I will get a Not Supported GUID exception

src/NHibernate/Dialect/MsSql2000Dialect.cs
lillo42: I think so, after this issue I can create a new PR remove this hard-code.

lillo42:
@fredericDelaporte  and @hazzik The firebase test is faild because database credential is not define. 

> FirebirdSql.Data.FirebirdClient.FbException (0x80004005): Your user name and password are not defined. Ask your database administrator to set up a Firebird login. ---> Your user name and password are not defined. Ask your database administrator to set up a Firebird login.

hazzik:
This happens some times (rarely)

fredericDelaporte:
The Travis and AppVeyor builds works on clean, fresh virtual machine images not having a lot of pre-installed software on it. Databases like Firebird have to be installed as part of the build process. Sometimes these installation steps fail. We just relaunch the build in such case. When not a team member, you cannot do this yourself, unless relaunching all the builds with some new commits.

(On the other hand, the Teamcity build works on one supplied virtual machine with all the software we need permanently installed on it. So it has no installation steps troubles. But sometimes a build do not leave it in a clean state and causes subsequent builds to fail, like I did recently when testing a SAP SQL Anywhere build on it.)

fredericDelaporte:
Sorry for the double post. GitHub encounters difficulties today. One hour after my first post, it was still not showing up on the page, so I have redone it.

lillo42:
@fredericDelaporte 
> As in written in an unanswered comment, I wonder, what is the use case for converting a GUID to a string representation on the db side rather than on the application side?

I don't know very well NHibernate Code, but I'm studying it. Is possible to NHibernate convert GUID on application side?

> In principle, why not doing that, provided that is not a complex thing to support. But apparently it is complex to support.
>
> So unless I am missing something, why not just advise to not convert Guid to string in projections, but instead get them as Guid and convert them by code? It looks far more reliable.
Where we can put this advise?

lillo42:
Should I use connection string of mysql with Old `Guids=True;` ? 
When I try with `Old Guids=True;` I can use this select: `select Id from Entity;`, but when I remove it my insert stop to work.

fredericDelaporte:
> Should I use connection string of mysql with `Old Guids=True;` ?
> When I try with `Old Guids=True;` I can use this select: select Id from Entity;, but when I remove it my insert stop to work.

This thing is a MySql data provider setting.

According to [there](https://stackoverflow.com/a/3117102/1178314):

> After version 6.1.1 you should add "old guids=true" to your connection string whenever you use BINARY(16) as your storage type. Else you should use CHAR(36)

It is also mentioned in the connector [documentation](https://dev.mysql.com/doc/connector-net/en/connector-net-6-10-connection-options.html):

> This option was introduced in Connector/NET 6.1.1. The back-end representation of a GUID type was changed from `BINARY(16)` to `CHAR(36)`. This was done to allow developers to use the server function `UUID()` to populate a `GUID` table - `UUID()` generates a 36-character string. Developers of older applications can add `Old Guids=true` to the connection string to use a GUID of data type `BINARY(16)`. 

So, deciding whether you need this setting or not depends on which type you use for storing your GUIDs.

In version 8, MySQL has gained functions for converting between the two representations: [`uuid-to-bin`](https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_uuid-to-bin) and [`bin-to-uuid`](https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_bin-to-uuid)

fredericDelaporte:
Dues to a formatting issue in the comment, I only see this question now.

> Where we can put this advise?

Well, it looks to me as such a specific need that I would tend to emit it only on issues or PR asking for supporting this.

src/NHibernate/Dialect/MySQL5Dialect.cs
```diff
@@ -1,4 +1,5 @@
 using System.Data;
+using NHibernate.Dialect.Function;
```
fredericDelaporte: There are no more changes in his file. This using is surely now undue.

But it is a bit strange that a Guid stored as a `binary(16)` succeeds the test with the base Dialect version of the function.

src/NHibernate/Linq/Functions/StringGenerator.cs
```diff
@@ -317,13 +317,20 @@ public IHqlGeneratorForMethod GetMethodGenerator(MethodInfo method)
 
 	public class ToStringHqlGeneratorForMethod : IHqlGeneratorForMethod
 	{
+		private static readonly System.Type s_guidType = typeof(Guid);
```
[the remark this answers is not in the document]
lillo42: sorry, next commit I will fix this

src/NHibernate.Test/NHSpecificTest/NH3426/Fixture.cs
```diff
@@ -0,0 +1,63 @@
...
+			}
+		}
+
+		[Test]
+		public void SelectGuidToString()
+		{
+			using (var session = OpenSession())
+			{
+				var list = session.Query<Entity>()
+					.Select(x => new { Id = x.Id.ToString() })
```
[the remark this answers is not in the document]
lillo42: I add this test, and this test works only with SQL Server ( I tried this test on db Postgres(fixes), SQLite and MySQL too).

I think SQLite can have same problem of MySQL and Oracle (big-endian processor).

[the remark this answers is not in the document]
fredericDelaporte: If it does not work within a `Where`, I do not think we can consider the feature as being done. (Unless the failure is unrelated, due by example to case sensitivity or lack of `ToUpper` support. But then a better way to test the where case is needed.)

> I think SQLite can have same problem of MySQL and Oracle (big-endian processor).

Having no way to test this, since it requires a separated database server with an adequate processor architecture, we will likely not investigate further this point.

src/NHibernate/Dialect/PostgreSQL82Dialect.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Data;
+using NHibernate.Dialect.Function;
```
hazzik: No changes now here

PR #1859, opened 2018-09-26
Fix filter & where fragment appended after lock hint

fredericDelaporte:
When a dialect has a read-only lock hint, the filter and/or collection's where fragment is added after the lock hint, which generates an invalid query.

This case happens with SQL Anywhere dialects.


fredericDelaporte:

PR #1864, opened 2018-10-01
Add an async enabled transaction synchronization

### LINKED ISSUE #1752: Async ISynchronization
I am using  [`NHibernate.Transaction.ISynchronization`](https://github.com/nhibernate/nhibernate-core/blob/5.1.3/src/NHibernate/Transaction/ISynchronization.cs) to run code after the transaction has committed. It would be great if there were async version of `AfterCompletion` and `BeforeCompletion` in this interface. I can see it would be easy enough to add directly to the code but all the async methods appear to be automatically generated and I am not sure how that works.

fredericDelaporte:
Fixes #1752


src/NHibernate/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
```diff
@@ -190,17 +190,31 @@ protected virtual void DropTemporaryTableIfNecessary(IQueryable persister, ISess
...
-						{
-							Isolater.DoIsolatedWork(work, session);
-						}
-						else
-						{
-							Isolater.DoNonTransactedWork(work, session);
-						}
-					}));
+					session.ConnectionManager.Transaction.RegisterSynchronization(
+						new AfterTransactionCompletion(
```
hazzik: Can we have a proper class instead of this delegate-based unpleasantness?

Issue #1721, opened 2018-05-31
DateTime + timestamptz + dynamic-update = always update the field even if the value didn't change.


MichaelBuen:
This is the continuation of the [problems](https://github.com/nhibernate/nhibernate-core/issues/1708) I encountered with DateTimeOffset/DateTime + timestamptz/timestamp combinations.

Saving UTC date using DateTime to a timestamptz field, it works with a work-around. However, the work-around causes UPDATE to a timestamptz field even if the value didn't change.

Is there a fix to prevent the UPDATE on unchanged property value?

# Here are the three combinations I have tried:

### TL;DR

Without work-around

|                               | Write | Read 
| ----------------------------- | ------| ---- 
| DateTimeOffset + timestamptz  |   OK  |  👎  
| DateTime + timestamptz        |   👎  |  OK  
| DateTime + timestamp          |   👎  |  👎  

With work-around*

|                               | Write | Read  | Notes                                                                                            
| ----------------------------- | ----- | ----- | -------------------------------------------------------------------------------------------------
| DateTimeOffset + timestamptz  |   OK  |  👎   | no work-around for read
| DateTime + timestamptz        |   OK* |  OK   | with dynamic-update, even if the field's value didn't change, NHibernate still updates the field 
| DateTime + timestamp          |   OK  |  OK*  | timestamp is not as convenient as timestamptz when looking at other countries' time              

### TL;Please Read the work-arounds:

**1. DateTimeOffset + timestamptz = Write fidelity OK. Read not OK, would cause cast exception error.**

**2. DateTime + timestamptz = Read fidelity OK. Write fidelity not OK.**

2018-05-04T17:37:00.000Z input is incorrectly persisted as 2018-05-04 17:37:00+08, which is equivalent to 2018-05-04 09:37:00+00, when written as Zulu format is 2018-05-04T09:37:00.000Z, which is not the same value from input anymore.

Would need this work-around for **Write***:

```c#
public class NpgsqlDriverExtended : NHibernate.Driver.NpgsqlDriver
{                
    protected override void InitializeParameter(DbParameter dbParam, string name, SqlType sqlType)
    {            
        base.InitializeParameter(dbParam, name, sqlType);
                    
        if (sqlType.DbType == System.Data.DbType.DateTime)
        {
            dbParam.DbType = System.Data.DbType.DateTimeOffset;
        }
    }
}
```

[...]

Issue #1708, opened 2018-05-27
UTC date is not sent as-is


MichaelBuen:
schedule field is of timestamptz type.

backing .NET property for schedule field is DateTime type.

User selected May 5, 2018 1:37 AM. My computer's local is Asia/Manila, which is +8.

So UTC date sent is **"2018-05-04T17:37:00.000Z"**

![image](https://user-images.githubusercontent.com/1811636/40582449-bc952a84-61a9-11e8-8bfd-94f7d577efd8.png)

What's shown from NHibernate log, **2018-05-04T17:37:00.0000000Z**

```sql
NHibernate: 
    select
        nextval ('job.job_id_seq')
NHibernate: 
    INSERT 
    INTO
        job.job
        (created_by_fk, created_date, position, description,
         schedule, duration, duration_type_fk, pay, is_pay_net, 
         needed, is_needed_open, 
         location, address_line_1, address_line_2, city_fk, zip_code, 
         location_type_fk, id) 
    VALUES
        (:p0, :p1, :p2, :p3, :p4, :p5, :p6, :p7, :p8, :p9, :p10, :p11, :p12, :p13, :p14, :p15, :p16, :p17);

:p0 = 1 [Type: Int32 (0:0:0)], 
:p1 = 2018-05-27T03:36:57.2189470Z [Type: DateTime (0:0:0)], 
:p2 = 'xxx' [Type: String (0:0:0)], 
:p3 = 'e' [Type: String (0:0:0)], 

:p4 = 2018-05-04T17:37:00.0000000Z [Type: DateTime (0:0:0)], 
:p5 = 1 [Type: Int32 (0:0:0)], 
:p6 = 'D' [Type: String (0:0:0)], 
:p7 = 123 [Type: Decimal (0:0:0)], 
:p8 = True [Type: Boolean (0:0:0)], 

:p9 = 324 [Type: Int32 (0:0:0)], 
:p10 = True [Type: Boolean (0:0:0)], 

:p11 = 'l' [Type: String (0:0:0)], 
:p12 = 'a1' [Type: String (0:0:0)], 
:p13 = 'a2' [Type: String (0:0:0)], 
:p14 = 991 [Type: Int32 (0:0:0)], 
:p15 = '1230' [Type: String (0:0:0)], 

:p16 = 'SC' [Type: String (0:0:0)], 
:p17 = 12 [Type: Int32 (0:0:0)]
```

[...]

PR #1866, opened 2018-10-02
Add UtcTicksType and UtcDbTimestampType

### LINKED ISSUE #1631: Create UtcTicks and UtcDbTimestamp types
Because of the web and internationalization of applications a UtcTicks type would be very useful.

fredericDelaporte:
Fixes #1631.

Like `DbTimestampType`, `UtcDbTimestampType` requires specific dialect support. Otherwise, it falls back to using the client's time.

I have added this support for dialects already supporting `DbTimestampType` and having a reliable way of getting the UTC time.

They are not numerous: many dialects lack support of `DbTimestampType`, probably because it has been overlooked for them. (And it does not cause test failures, since it falls back on client time.)

Cases supporting `DbTimestampType` but not `UtcDbTimestampType`:
- Firebird: no built-in support or workaround found. Some people mention UDF as a way but without examples. And anyway, NHibernate will not inject an UDF on the fly in the database. See [core-694](http://tracker.firebirdsql.org/browse/CORE-694).
- Informix: no-built-in support found, but documentation for this DB lacks discoverability. A quite ugly workaround exists, but with a precision lacking fractional seconds, which is quite unsuitable for a timestamp.
- Oracle8i: no built-in support prior to 9i. I have added it into the 9i dialect.
- PostgreSQL: no built-in support, sort of.
  It handles two datetime types: `timestamp` and `timestamp with time zone` (short `timestamptz`). Both do not store an offset or a timezone but just a datetime. But `timestamptz` is always stored in UTC and yielded back to the client in local time zone. So we can store in UTC with `timestamptz`, but it will not be read back in UTC.
  With `timestamp`, we can hack it by storing `timezone('UTC', current_timestamp)`. But `timestamp` are by default considered as being local (notably when casting to `timestamptz`), so it is not ideal.
  Overall we could get the type working if the underlying column is without time zone (so, `timestamp`, type mapped in the dialect currently for `DateTime`), but not if it is a `timestamptz`. As seen with #1708 and #1721, this can cause troubles to users.
- SQL Anywhere 10: no built-in support prior to version 12. I have added it into the 12 dialect.


src/NHibernate/Dialect/MsSql2000Dialect.cs
```diff
@@ -435,6 +435,16 @@ public override bool SupportsCurrentTimestampSelection
 			get { return true; }
 		}
 
+		/// <inheritdoc />
+		public override string CurrentUtcTimestampSQLFunctionName => "GETUTCDATE()";
```
fredericDelaporte: The Microsoft documentation does not document anything as supported prior to SQL 2008. But it seems this function was already existing in SQL 2000 (and even 7).

src/NHibernate/Type/TicksType.cs
```diff
@@ -9,46 +7,47 @@
...
-	/// This is the recommended way to "timestamp" a column.  
-	/// The System.DateTime.Ticks is accurate to 100-nanosecond intervals. 
+	/// This is the recommended way to "timestamp" a column, along with <see cref="UtcTicksType" />.
+	/// The System.DateTime.Ticks is accurate to 100-nanosecond intervals.
+	/// This type yields dates with an unspecified <see cref="DateTime.Kind"/>. On writes, it
+	/// does not perform any checks or conversions related to the kind of the date value to persist.
 	/// </remarks>
 	[Serializable]
-	public partial class TicksType : PrimitiveType, IVersionType, ILiteralType
+	public partial class TicksType : AbstractDateTimeType
```
fredericDelaporte: This type should have been refactored that way when `AbstractDateTimeType` has been put in place. In order to avoid duplicating the datetime kind handling, I have done it in this PR.

I am right thinking this is not a binary breaking change, provided `AbstractDateTimeType` inherits itself all the removed types and do not have "less virtual" interface implementations?

hazzik: Yes, that's fine

fredericDelaporte:
I was believing to have done a merge commit for merging after my rebase but it looks like a "rebase and merge", sorry for the mistake.

(Or maybe something has changed and GitHub does no more actually perform a merge when the PR is already based on top of master.)

PR #1868, opened 2018-10-06
Db side methods

### LINKED ISSUE #959: NH-4048 - Support non-deterministic/db-side-only methods in Linq
**Frédéric Delaporte** created an [issue](https://nhibernate.jira.com/browse/NH-4048) — <time datetime="2017-07-12T14:53:57.634Z" title="2017-07-12T14:53:57.634Z">12th July 2017, 14:53:57</time>:
> Some methods are non-deterministic and should not be evaluated in the expression tree, once and for all resulting rows, but instead in the database. This is the case for `Guid.NewGuid()` and `Random.Next()`.
> 
> Others may have distinct resulting values depending on the host on which they execute, and can change the query results depending on whether they are evaluated on db-client side or on db-server side. This is the case for `DateTime.Now`, `DateTime.UtcNow`, `DateTimeOffset.Now`, `DateTimeOffset.UtcNow`.
> 
> Currently they are always evaluated by NHibernate Linq provider before emitting the query with the resulting value as a parameter.
> 
> They should be left in the expression tree, and translated to appropriate HQL standard functions, then translated to corresponding database functions.
> 
> A user wishing to have them evaluated client side will then have to put their value in a variable then use that variable in its query.
> 
> By the way, for supporting this feature, when registering a method/property for the linq provider, we should be able to tell if it is partially evaluatable or not.
> 
---
**Frédéric Delaporte** added a comment — <time datetime="2017-07-13T17:38:52.403Z" title="2017-07-13T17:38:52.403Z">13th July 2017, 17:38:52</time>:
> If doing that, an option must be provided for reverting back to previous behavior. This change can be a major breaking changes, causing many working Linq queries to cease working.
> 
> `DateTime.Now.WhateverUnRegisteredMethod()` is currently working, due to it being evaluated and replaced by its returned value in the query. If `DateTime.Now` is no more evaluated but replaced by some sql `getDate()` function, it will then fail on trying to translate the unregistered method call on it.
>

fredericDelaporte:
This adds to the LINQ provider the ability to translate `DateTime.Now` and its variants (Utc, Today, Offset) to SQL calls instead of being pre-evaluated before running the query.
It does the same for `Guid.NewGuid`, `Random.Next` and `Random.NextDouble`.

For enabling these capabilities, matching HQL functions are required. This PR also adds them to all dialects whenever possible.

For being as few breaking as possible, when the dialect lacks support of required functions, the pre-evaluation takes place instead.
This looks quite suitable for `DateTime.Now` and its variants.

But for `Guid` and `Random`, I find it more debatable: when pre-evaluated, the same value will be used for each row of result, whereas when converted to SQL, different values for each row will be used. So this could cause bugs when an application targets a different database than previously tested, without an immediate obvious exception.
Maybe this will be a bad legacy, and maybe it would be better to throw a `NotSupportedException` in those cases, instead of falling back to pre-evaluation.

For queries where this change is breaking, it can be fixed either on a case by case basis or globally.
On a case by case basis, moving the expression newly converted to SQL out of the LINQ query thanks to a local variable will cause it to be evaluated once per request like it was prior to this PR.
Globally, a new boolean setting `linqtohql.legacy_preevaluation` allows to disable this new feature.

Fixes #959.


src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -55,8 +87,14 @@ public override bool IsEvaluatableMethodCall(MethodCallExpression node)
...
 				.GetCustomAttributes(typeof(LinqExtensionMethodAttributeBase), false)
 				.Cast<LinqExtensionMethodAttributeBase>().ToArray();
-			return attributes.Length == 0 ||
-				attributes.Any(a => a.PreEvaluation == LinqExtensionPreEvaluation.AllowPreEvaluation);
+			if (attributes.Length > 0)
+				return attributes.Any(a => a.PreEvaluation == LinqExtensionPreEvaluation.AllowPreEvaluation);
+
+			if (_sessionFactory == null || _sessionFactory.Settings.LinqToHqlLegacyPreEvaluation ||
+				!_sessionFactory.Settings.LinqToHqlGeneratorsRegistry.TryGetGenerator(node.Method, out var generator))
+				return true;
```
fredericDelaporte: This also triggers a lookup of `IRuntimeMethodHqlGenerator` generators: finding them is a O(n) loop already done for each method lacking a regular generator. With this change, it is now done twice per such method.

src/NHibernate/Linq/Visitors/SelectClauseNominator.cs
```diff
@@ -114,11 +114,14 @@ private bool IsRegisteredFunction(Expression expression)
 			if (expression.NodeType == ExpressionType.Call)
 			{
 				var methodCallExpression = (MethodCallExpression) expression;
-				IHqlGeneratorForMethod methodGenerator;
-				if (_functionRegistry.TryGetGenerator(methodCallExpression.Method, out methodGenerator))
+				if (_functionRegistry.TryGetGenerator(methodCallExpression.Method, out var methodGenerator))
```
fredericDelaporte: New lookup of `IRuntimeMethodHqlGenerator`...

Here it is done only for methods inside the select clause. This part is responsible for extracting things deemed not translatable to SQL out of the query, for running them with .Net Runtime on the results.
It considers that methods belonging to instances not part of the query cannot be translated. In general this is right, unless the translation does not care for the instance state, which is the case of the `Random` generator. So here we need to ask any registered generator for the method if it cares about the instance.

src/NHibernate.Test/Linq/MiscellaneousTextFixture.cs
```diff
@@ -27,7 +27,8 @@ public void WhereUsingGuidProperty()
         [Test(Description = "This sample uses Count to find the number of Orders placed before yesterday in the database.")]
         public void CountWithWhereClause()
         {
-            var q = from o in db.Orders where o.OrderDate <= DateTime.Today.AddDays(-1) select o;
+            var yesterday = DateTime.Today.AddDays(-1);
+            var q = from o in db.Orders where o.OrderDate <= yesterday select o;
```
fredericDelaporte: Here is an example of breaking change.
Previously, `DateTime.Today` was pre-evaluated and not translated to SQL, and `AddDays` too. But now, `DateTime.Today` is no more pre-evaluated, forcing `AddDays` to be translated. But there is no generator for it, so the query fails, unless moving all this into a local variable.

fredericDelaporte:
Or add it as an experimental feature, disabled by default till 6.0. (`linqtohql.legacy_preevaluation` would default to `true` at first, and be changed for defaulting to `false` only in 6.0.)

By the way, I have still not really make my mind about the fallback to pre-evaluation when the feature is enabled but the dialect does not support the required translation. Not fall-backing will cause a failure when using the feature, which should be easier and less surprising to diagnose than the consequences of a fallback to pre-evaluation. But that also reduces the portability of an application to other databases, since it is also likely that the fallback could cause less disturbance to the application than a straight failure.

fredericDelaporte:
Rebased, previous fixup squashed, one more fixup added for the last Oracle failure.

fredericDelaporte:
I have added some commits:

The first switch `linqtohql.legacy_preevaluation` to `true` by default, with a not in documentation stating it will likely be `false` in the next major version.

Following commits excepted the last are for an additional option for failing or falling back in case of lack of dialect support: `linqtohql.fallback_on_preevaluation`, `false` by default. Documented as having no effect with legacy pre-evaluation. And in case of failure, the exception message hint at this option.

The last one is just minor adjustments in the documentation of the new pre-evaluation.

src/NHibernate/nhibernate-configuration.xsd
```diff
@@ -145,6 +145,40 @@
...
+										<xs:enumeration value="linqtohql.legacy_preevaluation">
+											<xs:annotation>
+												<xs:documentation>
+													Whether to use the legacy pre-evaluation or not in Linq queries. true by default.
+
+													Legacy pre-evaluation is causing special properties or functions like DateTime.Now or Guid.NewGuid()
+													to be always evaluated with the .Net runtime and replaced in the query by parameter values.
+
+													The new pre-evaluation allows them to be converted to HQL function calls which will be run on the db
+													side. This allows by example to retrieve the server time instead of the client time, or to generate
```
hazzik: "for example"

src/NHibernate/Linq/ReWriters/ConditionalQueryReferenceExpander.cs
```diff
@@ -77,7 +77,7 @@ public void Transform(ResultOperatorBase resultOperator)
 			protected override Expression VisitMember(MemberExpression node)
 			{
 				var result = (MemberExpression) base.VisitMember(node);
-				if (QueryReferenceCounter.CountReferences(result.Expression) > 1)
+				if (result.Expression != null && QueryReferenceCounter.CountReferences(result.Expression) > 1)
```
fredericDelaporte: This additional change, required since #1880 is merged, is not a regression from #1880 since other places in the Linq provider were failing for the same reason in case of non-pre-evaluated static property and are also fixed by the db-side method PR. So it is also fixed here.

fredericDelaporte:
Rebased, more wording fixed, slightly squashed, commits regrouped for resolving conflicts the fixups were having. Stay as fixup the commit for not enabling this feature by default and the commits for failing or fall-backing.

hazzik:
Maybe we just need to add the ability for user register the system methods as not-preevaluateable, but  not register any functions (maybe even provide the generators and information how to register them).

fredericDelaporte:
> Maybe we just need to add the ability for user register the system methods as not-preevaluateable, but not register any functions (maybe even provide the generators and information how to register them).

It seems overly complex to me, when all the user has to do for those functions he does not want to get translated to SQL, is to call them outside of the query.

There is only the breaking change aspect of the feature we need to handle. And this PR does it by supplying configuration option, defaulting to legacy behavior for 5.x.

maca88:
In my opinion the default behaviour should be the following:

***DateTime/DateTimeOffset properties:***
- Should be pre-evaluated when located in the expression and not supported by the database
```C#
from o in db.Orders where o.OrderDate <= DateTime.Today // DateTime.Today should be pre-evaluated
```
- Should not be pre-evaluated when located in the expression and supported by the database
```C#
from o in db.Orders where o.OrderDate <= DateTime.Today // DateTime.Today should be converted to SQL
```
- Should be pre-evaluated when there are additional method calls/members afterwards that are not supported by the database
```C#
from o in db.Orders where o.OrderDate <= DateTime.Today.AddDays(1)
```
- Should not be pre-evaluated when there are additional method calls/members afterwards that are supported by the database
```C#
from o in db.Orders where o.OrderDate <= DateTime.Today.Day
```
- Should be pre-evaluated when located outside of the expression
```C#
var date = DateTime.Today;
from o in db.Orders where o.OrderDate <= date
```

***Guid and Random:***
- Should not be pre-evaluated when they are located in the expression and supported by the database
```C#
from o in db.Orders where o.Guid != new Guid()
```
- Should be pre-evaluated when located outside of the expression and they have no additional method calls/members afterwards
```C#
var guid = new Guid();
from o in db.Orders where o.Guid != guid 
```
- Should throw when they are located in a non select expression or sub-query and not supported by the database 
```C#
from o in db.Orders where o.Guid != new Guid()
```
- Should be evaluated on the client side when they are located in a select expression that is not part of a sub-query and not supported by the database 
```C#
select new Guid() from o in db.Orders
```
- Should throw when they are located outside/inside of a non select expression or sub-query and have additional method calls/members afterwards that are not supported by the database 
```C#
var random = new Random();
from o in db.Orders where o.OrderId != random.NextDouble().GetHashCode()
from o in db.Orders where o.Guid != new Guid().GetHashCode() 
```
- Should be evaluated on the client side when located outside/inside of a select expression that is not part of a sub-query and have additional method calls/members afterwards that are not supported by the database
```C#
var random = new Random();
select random.NextDouble().GetHashCode() from o in db.Orders
select new Guid().GetHashCode()  from o in db.Orders
```

So `LinqToHqlLegacyPreEvaluation` should by default be set to `false` and `LinqToHqlFallbackOnPreEvaluation` to `true`.


Instead of adding `IAllowPreEvaluationHqlGenerator.AllowPreEvaluation`, it would be more useful to have a method like `IEnumerable<string> GetSqlFunctions(MemberInfo member)` for `IHqlGeneratorForMethod` and `IHqlGeneratorForProperty` interfaces, which would return all sql function names that are used in the hql. Such method could be also used in other PRs like #2079 in order to detect whether a client side operation should be done instead. All `AllowPreEvaluation` methods in this PR have similar logic that could be written in one place by having such method.

I will give it a try and see what has to be done in order to have such behavior.

fredericDelaporte:
For a minor release (5.x), if we can avoid possible breaking change, we should.
This PR current state avoids possible breaking changes with its default configuration values.
I think we should set `LinqToHqlLegacyPreEvaluation` to `false` only in a next major version (6.0 if this gets merged in 5.x).

I am not convinced we should handle fallbacks for `DateTime` cases differently than for guid/random.  I would rather give one simple option, fallback for all or fail for all. `DateTime` fallback can be as wrong as guid/random fallback: when the application host is not in the same time zone than the database, the result will be just plain wrong.

> Should be pre-evaluated when there are additional method calls/members afterwards that are not supported by the database

And if that member ends up supported in some later version, end of pre-evaluation? That could be a serious breaking change, if the database is not in the same timezone, again.
I do not think it is a good idea, because it will likely bring more complexity into the logic for determining whether we should pre-evaluate or not, and it will also cause the end result harder to predict for the user.
I think we should keep it simple. And just fail if the `Now` or `Today` is translated but followed by something unsupported: the user will obviously see that when testing, and the fix, if pre-evaluation is fine for him, is easy: just compute it out of the query. (Or switch back to legacy behavior.)

maca88:
> I would rather give one simple option, fallback for all or fail for all. DateTime fallback can be as wrong as guid/random fallback

On second thought, I do agree that we should not provide a fallback for `DateTime/DateTimeOffset` properties, so it should throw no matter where they are located in the expression.

> And if that member ends up supported in some later version, end of pre-evaluation?

For the pre-evaluation I now agree with how it is handled by this PR. Also we need to consider that pre-evaluation is a hot path, as it is executed every time the query is executed, so we should keep it simple.

> I think we should keep it simple. And just fail if the Now or Today is translated but followed by something unsupported

I agree that it should fail when `Now/Today/UtcNow` are used in a non select statement or a select statement of a sub-query, but I think for the following examples:
```C#
db.Orders.Select(o => DateTime.Now.DayOfYear).ToList(); // Does not work
db.Orders.Select(o => DateTime.Now.AddDays(1)).ToList(); // Works
```
we should support client side evaluation when only `DateTime.Now` is supported by the database.

And for Guid and Random I think that we should allow client side evaluation when located in a non subquery select statement:
```C#
var random = new Random();
db.Orders.Select(o => random.NextDouble().GetHashCode()).ToList(); // Works when random is supported, throws otherwise
db.Orders.Select(o => Guid.NewGuid().GetHashCode()).ToList(); // Works when guid is supported, throws otherwise
```
even when `random` and `guid` are not supported by the database. Whether values are generated on the client or on the database doesn't matter, what matters is that the values are random/unique. What are your thoughts on it?
Even if you agree with it, the work should be done/tested in #2079 as client side evalaution is calculated by `SelectClauseHqlNominator` which is heavily modified in the mentioned PR.

fredericDelaporte:
> we should support client side evaluation when only `DateTime.Now` is supported by the database.

Assuming you mean the `Now` is still evaluated on db side, but then its value would be returned in the result set and the additional method/property call evaluated in runtime, it may be a nice to have, provided this does not add much more complexity to the Linq provider. Same for the guid/random case having for this case a fallback to runtime. It is acceptable since it preserves their feature, as in this case they would be run for each row of the result set, not just once for the whole query.

I just write "may" because this does introduce some discrepancy between when those expressions are used in a "final" `Select` and when they are used elsewhere in a Linq expression. As an user, when I was not knowing that NHibernate may evaluate in runtime some projection cases, I had a hard time understanding why some expression seemed supported by the Linq provider in the `Select` but not elsewhere. It was looking to me as a NHibernate bug. The answer was they were not supported at all by the Linq provider, but there were not even "seen" by the provider in those cases, and instead evaluated by the runtime after query execution. When you do not know this, it is a bit confusing.

So overall why not, but I think it would be better to handle it in a dedicated PR. (Maybe not in #2079 either, but in a PR depending on #2079.)

maca88:
> As an user, when I was not knowing that NHibernate may evaluate in runtime some projection cases, I had a hard time understanding why some expression seemed supported by the Linq provider in the Select but not elsewhere.

I had the same issue, but once understood it is very useful to have such feature. I saw many times developers using `ToList` before `Select` method because they thought that it wouldn't work or it actually didn't worked. Having a smart client side evaluation reduces the amount of columns that needs to be selected and does not require to have a `ToList` prior `Select` method call, which avoids iterating the collection twice. Unfortunately, there is also a downside of such feature, when the developer does not understand how NHiberante linq provider works and writes the following query:
```C#
db.Orders.Select(o => GetValue(o)).ToList();

private string GetValue(Order order)
{
  return order.Employee?.Superior?.FirstName;
}
```
the query will work but it will trigger additional queries, which can cause performance issues. This could be fixed by adding a configuration option whether to allow additional queries to be executed while calling `ToList` and could be overridden for specific queries, but it is a matter of another issue/PR.

Back on the topic, with the last commit two issues were addressed:
- Add `NullableExpressionDetector` support for static properties, by flagging them as not nullable.
- Fix `DateTime.Kind` for functions that are supported by this PR. For other functions like `current_time`, I think a separate PR should be made whether we want to change them.

src/NHibernate/Linq/Visitors/NullableExpressionDetector.cs
```diff
@@ -128,6 +128,13 @@ private bool IsNullable(MemberExpression memberExpression, BinaryExpression equa
 		{
 			if (_functionRegistry.TryGetGenerator(memberExpression.Member, out _))
 			{
+				// The expression can be null when the member is static (e.g. DateTime.Now).
+				// In such cases we suppose that the value cannot be null.
+				if (memberExpression.Expression == null)
+				{
+					return false;
```
fredericDelaporte: What if a Linq query references a static member of the class in which it is coded?
```
public SomeClass
{
  private int? _someValue;

  public void SomeMethod()
  {
    var results = session.Query<Bidule>().Where(b => b.Truc != _someValue).ToList();
  }
}
```

We should probably check what is the member type nullability, or assume it to be nullable for staying on the safe side.

It also seems to me this fix may deserve a PR on its own, especially if this one cannot be merged soon.

maca88: > What if a Linq query references a static member of the class in which it is coded?

Then it will be pre-evaluated and treated as a parameter, as far as I know the only static properties that are not pre-evaluated are introduced with this PR. Also `NoPreEvaluationAttribute` works only for methods so I think we should be good by saying that they are not nullable.

maca88: The only way to create a nullable static property that is not pre-evaluated is by creating a custom `IHqlGeneratorForProperty` and implement `IAllowPreEvaluationHqlGenerator` that is introduced by this PR. If we want to have this configurable we could extend `IHqlGeneratorForProperty` similar as we extended `IHqlGeneratorForMethod` by adding `AllowsNullableReturnType` method.

fredericDelaporte: I had forget this, you are right. This PR is a year and a half old already.

src/NHibernate/Dialect/SQLiteDialect.cs
```diff
@@ -94,6 +101,31 @@ protected virtual void RegisterFunctions()
 
 			// NH-3787: SQLite requires the cast in SQL too for not defaulting to string.
 			RegisterFunction("transparentcast", new CastFunction());
+
+			// This does not generate a valid uuid, but SelectGUIDString property already uses this.
+			// See https://www.ietf.org/rfc/rfc4122.txt
+			// System.Guid does not seem to care about the uuid validity, but this may still cause
+			// other issues.
+			RegisterFunction("new_uuid", new SQLFunctionTemplate(NHibernateUtil.Guid, "randomblob(16)"));
```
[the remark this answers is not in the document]
maca88: Maybe we could use [the expression from SO](https://stackoverflow.com/a/41649754), which generates valid UUIDs instead. I've tested it locally on a table with 100k rows and it didn't generate any duplicates ([fiddle](https://www.db-fiddle.com/f/bZ8YtHwtN216xz1xbTRWzP/0) with 10k rows). We could also use it for `SelectGUIDString`, but I think that should be done in a separate PR.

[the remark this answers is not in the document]
maca88: Also there is another issue, we have to take into consideration `_binaryGuid` as now the following query will return an invalid guid string when `_binaryGuid` is `false`:
```C#
session.Query<Entity>()
  .Select(x => new { Guid = Guid.NewGuid().ToString() })
  .ToList();
```
we would need to change the registration to something like:
```C#
RegisterFunction("new_uuid", new SQLFunctionTemplate(NHibernateUtil.Guid, _binaryGuid ? "randomblob(16)" : "hex(randomblob(16))"));
```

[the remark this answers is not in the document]
maca88: I've changed my mind about this, it is not just about validity but also about uniqueness of the uuid which we cannot guarantee that the two metnioned approaches would produce the same uniquness as RFC-4122 standard does. It seems like that in [Sqlite 3.31](https://sqlite.org/releaselog/3_31_0.html) they've added the `uuid` function, but the latest `System.Data.SQLite.Core` package contains a lesser Sqlite version `3.30.1`, which prevents us from adding it into the dialect.

fredericDelaporte:
Review dismissed due to having to resolve conflicting changes in Oracle 10g dialect.

fredericDelaporte:
Furthermore, rebased, since I consider merging rather than squashing, so I have squashed @maca88 fix commits into the three main commits, adding co-author.

PR #1869, opened 2018-10-06
Replace an O(n) lookup in LINQ query parsing by an O(1) one

fredericDelaporte:
The HQL generators registry already uses dictionaries to resolve generators for methods found in LINQ expression. But some generators do not declare statically their supported methods and instead have
to be queried with a loop on them for support of methods encountered in the LINQ query.

The result of this lookup can be cached for avoiding calling this loop again on the next query using the method.
If #1868 is merged, this change will compensate for #1868 causing up to three such lookups by method.


fredericDelaporte:
I have included the missing async trivial regen from another PR.

PR #1871, opened 2018-10-12
Fix property-ref on a component's property

### LINKED ISSUE #1824: property-ref on a component's property causes "wrong number of columns" error
In the example https://github.com/nhibernate/nhibernate-core/issues/1234, if I add a new property to `bar` `dynamic-component`, a "property mapping has wrong number of columns" error is thrown.

```
<class name="foo">
    <id type="int" name="Id">
      <column name="Id"/>
      <generator class="native"/>
    </id>

    <dynamic-component name="MyProps">
      <many-to-one name="DynPointer" class="bar" property-ref="MyProps.DynValString2"/>
    </dynamic-component>
  </class>

  <class name="bar">
    <id type="int" name="Id">
      <column name="Id"/>
      <generator class="native"/>
    </id>
    <dynamic-component name="MyProps">
      <property name="DynValString2" type="string" unique="1"/>
      <property name="AnotherProperty" type="string" />
    </dynamic-component>
  </class>
```

fredericDelaporte:
Fixes #1824


src/NHibernate/Cfg/Configuration.cs
```diff
@@ -221,7 +221,7 @@ public string GetIdentifierPropertyName(string className)
 			public IType GetReferencedPropertyType(string className, string propertyName)
 			{
 				PersistentClass pc = GetPersistentClass(className);
-				Property prop = pc.GetProperty(propertyName);
+				Property prop = pc.GetReferencedProperty(propertyName);
```
fredericDelaporte: `GetProperty` yields properties of the entity, it does not go inside components for yielding the component property.

hazzik:
@fredericDelaporte why does it show my name on the tests?

hazzik:
@fredericDelaporte can you check if #1214 is related?

fredericDelaporte:
> @fredericDelaporte why does it show my name on the tests?

Because I have cherry-pick from you 4431485, squashed in it bab942a, then renamed the files and added the property triggering the failure and added a non-dynamic component for allowing checking it was also having the trouble.

> @fredericDelaporte can you check if #1214 is related?

I tend to think no, but I will check the test case.

fredericDelaporte:
#211 (NH v4.1, [NH-3480](https://nhibernate.jira.com/browse/NH-3480)) has broken #1214 case, which hinders me for diagnosing it.

Reverting #211 by keeping only the `UseLHSPrimaryKey` path [here](https://github.com/nhibernate/nhibernate-core/blob/5.1.3/src/NHibernate/Persister/Collection/AbstractCollectionPersister.cs#L226) fixes #1214 previously working test.
By the way the `UseLHSPrimaryKey` path seems to do quite the opposite to what the property means. #211 causes the `property-ref` on the collection key column to be ignored when generating the insert SQL for the collection, using instead the primary key name of owner.
But for the failing one, the original reported failure is still not there after my partial local reversion of #211, and it looks to me due to some other #211 consequences I have not yet pinpointed (symptom: identifier value of ManyA is put in the many-to-many table instead of the referenced property value).

hazzik:
>Because I have cherry-pick from your 4431485,

Ok.

Any reasons not to merge yet?

fredericDelaporte:
I was still investigating #1214. With #378 getting in the way, I have not yet ascertained it was not bound to the case fixed here, although it does still look unlikely to me.

fredericDelaporte:
#1214 root cause is a mapping error: using a property-ref on a property not flagged as unique. But NHibernate does not detect this error when validating mappings, and do not raise a good error when attempting to resolve an entity by an undeclared unique key.
So #1214 is not related.

PR #1872, opened 2018-10-14
Fix property ref handling

fredericDelaporte:
#378 (NH-3480, v4.1) was a wrong fix, with an heavy impact on the property ref handling, and causing bugs in many-to-many cases.

This PR adds a many-to-many test demonstrating the introduced bugs, reverts most of #378, and fixes its distinctive cause.

#378 case was a one-to-many with property ref failure, requiring having the one side having its identifier mapped as a component. Otherwise, no failure was occurring. This was due to the component equality logic always assuming its operand were components of the right class. Fixing the equality logic is enough to solve the case and have the collection loading work as "good" as other cases not using an entity with a composite key.

Now, the root cause of the bug is the collection loading mechanism ending up resolving the owning entity with the collection key as an id, while it is not the id in the property ref case. Fortunately, resolving the owner there has no usage with default NHibernate implementation. (It may be an issue for collections of custom types.) So having the right owner or an owner at all here does not cause issues indeed.
(Update: additional commits have been added for resolving the owner correctly in any case.)

#378 fix was to cause the collection key to become the id even when a property ref was used. But this was invalid for a many-to-many and inefficient for one-to-many, because it requires additional joins.

Moreover I think the root cause can be fixed by checking if the collection yields the owner id or a unique key, and either resolve the id from the key or perform a lookup by unique key instead of a lookup by id. I will look into this in a separated PR. (Update: done as additional commits.)

About this fix, #378 is indeed a regression introduced in NHibernate v4.1, so ideally, it should be backported to it and to v5.0 and v5.1. But few people (or maybe no-one)  have complain about it, so I do not think it deserves being back-ported.

I would still rather have this fixed in 5.2.


src/NHibernate.Test/PropertyRef/ManyToManyPropertyRefFixture.cs
```diff
@@ -0,0 +1,137 @@
...
+			FROM	ManyBs manybs0_
+			left outer join ManyB manyb1_ on manybs0_.ManyANumber = manyb1_.Number
+			WHERE  manybs0_.ManyBNumber =6 /# ?p0 #/
+			 */
+
+			Assert.That(loadedManyA.ManyBs.Count, Is.EqualTo(3));
+		}
+
+		[Test, Ignore("Not fixed yet")]
+		public void Getting_a_ManyA_object_with_fetchmode_join_will_work()
```
fredericDelaporte: This test case is originally meant for #1214. This test here is now failing again like initially reported. #1214 cannot be fixed without first fixing the regression caused by #378.
The other test was not failing in the original context of #1214, and is the one demonstrating #378 regression.

src/NHibernate/Persister/Collection/AbstractCollectionPersister.cs
```diff
@@ -226,45 +225,16 @@ public AbstractCollectionPersister(Mapping.Collection collection, ICacheConcurre
...
-				{
-					keyColumnNames[k] = col.GetQuotedName(dialect);
-					keyColumnAliases[k] = col.GetAlias(dialect, table);
-					k++;
-				}
-				joinColumnNames = keyColumnNames;
-			}
-			else
-			{
-				keyType = collection.Owner.Key.Type;
```
fredericDelaporte: This change was breaking the many-to-many case with property-ref. It was causing NHibernate to try insert in the many-to-many table a column named after the entity id instead of using the referenced property name column name.

src/NHibernate/Type/CollectionType.cs
```diff
@@ -471,16 +471,39 @@ public override bool IsDirty(object old, object current, bool[] checkable, ISess
 		}
 
 		/// <summary>
-		/// Get the key value from the owning entity instance.
+		/// Get the key value from the owning entity instance. It is usually the identifier, but it might be some
+		/// other unique key, in the case of a property-ref.
 		/// </summary>
 		public object GetKeyOfOwner(object owner, ISessionImplementor session)
```
fredericDelaporte: Ceasing to return the referenced property value from this method causes bug in the many-to-many case, causing it to insert in the many-to-many table the identifier instead of the referenced property value.

src/NHibernate/Type/ComponentType.cs
```diff
@@ -600,6 +600,11 @@ public override bool IsEqual(object x, object y, ISessionFactoryImplementor fact
 			{
 				return false;
 			}
+
+			var componentType = EntityMode == EntityMode.Poco ? ComponentTuplizer.MappedClass : typeof(IDictionary);
+			if (!componentType.IsInstanceOfType(x) || !componentType.IsInstanceOfType(y))
+				return false;
```
fredericDelaporte: `GetPropertyValues` calls below can only fail if the parameter is not of the expected type. So an additional check on the type of object to compare together has been added in the `IsEqual` implementation.

fredericDelaporte:
> Moreover I think the root cause can be fixed by checking if the collection yields the owner id or a unique key, and either resolve the id from the key or perform a lookup by unique key instead of a lookup by id. I will look into this in a separated PR.

I will add this as an additional commit here instead. #378 was dodging the trouble by causing the collection key to hold the owner id instead of the referenced property. So the lookup which not handling the unique key case was no more failing, since it was always the id. Now that it is reverted and we are back to the collection key being actually the mapped referenced property value, better fix the lookup, even if prior to #378, having it failing was almost never causing issues. (Only in #378 case, it was throwing; in other cases, failing was meaning a null return or even another entity, but apparently without causing much issues.)

I am now checking the `VERY HACKISH WORKAROUND` which is back with the revert, in case it deserve an additional commit here.

src/NHibernate/Engine/StatefulPersistenceContext.cs
```diff
@@ -801,10 +809,9 @@ public virtual object GetLoadedCollectionOwnerOrNull(IPersistentCollection colle
 			object loadedOwner = null;
 			// TODO: an alternative is to check if the owner has changed; if it hasn't then
 			// return collection.getOwner()
-			object entityId = GetLoadedCollectionOwnerIdOrNull(ce);
-			if (entityId != null)
+			if (ce.LoadedKey != null)
```
fredericDelaporte: Now that `GetCollectionOwner(object key, ...` correctly handles unique key, this small hack here attempting to resolve the id from the key is no more needed.
Moreover, all other calls to `GetCollectionOwner` were transmitting the collection key without this hack, so not always the id. (Due to bad naming, `DefaultInitializeCollectionEventListener.InitializeCollectionFromCache could let think this one was supplying the identifier, but indeed it is supplying the key too.)

src/NHibernate/Loader/Loader.cs
```diff
@@ -981,9 +981,18 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
 						obj =
 							InstanceNotYetLoaded(rs, i, persister, key, lockModes[i], optionalObjectKey,
 												 optionalObject, hydratedObjects, session);
+
+						// IUniqueKeyLoadable.CacheByUniqueKeys caches all unique keys of the entity, regardless of
+						// associations loaded by the query. So if the entity is already loaded, it has forcibly already
+						// been cached too for all its unique keys, provided its persister implement it. With this new
+						// way of caching unique keys, it is no more needed to handle caching for alreadyLoaded path
+						// too.
+						(persister as IUniqueKeyLoadable).CacheByUniqueKeys(obj, session);
```
fredericDelaporte: Here is the change for avoiding cache misses when looking-up by unique key.

Due to #1226, this caching of unique key has a bit evolved.
Prior to #1226, the caching by unique key was done only for entities not yet loaded, when they were loaded through a query with eager loads referencing the entity by its unique key and as their RHS, excepted in the many-to-many case (which was failing and is still failing).
For fixing #1226, this caching has been moved for occurring also with entities already loaded.
Now with this PR, it is moved back to the path of entities not yet loaded, and no more done for this already loaded. This is done because now, instead of caching only by unique key used by RHS side of associations loaded by the query, it is done for all alternate unique keys declared in the entity. So it does no more depends on which associations are eagerly  loaded in the query loading the entity, the cache feeding will be done wherever the entity is actually loaded.

This is an overhead for session usages which were not doing any eager load, but I think it is acceptable, because anyway the unique key caching is actually done only in case of `property-ref` usage in mappings, feature which should be used as little as possible. (Mapped unique keys do not trigger this caching. Only the property target of a property-ref is cached as a unique key.)

src/NHibernate/Engine/BatchFetchQueue.cs
```diff
@@ -220,16 +220,16 @@ public object[] GetCollectionBatch(ICollectionPersister collectionPersister, obj
...
 		/// </summary>
 		/// <param name="collectionPersister">The persister for the collection role.</param>
-		/// <param name="id">A key that must be included in the batch fetch</param>
+		/// <param name="key">A key that must be included in the batch fetch</param>
 		/// <param name="batchSize">the maximum number of keys to return</param>
 		/// <param name="checkCache">Whether to check the cache for uninitialized collection keys.</param>
 		/// <param name="collectionEntries">An array that will be filled with collection entries if set.</param>
 		/// <returns>An array of collection keys, of length <paramref name="batchSize"/> (padded with nulls)</returns>
-		internal object[] GetCollectionBatch(ICollectionPersister collectionPersister, object id, int batchSize, bool checkCache,
+		internal object[] GetCollectionBatch(ICollectionPersister collectionPersister, object key, int batchSize, bool checkCache,
```
fredericDelaporte: Renaming for avoiding confusion. It may be something else than an identifier, it is a collection key, which can be a unique key property instead. The code calling this is actually supplying a key, not an id.

src/NHibernate/Event/Default/DefaultInitializeCollectionEventListener.cs
```diff
@@ -62,9 +62,10 @@ public virtual void OnInitializeCollection(InitializeCollectionEvent @event)
 		}
 
 		/// <summary> Try to initialize a collection from the cache</summary>
-		private bool InitializeCollectionFromCache(object id, ICollectionPersister persister, IPersistentCollection collection, ISessionImplementor source)
+		private bool InitializeCollectionFromCache(
+			object collectionKey, ICollectionPersister persister, IPersistentCollection collection,
+			ISessionImplementor source)
```
fredericDelaporte: Renaming `id` to `collectionKey` for avoiding confusion. It may be something else than an identifier, it is a collection key, which can be a unique key property instead. The code calling this is actually supplying a key, not an id.

src/NHibernate/Cfg/XmlHbmBinding/CollectionBinder.cs
```diff
@@ -904,6 +905,7 @@ private void BindKey(HbmKey keyMapping, Mapping.Collection model)
 			else
 			{
 				keyValue = (IKeyValue) model.Owner.GetProperty(propRef).Value;
+				mappings.AddUniquePropertyReference(model.OwnerEntityName, propRef);
```
fredericDelaporte: This enable caching of the referenced entity by property-ref done on LHS.

src/NHibernate/Cfg/XmlHbmBinding/CollectionBinder.cs
```diff
@@ -821,13 +821,14 @@ private void BindManyToMany(HbmManyToMany manyToManyMapping, Mapping.Collection
 			var restrictedLaziness = manyToManyMapping.lazySpecified ? manyToManyMapping.lazy : (HbmRestrictedLaziness?) null;
 			InitLaziness(restrictedLaziness, manyToMany, true);
 
-			if(!string.IsNullOrEmpty(manyToManyMapping.propertyref))
+			manyToMany.ReferencedEntityName = GetEntityName(manyToManyMapping, mappings);
+
+			if (!string.IsNullOrEmpty(manyToManyMapping.propertyref))
 			{
 				manyToMany.ReferencedPropertyName = manyToManyMapping.propertyref;
+				mappings.AddUniquePropertyReference(manyToMany.ReferencedEntityName, manyToMany.ReferencedPropertyName);
```
fredericDelaporte: This enable caching of the referenced entity by property-ref done on RHS in the many-to-many case.

src/NHibernate.Test/NHSpecificTest/Properties/Mappings.hbm.xml
```diff
@@ -27,6 +27,8 @@
       </key>
       <one-to-many class="Account"/>
     </set>
+    <!-- Having it mapped after the collection referencing it can cause trouble when resolving the collection, if
+         not accounted for. So keep it mapped after for checking this case is still supported. -->
```
fredericDelaporte: This pre-existing test is already checking the case of a collection having a property-ref on a "component like" (here a `properties` mapping) property mapped after the collection.

src/NHibernate.Test/PropertyRef/KeyEntityPropertyRef.hbm.xml
```diff
@@ -0,0 +1,24 @@
...
+                   namespace="NHibernate.Test.PropertyRef">
+  <class name="A" lazy="false">
+    <id name="Id" generator="assigned"/>
+    <property name="Name"/>
+    <set name="CItems">
+      <key column="cid" property-ref="C"/>
+      <one-to-many class="B"/>
+    </set>
+    <!-- Having it mapped after the collection referencing it was causing trouble when resolving the collection -->
+    <many-to-one name="C"/>
```
fredericDelaporte: A test for this case was lacking. And it was broken, with or without this PR first commits. Mapping the referenced many-to-one before the collection was working.

src/NHibernate.Test/PropertyRef/KeyEntityPropertyRefFixture.cs
```diff
@@ -0,0 +1,114 @@
...
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				var entity =
+					s
+						.Query<A>()
+						.Single(a => a.Id != _aWithItemsId);
+
+				Assert.That(NHibernateUtil.IsInitialized(entity.CItems), Is.False, "a without items");
+				Assert.That(entity.CItems, Has.Count.EqualTo(0), "a without items");
```
fredericDelaporte: An additional failure was occurring in case of empty collection, so testing for this case too. The pre-existing test for the component case was already testing both case, empty and non-empty collections.

src/NHibernate/Engine/TwoPhaseLoad.cs
```diff
@@ -84,15 +85,28 @@ public static void InitializeEntity(object entity, bool readOnly, ISessionImplem
...
+			var collectionToResolveIndexes = new List<int>();
 			for (int i = 0; i < hydratedState.Length; i++)
 			{
 				object value = hydratedState[i];
 				if (!Equals(LazyPropertyInitializer.UnfetchedProperty, value) && !(Equals(BackrefPropertyAccessor.Unknown, value)))
 				{
+					if (types[i].IsCollectionType)
+					{
+						// Resolve them last, because they may depend on other properties if they use a property-ref
+						collectionToResolveIndexes.Add(i);
```
fredericDelaporte: The change in `CollectionType.GetKeyOfOwner` renders this change here no more mandatory. But I think it is better to remove a code path relying on `CollectionType.GetKeyOfOwner` "best effort", which the change here does.

src/NHibernate/Engine/TwoPhaseLoad.cs
```diff
@@ -84,15 +85,28 @@ public static void InitializeEntity(object entity, bool readOnly, ISessionImplem
 				log.Debug("resolving associations for {0}", MessageHelper.InfoString(persister, id, session.Factory));
 
 			IType[] types = persister.PropertyTypes;
+			var collectionToResolveIndexes = new List<int>();
```
hazzik: Probably give it an initial capacity to avoid resizing 

```suggestion
			var collectionToResolveIndexes = new List<int>(hydratedState.Length);
```

src/NHibernate/Loader/Loader.cs
hazzik: Did you mean `(persister as IUniqueKeyLoadable)?.CacheByUniqueKeys(obj, session);`?

fredericDelaporte: Or `((IUniqueKeyLoadable) persister).CacheByUniqueKeys(obj, session);`. I have to check if it must always be castable or not.

fredericDelaporte: I think your version is the safer one. Maybe there is some loading cases, when there are no associations loaded or none with a unique key, where a custom persister could be something not implementing this interface, although all NHibernate `ILoadable` current implementations do also implement it.

fredericDelaporte:
Rebased, then three new commits added for addressing the review, each with its own async regen.

For merging, I intend to squash fixups only, then do a merge commit.

PR #1873, opened 2018-10-15
 Fix many-to-many with property-ref 

### LINKED ISSUE #1214: NH-2180 - Many-To-Many with Property-ref fails to get subitems with FetchMode Join
**Armin Landscheidt** created an [issue](https://nhibernate.jira.com/browse/NH-2180) — <time datetime="2010-04-18T05:28:36.566Z" title="2010-04-18T05:28:36.566Z">18th April 2010, 5:28:36</time>:
> When I try to get an item with a many-to-many relation and property-ref on both sides NHibernate produces an exception. If I use FetchMode Select there is no Problem but with Join I get this error.
> 
> I tried to figure out what exactly the problem is but I don't get it. It seems to be a problem in building the mappings. While loading the object I think that AbstractPersister should create an UniqueKeyLoader. The decision if it should creat the kealoader is made while the mappingbuilding.
> 
> The patch contains the two testcases with fetchmode select and join.
>

fredericDelaporte:
Fixes the second bug causing #1214.

### DIFF
--- a/src/NHibernate.Test/PropertyRef/ManyToManyPropertyRefFixture.cs
+++ b/src/NHibernate.Test/PropertyRef/ManyToManyPropertyRefFixture.cs
@@ -87,7 +87,7 @@ namespace NHibernate.Test.PropertyRef
 			Assert.That(loadedManyA.ManyBs.Count, Is.EqualTo(3));
 		}
 
-		[Test, Ignore("Not fixed yet")]
+		[Test]
 		public void Getting_a_ManyA_object_with_fetchmode_join_will_work()
 		{
 			ManyA loadedManyA;
@@ -131,7 +131,7 @@ System.Collections.Generic.KeyNotFoundException: Der angegebene Schlüssel war n
    bei NHibernate.Type.EntityType.LoadByUniqueKey(String entityName, String uniqueKeyPropertyName, Object key, ISessionImplementor session) in C:\Users\Armin\Projects\NHibernate\branches\2.1.x\nhibernate\src\NHibernate\Type\EntityType.cs:Zeile 552.
 			 */
 
-			Assert.That(loadedManyA.ManyBs.Count, Is.EqualTo(3));
+			Assert.That(loadedManyA.ManyBs, Has.Count.EqualTo(3).And.None.Null);
 		}
 	}
 }
--- a/src/NHibernate/Persister/Collection/BasicCollectionPersister.cs
+++ b/src/NHibernate/Persister/Collection/BasicCollectionPersister.cs
@@ -271,18 +271,36 @@ namespace NHibernate.Persister.Collection
 				IAssociationType elementType = (IAssociationType) ElementType;
 				if (rhs.Equals(elementType.GetAssociatedJoinable(Factory)))
 				{
-					return ManyToManySelectFragment(rhs, rhsAlias, lhsAlias, collectionSuffix);
+					return ManyToManySelectFragment(rhs, rhsAlias, lhsAlias, collectionSuffix, elementType);
 				}
 			}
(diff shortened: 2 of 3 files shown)


src/NHibernate.Test/PropertyRef/ManyToManyPropertyRefFixture.cs
```diff
@@ -131,7 +131,7 @@ bei System.ThrowHelper.ThrowKeyNotFoundException()
    bei NHibernate.Type.EntityType.LoadByUniqueKey(String entityName, String uniqueKeyPropertyName, Object key, ISessionImplementor session) in C:\Users\Armin\Projects\NHibernate\branches\2.1.x\nhibernate\src\NHibernate\Type\EntityType.cs:Zeile 552.
 			 */
 
-			Assert.That(loadedManyA.ManyBs.Count, Is.EqualTo(3));
+			Assert.That(loadedManyA.ManyBs, Has.Count.EqualTo(3).And.None.Null);
```
fredericDelaporte: `And.None.Null` is required because there was two bugs: one causing an exception (the reported one) and another one, prevented by the failure, causing the children to be null with this test case.

src/NHibernate/Cfg/XmlHbmBinding/CollectionBinder.cs
```diff
@@ -821,13 +821,14 @@ private void BindManyToMany(HbmManyToMany manyToManyMapping, Mapping.Collection
 			var restrictedLaziness = manyToManyMapping.lazySpecified ? manyToManyMapping.lazy : (HbmRestrictedLaziness?) null;
 			InitLaziness(restrictedLaziness, manyToMany, true);
 
-			if(!string.IsNullOrEmpty(manyToManyMapping.propertyref))
+			manyToMany.ReferencedEntityName = GetEntityName(manyToManyMapping, mappings);
+
+			if (!string.IsNullOrEmpty(manyToManyMapping.propertyref))
 			{
 				manyToMany.ReferencedPropertyName = manyToManyMapping.propertyref;
+				mappings.AddUniquePropertyReference(manyToMany.ReferencedEntityName, manyToMany.ReferencedPropertyName);
```
fredericDelaporte: Root cause of the reported bug: a many-to-many rhs is like a one-to-many and like an actual `one-to-many`, it needs to promote the property-ref as a unique property reference for allowing the loader to eager load it.

src/NHibernate/Persister/Collection/BasicCollectionPersister.cs
```diff
@@ -271,18 +271,36 @@ public override string SelectFragment(IJoinable rhs, string rhsAlias, string lhs
...
+			// handling of the not-found feature.)
+			var elementColumnNames = string.IsNullOrEmpty(elementType.RHSUniqueKeyPropertyName)
+				? rhs.KeyColumnNames
+				// rhs is the entity persister, it does not handle being referenced through an unique key by a
+				// collection and always yield its identifier columns as KeyColumnNames. We need to resolve the
+				// key columns instead.
+				// 6.0 TODO: consider breaking again that IJoinable.SelectFragment interface for transmitting
+				// the OuterJoinableAssociation instead of its Joinable property. This would allow to get the
+				// adequate columns directly instead of re-computing them.
+				: ((IPropertyMapping) rhs).ToColumns(elementType.RHSUniqueKeyPropertyName);
```
fredericDelaporte: Fix of the additional bug, see all the comments added in code for the rational.

fredericDelaporte:
This one is not in the 5.2 milestone yet, but I consider this as a follow up to #1872, and would rather have it in.

So,

PR #1875, opened 2018-10-16
Improve exception message in case of duplicated column

increddibelly:
I wasted a morning figuring this out. A more explicit exception would have helped me tremendously.
see [my post on stackoverflow](https://stackoverflow.com/questions/10689054/fluent-nhibernate-indexoutofrange/52834741#52834741)


fredericDelaporte:
The change does not compile.

Please also add a test case with something like `Assert.That(() => someAction, Throws.InstanceOf<PropertyValueException>());

increddibelly:
let me get this right.

hazzik:
Don’t need to create a new PR. You can update this one.

increddibelly:
I've fixed the build and added a test. I may need a little help getting the tests running - none of the tests ran on my machine due to missing database connection.

src/NHibernate.Test/Async/NHSpecificTest/MultipleFieldsMappedToOneColumn/Fixture.cs
```diff
@@ -0,0 +1,125 @@
...
+
+	public class CorrectlyMappedEntity
+	{
+		public virtual long Id { get; set; }
+		public virtual long FirstValue { get; set; }
+		public virtual long SecondValue {get; set; }
+	}
+
+	[TestFixture]
+	public class FixtureAsync : TestCaseMappingByCode
```
fredericDelaporte: The test should not be added in the `Async` subfolder. This folder contains only generated files, by the async generator. Move this in NHibernate.Test\NHSpecificTest, and rename it `Fixture`.

src/NHibernate.Test/Async/NHSpecificTest/MultipleFieldsMappedToOneColumn/Fixture.cs
```diff
@@ -0,0 +1,125 @@
...
+			{
+				using (var session = OpenSession())
+				using (var transaction = session.BeginTransaction())
+				{
+					
+					session.Save(e);
+					transaction.Commit();
+				}
+			}
+			catch (ArgumentOutOfRangeException arex)
```
fredericDelaporte: Your change is about getting something better than the arex, isn't it? So it should check for the expected result with your change, not for the old result.

By the way, better pin down which call trigger the exception and assert it.
Unfortunately with your test, it will be sometimes the `Save` other times the commit. This is due to using the `native` identifier strategy, which may change `Save` behavior depending on the target database.
So well, remove the try catch and use something like:

```c#
Assert.That(
    () =>
    {
        session.Save(e);
        transaction.Commit();
    },
    Throws.TypeOf<PropertyValueException>().And.Message.Contains("(Duplicate column mapping?)"));

File: src/NHibernate.Test/Async/NHSpecificTest/MultipleFieldsMappedToOneColumn/Fixture.cs [hunk 3]
```diff
@@ -0,0 +1,125 @@
...
+			{
+				result = arex;
+			}
+
+			// assert
+			Assert.That(result != null);
+			Assert.That(result.Message.Contains("(Duplicate column mapping?)"));
```
fredericDelaporte: Please do not use such assert. They generate very poor error messages. Better use
```c#
Assert.That(result, Is.Not.Null.And.Property("Message").Contains("(Duplicate column mapping?)"));
```
In case of constraint violation, the error message will be way better than just "expected true but was false".
Anyway with my previous comment, the above code should be removed.

src/NHibernate.Test/Async/NHSpecificTest/MultipleFieldsMappedToOneColumn/Fixture.cs
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		[Test]
+		public void ShouldNotThrow()
+		{
+			// arrange
+			ArgumentOutOfRangeException result = null;
+			var e = new BadlyMappedEntity
```
fredericDelaporte: That looks like a bit too fast copy-paste.

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -2456,6 +2454,10 @@ protected int Dehydrate(object id, object[] fields, bool[] includeProperty, bool
 						PropertyTypes[i].NullSafeSet(statement, fields[i], index, includeColumns[i], session);
 						index += ArrayHelper.CountTrue(includeColumns[i]); //TODO:  this is kinda slow...
 					}
+					catch (ArgumentOutOfRangeException arex)
```
fredericDelaporte: It seems the exception is an `IndexOutOfRangeException`, not an `ArgumentOutOfRangeException`.

Anyway, it would be better to avoid the `IndexOutOfRangeException` by detecting the root trouble sooner. It should be doable when this class is constructed, or maybe even sooner. Then the exception would likely be thrown when compiling mappings, which is better.

src/NHibernate.Test/Async/NHSpecificTest/MultipleFieldsMappedToOneColumn/Fixture.cs
increddibelly: done.

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
increddibelly: I'm looking at the error right now, it says:

> NHibernate.PropertyValueException : Error dehydrating property value for PatientService.Logic.Persistence.Models.PersistentPatient.EncryptedObsolete
>   ----> System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection.
> Parameter name: index

I agree, this should break when compiling mappings. That's a change I'm not comfortable with, though. I _can_ improve the error message :-)

increddibelly:
thank you Frederic. I'm sure you tire of having to explain this again and again, but I'm doing this while my normal work is compiling, so I hope we can meet halfway and get an improvement in.

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
fredericDelaporte: The error in TeamCity for SQL-Server is an `IndexOutOfRangeException`, see [here](https://teamcity.jetbrains.com/viewLog.html?buildId=1685546&buildTypeId=bt7) (use anonymous login option if you do not have a login). With NHibernate Teamcity configuration, all other databases builds depend on the "NHibernate (NHibernate)" build, and are not actually executed if this first one has failed, so we cannot check the error for other databases tested with Teamcity.
But on this previous appveyor [build](https://ci.appveyor.com/project/nhibernate/nhibernate-core/builds/19552572) for your PR, we can see the error is not the same depending on the target database:
- [PostgreSQL](https://ci.appveyor.com/project/nhibernate/nhibernate-core/builds/19552572/job/a9wubk2f86rjjo2p/tests): `ArgumentOutOfRangeException`.
- [Firebird](https://ci.appveyor.com/project/nhibernate/nhibernate-core/builds/19552572/job/xe92qy8gumnmo5ih/tests) apparently no exception but failure of asserts (which one, hard to say, since that were the previous assert, just checking boolean results).
- [MySql](https://ci.appveyor.com/project/nhibernate/nhibernate-core/builds/19552572/job/3fg1pe8a40xcx50c/tests): `IndexOutOfRangeException`.
- [SQLite](https://ci.appveyor.com/project/nhibernate/nhibernate-core/builds/19552572/job/etgwegxmlgwx0oj4/tests): `ArgumentOutOfRangeException`.
- [SQL-Server](https://ci.appveyor.com/project/nhibernate/nhibernate-core/builds/19552572/job/cnreul4cj6cj4cjq/tests): `IndexOutOfRangeException`.

So it is now clear that catching the exception and trying to translate it to a more meaningful error message is not appropriate: it will not be reliable, since the thrown exception depends on the target database.

By the way, all the links above are provided by GitHub on this PR. On a commit, click the status icon (red cross or green check) to see the list of builds, and click on their "Details" link. Or alternatively for the last commit, see the checks panel at the bottom of the PR.

fredericDelaporte:
> thank you Frederic. I'm sure you tire of having to explain this again and again, but I'm doing this while my normal work is compiling, so I hope we can meet halfway and get an improvement in.

The trouble is, your changes seems not tested locally. As they are currently, in my opinion, they can only fail, whatever the target database. You really should take the time to setup your local environment for testing your changes. That is way more efficient than blind pushing, waiting for builds and seeing them fail. And it avoids time consuming noise for reviewers.

Moreover, as explained by contributing guidelines, an async generation has to be done.

If for some reasons you cannot compile/test/generate async locally, better just open an issue than attempting a PR. But of course that will mean someone else will have to tackle the subject, which can take some time to start happening.

increddibelly:
Thank you very much Frederic, I wish I had more time to help.

alojoshka:
Is it possible to disable this error message wtih any setting and return to previous handling? We used this feature as a part of functionality in a huge project, using integer and object property mapped to the same column, just like it  is used in EF6, and it was very flexible option for setting high performance projections.

fredericDelaporte:
@alojoshka, no, there are no settings for this. The check is done at a point where otherwise an invalid update or insert query would have been generated. It does not seem to make sense to allow invalid queries to be generated.

But maybe you can fix this in your mappings instead, by switching one of the redundant property to `insert` and `update` `false`.

alojoshka:
Thanks, setting the duplicate column as .ReadOnly() in the fluent mapping solves the problem

PR #1878, opened 2018-10-18
Fix extra-lazy collection forgetting changes

### LINKED ISSUE #1874: Item in child collection not being removed
Removing an item from a collection mapped as extra-lazy has sometimes no effect: the item stays in the collection.
This happens when the remove is done while the collection is not yet initialized, and if it is enumerated after the remove. In such case, the enumeration not only yields the item, but also at later flush, the item is not deleted from the collection.

Corresponding Hibernate issue: https://hibernate.atlassian.net/browse/HHH-5742

Initial report follows:

---

Not sure if this is a bug or something I am doing but I have an entity with a child collection, exposed as an IEumerable<>  and I am having issues removing the item:

```
public class Parent {
    private ISet<ChildItem> _childItems;

    public virtual IEnumerable<ChildItem> ChildItems
    {
            get { return _childItems; }
    }

    public virtual void RemoveChildItem(ChildItem item)
    {
           ...
            _childItems.Remove(item);
           ...
    }
}
```
I do the following to remove the item:

`myChildItem.Parent.RemoveChildItem(myChildItem);`

If I stick a breakpoint after the `_childItems.Remove(item);` line then I discover that the _childItems collection still contains the item.

I think it has something with initialising the _childItems collection (extra lazy load) because if I add a check before it like:

`if (!_childItems.Any(c => c == item)) throw new Exception();`

then the `_childItems.Remove(item);` works correctly. Likewise if I put a breakpoint before the remove and inspect the collection (causing it to initialise), then it works.

fredericDelaporte:
Fixes #1874

### DIFF
--- a/src/NHibernate.Test/Async/Extralazy/ExtraLazyFixture.cs
+++ b/src/NHibernate.Test/Async/Extralazy/ExtraLazyFixture.cs
@@ -10,6 +10,7 @@
 
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework;
 
 namespace NHibernate.Test.Extralazy
@@ -31,6 +32,16 @@ namespace NHibernate.Test.Extralazy
 		protected override string CacheConcurrencyStrategy
 		{
 			get { return null; }
+		}
+
+		protected override void OnTearDown()
+		{
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				s.Delete("from System.Object");
+				t.Commit();
+			}
 		}
 
 		[Test]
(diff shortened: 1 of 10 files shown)


fredericDelaporte:
Possible breaking-change (as demonstrated by NH-739 failure): `AbstractPersistentCollection.AfterInitialize` does no more perform queued operation. Queued operation are now run by a later call to a new method, `ApplyPendingOperations`. Concrete custom implementations relying on the queued operation to be done by their base `AfterInitialize` will need to be changed accordingly.

hazzik:
@fredericDelaporte can you check if this has fixed #1295?

fredericDelaporte:
No unit tests are provided there, and the opener states he was not able to fully reproduce it outside of his application. Moreover he does not states explicitly his case was an extra-lazy case, but it seems it can only be one.
This said, the part he was reproducing is no more reproducible since this PR.

PR #1389, opened 2017-10-18
Add support for out/ref Nullable parameters of proxied methods

pmattmann:
Add support to out/ref Nullable<T> parameters of proxied methods.

```c#
public class MyEntity {
  private int? _data1 = 1;
  private int? _data2 = 2;

  public virtual void GetData(out int? d1, out int? d2) {
    d1 = _data1;
    d2 = _data2;
  }
}
```

PR #1884, opened 2018-10-24
Fix attempt of static proxies to call base method for abstract classes

hazzik:
Previously it was failing PEVerify similarly to #1728 but for abstract classes (this is possible in case of a polymorphic entities). Unlike interfaces the abstract base class can go into a situation when lazy initializer is not yet available, eg. code in a constructor.

### DIFF
--- a/src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs
+++ b/src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs
@@ -46,10 +46,10 @@ namespace NHibernate.Test.StaticProxyTest
 			public PublicInterfaceTestClass()
 			{
 				// Check access to properties from the default constructor do not fail once proxified
-				Assert.That(Id, Is.Zero);
-				Assert.That(Name, Is.Null);
 				Id = -1;
-				Name = string.Empty;
+				Assert.That(Id, Is.EqualTo(-1));
+				Name = "Unknown";
+				Assert.That(Name, Is.EqualTo("Unknown"));
 			}
 		}
 
@@ -63,10 +63,10 @@ namespace NHibernate.Test.StaticProxyTest
 			{
 				// Check access to properties from the default constructor do not fail once proxified
 				IPublic pub = this;
-				Assert.That(pub.Id, Is.Zero);
-				Assert.That(pub.Name, Is.Null);
 				pub.Id = -1;
-				pub.Name = string.Empty;
+				Assert.That(pub.Id, Is.EqualTo(-1));
+				pub.Name = "Unknown";
+				Assert.That(pub.Name, Is.EqualTo("Unknown"));
 			}
 		}
 
@@ -75,10 +75,10 @@ namespace NHibernate.Test.StaticProxyTest
 		{
 			protected AbstractTestClass()
 			{
-				Assert.That(Id, Is.Zero);
-				Assert.That(Name, Is.Null);
 				Id = -1;
+				Assert.That(Id, Is.Zero);
 				Name = "Unknown";
+				Assert.That(Name, Is.Null);
 			}
 
 			public abstract int Id { get; set; }
--- a/src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs
+++ b/src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs
@@ -33,13 +33,41 @@ namespace NHibernate.Test.StaticProxyTest
 
 		public interface IPublic
 		{
-			int Id { get; }
+			int Id { get; set; }
+			string Name { get; set; }
 		}
 
 		[Serializable]
 		public class PublicInterfaceTestClass : IPublic
 		{
 			public virtual int Id { get; set; }
+			public virtual string Name { get; set; }
+
+			public PublicInterfaceTestClass()
+			{
+				// Check access to properties from the default constructor do not fail once proxified
+				Assert.That(Id, Is.Zero);
+				Assert.That(Name, Is.Null);
+				Id = -1;
+				Name = string.Empty;
+			}
+		}
+
+		[Serializable]
+		public class PublicExplicitInterfaceTestClass : IPublic
+		{
+			int IPublic.Id { get; set; }
+			string IPublic.Name { get; set; }
+
+			public PublicExplicitInterfaceTestClass()
+			{
+				// Check access to properties from the default constructor do not fail once proxified
+				IPublic pub = this;
+				Assert.That(pub.Id, Is.Zero);
+				Assert.That(pub.Name, Is.Null);
+				pub.Id = -1;
+				pub.Name = string.Empty;
+			}
 		}
 
 		[Serializable]
@@ -172,7 +200,7 @@ namespace NHibernate.Test.StaticProxyTest
 				// By way of the "proxy" attribute on the "class" mapping, an interface to use for the
 				// lazy entity load proxy instead of the persistentClass can be specified. This is "translated" into
 				// having an additional interface in the interface list, instead of just having INHibernateProxy.
-				// (Quite a loosy semantic...)
+				// (Quite a loose semantic...)
 				new HashSet<System.Type> { typeof(INHibernateProxy), typeof(IPublic) },
 				null, null, null);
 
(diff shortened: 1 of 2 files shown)


hazzik:
@fredericDelaporte what if we move `ProxyBuilderHelper.CallDefaultBaseConstructor(IL, parentType);` in `ImplementConstructor` after we set `__lazyInitializer` and `__proxyInfo`?

fredericDelaporte:
> @fredericDelaporte what if we move `ProxyBuilderHelper.CallDefaultBaseConstructor(IL, parentType);` in `ImplementConstructor` after we set `__lazyInitializer` and `__proxyInfo`? 

I doubt this will be considered a valid program. It looks highly probable to me that the call to the base constructor has to be the first thing done, per the .Net specification.

I think we have to yield default(T) for the abstract case (with of course a special case for void).

According to LINQ Pad 5,
```c#
static T Default<T>()
{
    return default(T);
}
```

Generates this IL:
```
IL_0000:  nop         
IL_0001:  ldloca.s    00 
IL_0003:  initobj     UserQuery.T
IL_0009:  ldloc.0     
IL_000A:  stloc.1     
IL_000B:  br.s        IL_000D
IL_000D:  ldloc.1     
IL_000E:  ret         
```

000A to 000D looks to me as no-op, I do not really understood their purpose. But well, even if they are required, this generated IL looks not too bad, it should not be too hard to yield the default value.

Now I consider pursuing @PleasantD PR (#1882), because it sounds likely to me we will have to do this for fixing another valid case: a proxified class implementing explicitly an interface, without any implicit implementation for some members of the interface, and without having the interface being the proxy base type. In such case, the interface members can only be overridden by being re-implemented explicitly.
(But we may also have an additional difficulty to handle: the call forwarded to the delegated state would have to call the explicit implementation on the delegated state too. I think we already generate the adequate code, but a test checking this would be better.)

src/NHibernate/Proxy/NHibernateProxyBuilder.cs
```diff
@@ -393,9 +391,28 @@ private static void ImplementGetLazyInitializer(TypeBuilder typeBuilder, MethodI
...
 			IL.Emit(OpCodes.Bne_Un, skipBaseCall);
 
-			IL.Emit(OpCodes.Ldarg_0);
-			EmitCallMethod(IL, OpCodes.Call, method);
-			IL.Emit(OpCodes.Ret);
+			if (method.IsAbstract)
+			{
+				if (!method.ReturnType.IsValueType)
+				{
+					IL.Emit(OpCodes.Ldnull);
```
fredericDelaporte: Ok, I should have read your PR code files before commenting, you have already done the IL.

hazzik:
> Now I consider pursuing...

Only thing in the PR which is actual are the flags. The code we generate is valid (except the one fixed by this PR). However, even if we fix the flags and make a program "valid" it will still have the issues you have demonstrated before: the code will behave correctly _only_ is the a variable holding the proxy is declared as an interface. And, there could be some potential issues with hidden traps.

fredericDelaporte:
Ok, so it should be possible to initialize the lazy initializer before calling the base constructor. But then it will be a sizeable possible breaking change: this will cause proxies of classes initializing some of their properties in their constructor to force their initialization on construction, defeating the proxy purpose.

Of course, we can consider it as to be expected, since the class access its state during construction. Having it still lazy and actually ignoring those calls may defeat the constructor intended behavior. But for classes doing some initialization useful for new entities but unneeded for loaded one, like initializing collections, this would cause them to be no more actually lazy.

hazzik:
>Ok, so it should be possible to initialize the lazy initializer before calling the base constructor....

I've thought about this after I wrote my comment. Yes, it will trigger the initialization if a _property_ is accessed in the constructor, and we do not want this (as per what you've just said).

It's fairly common to have following code in the entities:

```csharp
public class A 
{
  ICollection<B> _children = new List<B>();

  public virtual IEnumerable<B> Children => _children; 
}
```
Or
```csharp
public class A 
{
  public virtual IEnumerable<B> Children { get; private set; } = new List<B>();
}
```
These two examples would behave as intended, however the following code would not:
```csharp
public class A 
{
  public virtual IEnumerable<B> Children { get; private set; }

  public A()
  {
    Children = new List<B>();
  }
}
```

But the later code is having a compilation warning "Access of virtual members in a constructor".

hazzik:
>is indeed more like this

I have implemented this on intention. Exactly because of the case you're describing in "Now I consider pursuing..." It is just easier to implement two methods explicit and implicit in case of interfaces than pick and choose what one kind method to implement.  

The case with ILazyCollectionInitializer will still generate semi-valid code which will behave correctly only when the instance is declared as the interface (it cannot be declared as proxy per se).

src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs
```diff
@@ -42,6 +42,17 @@ public class PublicInterfaceTestClass : IPublic
 			public virtual int Id { get; set; }
 		}
 
+		[Serializable]
+		public abstract class AbstractTestClass : IPublic
+		{
+			protected AbstractTestClass()
+			{
+				Id = -1;
+			}
```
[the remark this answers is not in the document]
hazzik: Done

src/NHibernate/Proxy/NHibernateProxyBuilder.cs
```diff
@@ -393,9 +391,28 @@ private static void ImplementGetLazyInitializer(TypeBuilder typeBuilder, MethodI
 			IL.Emit(OpCodes.Ldnull);
 			IL.Emit(OpCodes.Bne_Un, skipBaseCall);
 
-			IL.Emit(OpCodes.Ldarg_0);
-			EmitCallMethod(IL, OpCodes.Call, method);
-			IL.Emit(OpCodes.Ret);
+			if (method.IsAbstract)
+			{
```
fredericDelaporte: Thinking about it again, this could be an unlikely breaking change. That requires some weird class, but still.
Consider:
```c#
public interface IA
{
    int Id { get; set; }
}

public class A: IA
{

    public A()
    {
        IA ia = this;
        ia.Id = -1;
        Assert(ia.Id, Is.EqualTo(-1);
    }
}
```

With previous logic, it would work, since the interface call was redirected to the implicit implementation.

With the new logic, it will fail, since the interface call will now be no-op and yielding default value, even when an implicit implementation is available.

I think it is such a corner case that it is acceptable. Moreover supporting this case robustly would require a likely costly search for a matching implicit implementation on the base type, in case the interface is explicitly implemented by it. (I know that other case is not yet supported by the way, due to the way interface method implementations are currently declared.)

Still I add a possible breaking change here.

fredericDelaporte:
Still thinking about all that, it seems to me the static proxy lacks tests for #1389, `out` and `ref` parameters. The null initializer path has maybe some flaws for these cases too.

fredericDelaporte:
I have added some tests for the explicit interface case, since the change here is also enough to have it no more failing validation. I have also added checks for constructor calls in other cases, since this PR is about having those calls not failing (in the abstract base class case, but better check the more usual case too).

I do not think I will add tests for out or ref here even if they work (not checked yet), it seems a bit too unrelated.

src/NHibernate/Proxy/NHibernateProxyBuilder.cs
fredericDelaporte: v5.1 one. About possible breaking changes, only released versions count.

fredericDelaporte:
> > In previous version it fails with NullReferenceException.
>
> Actually, I've applied your last commit onto master and it is even worse than NullReferenceException. Explicit interface fails with StackOverflowException.

We cannot be speaking of the same thing. My comment is not about the commit I have added, it is about the example I have written in the comment.

The commit I have added test another case, which I know is failing onto master, but that your PR happens to fix altogether. That is why I have added it into your PR, since your PR also fix that other case.

src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs
```diff
@@ -114,7 +158,7 @@ public void VerifyProxyForClassWithInternalInterface()
 		}
 
 		[Test]
-		public void VerifyProxyForClassWithAdditionalInterface()
+		public void VerifyProxyOnInterface()
```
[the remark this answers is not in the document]
fredericDelaporte: Not really. This case tests a `proxy: IPublic`, while the other case tests `proxy: TestClass`, with `TestClass` being also `TestClass: IPublic`.

The names were not allowing to do the distinction. `ProxyOnInterface` seems to me well more explicit about the fact we test a proxy based on an interface, not a proxy based on a class which also happens to implement some interface.

src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs
```diff
@@ -158,7 +158,7 @@ public void VerifyProxyForClassWithInternalInterface()
 		}
 
 		[Test]
-		public void VerifyProxyOnInterface()
+		public void VerifyProxyForClassWithAdditionalInterface()
```
fredericDelaporte: This test generates a proxy class solely based on the interfaces. It is not actually a proxy of the class. Going to that name is going back to the confusion. Even worse combine with the other renaming.

We now have:

- `VerifyProxyForClassWithInterface`: does test a `class proxy: PublicInterfaceTestClass, INHibernateProxy, IPublic`.
- `VerifyProxyForClassWithAdditionalInterface`: does actually test a `class proxy: INHibernateProxy, IPublic`, without any direct relationship to `PublicInterfaceTestClass`. `Additional` is far from telling this.
   The fact that obtaining this requires to add the interface in the interface list of the proxy factory is a (bad) implementation detail. It seems way more important to me to describe the kind of proxy we test, rather than the way we obtain such proxy.

hazzik:
> My comment is not about the commit I have added, it is about the example I have written in the comment.

But the example in the comment and a test in the commit are [almost] the same.

hazzik:
I've applied tests to the 5.1 branch. The "class with explicit interface" fails with StackOverflowException. This is because it is not possible to call explicit interface members on a base class without reflection.

src/NHibernate/Proxy/NHibernateProxyBuilder.cs
```diff
@@ -393,9 +391,28 @@ private static void ImplementGetLazyInitializer(TypeBuilder typeBuilder, MethodI
...
+				{
+					var local = IL.DeclareLocal(method.ReturnType);
+					IL.Emit(OpCodes.Ldloca, local);
+					IL.Emit(OpCodes.Initobj, method.ReturnType);
+					IL.Emit(OpCodes.Ldloc, local);
+				}
+
+				IL.Emit(OpCodes.Ret);
+			}
+			else
```
hazzik: With the current tests this branch is not needed at all. If the `if` condition replace with `if(true)` all tests still pass.

I'm considering to make a breaking change and throw exception if `__lazyInitializer` is null.

hazzik:
I finally understood what you have meant, @fredericDelaporte and how the comment differs to the committed tests. What you describe in a comment as a "weird" class in fact could be fairly common code. However, ReSharper complains on such a code with "virtual call in constructor" [warning](https://www.jetbrains.com/help/rider/VirtualMemberCallInConstructor.html). 

Because `__lazyInitializer` is null only in the constructor calling base class methods does not make sense as the state change will be lost. And no-op should be fine [in some cases]. Consider following example:

```csharp
public class A
{
  public virtual int X { get; set; }
  public A()
  {
    X = -1;
  }
}

var a = new AProxy();
// a.HibernateLazyInitializer.SetImplementation(new A());
Assert.That(a.X, Is.Equal(-1));
```
This will pass, because the implementation will have a real constructor. 

So far, we have following options for a solution:

# always throw exception if `__lazyInitializer` is null. 
# throw exception if `__lazyInitializer` is null _and_ method is not void. Void methods are no-op. 
# always no-op.
# best effort in calling base class (chose an option from 1-3 if base method is not callable)
# set `__lazyInitializer` before calling a base constructor.

fredericDelaporte:
>I've applied tests to the 5.1 branch. The "class with explicit interface" fails with StackOverflowException. This is because it is not possible to call explicit interface members on a base class without reflection.

We are still not speaking of the same test.

This test does not fail on v5.1:

```c#
public interface IA
{
	int Id { get; set; }
}

public class A: IA
{

	public A()
	{
		IA ia = this;
		ia.Id = -1;
		Assert.That(ia.Id, Is.EqualTo(-1));
	}
}

[Test]
public void VerifyA()
{
	var factory = new StaticProxyFactory();
	factory.PostInstantiate(
		typeof(A).FullName,
		typeof(A),
		new HashSet<System.Type> {typeof(INHibernateProxy)},
		null, null, null);
	var proxy = factory.GetProxy(1, null);
	Assert.That(proxy, Is.Not.Null);
	Assert.That(proxy, Is.InstanceOf<IA>());
	Assert.That(proxy, Is.InstanceOf<A>());
	// Check interface and implicit implementations do both call the delegated state
	var state = new A { Id = 5 };
	proxy.HibernateLazyInitializer.SetImplementation(state);
	var pub = (IA) proxy;
	var ent = (A) proxy;
	Assert.That(pub.Id, Is.EqualTo(5), "IA.Id");
	Assert.That(ent.Id, Is.EqualTo(5), "A.Id");
	ent.Id = 10;
	Assert.That(pub.Id, Is.EqualTo(10), "IA.Id");
	Assert.That(state.Id, Is.EqualTo(10), "A.Id");
}
```

I was expecting it to fail with this PR, since the interface implementation has been changed for doing nothing or yielding default value. But instead, it does still work.

hazzik:
> But instead, it does still work. (The assert in the constructor still passes.)

Do you know why? It works because of #1882 :-) the compiler choses the override instead of explicit interface implementation because of incorrect attributes.

hazzik:
With option 3 user can see a NullReferenceException if they try access to the state.

fredericDelaporte:
Yes, that is why I consider what does currently this PR is still better.

src/NHibernate/Proxy/NHibernateProxyBuilder.cs
```diff
@@ -393,9 +395,76 @@ private static void ImplementGetLazyInitializer(TypeBuilder typeBuilder, MethodI
...
+				{
+					var local = IL.DeclareLocal(method.ReturnType);
+					IL.Emit(OpCodes.Ldloca, local);
+					IL.Emit(OpCodes.Initobj, method.ReturnType);
+					IL.Emit(OpCodes.Ldloc, local);
+				}
+
+				IL.Emit(OpCodes.Ret);
+			}
+			else if (method.IsPrivate)
```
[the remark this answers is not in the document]
fredericDelaporte: I was not considering to do the effort of calling a base explicit implementation, since that is already not easily doable in C# code, and previous releases were not supporting it anyway. So no-op would have suited me. But now that it is done and tested, why not keeping it. It is still better.

PR #1891, opened 2018-10-27
Recompute hashcodes on deserialization

### LINKED ISSUE #1089: NH-2755 - LockMode hash differs in x86 and 64bit OS
**Anton** created an [issue](https://nhibernate.jira.com/browse/NH-2755) — <time datetime="2011-06-01T00:34:54.000Z" title="2011-06-01T00:34:54.000Z">1st June 2011, 0:34:54</time>:
> When using LockMode over remoting between PCs with x86 and 64bit operating systems there is different result in GetHashCode.
> It leads to error "LockMode Upgrade not supported by NHibernate.Persister.Entity.SingleTableEntityPersister " when I pass LockMode.Upgrade from x86 to 64bit to do
> 
> session.Lock(this, lockMode)
> 
> It is caused by generating and storing hash code upon creating instance of LockMode.
> LockMode constructor uses a XOR (^) operand to generate and store hash code, and it has different behavior depends on OS type (x84, 64bit)
> 
---
**Anton** added a comment — <time datetime="2011-06-01T00:39:02.390Z" title="2011-06-01T00:39:02.390Z">1st June 2011, 0:39:02</time>:
> Here is callstack:
> 
> Server stack trace: 
>    в NHibernate.Persister.Entity.AbstractEntityPersister.GetLocker(LockMode lockMode) в D:\Projects\NHibernate.trunk\src\NHibernate\Persister\Entity\AbstractEntityPersister.cs:строка 1730 
>    в NHibernate.Persister.Entity.AbstractEntityPersister.Lock(Object id, Object version, Object obj, LockMode lockMode, ISessionImplementor session) в D:\Projects\NHibernate.trunk\src\NHibernate\Persister\Entity\AbstractEntityPersister.cs:строка 1734 
>    в NHibernate.Event.Default.AbstractLockUpgradeEventListener.UpgradeLock(Object entity, EntityEntry entry, LockMode requestedLockMode, ISessionImplementor source) в D:\Projects\NHibernate.trunk\src\NHibernate\Event\Default\AbstractLockUpgradeEventListener.cs:строка 71 
>    в NHibernate.Event.Default.DefaultLockEventListener.OnLock(LockEvent event) в D:\Projects\NHibernate.trunk\src\NHibernate\Event\Default\DefaultLockEventListener.cs:строка 55 
>    в NHibernate.Impl.SessionImpl.FireLock(LockEvent lockEvent) в D:\Projects\NHibernate.trunk\src\NHibernate\Impl\SessionImpl.cs:строка 2520 
>    в NHibernate.Impl.SessionImpl.Lock(Object obj, LockMode lockMode) в D:\Projects\NHibernate.trunk\src\NHibernate\Impl\SessionImpl.cs:строка 767 
> 
---
**Fabio Maulo** added a comment — <time datetime="2011-06-01T04:46:26.375Z" title="2011-06-01T04:46:26.375Z">1st June 2011, 4:46:26</time>:
> and the exception message ?
> 
---
**Anton** added a comment — <time datetime="2011-06-01T05:05:22.109Z" title="2011-06-01T05:05:22.109Z">1st June 2011, 5:05:22</time>:
> It is in issue description :)
> "LockMode Upgrade not supported by NHibernate.Persister.Entity.SingleTableEntityPersister"
> 
---
**Fabio Maulo** added a comment — <time datetime="2011-06-01T06:03:22.187Z" title="2011-06-01T06:03:22.187Z">1st June 2011, 6:03:22</time>:
> What you are doing is pass the NH's LockMode serializaed instance between 2 different machines, right ?
> 
---
**Anton** added a comment — <time datetime="2011-06-01T06:44:51.156Z" title="2011-06-01T06:44:51.156Z">1st June 2011, 6:44:51</time>:
> Yes, LockMode just passes from client to server PC, client is x86, server is 64bit.
> When using client 64bit and server 64bit - everything ok.
> 
> Stored hashcode in instance of LockMode differs at 86/64
> 
---
**Fabio Maulo** added a comment — <time datetime="2011-06-01T07:57:23.703Z" title="2011-06-01T07:57:23.703Z">1st June 2011, 7:57:23</time>:
> Well... that is not the only one place where NH store the hash code at ctor.
>

### LINKED ISSUE #1888: Second level cache key mismatch

fredericDelaporte:
Computed hashcode values may vary from one process to another. For being consistent inside a process, they must be re-computed after deserialization.

Fixes #1888, and also fixes #1089.

Possible breaking change: many classes, which were not serializing the session factory, do now serialize it. In case of cross-process serialization/deserialization, these session factories will need to be properly named, by setting `session_factory_name` setting in the configuration used to build them.


fredericDelaporte:
@gokhanabatay, may you check this change fixes your case?

src/NHibernate.Test/CacheTest/QueryKeyFixture.cs
```diff
@@ -100,6 +100,20 @@ public void NotEqualHashCodeWithFilters()
 			Assert.That(qk1.GetHashCode(), Is.Not.EqualTo(qvk1.GetHashCode()), "qk1 & qvk1");
 		}
 
+		[Test]
+		public void HashCodeWithFiltersAndSerialization()
```
fredericDelaporte: This new test does not demonstrate any failure in master. A multi-process test using .Net Core is likely required for this, but that will be a bit complex to setup.

fredericDelaporte: But after the change it fails... The hashcode computation is not stable itself. We have another bug there.

gokhanabatay: @fredericDelaporte  I think you figure it out, sorry for miss code but I am new to nhibernate and i want it to contribute too much :)

Did you fix it with deserialization call back?

gokhanabatay: @fredericDelaporte  I think FilterKey has own deserialization call back too.

fredericDelaporte: Not yet, but it is going to have it, since I do not see any other solution.

fredericDelaporte: No it is unrelated to the fields being readonly. Binary serialization has no issue with this.

My last fix was still not good. Maybe we will have better to cease caching hascodes.

fredericDelaporte:
We have two troubles here:

1. Hashcodes must not be serialized, as they are not guaranteed to be generated identical by different processes.
2. Hashcodes depending on iterating over dictionaries or sets cannot be reliably computed during deserialization, because dictionaries or sets may not be populated till the very end of the deserialization.

So a thing like `QueryKey` which holds a set of `FilterKey` triggers the `FilterKey` hashcode computation on deserializing. But we have not guarantees that the filter key inner dictionaries, on which its own hashcode computation depends, are ready.

We need to eliminate dictionaries and sets from the serialized state when they are used for hashcode computation, in order to ensure we can compute hashcodes of object without risking iterating over an not yet populated set or dictionary.

fredericDelaporte:
With the second fixup, tests now succeed (at least locally with netfx). But I do not consider the issue fixed yet. It has a reliability issue, which has to be addressed.

src/NHibernate.Test/CacheTest/QueryKeyFixture.cs
gokhanabatay: Also I am thinking the distributed cache scenario, "Multiple application single distributed cache" I did not test it but if it rely on HashCode, I think that they will also fail and cache the same result again right? Or just cache the query?

So maybe as you said we did not rely HashCode, just check Equals.

fredericDelaporte: No, I did not mean this. It is not an option in my opinion. Equal objects must have equal hashcodes, at least within the process in which they are compared. Any other solution is a workaround not fixing the amin issue. And moreover, since hashcode are essential for fast lookups, any other solution will imply a slow lookup logic.

fredericDelaporte:
That is the reliability issue I have written about. The fix will likely consist in serializing the `_filterParameters` as something else than a dictionary,
Implementing `IDeserializationCallback` will likely not do any good, because the `FilterKey` used by `QueryKey` are accessible only through a set. So in `QueryKey` own deserialization callback, we will not be able to iterate other filter keys for calling their deserialization callbacks soon enough: being in a set, their hashcode will already have been computed and used for storing them in the set, with a bad hashcode value.

src/NHibernate/Cache/CacheKey.cs
```diff
@@ -50,7 +56,22 @@ public override bool Equals(object obj)
 
 		public override int GetHashCode()
 		{
-			return hashCode;
+			// If the object is put in a set or dictionary during deserialization, the hashcode will not yet be
+			// computed. Compute the hashcode on the fly. So long as this happens only during deserialization, there
+			// will be no thread safety issues. For the hashcode to be always defined after deserialization, the
+			// deserialization callback is used.
+			return _hashCode ?? GenerateHashCode();
```
fredericDelaporte: I could have written it `_hashCode ?? (_hashCode = GenerateHashCode()).Value`, for lazily initializing it. But that is slightly more complicated code for a corner case, which should rarely happen: deserialization in which the object is put in a set or dictionary as part of the deserialization process. The constructor and the deserialization callback are the normal paths of initialization of this member.
They are mandatory, because solely relying on a lazy initialization could expose the code of `GenerateHashCode` to multi-threaded usage. It should work in most cases, but better reduce as much as possible the code having to support multi-threading.

src/NHibernate/Engine/AssociationKey.cs
```diff
@@ -40,7 +44,22 @@ public override bool Equals(object that)
...
+
+		/// <inheritdoc />
+		public void OnDeserialization(object sender)
+		{
+			_hashCode = GenerateHashCode();
+		}
+
+		private int GenerateHashCode()
+		{
+			return ownerKey.GetHashCode() ^ propertyName.GetHashCode() ^ ownerKey.EntityName.GetHashCode();
```
fredericDelaporte: Maybe for apparently simple cases like this one we would have better to always compute it on the fly instead of caching it. (`ownerKey.GetHashCode` computes a bit more than the two other ones which are just `string.GetHashCode()`.)

src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -211,14 +222,7 @@ protected HQLQueryPlanKey(System.Type queryTypeDiscriminator, string query, bool
 					filterNames = new HashSet<string>(enabledFilters.Keys);
 				}
 
-				unchecked
-				{
-					var hash = query.GetHashCode();
-					hash = 29 * hash + (shallow ? 1 : 0);
-					hash = 29 * hash + CollectionHelper.GetHashCode(filterNames, filterNames.Comparer);
```
fredericDelaporte: `filterNames.Comparer` is the default one here, the hashset is always constructed by this class, without a custom comparer. The new code does no more use it explicitly, as the hashset may not yet be available, if the hashcode computation has to be done during deserialization.

src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -286,11 +342,7 @@ public FilterQueryPlanKey(string query, string collectionRole, bool shallow, IDi
 					filterNames = new HashSet<string>(enabledFilters.Keys);
 				}
 
-				int hash = query.GetHashCode();
-				hash = 29 * hash + collectionRole.GetHashCode();
-				hash = 29 * hash + (shallow ? 1 : 0);
-				hash = 29 * hash + CollectionHelper.GetHashCode(filterNames, filterNames.Comparer);
```
fredericDelaporte: Same as previous comment.

src/NHibernate/LockMode.cs
```diff
@@ -11,11 +12,13 @@ namespace NHibernate
 	/// wish to explicitly specify lock levels.
 	/// </remarks>
 	[Serializable]
-	public sealed class LockMode
+	public sealed class LockMode : IDeserializationCallback
```
fredericDelaporte: I have missed that class in my first commit.

gokhanabatay: NHibernate Core Distributed Cache Library line 179 gets key with key.ToString() when i see this one I can say that we might have been used ToString() to compare filters. I think the order of these values are not important as you mentioned ? Am i wrong?

`private string GetCacheKey(object key)
		{
			var keyAsString = AppendHashcodeToKey
				? string.Concat(_cacheKeyPrefix, _fullRegion, ":", key.ToString(), "@", key.GetHashCode())
				: string.Concat(_cacheKeyPrefix, _fullRegion, ":", key.ToString());

			if (_maxKeySize < keyAsString.Length)
			{
				Log.Info(
					"Computing a hashed key for too long key '{0}'. This may cause collisions resulting into additional cache misses.",
					key);
				// Hash it for respecting max key size. Collisions will be avoided by storing the actual key along
				// the object and comparing it on retrieval.
				using (var hasher = new SHA256Managed())
				{
					var bytes = Encoding.UTF8.GetBytes(keyAsString);
					var computedHash = Convert.ToBase64String(hasher.ComputeHash(bytes));
					if (computedHash.Length <= _maxKeySize)
						return computedHash;

					if (!_hasWarnedOnHashLength)
					{
						// No lock for this field, some redundant logs will be less harm than locking.
						_hasWarnedOnHashLength = true;
						Log.Warn(
							"Hash computed for too long keys are themselves too long. They will be truncated, further " +
							"increasing the risk of collision resulting into additional cache misses. Consider using a " +
							"cache supporting longer keys. Hash length: {0}; max key size: {1}",
							computedHash.Length, _maxKeySize);
					}

					keyAsString = computedHash.Substring(0, _maxKeySize.Value);
				}
			}

			return _keySanitizer != null ? _keySanitizer(keyAsString) : keyAsString;
		}`

src/NHibernate/Cache/CacheKey.cs
gokhanabatay: I have some comments at QueryKey and FilterKey, but i test it and it works.
The cache problem will be fixed with this pr.

src/NHibernate/Cache/FilterKey.cs
```diff
@@ -53,7 +65,24 @@ public override bool Equals(object other)
 			}
 			if (!that._filterName.Equals(_filterName))
 				return false;
-			if (!CollectionHelper.DictionaryEquals<string, TypedValue>(that._filterParameters, _filterParameters))
+
+			// On deserialization, _filterParameters may be not yet populated. Use _serializedFilterParameters
+			// if it is defined.
+			var filterParameters = _filterParameters;
```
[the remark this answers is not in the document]
fredericDelaporte: Equals will be called during deserialization if two objects shares the same bucket (same hashcode) inside a dictionary or set.

src/NHibernate/Cache/FilterKey.cs
```diff
@@ -76,5 +105,21 @@ public static ISet<FilterKey> CreateFilterKeys(IDictionary<string, IFilter> enab
...
+		[OnSerializing]
+		private void OnSerializing(StreamingContext context)
+		{
+			_serializedFilterParameters = _filterParameters.ToArray();
+		}
+
+		/// <inheritdoc />
+		public void OnDeserialization(object sender)
+		{
+			_filterParameters = new Dictionary<string, TypedValue>(_serializedFilterParameters.Count());
```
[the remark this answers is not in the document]
fredericDelaporte: That works around the fact that deserialized dictionaries and sets cannot be used reliably in the deserialization callback. It relies on array implementation to not have that trouble, whatever the framework.

src/NHibernate/Cache/QueryKey.cs
```diff
@@ -127,12 +140,32 @@ public override bool Equals(object other)
 				}
 			}
 
-			if (!CollectionHelper.SetEquals(_filters, that._filters))
+			// On deserialization, _namedParameters and _filters may be not yet populated. Use their _serialization
+			// counterparts if they are defined.
+			if (_serializedFilters != null || that._serializedFilters != null)
```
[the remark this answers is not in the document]
fredericDelaporte: For this case to happen, you need to have in a deserialized dictionary keys or in a set two different objects having the same hashcode. Normally with a good hashcode algorithm, this happens almost never. So it is hard to test that case. Still it has to be supported.

fredericDelaporte:
So at least in the use case of @gokhanabatay, this PR solves the trouble of hashcode mismatch seen with his distributed cache usage.

That is still a change a bit heavier than what I would have expected. But it seems to me needed.

gokhanabatay:
> So at least in the use case of @gokhanabatay, this PR solves the trouble of hashcode mismatch seen with his distributed cache usage.
> 
> That is still a change a bit heavier than what I would have expected. But it seems to me needed.

I agree thank you @fredericDelaporte I will wait to merge at master branch.

src/NHibernate/Cache/FilterKey.cs
```diff
@@ -8,10 +10,16 @@
...
 	{
 		private readonly string _filterName;
-		private readonly Dictionary<string, TypedValue> _filterParameters = new Dictionary<string, TypedValue>();
+
+		// Sets and dictionaries are populated last during deserialization, causing them to be potentially empty
+		// during the deserialization callback. This causes them to be unreliable when used in hashcode computations.
+		// So better serialize them as other structures and use those other structures in hashcode computation.
+		[NonSerialized]
+		private Dictionary<string, TypedValue> _filterParameters = new Dictionary<string, TypedValue>();
+		private IEnumerable<KeyValuePair<string, TypedValue>> _serializedFilterParameters;
```
hazzik: What is the point declare it as IEnumerable?

fredericDelaporte: This avoids a cast in `GetHashCode` and `Equals` implementations, when doing things like `var filterParameters = _serializedFilterParameters ?? _filterParameters;`.

gokhanabatay:
@hazzik 
@fredericDelaporte 
I post the issue to dotnet team please look and watch further action;

Deserialize binary[] to ISet<T> wrong Internal HashCode has been stored #2027
[https://github.com/dotnet/core/issues/2027](url)

fredericDelaporte:
The PR here precisely dodge the issue you have reported in dotnet core repository. That is the reason for having serialized dictionaries and sets content with arrays in this PR.

The fact that hashcodes are not computed with same values across processes is also the reason for recomputing them on deserialization instead of using the previously stored value in serialized data.

About the use of hashcode in keys, the issue is already known and handled, see nhibernate/NHibernate-Caches#53. This is not yet released.

But a static configuration property is already available: `CoreDistributedCacheProvider.AppendHashcodeToKey`. Set it to `false`.

src/NHibernate/Cache/FilterKey.cs
```diff
@@ -8,10 +10,16 @@
...
+	public class FilterKey : IDeserializationCallback
 	{
 		private readonly string _filterName;
-		private readonly Dictionary<string, TypedValue> _filterParameters = new Dictionary<string, TypedValue>();
+
+		// Sets and dictionaries are populated last during deserialization, causing them to be potentially empty
+		// during the deserialization callback. This causes them to be unreliable when used in hashcode computations.
+		// So better serialize them as other structures and use those other structures in hashcode computation.
+		[NonSerialized]
+		private Dictionary<string, TypedValue> _filterParameters = new Dictionary<string, TypedValue>();
```
hazzik: Since `_filterParameters` field is not used outside of the class it would be easier to _just_ store it as a serializable collection (eg array).

src/NHibernate/Cache/QueryKey.cs
```diff
@@ -11,23 +13,33 @@
...
 		private readonly object[] _values;
 		private readonly int _firstRow = RowSelection.NoValue;
 		private readonly int _maxRows = RowSelection.NoValue;
+
+		// Sets and dictionaries are populated last during deserialization, causing them to be potentially empty
+		// during the deserialization callback. This causes them to be unreliable when used in hashcode computations.
+		// So better also serialize them as other structures and use those other structures in hashcode computation.
+		// We nonetheless need to also serialize them, because they are externally supplied and may use a custom
+		// comparer, implementation, ...
 		private readonly IDictionary<string, TypedValue> _namedParameters;
```
hazzik: Since `_namedParameters` field is not used outside of the class it would be easier to just change its type to a serializable collection.

src/NHibernate/Cache/QueryKey.cs
```diff
@@ -11,23 +13,33 @@
...
 		private readonly ISet<FilterKey> _filters;
+		private IEnumerable<KeyValuePair<string, TypedValue>> _serializedNamedParameters;
+		private IEnumerable<FilterKey> _serializedFilters;
```
hazzik: After deserialization callback is dropped from `FilterKey` this would not be needed.

src/NHibernate/Engine/AssociationKey.cs
hazzik: Please always compute this.

src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -178,12 +180,21 @@ private ParameterMetadata BuildNativeSQLParameterMetadata(string sqlString)
...
 			private readonly string query;
 			private readonly bool shallow;
-			private readonly HashSet<string> filterNames;
-			private readonly int hashCode;
+
+			// Sets and dictionaries are populated last during deserialization, causing them to be potentially empty
+			// during the deserialization callback. This causes them to be unreliable when used in hashcode computations.
+			// So better serialize them as other structures and use those other structures in hashcode computation.
+			[NonSerialized]
+			private HashSet<string> filterNames;
```
hazzik: Same as FilterKey and others.

src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -258,18 +270,62 @@ public bool Equals(HQLQueryPlanKey that)
...
 			private readonly string collectionRole;
 			private readonly bool shallow;
-			private readonly HashSet<string> filterNames;
-			private readonly int hashCode;
+
+			// Sets and dictionaries are populated last during deserialization, causing them to be potentially empty
+			// during the deserialization callback. This causes them to be unreliable when used in hashcode computations.
+			// So better serialize them as other structures and use those other structures in hashcode computation.
+			[NonSerialized]
+			private HashSet<string> filterNames;
```
hazzik: Same

src/NHibernate/LockMode.cs
```diff
@@ -126,7 +126,25 @@ public bool Equals(LockMode other)
 
 		public override int GetHashCode()
 		{
-			return hashcode;
+			// If the object is put in a set or dictionary during deserialization, the hashcode will not yet be
+			// computed. Compute the hashcode on the fly. So long as this happens only during deserialization, there
+			// will be no thread safety issues. For the hashcode to be always defined after deserialization, the
+			// deserialization callback is used.
+			return _hashCode ?? GenerateHashCode();
```
hazzik: Just need to always recompute.

src/NHibernate/Cache/FilterKey.cs
fredericDelaporte: This will have a performance impact on `Equals`, switching an O(n) algorithm for an an O(n²) one. That said, `n` should be always quite low in most cases, so it may not be noticeable at all.

src/NHibernate/Cache/QueryKey.cs
fredericDelaporte: This assumes the code here do not have to care about the possibility this externally supplied dictionary may use a custom comparer. Currently that does not happen and all these are quite internal things, so well, why not.

src/NHibernate/Engine/EntityKey.cs
```diff
@@ -88,24 +106,9 @@ private int GenerateHashCode()
...
 		public override string ToString()
 		{
 			return "EntityKey" + MessageHelper.InfoString(factory.GetEntityPersister(EntityName), Identifier, factory);
 		}
-
-		/// <summary>
-		/// To use in deserialization callback
-		/// </summary>
-		/// <param name="sessionFactory"></param>
-		internal void SetSessionFactory(ISessionFactoryImplementor sessionFactory)
```
fredericDelaporte: This was dead code.

fredericDelaporte:
Changes done.

fredericDelaporte:
The filed type change impacts `ToString` of the keys. It is now fixed.

src/NHibernate/Impl/Printer.cs
```diff
@@ -72,11 +71,16 @@ public string ToString(IType[] types, object[] values)
 
 		public string ToString(IDictionary<string, TypedValue> namedTypedValues)
 		{
-			IDictionary<string, string> result = new Dictionary<string, string>(namedTypedValues.Count);
+			return ToString(namedTypedValues.AsEnumerable());
+		}
+
+		internal string ToString(IEnumerable<KeyValuePair<string, TypedValue>> namedTypedValues)
+		{
+			IDictionary<string, string> result = new Dictionary<string, string>(namedTypedValues.Count());
```
hazzik: Not happy with this line at all.

fredericDelaporte: It does the same than previous code though.

fredericDelaporte: I have added a fixup with another implementation, leaving the old method (still used at other places) untouched.

fredericDelaporte:
I have qualified the issue fixed by this PR, #1888, as a major bug, since this causes distributed caches to not be actually shared among processes using them. There is no point in using a distributed cache in such case.

Is there anything which still prevents validating this PR?

hazzik:
> Is there anything which still prevents validating this PR?

I have not gone through all the code yet.

PR #1898, opened 2018-11-05
Specify the precise version of dependencies to avoid compiler warning N…

jaykay-design:
Explanation of why a package author should specify an existing lower bound version of a dependency: https://github.com/NuGet/Home/issues/5764

Currently a non existing version is indicated:
`      <dependency id="Iesi.Collections" version="[4,5)" />`

but it should be
`      <dependency id="Iesi.Collections" version="[4.0.0.4000,5)" />`


hazzik:
It’s for 4.1. Are we going to release 4.1?

fredericDelaporte:
Hum, no, I do not think so. And 5.2 already targets an existing version.

PR #1900, opened 2018-11-08
Do not generate FK on non-generated unique constraint

fredericDelaporte:
When an unique key is not generated due to the database not supporting a `null` value in an unique key, foreign keys on the unique key must not be generated either.

This is an old bug of the feature consisting in not generating unique constraints for nullable columns when the target database does not support null in unique. But as no dialects were enabling this feature, tests were not showcasing it.

#1854 enables a dialect with this feature, and its build fails due to FK being generated for non-generated unique constraints, see [here](https://teamcity.jetbrains.com/viewLog.html?buildId=1749931&buildTypeId=NHibernate_NHibernateSapSqlAnywhere).


fredericDelaporte:
This PR is also a kind of follow-up to #1855, as the trouble here can happen only when using the feature fixed for another trouble there. It would be better to have that feature (I hope) fully fixed in 5.2.

PR #1904, opened 2018-11-14
Lazy initialize BatchingEntityLoader and BatchingCollectionInitializer

hazzik:
Replaces #400


hazzik:
Agree. Better to optimize the code in batchers instead.

PR #1905, opened 2018-11-14
Improve support of Npgsql 4

hazzik:
We have some compatibility issues with the newer driver, so need to upgrade


hazzik:
Most likely changes required are similar to https://github.com/nhibernate/nhibernate-envers/pull/2

fredericDelaporte:
So, Npgsql 4 still has issues with connections released from the second phase.
As explained [here](https://github.com/nhibernate/nhibernate-core/pull/1401#issuecomment-339322323), the TeamCity build agent may require some manual actions for allowing PostgreSQL other builds to succeed.

> The 3.2.5 failure may lead to pending prepared transaction on PostgreSQL db, causing it to prevent nhibernate database drop, wrecking all subsequent builds whatever the PR. It is then required to use the pgadmin tool on build agent for rollbacking pending prepared transactions (see [here](https://stackoverflow.com/q/25077581/1178314)) and dropping the db.

As suggested [here](https://github.com/nhibernate/nhibernate-core/pull/1401#issue-148371413), this PR may have to introduce a new (TestDialect?) property for ignoring tests involving connections used in commit phases. Then something like the following could be added into `SystemTransactionFixtureBase`:
```c#
	=> (TestDialect.SupportsConnectionUseOnSystemTransactionPrepare ||
			!UseConnectionOnSystemTransactionPrepare) &&
		base.AppliesTo(dialect);
```

src/NHibernate/Dialect/PostgreSQLDialect.cs
```diff
@@ -329,6 +329,16 @@ public override string CurrentTimestampSelectString
...
 
+		/// <summary>
+		/// Does this dialect supports distributed transaction? <c>false</c>.
+		/// </summary>
+		/// <remarks>
+		/// Npgsql since its version 3.2.5 version has race conditions: it fails handling the threading involved with
+		/// distributed transactions. This causes a bunch of distributed tests to be flaky with Npgsql. Individually,
+		/// they usually succeed, but run together, some of them fail. The trouble was not occuring with Npgsql 3.2.4.1.
+		/// </remarks>
+		public override bool SupportsDistributedTransactions => false;
```
hazzik: I don't quite agree with this change. It does support distributed transactions, but has some bugs. Also, not related, but, it seems that this property should be on a driver, not on a dialect.

fredericDelaporte: Sometimes that is more driver related, other times that just the db, whatever the driver, so then it is more dialect related.

Yes, this property is informational, it currently does not have any effect elsewhere than tests (or user code if some users do something with it).

From my viewpoint, Npgsql does no more properly support distributed transactions, because it can not even reliably close the connection right after disposing a distributed scope since Npgsql 3.2.5. Sometimes, a race condition will cause the connection to be still executing the second phase, and the close will blow up with an exception.

hazzik: Ok.

hazzik:
I think for the most of the providers this will be no-op as types do typically match to the expected type. Also, most of the types already do some conversation on set, so it is not a big deal.

PR #1910, opened 2018-11-25
Add an overridable ADO BeginTransaction method

### LINKED ISSUE #1908: Control over BeginTransaction in AdoTransaction
It should be possible to control how the DbTransaction is created for generic Database Drivers.

Firebird for example implements its own `DbConnection`: `FbConnection`.
In this class, there is a Firebird specific overload for the `BeginTransaction` Method that lets you use specific Firebird `FbTransactionOptions`.

Of course, there is no way for NHibernate to consider such generic implementations.

I have a suggestion for giving the User an optional control over this matter:

Inside [AdoTransaction](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Transaction/AdoTransaction.cs#L149) on the Method `Begin(IsolationLevel isolationLevel)` there is the following code where the `DbTransaction` is created:

```C#
if (isolationLevel == IsolationLevel.Unspecified)
{
  trans = session.Connection.BeginTransaction();
}
else
{
  trans = session.Connection.BeginTransaction(isolationLevel);
}
```

How about moving that code to a protected virtual Method like this:

```C#
protected virtual DbTransaction BeginTransaction(IsolationLevel isolationLevel, ISessionImplementor sessionImplementor)
{
    if (isolationLevel == IsolationLevel.Unspecified)
    {
        return sessionImplementor.Connection.BeginTransaction();
    }
    else
    {
        return sessionImplementor.Connection.BeginTransaction(isolationLevel);
    }
}
```

And changing the mentioned code above using that Method:

```C#
trans = BeginTransaction(isolationLevel, session);
```

This way, i can derive from `AdoTransaction` and override my own specific implementation.
For example with the specific Firebird `Wait` Option:

```C#
public class FirebirdAdoTransaction : AdoTransaction
{
    public FirebirdAdoTransaction(ISessionImplementor session) : base(session)
    {
    }

    protected override DbTransaction BeginTransaction(IsolationLevel isolationLevel, ISessionImplementor sessionImplementor)
    {
        if (sessionImplementor.Connection is FbConnection firebird)
        {
            var transactionOptions = new FbTransactionOptions
            {
                TransactionBehavior = FbTransactionBehavior.Wait
            };

            if (isolationLevel == IsolationLevel.ReadCommitted)
            {
                transactionOptions.TransactionBehavior |= FbTransactionBehavior.ReadCommitted;
            }

            return firebird.BeginTransaction(transactionOptions);
        }
        else
        {
            return base.BeginTransaction(isolationLevel, sessionImplementor);
        }
    }
}
```

Then, i only have to create my own `TransactionFactory`, using the `FirebirdAdoTransaction` and change my NHibernate Settings `"transaction.factory_class"` to this Factory:

```C#
public class FirebirdTransactionFactory : AdoNetWithSystemTransactionFactory
{
    public override ITransaction CreateTransaction(ISessionImplementor session)
    {
        return new FirebirdAdoTransaction(session);
    }
}
```

I think this is a good addition in general, since you never know what generic implementation of `DbConnection` is used by the Database driver and this `protected virtual` method gives the user full control over driver related transaction options.

Also: No breaking change.

EDIT:
Changed the name of the `protected virtual` method to `BeginTransaction` like suggested by **hazzik**

fredericDelaporte:
Fixes #1908 

I have done something else than @greenlight2k suggestion, because looking into it again, it appeared to me `BeginTransaction` should belong to the driver.

### DIFF
--- a/src/NHibernate/Driver/DriverExtensions.cs
+++ b/src/NHibernate/Driver/DriverExtensions.cs
@@ -36,7 +36,7 @@ namespace NHibernate.Driver
 			var beginMethod = driver.GetType().GetMethod(
 				nameof(DriverBase.BeginTransaction),
 				new[] { typeof(IsolationLevel), typeof(DbConnection) });
-			if (beginMethod != null)
+			if (beginMethod != null && beginMethod.ReturnType == typeof(DbTransaction) && !beginMethod.IsStatic)
 			{
 				return (DbTransaction) beginMethod.Invoke(driver, new object[] { isolationLevel, connection });
 			}
--- a/src/NHibernate/Driver/DriverExtensions.cs
+++ b/src/NHibernate/Driver/DriverExtensions.cs
@@ -7,10 +7,6 @@ namespace NHibernate.Driver
 {
 	public static class DriverExtensions
 	{
-		private static readonly INHibernateLogger Log = NHibernateLogger.For(typeof(DriverExtensions));
-		// No threading protection on this member: we do not care, at worst it will cause some duplicated warning.
-		private static bool _hasWarnedForMissingBeginTransaction;
-
 		internal static void AdjustParameterForValue(this IDriver driver, DbParameter parameter, SqlType sqlType, object value)
 		{
 			var adjustingDriver = driver as IParameterAdjuster;
@@ -32,27 +28,8 @@ namespace NHibernate.Driver
 				return driverBase.BeginTransaction(isolationLevel, connection);
 			}
 
-			// Use reflection for supporting custom drivers.
-			var beginMethod = driver.GetType().GetMethod(
-				nameof(DriverBase.BeginTransaction),
-				new[] { typeof(IsolationLevel), typeof(DbConnection) });
-			if (beginMethod != null && beginMethod.ReturnType == typeof(DbTransaction) && !beginMethod.IsStatic)
-			{
-				return (DbTransaction) beginMethod.Invoke(driver, new object[] { isolationLevel, connection });
-			}
-
-			if (!_hasWarnedForMissingBeginTransaction)
-			{
-				// Ideally it should be a dictionary per driver class, but keep it simple. Drivers not deriving from
-				// DriverBase are already not common, so applications which would use two distinct such drivers are
-				// even less common.
-				_hasWarnedForMissingBeginTransaction = true;
-				Log.Warn(
-					"Driver {0} is obsolete. It should implement a 'public DbTransaction BeginTransaction(" +
-					"IsolationLevel, DbConnection) method.",
-					driver);
-			}
-
+			// So long for custom drivers not deriving from DriverBase, they will have to wait for 6.0 if they
+			// need the feature.
 			if (isolationLevel == IsolationLevel.Unspecified)
 			{
 				return connection.BeginTransaction();
--- a/src/NHibernate/Driver/DriverExtensions.cs
+++ b/src/NHibernate/Driver/DriverExtensions.cs
@@ -1,16 +1,63 @@
+using System.Data;
 using System.Data.Common;
 using NHibernate.AdoNet;
 using NHibernate.SqlTypes;
-using NHibernate.Type;
 
 namespace NHibernate.Driver
 {
-	internal static class DriverExtensions
+	public static class DriverExtensions
 	{
+		private static readonly INHibernateLogger Log = NHibernateLogger.For(typeof(DriverExtensions));
+		// No threading protection on this member: we do not care, at worst it will cause some duplicated warning.
+		private static bool _hasWarnedForMissingBeginTransaction;
+
 		internal static void AdjustParameterForValue(this IDriver driver, DbParameter parameter, SqlType sqlType, object value)
 		{
 			var adjustingDriver = driver as IParameterAdjuster;
 			adjustingDriver?.AdjustParameterForValue(parameter, sqlType, value);
+		}
+
+		// 6.0 TODO: merge into IDriver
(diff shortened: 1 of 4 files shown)


src/NHibernate/Driver/DriverExtensions.cs
```diff
@@ -1,16 +1,63 @@
...
+			{
+				return (DbTransaction) beginMethod.Invoke(driver, new object[] { isolationLevel, connection });
+			}
+
+			if (!_hasWarnedForMissingBeginTransaction)
+			{
+				// Ideally it should be a dictionary per driver class, but keep it simple. Drivers not deriving from
+				// DriverBase are already not common, so applications which would use two distinct such drivers are
+				// even less common.
+				_hasWarnedForMissingBeginTransaction = true;
```
fredericDelaporte: Maybe we should not even care about warning, and just silently fallback to the inline implementation.

greenlight2k:
Since the Driver is also creating the DbConnection, this solution seems appropriate to me.

The reflection stuff however seems not very intentional: How should any custom implementation of `IDriver` know that it can implement that BeginTransaction method with that specific syntax?
And that it will be invoked per reflection by this Extension? This is basically a black box.

I understand that you can't just put it into the IDriver interface because of breaking change,
but just trying to check if the custom implementation has that specific method, invoking it and force cast the return value to DbTransaction seems pretty dangerous to me. You dont even know if that method has any return type at all.

I would leave that reflection call out and just use the lines after that with the basic transaction begin to the DbConnection.

greenlight2k:
But if you want to leave it in, at least check that methods return type for DbTransaction.

fredericDelaporte:
> How should any custom implementation of `IDriver` know that it can implement that BeginTransaction method with that specific syntax?
> And that it will be invoked per reflection by this Extension?

That is the reason for the log warning when the method is not found. Yes it may remain unseen, since an application can use NHibernate without enabling NHibernate logging.

> You dont even know if that method has any return type at all.

This could be checked.

> I would leave that reflection call out

Then, there will be no incentive to prepare a custom `IDriver` for the breaking change which would come in 6.0. And no way to benefit of this `BeginTransaction` pattern with a custom `IDriver` until 6.0.

There is another way we handle this: using an additional "transitional" new interface instead of that reflection stuff. We then emit the warning for classes not implementing that new interface. With transitional interfaces, there are still no better cue than the log warning for custom implementations.

But I find that adding transitional interfaces for such changes bloats the NHibernate API with temporary interfaces that ideally will have to be obsoleted only in 6.0, since they cannot be obsoleted sooner. And so they would be dropped only in 7.0. We may also decide to drop them directly in 6.0, but that is not a very nice practice as this would be a compilation breaking change without any way to get ready for it before 6.0 release.
So I have started using this reflection based solution instead, when we have to dodge a breaking change by changing an interface.

greenlight2k:
Ah, I see. Makes sense now.

src/NHibernate/Driver/DriverExtensions.cs
```diff
@@ -1,16 +1,63 @@
...
+		/// <returns>The started <see cref="DbTransaction" />.</returns>
+		public static DbTransaction BeginTransaction(this IDriver driver, IsolationLevel isolationLevel, DbConnection connection)
+		{
+			if (driver is DriverBase driverBase)
+			{
+				return driverBase.BeginTransaction(isolationLevel, connection);
+			}
+
+			// Use reflection for supporting custom drivers.
+			var beginMethod = driver.GetType().GetMethod(
```
hazzik: The method is usually heavily used, so this needs to be cached at least into a compiled delegate (`MethodBase.Invoke` is bloody [slow](http://www.thejoyofcode.com/Performance_of_Method.Invoke_vs_a_Delegate.aspx)). However, I would encourage just remove the reflection from here.

hazzik: I do understand that this is "not common" to have a `IDriver` which does not extend `DriverBase`, but if someone does have this situation - they will suffer a lot.

src/NHibernate/Driver/DriverExtensions.cs
```diff
@@ -36,7 +36,7 @@ public static DbTransaction BeginTransaction(this IDriver driver, IsolationLevel
 			var beginMethod = driver.GetType().GetMethod(
 				nameof(DriverBase.BeginTransaction),
 				new[] { typeof(IsolationLevel), typeof(DbConnection) });
-			if (beginMethod != null)
+			if (beginMethod != null && beginMethod.ReturnType == typeof(DbTransaction) && !beginMethod.IsStatic)
```
hazzik: `beginMethod.IsStatic` should not be here. `GetMethod` has an overload to specify the flags of the method you want to find. (Hint: `BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic`).

src/NHibernate/Driver/DriverExtensions.cs
fredericDelaporte: A cached compiled delegate would have to be done per driver type, something with a concurrent dictionary, and a GetOrAdd which would call the method if the driver has it, or call the inline implementation. At least it would handle more properly the warning. But well, it seems not worth it to do all this just for being slightly nicer by giving a hint for custom implementers, if they check NHibernate log, about getting ready for 6.0.

src/NHibernate/Driver/DriverExtensions.cs
fredericDelaporte: I know, there are just a lot more parameters to these overloads, so I have been lazy.

PR #1916, opened 2018-11-29
Fix regressions introduced by Conditional and Coalesce expansion (#1880)

PleasantD:
We noticed a couple of regressions caused by #1880 

The coalesce and conditional nomination was a little over-zealous and some expressions were getting rewritten incorrectly.
Specifically when there were object casts in place.


fredericDelaporte:
As #1880 is not yet released, I am labelling this as a fix (of #1880) rather than as a bug of its own. (In other words, it will not be put in release notes.)

PR #1923, opened 2018-12-06
Obsolete StringHelper.Replace

hazzik:
`StringHelper.Replace*` methods are highly ineffective when the input string has many occurrences of a placeholder.

### DIFF
--- a/src/NHibernate/Impl/AbstractQueryImpl.cs
+++ b/src/NHibernate/Impl/AbstractQueryImpl.cs
@@ -10,6 +10,7 @@ using NHibernate.Transform;
 using NHibernate.Type;
 using NHibernate.Util;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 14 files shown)


src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -265,7 +266,10 @@ private string ExpandParameterList(string query, string name, TypedValue typedLi
 
 			var paramPrefix = isJpaPositionalParam ? StringHelper.SqlParameter : ParserHelper.HqlVariablePrefix;
 
-			return StringHelper.Replace(query, paramPrefix + name, string.Join(StringHelper.CommaSpace, aliases), true);
+			return Regex.Replace(
+				query,
+				$@"{paramPrefix}{name}\b",
```
bahusoid: Shouldn't this place and others be wrapped in  `Regex.Escape` like `Regex.Escape(paramPrefix+name)`?

hazzik: This is unlikely that parameter name will contain any special regex characters.

src/NHibernate/Util/StringHelper.cs
```diff
@@ -83,11 +83,15 @@ public static string Repeat(string str, int times)
 			return buf.ToString();
 		}
 
+		//TODO: Inline
 		public static string Replace(string template, string placeholder, string replacement)
```
fredericDelaporte: Why not obsoleting and inlining it right now?

hazzik: Did not want to change much of the code.

fredericDelaporte: Well, if not now, why later? Better even remove the TODO then.

hazzik: TODO removed

PR #1926, opened 2018-12-06
Fix DML insert with filter

### LINKED ISSUE #1921: DML insert fails when a filter is enabled
When a HQL `Insert` is attempted from a selected entity having a filter enabled, an invalid query is generated.

Consider following pseudo HQL code: `insert into Target (...) select ... from Source`. If `Source` has a filter enabled, the filter fragment is appended to the query, but its parameters are neither processed nor have their values supplied.

`update` and `delete` do not fail with filters: they completely ignore them.

Initial report:
---

Insert select dml style statement with filter data is not bind to sql parameters ORA-01008 : Not all variables bound

collectionRole is passed null
image

So is filter enabled parameter setted to false by doing collectionRole != null, but enabled filter is not empty its filled with my filter.
image

It is added to where criteria but parameters is empty, and it's not added to parameter collection.
image

```c#
 var query = ISession.CreateQuery(
                $@"INSERT INTO {typeof(Session)}(MbrId, UserCode, OpenDate, IsClosed, ExpireDateTime,
                RemoteIpAddress, RemotePort, LocalIpAddress, LocalPort, Claims)
                SELECT E.MbrId, E.UserCode, E.OpenDate, E.IsClosed, E.ExpireDateTime,
                E.RemoteIpAddress, E.RemotePort, E.LocalIpAddress, E.LocalPort, E.Claims from {typeof(Session)} AS E");

            processedCount = query.ExecuteUpdate();
```

fredericDelaporte:
Fixes #1921


src/NHibernate.Test/NHSpecificTest/GH1921/Fixture.cs
```diff
@@ -0,0 +1,85 @@
...
+			using (var transaction = session.BeginTransaction())
+			{
+				if (filtered)
+					session.EnableFilter("NameFilter").SetParameter("name", "Bob");
+				var rowCount = session.CreateQuery("update Entity e set Name = 'newName'").ExecuteUpdate();
+				transaction.Commit();
+
+				// If the DML has to take the filter into account, then below Assert should be
+				// Assert.That(rowCount, Is.EqualTo(filtered ? 1 : 2));
+				Assert.That(rowCount, Is.EqualTo(2));
```
fredericDelaporte: Currently, update and delete do not care at all about filters enabled in the session. They are ignored.
So likely the fix for the insert case should be to ignore them completely too.

Otherwise, if the insert case is fixed by actually supporting filters, then the update and delete would need to support it too. And it would be a breaking change for those already using an update or delete with a filter enabled: their update or delete would change by taking into account the filter.

src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
```diff
@@ -248,6 +248,14 @@ private IASTNode GenerateVersionPropertyNode(IQueryable persister)
 			return versionPropertyNode;
 		}
 
+		partial void EnterRule_insertStatement()
+		{
+			// Filters are not supported by DML, and they wreck the select part of an insert.
+			// Unfortunately I do not see any better place to handle this. This hack requires
+			// EnabledFilters to not be the same instance than the one held by the session.
+			_qti.EnabledFilters.Clear();
```
fredericDelaporte: This is hackish. Any better idea is welcome.

The trouble is, the query translator has no idea it is a DML query before having analysed it. But the analyse does already incorporate filters in the AST if any is enabled. So ideally we need to ignore filters before analysing the query if it is a DML, but we cannot know it is a DML before having analysed it.

In the analyse, the very first thing happening is finding the insert statement. So the hack is to clear the enabled filters at this point, in order to ignore them.

(We may instead set a new flag on the query translator for telling it very early it is a DML, and then the query translator could yield an empty `EnabledFilters` if it is a DML. It would avoid the dictionary copy, but overall the fix would still be a hack.)

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -55,7 +55,9 @@ public partial class QueryTranslatorImpl : IFilterTranslator
 			_stageOneAst = parsedQuery;
 			_compiled = false;
 			_shallowQuery = false;
-			_enabledFilters = enabledFilters;
+			// Copy the filters, because a hack into EnterRule_insertStatement may clear them. It
+			// should be done only for the current query translator, not for the whole session.
+			_enabledFilters = new Dictionary<string, IFilter>(enabledFilters);
```
fredericDelaporte: The hack moreover requires a dictionary duplication, in order to avoid wrecking the session enabled filters.

src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
gokhanabatay: If we are thinking about backward compability we could add a default parameter to ExecuteUpdate?
ExecuteUpdate(bool enableFilters = false)

fredericDelaporte: The reason for this is explained [here](https://github.com/nhibernate/nhibernate-core/issues/1921#issuecomment-444941560).

fredericDelaporte:
Replaced by #1931.

PR #1935, opened 2018-12-10
Fix generation of version update statement for joined subclasses

### LINKED ISSUE #1269: NH-3069 - Cannot use Session.Lock with Version column on abstract base class
**Richard Lowden** created an [issue](https://nhibernate.jira.com/browse/NH-3069) — <time datetime="2012-02-21T13:53:02.939Z" title="2012-02-21T13:53:02.939Z">21st February 2012, 13:53:02</time>:
> When a version column is on an abstract base class and a joined subclass is used, when trying to use Session.Lock(entity, LockMode.Force) where entity is an instance of the subclass the incorrect SQL is generated.
> 
> The statement to try and increment the version number creates an update statement with the table name for the base class as expected, but the id column name as the column name from the sub-class table.
> 
> I believe the issue is in:
> - NHibernate.Persister.Entity.AbstractEntityPersister#GenerateVersionIncrementUpdateString()
> 
> 
> 
---
**Richard Lowden** added a comment — <time datetime="2012-02-21T13:55:44.461Z" title="2012-02-21T13:55:44.461Z">21st February 2012, 13:55:44</time>:
> Adding test case
> 
---
**Adam Rivera** added a comment — <time datetime="2012-05-25T18:05:00.583Z" title="2012-05-25T18:05:00.583Z">25th May 2012, 18:05:00</time>:
> Same issue here. :-(
> 
---

* [NH-3069.zip](https://github.com/nhibernate/nhibernate-core/files/2664709/NH-3069.zip)

hazzik:
Tests contributed by @lowds

Fixes #1269


hazzik:
Thanks

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -1722,13 +1722,13 @@ public object ForceVersionIncrement(object id, object currentVersion, ISessionIm
 		private SqlCommandInfo GenerateVersionIncrementUpdateString()
 		{
 			SqlUpdateBuilder update = new SqlUpdateBuilder(Factory.Dialect, Factory);
-			update.SetTableName(GetTableName(0));
+			update.SetTableName(RootTableName);
```
fredericDelaporte: There is a property, `VersionTableName`, which looks dedicated to this and is used in `GenerateSelectVersionString`. (This other method directly uses the `rootTableKeyColumnNames` field instead of `RootTableIdentifierColumnNames` which delegates to `RootTableKeyColumnNames`. All this seems a bit messy.)

hazzik: I've decided to obsolete `VersionedTableName` instead of using it.

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -1722,13 +1723,13 @@ public object ForceVersionIncrement(object id, object currentVersion, ISessionIm
...
 			if (Factory.Settings.IsCommentsEnabled)
 			{
 				update.SetComment("forced version increment");
 			}
 			update.AddColumn(VersionColumnName, VersionType);
-			update.SetIdentityColumn(IdentifierColumnNames, IdentifierType);
+			update.SetIdentityColumn(RootTableIdentifierColumnNames, IdentifierType);
```
fredericDelaporte: Shouldn't it be changed to `RootTableKeyColumnNames` for consistency with `GenerateSelectVersionString`? (Or eventually the other way round, but I do not see which added value has `RootTableIdentifierColumnNames`. It is overridable contrary to `RootTableKeyColumnNames`, but never overridden in NHibernate code base.)

hazzik: `RootTableIdentifierColumnNames`

hazzik: There are also `SelectLockingStrategy` and `UpdateLockingStrategy`. I'm making everything consistent to `UpdateLockingStrategy`.

hazzik:
In case we're going to release any more patch version of 5.2 branch, this could be included as well.

PR #1936, opened 2018-12-11
Upgrade AsyncGenerator to 0.13.1

fredericDelaporte:
NHibernate currently used a 0.8 version having received many backports of fixes from newer versions.

I think we should upgrade it to latest version instead of going on relying on backports in case of new needs / new unsupported patterns used in NHibernate.

This update also refines the cases where generated async code is guarded by `try catch` clauses. (Some are removed, other are added.)

By the way, I let the .DotSettings file get updated by Rider here: it does no more remove the property we were not willing to let it remove.


src/NHibernate/Async/Loader/Loader.cs
```diff
@@ -334,7 +334,7 @@ private async Task<IList> DoQueryAsync(ISessionImplementor session, QueryParamet
 
 		internal async Task InitializeEntitiesAndCollectionsAsync(
 			IList hydratedObjects, DbDataReader reader, ISessionImplementor session, bool readOnly,
-			CacheBatcher cacheBatcher = null, CancellationToken cancellationToken = default(CancellationToken))
+			CacheBatcher cacheBatcher = null, CancellationToken cancellationToken)
```
fredericDelaporte: @maca88, some invalid code is generated here by the latest version of the AsyncGenerator, 0.13.0, by added a non-optional parameter after an optional one. Was the fix in 0.8.2.9 not done in latest versions of AsyncGenerator? (maca88/AsyncGenerator#115)

maca88: Is was done but not released. Now I've released it in version 0.13.1.

fredericDelaporte: Thanks, upgrade chnaged to 0.13.1 and commit forced-pushed.

fredericDelaporte:
@maca88, @bahusoid, having merged this PR may cause some async generation changes in pending PR. So better updates them and check an async generation.

PR #1939, opened 2018-12-18
Support Future for collection filters

### LINKED ISSUE #722: NH-1953 - Support Future for collection filters
**Marko Podgorsek** created an [issue](http://nhibernate.jira.com/browse/NH-1953) — <time datetime="2009-09-02T06:57:10.844Z" title="2009-09-02T06:57:10.844Z">2nd September 2009, 6:57:10</time>:
> Currently, `Future<>` doesn't work on `CreateFilter`
>
> I have a class:
> 
> ```csharp
> public class User
> {
>       public virtual IList<Tag> Tags { get; protected set; }
> }
> ```
> 
> 
> public class Tag .... //not important
> 
> Mapping for User -- the Tags part:
> 
> ```xml
> <bag lazy="true" name="Tags" table="USER_TAG">
>       <key>
>         <column name="ID_USER" />
>       </key>
>       <many-to-many class="BeriMe.Biz.Domain.Tag, BeriMe.Biz, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null">
>         <column name="ID_TAG" />
>       </many-to-many>
> </bag>
> ```
> 
> 
> If I do this:
> 
> ```csharp
> User u = db.Get<User>(5);
> 
> var t = db.CreateFilter(u.Tags, "")
>                 .SetFirstResult(4)
>                 .SetMaxResults(3)
>                 .Future<Tag>();    //If I say List<Tag>() it works OK.
> 
>             foreach (var x in t) { } //crash
> 
> ```
> 
> Error:
> NHibernate.Hql.Ast.ANTLR.QuerySyntaxException: Exception of type 'Antlr.Runtime.MismatchedTreeNodeException' was thrown. [].
> 
> Error from log:
> ERROR NHibernate.Hql.Parser - MismatchedTreeNodeException(3!=2)
> 
> 
> 
---
**Fabio Maulo** added a comment — <time datetime="2009-09-02T23:20:15.062Z" title="2009-09-02T23:20:15.062Z">2nd September 2009, 23:20:15</time>:
> I think that was never supported.
>

bahusoid:
Fixes #722

### DIFF
--- a/src/NHibernate/Impl/ExpressionQueryImpl.cs
+++ b/src/NHibernate/Impl/ExpressionQueryImpl.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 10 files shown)


hazzik:
Need to override ToList in LinqFilterImpl

bahusoid:
Oh.. Thanks. Will fix tommorow. But why firebird only?

hazzik:
Because it does not support multi-query and that code path executes `List<T>` instead of `List`.

src/NHibernate/Impl/ExpressionQueryImpl.cs
```diff
@@ -125,6 +125,32 @@ public override IList List()
 			}
 		}
 
+		public override IList<T> List<T>()
```
hazzik: Lack of this override caused to call `Session.List` and this resulted in a lost `collection` parameter. So the query applied to the whole table.

fredericDelaporte: For correctness of the class, shouldn't a `void List(IList results)` override be added too? (And same for `CollectionFilterImpl`.)

bahusoid: Done. I've also added missing Enumerable ovverrides to `ExpressionQueryImpl` to throw NotSupportedException.

PR #341, opened 2014-09-22
NH-3704 - Allow Setting Dynamic Component Templates From Dictionary

rjperes:
#913 ([NH-3704](https://nhibernate.jira.com/browse/NH-3704))
With unit tests

PR #1941, opened 2018-12-20
Allow setting dynamic component templates from dictionary in ByCode

### LINKED ISSUE #913: NH-3704 - Allow Setting Dynamic Component Templates From Dictionary
**Ricardo Peres** created an [issue](https://nhibernate.jira.com/browse/NH-3704) — <time datetime="2014-09-21T09:18:33.863Z" title="2014-09-21T09:18:33.863Z">21st September 2014, 9:18:33</time>:
> Currently, when mapping a dynamic component with mapping by code, we need to specify the dynamic component template (the columns to store) as an anonymous type, which isn't very good for dynamic scenarios, where these columns and their types are obtained dynamically. My proposal is to use a dictionary (`IDictionary<string, Type>` or `IDictionary<string, object>`) for that.
> Pull request will follow shortly.
> 
---
**Ricardo Peres** added a comment — <time datetime="2014-09-22T00:05:55.177Z" title="2014-09-22T00:05:55.177Z">22nd September 2014, 0:05:55</time>:
> Pull request: #341
>

fredericDelaporte:
Fixes #913 - NH-3704

Replaces #341. It is a rebase of it, with fixes required due to changes in NHibernate.

The original PR repository being deleted, I do not see any way to work with it other than opening a new PR.

The original PR was put in the 6.0 milestone, but it does not have breaking changes. It can be put in the 5.3 milestone.

There was a discussion about the implementation relying on dynamically building a type matching the component. It allows to re-use the existing By-Code logic working on anonymous types, rather than coding a whole new code path for mapping dynamic components by code. It is a bit runtime heavy way of doing things, but it seems doing it another way will require a bit to much changes in By-Code.

If approved, it should be manually squashed before merging, in order to keep Ricardo as the main author.

### DIFF
--- a/src/NHibernate.Test/MappingByCode/ExplicitMappingTests/DynamicComponentMappingTests.cs
+++ b/src/NHibernate.Test/MappingByCode/ExplicitMappingTests/DynamicComponentMappingTests.cs
@@ -6,7 +6,7 @@ using NHibernate.Cfg.MappingSchema;
 using NHibernate.Mapping.ByCode;
 using NUnit.Framework;
 
-namespace NHibernate.Test.MappingByCode.ExpliticMappingTests
+namespace NHibernate.Test.MappingByCode.ExplicitMappingTests
 {
 	[TestFixture]
 	public class DynamicComponentMappingTests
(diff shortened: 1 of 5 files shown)


src/NHibernate.Test/MappingByCode/ExplicitMappingTests/DynamicComponentMappingTests.cs
```diff
@@ -33,6 +33,137 @@ private class PersonWithGenericInfo
...
+			var mapper = new ModelMapper();
+			mapper.Class<Person>(
+				map =>
+				{
+					map.Id(x => x.Id, idmap => { });
+					map.Component(
+						x => x.Info,
+						new Dictionary<string, System.Type>
+							{ { "MyInt", typeof(int) }, { "MyDate", typeof(DateTime) } },
+						z => { });
```
hazzik: This PR isn't feature complete. Further configuration is not possible:
```suggestion
						z => z.Property("MyInt").Column("MyIntColumn"));
```

It throws: 
```
NHibernate.MappingException : Member not found. The member 'MyInt' does not exists in type System.Collections.Generic.Dictionary`2[[System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089],[System.Type, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]
```

bahusoid:
Wouldn't it be simpler to add method that takes `dynamic` component
```C#
public static void Component<TContainer>(
			this IBasePlainPropertyContainerMapper<TContainer> mapper,
			Expression<Func<TContainer, IDictionary<string, object>>> property,
			dynamic dynamicComponentTemplate,
			Action<IDynamicComponentMapper<dynamic>> mapping)
```
			
			
And add additional extension for dictionary that wraps dictionary to dynamic via ExpandoObject
```C#
public static void Component<TContainer, dynamic>(
			this IBasePlainPropertyContainerMapper<TContainer> mapper,
			Expression<Func<TContainer, IDictionary<string, object>>> property,
			IDictionary<string, Type> dynamicComponentTemplate,
			Action<IDynamicComponentMapper<dynamic>> mapping)
{
//create ExpandoObject from dictionary and call dynamic overload
}
```

Just a quick idea - I didn't really thought it through.

bahusoid:
Ok. I've made a quick look and that's not how `ExpandoObject` works. So it seems dynamic won't  really help here. But still I think it's better to have separate method that take IDictionary parameter. It would also allow to provide different type for mapping action if required. Something like:
```C#
public static void Component<TContainer>(
			this IBasePlainPropertyContainerMapper<TContainer> mapper,
			Expression<Func<TContainer, IDictionary<string, object>>> property,
			IDictionary<string, Type> dynamicComponentTemplate,
			Action<IDynamicComponentMapper<object>> mapping) 
```

fredericDelaporte:
Feel free to try your suggestions directly here if you wish. I have less time available currently for contributing, and I am not prioritizing this PR over other PRs. I will probably not look into the trouble raised by Alexander before next week-end.

bahusoid:
> Feel free to try your suggestions directly here if you wish

Ok I've checked the code and my suggestions are not really helpful. I don't know how to make this delegate work without huge refactoring

src/NHibernate/Mapping/ByCode/Impl/CustomizersImpl/PropertyContainerCustomizer.cs
```diff
@@ -118,6 +149,34 @@ protected virtual void RegisterDynamicComponentMapping<TComponent>(Expression<Fu
...
+			MemberInfo member = TypeExtensions.DecodeMemberAccessExpression(property);
+			MemberInfo memberOf = TypeExtensions.DecodeMemberAccessExpressionOf(property);
+			RegisterDynamicComponentMapping(template, mapping, member, memberOf);
+		}
+
+		protected void RegisterDynamicComponentMapping<TComponent>(IDictionary<string, System.Type> template, Action<IDynamicComponentMapper<TComponent>> mapping, params MemberInfo[] members)
+		{
+			foreach (var member in members)
+			{
+				mapping(new DynamicComponentCustomizer<TComponent>(template, explicitDeclarationsHolder, CustomizersHolder, new PropertyPath(PropertyPath, member)));
```
hazzik: Here, inside `new DynamicComponentCustomizer<TComponent>(...)` for every member we call `CreateType()`

hazzik:
@fredericDelaporte @bahusoid 
image

src/NHibernate/Mapping/ByCode/Impl/CustomizersImpl/PropertyContainerCustomizer.cs
```diff
@@ -508,6 +661,14 @@ public void Component<TComponent>(string notVisiblePropertyOrFieldName, TCompone
...
 
+		protected virtual MemberInfo GetRequiredPropertyOrFieldByName(string memberName)
+		{
+#pragma warning disable 618
+			return GetPropertyOrFieldMatchingNameOrThrow(memberName);
+#pragma warning restore 618
+		}
+
+		[Obsolete("Please use GetRequiredPropertyOrFieldByName instead.")]
 		public static MemberInfo GetPropertyOrFieldMatchingNameOrThrow(string memberName)
```
hazzik: Having this method static really sucks.

src/NHibernate.Test/MappingByCode/ExplicitMappingTests/DynamicComponentMappingTests.cs
```diff
@@ -33,6 +33,141 @@ private class PersonWithGenericInfo
...
+			var mapper = new ModelMapper();
+			mapper.Class<Person>(
+				map =>
+				{
+					map.Id(x => x.Id, idmap => { });
+					map.Component(
+						x => x.Info,
+						new Dictionary<string, System.Type>
+							{{"MyInt", typeof(int)}, {"MyDate", typeof(DateTime)}},
+						(IDynamicComponentMapper<object> z) => { z.Property("MyInt", pm => pm.Column("MY_COLUMN")); });
```
hazzik: This is not really discoverable at the moment.

src/NHibernate/Mapping/ByCode/Impl/CustomizersImpl/PropertyContainerCustomizer.cs
```diff
@@ -94,14 +94,59 @@ protected void RegisterComponentMapping<TComponent>(Action<IComponentMapper<TCom
...
+		/// <param name="property">The property to map.</param>
+		/// <param name="dynamicComponentTemplate">The template for the component. It should either be a (usually
+		/// anonymous) type having the same properties than the component, or an
+		/// <c>IDictionary&lt;string, System.Type&gt;</c> of property names with their type.</param>
+		/// <param name="mapping">The mapping of the component.</param>
+		/// <typeparam name="TComponent">The type of the template.</typeparam>
 		public void Component<TComponent>(Expression<Func<TEntity, IDictionary>> property, TComponent dynamicComponentTemplate, Action<IDynamicComponentMapper<TComponent>> mapping)
 		{
-			RegisterDynamicComponentMapping(property, mapping);
+			if (dynamicComponentTemplate is IDictionary<string, System.Type> dicTemplate && mapping is Action<IDynamicComponentMapper<object>> objectMapping)
```
bahusoid: Is it still valid check? It seems separate method needs to be used for dictionary.

hazzik: >Is it still valid check?

Yes

> It seems separate method needs to be used for dictionary.

Unfortunately this is not really possible because the generic instance overload takes precedence over extension methods.

bahusoid: Hm... But `TComponent` here is  `Dictionary<string, Type>`.
I don't get how `Action<IDynamicComponentMapper<Dictionary<string, Type>>>` can be casted to `Action<IDynamicComponentMapper<object>> objectMapping`

hazzik: It's easy. `typeof(TComponent) == typeof(object)`.

bahusoid: But it means user needs to explicitly cast to `object` to make it work. Not obvious at all.  This I'm pretty sure will lead to many confusions. 
Wouldn't it be better to simply introduce separate method for dictionary? Something like `ComponentForDic`

hazzik: > Not obvious at all. This I'm pretty sure will lead to many confusions.

This is what I've meant [here](https://github.com/nhibernate/nhibernate-core/pull/1941#discussion_r268542709).

bahusoid: So  solution would be to add separate method. This avoids castings here and makes new method discoverable and user doesn't need to know that to make it work explicit cast to object is required. 

What disadvantages do you see?

hazzik: I've solved this by just checking for dictionary with any type of mapper. So, it's not a problem anymore.

src/NHibernate.Test/MappingByCode/ExplicitMappingTests/DynamicComponentMappingTests.cs
hazzik: Solved. Now one can use just `z => ... `

src/NHibernate/Mapping/ByCode/Impl/CustomizersImpl/PropertyContainerCustomizer.cs
bahusoid: Cool

fredericDelaporte:
Rebased and squashed, for allowing merging while keeping its original author.

PR #1942, opened 2018-12-20
WIP - Use Lambda Expressions for field/property accessor when UseReflectionOptimizer enabled

maca88:
The performance tests shows up to 10 times faster than before for getting or setting a property/field.

### DIFF
--- a/src/NHibernate/Properties/FieldAccessor.cs
+++ b/src/NHibernate/Properties/FieldAccessor.cs
@@ -163,6 +163,7 @@ namespace NHibernate.Properties
 			private readonly FieldInfo field;
 			private readonly System.Type clazz;
 			private readonly string name;
+			private readonly Lazy<Func<object, object>> _getDelegate;
 
 			/// <summary>
 			/// Initializes a new instance of <see cref="FieldGetter"/>.
@@ -175,6 +176,13 @@ namespace NHibernate.Properties
 				this.field = field;
 				this.clazz = clazz;
 				this.name = name;
+
+				if (!Cfg.Environment.UseReflectionOptimizer)
+				{
+					return;
+				}
+
+				_getDelegate = new Lazy<Func<object, object>>(() => PropertyAccessorFactory.CreateGetDelegate(this, clazz));
 			}
 
 			#region IGetter Members
@@ -190,7 +198,9 @@ namespace NHibernate.Properties
 			{
 				try
 				{
-					return field.GetValue(target);
+					return _getDelegate == null
+						? field.GetValue(target)
+						: _getDelegate.Value(target);
 				}
 				catch (Exception e)
 				{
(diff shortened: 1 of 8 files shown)


hazzik:
I would rewrite this using dynamic lambda.

maca88:
Done, the performance is more or less the same.
I realized that I mistakenly pushed the branch to this repository, should I remove it and create a new PR?

hazzik:
>I realized that I mistakenly pushed the branch to this repository, should I remove it and create a new PR?

No, it's fine

fredericDelaporte:
When `UseReflectionOptimizer` is true, the `Getter.Get` or `Setter.Set` are not actually used, are they? Normally IL generated getters/setters built from their `Emit` methods are used instead. So it seems to me there are no points optimizing `Getter.Get` or `Setter.Set` when `UseReflectionOptimizer` is true.

Am I missing something?

maca88:
> When UseReflectionOptimizer is true, the Getter.Get or Setter.Set are not actually used, are they? 

Yes they still are as `IAccessOptimizer` has only methods for setting/getting all properties. Here are the methods that are using it regardless `UseReflectionOptimizer`:
```
IEntityTuplizer.GetIdentifier
IEntityTuplizer.SetIdentifier
IEntityTuplizer.GetPropertyValue
IEntityTuplizer.SetPropertyValue
IEntityTuplizer.GetVersion
ITuplizer.GetPropertyValue
```

And also `IEntityTuplizer.SetPropertyValues` and `IEntityTuplizer.GetPropertyValues` when the entity has lazy properties.

src/NHibernate/Properties/FieldAccessor.cs
```diff
@@ -158,11 +160,13 @@ private string GetFieldName(string propertyName)
...
 		/// </summary>
 		[Serializable]
-		public sealed class FieldGetter : IGetter, IOptimizableGetter
+		public sealed class FieldGetter : IGetter, IOptimizableGetter, IDeserializationCallback
 		{
 			private readonly FieldInfo field;
 			private readonly System.Type clazz;
 			private readonly string name;
+			[NonSerialized]
+			private Lazy<Func<object, object>> _getDelegate;
```
fredericDelaporte: Using a lazy here or in property accessor looks like #1904 laziness, which was rejected: it may reduce start-up time but will cause warming-up more hard to achieve.

It looks even more debatable than #1904, because #1904 was more likely to allow not initializing things never actually used by the application. But about getter/setter used notably for accessing an entity identifier, having them never used would require the application to map useless entities, which should normally not be done.

maca88: I've made an alternative approach (#1944) that avoids this problem.

fredericDelaporte:
Replaced by #1944.

PR #1944, opened 2018-12-21
Extend IAccessOptimizer to support getting/setting single property value

maca88:
This is an alternative approach (#1942) to avoid reflection when `UseReflectionOptimizer` enabled. Here I was forced to use IL as `IOptimizableSetter` exposes only an `Emit` method. I think that this approach is better, but is also more invasive and contains some ugly code to avoid a breaking change.


src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
```diff
@@ -66,6 +68,8 @@ public PocoEntityTuplizer(EntityMetamodel entityMetamodel, PersistentClass mappe
 				if (property.UnwrapProxy)
 					unwrapProxyPropertyNames.Add(property.Name);
 			}
+
+			isBytecodeProviderImpl = Cfg.Environment.BytecodeProvider is BytecodeProviderImpl;
```
fredericDelaporte: It should be moved into `SetReflectionOptimizer`, because nothing guarantees the byte code provider will be the same on deserialization. By the way the test could be on the `optimizer.AccessOptimizer` concrete implementation instead, which would be more straightforward.

src/NHibernate/Tuple/Component/PocoComponentTuplizer.cs
```diff
@@ -52,6 +54,7 @@ public PocoComponentTuplizer(Mapping.Component component)
 				parentGetter = parentProperty.GetGetter(componentClass);
 			}
 
+			isBytecodeProviderImpl = Cfg.Environment.BytecodeProvider is BytecodeProviderImpl;
```
fredericDelaporte: It should be moved into `SetReflectionOptimizer`, because nothing guarantees the byte code provider will be the same on deserialization. By the way the test could be on the `optimizer.AccessOptimizer` concrete implementation instead, which would be more straightforward.

src/NHibernate/Properties/FieldAccessor.cs
```diff
@@ -331,6 +331,11 @@ public void Emit(ILGenerator il)
 			{
 				il.Emit(OpCodes.Stfld, field);
 			}
+
+			internal bool CanEmit()
+			{
+				return !field.IsInitOnly;
```
fredericDelaporte: Is this actually supported only for id/parent properties? Otherwise I do not understand how it works for the optimized `setters` we currently already have.

src/NHibernate/Bytecode/Lightweight/BytecodeProviderImpl.cs
```diff
@@ -24,9 +24,25 @@ public class BytecodeProviderImpl : AbstractBytecodeProvider
...
+		/// capable of generating reflection optimization components.
+		/// </summary>
+		/// <param name="mappedClass">The class to be reflected upon.</param>
+		/// <param name="getters">All property getters to be accessed via reflection.</param>
+		/// <param name="setters">All property setters to be accessed via reflection.</param>
+		/// <param name="specializedGetter">The specialized getter for the given type.</param>
+		/// <param name="specializedSetter">The specialized setter for the given type.</param>
+		/// <returns>The reflection optimization delegate.</returns>
+		internal IReflectionOptimizer GetReflectionOptimizer(System.Type mappedClass, IGetter[] getters, ISetter[] setters,
+																	IGetter specializedGetter, ISetter specializedSetter)
```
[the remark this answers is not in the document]
maca88: Done.

src/NHibernate/Bytecode/Lightweight/ReflectionOptimizer.cs
```diff
@@ -45,7 +56,29 @@ public ReflectionOptimizer(System.Type mappedType, IGetter[] getters, ISetter[]
...
+			}
+
+			accessOptimizer = new AccessOptimizer(
+				getInvoker,
+				setInvoker,
+				getMethods,
+				setMethods,
+				new Getter(
+					specializedGetter,
+					specializedGetter != null ? GenerateGetPropertyValueMethod(specializedGetter) : null),
```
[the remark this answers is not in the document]
maca88: I've removed it.

src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
maca88: Moved to into `SetReflectionOptimizer`. 

> By the way the test could be on the optimizer.AccessOptimizer concrete implementation instead

Is it not enough as the user can use the obsolete constructor that does not have the specialized getter/setter and it would throw a NRE on (Set\Get)SpecializedPropertyValue. We have to use the same condition as in the `GetReflectionOptimizer` extension method in order to avoid that.

src/NHibernate/Properties/FieldAccessor.cs
maca88: My mistake, I didn't use the owner argument when creating a `DynamicMethod` and without the owner you can't set a readonly field. I forgot to use the `CreateDynamicMethod` in `ReflectionOptimizer` when copy/pasting the code from the previous PR.

src/NHibernate/Properties/FieldAccessor.cs
```diff
@@ -331,6 +331,11 @@ public void Emit(ILGenerator il)
 			{
 				il.Emit(OpCodes.Stfld, field);
 			}
+
+			internal bool CanEmit()
```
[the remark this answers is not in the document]
maca88: I've removed it as there is no need for it anymore.

maca88:
Here are some performance numbers of the added [Get\Set]PropertyValue methods:
```
IGetter.Get total time for 100000 iterations: 131ms
IAccessOptimizer.GetPropertyValue total time for 100000 iterations: 37ms
IAccessOptimizer.GetPropertyValues total time for 100000 iterations: 16ms

ISetter.Set total time for 100000 iterations: 242ms
IAccessOptimizer.SetPropertyValue total time for 100000 iterations: 49ms
IAccessOptimizer.SetPropertyValues total time for 100000 iterations: 23ms
```

src/NHibernate/Bytecode/Lightweight/Getter.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using NHibernate.Properties;
+
+namespace NHibernate.Bytecode.Lightweight
+{
+	/// <summary>
+	/// Contains the <see cref="IGetter"/> instance with an optional optimized delegate.
+	/// </summary>
+	public class Getter
```
fredericDelaporte: Why not hiding the fact it is optimized or not?

It could be something like:

```c#
public class OptimizedGetter
{
	private readonly Func<object, object> _getter;

	public OptimizedGetter(IGetter @default) : this(@default, null)
	{
	}

	public OptimizedGetter(IGetter @default, Func<object, object> optimized)
	{
		_getter = optimized ?? @default.Get;
	}

	public object Get(object target)
	{
		return _getter(target);
	}
}
```

It would allow a more straightforward implementation of the access optimizer.

Of course the same should then be done for setters.

maca88: > Why not hiding the fact it is optimized or not?

I agree, I think it is even better to use a delegate to be consistent with [Get/Set]PropertyValues, which is what I've done. If you think that it would be better to have a class I will change to it.

src/NHibernate/Bytecode/Lightweight/ReflectionOptimizer.cs
```diff
@@ -45,7 +56,26 @@ public ReflectionOptimizer(System.Type mappedType, IGetter[] getters, ISetter[]
...
+				setMethods[i] = GenerateSetPropertyValueMethod(setters[i]) ?? setters[i].Set;
+			}
+
+			accessOptimizer = new AccessOptimizer(
+				getInvoker,
+				setInvoker,
+				getMethods,
+				setMethods,
+				GenerateGetPropertyValueMethod(specializedGetter),
+				GenerateSetPropertyValueMethod(specializedSetter)
```
fredericDelaporte: I am thinking about it a bit late, but it does not look to me guaranteed that the specializedGetter/Setter will be optimizable. In such case, we need a fallback here, something like `?? specializedGetter?.Get`. Same for the setter.

maca88: You are right, I somehow overlooked it. `specializedGetter?.Get` does not work so I added a ternary if instead with a todo for removing it once the obsolete constructor will be removed

src/NHibernate/Bytecode/Lightweight/ReflectionOptimizer.cs
```diff
@@ -132,6 +167,46 @@ private static void EmitCastToReference(ILGenerator il, System.Type type)
 		private static readonly MethodInfo GetterCallbackInvoke = ReflectHelper.GetMethod<GetterCallback>(
 			g => g.Invoke(null, 0));
 
+		private GetPropertyValueInvoker GenerateGetPropertyValueMethod(IGetter getter)
+		{
+			if (!(getter is IOptimizableGetter optimizableGetter))
+			{
+				return null;
```
fredericDelaporte: We should put the fallback right here rather than yielding `null`. It simplifies the code and aligns with `GenerateGetPropertyValuesMethod`, which also handles the fallback itself.

src/NHibernate/Bytecode/Lightweight/ReflectionOptimizer.cs
```diff
@@ -163,10 +163,10 @@ private static void EmitCastToReference(ILGenerator il, System.Type type)
 
 		private GetPropertyValueInvoker GenerateGetPropertyValueMethod(IGetter getter)
 		{
+			if (getter == null)
+				return null;
```
fredericDelaporte: I forgot the `null` case... Now that we do no more yield just null when the argument is not optimizable, the case it is null has to be handled separately, because the null coalesce operator does not work on method groups. (And the ternary requires some explicit casting.)

PR #1946, opened 2018-12-24
Port Hibernate's BytecodeEnhancementMetadata

maca88:
The idea of the `BytecodeEnhancementMetadata` is to have all information about lazy and unwrap-proxy properties and methods related to the field interceptor into one place that can be accessed via `IEntityPersister` or `EntityMetamodel`. This class is also a prerequisite to port the Hibernate lazy [attribute fetch groups](https://github.com/hibernate/hibernate-orm/commit/1e44e7420b7af10adc9547a461f02a6d979624f1) feature.


hazzik:
Rebased

src/NHibernate/Engine/EntityEntry.cs
```diff
@@ -265,7 +269,7 @@ public bool RequiresDirtyCheck(object entity)
 			return
 				IsModifiableEntity()
 				&& (Persister.HasMutableProperties || !FieldInterceptionHelper.IsInstrumented(entity)
```
fredericDelaporte: Shouldn't `FieldInterceptionHelper.IsInstrumented(entity)` be changed for `Persister.IsInstrumented`?

src/NHibernate/Intercept/FieldInterceptionHelper.cs
```diff
@@ -32,6 +33,7 @@ public static IFieldInterceptor ExtractFieldInterceptor(object entity)
 			return fieldInterceptorAccessor == null ? null : fieldInterceptorAccessor.FieldInterceptor;
 		}
 
+		[Obsolete("Use IBytecodeEnhancementMetadata.InjectInterceptor method instead")]
```
fredericDelaporte: I think `IsInstrumented(object entity)` should be obsoleted too: all its remaining calls in NHibernate project can be replaced by `persister.IsInstrumented` calls, or simply removed. Two calls in tests would have to be replaced by an interface check.

src/NHibernate/Engine/EntityEntry.cs
maca88: There is no gaurantee that when `Persister.IsInstrumented` is true the entity is actually instrumental. When the entity is created by the user (`session.Save(new Book())`) the entity won't be instrumental and in this particular case is not the same if we change to `Persister.IsInstrumented`. For the other cases we can do it but we have to add an additinal null-conditional (?.) operator when calling the interceptor methods (`interceptor?.MarkDrity()`). The `Persister.IsInstrumented` only tell us that the entity will be instrumented when the entity is created by NHibernate.
We could move the `IsInstrumented` method in `IBytecodeEnhancementMetadata` or just use `entity is IFieldInterceptorAccessor` for this particular case.

maca88:
I've addressed all the issues from the comments and obsoleted also `FieldInterceptionHelper.ExtractFieldInterceptor` method.

src/NHibernate/Engine/EntityEntry.cs
fredericDelaporte: But then here, we can write it as `Persister.HasMutableProperties || !Persister.IsInstrumented || (Persister.ExtractFieldInterceptor(entity)?.IsDirty ?? true)`. So no need for the direct type checking, if we cease using `FieldInterceptionHelper.IsInstrumented(entity)`.


(But in fact I suspect the current condition is wrong. I think it should have been
`Persister.HasMutableProperties || FieldInterceptionHelper.IsInstrumented(entity) && Persister.ExtractFieldInterceptor(entity).IsDirty`. (So, rewritten `Persister.HasMutableProperties || Persister.IsInstrumented && (Persister.ExtractFieldInterceptor(entity)?.IsDirty ?? false)`.)
This dates back to the port from Hibernate, but on Hibernate side, the check on field interceptor has been completely removed since then. (So, just `Persister.HasMutableProperties` remains, the "or" are removed.) That is maybe a bit too unrelated for being changed though.)

maca88: > Hibernate side, the check on field interceptor has been completely removed since then

I've checked Hibernate code and the check [on field interceptor is still there](https://github.com/hibernate/hibernate-orm/blob/e5dc635a52362f69b69acb8d5b166b69b165dbbd/hibernate-core/src/main/java/org/hibernate/engine/internal/AbstractEntityEntry.java#L344-L346), but they moved it to `SelfDirtinessTracker` interface instead. They also added a check for collection ([HHH-11293](https://hibernate.atlassian.net/browse/HHH-11293)) as a collection can change and the interceptor will not detect it, probably we should add it too.

> I think it should have been
Persister.HasMutableProperties || FieldInterceptionHelper.IsInstrumented(entity) && Persister.ExtractFieldInterceptor(entity).IsDirty

[In Hibernate](https://github.com/hibernate/hibernate-orm/blob/e5dc635a52362f69b69acb8d5b166b69b165dbbd/hibernate-core/src/main/java/org/hibernate/engine/internal/AbstractEntityEntry.java#L354-L358) `true` will be returned when the entity is not instrumented even if `HasMutableProperties` is `false`.

maca88: If we want to have the same logic as Hibernate does we should rewrite the condition to:
```
return
  IsModifiableEntity()
  && (!Persister.IsInstrumented || Persister.HasCollections
  || (Persister.EntityMetamodel.BytecodeEnhancementMetadata.ExtractInterceptor(entity)?.IsDirty ?? true));
```

PR #1947, opened 2018-12-26
Partially port Hibernate's current field interceptor mechanism

maca88:
As [already discussed](https://github.com/nhibernate/nhibernate-core/pull/1943#issuecomment-449585501), the Hibernate field interceptor works differently that the current NHibernate's one and this PR is aimed to shrink the differences between the two. The most important difference that this PR ported is when the field interceptor is injected. In NHibernate the interceptor is injected after the entity is hydrated and in Hibernate it is injected right after the entity is instantiated. By injecting the interceptor at the entity creation the interceptor will track all uninitialized properties and we don't need to calculate them after the entity is initialized, which is what #1922 currently does. Another mechanism that this PR aligns is the ability to retrieve an initialized unwrapped proxy after the session is closed:
```
using (var s = OpenSession())
{
  order = s.Get<Order>(1);
  payment = order.Payment; // Initialize Payment
}
Assert.That(order.Payment, Is.EqualTo(payment));
```
The current code will throw as the payment was not set upon initialization.

Also, another thing that was done in this PR and does not work in Hibernate is to update the state of the initialized lazy="no-proxy" properties when they are set, for example:
```
using (var s = OpenSession())
{
  var order = s.Get<Order>(1);
  Assert.That(order.Payment is WireTransfer, Is.True); // Load property
  Assert.That(NHibernateUtil.IsPropertyInitialized(order, "Payment"), Is.True);
  order.Payment = s.Load<Payment>(2); // Load a different payment
  Assert.That(NHibernateUtil.IsPropertyInitialized(order, "Payment"), Is.False); // The initialized state of the Payment was updated
}
```
I think it is more correct to update the state as otherwise we are not able to detect if the `Payment` is really initialized or not.

There is still some differences that this PR does not address:
1. In Hibernate properties with lazy="no-proxy" are treated the same way as lazy properties, which means that the property will not be present in the query when the entity is retrieved from the database:
```
doInHibernate( this::sessionFactory, s -> {
  a = s.get( A.class, 1L );
  b = a.b; // Initialize b (b is a many-to-one relaton mapped as lazy="no-proxy")
  assertTrue(Hibernate.isPropertyInitialized(a, "b"));
} );
```
The above code will produce 3 queries:
- getting the entity `a` without containg the `b` property
- getting the `b` property
- getting the `b` entity
2. In Hibernate uninitialized lazy properties can optionally be accessed when the session is closed, a read only session is created by the interceptor to retrieve the value and then disposed.

I think that NHibernate behavior for lazy="no-proxy" is more practical/usefull as I don't see the point of not selecting the id of `b` as the main point of lazy propeties is to avoid expensive formulas or columns with large data.

The breaking change of this PR is the changed behavior of `NHibernateUtil.IsPropertyInitialized` which now returns the actual initialization state for the property.

EDIT:

By injecting the interceptor at the entity creation the issue of fetching a property mapped as lazy="no-proxy" for an entity that is not yet loaded is now working (#1267 second test).

Possible breaking change: assigning an uninitialized proxy to a `no-proxy` property will no more trigger the proxy initialization. Moreover, reading the property afterwards will no more unwrap the assigned proxy, but will yield it.


fredericDelaporte:
> I think it is more correct to update the state as otherwise we are not able to detect if the Payment is really initialized or not.

I am not convinced of this, because we have two methods: `IsPropertyInitialized`, about the property state, and `IsInitialized`, about a proxy state, whatever the property it comes from (if any).

Of course, using `IsInitialized` directly without checking first `IsPropertyInitialized` on a `no-proxy` property will force the property load if it was not already loaded.
But putting an uninitialized proxy  in a `no-proxy` property looks to me as quite a special case anyway, since `no-proxy` properties are meant to avoid having proxies yielded by the property. So having to use two methods in order to fully check that special case does not look as an issue to me.

> which now returns the actual initialization state for the property

I do not really see that as "actual initialization state for the property". For me, the actual state is what current behavior yields. Putting an uninitialized proxy in the property causes the property to be initialized. But with a thing that is itself not initialized.

maca88:
> But putting an uninitialized proxy in a no-proxy property looks to me as quite a special case anyway

Not anymore as the interceptor now is injected from the entity creation so the interceptor will recieve a proxy for a `no-proxy` relation in `TwoPhaseLoad.InitializeEntity` if the relation was not fetched. So the interceptor has to be smarter in order to avoid loading all `no-proxy` relations in 2-phase load.
About updating the state there is one scenario that would be helpful which is serialization. In order to avoid serializing a not loaded relation a mechanism is needed that is able to check whether the property is loaded or not and if the mechanim does not have the actual state an uninitialized relation could be loaded which could cause performance issues. Indeed, normally if the relation exists we would want to include it in the serialization result, but when using a clinet side ORM like `BreezeJS` or `JayData` which are aware of entity relations and their loaded state we don't want the uninitialized relation to be loaded as I can be loaded from the client side ORM when needed.

fredericDelaporte:
The fact that the proxy id is retrieved immediately is an implementation detail. I do not think `IsPropertyInitialized` on a `no-proxy` association has to care about it.

The normal behavior of a `no-proxy` association is to yield a hydrated, un-proxied entity. For doing this, it may use a proxy under the hood, but that is just a hidden implementation detail. `IsPropertyInitialized` has only to tell whether accessing the property will yield its value without triggering a db access, or not.

When only the association id is loaded, it will cause a db access: the property has to be considered non-initialized.
When the hidden proxy inner entity state has already been loaded, it will just yield that entity state without hitting the db: the property can be considered initialized.

>> But putting an uninitialized proxy in a no-proxy property looks to me as quite a special case anyway
>
>Not anymore as the interceptor is now injected from the entity creation so the interceptor will receive a proxy for a `no-proxy` relation in `TwoPhaseLoad.InitializeEntity` if the relation was not fetched.

I hope this is still a hidden implementation detail: I mean, when reading the property, we still get an unproxied entity. Otherwise the purpose of `no-proxy` is defeated by the change here.

In such case, when an user put an uninitialized proxy in a no-proxy property, this is still an edge and a bit weird case. I have not check the behavior of the property in such case by the way, but if, when reading it afterwards, it yields an hydrated un-proxied entity instead of the proxy the user has assigned to it, that would be a debatable behavior. It should yield back what the user has assigned to it instead, for not breaking the unicity of the entity when working with a single session.

fredericDelaporte:
I have "resolved" my first comment because it was based on a false assumption.

>> I think it is more correct to update the state as otherwise we are not able to detect if the Payment is really initialized or not.
>
>I am not convinced of this, because we have two methods: `IsPropertyInitialized`, about the property state, and `IsInitialized`, about a proxy state, whatever the property it comes from (if any).

All this was about the case where the user assigns a proxy to a `no-proxy` property. I was assuming accessing the property afterwards would have yielded back the proxy, but this is not the case. It yields the unwrapped entity.

In such case, I agree the `IsPropertyInitialized` must yield `false` if the proxy was not initialized, because accessing the property will trigger a db-access.

About whether the `no-proxy` behavior should be changed for yielding back what the user has assigned to it, or not, better handle that in another PR. (But well, although I think this behavior debatable, it can be argued that it allows to have  `no-proxy` somewhat strictly `no-proxy`. Why not, I do not think I would open this another PR myself.)

maca88:
> The fact that the proxy id is retrieved immediately is an implementation detail. I do not think IsPropertyInitialized on a no-proxy association has to care about it.

Fine by me.

> IsPropertyInitialized has only to tell whether accessing the property will yield its value without triggering a db access, or not.

This is exactly how the current behavior of this PR works.

> I hope this is still a hidden implementation detail: I mean, when reading the property, we still get an unproxied entity

Of course, this is the whole purpose of `no-proxy`.

> I have not check the behavior of the property in such case by the way

Currently, NHibernate will initialize the entity on set as it does not return earlier when the `setter` argument of the `Intercept` method is `true` like this PR does. 

>  It should yield back what the user has assigned to it instead, for not breaking the unicity of the entity when working with a single session.

I am ok with this behavior but we would still need to list this as a breaking change as currently the proxy will be initialized on set. This behavior could be achieved by checking if the session is set on the interceptor or not as the session is set in the `AfterInitialize` method which is after the 2-phase load.

In Hibernate when setting a proxy on a initialized `no-proxy` property works the same as you described that it should be working:
```
doInHibernate( this::sessionFactory, s -> {
  child = s.get( Child.class, 1L );
  assertThat( child, notNullValue() );

  assertFalse( isPropertyInitialized( child, "parent" ) );
  // Initialize parent no-proxy property
  assertThat(child.parent, instanceOf(Parent.class));
  assertTrue( isPropertyInitialized( child, "parent" ) );

  Parent p = s.load(Parent.class, 2L);
  assertFalse(Hibernate.isInitialized(p));
  // Set a proxy to the no-proxy property
  child.parent = p;
  assertFalse(Hibernate.isInitialized(p)); // set will not trigger the proxy to be initialized
  assertTrue( isPropertyInitialized( child, "parent" ) );

  assertTrue(Hibernate.isInitialized(child.parent)); //  fails as an uninitialized proxy is returned

} );
```

but in the following scenario the behavior is quite unexpected (at least for me):
```
doInHibernate( this::sessionFactory, s -> {
  child = (Child)s.createQuery("from Child fetch all properties where id = :id") // This is the same as s.Get<Child>(1) in NHibernate as no-proxy properties are always retrieved
      .setParameter("id", 1L)
      .uniqueResult();

  assertTrue( isPropertyInitialized( child, "parent" ) );
  assertTrue(Hibernate.isInitialized(child.parent)); // fails as an uninitialized proxy is returned
} );
```
I think that is not how it should work as the `no-proxy` property was not set by the user.

src/NHibernate.Test/LazyOneToOne/LazyOneToOneTest.cs
```diff
@@ -67,8 +67,8 @@ public void Lazy()
 
 			s = OpenSession();
 			t = s.BeginTransaction();
-			p = s.Get<Person>("Gavin");
-			Assert.That(!NHibernateUtil.IsPropertyInitialized(p, "Employee"));
+			p = s.Get<Person>("Gavin"); // The default loader will fetch the employee
+			Assert.That(NHibernateUtil.IsPropertyInitialized(p, "Employee"));
```
fredericDelaporte: Isn't that a regression? `Employee` property was supposed to be lazy, but it looks to be no more.

src/NHibernate/Proxy/StaticProxyFactory.cs
```diff
@@ -43,6 +43,11 @@ public override INHibernateProxy GetProxy(object id, ISessionImplementor session
 		}
 
 		public override object GetFieldInterceptionProxy(object instanceToWrap)
```
fredericDelaporte: In order to not forget removing it, I think it should be obsoleted, and its base too.

src/NHibernate/Proxy/IProxyFactory.cs
```diff
@@ -49,6 +50,37 @@ public interface IProxyFactory
...
+		// 6.0 TODO: Remove
+		internal static object GetFieldInterceptionProxy(this IProxyFactory proxyFactory, Func<object> instantiateFunc)
+		{
+			if (proxyFactory is StaticProxyFactory staticProxyFactory)
+			{
+				return staticProxyFactory.GetFieldInterceptionProxy();
+			}
+
+#pragma warning disable 618
+			return proxyFactory.GetFieldInterceptionProxy(instantiateFunc?.Invoke());
```
fredericDelaporte: I do not think this case is tested, especially when called by `PocoEntityInstantiator.CreateInstance`. It seems to me it can only result in a stack overflow.
Maybe an obsolete no-proxy test should be added with the dynamic proxy factory instead of the static proxy one.

fredericDelaporte:
>>  It should yield back what the user has assigned to it instead, for not breaking the unicity of the entity when working with a single session.
>
>I am ok with this behavior but we would still need to list this as a breaking change as currently the proxy will be initialized on set. This behavior could be achieved by checking if the session is set on the interceptor or not as the session is set in the `AfterInitialize` method which is after the 2-phase load.

As explained in [this comment](https://github.com/nhibernate/nhibernate-core/pull/1947#issuecomment-451758203), I do not think this PR should change the property behavior of yielding the unwrapped entity instead of the proxy assigned by the user to the property.

> Currently, NHibernate will initialize the entity on set as it does not return earlier when the `setter` argument of the `Intercept` method is `true` like this PR does. 

So, as you write, small breaking change: assigning an uninitialized proxy to a `no-proxy` property will no more trigger the proxy initialization. It should not be really a trouble because the main reason for doing such an assignment is to try avoiding loading the entity.

src/NHibernate.Test/LazyOneToOne/LazyOneToOneTest.cs
maca88: The problem here is that the current `IsPropertyInitialized` does not work correctly for this case as the `Employeee` is eager loaded when the `Person` is retrieved, but as the interceptor is currently injected after the entity initialization it doesn't know that the `Employee` was already initialized in the 2-phase load. If the `Employee` wasn't meant to be eager loaded (by using a join) when retrieving the `Persion`, then it should be treated as a different issue.

fredericDelaporte: Do you mean that for some reason, `no-proxy` on non-inverse `one-to-one` is broken since long and always eager load the employee anyway? (But an additional bug was causing this eager load to be "unseen" by the property.)

src/NHibernate.Test/GhostProperty/GhostPropertyFixture.cs
```diff
@@ -84,6 +89,49 @@ public void CanGetActualValueFromLazyManyToOne()
...
+		[Test]
+		public void SetInitializedProxyShouldNotResetPropertyInitialization()
+		{
+			using (var s = OpenSession())
+			{
+				var order = s.Get<Order>(1);
+				var payment = s.Load<Payment>(2);
+				Assert.That(order.Payment is WireTransfer, Is.True); // Load property
+				Assert.That(NHibernateUtil.IsPropertyInitialized(order, "Payment"), Is.True);
+				s.Get<Payment>(2); // Load the uninitialized payment
```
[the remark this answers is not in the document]
maca88: Done.

src/NHibernate/Proxy/StaticProxyFactory.cs
maca88: Done.

src/NHibernate/Proxy/IProxyFactory.cs
maca88: You are right, I've fixed it and added obsolete tests for dynamic proxies.

maca88:
> I do not think this PR should change the property behavior of yielding the unwrapped entity instead of the proxy assigned by the user to the property.

With the latest commit the interceptor was modified to yield a proxy instead of unwrapping it when it is set by the user.

src/NHibernate/Intercept/AbstractFieldInterceptor.cs
```diff
@@ -164,13 +165,22 @@ private bool IsUninitializedProperty(string fieldName)
...
 
-		private object InitializeOrGetAssociation(INHibernateProxy value, string fieldName)
+		private object InitializeOrGetAssociation(object target, INHibernateProxy value, string fieldName)
 		{
 			if(value.HibernateLazyInitializer.IsUninitialized)
 			{
 				value.HibernateLazyInitializer.Initialize();
 				value.HibernateLazyInitializer.Unwrap = true; // means that future Load/Get from the session will get the implementation
 				loadedUnwrapProxyFieldNames.Add(fieldName);
+				// Set the property value in order to be accessible when the session is closed
```
fredericDelaporte: I think this should be done out of the `if`. Otherwise, this test would fail:

```c#
[Test]
public void CanGetInitializedLazyManyToOneAfterClosedSession()
{
	Order order;
	Payment payment;

	using (var s = OpenSession())
	{
		payment = s.Load<Payment>(1);
		NHibernateUtil.Initialize(payment);
		order = s.Get<Order>(1);
		// Initialize Payment property
		payment = order.Payment;
	}

	Assert.That(order.Payment, Is.EqualTo(payment));
}
```

Maybe it should be put in directly in `Intercept`, in the `if (value.IsProxy() && IsUninitializedAssociation(fieldName))`.

src/NHibernate/Intercept/AbstractFieldInterceptor.cs
```diff
@@ -113,17 +114,22 @@ public object Intercept(object target, string fieldName, object value, bool sett
...
 				}
 
-				if (IsUninitializedAssociation(fieldName))
+				if (!unwrapProxyFieldNames.Contains(fieldName))
+				{
+					return value;
+				}
+
+				// When a proxy is set by the user which we know when the session is set, we should not unwrap it
+				if (session != null || !value.IsProxy() || NHibernateUtil.IsInitialized(value))
```
fredericDelaporte: Since on get, we assign from the proxy its unwrapped entity to the property in order to allow access after session closing, the `NHibernateUtil.IsInitialized(value)` condition should be removed in order to allow the unwrapping to take place even when the entity was proxy loaded prior to loading the parent entity having a `no-proxy` property on it.

src/NHibernate/Intercept/AbstractFieldInterceptor.cs
maca88: The assert won't fail as the objects are equal, but if the assert is modified to:
```
Assert.That(order.Payment is WireTransfer, Is.True);
```
then it would fail as the proxy was not unwrapped on get. By appling your suggestions the test will start working but there is another scenario that it would still fail. If in the above test the following line is removed:
```
// Initialize Payment property
payment = order.Payment;
```
the test would fail as the proxy was not unwrapped and set. This could be fixed by using the following code:
```
if (session != null || !value.IsProxy())
{
  loadedUnwrapProxyFieldNames.Add(fieldName);
}
else if (value is INHibernateProxy proxy)
{
  // Unwap the proxy if it is initialized
  var unwrappedProxy = UnwrapOrNull(proxy);
  if (unwrappedProxy != null)
  {
    loadedUnwrapProxyFieldNames.Add(fieldName);
    return unwrappedProxy;
  }
}
//...
private static object UnwrapOrNull(INHibernateProxy proxy)
{
  if (!proxy.HibernateLazyInitializer.IsUninitialized && proxy.HibernateLazyInitializer.Session != null)
  {
    return proxy.HibernateLazyInitializer.GetImplementation(proxy.HibernateLazyInitializer.Session);
  }

  return null;
}

```
The problem is that the current implementation for set interception looks like:
```
if (this.__fieldInterceptor != null)
{
  this.__fieldInterceptor.MarkDirty();
  this.__fieldInterceptor.Intercept(this, <ReflectHelper.GetPropertyName(setter)>, value, true);
}
base.<setter>(value);
``` 
which prevents us from overriding the value that will be set. If you agree I would change the implementation so that it would work similar like the current getter:
```
if (this.__fieldInterceptor != null)
{
  this.__fieldInterceptor.MarkDirty();
  var result = this.__fieldInterceptor.Intercept(this, <ReflectHelper.GetPropertyName(setter)>, value, true);
  if (result != AbstractFieldInterceptor.InvokeImplementation)
  {
    value = (<getter.ReturnType>)result;
  }
}
base.<setter>(value);
```
By doing this we would be able to set the unwrapped proxy when populating values in the 2-phase load and both test would pass.

maca88:
> For some reasons, dd9204b has broken the field-interceptor proxy serialization

Actually [this commit](https://github.com/nhibernate/nhibernate-core/pull/1947/commits/c3e6c70291f04f4448111faacea0845574c159a0) somehow broke the following test: `StaticProxyFixture.SessionSerializationWithLazyProperty`. The strange thing is that on my local machine this test passes with both `net461` and `netcoreapp2.0`, so I am currently not able to reproduce it in order to see what is wrong.

src/NHibernate/Intercept/AbstractFieldInterceptor.cs
fredericDelaporte: I do not think the second case should be supported. The lazy property has not been accessed at all before closing the session. Having a failure in such case, even if we could avoid it, is to be expected. Not failing in this second case could be a troublesome legacy for later changes.

maca88: > I do not think the second case should be supported.

I am ok with the decision as if we would support the second case we would also need to support the case when the `no-proxy` property is initialized after the parent is loaded to be consistent, example:
```
payment = s.Load<Payment>(1);
order = s.Get<Order>(1);
NHibernateUtil.Initialize(payment);
```
Which would be harder to implement as the parent would need to "subscribe" to the persistence context in order to initialize the `no-proxy` property once it is loaded.

src/NHibernate.Test/GhostProperty/GhostPropertyDynamicProxyFixture.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+	// Since 5.3
+	[Obsolete]
+	[TestFixture]
+	public class GhostPropertyDynamicProxyFixture : GhostPropertyFixture
+	{
+		protected override void Configure(Cfg.Configuration configuration)
+		{
+			base.Configure(configuration);
+			configuration.SetProperty(Cfg.Environment.ProxyFactoryFactoryClass, typeof(DefaultProxyFactoryFactory).FullName);
```
fredericDelaporte: For being future proof, better save here `Cfg.Environment.BytecodeProvider.ProxyFactoryFactory.GetType().FullName` in some field, and use that to restore it after the test.

PR #1948, opened 2018-12-26
Fix SELECT N+1 for not-found=ignore

lillo42:
Fixes: https://github.com/nhibernate/nhibernate-core/issues/1059


src/NHibernate/Type/ManyToOneType.cs
```diff
@@ -91,6 +91,12 @@ public override object Hydrate(DbDataReader rs, string[] names, ISessionImplemen
 			object id = GetIdentifierOrUniqueKeyType(session.Factory)
 				.NullSafeGet(rs, names, session, owner);
 			ScheduleBatchLoadIfNeeded(id, session);
+
+			if (id != null && !id.GetType().IsClass && IsAlwaysDirtyChecked)
```
hazzik: Need to have test with non value type identifiers (strings for example).

hazzik: If you change type of IDs in the test to string the test is failing.

hazzik:
@lillo42 if you add a following test that does not corrupt the employee table (without `ExecuteStatement`) it fails:

```csharp
[Test]
public void TestDepartmentIsNotNull()
{
	var statistics = Sfi.Statistics;
	statistics.Clear();

	{
		var employee = session.Get<Employee>(employeeId);
		Assert.That(employee.Department, Is.Not.Null); // <--- here
		Assert.That(statistics.PrepareStatementCount, Is.EqualTo(1));
		transaction.Commit();
	}
}
```

src/NHibernate/Type/EntityType.cs
```diff
@@ -87,6 +88,22 @@ public override bool IsEqual(object x, object y, ISessionFactoryImplementor fact
 
 		public virtual bool IsNull(object owner, ISessionImplementor session)
 		{
+			if (IsNullable && !string.IsNullOrEmpty(associatedEntityName))
+			{
+				EntityEntry entry = session.PersistenceContext.GetEntry(owner);
+				int position = -1;
+				for (int i = 0; i < entry.Persister.PropertyTypes.Length; i++)
+				{
+					if (entry.Persister.PropertyTypes[i].Name == associatedEntityName)
```
hazzik: This would not work if "owner" has more than one property referencing this entity.

src/NHibernate/Loader/Loader.cs
```diff
@@ -947,22 +947,47 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
+						{
+							if (i == j)
+							{
+								continue;
+							}
+
+							int count = -1;
+							for (int z = 0; z < persisters[j].PropertyTypes.Length; z++)
+							{
+								if (persisters[j].PropertyTypes[z].Name == persister.EntityName)
```
hazzik: Same as below.

src/NHibernate/Loader/Loader.cs
```diff
@@ -944,25 +944,68 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
+							if (i == j)
+							{
+								continue;
+							}
+
+							int position = -1;
+							int count = 0;
+							for (int z = 0; z < persisters[j].PropertyTypes.Length; z++)
+							{
+								if (persisters[j].PropertyTypes[z].Name == persister.EntityName && ++count == total)
```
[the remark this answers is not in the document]
bahusoid: Got a look inside. Actually it  seems I'm wrong and it should work.

src/NHibernate/Type/EntityType.cs
```diff
@@ -441,10 +442,7 @@ public string RHSUniqueKeyPropertyName
 			get { return uniqueKeyPropertyName; }
 		}
 
-		public virtual string PropertyName
-		{
-			get { return null; }
-		}
+		public virtual string PropertyName { get; set; }
```
bahusoid: This looks hacky. If property name is really required to be set in runtime - I think you should introduce instead some set method in `ManyToOne`

But it seems you should be able to set it similair way it's set for `OneToOneType`. Via the following chain: `HbmManyToOne.Name` -> `ManyToOne.PropertyName` (add property) -> `ManyToOneType.PropertyName` (add ctor parameter)

See `PropertiesBinder.BindManyToOne`  for a starting point.

src/NHibernate/Loader/Loader.cs
```diff
@@ -944,25 +944,68 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
 				{
-					// do nothing
-					/* TODO NH-1001 : if (persisters[i]...EntityType) is an OneToMany or a ManyToOne and
-					 * the keys.length > 1 and the relation IsIgnoreNotFound probably we are in presence of
-					 * an load with "outer join" the relation can be considerer loaded even if the key is null (mean not found)
-					*/
+					if (cols > 1)
+					{
+						/*
+						* the keys.length > 1 and the relation IsIgnoreNotFound probably we are in presence of
```
bahusoid: In original comment is also mentioned `OneToMany type. But I see only changes related to ManyToOne. OneToMany case still not fixed?

lillo42: I will add test for this case.

src/NHibernate/Loader/Loader.cs
```diff
@@ -944,25 +944,68 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
+				{
+					entityCounter[persister.EntityName]++;
+				}
+				else
+				{
+					entityCounter.Add(persister.EntityName, 1);
+				}
+
+				int total = entityCounter[persister.EntityName];
+
```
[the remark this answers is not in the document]
lillo42: Code simplified

src/NHibernate/Type/ManyToOneType.cs
```diff
@@ -19,20 +19,37 @@ public partial class ManyToOneType : EntityType
...
 
 		public ManyToOneType(string className, bool lazy)
 			: base(className, null, !lazy, false)
+		{
+			ignoreNotFound = false;
+			isLogicalOneToOne = false;
+			PropertyName = null;
+		}
+		
+		public ManyToOneType(string className, string propertyName, bool lazy)
```
bahusoid: Is this ctor used? `propertyName` param is not used

src/NHibernate/Mapping/OneToMany.cs
```diff
@@ -23,7 +23,7 @@ public OneToMany(PersistentClass owner)
 
 		private EntityType EntityType
 		{
-			get { return TypeFactory.ManyToOne(ReferencedEntityName, null, false, false, IsIgnoreNotFound, false); }
+			get { return TypeFactory.ManyToOne(ReferencedEntityName, null, false, false, IsIgnoreNotFound, false, null); }
```
bahusoid: This strange. Why `OneToMany` mapping returns `ManyToOne` entity type? If it's not a bug we need to add explcit comment about it. Looks very confusing.

src/NHibernate/Engine/EntityKey.cs
```diff
@@ -25,6 +25,7 @@ public EntityKey(object id, IEntityPersister persister)
 			identifier = id ?? throw new AssertionFailure("null identifier");
 			_persister = persister;
 			_hashCode = GenerateHashCode(persister, id);
+
```
bahusoid: You should get rid of such formatting only changes in this pr

src/NHibernate.Test/NHSpecificTest/NH1001/Mappings.hbm.xml
```diff
@@ -3,7 +3,7 @@
 				   namespace="NHibernate.Test.NHSpecificTest.NH1001"
 				   assembly="NHibernate.Test">
 
-	<class name="Department" table="DEPARTMENTS" lazy="false">
+	<class name="Department" table="DEPARTMENTS" lazy="false" entity-name="FooDepartment">
```
bahusoid: This `class` xml element needs to stay. I meant to add new xml element `class` with `entity-name`

src/NHibernate/Type/TypeFactory.cs
```diff
@@ -843,9 +843,9 @@ public static EntityType ManyToOne(string persistentClass, bool lazy)
 		/// <summary>
 		/// A many-to-one association type for the given class and cascade style.
 		/// </summary>
-		public static EntityType ManyToOne(string persistentClass, string uniqueKeyPropertyName, bool lazy, bool unwrapProxy, bool ignoreNotFound, bool isLogicalOneToOne)
+		public static EntityType ManyToOne(string persistentClass, string uniqueKeyPropertyName, bool lazy, bool unwrapProxy, bool ignoreNotFound, bool isLogicalOneToOne, string propertyName)
```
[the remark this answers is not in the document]
lillo42: Would be better if I create a new constructor?

src/NHibernate/Type/ManyToOneType.cs
lillo42: I forgot to remove this ctor, sorry

src/NHibernate/Type/ManyToOneType.cs
```diff
@@ -19,20 +19,37 @@ public partial class ManyToOneType : EntityType
...
+			: base(className, null, !lazy, false)
 		{
 			ignoreNotFound = false;
 			isLogicalOneToOne = false;
 		}
 
-		public ManyToOneType(string entityName, string uniqueKeyPropertyName, bool lazy, bool unwrapProxy, bool ignoreNotFound, bool isLogicalOneToOne)
+		public ManyToOneType(
```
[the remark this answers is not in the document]
lillo42: Ok, I will change

src/NHibernate/Type/TypeFactory.cs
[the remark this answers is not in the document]
lillo42: Should I create a new method?

src/NHibernate/Loader/Loader.cs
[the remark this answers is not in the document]
lillo42: Is there any way to get custom entity name from some entity or entity to custom entity name?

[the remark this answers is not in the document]
bahusoid: I'm not sure I understand the question. Do you want to retrieve entity name for the given entity instance?
```C#
var entity = session.Get("EntityName", id);
var entityName = GetEntityNameFromInstance(entity); // <-- Is it what are you asking for?
```

I don't know if it's possible. Where do you need such logic?
It's the main idea of custom entity name - it's possible to have the same class for different entities. So you can't obtain entity name by class type - you might have multiple entity names associated with this type.

src/NHibernate/Type/TypeFactory.cs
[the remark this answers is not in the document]
bahusoid: Yeah - same as with constructor

bahusoid:
@lillo42 Do you have any pending changes? It's too many commits I want to squash them all.

lillo42:
@bahusoid I have more one, this commit will add one to many case, but I changed many files.

``` git
modified:   src/NHibernate.Test/NHSpecificTest/NH1001/Employee.cs
	modified:   src/NHibernate.Test/NHSpecificTest/NH1001/Fixture.cs
	modified:   src/NHibernate.Test/NHSpecificTest/NH1001/Mappings.hbm.xml
	modified:   src/NHibernate.Test/NHSpecificTest/NH1001/Phone.cs
	modified:   src/NHibernate/Bytecode/ICollectionTypeFactory.cs
	modified:   src/NHibernate/Cfg/XmlHbmBinding/CollectionBinder.cs
	modified:   src/NHibernate/Loader/Loader.cs
	modified:   src/NHibernate/Mapping/Array.cs
	modified:   src/NHibernate/Mapping/Bag.cs
	modified:   src/NHibernate/Mapping/Collection.cs
	modified:   src/NHibernate/Mapping/IdentifierBag.cs
	modified:   src/NHibernate/Mapping/List.cs
	modified:   src/NHibernate/Mapping/Map.cs
	modified:   src/NHibernate/Mapping/Set.cs
	modified:   src/NHibernate/Type/ArrayType.cs
	modified:   src/NHibernate/Type/CollectionType.cs
	modified:   src/NHibernate/Type/CustomCollectionType.cs
	modified:   src/NHibernate/Type/DefaultCollectionTypeFactory.cs
	modified:   src/NHibernate/Type/GenericBagType.cs
	modified:   src/NHibernate/Type/GenericIdentifierBagType.cs
	modified:   src/NHibernate/Type/GenericListType.cs
	modified:   src/NHibernate/Type/GenericMapType.cs
	modified:   src/NHibernate/Type/GenericOrderedSetType.cs
	modified:   src/NHibernate/Type/GenericSetType.cs
	modified:   src/NHibernate/Type/GenericSortedDictionaryType.cs
	modified:   src/NHibernate/Type/GenericSortedListType.cs
	modified:   src/NHibernate/Type/GenericSortedSetType.cs
	modified:   src/NHibernate/Type/TypeFactory.cs
```

bahusoid:
Ok. Then will do it later. But could you please clarify what's the problem with one-to-many case? When I think about it I don't quite understand what could be wrong. Shouldn't entities  with not found Id be simply ignored and skipped from collections? It's not how it's done?

lillo42:
In Loader class I added this code

```csharp

if (type is CollectionType collection && collection.EntityName == persister.EntityName && ++count == total && collection.IsNullable)
{
    found = true;
    session.PersistenceContext.AddNullProperty(keys[j], collection.PropertyName);
    break;
}
```

but CollectionType doesn't have Properties:

```EntityName, PropertyName and IsNullable```

So I add this properties

**Comment by bahusoid** [comment]
I was asking not about the fix but how the issue manifests itself. Many-to-one case is clear to me:
```C#
var employee = session.Get<Employee>(employeeId);
```
So `employee.Department` isn't loaded to session and additional query is exectuted to check if it's exists.

But with one-to-many collections it's a bit different story. For bag/set mapping you specify the column on the other end (so for `employee.Phones`  there is `Phone.EmployeeId` column). And I don't quite understand what `not-found` means here. Query for phones should specifically load only phone records with given EmployeeId. What is "null" property here?

Well I can wait and see your tests.


`

lillo42:
@bahusoid I pushed the commits

src/NHibernate.Test/NHSpecificTest/NH1001/Fixture.cs
```diff
@@ -158,5 +179,26 @@ public void AddressNull()
...
+			statistics.Clear();
+
+			using (ISession session = OpenSession())
+			{
+				var employee = session.Get<Employee>(employeeId);
+				Assert.That(employee.Department1, Is.Not.Null);
+				Assert.That(employee.Department2, Is.Not.Null);
+				Assert.That(employee.Department3, Is.Not.Null);
+				Assert.That(employee.Address, Is.Not.Null);
+				Assert.That(employee.Phones, Is.Null);
```
bahusoid: I'm afraid but I don't think this is correct behavior. Why would you want you `Phones` collection be initialized with null? It should be simply empty collection. And I believe it's the way it's already working without any fixes.

As I already said I don't understand at all what's use for [`not-found`  in one-to-many](https://nhibernate.info/doc/nhibernate-reference/collections.html#collections-onetomany) association
@hazzik , @fredericDelaporte  Can you shed some light on it?

src/NHibernate/Loader/Loader.cs
```diff
@@ -944,25 +944,59 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
+							{
+								continue;
+							}
+
+							bool found = false;
+							int count = 0;
+							for (int z = 0; z < persisters[j].PropertyTypes.Length; z++)
+							{
+								IType type = persisters[j].PropertyTypes[z];
+								if (type.Name == persister.EntityName && ++count == total
```
bahusoid: I'm afraid but we can't detect owner by order of persisters in result as order can be changed in query.
Here is the test that breaks this check:
```C#
[Test]
public void Department2IsNull_QueryOver()
{
	ExecuteStatement($"UPDATE EMPLOYEES SET DEPARTMENT_ID_1 = 11, DEPARTMENT_ID_2 = 99999 WHERE EMPLOYEE_ID = {employeeId}");

	IStatistics statistics = Sfi.Statistics;
	statistics.Clear();

	{
		var employee = session.QueryOver<Employee>()
			.Fetch(SelectMode.Fetch, e => e.Department2)
			.Fetch(SelectMode.Skip, e => e.Department1, e => e.Department3, e => e.Address)
			.Where(e => e.Id == employeeId)
			.SingleOrDefault();

		Assert.That(employee.Department1, Is.Not.Null);
		Assert.That(employee.Department2, Is.Null);
		Assert.That(employee.Department3, Is.Not.Null);
		Assert.That(employee.Address, Is.Not.Null);
		Assert.That(statistics.PrepareStatementCount, Is.EqualTo(1));
	}
}
```
We need some other way to properly find owner property.

src/NHibernate/Loader/Loader.cs
```diff
@@ -944,25 +944,59 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
-					 * the keys.length > 1 and the relation IsIgnoreNotFound probably we are in presence of
-					 * an load with "outer join" the relation can be considerer loaded even if the key is null (mean not found)
-					*/
+					if (cols > 1)
+					{
+						/*
+						* the keys.length > 1 and the relation IsIgnoreNotFound probably we are in presence of
+						* an load with "outer join" the relation can be considerer loaded even if the key is null (mean not found)
+						*/
+						for (int j = 0; j < persisters.Length; j++)
```
lillo42: Should I change ```persisters``` to ```CollectionOwners```?

src/NHibernate/Loader/Loader.cs
```diff
@@ -980,22 +981,63 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
-					// do nothing
-					/* TODO NH-1001 : if (persisters[i]...EntityType) is an OneToMany or a ManyToOne and
-					 * the keys.length > 1 and the relation IsIgnoreNotFound probably we are in presence of
-					 * an load with "outer join" the relation can be considerer loaded even if the key is null (mean not found)
+					if (cols == 1 || CollectionOwners == null)
+					{
+						continue;
+					}
+
+					int position = GetPosition(i);
```
hazzik: I suspect that `position` always equals to `i+1`.

hazzik: Ok, except QueryOver trick with skip...

src/NHibernate.Test/NHSpecificTest/NH1001/Fixture.cs
```diff
@@ -12,54 +12,217 @@ protected override void Configure(Configuration configuration)
...
+				var employee = session.QueryOver<Employee>()
+					.Fetch(SelectMode.Fetch, e => e.Department2)
+					.Fetch(SelectMode.Skip, e => e.Department1, e => e.Department3, e => e.Address)
+					.Where(e => e.Id == employeeId)
+					.SingleOrDefault();
+
+				Assert.That(employee.Department1, Is.Not.Null);
+				Assert.That(employee.Department2, Is.Null);
+				Assert.That(employee.Department3, Is.Not.Null);
+				Assert.That(employee.Address, Is.Not.Null);
```
hazzik: In this test, how many statements do we expect to be executed? /cc @bahusoid

src/NHibernate/Loader/Loader.cs
```diff
@@ -980,22 +981,63 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
+						continue;
+					}
+					
+					/*
+					* the keys.length > 1 and the relation IsIgnoreNotFound probably we are in presence of
+					* an load with "outer join" the relation can be considerer loaded even if the key is null (mean not found)
 					*/
+					
+					
+					foreach (int owner in CollectionOwners)
```
hazzik: `CollectionOwners` is not empty in the current tests because Employee has a "Phones" collection.

src/NHibernate/Loader/Loader.cs
bahusoid: My main point was to provide example when order of entities in result set is different from mapping (so mapping based index check is wrong).  IMHO number of queries are not relevant - but I can double check later if you think it's useful (just let me know).
Theoretically this fix should save 1 additional request for checking Department2 but all other `not-found` ignore properties should still trigger requests.
So I think 4 statements in total : 1 for main query + 3 for Department1, Department3, Address "not found" ignore check.

And regarding current fix. This Aliases parse logic doesn't feel reliable and looks like depends on some internal knowledge of how aliases are generated. I would avoid such dependencies.

src/NHibernate.Test/NHSpecificTest/NH1001/Fixture.cs
bahusoid: My main point was to provide example when order of entities in result set is different from mapping (so mapping based index check is wrong).  IMHO number of queries are not relevant - but I can double check later if you think it's useful (just let me know).
Theoretically this fix should save 1 additional request for checking Department2 but all other `not-found` ignore properties should still trigger requests.
So I think 4 statements in total : 1 for main query + 3 for Department1, Department3, Address "not found" ignore check.

And regarding current fix. This Aliases parse logic doesn't feel reliable and looks like depends on some internal knowledge of how aliases are generated. I would avoid such dependencies.

src/NHibernate.Test/NHSpecificTest/NH1001/Fixture.cs
```diff
@@ -207,19 +205,44 @@ public void Department2IsNull_QueryOver()
...
+
+			using (var session = OpenSession())
+			{
+				var employee = session.QueryOver<Employee>()
+					.Fetch(SelectMode.Fetch, e => e.Department2)
+					.Fetch(SelectMode.Skip, e => e.Department1, e => e.Department3, e => e.Address, e => e.Phones)
+					.Where(e => e.Id == employeeId)
+					.SingleOrDefault();
 				Assert.That(employee.Department2, Is.Null);
+				Assert.That(statistics.PrepareStatementCount, Is.EqualTo(5), "Employee, Department1, Department3, Address, and Phones");
```
bahusoid: Are you sure regarding Phones? Additional query is required only for "single property" mappings (like many-to-one)  - because `not-found` property needs to be initialized with null in case not found case.  `Phones` shouldn't trigger "not found" check statement.  That's something I was trying to discuss [here](https://github.com/nhibernate/nhibernate-core/pull/1948#discussion_r249247393)

So I would say it should be 4

hazzik: Yes, I'm sure about phones. It is mapped not lazily.

hazzik: The queries generated (except for `Department2`) not because of "not found", but because they are not lazy.

bahusoid: No.. `SelecMode.Skip` overrides mapping. That's the whole point of using it. You've made `Phones` lazy in query.

hazzik: These are the queries generated
```sql
-- 1
    SELECT
        this_.EMPLOYEE_ID as employee1_3_1_,
        this_.FIRST_NAME as first2_3_1_,
        this_.LAST_NAME as last3_3_1_,
        this_.DEPARTMENT_ID_1 as department4_3_1_,
        this_.DEPARTMENT_ID_2 as department5_3_1_,
        this_.DEPARTMENT_ID_3 as department6_3_1_,
        this_.ADDRESS_ID as address7_3_1_,
        department2_.DEPARTMENT_ID as department1_0_0_,
        department2_.DEPARTMENT_NAME as department2_0_0_ 
    FROM
        EMPLOYEES this_ 
    left outer join
        DEPARTMENTS department2_ 
            on this_.DEPARTMENT_ID_2=department2_.DEPARTMENT_ID 
    WHERE
        this_.EMPLOYEE_ID = @p0;
    @p0 = 1 [Type: Int32 (0:0:0)]
--2 Department1
    SELECT
        department0_.DEPARTMENT_ID as department1_0_0_,
        department0_.DEPARTMENT_NAME as department2_0_0_ 
    FROM
        DEPARTMENTS department0_ 
    WHERE
        department0_.DEPARTMENT_ID=@p0;
    @p0 = 11 [Type: Int32 (0:0:0)]
-- 3 Department2 - unwanted 
    SELECT
        department0_.DEPARTMENT_ID as department1_0_0_,
        department0_.DEPARTMENT_NAME as department2_0_0_ 
    FROM
        DEPARTMENTS department0_ 
    WHERE
        department0_.DEPARTMENT_ID=@p0;
    @p0 = 99999 [Type: Int32 (0:0:0)]
-- 4 Department 3 
    SELECT
        department0_.DEPARTMENT_ID as department1_0_0_,
        department0_.DEPARTMENT_NAME as department2_0_0_ 
    FROM
        DEPARTMENTS department0_ 
    WHERE
        department0_.DEPARTMENT_ID=@p0;
    @p0 = 13 [Type: Int32 (0:0:0)]
-- 5 Address  
    SELECT
        addressemp0_.ADDRESS_ID as address1_1_0_,
        addressemp0_.ADDRESS_NAME as address2_1_0_ 
    FROM
        ADDRESSES addressemp0_ 
    WHERE
        addressemp0_.ADDRESS_ID=@p0;
    @p0 = 15 [Type: Int32 (0:0:0)]
-- 6 Phones 
    SELECT
        phones0_.EMPLOYEE_ID as employee3_2_1_,
        phones0_.PHONE_ID as phone1_2_1_,
        phones0_.PHONE_ID as phone1_2_0_,
        phones0_.PHONE_NUMBER as phone2_2_0_ 
    FROM
        PHONES phones0_ 
    WHERE
        phones0_.EMPLOYEE_ID=@p0;
    @p0 = 1 [Type: Int32 (0:0:0)]
```

If phones is not skipped then last query is not generated but embedded in the first query.

bahusoid: Well. That doesn't look right to me. Maybe something in mapping (like lazy="false" for Employee) triggers such behavior or it's some issue with SelecMode.Skip. But it's off topic.

hazzik:
Ok, I've found the super easy fix. I've just removed restrictions on the IsOneToOne... And that's it. I do not really know why this restriction is there, or who put them there.

hazzik:
I've done regression test on `not-found="exception"` and it seems to be working as expected.

lillo42:
@hazzik Thank you, it's very easy way to fix

src/NHibernate.Test/NHSpecificTest/NH1001/Fixture.cs
bahusoid: Ah.. Indeed all classes and `Phones` property is mapped with `lazy="false"`. 
But `SkipMode` overrides only fetch mode (which is `fetch="join"` in our case).

src/NHibernate/Cfg/XmlHbmBinding/CollectionBinder.cs
```diff
@@ -192,6 +192,7 @@ private void BindPrimitiveArray(HbmPrimitiveArray arrayMapping, PrimitiveArray m
 				var oneToMany = new OneToMany(model.Owner);
 				model.Element = oneToMany;
 				BindOneToMany(oneToManyMapping, oneToMany);
+				oneToMany.PropertyName = collectionMapping.Name;
```
bahusoid: This fix doesn't affect `one-to-many` case so I think we should undo this change

src/NHibernate/Mapping/OneToMany.cs
```diff
@@ -59,6 +59,8 @@ public IType Type
 		{
 			get { return EntityType; }
 		}
+		
+		public string PropertyName { get; set; }
```
bahusoid: I think this new property should be removed as we don't need collection property name - we should just supply null to `EntityType` here

src/NHibernate.sln.DotSettings
```diff
@@ -410,4 +410,4 @@
 	<s:String x:Key="/Default/PatternsAndTemplates/StructuralSearch/Pattern/=EDE6DA5BB1FD144F81E051772A257749/ReplaceComment/@EntryValue">Update to fluent syntax.</s:String>
 	<s:String x:Key="/Default/PatternsAndTemplates/StructuralSearch/Pattern/=EDE6DA5BB1FD144F81E051772A257749/ReplacePattern/@EntryValue">Assert.That($greater$, Is.GreaterThanOrEqualTo($lesser$));</s:String>
 	<s:String x:Key="/Default/PatternsAndTemplates/StructuralSearch/Pattern/=EDE6DA5BB1FD144F81E051772A257749/SearchPattern/@EntryValue">Assert.GreaterOrEqual($greater$, $lesser$);</s:String>
-	<s:String x:Key="/Default/PatternsAndTemplates/StructuralSearch/Pattern/=EDE6DA5BB1FD144F81E051772A257749/Severity/@EntryValue">HINT</s:String></wpf:ResourceDictionary>
\ No newline at end of file
+	<s:String x:Key="/Default/PatternsAndTemplates/StructuralSearch/Pattern/=EDE6DA5BB1FD144F81E051772A257749/Severity/@EntryValue">HINT</s:String></wpf:ResourceDictionary>
```
bahusoid: Better to undo

src/NHibernate/Type/CollectionType.cs
```diff
@@ -673,4 +673,4 @@ protected internal virtual string RenderLoggableString(object value, ISessionFac
 			return CollectionPrinter.ToString(list);
 		}
 	}
-}
+}
```
bahusoid: Better to undo

src/NHibernate/Type/GenericOrderedSetType.cs
```diff
@@ -23,7 +23,6 @@ public GenericOrderedSetType(string role, string propertyRef)
 			: base(role, propertyRef)
 		{
 		}
-
```
bahusoid: Better to undo

hazzik:
>async regen is missing

Fixed

src/NHibernate/Loader/Loader.cs
```diff
@@ -778,23 +778,25 @@ protected virtual object[] GetResultRow(Object[] row, DbDataReader rs, ISessionI
 		/// </summary>
 		private void RegisterNonExists(EntityKey[] keys, ISessionImplementor session)
 		{
-			int[] owners = Owners;
-			if (owners != null)
+			var owners = Owners;
```
bahusoid: >For missing objects associated by one-to-one with...

Comment for method should be adjusted as it's not one-to-one specific anymore.

Same applies for protected `Owners` property:
>An array of indexes of the entity that owns a one-to-one association

For one-to-one - null key is possible only for not-found ignore mapping, but for many-to-one we may have many valid null keys. So I it would be good to improve this place to call AddNullProperty only if association is mapped as not-found ignore. Just a thought..

src/NHibernate/Loader/Loader.cs
```diff
@@ -778,23 +778,25 @@ protected virtual object[] GetResultRow(Object[] row, DbDataReader rs, ISessionI
...
-														 && ownerAssociationTypes[i].IsOneToOne;
-							if (isOneToOneAssociation)
+							int owner = owners[i];
+							if (owner > -1)
 							{
-								session.PersistenceContext.AddNullProperty(ownerKey, ownerAssociationTypes[i].PropertyName);
+								var ownerKey = keys[owner];
+								if (ownerKey != null)
+								{
+									session.PersistenceContext.AddNullProperty(ownerKey, ownerAssociationType.PropertyName);
```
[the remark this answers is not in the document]
hazzik: It could be, but I doubt it would be the case. Need a test.

hazzik:
I'm going to squash commits.

src/NHibernate/Loader/Loader.cs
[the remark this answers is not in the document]
hazzik: I've added a null check.

src/NHibernate/Loader/Loader.cs
hazzik: I've updated comments.

>it would be good to improve this place to call AddNullProperty only if association is mapped as not-found ignore. Just a thought..

It actually does not matter. I think it saves some calls in `ResolveIdentifier`. However, I think we need to add some property-ref tests.

bahusoid: >It actually does not matter.

I meant only wasting memory. This check would save some.

src/NHibernate/Loader/Loader.cs
```diff
@@ -982,15 +984,8 @@ private void CheckVersion(int i, IEntityPersister persister, object id, object e
...
-				if (key == null)
-				{
-					// do nothing
-					/* TODO NH-1001 : if (persisters[i]...EntityType) is an OneToMany or a ManyToOne and
-					 * the keys.length > 1 and the relation IsIgnoreNotFound probably we are in presence of
-					 * an load with "outer join" the relation can be considerer loaded even if the key is null (mean not found)
-					*/
-				}
-				else
+				// if key is null do nothing
```
bahusoid: Maybe instead of commenting the obvious we should clarify where null key is handled. Something like:
```suggestion
				// null keys are handled in RegisterNonExists
```

src/NHibernate/Loader/Loader.cs
bahusoid: Actually it seems check to avoid unnecessary filling dictionary with null keys is already available via `ownerAssociationType.IsNullable`. I think we should add it here

src/NHibernate/Loader/Loader.cs
```diff
@@ -772,29 +772,31 @@ protected virtual object[] GetResultRow(Object[] row, DbDataReader rs, ISessionI
...
+				for (var i = 0; i < keys.Length; i++)
 				{
-					int owner = owners[i];
-					if (owner > -1)
+					if (keys[i] == null)
 					{
-						EntityKey ownerKey = keys[owner];
-						if (keys[i] == null && ownerKey != null)
+						var ownerAssociationType = ownerAssociationTypes[i];
+						if (ownerAssociationType?.PropertyName != null)
```
bahusoid:
```suggestion
						if (ownerAssociationType?.PropertyName != null && ownerAssociationType.IsNullable)
```

src/NHibernate/Async/Loader/Loader.cs
```diff
@@ -625,15 +625,8 @@ private async Task CheckVersionAsync(int i, IEntityPersister persister, object i
...
-				if (key == null)
-				{
-					// do nothing
-					/* TODO NH-1001 : if (persisters[i]...EntityType) is an OneToMany or a ManyToOne and
-					 * the keys.length > 1 and the relation IsIgnoreNotFound probably we are in presence of
-					 * an load with "outer join" the relation can be considerer loaded even if the key is null (mean not found)
-					*/
-				}
-				else
+				// if key is null do nothing
```
bahusoid:
```suggestion
				// null keys are handled in RegisterNonExists
```

hazzik:
Need to also have a test for property-ref (by unique key instead of PK)

hazzik:
There is possible regression for property-ref not found=exception case.

hazzik:
Ok, this is not regression. Checked and it did not work all the way down to 4.0.4.

src/NHibernate.Test/NHSpecificTest/NH1001/Fixture.cs
```diff
@@ -191,15 +218,50 @@ public void Department3IsNotFound()
...
+				Assert.That(employee.Department3, Is.Not.Null);
+				Assert.That(employee.Department4, Is.Not.Null);
+				Assert.That(employee.Department5, Is.Null);
+				Assert.That(employee.Address, Is.Not.Null);
+				Assert.That(statistics.PrepareStatementCount, Is.EqualTo(1));
+			}
+		}
+
+		[Test, Ignore("Not fixed yet")]
+		public void Department5IsNotFound()
```
hazzik: This is not a regression here. It does not work all the way down to 4.0.4.

src/NHibernate/Type/ManyToOneType.cs
```diff
@@ -144,6 +157,18 @@ private bool IsIdentifier(object value, ISessionImplementor session)
 			return value.GetType() == identifierType.ReturnedClass;
 		}
 
+		public override bool IsNull(object owner, ISessionImplementor session)
+		{
+			if (IsNullable && !string.IsNullOrEmpty(PropertyName))
+			{
+				EntityEntry entry = session.PersistenceContext.GetEntry(owner);
+
+				return session.PersistenceContext.IsPropertyNull(entry.EntityKey, PropertyName);
```
fredericDelaporte, line 166: This causes #2627 regression. During a merge, the owner does not belong to the session, causing `GetEntry` to yield `null`.

The `OneToOne` type has a different implementation which does not rely on the owner belonging to session, using a list of known `null` associations, which this PR seems to update with `not-found` cases. So why not having used the `OneToOne` implementation here?

https://github.com/nhibernate/nhibernate-core/blob/33f90d3742f1a2d5a2aa2368edbaf667663fae1a/src/NHibernate/Type/OneToOneType.cs#L77-L85

Issue #1950, opened 2018-12-28
Caching of Property Dialect in class Configuration.Mapping


ValResnick:
This Property should be cached, because the creation of Dialect can be expensive. 

This is our scenario: We have a base class with a lot of subclasses (over 400).

The class `UnionSubclassEntityPersister` creates for our base class  a very long (over 20 Million characters) SubQuery in `GenerateSubquery`. 

```c#
// from UnionSubclassEntityPersister
subquery = GenerateSubquery(persistentClass, mapping);
```

Everytime `GetSqlTypeCode` is called for an instance of `NullableType` the Dialect will be created again.

```c#

// from UnionSubclassEntityPersister.GenerateSubquery
var sqlType = col.GetSqlTypeCode(mapping);

```
```c#
// from NullableType
static SqlType OverrideSqlType(IMapping mapping, SqlType type)
{
   return mapping != null ? mapping.Dialect.OverrideSqlType(type) : type;
}
```

This is a possible solution for the problem.

```c#
[Serializable]
private class Mapping : IMapping
{
   private readonly Configuration configuration;

   [NonSerialized]
   private readonly Lazy<Dialect.Dialect> _lazyDialect;

   public Mapping(Configuration configuration)
   {
      this.configuration = configuration;
				
      _lazyDialect = new Lazy<Dialect.Dialect>(() => NHibernate.Dialect.Dialect.GetDialect(this.configuration.Properties));
   }
   
   ...

   public Dialect.Dialect Dialect => _lazyDialect.Value;
}
```

fredericDelaporte:
Have you checked there is still the issue with NHibernate v5.2? There have been some changes done on that precise subject. See #1703.

ValResnick:
Today i tried to upgrade our project to V5.2.1. That's the reason i have created this issue :-)

fredericDelaporte:
Ok, it is already lazy in [`Mappings`](https://github.com/nhibernate/nhibernate-core/blob/ce73fc08719d5f8494d9755d94e0ef88f1de3ece/src/NHibernate/Cfg/Mappings.cs#L92), but the trouble is in [`Mapping`](https://github.com/nhibernate/nhibernate-core/blob/ce73fc08719d5f8494d9755d94e0ef88f1de3ece/src/NHibernate/Cfg/Configuration.cs#L187).

This dates back to 1464f48, #703, v5.0.

fredericDelaporte:
[...]

PR #1951, opened 2018-12-28
Cache Dialect when creating a session factory

maca88:
Fixes: #1950

This is a quick fix for the issue as a proper solution would probably require to refactor the logic for creating a session factory.

### DIFF
--- a/src/NHibernate/Cfg/Configuration.cs
+++ b/src/NHibernate/Cfg/Configuration.cs
@@ -239,6 +239,40 @@ namespace NHibernate.Cfg
 				NHibernate.Dialect.Dialect.GetDialect(configuration.Properties);
 		}
 
+		[Serializable]
+		private class StaticDialectMappingWrapper : IMapping
+		{
+			private readonly IMapping _mapping;
+
+			public StaticDialectMappingWrapper(IMapping mapping)
+			{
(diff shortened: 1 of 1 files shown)


fredericDelaporte:
This fix the session factory building case, but maybe there are other cases which may suffer of re-creating the dialect.
I was thinking about doing something more wide but also a bit more hackish: an internal lazy property on `Mapping`, but reinitialized on some `Configuration` method calls like `SecondPassCompile`.

maca88:
I've tried to move the `Dialect` property from `IMapping` to `ISessionFactoryImplementor` and the only usage that was problematic is the one from the issue. The `IMapping` implementation in `SessionFactoryImpl` uses the dialect from the `Settings` class that has it stored in a property. The only recreation that I additionally found can happen at:
https://github.com/nhibernate/nhibernate-core/blob/ce73fc08719d5f8494d9755d94e0ef88f1de3ece/src/NHibernate/Cfg/Configuration.cs#L1709
Where both `BuildSettings` and `GetDerivedProperties` use the `Dialect.GetDialect` method. As those methods are called only once per session factory I don't think that they are problematic.

fredericDelaporte:
Back-ported to 5.2.4.

PR #1952, opened 2019-01-02
Add support for caching fetched relations with linq query provider

### LINKED ISSUE #1195: NH-4078 - LINQ fetched collections aren't cached
**Tahir K Ahmadov** created an [issue](https://nhibernate.jira.com/browse/NH-4078) — <time datetime="2017-09-12T15:20:16.392Z" title="2017-09-12T15:20:16.392Z">12th September 2017, 15:20:16</time>:
> Config:
> 
> 
> ```c#
> var automappingConfiguration = new AutomappingConfiguration();
> return Fluently.Configure()
>     .Database(MsSqlConfiguration.MsSql2012.ConnectionString(
>         ConfigurationManager.ConnectionStrings<"Main">.ConnectionString))
>     .Cache(c => c.UseQueryCache().UseSecondLevelCache().ProviderClass<SysCacheProvider>())
>     .Mappings(m =>
>         m.AutoMappings.Add(
>             AutoMap.AssemblyOf<User>(automappingConfiguration)
>                 .IgnoreBase<EntityBase>()
>                 .Conventions
>                 .Setup(c =>
>                 {
>                     c.Add(ForeignKey.Format((member, type) =>
>                     {
>                         // standard foreign key, straight forward
>                         if (member != null)
>                             return member.Name <ins> "Id";
> 
>                         // member is null, because it's trying to find the "reverse" column name for a collection
>                         // it doesn't find the "reverse" property when it exists, so we find it ourselves
>                         var properties = type.GetProperties();
>                         var collectionProperties = properties.Where(p => p.PropertyType.IsGenericType && new[] { typeof(ISet<>), typeof(IList<>) }.Contains(p.PropertyType.GetGenericTypeDefinition()));
>                         // unfortunately, we don't know which collection property it is and have to guess and go with the first one
>                         var firstCollectionProperty = collectionProperties.First();
>                         var collectionItem = firstCollectionProperty.PropertyType.GetGenericArguments()<0>;
>                         var reversePropertyInfo = collectionItem.GetProperties().First(p => p.PropertyType == type);
>                         return reversePropertyInfo.Name </ins> "Id";
>                     }));
>                     c.Add<EnumConvention>();
>                     c.Add(Cache.Is(ci => ci.ReadWrite()));
>                 })
>             )
>     )
>     .BuildSessionFactory();
> 
> ```
> 
> 
> 
> 
> Query:
> 
> ```c#
> session.Query<User>()
>     .FetchMany(u => u.DataSources).ThenFetchMany(uds => uds.ParameterLimits).ThenFetchMany(udsp => udsp.Options).ThenFetch(udspo => udspo.Option)
>     .FetchMany(u => u.DataSources).ThenFetchMany(uds => uds.ParameterLastValues).ThenFetchMany(udsplv => udsplv.Options).ThenFetch(o => o.Option)
>     .Where(u => u.Email == email).ToList().First()
> 
> ```
> 
> 
> Cold cache - correct operation (almost: I have to run around and dedupe some collections, and First() doesn't work - it limits all collections to 1 item only, too - have to do .Where(...).ToList().First(); perhaps I should create another bug for these).
> 
> Warm cache - user is retrieved from cache, but then all other collections are retrieved as if lazy loaded:
> 
> 
> 
> 
> ```sql
> 
> exec sp*executesql N'SELECT datasource0_.UserId as UserId2_2_1_, datasource0_.Id as Id1_2_1_, datasource0_.Id as Id1_2_0_, datasource0_.UserId as UserId2_2_0_, datasource0_.DataSourceId as DataSo3_2_0_, datasource0_.LastGraphModuleId as LastGr4_2_0_, datasource0_.LastGridModuleId as LastGr5_2_0_ FROM <UserDataSource> datasource0_ WHERE datasource0*.UserId=@p0',N'@p0 int',@p0=1
> exec sp*executesql N'SELECT parameterl0_.UserDataSourceId as UserDa2_5_1_, parameterl0_.Id as Id1_5_1_, parameterl0_.Id as Id1_5_0_, parameterl0_.UserDataSourceId as UserDa2_5_0_, parameterl0_.DataSourceParameterId as DataSo3_5_0_ FROM [UserDataSourceParameterLimit] parameterl0_ WHERE parameterl0*.UserDataSourceId=@p0',N'@p0 int',@p0=1
> 
> ```
> 
> 
> 
---
**Frédéric Delaporte** added a comment — <time datetime="2017-09-12T16:48:43.211Z" title="2017-09-12T16:48:43.211Z">12th September 2017, 16:48:43</time>:
> > (almost: I have to run around and dedupe some collections, and First() doesn't work - it limits all collections to 1 item only, too - have to do .Where(...).ToList().First(); perhaps I should create another bug for these)
> 
> Cartesian products occur when fetching multiple collections (many `FetchMany`): do not fetch multiple collections. This is an old known limitation of fetching with NHibernate.
> 
> `First` applies to the whole query and as such does not actually support fetching collections.
> 
> Collection fetching support in NHibernate is far from some other ORM (which do instead generate quite complex queries for supporting it better, sometime at the expense of generating badly performing queries). But I suspect this is partly because NHibernate provides other efficient ways of avoiding N+1 fetching issues. You may read my StackOverflow answer on those other ways [here](https://stackoverflow.com/a/36070727/1178314).
> 
> Regarding the bug you report, please provide a simple test case demonstrating this. Are you sure your current auto-mapping correctly flag each collection and each involved entity as cacheable? (I am not used to auto-mapping.)
> 
> It looks to me it is well likely to be a "work as intended" case. Eager fetching do not use cache. Eager fetching always fetch. Eager fetching can not guess if the collection it will have to fetch are already sitting in the cache or not. It needs to have retrieved the queried root entities list first, but that is in contradiction with its contract, which is to retrieve the fetched collections/entities altogether. Added to the fact that it has currently limitations such as duplicating results in case of multiple collection fetching, I would not be surprised if putting its result to cache for fetched collections was disabled. (I would have to check that point.) 
> 
> Switching to minor.
> 
---
**Tahir K Ahmadov** added a comment — <time datetime="2017-09-12T17:11:26.621Z" title="2017-09-12T17:11:26.621Z">12th September 2017, 17:11:26</time>:
> Regarding the .First() limitation, I think there is quite a big design flaw: instead of supporting it properly (which may or may not be desirable from performance perspective as you mentioned) OR refusing to execute it with an exception, NH just returns some result which is incorrect. The standard here is how would an IEnumerable.First() call work; obviously NH's First() behavior is different, therefore wrong. It should just throw and exception and explain to the user that it's not supported and why/etc.
> Similarly, when I'm using the .Where().ToList().First() workaround, collections getting multiple items is incorrect behavior; if NH cannot fetch multiple collections in one LINQ query, it should throw an exception, not quietly produce incorrect results.
> Regarding my bug, yes all entities are cached (because I don't have any special logic for User class as opposed to all of the other classes). I'll try to put together a test case sometime today or tomorrow. 
> Overall, unfortunately it seems that the collection and caching features are a long way from being production ready. I don't want to come off as being too critical but I downloaded NH with the expectation that a lot of these scenarios have been worked out over the 10+ years of its existence, yet we're encountering big limitations in very commonly used, basic features, such as collections and caching.
> 
---
**Frédéric Delaporte** added a comment — <time datetime="2017-09-12T17:53:47.374Z" title="2017-09-12T17:53:47.374Z">12th September 2017, 17:53:47</time>:
> No, Linq-to-Object is not a standard. Better not assume other Linq-to-xxx will behave like Linq-to-object. Anyway the `First ` case is already reported as NH-3295 which is somewhat almost a dup of NH-3204, which just confirms that current NHibernate does not support mixing paging and fetching collections. (`First` is just a special case of paging.)
> 
> About your rant on NH, as I have tried to explain you, caching and fetching does not play well together. This is a bit contradictory to mix them. Normal use case is caching with lazy-loading. Are you experiencing issues when using caching and lazy-loading? (You may, since there is a number of errors a user can do. Such as not using transactions: cache is disabled when not using transaction.) But this is not the right place here for discussing all that. If you need support with caching, there is the [nh-users group](https://groups.google.com/forum/#!forum/nhusers) or [nhibernate.info](https://nhibernate.info) and the [reference documentation](https://nhibernate.info/doc/nhibernate-reference/index.html), or StackOverflow. If you want to discuss design, there is the [nh-development](https://groups.google.com/forum/#!forum/nhibernate-development) group.
>

maca88:
This PR fixes #1195 as it adds the ability to cache all fetched relations (many-to-one/one-to-many). The solution uses a special query cache result builder that appends extra values when creating the cache list that is then passed to the `IBatchableQueryCache.(Put\PutMany)` method. When fetching a many-to-one relation an additional value is added to the row of the cache list which is the id of the fetched relation. When fetching a one-to-many relation two values are added, the first is the id of the fetched element and the second is the collection id, which is the id of the entity that contains that collection. When the cached list is retrieved form the cache, the query cache will assembly and initialize all objects from the cached query and afterwards a special method is called that will remove the extra values that were added by the result builder.

### DIFF
--- a/src/AsyncGenerator.yml
+++ b/src/AsyncGenerator.yml
@@ -104,9 +104,6 @@
     - conversion: Ignore
       name: GetEnumerator
       containingTypeName: IFutureEnumerable
-    # 6.0 TODO: Remove QueryCacheResultBuilder from ignore
-    - conversion: Ignore
-      containingTypeName: QueryCacheResultBuilder
     - conversion: ToAsync
       name: ExecuteReader
       containingTypeName: IBatcher
--- a/src/AsyncGenerator.yml
+++ b/src/AsyncGenerator.yml
@@ -104,6 +104,9 @@
     - conversion: Ignore
       name: GetEnumerator
       containingTypeName: IFutureEnumerable
+    # 6.0 TODO: Remove QueryCacheResultBuilder from ignore
+    - conversion: Ignore
+      containingTypeName: QueryCacheResultBuilder
     - conversion: ToAsync
       name: ExecuteReader
       containingTypeName: IBatcher
--- a/src/NHibernate/Cache/QueryCacheResultBuilder.cs
+++ b/src/NHibernate/Cache/QueryCacheResultBuilder.cs
@@ -27,20 +27,19 @@ namespace NHibernate.Cache
 			_resultTypes = loader.ResultTypes;
 			_cacheTypes = loader.CacheTypes;
 
-			if (loader.EntityFetches == null)
+			if (loader.EntityFetches != null)
 			{
-				return;
-			}
-
-			for (var i = 0; i < loader.EntityFetches.Length; i++)
-			{
-				if (loader.EntityFetches[i])
+				for (var i = 0; i < loader.EntityFetches.Length; i++)
 				{
-					_entityFetchIndexes.Add(i);
+					if (loader.EntityFetches[i])
+					{
+						_entityFetchIndexes.Add(i);
+					}
 				}
+
+				_hasFetches = _entityFetchIndexes.Count > 0;
 			}
 
-			_hasFetches = _entityFetchIndexes.Count > 0;
 			if (loader.CollectionFetches == null)
 			{
 				return;
--- a/src/NHibernate/Cache/QueryCacheResultBuilder.cs
+++ b/src/NHibernate/Cache/QueryCacheResultBuilder.cs
@@ -1,0 +1,139 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using NHibernate.Engine;
+using NHibernate.Persister.Collection;
+using NHibernate.Type;
+
+namespace NHibernate.Cache
+{
+	/// <summary>
+	/// A builder that builds a list from a query that can be passed to <see cref="IBatchableQueryCache"/>.
+	/// </summary>
+	public sealed class QueryCacheResultBuilder
+	{
+		private readonly IType[] _resultTypes;
+		private readonly IType[] _cacheTypes;
+		private readonly ICollectionPersister[] _collectionPersisters;
+		private readonly List<int> _entityFetchIndexes = new List<int>();
+		private readonly List<int> _collectionFetchIndexes = new List<int>();
+		private readonly bool _hasFetches;
+		private readonly bool _hasCollectionFetches;
+		private readonly ISessionImplementor _session;
+
+		internal QueryCacheResultBuilder(Loader.Loader loader, ISessionImplementor session)
+		{
+			_resultTypes = loader.ResultTypes;
+			_cacheTypes = loader.CacheTypes;
+			_collectionPersisters = loader.GetCollectionPersisters();
+			_session = session;
+
+			if (loader.EntityFetches == null)
+			{
+				return;
+			}
+
+			for (var i = 0; i < loader.EntityFetches.Length; i++)
+			{
+				if (loader.EntityFetches[i])
+				{
+					_entityFetchIndexes.Add(i);
+				}
+			}
+
+			_hasFetches = _entityFetchIndexes.Count > 0;
+			if (loader.CollectionFetches == null)
+			{
+				return;
+			}
+
+			for (var i = 0; i < loader.CollectionFetches.Length; i++)
+			{
+				if (loader.CollectionFetches[i])
+				{
+					_collectionFetchIndexes.Add(i);
+				}
+			}
+
+			_hasCollectionFetches = _collectionFetchIndexes.Count > 0;
+		}
+
+		internal IList Result { get; } = new List<object>();
+
+		internal void AddRow(object result, object[] entities, object[] collectionKeys)
+		{
+			if (!_hasFetches)
+			{
+				Result.Add(result);
+				return;
+			}
+
+			var row = new object[_cacheTypes.Length];
+			if (_resultTypes.Length == 1)
+			{
+				row[0] = result;
+			}
+			else
+			{
+				Array.Copy((object[]) result, 0, row, 0, _resultTypes.Length);
+			}
+
+			var i = _resultTypes.Length;
+			foreach (var index in _entityFetchIndexes)
+			{
+				row[i++] = entities[index];
+			}
+
+			foreach (var index in _collectionFetchIndexes)
+			{
+				row[i++] = collectionKeys[index];
+			}
+
+			Result.Add(row);
+		}
+
+		internal IList GetResultList(IList cacheList)
+		{
+			if (!_hasFetches)
+			{
+				return cacheList;
+			}
+
+			var result = new List<object>();
+
+			foreach (object[] cacheRow in cacheList)
+			{
+				if (_resultTypes.Length == 1)
+				{
+					result.Add(cacheRow[0]);
+				}
+				else
+				{
+					var row = new object[_resultTypes.Length];
+					Array.Copy(cacheRow, 0, row, 0, _resultTypes.Length);
+					result.Add(row);
+				}
+
+				if (!_hasCollectionFetches)
+				{
+					continue;
+				}
+
+				var i = _cacheTypes.Length - _collectionFetchIndexes.Count;
+				foreach (var index in _collectionFetchIndexes)
+				{
+					var persister = _collectionPersisters[index];
+					var key = cacheRow[i];
+					var collection = _session.PersistenceContext.GetCollection(new CollectionKey(persister, key));
+					collection.ForceInitialization();
+					i++;
+				}
+			}
+
+			return result;
+		}
+	}
+}
--- a/src/NHibernate/Cache/QueryCacheResultBuilder.cs
+++ b/src/NHibernate/Cache/QueryCacheResultBuilder.cs
@@ -53,6 +53,8 @@ namespace NHibernate.Cache
 					_collectionFetchIndexes.Add(i);
 				}
 			}
+
+			_hasFetches = _hasFetches || _collectionFetchIndexes.Count > 0;
 		}
 
 		internal IList Result { get; } = new List<object>();
--- a/src/NHibernate/Cache/QueryCacheResultBuilder.cs
+++ b/src/NHibernate/Cache/QueryCacheResultBuilder.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using NHibernate.Collection;
 using NHibernate.Engine;
 using NHibernate.Persister.Collection;
 using NHibernate.Type;
@@ -17,19 +18,14 @@ namespace NHibernate.Cache
 	{
 		private readonly IType[] _resultTypes;
 		private readonly IType[] _cacheTypes;
-		private readonly ICollectionPersister[] _collectionPersisters;
 		private readonly List<int> _entityFetchIndexes = new List<int>();
 		private readonly List<int> _collectionFetchIndexes = new List<int>();
 		private readonly bool _hasFetches;
-		private readonly bool _hasCollectionFetches;
-		private readonly ISessionImplementor _session;
 
-		internal QueryCacheResultBuilder(Loader.Loader loader, ISessionImplementor session)
+		internal QueryCacheResultBuilder(Loader.Loader loader)
 		{
 			_resultTypes = loader.ResultTypes;
 			_cacheTypes = loader.CacheTypes;
-			_collectionPersisters = loader.GetCollectionPersisters();
-			_session = session;
 
 			if (loader.EntityFetches == null)
 			{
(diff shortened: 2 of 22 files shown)


src/AsyncGenerator.yml
```diff
@@ -104,6 +104,9 @@
     - conversion: Ignore
       name: GetEnumerator
       containingTypeName: IFutureEnumerable
+    # 6.0 TODO: Remove QueryCacheResultBuilder from ignore
+    - conversion: Ignore
+      containingTypeName: QueryCacheResultBuilder
```
maca88: Prevents from generating `ProcessResultsAsync` in `IQueryBatchItem`.

maca88: Not needed anymore as the initialization logic is now in the query cache

src/NHibernate/Multi/ICachingInformation.cs
```diff
@@ -68,4 +74,21 @@ public interface ICachingInformation
...
 		void SetCacheBatcher(CacheBatcher cacheBatcher);
 	}
+
+	internal static class CachingInformationExtensions
+	{
+		// 6.0 TODO: Move to ICachingInformation as a Property
+		public static IType[] GetCacheTypes(this ICachingInformation cachingInformation)
+		{
+			var loaderProperty = cachingInformation.GetType().GetProperty("Loader");
+			if (loaderProperty?.GetValue(cachingInformation) is Loader.Loader loader)
```
fredericDelaporte: Avoiding the reflection performance hit is surely worth switching `protected class QueryInfo` to `protected internal` in order to do a cast here instead.

src/NHibernate/Cache/QueryCacheResultBuilder.cs
```diff
@@ -0,0 +1,118 @@
...
+			}
+
+			for (var i = 0; i < loader.CollectionFetches.Length; i++)
+			{
+				if (loader.CollectionFetches[i])
+				{
+					_collectionFetchIndexes.Add(i);
+				}
+			}
+		}
```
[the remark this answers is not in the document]
maca88: It is not needed as a collection cannot be fetched without its owner which will be always located in `Loader.EntityFetches`.

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -320,39 +320,49 @@ public void Destroy()
...
+							collectionTypes.Add(i, collectionType);
+						}
+						else
+						{
+							nonCollectionTypeIndexes.Add(i);
+						}
+					}
+
 					// Skip first element, it is the timestamp
-					var rows = new List<object[]>(cacheable.Count - 1);
```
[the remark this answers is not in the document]
maca88: I created a [simple test](https://dotnetfiddle.net/K1xsJX) and it seems that having an additional list is worse than having a double cast.

src/NHibernate/Multi/ICachingInformation.cs
maca88: Unfortunately it is not possible as `QueryInfo` is a nested class of a generic type and the generic argument would be needed in order to get rid of the reflection but the callers are not able to provide it.

bahusoid: Well as an option you can introduce temp interface. Somethinkg like:
// 6.0 TODO: merge into 'ICachingInformation'.
public interface ICachingInformationWithPropertyFetch
{
	IType[] GetCacheTypes();
}

src/NHibernate/Cache/QueryCacheResultBuilder.cs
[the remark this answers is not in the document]
bahusoid: It possible to fetch collection without owner in criteria with `SelectMode.ChildFetch` so it's better to support this case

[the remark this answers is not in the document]
maca88: Ok, I've added it.

src/NHibernate/Multi/ICachingInformation.cs
maca88: Added the proposed interface.

src/NHibernate/Cache/QueryCacheResultBuilder.cs
```diff
@@ -0,0 +1,120 @@
...
+		private readonly List<int> _entityFetchIndexes = new List<int>();
+		private readonly List<int> _collectionFetchIndexes = new List<int>();
+		private readonly bool _hasFetches;
+
+		internal QueryCacheResultBuilder(Loader.Loader loader)
+		{
+			_resultTypes = loader.ResultTypes;
+			_cacheTypes = loader.CacheTypes;
+
+			if (loader.EntityFetches == null)
```
bahusoid: This shortcut needs to be removed to support collection fetches without owner.

maca88: Done.

fredericDelaporte:
I should have not validated it yet maybe: I have not checked (and not seen tests about) what would happen in case of Cartesian products due to the fetches. Won't it cache, in case of bags, duplicated collection items? And causes wrong data to be cached?

maca88:
I tested it by modifying the mappings to use bags instead and when I tried to fetch multiple bags I got an `QueryException` with message: `Cannot simultaneously fetch multiple bags`. Then I tried to change only one collection to be a bag and the bag collection had 16 items (where only 3 were unique) when the query was executed in the database and the same amount of items when executed for the second time where the results were retrieved from the cache. This means that the bag is cached with 16 items instead of 3 in this case, but the issue is not caused by this PR as the collection caching was not modified in this PR and the same issue occurs even if the query is not cached. This PR caches only the collection key which is then retrieved from the collection cache when assembling the query result from cache. I think that this issue is related to the `FetchMany` method which allows to create queries with cartesian products.

bahusoid:
> what would happen in case of Cartesian products due to the fetches

I'm also interested in this case. Could you please clarify how collections owners are cached in this case. Are they cached per row - the same owner is serialized for each row and then put to cache? Or there is some smart logic that serializes owner only once or even unique owner is put to cache only once?

maca88:
In order to explain how this PR works I will use a simple one to many fetch first, for example:
```
order = s.Query<Order>()
  .WithOptions(o => o.SetCacheable(true))
  .FetchMany(x => x.OrderLines)
  .Where(x => x.OrderId == 10248)
  .ToList()
  .First();
```
When the query is executed for the first time it will cache the following data:
```
[
  1234679, -- timestamp
  [10248, 1, 10248], -- The first value is the order key, the second is the order line key and the last one is the collection key
  [10248, 2, 10248],
  [10248, 3, 10248]
]
```
Now one would ask the following questions:
> Why are collection owners cached if this PR does not support caching a collection without its owner?

When I was doing this PR I didn't find an easy way to detect where the owner will be among the cached values especially when caching multiple collections.

> Why are the collection items cached if the collection will be retrieved from the collection cache that have this information?

The idea was to be able to batch fetch all collection items at once when batch fetching is enabled so if we fetched 3 collection we would be able to get all collection items in one cache call, which is done in #1955.
Additionaly, the mentioned PR uses this information to know if it should batch fetch all of them when one of them is requested or only batch fetch them by the defined batch size.

When the query is executed for the second time the query result will be retrieved from the cache and the cached entites will be assembled by the `TypeHelper.Assemble` and for collections the newly added static method `TypeHelper.InitializeCollections` will be called that uses `ICollectionPersistent.ForceInitialization` method to initialize them. In our case the `TypeHelper.InitializeCollections` will be called 3 times for the same collection which is fine as `ICollectionPersistent.ForceInitialization` does nothing when the collection is already initialized.

Now if we have a cacheable query with a cartesian product the only difference is that we will have a lot more duplicated values. For the [test I've made](https://github.com/maca88/nhibernate-core/commit/98d7f2d4b4f9f464d822f047b13b1a119e92d4af), the following data is cached:
```
[
  6346353453645824,
  [10248,"VINET",10248,1,"VINET",10248],
  [10248,"VINET",10248,2,"VINET",10248],
  [10248,"VINET",10248,3,"VINET",10248],
  [10248,"VINET",10274,1,"VINET",10248],
  [10248,"VINET",10274,2,"VINET",10248],
  [10248,"VINET",10274,3,"VINET",10248],
  [10248,"VINET",10295,1,"VINET",10248],
  [10248,"VINET",10295,2,"VINET",10248],
  [10248,"VINET",10295,3,"VINET",10248],
  [10248,"VINET",10737,1,"VINET",10248],
  [10248,"VINET",10737,2,"VINET",10248],
  [10248,"VINET",10737,3,"VINET",10248],
  [10248,"VINET",10739,1,"VINET",10248],
  [10248,"VINET",10739,2,"VINET",10248],
  [10248,"VINET",10739,3,"VINET",10248]
]
```

> With this PR, we would now have wrong results cached for the bag, and later loading of the owner followed by a lazy load of the bag would retrieve these wrong results from cache.

[I've made a test](https://github.com/maca88/nhibernate-core/commit/e6604a8e07b0477527c99dd31e1c5a4169e371de) on the current master that shows that the bag can be cached with the duplicated elements, so the issue already exists.

fredericDelaporte:
Rebase for resolving conflicts.

PR #1954, opened 2019-01-02
Move discovery of executedSpaces to RegisterCleanupActions

### LINKED ISSUE #1953: Query space invalidation doesn't work for bulk actions
The simplification done for #1467 was too simplified. Now, query space invalidation doesn't happen for bulk updates at all.

I set it to Critical, since it makes query cache usage unusable together with bulk actions.

gliljas:
Since BulkOperationCleanupAction isn't added among the regular executables, its queryspaces are ignored. The original implementation had AddSpacesToInvalidate on AfterTransactionCompletionProcessQueue, but we managed to simplify that away.

This fix doesn't readd that, but instead just moves the "discovery" to RegisterCleanupActions.

Fixes #1953

### DIFF
--- a/src/NHibernate.Test/Async/SecondLevelCacheTest/InvalidationTests.cs
+++ b/src/NHibernate.Test/Async/SecondLevelCacheTest/InvalidationTests.cs
@@ -15,6 +15,7 @@ using System.Reflection;
 using NHibernate.Cache;
 using NHibernate.Cfg;
 using NHibernate.Impl;
+using NHibernate.Linq;
 using NHibernate.Test.SecondLevelCacheTests;
 using NSubstitute;
 using NUnit.Framework;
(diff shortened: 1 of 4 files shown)


fredericDelaporte:
Feel free to not create an issue when intending to do a PR: the PR is then the issue by the way.

As this is a regression of 5.1, patches should be done on 5.2.x and 5.1.x, especially if it should be considered critical.

Could you then re-target 5.2.x rather than master ? (It will have to be back-ported to 5.1.x once validated on 5.2.x.)

hazzik:
I want to include #1269 to 5.2.2 as well.

fredericDelaporte:
> I want to include #1269 to 5.2.2 as well (it is a first commit after 5.2.1).

Yes I was remembering. I think 5.2.x should be set to 92d9b07, since it currently lies at the previous commit. But I consider doing this only once this PR here is validated (and of course before merging it).

PR #1955, opened 2019-01-04
Optimize batchable cache calls for cached queries

maca88:
This PR tends to minimize 2nd level cache calls when retrieving entities/collections from the batchable cache when assembling objects from a cached query. Today, when a cached query is retrieved from the cache the same batching logic is applied as when a lazy entity is initialized, which means that the entity batch fetch size will be used that can be far less that the query result count. To solve this problem a special queue was created that is initialized by the query cache and will replace the current one in order to optimize the cache calls. Once the query cache is done with the assembling the special queue is removed and the old one takes its place.
There is still one optimization that can be done but would require to have the `CacheBatcher` on a sesssion in order to combine multiple entity/collection batch loads when they are not present in the cache. Additionaly a special setting could be added that would allow to batch a bigger number of entities when batching from the database entities that are present in the cached query. 

This PR should be reviewed after #1952

### DIFF
--- a/src/NHibernate/Engine/QueryCacheBatchQueue.cs
+++ b/src/NHibernate/Engine/QueryCacheBatchQueue.cs
@@ -188,11 +188,6 @@ namespace NHibernate.Engine
 		/// <param name="key">The entity key.</param>
 		internal void AddEntityKey(EntityKey key)
 		{
-			if (!key.IsBatchLoadable)
-			{
-				return;
-			}
-
 			if (!_queryEntityKeys.TryGetValue(key.EntityName, out var querySet))
 			{
 				querySet = new HashSet<EntityKey>();
@@ -209,11 +204,6 @@ namespace NHibernate.Engine
 		/// <param name="ce">The collection entry.</param>
 		internal void AddCollection(ICollectionPersister persister, CollectionKey ce)
 		{
-			if (persister.GetBatchSize() <= 1)
-			{
-				return;
-			}
-
 			if (!_queryCollectionKeys.TryGetValue(persister.Role, out var querySet))
 			{
 				querySet = new Dictionary<CollectionKey, CollectionEntry>();
--- a/src/NHibernate/Engine/QueryCacheBatchQueue.cs
+++ b/src/NHibernate/Engine/QueryCacheBatchQueue.cs
@@ -1,0 +1,271 @@
+﻿using System;
+using System.Collections.Generic;
+using NHibernate.Persister.Collection;
+using NHibernate.Persister.Entity;
+
+namespace NHibernate.Engine
+{
+	/// <summary>
+	/// A batcher used to retrieve a batch of entity or collection keys that are present in the cached query.
+	/// </summary>
+	internal class QueryCacheBatchQueue
+	{
+		private readonly IPersistenceContext _persistenceContext;
+
+		/// <summary>
+		/// Used to hold information about the entities that are currently eligible for batch-fetching. Ultimately
+		/// used by <see cref="GetEntityBatch" /> to build entity load batches.
+		/// </summary>
+		private readonly IDictionary<string, HashSet<EntityKey>> _queryEntityKeys;
+
+		/// <summary>
+		/// Used to hold information about entity keys that were checked in the cache.
+		/// </summary>
+		private readonly IDictionary<string, HashSet<EntityKey>> _queryCheckedEntityKeys;
+
+		/// <summary>
+		/// Used to hold information about collection entries that are currently eligible for batch-fetching. Ultimately
+		/// used by <see cref="GetCollectionBatch" /> to build collection load batches.
+		/// </summary>
+		private readonly IDictionary<string, IDictionary<CollectionKey, CollectionEntry>> _queryCollectionKeys;
+
+		/// <summary>
+		/// Used to hold information about collection keys that were checked in the cache.
+		/// </summary>
+		private readonly IDictionary<string, HashSet<CollectionKey>> _queryCheckedCollectionKeys;
+
+		/// <summary>
+		/// Used to hold information about collection entries that were checked in the cache.
+		/// </summary>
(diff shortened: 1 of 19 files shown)


src/NHibernate/Util/CollectionHelper.cs
```diff
@@ -223,6 +223,8 @@ public IEnumerator GetEnumerator()
 
 		internal static ISet<T> EmptySet<T>() => EmptyReadOnlySet<T>.Instance;
 
+		internal static T[] EmptyArray<T>() => EmptyArrayClass<T>.Instance;
```
hazzik: This is not required. The fx already has `Array.Empty<T>()`.

hazzik:
@maca88 can you please rebase?

maca88:
Rebased, removed `EmptyArrayClass` and added an extra test for fetching a many-to-one relation.

maca88:
Rebased.

src/NHibernate/Engine/QueryCacheBatchQueue.cs
```diff
@@ -0,0 +1,271 @@
...
+			return result;
+		}
+
+		/// <summary>
+		/// Adds the entity to the batch.
+		/// </summary>
+		/// <param name="key">The entity key.</param>
+		internal void AddEntityKey(EntityKey key)
+		{
+			if (!key.IsBatchLoadable)
```
fredericDelaporte: This condition is indeed `batchSize <= 1`. In the context of caching or retrieving from cache a query's results, I do not think `batchSize` is relevant. `batchSize` is relevant to lazy loads, it decides whether to pro-actively load more data than directly asked by the current operation or not.
For a cacheable query, there is no reason to prevent a multiple get of its results, they will all be attempted anyway.
But well, removing the condition here will not change anything currently, since anyway the call to `AddEntityKey` is already ruled-out by the same condition in `ManyToOneType`.

It seems to me all the already existing checks on `batchSize` should be disabled/avoided by this PR when a query cache batch queue is enabled, and no new ones should be added. But maybe doing this may add to much complexity.

src/NHibernate/Engine/QueryCacheBatchQueue.cs
```diff
@@ -0,0 +1,271 @@
...
+		}
+
+		/// <summary>
+		/// Adds the collection to the batch.
+		/// </summary>
+		/// <param name="persister">The collection persister.</param>
+		/// <param name="ce">The collection entry.</param>
+		internal void AddCollection(ICollectionPersister persister, CollectionKey ce)
+		{
+			if (persister.GetBatchSize() <= 1)
```
fredericDelaporte: Same remarks than previous one.

src/NHibernate/Engine/QueryCacheBatchQueue.cs
maca88: I've modified the code to allow batching regardless of `batchSize` and I don't think that it added too much complexity.

src/NHibernate/Event/Default/DefaultInitializeCollectionEventListener.cs
```diff
@@ -80,24 +80,20 @@ public virtual void OnInitializeCollection(InitializeCollectionEvent @event)
 			}
 
 			var batchSize = persister.GetBatchSize();
-			if (batchSize > 1 && persister.Cache.PreferMultipleGet())
+			CollectionEntry[] collectionEntries = null;
+			var collectionBatch = source.PersistenceContext.BatchFetchQueue.QueryCacheQueue
+			                            ?.GetCollectionBatch(persister, collectionKey, out collectionEntries);
+			if ((collectionBatch != null || batchSize > 1) && persister.Cache.PreferMultipleGet())
```
fredericDelaporte: I should have think about it sooner, but the same applies to `PreferMultipleGet`. See its xml doc remarks:
```c#
/// When <see langword="true" />, NHibernate will attempt to get other non initialized proxies or
/// collections from the cache instead of only getting the proxy or collection which initialization
/// is asked for. If this cache implementation does not benefit from batching together get operations,
/// this may result in a performance loss.
/// </para>
/// <para>
/// When <see langword="false" />, NHibernate will still call <see cref="GetMany" /> when it has many
/// gets to perform. Its <see cref="CacheBase" /> default implementation is adequate for this case.
/// </para>
```

So the condition should be:
```suggestion
			if (collectionBatch != null || batchSize > 1 && persister.Cache.PreferMultipleGet())
```

src/NHibernate/Event/Default/DefaultLoadEventListener.cs
```diff
@@ -425,23 +425,19 @@ protected virtual object LoadFromSecondLevelCache(LoadEvent @event, IEntityPersi
 			}
 			ISessionFactoryImplementor factory = source.Factory;
 			var batchSize = persister.GetBatchSize();
-			if (batchSize > 1 && persister.Cache.PreferMultipleGet())
+			var entityBatch = source.PersistenceContext.BatchFetchQueue.QueryCacheQueue
+			                        ?.GetEntityBatch(persister, @event.EntityId);
+			if ((entityBatch != null || batchSize > 1) && persister.Cache.PreferMultipleGet())
```
fredericDelaporte:
```suggestion
			if (entityBatch != null || batchSize > 1 && persister.Cache.PreferMultipleGet())
```

maca88:
I forgot to re-run the async generator for the last commit, should we create a PR to fix that or do a force push on the last commit in the master?

fredericDelaporte:
Force pushes are not allowed on master. (I could lift the restriction but I would not do it. On patch branches I may allow myself to lift it for dropping more serious mishap.)

For merged async regeneration lacks, the handling varies. When it has no impact on the feature (like namespace or test only changes), we tend to just let some later PR fix it with its own async regeneration. When it has an impact on the feature, better do a separate fix.

So I have done #2063.

PR #1959, opened 2019-01-06
Backport Query space invalidation doesn't work for bulk actions

fredericDelaporte:
This backports #1953.

Some changes had to be done due to #1392 being done only in 5.2.0.

This PR also directly includes the release changes. Should they stay as a separated commits? Or does it seem right to squash it all together? I would tend to squash it all.


hazzik:
>Should they stay as a separated commit?
Yes please.

PR #1965, opened 2019-01-10
Fix code sample in docs, section 10.4.2

stefanhamburger:
There is no variable `mate` in [this code sample](https://nhibernate.info/doc/nhibernate-reference/manipulatingdata.html#manipulatingdata-updating-detached), so the code `secondSession.Update(mate);` won't compile.

The line can easily be removed. It may have been copied & pasted from a similar sample a few lines below.


fredericDelaporte:
Thanks.

PR #1972, opened 2019-01-13
Port Hibernate's EntityKey optimization

maca88:
This PR ports Hibernate's [HHH-8682](https://hibernate.atlassian.net/browse/HHH-8682), that aims to have an `EntityKey` as smallest as possible. In addition the `ManyToOneType.ScheduleBatchLoadIfNeeded` was modified to not generate an entity key when batching is disabled in order to reduce garbage collection. Ideally, an object pool could be implemented in order to prevent generating keys just to check if the key exist in the persistence context or not as from profiling, 35 percent of the time running the "set fetches" test is spent by the GC.

Here are some benchmark results from `RawDataAccessBencher`, before:
```
Async eager Load fetches
-------------------------
# of elements fetched: 6768 (4768 + 1000 + 1000).    Fetch took: 90.673.920,00ms. Allocated bytes: 90673920.

Change tracking fetches, set fetches (20 runs), no caching
------------------------------------------------------------------------------
NHibernate v5.2.0.0 (v5.2.2.0) : 2.050,20ms (58,98ms)   Enum: 1,71ms (0,08ms)

Memory usage, per iteration
------------------------------------------------------------------------------
NHibernate v5.2.0.0 (v5.2.2.0) : 195.868 KB (200.569.192 bytes)
```

after:
```
Async eager Load fetches
-------------------------
# of elements fetched: 6768 (4768 + 1000 + 1000).    Fetch took: 87.064.984,00ms. Allocated bytes: 87064984.

Change tracking fetches, set fetches (20 runs), no caching
------------------------------------------------------------------------------
NHibernate v5.2.0.0 (v5.2.2.0) : 1.853,30ms (57,91ms)   Enum: 1,67ms (0,06ms)

Memory usage, per iteration
------------------------------------------------------------------------------
NHibernate v5.2.0.0 (v5.2.2.0) : 173.056 KB (177.209.960 bytes)
```

### DIFF
--- a/src/NHibernate/Engine/EntityKey.cs
+++ b/src/NHibernate/Engine/EntityKey.cs
@@ -12,7 +12,7 @@ namespace NHibernate.Engine
 	/// and the identifier space (eg. tablename)
 	/// </summary>
 	[Serializable]
-	public sealed class EntityKey : IDeserializationCallback, ISerializable
+	public sealed class EntityKey : IDeserializationCallback, ISerializable, IEquatable<EntityKey>
 	{
 		private readonly object identifier;
 		private readonly IEntityPersister _persister;
@@ -49,12 +49,19 @@ namespace NHibernate.Engine
 
 		public override bool Equals(object other)
 		{
-			var otherKey = other as EntityKey;
-			if(otherKey==null) return false;
+			return other is EntityKey otherKey && Equals(otherKey);
+		}
+
+		public bool Equals(EntityKey other)
+		{
+			if (other == null)
+			{
+				return false;
+			}
 
 			return
-				otherKey.RootEntityName.Equals(RootEntityName)
-				&& _persister.IdentifierType.IsEqual(otherKey.Identifier, Identifier, _persister.Factory);
+				other.RootEntityName.Equals(RootEntityName)
+				&& _persister.IdentifierType.IsEqual(other.Identifier, Identifier, _persister.Factory);
 		}
 
 		public override int GetHashCode()
--- a/src/NHibernate/Engine/EntityKey.cs
+++ b/src/NHibernate/Engine/EntityKey.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Runtime.Serialization;
+using System.Security;
 using NHibernate.Impl;
 using NHibernate.Persister.Entity;
 using NHibernate.Type;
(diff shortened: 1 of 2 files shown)


fredericDelaporte:
For merging, I intend to squash and merge in order to reword the commit.

hazzik:
Need to implement IEquitable to squeeze more performance.

maca88:
Done.

hazzik:
> Done

Can you run a test again?

maca88:
The results are more or less the same: 
```
Async eager Load fetches
-------------------------
# of elements fetched: 6768 (4768 + 1000 + 1000).    Fetch took: 87.025.824,00ms. Allocated bytes: 87025824.

Results per framework. Values are given as: 'mean (standard deviation)'
==============================================================================

Change tracking fetches, set fetches (20 runs), no caching
------------------------------------------------------------------------------
NHibernate v5.2.0.0 (v5.2.2.0) : 1.851,49ms (51,74ms)   Enum: 1,72ms (0,11ms)

Memory usage, per iteration
------------------------------------------------------------------------------
NHibernate v5.2.0.0 (v5.2.2.0) : 173.071 KB (177.225.576 bytes)
```

bahusoid:
@maca88  I've prepared a patch for this branch with EntityKey class->struct conversion:
[0001-Convert-EntityKey-to-struct.zip](https://github.com/nhibernate/nhibernate-core/files/2754874/0001-Convert-EntityKey-to-struct.zip)
It can be applied via `git am < 0001-Convert-EntityKey-to-struct.patch` for unzipped file. Could you please check how it affects benchmark results.

bahusoid:
Interesting. Thanks! Though I'm not sure I'm following you regarding diffreneces for `string` vs `int` for identifier. From what I know stored value in field should not affect struct size in any way.

maca88:
As string is not a value type, a reference to the string will be stored in the struct which on x64 platform is 8 bytes. So if we combine all fields we get 20 bytes:
identifier (string): 8 bytes (reference)
persister: 8 bytes (reference)
hashcode: 4 byte (value)

bahusoid:
What I'm saying is `EntityKey.identifier` is of type `object`. I expect it always to be 8 bytes on 64 bit. It doesn't matter what value you put in it -  boxed int or string

maca88:
Yea you are right, I messed up.

hazzik:
>By just appling your patch the overall performance wasn't changed

But memory usage dropped by ~6 megabytes.

fredericDelaporte:
> Another option for optimizing such cases is to use the object pooling that is frequently [used in games](https://www.raywenderlich.com/847-object-pooling-in-unity).

This pooling implies mutating the cache keys. I find it quite debatable: it has a good potential for serious bugs. If a key, which has been used for storing caching something, is mutated, things will obviously go wrong.

maca88:
> This pooling implies mutating the cache keys. I find it quite debatable: it has a good potential for serious bugs

Indeed, it may be the fastest but for this case using a struct would probably be a better choice which would also lower the memory usage as @hazzik pointed out.

PR #1977, opened 2019-01-15
Remove IDeserializationCallback from EntityKey

bahusoid:
Follow up to #1972


bahusoid:
Why breaking change? This is serialization specific inerface. What kind of usageare you expect in user code?

bahusoid:
I can leave it with empty implementation if you think it can break anything

bahusoid:
For the future: If possible always consider implementing new interfaces explicitly to avoid breaking changes removing it

PR #1979, opened 2019-01-15
Refactor sequential select

maca88:
This PR moves the generated sequential statements form the root entity to its subclasses which then simplifies the `ILoadable.Hydrate` method as the root persister parameter is not needed anymore.

### DIFF
--- a/src/NHibernate/Async/Persister/Entity/AbstractEntityPersister.cs
+++ b/src/NHibernate/Async/Persister/Entity/AbstractEntityPersister.cs
@@ -36,6 +36,7 @@ using Array=System.Array;
 using Property=NHibernate.Mapping.Property;
 using NHibernate.SqlTypes;
 using System.Linq;
+using System.Linq.Expressions;
 using NHibernate.Bytecode;
 
 namespace NHibernate.Persister.Entity
@@ -339,7 +340,21 @@ namespace NHibernate.Persister.Entity
 		/// Unmarshall the fields of a persistent instance from a result set,
 		/// without resolving associations or collections
 		/// </summary>
-		public async Task<object[]> HydrateAsync(DbDataReader rs, object id, object obj, ILoadable rootLoadable,
+		public Task<object[]> HydrateAsync(DbDataReader rs, object id, object obj, ILoadable rootLoadable,
+			string[][] suffixedPropertyColumns, bool allProperties, ISessionImplementor session, CancellationToken cancellationToken)
+		{
+			if (cancellationToken.IsCancellationRequested)
+			{
+				return Task.FromCanceled<object[]>(cancellationToken);
+			}
+			return HydrateAsync(rs, id, obj, suffixedPropertyColumns, allProperties, session, cancellationToken);
+		}
+
+		/// <summary>
+		/// Unmarshall the fields of a persistent instance from a result set,
+		/// without resolving associations or collections
+		/// </summary>
+		public async Task<object[]> HydrateAsync(DbDataReader rs, object id, object obj,
 			string[][] suffixedPropertyColumns, bool allProperties, ISessionImplementor session, CancellationToken cancellationToken)
 		{
 			cancellationToken.ThrowIfCancellationRequested();
@@ -348,9 +363,7 @@ namespace NHibernate.Persister.Entity
 				log.Debug("Hydrating entity: {0}", MessageHelper.InfoString(this, id, Factory));
 			}
 
-			AbstractEntityPersister rootPersister = (AbstractEntityPersister)rootLoadable;
-
-			bool hasDeferred = rootPersister.HasSequentialSelect;
+			bool hasDeferred = HasSequentialSelect;
 			DbCommand sequentialSelect = null;
 			DbDataReader sequentialResultSet = null;
 			bool sequentialSelectEmpty = false;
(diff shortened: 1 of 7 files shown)


src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -4491,5 +4520,49 @@ public string GetInfoString()
 			return MessageHelper.InfoString(this);
 		}
 		#endregion
+
+		internal SqlString GenerateSequentialSelect(ILoadable persister)
```
hazzik: The argument should be `AbstractEntityPersister`.

maca88: Done.

maca88:
> This sequential select feature such a mess.....

And also a quite old one that hasn't changed since 2007 even in Hibernate.

I've restored the `HasSequentialSelect` behavior in order to avoid a breaking change and deprecate it in favor of `GetSequentialSelect`.

fredericDelaporte:
> This sequential select feature such a mess.....

What that feature is about? Eager fetch mode `select` on associations?

hazzik:
> What that feature is about?

It's about eager joins inside of an entity. See here: https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate.Test/Join/JoinTest.cs#L47. It is a kind of an in-between of table-per-hierarchy and joined-subclass features.

About mapping like this:
https://github.com/nhibernate/nhibernate-core/blob/1711b5c5162e9981fb6d83aaae8810b09aa97876/src/NHibernate.Test/Join/Person.hbm.xml#L7-L72

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -2590,63 +2612,65 @@ protected int Dehydrate(object id, object[] fields, bool[] includeProperty, bool
 		/// </summary>
 		public object[] Hydrate(DbDataReader rs, object id, object obj, ILoadable rootLoadable,
 			string[][] suffixedPropertyColumns, bool allProperties, ISessionImplementor session)
```
fredericDelaporte: This method should be obsoleted.

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -4491,5 +4517,48 @@ public string GetInfoString()
...
+				if (IsSubclassTableSequentialSelect(propTableNumber) && !IsSubclassTableLazy(propTableNumber))
+				{
+					tableNumbers.Add(propTableNumber);
+				}
+			}
+			if (tableNumbers.Count == 0)
+				return null;
+
+			//figure out which columns are needed (excludes lazy-properties)
+			var columnNumbers = new List<int>();
```
fredericDelaporte: The comment lets think some filter on `subclassColumnLazyClosure` is done by the code, but it is not the case. Are the lazy columns really excluded here? I do not think they are.

But it seems to be an old issue of the code, just moved "as-is" from `SingleTableEntityPersister`. If there is something to fix here, better do it in another PR. (Excepted if the fix is just changing the comment.)

There is the same issue for formulas below.

maca88: > Are the lazy columns really excluded here?

No, they are not. When the filter will be added the #1922 would need to be enhanced to support it as we can [fetch a lazy property of a subclass](https://github.com/nhibernate/nhibernate-core/pull/1922/files#diff-03963f8bf0b4f656910fabd85135c7bcR678). In addition the current `fetch all properties` would also need to be enhanced to support it.

fredericDelaporte:
There are some conflicts to resolve with #1922 merge into master. For `AbstractEntityPersister`, they do not look trivial.

maca88:
Conflicts are resolved.

src/NHibernate/Async/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -400,14 +401,14 @@ protected Task<int> DehydrateAsync(object id, object[] fields, bool[] includePro
 		/// Unmarshall the fields of a persistent instance from a result set,
 		/// without resolving associations or collections
 		/// </summary>
-		public Task<object[]> HydrateAsync(DbDataReader rs, object id, object obj, ILoadable rootLoadable,
+		public Task<object[]> HydrateAsync(DbDataReader rs, object id, object obj,
```
fredericDelaporte: Added in #1922 so changing it is ok.

src/NHibernate/Async/Persister/Entity/ILoadable.cs
```diff
@@ -38,7 +38,7 @@ public static partial class LoadableExtensions
 		/// </summary>
 		//6.0 TODO: Merge into ILoadable
 		public static Task<object[]> HydrateAsync(
-			this ILoadable loadable, DbDataReader rs, object id, object obj, ILoadable rootLoadable,
+			this ILoadable loadable, DbDataReader rs, object id, object obj,
```
fredericDelaporte: Same than previous comment.

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -2685,10 +2707,10 @@ protected int Dehydrate(object id, object[] fields, bool[] includeProperty, bool
 		/// Unmarshall the fields of a persistent instance from a result set,
 		/// without resolving associations or collections
 		/// </summary>
-		public object[] Hydrate(DbDataReader rs, object id, object obj, ILoadable rootLoadable,
+		public object[] Hydrate(DbDataReader rs, object id, object obj,
```
fredericDelaporte: Same than previous comment.

src/NHibernate/Persister/Entity/ILoadable.cs
```diff
@@ -83,14 +83,18 @@ public static partial class LoadableExtensions
 		/// </summary>
 		//6.0 TODO: Merge into ILoadable
 		public static object[] Hydrate(
-			this ILoadable loadable, DbDataReader rs, object id, object obj, ILoadable rootLoadable,
+			this ILoadable loadable, DbDataReader rs, object id, object obj,
```
fredericDelaporte: Same than previous comment.

PR #1980, opened 2019-01-16
Cache proxy activators in static proxy factory

bahusoid:


bahusoid:
Dumb question: can proxy or proxy field activators be the same for different persisters? If not we can also get rid of static dictionaries.

hazzik:
I did not understand the question

bahusoid:
Question arises from the following peace of code:
https://github.com/nhibernate/nhibernate-core/blob/1711b5c5162e9981fb6d83aaae8810b09aa97876/src/NHibernate/Proxy/StaticProxyFactory.cs#L28-L32

So why do we need `Cache` dictionary here? Can't we directly create activator and cache it inside `StaticProxyFactory`? Or it's possible that mulitple persisters have the same `_cacheEntry`?

bahusoid:
Hm.. Actually Lazy is damn slow for some reasons and perform worse than dic lookup on small count of objects. Pitty

PR #1984, opened 2019-01-17
Reduce SessionIdLoggingContext creation

bahusoid:
1) Do not create `SessionIdLoggingContext` if `SessionId` is empty or not changed;
2) Set context on session creation;
It's the simplest way to avoid objects creation in most common scenarios - operations scoped to single session:
```C#
using(var session = OpenSession())
{
session.Method1();
...
using(var session2 = OpenSession())
{
....
}
...
session.MethodN()
}
```

This change may lead to scenarios when SessionId is not properly reset:
```C#
session1 = Open();
session2 = Open();

session1.Dispose();// <- SessionId is set to null
...
session2.Dispose();// <- SessionId is set to value from session1
```
But what's harm in it? It's barely valid scenarios and inside session methods it will be properly set again by `BeginContext` methods.

### DIFF
--- a/src/NHibernate/Impl/SessionIdLoggingContext.cs
+++ b/src/NHibernate/Impl/SessionIdLoggingContext.cs
@@ -15,16 +15,36 @@ namespace NHibernate.Impl
 #else
 		private const string LogicalCallContextVariableName = "__" + nameof(SessionIdLoggingContext) + "__";
 #endif
-		private readonly Guid? _oldSessonId;
-		private readonly bool _hasChanged;
-		
+		private readonly Guid? _oldSessionId;
+		private bool _hasChanged;
+
+		[Obsolete("Please use SessionIdLoggingContext.CreateOrNull instead.")]
 		public SessionIdLoggingContext(Guid id)
 		{
 			if (id == Guid.Empty) return;
-			_oldSessonId = SessionId;
-			if (id == _oldSessonId) return;
+			_oldSessionId = SessionId;
+			if (id == _oldSessionId) return;
 			_hasChanged = true;
 			SessionId = id;
+		}
+
+		private SessionIdLoggingContext(Guid newId, Guid? oldId)
+		{
+			SessionId = newId;
+			_oldSessionId = oldId;
+			_hasChanged = true;
+		}
+
+		public static IDisposable CreateOrNull(Guid id)
+		{
+			if (id == Guid.Empty)
+				return null;
+			var oldId = SessionId;
+
+			if (oldId == id)
+				return null;
+
+			return new SessionIdLoggingContext(id, oldId);
 		}
 
 		/// <summary>
@@ -58,7 +78,8 @@ namespace NHibernate.Impl
 		{
 			if (_hasChanged)
 			{
-				SessionId = _oldSessonId;
+				SessionId = _oldSessionId;
+				_hasChanged = false;
 			}
 		}
 	}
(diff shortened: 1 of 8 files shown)


src/NHibernate/Impl/SessionIdLoggingContext.cs
```diff
@@ -15,18 +15,38 @@ public class SessionIdLoggingContext : IDisposable
...
+		{
+			SessionId = newId;
+			_oldSessionId = oldId;
+			_hasChanged = true;
+		}
+
+		public static IDisposable CreateOrNull(Guid id)
+		{
+			if (id == Guid.Empty)
+				return null;
```
fredericDelaporte: This shortcut will not play nice with cases where usages of sessions having an id are mixed with sessions not having one. That is likely an extreme corner case, though.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -183,7 +183,8 @@ void IDeserializationCallback.OnDeserialization(object sender)
 		internal SessionImpl(SessionFactoryImpl factory, ISessionCreationOptions options)
 			: base(factory, options)
 		{
-			using (BeginContext())
+			_context = SessionIdLoggingContext.CreateOrNull(SessionId);
```
fredericDelaporte: I think we should explain here in code why it is not disposed immediately. Something like:
```c#
// This context is disposed only on session own disposal. This greatly reduces the number of context switches
// for most usual session usages. It may cause an irrelevant session id to be set back on disposal, but since all
// session entry points are supposed to set it, it should not have any consequences.
```

src/NHibernate/Impl/StatelessSessionImpl.cs
```diff
@@ -31,7 +31,8 @@ public partial class StatelessSessionImpl : AbstractSessionImpl, IStatelessSessi
 		internal StatelessSessionImpl(SessionFactoryImpl factory, ISessionCreationOptions options)
 			: base(factory, options)
 		{
-			using (BeginContext())
+			_context = SessionIdLoggingContext.CreateOrNull(SessionId);
```
fredericDelaporte: Same as previous comment.

src/NHibernate/Impl/SessionIdLoggingContext.cs
bahusoid: Well this case was not supported anyway:
https://github.com/nhibernate/nhibernate-core/blob/1711b5c5162e9981fb6d83aaae8810b09aa97876/src/NHibernate/Impl/SessionIdLoggingContext.cs#L21-L28

As you can see `_hasChanged` is never set for sessions without `id`

fredericDelaporte: ok

src/NHibernate/Impl/SessionImpl.cs
bahusoid: Done

hazzik:
It did help. But not too much. About 2mb of memory and 20ms. 😞.

hazzik:
I've double-checked my records. SessionIdLoggingContext was responsible only for 2mb of memory traffic (ooops, sorry), which is fixed by this PR.

bahusoid:
I've added `ProcessHelper` pooling. `SessionIdLoggingContext.CreateOrNull` already avoids any unnecessary objects creations - so I think it's good in a current state.

PR #1990, opened 2019-01-24
Replace <PackageLicenseUrl /> with <PackageLicenseExpression />

### LINKED ISSUE #1973: Investigate licenseUrl deprecation
Pushing the Nuget package yields a warning about `licenseUrl` being deprecated.
See [here](https://docs.microsoft.com/en-us/nuget/reference/nuspec#licenseurl), we should use `license` instead. These are Nuget properties. We currently specify the .csproj property `PackageLicenseUrl`.

What should be done should be investigated.

hazzik:
Closes #1973

### DIFF
--- a/build-common/NHibernate.props
+++ b/build-common/NHibernate.props
@@ -24,7 +24,7 @@
     <PackageProjectUrl>http://nhibernate.info</PackageProjectUrl>
     <PackageIconUrl>https://raw.githubusercontent.com/nhibernate/nhibernate-core/master/logo/NHibernate-NuGet.png</PackageIconUrl>
     <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
-    <PackageLicenseUrl>https://raw.githubusercontent.com/nhibernate/nhibernate-core/master/LICENSE.txt</PackageLicenseUrl>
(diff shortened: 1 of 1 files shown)


build-common/NHibernate.props
```diff
@@ -24,7 +24,7 @@
     <PackageProjectUrl>http://nhibernate.info</PackageProjectUrl>
     <PackageIconUrl>https://raw.githubusercontent.com/nhibernate/nhibernate-core/master/logo/NHibernate-NuGet.png</PackageIconUrl>
     <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
-    <PackageLicenseUrl>https://raw.githubusercontent.com/nhibernate/nhibernate-core/master/LICENSE.txt</PackageLicenseUrl>
+    <PackageLicenseFile>NHibernate.license.txt</PackageLicenseFile>
```
fredericDelaporte: Can't we use [`PackageLicenseExpression`](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/nuget#important-nuget-package-metadata) with value `LGPL-2.1-only` instead? This should allow Nuget to directly states the package license, instead of having to read the file.

hazzik: Done. `LGPL-2.1-only` it is.

PR #1995, opened 2019-01-27
Fix #1994 (many-to-many filter in hql)

### LINKED ISSUE #1994: Extra Select for every "outfiltered" Element
We use the LINQ Provider with the FetchMany extension to reduce the roundtrips to the database and the amount of data. Unfortunately it is not possible to filter the fetched association like `FetchMany(x => x.MyAssociation.Where(...))` , so we tried to use the Filter in the Mapping files. This seems to work, but the problem is, that for every out filtered element, there is a select statement in the log file. Is it really necessary to create a extra select there? We use the Version 5.2.1.

This is the created select statement:

```
    select
        asset0_.pdoid as pdoid1_0_0_,
        dokument2_.pdoid as pdoid1_0_1_,
        asset0_.IsDeleted as isdeleted2_0_0_,
        dokument2_.IsDeleted as isdeleted2_0_1_,
        dokumente1_.AssetId as assetid1_4_0__,
        dokumente1_.DokumentId as dokumentid2_4_0__ 
    from
        Asset asset0_ 
    left outer join
        asset_to_dokument dokumente1_ 
            on asset0_.pdoid=dokumente1_.AssetId 
    left outer join
        Dokument dokument2_ 
            on dokumente1_.DokumentId=dokument2_.pdoid 
            and dokument2_.IsDeleted = @p0 
    where
        asset0_.IsDeleted = @p0;
    @p0 = False [Type: Boolean (0:0:0)]

```

And for every deleted "Dokument" (for example 1000 elements) there is a statement like this:

```
  SELECT
        dokument0_.pdoid as pdoid1_0_0_,
        dokument0_.IsDeleted as isdeleted2_0_0_ 
    FROM
        Dokument dokument0_ 
    WHERE
        dokument0_.pdoid=@p0;
    @p0 = 32783 [Type: Int64 (0:0:0)]
```

This is my model:

```
public class Base
{
   public virtual long Key { get; set; }

   public virtual bool IsDeleted { get; set; }
}

public class Asset : Base
{
   public virtual ISet<Dokument> Dokumente { get; set; }
}

public class Dokument : Base
{
   public virtual ISet<Asset> Assets { get; set; }
}
```

This is my mapping:
```
<class name="NHibernate.DomainModel.Base, NHibernate.DomainModel" abstract="true" dynamic-update="true" lazy="true">
    <id name="Key" column="pdoid" unsaved-value="0" >
      <generator class="hilo" />
    </id>

    <property name="IsDeleted" column="IsDeleted" />

    <filter name="deletedFilter" condition="IsDeleted = :deletedParam"/>

  </class>

<union-subclass name="NHibernate.DomainModel.Asset, NHibernate.DomainModel" table="Asset" lazy="true" extends="NHibernate.DomainModel.Base">
  
    <set name="Dokumente" table="asset_to_dokument" inverse="true" lazy="true">
       <key column="AssetId"/>
       <many-to-many class="NHibernate.DomainModel.Dokument, NHibernate.DomainModel" column="DokumentId">
         <filter name="deletedFilter" condition="IsDeleted = :deletedParam"/>
       </many-to-many>
     </set>
</union-subclass>

<union-subclass name="NHibernate.DomainModel.Dokument, NHibernate.DomainModel" table="Dokument" lazy="true" extends="NHibernate.DomainModel.Base">

    <set name="Assets" table="asset_to_dokument" inverse="false" lazy="true">
      <key column="DokumentId"/>
      <many-to-many class="NHibernate.DomainModel.Asset, NHibernate.DomainModel" column="AssetId"/>
    </set>
</union-subclass>

<filter-def name="deletedFilter">
    <filter-param name="deletedParam" type="bool"/>
</filter-def>

```

This is my LINQ Statement:
```
using (ITransaction t = s.BeginTransaction())
{
   s.EnableFilter("deletedFilter").SetParameter("deletedParam", false);
   var assets = s.Query<Asset>()
                       .FetchMany(x => x.Dokumente)
                       .ToList();
}
```

bahusoid:
Filter doesn't work properly for many-to-many  association in hql queries #1994

QueryOver generated query (works correctly):
```c#
s.QueryOver<Asset>().Fetch(SelectMode.Fetch, x => x.Documents).List<Asset>();
```
```sql
SELECT this_.pdoid AS pdoid1_0_1_
	,this_.IsDeleted AS isdeleted2_0_1_
	,documents2_.AssetId AS assetid1_2_3_ -- COLLECTION PERSISTER
	,document3_.pdoid AS documentid2_2_3_ -- COLLECTION PERSISTER.  REPLACED documents2_.DocumentID. In LINQ query it's still documents2_.DocumentID
	,document3_.pdoid AS pdoid1_0_0_
	,document3_.IsDeleted AS isdeleted2_0_0_
FROM Asset this_
LEFT OUTER JOIN asset_to_document documents2_ ON this_.pdoid = documents2_.AssetId
LEFT OUTER JOIN Document document3_ ON documents2_.DocumentId = document3_.pdoid
	AND document3_.IsDeleted = @p0
WHERE this_.IsDeleted = @p0
```

Query generated query (wrong):
```c#
s.Query<Asset>().FetchMany(x => x.Documents).ToList();
```
```sql
SELECT asset0_.pdoid AS pdoid1_0_0_
	,document2_.pdoid AS pdoid1_0_1_-- ENTITY PERISTER, NULL if filtered
	,asset0_.IsDeleted AS isdeleted2_0_0_
	,document2_.IsDeleted AS isdeleted2_0_1_
	,documents1_.AssetId AS assetid1_2_0__  -- COLLECTION PERSISTER
	,documents1_.DocumentId AS documentid2_2_0__  -- COLLECTION PERSISTER, NOT NULL even if filtered  so collection still tries to load filtered element
FROM Asset asset0_
LEFT OUTER JOIN asset_to_document documents1_ ON asset0_.pdoid = documents1_.AssetId
LEFT OUTER JOIN Document document2_ ON documents1_.DocumentId = document2_.pdoid
	AND document2_.IsDeleted = @p0
WHERE asset0_.IsDeleted = @p0
```
As you can see in QueryOver case all document related fields in select  statement are only from `Document` table with `document3` alias. And in query case `documents1_.DocumentId` is added from bridge table which I believe leads to this bug.

Implemented fix: substitute alias for collection persister with entity perister for filtered many-to-many collections. So documents for collection are populated from filtered entity table instead of bridge table.

Fixes #1994


hazzik:
@bahusoid do you work on a fix?

bahusoid:
No. I made a quick look  - hql query generated by completly different code parts which I know very little about.

But my findings is the following:
QueryOver select statement ended in the following `SelectFragment` calls:
https://github.com/nhibernate/nhibernate-core/blob/f5e2431f08a6df62ca735f1bf4bfb4c5546c9b21/src/NHibernate/Loader/JoinWalker.cs#L1095-L1116
This is the call that handles many-to-many case (see `BasicCollectionPersister.SelectFragment`)

And for hql it's different call:
https://github.com/nhibernate/nhibernate-core/blob/f5e2431f08a6df62ca735f1bf4bfb4c5546c9b21/src/NHibernate/Hql/Ast/ANTLR/Tree/FromElementType.cs#L221

I have no idea how provide all required data from this context

hazzik:
Cool, I'll try then

hazzik:
The query itself is correct. However, there is a second query issued to fetch documents:

```sql
QUERY START!!!
NHibernate: 
    select
        asset0_.pdoid as pdoid1_0_0_,
        document2_.pdoid as pdoid1_0_1_,
        asset0_.IsDeleted as isdeleted2_0_0_,
        document2_.IsDeleted as isdeleted2_0_1_,
        documents1_.AssetId as assetid1_2_0__,
        documents1_.DocumentId as documentid2_2_0__ 
    from
        Asset asset0_ 
    left outer join
        asset_to_document documents1_ 
            on asset0_.pdoid=documents1_.AssetId 
    left outer join
        Document document2_ 
            on documents1_.DocumentId=document2_.pdoid 
            and document2_.IsDeleted = @p0 
    where
        asset0_.IsDeleted = @p0;
    @p0 = False [Type: Boolean (0:0:0)]
NHibernate: 
    SELECT
        document0_.pdoid as pdoid1_0_0_,
        document0_.IsDeleted as isdeleted2_0_0_ 
    FROM
        Document document0_ 
    WHERE
        document0_.pdoid=@p0;
    @p0 = 4ece3d5c-dc7b-4998-b9a2-a9e2017532fa [Type: Guid (0:0:0)]
QUERY STOP!!!
```

hazzik:
btw, query over does not work as well when you have mixed documents.

bahusoid:
> The query itself is correct.

Maybe but QueryOver version is safer :)  `documents1_.DocumentId` is the source for filtered document id and it's missing in QueryOver.

bahusoid:
Well QueryOver case is also broken if `set` in mapping is replaced with `bag` due to 34c7e083795b7dd4546317ad884717721cce425d (which is about adding null to collection for each not found record if `not-found = "ignore"`)
Filtered element is also returned as null and added to collection. I think that this commit should be simply reverted. Why would anyone want to have null objects in collection? I would expect `not-found ="ignore"` to actually ignore not-found records and I would vote to make this default behavior.

And it seems `not-found="exception"` is also broken after 385a8611b0490590b42f4adbcbe8759d2a90de34  for fetched collections in QueryOver. Not found elements are treated as "ignore" - returned as null for bags and skipped for sets.

fredericDelaporte:
34c7e08 is [NH-750](https://nhibernate.jira.com/browse/NH-750), which was about not failing in case of a not-found element when `not-found` is set as `ignore`. Prior to NH-750, it was throwing. It is not explicitly about putting `null` in the collection in such case.

> I would expect `not-found="ignore"` to actually ignore not-found records and I would vote to make this default behavior.

It seems to me that the point of putting `null` in the collection or filtering out not-found elements was not discussed, but at least it matches [the documentation](https://nhibernate.info/doc/nhibernate-reference/collections.html#collections-ofvalues):

> `not-found` (optional - defaults to `exception`): specifies how foreign keys that reference missing rows will be handled. `ignore` will treat a missing row as a null association.

That is in fact just a copy-paste of the `one-to-many` case.

I agree it does not look very useful to have those `null` elements in the collection. But we have to be careful about the flush behavior. Not having them could only mean they would be deleted from the database. We have to check how it behaves currently. If currently, these not-found elements are left untouched by flushes (unless explicitly removed from the collection by the application), changing this could be a serious breaking change for some applications.

> And it seems `not-found="exception"` is also broken (looks like after 385a861 but I might be wrong) for fetched collections in QueryOver. Not found elements are treated as "ignore" - returned as null for bags and skipped for sets.

I have chosen to keep a left-join, instead of switching to an inner-join, explicitly for allowing to still throw for the `not-found` case. But I have not checked it was actually working, neither before nor after. I think it was already broken. This point should be handled as a separated issue I think.

bahusoid:
>It is not explicitly about putting null in the collection in such case.

Oups. Wrong commit. It's actually hard to pin-point exact commit. But I meant undo only this change which commented null check:
https://github.com/nhibernate/nhibernate-core/blob/f0f2a53fb778daa06dc7463625ff555ea6fb51ce/src/NHibernate/Collection/Generic/PersistentGenericBag.cs#L451-L455

>Not having them could only mean they would be deleted from the database. But we have to be careful about the flush behavior. 

I don't think that flush behavior can be changed in any way. As null objects have no connection to broken bridge records. But yeah we need more tests with many-to-many cases.

>I have chosen to keep a left-join, instead of switching to an inner-join,

Inner join will work only for lazy initialization. For fetches inner join is not an option. You don't want to filter owner elements without children.

It's my understading not-found exception throwing is handled by collection persister `AbstractCollectionPersister.ReadElement`. Collection pesister expect to have exact not found ID to throw (simply relies on IType.NullSafeGet without any specific checks). But now in QueryOver it's replaced with column from element table which is NULL if not found:
```sql
SELECT this_.pdoid AS pdoid1_0_1_
	,this_.IsDeleted AS isdeleted2_0_1_
	,documents2_.AssetId AS assetid1_2_3_ -- COLLECTION PERSISTER
	,document3_.pdoid AS documentid2_2_3_ -- COLLECTION PERSISTER.  REPLACED documents2_.DocumentID. In LINQ query it's still documents2_.DocumentID
	,document3_.pdoid AS pdoid1_0_0_
	,document3_.IsDeleted AS isdeleted2_0_0_
FROM Asset this_
LEFT OUTER JOIN asset_to_document documents2_ ON this_.pdoid = documents2_.AssetId
LEFT OUTER JOIN Document document3_ ON documents2_.DocumentId = document3_.pdoid
	AND document3_.IsDeleted = @p0
WHERE this_.IsDeleted = @p0
```
Didn't dig when this replacement was added but that's when `not-found ="exception"` case  was broken.

fredericDelaporte:
> Inner join will work only for lazy initialization. For fetches inner join is not an option. You don't want to filter owner elements without children.

That is a many-to-many, and the inner join would be on the one-to-many side of the association, which would not filter out the owner. (It would be "owner left-join (association-table inner-join children)".) Anyway, that is a bit off-topic.

> Didn't dig when this replacement was added but that's when `not-found="exception"` case was broken.

Ok, then it indeed looks like 385a861, which was done to handle correctly the property-ref case, where the association table does not hold the identifier, and so where it is wrong to use its foreign key as the identifier. It is then a 5.2 regression. It could be fixed by using the old code when there is no property-ref, but that would mean the property-ref case would not support `not-found="exception"` (which is still better than failing like it was prior to 385a861). It would be better to find a way to support all cases.

bahusoid:
>That is a many-to-many, and the inner join would be on the one-to-many side of the association, which would not filter out the owner

Hm.. But from what I know any inner join in SQL Server makes query to return only existing records. Nesting doesn't matter:
```sql
select * from Asset a
left outer join asset_to_document ad on ad.AssetId = a.pdoid
inner join Document d on d.pdoid = ad.DocumentId and d.IsDeleted = 0
```

For instance the SQL above would return only assets without deleted documents.  It won't return  assets without documents or with only deleted documents  as you suggest. Or did you have in mind some different query?

fredericDelaporte:
This:
```sql
select * from Asset a
left outer join (asset_to_document ad
inner join Document d on d.pdoid = ad.DocumentId and d.IsDeleted = 0) on ad.AssetId = a.pdoid
```
But not sure it works with all databases. (It works with SQL-Server.)

bahusoid:
Cool. But I doubt that's Nhibernate can generate such joins now.

fredericDelaporte:
Still off-topic, but I have further checked the assertion about an alleged `not-found` regression in 5.2.x: the behavior of not throwing despite of `not-found="exception"` is already found in 5.1.x on `many-to-many`. So that is not a regression of 5.2.x. (By the way 385a861 was not doing that replacement, that is #1872, more precisely the revert commit of #378, 5456d85.) It looks like this case is broken since longer. Anyway, for discussing more about it, a dedicated issue should be opened I think.

bahusoid:
Found an easy fix for this issue: do not mess with SQL generation as it's done in QueryOver/Entity Loader and instead substitute aliases for collection persister with entity persister for filtered many-to-many collection.

hazzik:
@bahusoid I made some changes, could you please verify?

bahusoid:
@hazzik  looks good to me

PR #1996, opened 2019-01-27
Enhance nullability check for "==" and "!=" operators for LINQ provider

### LINKED ISSUE #1860: LINQ "==" operator generates OR with IS NULL
When I do this query:

```
var query = from entity in session.Query<Entity>()
            where entity.Status != 2
            select entity;
```

NHibernate generate this SQL:

```
SELECT ...
FROM TABLE_NAME ALIAS
WHERE ALIAS.SITFIL <> 2 
       or ALIAS.SITFIL is null /*this 'or' is the problem*/
```

Why NHibernate do this?
The 'or' causes a performance issue on application I want to generate just:

```
SELECT ...
FROM TABLE_NAME ALIAS
WHERE ALIAS.SITFIL <> 2 
```

Without 'Or field is null'
The mapping not is nullable:

```
[Property(Column = "SITFIL", NotNull = true)]
public virtual TypeEnumOfProperty Status { get; set; }
```

The enum:

```
public enum TypeEnumOfProperty
{
    [Display(Name = "Ativa", Description ="Ativo")]
    Ativa = 0,
    [Display(Name = "Inativa", Description = "Inativo")]
    Inativa = 1,
    [Display(Name = "Inativa e oculta", Description = "Inativo e oculto")]
    InativaOculta = 2,
}
```
If I change the TypeEnumOfProperty to int the SQL is generated like I want, if change to TypeEnumOfProperty or long then was generated with the 'or'.

If I use 'equals' on comparision, NHibernate generates the where clause without 'or'.

maca88:
This PR adds a slightly complex logic that determines if it should add a `OR <Property> IS NULL` check for `==` and `!=` operators, fixes #1860. The logic will not add the additional null check for the following scenarios:
- The property is a value type.
Example: `s.Query<Order>().Where(o => o.Number != 5)`
- The property is mapped as not null.
Example: `s.Query<Order>().Where(o => o.Status != "NEW")`
- The property has a not null check before or after the `!=` operator.
Example: `s.Query<Order>().Where(o => o.Status != null && o.Status != "NEW")`
- The nullable property has a `HasValue` check before or after the `!=` operator.
Example: `s.Query<Order>().Where(o => o.Number.HasValue && o.Number != 3)`
- The expression for the `!=` operator is not null.
Example: `s.Query<Order>().Where(o => (o.NotNullProperty + o.NotNullProperty2) != "TEST")`
- Two nullable properties are compared with `==` operator and one of them or both have a not null check.
Example: `s.Query<Order>().Where(o => o.NullProperty != null && o.NullProperty == o.NullProperty2)`

There are some scenarios where the logic will add a null check where currently it is not added:
- Using a subquery with an aggregate method that can return null when the subquery has no records.
Example: `s.Query<Order>().Where(o => o.Rows.Max(r => r.Number) != 0)`
- Using a property that is translated to a sql function on a property that can be null.
Example: `s.Query<Order>().Where(o => o.Status.Length != 3)`

In addition, the `Equals` method was modified to use the same logic as the `==` operator.


src/NHibernate.Test/Async/Linq/NullComparisonTests.cs
```diff
@@ -260,13 +265,7 @@ public async Task NullEqualityWithNotNullAsync()
 			q = session.Query<AnotherEntityRequired>().Where(o => (o.NullableOutput + o.Output) == o.Output);
 			await (ExpectAsync(q, Does.Not.Contain("is null").IgnoreCase));
 
-			q = session.Query<AnotherEntityRequired>().Where(o => (o.Input + o.Output) == o.Output);
```
maca88: Changed to `o.Output + o.Output` as otherwise it fails in oracle as in oracle concatenating `null` with a non empty string will not result in a `null` value like in the other databases. I think that we should add an additional logic when concatenating strings in order to have a consistent behaviour across various databases and possible align with Linq-to-Object so that `null + "test"` will not result in a `null` value.

src/NHibernate.Test/Async/Linq/NullComparisonTests.cs
```diff
@@ -134,18 +132,6 @@ public async Task NullInequalityWithNotNullAsync()
 			q = session.Query<AnotherEntityRequired>().Where(o => (o.Input + o.Output) != o.Output);
 			await (ExpectAsync(q, Does.Not.Contain("is null").IgnoreCase, BothSame, BothDifferent));
 
-			q = session.Query<AnotherEntityRequired>().Where(o => o.RelatedItems.Count != 1);
-			await (ExpectAsync(q, Does.Not.Contain("is null").IgnoreCase));
-
-			q = session.Query<AnotherEntityRequired>().Where(o => o.RelatedItems.Max(r => r.Id) != 0);
-			await (ExpectAllAsync(q, Does.Contain("is null").IgnoreCase));
-
-			q = session.Query<AnotherEntityRequired>().Where(o => o.RelatedItems.Take(10).All(r => r.Output != null) != o.NullableBool);
```
maca88: Removed the `Take` call as it produces an invalid sql statement for Oracle.

src/NHibernate.DomainModel/Northwind/Mappings/AnotherEntityRequired.hbm.xml
```diff
@@ -0,0 +1,23 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<hibernate-mapping xmlns="urn:nhibernate-mapping-2.2" namespace="NHibernate.DomainModel.Northwind.Entities" assembly="NHibernate.DomainModel">
+  <class name="AnotherEntityRequired" table="AnotherEntity" mutable="false" schema-action="none">
+    <id name="Id">
+      <generator class="native" />
+    </id>
+    <property name="Output" not-null="true" />
+    <property name="Input" not-null="true" />
+    <property name="NullableOutput" formula="Output" lazy="true" />
```
maca88: I've had to use lazy as otherwise an invalid query is generated due to the missing table alias for the column used in the formula.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -458,10 +564,188 @@ private HqlTreeNode TranslateEqualityComparison(BinaryExpression expression)
 					_hqlTreeBuilder.IsNull(rhs2)));
 		}
 
-		static bool IsNullable(HqlExpression original)
+		private bool IsNullable(Expression expression, BinaryExpression equalityExpression)
```
hazzik: Would it be a clearer code if it put into a expression tree visitor?

maca88: Not really as the problem with the expression visitor is that its metohds are all void and converting for example `IsNullable(binaryExpression.Left) || IsNullable(binaryExpression.Right)` would require more code. Nonetheless, I simplified the code by removing the unneeded `while` statement, split the larger cases into separated methods and moved all logic into a separated class as `HqlGeneratorExpressionVisitor` is quite large without the nullable check.

src/NHibernate.Test/Linq/NullComparisonTests.cs
```diff
@@ -17,6 +18,285 @@ public class NullComparisonTests : LinqTestCase
...
+
+			q = session.Query<AnotherEntityRequired>().Where(o => o.RelatedItems.All(r => r.Output != null) != (o.NullableOutput.Length > 0));
+			Expect(q, Does.Not.Contain("or case").IgnoreCase);
+		}
+
+		[Test]
+		public void NullEqualityWithNotNull()
+		{
+			var q = session.Query<AnotherEntityRequired>().Where(o => o.Input == null);
+			Expect(q, Does.Not.Contain("or is null").IgnoreCase, OutputSet, BothNull);
```
fredericDelaporte: Can we ever have `or is null` ? Shouldn't we have an expression between `or` and `is null`?

src/NHibernate/Linq/Functions/BaseHqlGeneratorForMethod.cs
```diff
@@ -7,10 +7,17 @@
...
 
-        public abstract HqlTreeNode BuildHql(MethodInfo method, Expression targetObject, ReadOnlyCollection<Expression> arguments, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor);
-    }
-}
\ No newline at end of file
+		public abstract HqlTreeNode BuildHql(MethodInfo method, Expression targetObject, ReadOnlyCollection<Expression> arguments, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor);
+
+		public virtual bool AllowsNullableReturnType(MethodInfo method) => true;
+
+		bool IHqlGeneratorForMethodExtended.AllowsNullableReturnType(MethodInfo method)
```
fredericDelaporte: I do not get why an explicit implementation is needed here.

src/NHibernate/Linq/Visitors/NullableExpressionDetector.cs
```diff
@@ -0,0 +1,309 @@
...
+
+					return true;
+				case NhAggregatedExpression aggregatedExpression:
+					return aggregatedExpression.AllowsNullableReturnType;
+				default:
+					return true; // a query can return null and we cannot calculate it as it is not yet executed
+			}
+		}
+
+		private string TryGetEntityName(MemberExpression memberExpression)
```
maca88: After doing #2079 I found out that this method does not cover all cases. I will update this method and try to add additional tests that cover the edge cases.

src/NHibernate.Test/Linq/NullComparisonTests.cs
maca88: In this case the `Input` property is mapped as not null so the `is null` is not needed. The `or is null` is added when comparing a nullable property. I will add additional tests for nullable properties.

fredericDelaporte: My point is `or is null` is not a valid SQL expression, so I doubt it can ever be added into the generated query. So I do not think this assert can fail, be it with master state or with this PR state.

maca88: You are right, I didn't recognized that. Now it is fixed.

src/NHibernate/Linq/Functions/BaseHqlGeneratorForMethod.cs
maca88: Removed as it is not needed.

src/NHibernate/Persister/Entity/IEntityPersister.cs
```diff
@@ -623,5 +624,17 @@ public static int GetBatchSize(this IEntityPersister persister)
 
 			return 1;
 		}
+
+		internal static bool IsIdentifierMember(this IEntityPersister entityPersister, string memberPath)
```
maca88: Copied from #2079. Added as an extension method so that can be used by other expression visitors.

src/NHibernate/Tuple/Entity/EntityMetamodel.cs
```diff
@@ -411,14 +411,20 @@ private bool HasPartialUpdateComponentGeneration(Mapping.Component component)
 
 		private void MapPropertyToIndex(Mapping.Property prop, int i)
 		{
-			propertyIndexes[prop.Name] = i;
-			Mapping.Component comp = prop.Value as Mapping.Component;
-			if (comp != null)
+			MapPropertyToIndex(null, prop, i);
+		}
+
+		private void MapPropertyToIndex(string path, Mapping.Property prop, int i)
```
maca88: Copied from #2079. It is a fix for including nested components.

src/NHibernate/Util/ExpressionsHelper.cs
```diff
@@ -15,5 +17,40 @@ public static class ExpressionsHelper
 			}
 			return ((MemberExpression)expression.Body).Member;
 		}
+
+		internal static string TryGetEntityName(ISessionFactoryImplementor sessionFactory, MemberExpression memberExpression, out string memberPath)
```
maca88: Copied from #2079. Added as a static helper method so that can be used by other expression visitors.

src/NHibernate.DomainModel/Northwind/Mappings/User.hbm.xml
```diff
@@ -7,11 +7,19 @@
       <generator class="assigned" />
     </id>
 
-    <property name="Name" type="AnsiString" />
+    <property name="Name" type="AnsiString" not-null="true" />
     <property name="InvalidLoginAttempts" type="Int32" />
     <property name="RegisteredAt" type="DateTime" />
     <property name="LastLoginDate" type="DateTime" />
 
+    <many-to-one name="CreatedBy" class="User" not-null="true" lazy="false">
```
maca88: I had to add `lazy="false"`, otherwise some tests in `QueryReuseTests` (`CanReuseAfterFirst`, `CanReuseAfterFirstOrDefault`, `CanReuseAfterSingle`,  `CanReuseAfterSingleOfDefault`) would fail.

src/NHibernate/Persister/Entity/IEntityPersister.cs
maca88: This method is now replaced by `EntityMetamodel.GetIdentifierPropertyType`.

src/NHibernate/Persister/Entity/IEntityPersister.cs
```diff
@@ -10,6 +10,7 @@
 using System.Collections.Generic;
 using NHibernate.Intercept;
 using NHibernate.Util;
+using System.Linq;
```
hazzik: Prob needs to be reverted.

maca88: Now its gone.

maca88:
Rebased and fixed `TryGetEntityName` method to work with custom entity names. I am not sure what to do with the current CodeFactor issues, I think that we should allow those issues for the test project and increase the allowed complexity a bit :)

hazzik:
>I am not sure what to do with the current CodeFactor issues, I think that we should allow those issues for the test project and increase the allowed complexity a little bit :)

I will agree with CodeFactor that this method is too complex (and too hard for me to understand).

maca88:
Ok, I will try to make the method more readable by adding documentation, split into multiple methods, try to simplify the logic and add isolated tests.

maca88:
I tried my best to make the method more readable and to reduce its complexity in #2036. I've added several isolated tests in order to cover all kind of mappings. I think that the method is now more readable but the complexity didn't dropped, it even increased so I am not sure what to do about it. Should I split it into more smaller methods or replace the method with a [method object](https://sourcemaking.com/refactoring/replace-method-with-method-object)?

maca88:
Rebased and updated code to use `ExpressionHelper.TryGetMappedNullability`.

src/NHibernate.Test/Linq/NullComparisonTests.cs
```diff
@@ -17,6 +18,320 @@ public class NullComparisonTests : LinqTestCase
...
+			Expect(q, Does.Not.Contain("is null").IgnoreCase);
+
+			q = session.Query<AnotherEntityRequired>().Where(o => o.NullableAnotherEntityRequiredId.HasValue && o.NullableAnotherEntityRequiredId.Value != 0);
+			ExpectAll(q, Does.Not.Contain("is null").IgnoreCase);
+
+			q = session.Query<AnotherEntityRequired>().Where(o => o.NullableAnotherEntityRequiredId.HasValue || o.NullableAnotherEntityRequiredId.Value != 0);
+			ExpectAll(q, Does.Contain("is null").IgnoreCase);
+
+			q = session.Query<AnotherEntityRequired>().Where(o => o.NullableOutput != null && o.NullableOutput != "test");
+			Expect(q, Does.Not.Contain("is null").IgnoreCase, BothDifferent, BothSame, OutputSet);
```
fredericDelaporte: Inverting tests fails here. Shouldn'it be supported too?
```
q = session.Query<AnotherEntityRequired>().Where(o => o.NullableOutput != "test" && o.NullableOutput != null);
Expect(q, Does.Not.Contain("is null").IgnoreCase, BothDifferent, BothSame, OutputSet);
```
I have seen your summary is just listing the case:

> The nullable property has a HasValue check before the != operator.

But having such a check after yields the same result set and it does not make much sense to have the `is null` check in SQL here.
From a Linq-To-Object stand-point, you can argue that Linq would first consider null as passing then the "and also" statement would ruled it out, so in some way this is translating to SQL what does Linq-To-Object. But I think we should align on Linq-To-Object at most about which result set is emitted. I do not think we should align on it also for how it gets it.

Has this choice been done because it would be overly complex to handle otherwise?

src/NHibernate/Linq/Visitors/NullableExpressionDetector.cs
```diff
@@ -0,0 +1,288 @@
...
+		}
+
+		public void SearchForNotNullMemberChecks(BinaryExpression expression)
+		{
+			// Check for a member not null check that has a not equals expression
+			// Example: o.Status != null && o.Status != "New"
+			// Example: (o.Status != null && o.OldStatus != null) && (o.Status != o.OldStatus)
+			// Example: (o.Status != null && o.OldStatus != null) && (o.Status == o.OldStatus)
+			if (
+				expression.NodeType != ExpressionType.AndAlso ||
```
fredericDelaporte: I would include the `ExpressionType.And` case too here. I do not see any value in not optimizing the generated query for non-short-circuiting `&`. I think we should optimize all we can as long as the queries yielded result sets stay the same.

src/NHibernate.Test/Linq/NullComparisonTests.cs
maca88: > Inverting tests fails here. Shouldn'it be supported too?

It should be, but I guess I didn't considering them when I made this PR. The logic is now modified to support them.

> I have seen your summary is just listing the case
The nullable property has a HasValue check before the != operator.

Description updated.

src/NHibernate/Linq/Visitors/NullableExpressionDetector.cs
maca88: > I would include the ExpressionType.And case too here

`ExpressionType.And` is now added, but I wasn't able to add any tests as it seems that `QueryTranslatorImpl` does not support such cases.

fredericDelaporte: You mean, using simple `&` is not supported by the Linq provider? (Or maybe ambiguous with binary "and".) I have never checked that indeed.

maca88: > You mean, using simple & is not supported by the Linq provider?

It is supported but not when the left or the right expression is a binary expression. For example:
```
var list =  session.Query<Product>().Where(o => o.Discontinued & o.Discontinued).ToList(); // Works
var list2 =  session.Query<Product>().Where(o => o.Discontinued & o.Name != null).ToList(); // Throws
```
the first query does work but the second doesn't. It seems like that neither Sql Server 2017 supports it:
`SELECT * FROM Products where Discontinued = 1 & (ProductName != 'test')` fails with error: `Incorrect syntax near '!'.` ([fiddle](http://sqlfiddle.com/#!18/41386/1))

fredericDelaporte: I do not get the point with Sql Server. In T-SQL `&` is for "binary and", while `&` in C# has a dual usage : it may be the binary operator or the non-short-circuiting "logical and" (when both side are boolean expressions), which is just `and` in T-SQL. T-SQL does not have a short-circuiting "logical and" operator.
The correct translation for C# `&` when both sides are boolean is the T-SQL `and`, not `&`. So it should be supported I think, but that is not the subject of this PR.

Overall I do not think we need to change anything more in this PR about this, so I am off-topic.

hazzik: > T-SQL does not have a short-circuiting "logical and" operator.

This is because this concept does not apply to the SQL server.

maca88: You are right, I got confused. Indeed, this should be covered in a separate PR.

fredericDelaporte: >>T-SQL does not have a short-circuiting "logical and" operator.
>
> This is because this concept does not apply to the SQL server.

And to SQL at all I think indeed. It is more an iterative concept than a thing for data manipulation.

PR #1999, opened 2019-01-30
Optimize DistinctRootEntityResultTransformer

bahusoid:
Use comparer instead of object wrapping
Added shortcut exit
Avoid ToList conversion for LINQ

### DIFF
--- a/src/NHibernate/Transform/DistinctRootEntityResultTransformer.cs
+++ b/src/NHibernate/Transform/DistinctRootEntityResultTransformer.cs
@@ -12,24 +12,16 @@ namespace NHibernate.Transform
 		private static readonly INHibernateLogger log = NHibernateLogger.For(typeof(DistinctRootEntityResultTransformer));
 		private static readonly object Hasher = new object();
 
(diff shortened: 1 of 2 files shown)


src/NHibernate/Transform/DistinctRootEntityResultTransformer.cs
```diff
@@ -77,13 +72,10 @@ public bool IsTransformedValueATupleElement(String[] aliases, int tupleLength)
...
-			{
-				return false;
-			}
-			// NH-3957: do not rely on hashcode alone.
-			// Must be the exact same type
-			return obj.GetType() == typeof(DistinctRootEntityResultTransformer);
+			if (ReferenceEquals(obj, this))
+				return true;
+
+			return obj is DistinctRootEntityResultTransformer;
```
hazzik: This is not equivalent because the class is not sealed.

bahusoid: this equality issue affects all transformers. I'm going to fix it in separate pr

PR #2000, opened 2019-01-31
Fixed Equals method for transformers

bahusoid:
Fixed possible equality issue:
```C#
public class MyTransformer: DistinctRootEntityResultTransformer {}
var myTransformer = new MyTransformer();
var distinctTransformer = new DistinctRootEntityResultTransformer();
myTransformer.Equals(distinctTransformer); // true. Should be false
distinctTransformer.Equals(myTransformer); // false
```
Also reuse instances when possible

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3957/ResultTransformerEqualityFixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3957/ResultTransformerEqualityFixture.cs
@@ -18,22 +18,22 @@ namespace NHibernate.Test.NHSpecificTest.NH3957
 		{
 			var transf1 = new AliasToEntityMapResultTransformer();
 			var transf2 = new AliasToEntityMapResultTransformer();
-			var custom = new CustomAliasToEntityMapResultTransformer();
-			HashSet<IResultTransformer> set = new HashSet<IResultTransformer>() { transf1, transf2, custom, };
+			HashSet<IResultTransformer> set = new HashSet<IResultTransformer>() { transf1, transf2, };
 
-			Assert.That(set.Count, Is.EqualTo(2));
+			Assert.That(set.Count, Is.EqualTo(1));
 			Assert.IsTrue(transf1.Equals(transf2));
 			Assert.IsTrue(transf2.Equals(transf1));
-			Assert.False(custom.Equals(transf1));
-			Assert.False(transf1.Equals(custom));
 		}
 
 		[Test]
 		public void AliasToEntityMapAndDistinctRootEntityInequality()
 		{
 			var transf1 = new AliasToEntityMapResultTransformer();
-			var transf2 = new DistinctRootEntityResultTransformer();
+			var transf2 = new CustomAliasToEntityMapResultTransformer();
+			HashSet<IResultTransformer> set = new HashSet<IResultTransformer>() { transf1, transf2, };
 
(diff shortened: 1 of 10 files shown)


src/NHibernate.Test/NHSpecificTest/NH3957/ResultTransformerEqualityFixture.cs
```diff
@@ -8,55 +7,25 @@ namespace NHibernate.Test.NHSpecificTest.NH3957
...
-		{
-			// Restore those types hashcode. (Avoid impacting perf of other tests, avoid second level query cache
-			// issues if it was holding cached entries (but would mean some tests have not cleaned up properly).)
-			foreach(var backup in _hasherBackup)
-			{
-				TweakHashcode(backup.Key, backup.Value);
-			}
-		}
-
+		public class CustomAliasToEntityMapResultTransformer : AliasToEntityMapResultTransformer { }
```
bahusoid: This custom class provides also hashcode collision with current implementaion. Not sure why TweakHashcode was required

fredericDelaporte: See #564, it is required to check that two different result transformers will not be considered equal if they happen to have the same hashcode, which was NH-3957 bug. The collision for different instances of the same transformer is intentional and mandatory since they are considered equal.

Removing the tweak causes the `Inequality` suffixed tests to no more test the NH-3957 bug. You should change them to compare with the corresponding `Custom` descendant you have added. That is still a step away from what was NH-3957, but I do not see any other solution with the `Instance` optimization. NH-3957 was a random hashcode collision between different unrelated transformers causing them to wrongly consider themselves equals.

bahusoid: >it is required to check that two different result transformers will not be considered equal if they happen to have the same hashcode,.. Removing the tweak causes the Inequality suffixed tests to no more test the NH-3957 bug

But I'm already checking `Inequality`:
[`Assert.False(custom.Equals(transf1));`](https://github.com/nhibernate/nhibernate-core/pull/2000/files#diff-012db71c0b0b1c67464a20c007d76d40R53)
[`Assert.False(transf1.Equals(custom));`](https://github.com/nhibernate/nhibernate-core/pull/2000/files#diff-012db71c0b0b1c67464a20c007d76d40R54)

 I've just put those checks in the wrong tests.
>That is still a step away from what was NH-3957

I still see no difference between inequality check for custom implemented classes with collisions to existing ones ( `custom` and `transf1`) and TweakHashcode hack for existing classes. IMHO they still fully cover the case:
>that two different result transformers will not be considered equal if they happen to have the same hashcode

src/NHibernate/Linq/IntermediateHqlTree.cs
```diff
@@ -110,7 +110,7 @@ public void AddDistinctRootOperator()
 			if (!_hasDistinctRootOperator)
 			{
 				Expression<Func<IEnumerable<object>, IList>> x =
-					l => new DistinctRootEntityResultTransformer().TransformList(l.ToList());
+					l => DistinctRootEntityResultTransformer.Instance.TransformList(l.ToList());
```
bahusoid: I also want to get rid of this ToList conversion. Will undo this change and make it in my other PR

PR #2006, opened 2019-02-06
Skip Topological sorting if not required

bahusoid:


bahusoid:
Hm. It seems there is no need to split dependent alias calculation and sorting across different methods. I'm going to remove `_dependentAliases` field

src/NHibernate/Loader/JoinWalker.cs
```diff
@@ -840,6 +851,21 @@ protected JoinFragment MergeOuterJoins(IList<OuterJoinableAssociation> associati
 			return outerjoin;
 		}
 
+		private IList<OuterJoinableAssociation> GetSortedAssociations(IList<OuterJoinableAssociation> associations)
+		{
+			var indexes = GetTopologicalSortOrder(associations);
+			if (indexes == null)
```
hazzik: Can we remove null-driven logic here? Eg move conditions and dependents calculation from `GetTopologicalSortOrder` to here.

src/NHibernate/Loader/JoinWalker.cs
```diff
@@ -840,6 +851,21 @@ protected JoinFragment MergeOuterJoins(IList<OuterJoinableAssociation> associati
 			return outerjoin;
 		}
 
+		private IList<OuterJoinableAssociation> GetSortedAssociations(IList<OuterJoinableAssociation> associations)
```
hazzik: This can be made static.

hazzik:
Can you also review #419?

bahusoid:
Regarding #419. I know why and what this sorting does. But I have no idea how :) So I can't really say if it's any better then existing algorithm.

src/NHibernate/Loader/JoinWalker.cs
bahusoid: Done

bahusoid:
Something is fishy with MergeDeliveryNodeAsync test. It's failed to often on appveyor and takes  7 min when failed

PR #2010, opened 2019-02-16
Add new collection operation queue mechanism

maca88:
The main purpose of this PR is to minimize flushes when several operations are done to an uninitialized collection. Currently, a flush will occur after a second operation as [already explained](https://github.com/nhibernate/nhibernate-core/issues/1874#issuecomment-431030704) by @fredericDelaporte, which can lead to various unexpected behavior like throwing a `PropertyValueException` with a message `not-null property references a null or transient value`, when adding an item to the collection or checking its existence, which occurred to me several times in the past.
The new mechanism in this PR replaces the current one which eliminates the need for a flush for all collections except for two cases when using `PersistentGenericList<>`:
1. Using an indexed operation after a non indexed one (e.g. use `IList<>.RemoveAt` after using `IList<>.Add`)
2. Using `IList<>.Remove` after using an indexed operation (e.g. use `IList<>.Remove` after using `IList<>.Insert`)

We could eliminate the flushing also for the above cases if we would add an additional method to `ICollectionPersister`, which would return the index of the given element that could be also used to queue `IList<>.IndexOf` operation. This can be done in a separate PR if desired.

In addition an optimization was done for `PersistentGenericSet<>` when using `ICollection<>.Add` or `ISet<>.Add` for a transient entity (only when we are able to detect that it is transient) which will not trigger a query for checking its existence.

This PR contains several breaking changes (applies for collections mapped with `inverse="true"`):
- ~~Calling `IList.Add` for an uninitialized collection mapped as `lazy="true"` will trigger the collection initialization and return the correct index, currently the collection is not initialized and `-1` is returned (applies for Bag and List)~~
- ~~Calling `IList.Add` for an uninitialized collection mapped as `lazy="extra"` will return the correct index, currently `-1` is returned (applies for Bag and List)~~
- Calling `IList.RemoveAt` or `IList<>.RemoveAt` with a negative number will throw an `ArgumentOutOfRangeException` in order to mimic `List<>.RemoveAt` behavior, currently it throws a `IndexOutOfRangeException` (applies for List)
- Calling `IList.RemoveAt` or `IList<>.RemoveAt` with a number that is equal or higher that the current collection size for an uninitialized collection mapped as `lazy="extra"` will throw an `ArgumentOutOfRangeException`, currently it allows to queue the operation without throwing an exception (applies for List)
- Calling `IList.Insert` or `IList<>.Insert` with a negative number will throw an `ArgumentOutOfRangeException` in order to mimic `List<>.Insert` behavior, currently it throws an `IndexOutOfRangeException` (applies for List)
- Calling `IList.Insert` or `IList<>.Insert` with a number that is higher that the current collection size for an uninitialized collection mapped as `lazy="extra"` will throw an `ArgumentOutOfRangeException`, currently it allows to queue the operation without throwing an exception (applies for List)
- Getting or setting a value for `IList.this[int index]` or `IList<>.this[int index]` with a negative number will throw an `ArgumentOutOfRangeException` in order to mimic `List<>.this[int index]` behavior, currently it throws an `IndexOutOfRangeException` (applies for List)
- setting a value for `IList.this[int index]` or `IList<>.this[int index]` with a number that is equal or higher that the current collection size for an uninitialized collection mapped as `lazy="extra"` will throw an `ArgumentOutOfRangeException`, currently it allows to queue the operation without throwing an exception (applies for List)
- Calling `IDictionary<,>.Add` or `ICollection<>.Add` for an uninitialized collection mapped as `lazy="extra"` with a key that already exists will throw an `ArgumentException`, currently it allows to queue the operation without throwing an exception (applies for Map)
- Calling `IDictionary<,>.Remove` or `ICollection<>.Remove` for an uninitialized collection mapped as `lazy="extra"` with a key that does not exist will return false, currently it returns true (applies for Map)
- Collection dirtiness will be evaluated by `EqualityComparer<TValue>.Default` when setting an existing key value with `IDictionary<,>.this[]` for an initialized collection, currently the old and the new value are compared by using `ReferenceEquals` (applies for Map)
- Calling `ISet<>.Add` for an uninitialized collection mapped as `lazy="extra"` with a transient element that already exists in the set will return false, currently it returns true (applies for Set)


src/NHibernate/Collection/Trackers/BagQueueOperationTracker.cs
```diff
@@ -0,0 +1,61 @@
...
+					// Since this condition happens with transient instances added in the bag then saved, ReferenceEquals
+					// is enough to match them.
+					// This solution is a workaround, the root cause is not fixed. The root cause is the insertion on save
+					// done without caring for pending operations of one-to-many collections. This root cause could be fixed
+					// by triggering a full flush there before the insert (currently it just flushes pending inserts), or
+					// maybe by flushing just the dirty one-to-many non-initialized collections (but this can be tricky).
+					// (It is also likely one-to-many lists have a similar issue, but nothing is done yet for them. And
+					// their case is more complex due to having to care for the indexes and to handle many more delayed
+					// operation kinds.)
+					if (CollectionPersister.IsOneToMany && loadedCollection.Any(l => ReferenceEquals(l, toAdd)))
```
maca88: Copied from `PersistentGenericBag<>.SimpleAddDelayedOperation`. Instead of the proposed fixes I will try to solve the issue by removing the queued operation for the entity that is immediately inserted because of the id generator from all uninitialized collections that currently contain it. I will try to make a separated PR on top of this one if the mentioned solution will work.

fredericDelaporte:
From reading this PR description, it seems to me that this PR will cause more operations on extra-lazy collections to cause some database queries, for the sake of better honouring the collection contract.

Returning the correct index when adding to an uninitialized list is surely done by initializing the list count of elements by querying this count in the database. (I guess the next adds, or checks for insert/remove at/set at index range, will not re-trigger a count in db.)
So with this PR, the first add/insert/remove at/set at on an uninitialized extra lazy indexed collection will trigger a count, while I guess it was not doing that previously.

Returning `false` when adding to a set an element it already has surely requires to query the db for its existence in the set. So each add on uninitialized set would trigger a check in db for the element existence.
Same for immediately throwing in the map case.

The only documentation for extra lazy seems currently to be:

> _"Extra-lazy" collection fetching_ - individual elements of the collection are accessed from the database as needed. NHibernate tries not to fetch the whole collection into memory unless absolutely needed (suitable for very large collections) 

(From [here](https://nhibernate.info/doc/nhibernate-reference/performance.html#performance-fetching).)

So, nothing contradicting what this PR does.

But it seemed to me there was a choice made for extra lazy to somewhat break collection contracts when it allows avoiding a database query.
For the extra `count` query, presumably occurring only on first operation, well, why not.
But for set or map, having a query triggered at each add does quite undermine the usefulness of extra lazy. Maybe we should document those contract violations instead of fixing them.

maca88:
> Returning the correct index when adding to an uninitialized list is surely done by initializing the list count

Correct.

> I guess the next adds, or checks for insert/remove at/set at index range, will not re-trigger a count in db

Again correct, the count query is triggered only once when the requested operation needs it and it is stored for afterward usage.

> So with this PR, the first add/insert/remove at/set at on an uninitialized extra lazy indexed collection will trigger a count

That is partially correct, `Add` will not trigger a count if it is the first operation to be executed as it does not need it. The List queue mechanism is composed by two tracker (`NonIndexedListQueueOperationTracker` and `IndexedListQueueOperationTracker`) and the first operation will decide which tracker will be used. So when `Add` is the first operation, the `NonIndexedListQueueOperationTracker` will be used which does not use `Count` and when an indexed operation is triggered then a flush is done and the tracker is then switched to `IndexedListQueueOperationTracker` which does use the `Count` for the `Add` operation.

> Returning `false` when adding to a set an element it already has surely requires to query the db for its existence in the set

I've updated the set breaking change as it was wrongly written. The query for checking an element existance is already done by the current code:
https://github.com/nhibernate/nhibernate-core/blob/6cb478fd7644507941140dd05b33e2fa9f66702e/src/NHibernate/Collection/Generic/PersistentGenericSet.cs#L308-L325

The problem with the current code is when a new entity is added twice in the set, both times `true` is returned (line 325) when it should be returned only the first time. 

> Still I guess that in one-to-many cases, where an element cannot appears many times in the list, the returned/checked index will be wrong if some add/insert operation is done on an element already in the list, followed by some more adds/inserts/removes.

I didn't test this scenario yet, but the same issue applies for the current code as there is no existance check when adding an item to an indexed collection.

> But for a set or map, having a query triggered at each add does quite undermine the usefulness of extra lazy.

As already mentioned this check is already in the current code and that is why I optimized `ICollection<>.Add` for the set in order to skip that check when possible.

bahusoid:
>* Calling `IList.Add` for an uninitialized collection mapped as `lazy="true"` will trigger the collection initialization and return the correct index, currently the collection is not initialized and `-1` is returned (applies for Bag and List)

This potentially could be a pretty serious drawback. As if I understand correctly there is now no way to "lazy add" via `IList` interface.

maca88:
> As if I understand correctly there is now no way to "lazy add" via `IList` interface.

It is possible only when the collection is mapped as `lazy="extra"`.

maca88:
With the latest commit the set optimization for `ICollection<>.Add` is now also applied for `ISet<>.Add`.

maca88:
> Still I guess that in one-to-many cases, where an element cannot appears many times in the list, the returned/checked index will be wrong if some add/insert operation is done on an element already in the list, followed by some more adds/inserts/removes.

I've added some tests that show how this PRs works when an element is added multiple times. The behavior is indeed different that the current one for lists as currently, when checking the `Count` of the list or when retrieving an element by its index a flush is triggered that will clear the queue and take care of duplicate elements. With this PR the flush is not triggered, which means that the collection will work exactly like `List<>` does, `Count` will return the number of all elements including duplicated. In order to obtain the old behavior in this case, the user would need to trigger the flush manually before checking the `Count` of the list or retrieving an element by its index. If we would like to mimic the old behavior for lists and bags we would need to check the element existance for each addition, similar as for a set, which is not worth it. For me, adding an element multiple times into a list is an user error.

src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs
```diff
@@ -156,6 +156,92 @@ public void ListAdd(bool initialize)
...
+				Assert.That(NHibernateUtil.IsInitialized(gavin.Companies), Is.False);
+
+				// Readd items
+				Sfi.Statistics.Clear();
+				for (var i = 0; i < 5; i++)
+				{
+					gavin.Companies.Add(addedItems[i]);
+				}
+
+				Assert.That(gavin.Companies.Count, Is.EqualTo(10));
```
maca88: By running the test with the current code, `Count` would be equal to `5`.

src/NHibernate.Test/Async/Extralazy/ExtraLazyFixture.cs
```diff
@@ -34,13 +38,2207 @@ protected override string CacheConcurrencyStrategy
...
+
+				await (t.CommitAsync(cancellationToken));
+			}
+		}
+
+		[TestCase(false, false)]
+		[TestCase(false, true)]
+		[TestCase(true, false)]
+		[TestCase(true, true)]
+		public async Task ListAddDuplicatedAsync(bool initialize, bool flush, CancellationToken cancellationToken = default(CancellationToken))
```
hazzik: @maca88 can you please update AsyncGenerator's NUnit plugin not to generate cancellation token for `[TestCase]`?

maca88: Done. Also updated the NHibernate configuration #2126.

hazzik: I just realised that this is because the test method lacks `[Theory]` attribute.

hazzik:
@fredericDelaporte @bahusoid any conclusions on this PR?

bahusoid:
> * Calling IList.Add for an uninitialized collection mapped as lazy="true" will trigger the collection initialization and return the correct index, currently the collection is not initialized and -1 is returned (applies for Bag and List)
> * Calling IList.Add for an uninitialized collection mapped as lazy="extra" will return the correct index, currently -1 is returned (applies for Bag and List)

As I've already said I don't like both changes. IMHO it limits functionality (kills delayed add feature for lazy `IList` and triggers in most cases unnecessary query for extra lazy `IList` )  and adds no practical value as you can retrieve count via `IList.Count`. It's the case when contract violation is better than unsupported feature.

maca88:
@bahusoid I will revert those changes and update the tests.

maca88:
Reverted `IList.Add` behavior and updated the breaking changes.

src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs
```diff
@@ -23,12 +26,2206 @@ protected override string CacheConcurrencyStrategy
...
+				}
+
+				Assert.That(gavin.Companies.Count, Is.EqualTo(15));
+				Assert.That(Sfi.Statistics.FlushCount, Is.EqualTo(0));
+				Assert.That(Sfi.Statistics.PrepareStatementCount, Is.EqualTo(0));
+				Assert.That(NHibernateUtil.IsInitialized(gavin.Companies), Is.False);
+
+				// Try insert invalid indexes
+				Assert.Throws<ArgumentOutOfRangeException>(() => gavin.Companies.RemoveAt(-1));
+				Assert.Throws<ArgumentOutOfRangeException>(() => gavin.Companies.RemoveAt(20));
```
fredericDelaporte: Insert? This looks like removes.

src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs
```diff
@@ -23,12 +26,2206 @@ protected override string CacheConcurrencyStrategy
...
+
+				t.Commit();
+			}
+		}
+
+		[TestCase(false, false)]
+		[TestCase(false, true)]
+		[TestCase(true, false)]
+		[TestCase(true, true)]
+		public void ListInsertDuplicated(bool initialize, bool flush)
```
fredericDelaporte: I would also set this as explicitly run only.

hazzik:
[CodeFactor](https://www.codefactor.io/repository/github/nhibernate/nhibernate-core) found multiple issues last seen at febd96f3b25cc43d6c6f060fef7f7f2bdb11f7e7:

#### The code must not contain multiple blank lines in a row.
[src\NHibernate.Test\Extralazy\ExtraLazyFixture.cs:1820
](https://github.com/nhibernate/nhibernate-core/blob/febd96f3b25cc43d6c6f060fef7f7f2bdb11f7e7/src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs#L1820)[src\NHibernate.Test\Extralazy\ExtraLazyFixture.cs:106
](https://github.com/nhibernate/nhibernate-core/blob/febd96f3b25cc43d6c6f060fef7f7f2bdb11f7e7/src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs#L106)[src\NHibernate.Test\Extralazy\ExtraLazyFixture.cs:1031
](https://github.com/nhibernate/nhibernate-core/blob/febd96f3b25cc43d6c6f060fef7f7f2bdb11f7e7/src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs#L1031)[src\NHibernate.Test\Extralazy\ExtraLazyFixture.cs:1489
](https://github.com/nhibernate/nhibernate-core/blob/febd96f3b25cc43d6c6f060fef7f7f2bdb11f7e7/src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs#L1489)
#### Complex Code
[src\NHibernate.Test\Extralazy\ExtraLazyFixture.cs
](https://github.com/nhibernate/nhibernate-core/blob/febd96f3b25cc43d6c6f060fef7f7f2bdb11f7e7/src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs)[src\NHibernate\Collection\Generic\PersistentGenericList.cs
](https://github.com/nhibernate/nhibernate-core/blob/febd96f3b25cc43d6c6f060fef7f7f2bdb11f7e7/src/NHibernate/Collection/Generic/PersistentGenericList.cs)
#### Insert parentheses within the conditional AND and OR expressions to declare the operator precedence.
[src\NHibernate\Collection\AbstractPersistentCollection.cs:1027
](https://github.com/nhibernate/nhibernate-core/blob/febd96f3b25cc43d6c6f060fef7f7f2bdb11f7e7/src/NHibernate/Collection/AbstractPersistentCollection.cs#L1027)
#### A C# document may only contain a single class at the root level unless all of the classes are partial and are of the same type.
[src\NHibernate\Collection\Trackers\AbstractCollectionQueueOperationTracker.cs:78
](https://github.com/nhibernate/nhibernate-core/blob/febd96f3b25cc43d6c6f060fef7f7f2bdb11f7e7/src/NHibernate/Collection/Trackers/AbstractCollectionQueueOperationTracker.cs#L78)

src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs
```diff
@@ -23,12 +26,2206 @@ protected override string CacheConcurrencyStrategy
...
+				Assert.That(Sfi.Statistics.PrepareStatementCount, Is.EqualTo(0));
+				Assert.That(NHibernateUtil.IsInitialized(gavin.Companies), Is.False);
+
+				// Test adding companies with IList interface
+				Sfi.Statistics.Clear();
+				for (var i = 10; i < 15; i++)
+				{
+					var item = new Company($"c{i}", i, gavin);
+					addedItems.Add(item);
+					Assert.That(((IList) gavin.Companies).Add(item), Is.EqualTo(-1));
```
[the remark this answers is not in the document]
maca88: > I think it would be better to not assert anything here

I do agree, the assert is now removed.

src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs
```diff
@@ -23,12 +26,2206 @@ protected override string CacheConcurrencyStrategy
...
+
+				// Check existance of added companies
+				Sfi.Statistics.Clear();
+				foreach (var item in addedItems.Skip(5))
+				{
+					Assert.That(gavin.Companies.Contains(item), Is.True);
+				}
+
+				Assert.That(Sfi.Statistics.FlushCount, Is.EqualTo(0));
+				Assert.That(Sfi.Statistics.PrepareStatementCount, Is.EqualTo(0));
```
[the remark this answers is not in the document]
maca88: Added labels to all of them.

src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs
```diff
@@ -23,12 +26,2206 @@ protected override string CacheConcurrencyStrategy
...
+
+				t.Commit();
+			}
+		}
+
+		[TestCase(false, false)]
+		[TestCase(false, true)]
+		[TestCase(true, false)]
+		[TestCase(true, true)]
+		public void ListAddDuplicated(bool initialize, bool flush)
```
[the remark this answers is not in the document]
maca88: Marked as `Explicit`.

src/NHibernate.Test/Extralazy/ExtraLazyFixture.cs
maca88: Marked as `Explicit`.

src/NHibernate/Collection/Trackers/MapQueueOperationTracker.cs
```diff
@@ -0,0 +1,212 @@
...
+			}
+
+			if (existsInDb == true)
+			{
+				GetOrCreateOrphanMap()[elementKey] = oldElement;
+			}
+
+			// We don't want to have non database keys in the removal queue
+			if (_queue?.Remove(elementKey) == true |
+				(_orphanMap?.ContainsKey(elementKey) == true && GetOrCreateRemovalQueue().Add(elementKey)))
```
fredericDelaporte: Why can't we just use `existsInDb == true` instead of `_orphanMap?.ContainsKey(elementKey) == true`?

src/NHibernate/Collection/Trackers/MapQueueOperationTracker.cs
```diff
@@ -0,0 +1,212 @@
...
+		/// <inheritdoc />
+		public override bool TryGetDatabaseElementByKey(TKey elementKey, out TValue element)
+		{
+			if (_orphanMap == null)
+			{
+				element = default(TValue);
+				return false;
+			}
+
+			return _orphanMap.TryGetValue(elementKey, out element);
```
fredericDelaporte: In which case is-it useful to retrieve a persisted element from the to be orphaned ones? The only place where `TryGetDatabaseElementByKey` is currently called does previously check `IsElementKeyQueuedForDelete` and short-circuit in such case. So it looks to me this `TryGetDatabaseElementByKey` method is unneeded and could be removed.

src/NHibernate/Collection/Trackers/SetQueueOperationTracker.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+using NHibernate.Persister.Collection;
+
+namespace NHibernate.Collection.Trackers
+{
+	/// <summary>
+	/// A tracker that is able to track changes that are done to an uninitialized map.
```
fredericDelaporte: `set`, not `map` here.

src/NHibernate/Collection/Generic/PersistentGenericBag.cs
```diff
@@ -501,6 +546,8 @@ private static int CountOccurrences(object element, IEnumerable list, IType elem
 			return result;
 		}
 
+		// Since v5.3
+		[Obsolete("This class has no more usages in NHibernate and will be removed in a future version.")]
 		private sealed class ClearDelayedOperation : IDelayedOperation
```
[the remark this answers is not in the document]
maca88: Even if they are private the `Obsolete` attribute was needed in order to compile the code. As those classes are not needed anymore, I have removed them.

src/NHibernate/Collection/Generic/PersistentGenericBag.cs
```diff
@@ -119,15 +133,26 @@ int IList.IndexOf(object value)
...
-							? _gbag.Count
-							: 0;
-			return count - 1;
+			var val = (T) value;
+			var queueOperationTracker = GetOrCreateQueueOperationTracker();
+			if (queueOperationTracker != null)
+			{
+				QueueAddElement(val);
+			}
+			else
```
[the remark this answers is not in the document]
maca88: > CreateQueueOperationTracker for returning null, or I am wrong?

You are correct and I do agree that the checks are not needed for the derived classes, so they are now removed.

src/NHibernate/Collection/Trackers/AbstractCollectionQueueOperationTracker.cs
```diff
@@ -0,0 +1,87 @@
...
+		/// <param name="element">The output variable for the element.</param>
+		/// <returns>True whether the element was found, false otherwise.</returns>
+		public abstract bool TryGetElementAtIndex(int index, out T element);
+
+		/// <summary>
+		/// Calculates the element index where it currently lies in the database by taking into the consideration the queued operations.
+		/// </summary>
+		/// <param name="index">The effective index that will be when all operations would be flushed.</param>
+		/// <returns>The element index in the database or -1 if the index represents a transient element.</returns>
+		public abstract int CalculateDatabaseElementIndex(int index);
```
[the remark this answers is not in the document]
maca88: Renamed to `GetDatabaseElementIndex`.

src/NHibernate/Collection/Trackers/AbstractQueueOperationTracker.cs
```diff
@@ -0,0 +1,106 @@
...
+			nameof(AbstractCollectionQueueOperationTracker<int>.RemoveElementAtIndex),
+			nameof(AbstractCollectionQueueOperationTracker<int>.AddElementAtIndex),
+			nameof(AbstractCollectionQueueOperationTracker<int>.SetElementAtIndex),
+			nameof(AbstractCollectionQueueOperationTracker<int>.TryGetElementAtIndex)
+		};
+
+		/// <summary>
+		/// The number of elements that the collection have in the database.
+		/// </summary>
+		public virtual int DatabaseCollectionSize { get; protected internal set; } = -1;
```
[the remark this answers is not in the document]
maca88: Modified the return type to `int?`.

src/NHibernate/Collection/Trackers/ClearedListQueueOperationTracker.cs
```diff
@@ -0,0 +1,142 @@
...
+				{
+					loadedCollection.Add(toAdd);
+				}
+			}
+		}
+
+		/// <inheritdoc />
+		public override int CalculateDatabaseElementIndex(int index)
+		{
+			return -1;
```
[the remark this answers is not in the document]
maca88: Modified the return type to int?.

src/NHibernate/Collection/Generic/PersistentGenericMap.cs
```diff
@@ -108,6 +115,13 @@ public override void BeforeInitialize(ICollectionPersister persister, int antici
 			WrappedMap = (IDictionary<TKey, TValue>)persister.CollectionType.Instantiate(anticipatedSize);
 		}
 
+		public override void ApplyQueuedOperations()
+		{
+			var queueOperation = (MapQueueOperationTracker<TKey, TValue>) QueueOperationTracker;
```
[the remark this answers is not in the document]
maca88: Now it is casted to `AbstractMapQueueOperationTracker`.

src/NHibernate/Collection/Trackers/MapQueueOperationTracker.cs
```diff
@@ -0,0 +1,212 @@
...
+			{
+				return _queue?.Remove(elementKey) ?? false;
+			}
+
+			// We can have the following scenarios:
+			// 1. remove a key that exists in db and it is not in the queue and removal queue (decrease queue size)
+			// 2. remove a key that exists in db and it is in the queue (decrease queue size)
+			// 3. remove a key that does not exist in db and it is not in the queue (don't decrease queue size)
+			// 4. remove a key that does not exist in db and it is in the queue (decrease queue size)
+			// 5. remove a key that exists in db and it is in the removal queue (don't decrease queue size)
```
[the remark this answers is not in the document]
maca88: Removed `it` from the comments.

src/NHibernate/Collection/Trackers/MapQueueOperationTracker.cs
maca88: When removing an element that exists in the database twice (remove, add, remove), in the second remove `existsInDb` will be `null` because the element is in the queue, that is why we cannot rely on `existsInDb`. An example can be found in [MapRemove](https://github.com/nhibernate/nhibernate-core/pull/2010/files#diff-00e2aadba38d74d8484b0e6fb6071d90R2178) test.

src/NHibernate/Collection/Trackers/MapQueueOperationTracker.cs
maca88: I didn't found any example where this would be used, so I removed it.

src/NHibernate/Collection/Trackers/SetQueueOperationTracker.cs
maca88: Fixed.

fredericDelaporte:
@bahusoid, you were having some concerns about some changes in this PR, which have been reverted. Do you have some other concerns?

bahusoid:
> Do you have some other concerns?

Nope. I can't make proper code review but in principle it's all good changes

fredericDelaporte:
We have forgotten to label this as "possible breaking change" and so it has not be mentioned in the possible breaking changes section of the release notes.
May be related to #2476.

maca88:
I've added the missing possible breaking change that is related to #2476.

fredericDelaporte:
I am not seeing where you have added that. I am not seeing a change on the release notes. (I usually include missing possible breaking changes in the release pr of the next patch, with a note in that release notes about completing the possible breaking changes of the minor.)

maca88:
I've added the missing one in the description of this PR and now I've made a PR for them. Feel free to reword the added notes, as I am not that good at writing :)

Issue #1983, opened 2019-01-17
Unable to correctly resolve the owner key with null value property


neseih:
Since NHibernate version 5.2 (commit: 8124cb5d75a8f76cde217a67b29b03f441057751) `GetKeyOfOwner` in CollectionType.cs throws an exception if the property value (in the database) is null.

Exception:
```
AssertionFailure: Unable to correctly resolve the owner key, property Property for collection SomeChildrenList.
Unresolved value '', key type 'NHibernate.Type.StringType', owner 'Example.SomeEntity'
```

Mapping (Example.SomeEntity):
``` c#
mapping.HasMany(m => m.SomeChildrenList)
.PropertyRef("Property")     // Type: String, can be null
.KeyColumns.Add("ReferencedProperty")
.NotFound.Ignore();
```

There is no ForeignKeyConstraint on Property and ReferencedProperty and both properties are not keys.

fredericDelaporte:
Which way is it? Does it throw if the referenced entity has the property null, or if if the referencing entity has its association null?

neseih:
It throws if the referencing entity has a property with null. In the example, it would be "Property" that is null and throws the exception when we access someEntity.SomeChildrenList.

fredericDelaporte:
Considering this case again, it seems to me that somewhat supporting `null` for a key (the property ref) was an unexpected side effect of #378, which was also causing bugs and has been reverted in #1872.

Associating entities with a `null` value is quite a debatable semantic. Indeed, on 5.1.x, it does not work either. Accessing the collection does not throw, but adding items to it is ignored. They get lost. I do not think this behavior was better than throwing right away.

So this could be fixed in two ways:
- Just document this end of partial `null` support as a possible breaking change.
- Support that special case again.

neseih:
Hi @fredericDelaporte,

thanks for investigating. For me, readonly access would be enough as I can't have a ForeignKey for different reasons. So if the property ref is null and empty (or null) collection is returned. That worked in previous releases and would be a non breaking change.

fredericDelaporte:
The more I consider this, the more I think we should:

> document this end of partial `null` support as a possible breaking change.

From a correctness standpoint, there is no point introducing back a bugged behavior, where one could accessed a non-functionnal collection without any error be raised, although anything done on the collection was ignored. That breaks the least surprise principle. When something do not work, it should throw, rather than silently doing nothing.

[...]

PR #2011, opened 2019-02-17
 Use a statically resolved dialect when building the session factory 

fredericDelaporte:
This change is more a trivial optimization: when building the session factory, the dialect has no reason to change, and so we can use the same dialect instance from the start of the process to its end.
#1951, merged in v5.3, was already doing most of the work, by supplying to the built session factory an `IMapping` resolving the dialect at instantiation rather than dynamically.

This PR cherry-pick #1951 into 5.2.x, and furthermore switch `Configuration.mapping` member to the mapping having a static dialect, for the duration of the build.

So if anything does use the `mapping.Dialect` in steps prior to factory instantiation, it will now use that static one instead of instantiating it.

Well, that does probably not actually optimize anything in most cases as far as I know.

But it allows to restore a hack used by NHibernate.Spatial, which relies on having its dialect instantiated early in the factory building process. Without that hack, Spatial users cannot upgrade to NHibernate 5.2.x.
See nhibernate/NHibernate.Spatial#104.

That is why I issue this PR, targeting the 5.2.x branch.

### DIFF
--- a/src/NHibernate/Cfg/Configuration.cs
+++ b/src/NHibernate/Cfg/Configuration.cs
@@ -239,6 +239,40 @@ namespace NHibernate.Cfg
 				NHibernate.Dialect.Dialect.GetDialect(configuration.Properties);
 		}
 
+		[Serializable]
+		private class StaticDialectMappingWrapper : IMapping
+		{
+			private readonly IMapping _mapping;
+
+			public StaticDialectMappingWrapper(IMapping mapping)
+			{
+				_mapping = mapping;
+				Dialect = mapping.Dialect;
+			}
+
+			public IType GetIdentifierType(string className)
+			{
+				return _mapping.GetIdentifierType(className);
+			}
+
+			public string GetIdentifierPropertyName(string className)
+			{
+				return _mapping.GetIdentifierPropertyName(className);
+			}
+
+			public IType GetReferencedPropertyType(string className, string propertyName)
+			{
+				return _mapping.GetReferencedPropertyType(className, propertyName);
+			}
+
+			public bool HasNonIdentifierPropertyNamedId(string className)
+			{
+				return _mapping.HasNonIdentifierPropertyNamedId(className);
+			}
+
+			public Dialect.Dialect Dialect { get; }
+		}
+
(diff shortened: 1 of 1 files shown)


hazzik:
I don't think this will help spatial, the error happens long before `BuildSessionFactory` call.

hazzik:
Ok, I've checked the stack trace. It should work.

fredericDelaporte:
I had checked with a locally generated NuGet package with those changes, and an upgraded Spatial.

Still, @peetw, can you also check this change solves nhibernate/NHibernate.Spatial#104, just to be sure? Releasing a patch takes some time, so it is better to double check.

hazzik:
Could you also include the release notes in the PR?

peetw:
Thank you both for looking into this - much appreciated! I was away for the weekend so apologies for not replying sooner. I should have some time today to test whether the PR fixes the issue.

fredericDelaporte:
> Could you also include the release notes in the PR?

I have not put them directly due to not having definitely decided what to do for #1983 (but most likely I will just add a possible breaking change notice), and also due to the suspected regression on `not-found` discussed in off-topic comments of #1995. If there is actually a regression there, it would then be better to fix it in 5.2.4 too.

peetw:
@fredericDelaporte Using the NuGet package from the [build artifacts](https://teamcity.jetbrains.com/viewLog.html?buildId=1982568&buildTypeId=NHibernate_NHibernateReleasePackage&tab=artifacts#%2FNHibernate-5.2.3%2Fnuget_gallery), I have confirmed that this PR fixes the issue reported in [nhibernate/NHibernate.Spatial#104](https://github.com/nhibernate/NHibernate.Spatial/issues/104) - thanks!

fredericDelaporte:
Release commit added, adding by the way a lacking possible breaking change.

hazzik:
The issue type has to be bug or task. Let’s call it bug.

fredericDelaporte:
I have adjusted the releasenotes file accordingly.

fredericDelaporte:
Release commit dropped.

PR #2014, opened 2019-02-20
Obsolete Environment.Properties

hazzik:
The bigger idea is to gradually move away from static configurations.

### DIFF
--- a/src/NHibernate/Cfg/Configuration.cs
+++ b/src/NHibernate/Cfg/Configuration.cs
@@ -171,7 +171,9 @@ namespace NHibernate.Cfg
 			propertyReferences = new List<Mappings.PropertyReference>();
 			FilterDefinitions = new Dictionary<string, FilterDefinition>();
 			interceptor = EmptyInterceptor.Instance;
+#pragma warning disable 618
 			properties = Environment.Properties;
+#pragma warning restore 618
 			auxiliaryDatabaseObjects = new List<IAuxiliaryDatabaseObject>();
 			SqlFunctions = new Dictionary<string, ISQLFunction>();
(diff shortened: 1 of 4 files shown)


src/NHibernate/Dialect/Dialect.cs
```diff
@@ -148,15 +148,18 @@ protected Dialect()
 		public static Dialect GetDialect()
```
fredericDelaporte: Shouldn't this one be obsoleted too ? How would it work without `Environment.Properties` ?

src/NHibernate/Cfg/Configuration.cs
```diff
@@ -171,7 +171,9 @@ protected void Reset()
 			propertyReferences = new List<Mappings.PropertyReference>();
 			FilterDefinitions = new Dictionary<string, FilterDefinition>();
 			interceptor = EmptyInterceptor.Instance;
+#pragma warning disable 618
 			properties = Environment.Properties;
```
fredericDelaporte: For fully preparing its removal, the code here should read the configuration file. (Or eventually cache statically in `Configuration` that file read result as a dictionary.)

hazzik: I decided to split this into several PRs. This deserves full re-implementation.

src/NHibernate/Dialect/Dialect.cs
hazzik: I decided to split this into several PRs.

hazzik:
The reason why this is just obsoleted is because I wanted to make this PR focused as much as possible.

PR #2015, opened 2019-02-21
Implement AliasToBeanCompiledResultTransformer

bahusoid:
AliasToBean transformation as compiled expression.

Original test `AliasToBeanResultTransformerFixture` was refactored to base class `AliasToBeanFixtureBase`. As in final commit it's unclear what's been changed, you can see changes [here](https://github.com/nhibernate/nhibernate-core/commit/ddd956e946b41c46c39b589e199449946d069478#diff-800a2cf725afba25e0808c18f73f6787).

I didn't replace/update original transformer as it seems possible to reuse it across different queries (it's undocumented and arguably useful feature but still...).  And new  transformer must be used only for the same query as it compiles delegate for aliases supplied by first transformation.

[Perf  test](https://github.com/nhibernate/nhibernate-core/compare/master...bahusoid:aliasToBeanCompiled?expand=1#diff-2b35825d5bd68d36dcdf78015340c5daR26) for 5 aliases and 50 000 transformations on my machine:
300 ms vs 15 ms for Compiled version


bahusoid:
Maybe this PR should be postponed and made as default `AliasToBeanResultTransformer' implementation in 6.0.

As I don't see any noticeable perf loss by expression compilation.

fredericDelaporte:
> Maybe this PR should be postponed and made as default `AliasToBeanResultTransformer' implementation in 6.0.
>
> As I don't see any noticeable perf loss by expression compilation.

But it forbids sharing it between different queries yielding the same result as you have explained, so I am unsure.

bahusoid:
> But it forbids sharing it between different queries yielding

The thing is that it saves very little from my tests. The main hit is took only by first instantiation where it seems .NET caches some reflection related information for result type. So I thought it might go as breaking 6.0 change  - in 6.0 you can reuse transformer for perf but only for the same query.

But if you think it's OK to provide two different options for solving the same problem - let's keep it as is.

fredericDelaporte:
Ok, better have only one. Such a feature does not deserve many options.

bahusoid:
Ok. Will do as separate PR and keep this one just in case (as it allows to easily compare perf for both versions)

bahusoid:
Replaced with #2021

Issue #1971, opened 2019-01-13
Add SourceLink to allow NuGet package debugging


cremor:
Have you thought about adding [SourceLink](https://github.com/dotnet/sourcelink) to NHiberante? That would make it easier to debug NHibernate in a project that references it.

As I see it, you'd only have to add the few things to the project file as explained in the readme of SourceLink. And you could then remove `RepositoryUrl` and `RepositoryType` from NHibernate.props since they are set automatically by SourceLink.

Related since the user will need to somehow get pdb files to allow SourceLink to work:
- [The official Open-source library guidance about symbol packages](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/nuget#symbol-packages)
- #1056

fredericDelaporte:
I have seen this as build warnings in NHibernate prior to releasing 5.2, but at that point it seemed a bit early to do the move.

Now it may be time. But compared to the legacy symbol package that NHibernate currently publishes, it does not embed sources. And due to the way it is generated, this new symbol package cannot be generated along the legacy one in a single package generation: either it generates the legacy format, or the new format, but not both.

This is a bit troublesome since the new format requires at least Visual Studio 2017 15.9 for being consumed, while the legacy one can be consumed by much older VS.

I do not see why we should cease providing symbols for older VS, since NHibernate can be used with quite old ones.

Maybe can we publish both symbol formats, but that will require a "double" package generation: one for the legacy format, one for the new one.

cremor:
> But compared to the legacy symbol package that NHibernate currently publishes, it does not embed sources.

Until I read your message in #1056 I didn't know that NHibernate publishes a (now legacy) symbol package. With that knowledge, the current situation is far better already.
But as I said in #1056 I think this should be documented. This describes the current situation perfectly: [Package Consuming Experience](https://github.com/NuGet/Home/wiki/NuGet-Package-Debugging-&-Symbols-Improvements#package-consuming-experience)

> And due to the way it is generated, this new symbol package cannot be generated along the legacy one in a single package generation: either it generates the legacy format, or the new format, but not both.
> This is a bit troublesome since the new format requires at least Visual Studio 2017 15.9 for being consumed, while the legacy one can be consumed by much older VS.

[...]

PR #2018, opened 2019-02-24
Update symbol package format

fredericDelaporte:
NHibernate goes on building symbol.nupkg since v5.0 although they are no more consumable with library built from `Sdk` projects (new .csproj format).

We should switch to building snupkg.

Related: #1971. (Sourcelink cannot be added yet because it implies adding a pre-release dependency.)


fredericDelaporte:
I have launched the release build and intend to check its artefacts before merging.

fredericDelaporte:
The artifacts are not correct. I guess the sdk has to be updated on the build server.

fredericDelaporte:
Closed because we can indeed already add SourceLink, and I do not think we should have two PRs for this.

PR #2019, opened 2019-02-24
Update symbol package format and add Sourcelink

### LINKED ISSUE #1971: Add SourceLink to allow NuGet package debugging

fredericDelaporte:
NHibernate goes on building symbol.nupkg since v5.0 although they are no more consumable with library built from `Sdk` projects (new .csproj format).

We should switch to building snupkg, and to ease furthermore debugging, add SourceLink as requested by #1971.

Fixes #1971.
Replaces #2018.

### DIFF
--- a/default.build
+++ b/default.build
@@ -319,21 +319,10 @@
 			<in>
 				<items>
 					<include name="${nuget.nupackages.dir}/*.nupkg"/>
-					<exclude name="${nuget.nupackages.dir}/*.symbols.nupkg"/>
 				</items>
 			</in>
 			<do>
 				<echo message="nuget push -source https://nuget.org/ ${path::get-file-name(filename)} ${environment::newline()}" file="${nuget.nupackages.pushbatfile}" append="true"/>
-			</do>
-		</foreach>
-		<foreach item="File" property="filename">
-			<in>
-				<items>
-					<include name="${nuget.nupackages.dir}/*.symbols.nupkg"/>
-				</items>
-			</in>
-			<do>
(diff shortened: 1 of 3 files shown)


default.build
```diff
@@ -319,23 +319,12 @@
 			<in>
 				<items>
 					<include name="${nuget.nupackages.dir}/*.nupkg"/>
-					<exclude name="${nuget.nupackages.dir}/*.symbols.nupkg"/>
 				</items>
 			</in>
 			<do>
 				<echo message="nuget push -source https://nuget.org/ ${path::get-file-name(filename)} ${environment::newline()}" file="${nuget.nupackages.pushbatfile}" append="true"/>
```
fredericDelaporte: This url is not correct for pushing snupkg. They are ignored on it, even when explicitly pushing them.
I am going to PR the required url for pushing snupkg along with the regular package.
Seen after publishing NHibernate.Caches 5.6, done nhibernate/NHibernate-Caches#74 for fixing it there.

PR #2020, opened 2019-02-25
Throw for DML on filter

bahusoid:
As otherwise DML is executed on the whole table.
Though it would be cool to support it.

### DIFF
--- a/src/NHibernate.Test/Async/LinqBulkManipulation/Fixture.cs
+++ b/src/NHibernate.Test/Async/LinqBulkManipulation/Fixture.cs
@@ -1158,6 +1158,18 @@ namespace NHibernate.Test.LinqBulkManipulation
 			}
 		}
 
+		[Test]
+		public async Task DeleteOnFilterThrowsAsync()
+		{
+			using (var s = OpenSession())
+			using (s.BeginTransaction())
+			{
+				var a = await (s.Query<SimpleEntityWithAssociation>().Take(1).SingleOrDefaultAsync());
+				var query = a.AssociatedEntities.AsQueryable();
+				Assert.That(() => query.Delete(), Throws.InstanceOf<NotSupportedException>());
+			}
+		}
+
 		#endregion
 	}
 }
--- a/src/NHibernate/Linq/DefaultQueryProvider.cs
+++ b/src/NHibernate/Linq/DefaultQueryProvider.cs
@@ -282,6 +282,9 @@ namespace NHibernate.Linq
 
 		public int ExecuteDml<T>(QueryMode queryMode, Expression expression)
 		{
+			if (Collection != null)
+				throw new NotSupportedException("DML operations are not supported for filters.");
+
 			var nhLinqExpression = new NhLinqDmlExpression<T>(queryMode, expression, Session.Factory);
 
 			var query = Session.CreateQuery(nhLinqExpression);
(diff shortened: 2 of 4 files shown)


src/NHibernate.Test/Async/LinqBulkManipulation/Fixture.cs
```diff
@@ -1158,6 +1158,18 @@ public async Task DeleteSyntaxWithCompositeIdAsync()
...
 
+		[Test]
+		public async Task DeleteOnFilterThrowsAsync()
+		{
+			using (var s = OpenSession())
+			using (s.BeginTransaction())
+			{
+				var a = await (s.Query<SimpleEntityWithAssociation>().Take(1).SingleOrDefaultAsync());
+				var query = a.AssociatedEntities.AsQueryable();
+				Assert.That(() => query.Delete(), Throws.InstanceOf<NotSupportedException>());
```
bahusoid: @maca88 I've just noticed that async versions of DML extension methods are not used at all in async tests. Is it a bug?

hazzik:
Should probably go all the way down to 5.0.x

hazzik:
Too soon, @bahusoid. @fredericDelaporte your thoughts?

bahusoid:
Anyway my branch manipulations seems to broke CI builds and for 5.0.x it's better to create new PR

fredericDelaporte:
Technically, I think we should anyway target first 5.2.x, then eventually backport (cherry-pick) to previous ones.
We have already backported some fixes on 5.0.x rather than developing them on it directly, so developing on it then merging up may cause conflicts/history glitches.
About 5.1.x, I think we have already done some backport too, but I have not checked. Anyway, we can start backporting on it instead of developing fixes on it directly, if it is not already started.

So this PR here would go in 5.2.4, and #2011 should drop its release commit for letting it go in its own PR.

src/NHibernate.Test/Async/LinqBulkManipulation/Fixture.cs
maca88: Yes it's a bug and I've created [an issue](https://github.com/maca88/AsyncGenerator/issues/121) for it.

maca88: Published a new release `0.13.2` that has the bug fixed.

src/NHibernate/Linq/DefaultQueryProvider.cs
```diff
@@ -282,6 +282,9 @@ public virtual void SetResultTransformerAndAdditionalCriteria(IQuery query, NhLi
 
 		public int ExecuteDml<T>(QueryMode queryMode, Expression expression)
 		{
+			if (Collection != null)
+				throw new NotSupportedException("DML operations are not supported for filters.");
```
hazzik: Should probably be `InvalidOperationException`. However, I'm not sure.

bahusoid: As potentially it could be implemented I think NotSupported/NotImplemented exceptions are more appropriate.

hazzik: NotSupported is fine

src/NHibernate.Test/Async/LinqBulkManipulation/Fixture.cs
fredericDelaporte: I do not think we should upgrade the async generator on 5.2x (and even less on previous branches). So we will have to take care of regenerating async files as part of the merge to master I think, in order to take into account the fix now available in master.

bahusoid: Same thoughts

PR #2022, opened 2019-02-26
Make CancellationToken optional for async Linq DML queries

hazzik:


hazzik:
I've included AsyncGenerator update to 0.13.2 here as it affects only one and the same file.

PR #2024, opened 2019-02-26
Refactor to simplify netfx retargeting

bahusoid:
I was checking some stuff under .NET 4.7 and have had some struggle to make it compile. So let's simplify it a bit..

### DIFF
--- a/src/NHibernate.Test.VisualBasic/NHibernate.Test.VisualBasic.vbproj
+++ b/src/NHibernate.Test.VisualBasic/NHibernate.Test.VisualBasic.vbproj
@@ -1,4 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <NhVbNet>True</NhVbNet>
+  </PropertyGroup>
   <Import Project="../../build-common/NHibernate.props" />
   <PropertyGroup>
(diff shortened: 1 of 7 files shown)


src/NHibernate.Test.VisualBasic/NHibernate.Test.VisualBasic.vbproj
```diff
@@ -13,8 +16,8 @@
     <OutputType>Exe</OutputType>
     <GenerateProgramFile>false</GenerateProgramFile>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetFramework)' == 'net461'">
-    <DefineConstants>NET461,$(DefineConstants)</DefineConstants>
+  <PropertyGroup Condition="$(TargetFramework.StartsWith('net4'))">
+    <DefineConstants>NETFX,$(DefineConstants)</DefineConstants>
```
fredericDelaporte: 👍

bahusoid:
> Why not adding a build variable (like `NhAppTargetFrameworks`, so maybe `NhNetFxTarget`) instead, and using it everywhere we currently hard-code `net461`

In theory NHibernate can build multiple netfx targets. So I think `NhAppTargetFrameworks` is sufficient and we don't need to introduce separate single netfx target property.

PR #2026, opened 2019-02-28
ComponentType will always return a non-null collection, even if all the entries are null.

czema:
If all the columns in a dynamic component are null, then rather than returning an IDictionary with keys and null values, it just returns a null reference.  Having even one column with a non-null value will cause the dictionary to be populated. 

This causes two problems: first is you have an unexpected null reference, and if you want to create a dictionary you won't know what keys to add to it.  Second, if you do add a reference to a dictionary with the correct keys, NHibernate will attempt to INSERT rather than UPDATE.  This will either cause a duplicate record or a primary key violation.

### DIFF
--- a/src/NHibernate/Type/ComponentType.cs
+++ b/src/NHibernate/Type/ComponentType.cs
@@ -490,10 +490,11 @@ namespace NHibernate.Type
 				begin += length;
 			}
 
-			if (ReturnedClass.IsValueType)
-				return values;
-			else
-				return notNull ? values : null;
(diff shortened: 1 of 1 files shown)


fredericDelaporte:
I guess you have replaced it with #2027. If that was just for doing some adjustments in the commit, there is no need for opening a new PR. Just add more commits, or, especially when no review has already been done, amend the current ones and force push them.

(Of course, now, go on with #2027.)

PR #2027, opened 2019-02-28
ComponentType will always return a non-null collection, even if all the entries are null.

czema:
If all the columns in a dynamic component are null, then rather than returning an IDictionary with keys and null values, it just returns a null reference. Having even one column with a non-null value will cause the dictionary to be populated.

This causes two problems: first is you have an unexpected null reference, and if you want to create a dictionary you won't know what keys to add to it. Second, if you do add a reference to a dictionary with the correct keys, NHibernate will attempt to INSERT rather than UPDATE. This will either cause a duplicate record or a primary key violation.

### DIFF
--- a/src/NHibernate/Type/ComponentType.cs
+++ b/src/NHibernate/Type/ComponentType.cs
@@ -468,7 +468,6 @@ namespace NHibernate.Type
 		public override object Hydrate(DbDataReader rs, string[] names, ISessionImplementor session, object owner)
 		{
 			int begin = 0;
-			bool notNull = false;
 			object[] values = new object[propertySpan];
 			for (int i = 0; i < propertySpan; i++)
 			{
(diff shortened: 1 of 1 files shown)


hazzik:
This is valid documented behavior.
 
From the documentation [Chapter 8. Component Mapping](https://nhibernate.info/doc/nhibernate-reference/components.html#components-dependentobjects):

> if all component columns are null, then the entire component is null

czema:
Does the behavior make sense?

fredericDelaporte:
> first is you have an unexpected null reference, and if you want to create a dictionary you won't know what keys to add to it.

Well, your application has anyway to know them at some point, for inserting the data, isn't it? So I am not convinced this should be considered an issue.

> Second, if you do add a reference to a dictionary with the correct keys, NHibernate will attempt to INSERT rather than UPDATE. This will either cause a duplicate record or a primary key violation.

This sounds as a bug. Can you PR a test case demonstrating it please?

By the way there are no test case added or adapted for testing your changes in this PR, and numerous existing ones are failing. It looks like you have not setup your local environment for testing. You should, because iterating over the CI for fixing things is generally quite inefficient.

hazzik:
I've created the issue #2112. Please continue any discussions there. In the current state the PR is not acceptable.

PR #2034, opened 2019-03-01
Throw for DML on filter

fredericDelaporte:
Backport to 5.1.x, fixes #2020


fredericDelaporte:
Release commit dropped, since another regression is to be fixed.

PR #2036, opened 2019-03-02
Reduce cast usage for aggregate functions

### LINKED ISSUE #2029: Incorrect SQL for cast inside an aggregate (MS SQL)
Version: 5.2.3 from NuGet

Code to reproduce:

```c#
session.Save(new TestClass { Int32Prop = Int32.MaxValue });
session.Save(new TestClass { Int32Prop = Int32.MaxValue });

session.Query<TestClass>()
    .GroupBy(i => 1)
    .Select(g => new {
        s = g.Sum(i => (long)i.Int32Prop)
    })
    .ToArray();
```

Generated SQL
```sql
select cast(sum(testclass0_.Int32Prop) as BIGINT) as col_0_0_ from TestClass testclass0_
```

fails with
```
Arithmetic overflow error converting expression to data type int.
```

Looks like `sum` and `cast` should be swapped.
This runs fine:

```sql
select sum(cast(testclass0_.Int32Prop as BIGINT)) as col_0_0_ from TestClass testclass0_
```

Test class and mapping:

```c#
public class TestClass {
    public virtual int? Int32Prop { get; set; }
}

public class TestClassMap : ClassMap<TestClass> {
    public TestClassMap() {
        Table("TestClass");
        Id(i => i.Id);
        Map(i => i.Int32Prop);
    }
}
```

maca88:
This PR uses the dialect specific SUM function to determine whether cast is needed or not and also takes nullable types into consideration, which fixes #2029.
Note:
The cast usage was not reduced for SQLite as it needs it when `sum` expression contains a parameter.

Edit: this PR has been extended to other aggregate functions.


src/NHibernate.Test/NHSpecificTest/GH2029/Fixture.cs
```diff
@@ -0,0 +1,181 @@
...
+		{
+			if (source == null)
+			{
+				return 0;
+			}
+			int n = 0, count = 0;
+			while ((n = source.IndexOf(substring, n, StringComparison.InvariantCulture)) != -1)
+			{
+				n += substring.Length;
+				++count;
```
fredericDelaporte: In c#, pre-increment is not better than post-increment, performance wise, while being generally considered less readable.
So better use the more usual `count++;` syntax.

src/NHibernate.Test/NHSpecificTest/GH2029/Fixture.cs
```diff
@@ -0,0 +1,181 @@
...
+			using (session.BeginTransaction())
+			using (var sqlLog = new SqlLogSpy())
+			{
+				var groups = session.Query<TestClass>()
+					   .GroupBy(i => 1)
+					   .Select(g => new
+					   {
+						   s = g.Sum(i => (long) i.NullableInt32Prop)
+					   })
+					   .ToArray();
```
fredericDelaporte: Has `ToArray` be used on purpose to avoid async generation?
Since we may someday support `ToArrayAsync`, I think we should use `ToList` here for now, in order to avoid changes here if `ToArrayAsync` is added.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -263,6 +265,16 @@ protected HqlTreeNode VisitNhMax(NhMaxExpression expression)
 
 		protected HqlTreeNode VisitNhSum(NhSumExpression expression)
 		{
+			var type = expression.Type.UnwrapIfNullable();
+			var nhType = TypeFactory.GetDefaultTypeFor(type);
+			if (nhType != null && _parameters.SessionFactory.SQLFunctionRegistry.FindSQLFunction("sum")
```
hazzik: I think some of this logic is not necessary.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -477,8 +489,10 @@ protected HqlTreeNode VisitUnaryExpression(UnaryExpression expression)
 				case ExpressionType.Convert:
 				case ExpressionType.ConvertChecked:
 				case ExpressionType.TypeAs:
-					if ((expression.Operand.Type.IsPrimitive || expression.Operand.Type == typeof(Decimal)) &&
-						(expression.Type.IsPrimitive || expression.Type == typeof(Decimal)))
+					var operandType = expression.Operand.Type.UnwrapIfNullable();
```
hazzik: Same here

maca88:
I've made a more sophisticated logic that can be used also for other aggregates like `COUNT` and `AVG`. Instead of checking the `ReturnedClass` property, the logic is now comparing `SqlTypes` which should be more correct as the cast is done on the database level. The logic was applied also for `AVG` but not for `COUNT` as I found out that the registered function "count_big" for `MsSql2000Dialect` is not used for hql queries and would not add a cast for a `long` type (`COUNT` in Sql Server [returns an `int`](https://docs.microsoft.com/en-us/sql/t-sql/functions/count-transact-sql?view=sql-server-2017#return-types)). It seems that [NH-865](https://nhibernate.jira.com/browse/NH-865) was never tested. I will try to address the count issue for Sql Server in a separate PR.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -584,5 +588,96 @@ protected HqlTreeNode VisitNewArrayExpression(NewArrayExpression expression)
...
+			}
+
+			var persister = _parameters.SessionFactory.GetEntityPersister(entityName);
+			var index = persister.EntityMetamodel.GetPropertyIndexOrNull(memberExpression.Member.Name);
+			return !index.HasValue
+				? TypeFactory.GetDefaultTypeFor(expression.Type) // Not mapped
+				: persister.EntityMetamodel.PropertyTypes[index.Value];
+		}
+
+		private string TryGetEntityName(MemberExpression memberExpression)
```
maca88: Copied from #1996.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -477,13 +485,12 @@ protected HqlTreeNode VisitUnaryExpression(UnaryExpression expression)
...
-					{
-						return _hqlTreeBuilder.Cast(VisitExpression(expression.Operand).AsExpression(), expression.Type);
-					}
-
-					return VisitExpression(expression.Operand);
+					return IsCastRequired(expression.Operand, expression.Type, out var existType)
+						? _hqlTreeBuilder.Cast(VisitExpression(expression.Operand).AsExpression(), expression.Type)
+						// Make a transparent cast when an IType exists, so that it can be used to retrieve the value from the data reader
+						: existType
+							? _hqlTreeBuilder.TransparentCast(VisitExpression(expression.Operand).AsExpression(), expression.Type)
```
maca88: I have to recheck this as `TransparentCast` does support only a few types.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
maca88: After doing #2079 I found out that this method does not cover all cases. I will update this method and try to add additional tests that cover the edge cases.

maca88:
Set to WIP as I need to revisit this PR, see the comments above.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
maca88: Added an additional check for calling `TransparentCast `, as there is an edge case when casting to a custom registered type.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
maca88: This method was improved and exposed for internal usage by #1996. I am thinking to leave this PR as WIP until #1996 get merged in order to replace this method with the new one.

maca88:
Rebased and updated `TryGetEntityName` (from #1996) and `SupportsType` (from #2079) methods.

src/NHibernate/Util/ExpressionsHelper.cs
```diff
@@ -225,10 +213,24 @@ public static class ExpressionsHelper
 							out currentComponentType);
 						break;
 					case IAbstractComponentType componentType:
+						// Concatenate the component property path in order to be able to use EntityMetamodel.GetPropertyType to retrieve the type.
+						// As GetPropertyType supports only components, do not concatenate when dealing with collection composite elements or elements.
+						if (!currentType.IsAnyType)
```
maca88: Thanks for cleaning the code. I've removed this condition as it is no longer needed because `AnyType` also implements `IAssociationType`, so it will be handled by the above `case`.

src/NHibernate/Util/ExpressionsHelper.cs
```diff
@@ -15,5 +27,572 @@ public static class ExpressionsHelper
...
+						currentComponentType = null;
+						break;
+				}
+			}
+			
+			// When traversed to the top of the expression, return the current tracking values
+			if (memberPaths.Count == 0)
+			{
+				memberPath = currentEntityPersister != null || currentComponentType != null ? member.Path : null;
+				mappedType = GetType(currentEntityPersister, currentType, member, sessionFactory, out _);
```
hazzik: Are you using `GetType` method in other PRs? If no, I think the last argument needs to be removed.

maca88: No, I've removed it.

maca88:
I've updated the `TryGetMappedType` method to support polymorphic queries, coalesce and conditional expressions, which will be needed for #2079 in order to detect parameter types.

src/NHibernate/Dialect/Function/CastFunction.cs
```diff
@@ -50,13 +50,8 @@ public SqlString Render(IList args, ISessionFactoryImplementor factory)
 				{
 					throw new QueryException("invalid NHibernate type for cast(), was:" + typeName);
 				}
+
 				sqlType = factory.Dialect.GetCastTypeName(sqlTypeCodes[0]);
-				if (sqlType == null)
-				{
-					//TODO: never reached, since GetTypeName() actually throws an exception!
```
maca88: It doesn't make sense using the type name when the dialect doesn't support the cast. If the cast would be supported, `GetCastTypeName` should return it.

maca88:
Rebased.

Issue #2031, opened 2019-03-01
Combining an implicit JOIN with a LEFT JOIN which requires a WITH clause triggers an exception.


bahusoid:
Original problematic query from hibenate [issue](https://hibernate.atlassian.net/browse/HHH-7321) looks like:
```SQL
SELECT s.studentId, q.questionId, sa.answer
FROM Student s, Question q
LEFT JOIN s.studentAnswerList AS sa WITH sa.question.questionId = q.questionId
```
Test simulating such query was added here:
https://github.com/nhibernate/nhibernate-core/blob/11bc49d412fda5630a6ac576d6bdd9d940220975/src/NHibernate.Test/Hql/EntityJoinHqlTest.cs#L124-L138

NHibernate currently generates joins in wrong order 
```SQL
FROM fromTable
LEFT JOIN leftJoin ON leftJoin.someId = thetaJoin.someId
,  thetaJoin
```

bahusoid:
I understood the root of the issue. Parser converts flat list of joins to parent-child tree:
1 fromTable -> 2 thetaJoin -> 3 leftJoin

is transformed to:
1 fromTable -> 1.1 leftJoin (as leftJoin is mapped association for fromTable)
1.2 thetaJoin

And final join list is constructed by flattening this tree: fromTable -> leftJoin -> thetaJoin

That means that this issue affects entity join too if it's referenced in association:
```SQL
FROM  fromTable
INNER JOIN entityJoin  ...
LEFT JOIN  association WITH entityJoin.SomeID =  association.SomeID
```

It seem some kind of topological sorting is required here. Or at least user defined order of joins shouldn't be affected .

hazzik:
Is it this one? https://hibernate.atlassian.net/browse/HHH-1969

bahusoid:
Looking for provided SQL - very likely. But this issue was closed as duplicate of https://hibernate.atlassian.net/browse/HHH-1480 without much investigations. So I can't be sure.

hazzik:
Based on a discussion in #2037 here should be "CROSS JOIN". Otherwise it would not be possible to re-order the joins in any way which will allow the query to execute.

bahusoid:
>  should be "CROSS JOIN". Otherwise it would not be possible to re-order the joins in any way which will allow the query to execute.

So to me now it looks like 2 issues:
1) Invalid query - as this query can't be written correctly in SQL with implicit joins.
[...]

PR #2037, opened 2019-03-03
Use proper inner join in Linq instead of cross-join and where clauses

### LINKED ISSUE #1128: NH-3210 - NHibernate Linq Provider does cross join or left outer join and not inner join (even if outer-join=false on many-to-one mapping)
**Taavi KÃµosaar** created an [issue](https://nhibernate.jira.com/browse/NH-3210) — <time datetime="2012-07-10T13:41:08.652Z" title="2012-07-10T13:41:08.652Z">10th July 2012, 13:41:08</time>:
> ```csharp
> var st1 = from ultt in _nHibernateSession.Query<UnitLeadTimeTemplate>()
>           where ultt.Unit.Id == unitId
>           select ultt.LeadTimeTemplate;
> st1.ToList();
> 
> var st2 = from ultt in _nHibernateSession.Query<UnitLeadTimeTemplate>()
>           from ltt in _nHibernateSession.Query<LeadTimeTemplate>()
>           where ultt.Unit.Id == unitId
>           select ltt;
> st2.ToList();
> 
> var st3 = from ultt in _nHibernateSession.Query<UnitLeadTimeTemplate>()
>           join ltt in _nHibernateSession.Query<LeadTimeTemplate>() 
>             on ultt.LeadTimeTemplate.Id equals ltt.Id
>           where ultt.Unit.Id == unitId
>           select ltt;
> st3.ToList();
> ```
>
> ```sql
> -- statement #1
> select leadtimete1_.lead_time_template_id as lead1_24_,
>        leadtimete1_.Name                  as Name24_,
>        leadtimete1_.Description           as Descript3_24_
> from   unit_lead_time_template unitleadti0_
>        left outer join lead_time_template leadtimete1_
>          on unitleadti0_.lead_time_template_id = leadtimete1_.lead_time_template_id
> where  unitleadti0_.unit_id = 8378 /** @p0 **/
> 
> -- statement #2
> select leadtimete1_.lead_time_template_id as lead1_24_,
>        leadtimete1_.Name                  as Name24_,
>        leadtimete1_.Description           as Descript3_24_
> from   unit_lead_time_template unitleadti0_,
>        lead_time_template leadtimete1_
> where  unitleadti0_.unit_id = 8378 /** @p0 **/
> 
> -- statement #3
> select leadtimete1_.lead_time_template_id as lead1_24_,
>        leadtimete1_.Name                  as Name24_,
>        leadtimete1_.Description           as Descript3_24_
> from   unit_lead_time_template unitleadti0_,
>        lead_time_template leadtimete1_
> where  leadtimete1_.lead_time_template_id = unitleadti0_.lead_time_template_id
>        and unitleadti0_.unit_id = 8378 /** @p0 **/
>``` 
>
> Even though in the mapping of UnitLeadTimeTemplate i have configured:
> Many-To-One with `not-null="true"` and `outer-join="false"`
> 
> There doesnt seem to be any consistent way to force inner-join to happen with linq query implementation in NH.
> 
>

hazzik:
Closes #1128

### DIFF
--- a/src/NHibernate/Hql/Ast/HqlTreeBuilder.cs
+++ b/src/NHibernate/Hql/Ast/HqlTreeBuilder.cs
@@ -363,6 +363,11 @@ namespace NHibernate.Hql.Ast
 			return new HqlJoin(_factory, expression, @alias);
 		}
 
+		public HqlInnerJoin InnerJoin(HqlExpression expression, HqlAlias @alias)
+		{
+			return new HqlInnerJoin(_factory, expression, @alias);
+		}
+
 		public HqlAny Any()
 		{
 			return new HqlAny(_factory);
--- a/src/NHibernate/Hql/Ast/HqlTreeNode.cs
+++ b/src/NHibernate/Hql/Ast/HqlTreeNode.cs
@@ -822,9 +822,18 @@ namespace NHibernate.Hql.Ast
 		}
 	}
 
+	public class HqlInnerJoin : HqlTreeNode
+	{
+		public HqlInnerJoin(IASTFactory factory, HqlExpression expression, HqlAlias @alias) 
+			: base(HqlSqlWalker.JOIN, "join", factory, new HqlInner(factory), expression, @alias)
+		{
+		}
+	}
+
 	public class HqlLeftJoin : HqlTreeNode
 	{
-		public HqlLeftJoin(IASTFactory factory, HqlExpression expression, HqlAlias @alias) : base(HqlSqlWalker.JOIN, "join", factory, new HqlLeft(factory), expression, @alias)
+		public HqlLeftJoin(IASTFactory factory, HqlExpression expression, HqlAlias @alias)
+			: base(HqlSqlWalker.JOIN, "join", factory, new HqlLeft(factory), expression, @alias)
 		{
 		}
 	}
@@ -872,6 +881,14 @@ namespace NHibernate.Hql.Ast
 	{
 		public HqlBitwiseAnd(IASTFactory factory, HqlExpression lhs, HqlExpression rhs)
 			: base(HqlSqlWalker.BAND, "band", factory, lhs, rhs)
+		{
+		}
+	}
+
+	public class HqlInner : HqlTreeNode
+	{
+		public HqlInner(IASTFactory factory)
+			: base(HqlSqlWalker.INNER, "inner", factory)
 		{
 		}
 	}
--- a/src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
+++ b/src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
@@ -346,7 +346,7 @@ namespace NHibernate.Linq.Visitors
 			HqlTreeNode hqlJoin;
 			if (joinClause.IsInner)
 			{
-				hqlJoin = _hqlTree.TreeBuilder.Join(expression, alias);
+				hqlJoin = _hqlTree.TreeBuilder.InnerJoin(expression, alias);
 			}
 			else
 			{
@@ -497,15 +497,17 @@ namespace NHibernate.Linq.Visitors
 
 		public override void VisitJoinClause(JoinClause joinClause, QueryModel queryModel, int index)
 		{
+			var querySourceName = VisitorParameters.QuerySourceNamer.GetName(joinClause);
 			var equalityVisitor = new EqualityHqlGenerator(VisitorParameters);
 			var whereClause = equalityVisitor.Visit(joinClause.InnerKeySelector, joinClause.OuterKeySelector);
 
-			_hqlTree.AddWhereClause(whereClause);
-
-			_hqlTree.AddFromClause(
-				_hqlTree.TreeBuilder.Range(
-					HqlGeneratorExpressionVisitor.Visit(joinClause.InnerSequence, VisitorParameters),
-					_hqlTree.TreeBuilder.Alias(joinClause.ItemName)));
+			var expression = HqlGeneratorExpressionVisitor.Visit(joinClause.InnerSequence, VisitorParameters).AsExpression();
+			var hqlJoin = _hqlTree.TreeBuilder.InnerJoin(
+				expression,
+				_hqlTree.TreeBuilder.Alias(querySourceName)
+			);
+			hqlJoin.AddChild(_hqlTree.TreeBuilder.With(whereClause.ToBooleanExpression()));
+			_hqlTree.AddFromClause(hqlJoin);
 		}
 
 		public override void VisitGroupJoinClause(GroupJoinClause groupJoinClause, QueryModel queryModel, int index)


src/NHibernate/Hql/Ast/HqlTreeBuilder.cs
```diff
@@ -363,6 +363,11 @@ public HqlJoin Join(HqlExpression expression, HqlAlias @alias)
 			return new HqlJoin(_factory, expression, @alias);
 		}
 
+		public HqlInnerJoin InnerJoin(HqlExpression expression, HqlAlias @alias)
+		{
+			return new HqlInnerJoin(_factory, expression, @alias);
```
hazzik: Unfortunately new join feature does not support implicit join types, so have to be explicit.

hazzik:
The description of #1128 is vague, so this change will make the issue obsolete. Technically the previous behaviour is correct despite being sub-optimal.

hazzik:
Depends on https://hibernate.atlassian.net/browse/HHH-11383 and #2031

bahusoid:
> Depends on HHH-11383 (hibernate/hibernate-orm@327d0a9 + hibernate/hibernate-orm@5ac46eb) and #2031

It seems the only change in hibernate is commented exception - why don't you do it right in this pr?

And regarding #2031 - I dig a little bit more in to it and I was hoping that I only need to adapt existing topological sorting to `JoinProcessor.ProcessJoins`
https://github.com/nhibernate/nhibernate-core/blob/95c43e4f08e3d16c3277a11bc7bfa7f6de30ac74/src/NHibernate/Hql/Ast/ANTLR/Util/JoinProcessor.cs#L99-L100

But proper sorting here didn't change a thing :(  It seems the final order of joins is ruled by how original AST tree is traversed by the following code (and apparently hibernate do it somehow correctly ):
https://github.com/nhibernate/nhibernate-core/blob/95c43e4f08e3d16c3277a11bc7bfa7f6de30ac74/src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs#L626

Also current tree parser requires you to write joins in correct orders - otherwise parser throws `InvalidPathException`. So it seems additional topological sorting is not required at all - instead we need some proper way of traversing existing query tree or change the way it's created.

So are you sure that this PR depends on this issue? As from my understanding existing tests shouldn't suffer from this issue.

hazzik:
> It seems the only change in hibernate is commented exception - why don't you do it right in this pr?

Didn’t have time yet. No

hazzik:
>So are you sure that this PR depends on this issue? As from my understanding existing tests shouldn't suffer from this issue.

Somehow.

Dlinq9 test:

```csharp
var actual =
	(from o in db.Orders
	 from p in db.Products
	 join d in db.OrderLines
		on new {o.OrderId, p.ProductId} equals new {d.Order.OrderId, d.Product.ProductId}
		into details
	 from d in details
	 select new {o.OrderId, p.ProductId, d.UnitPrice}).ToList();
```

Is generating following SQL:

```sql
select
    order0_.OrderId as col_0_0_,
    product1_.ProductId as col_1_0_,
    orderline2_.UnitPrice as col_2_0_
from
    Orders order0_,
    Products product1_
inner join OrderLines orderline2_
    on (orderline2_.OrderId=order0_.OrderId and orderline2_.ProductId=product1_.ProductId)
```

However I'm not sure how the SQL should look like.

bahusoid:
Shouldn't you change eliminate theta-style joins?

hazzik:
> Shouldn't you change eliminate theta-style joins?

Only second join in this query.

bahusoid:
Well from ordering standpoint everything is correct. Isn't it? So if it throws it's something else..

hazzik:
It throws:

"System.Data.SqlClient.SqlException : The multi-part identifier "order0_.OrderId" could not be bound."

bahusoid:
It seems SQL Server doesn't like such mix of theta-style joins and inner joins. If convert to bogus inner join it works:
```SQl
select
    order0_.OrderId as col_0_0_,
    product1_.ProductId as col_1_0_,
    orderline2_.UnitPrice as col_2_0_
from
    Orders order0_
inner join Products product1_ on 1=1
inner join OrderLines orderline2_
    on (orderline2_.OrderId=order0_.OrderId and orderline2_.ProductId=product1_.ProductId)
```
Maybe we should completely get rid of theta-style joins in Linq?

hazzik:
>Maybe we should completely get rid of theta-style joins in Linq?

What would be SQL for that query?

hazzik:
SQL server is not alone there, actually. Postgres throws a similar exception:
```
----> Npgsql.PostgresException : 42P01: invalid reference to FROM-clause entry for table "order0_"
```

bahusoid:
I've posted  example SQL in the same comment... Posting it again:

```SQL
select
    order0_.OrderId as col_0_0_,
    product1_.ProductId as col_1_0_,
    orderline2_.UnitPrice as col_2_0_
from
    Orders order0_
inner join Products product1_ on 1=1
inner join OrderLines orderline2_
    on (orderline2_.OrderId=order0_.OrderId and orderline2_.ProductId=product1_.ProductId)
```

hazzik:
Actually, these are called not "theta join" but "implicit join"

bahusoid:
> these are called not "theta join" but "implicit join"

Ok. Whatever suits you :) Though I'm pretty sure that I got "theta style join"  from some docs about hql

hazzik:
Ok, found correct issue: [HHH-1480](https://hibernate.atlassian.net/browse/HHH-1480).

bahusoid:
I'm not sure it's a good fix. It surely breaks mix of implicit and right joins in hql - see https://hibernate.atlassian.net/browse/HHH-5352

bahusoid:
There is still ordering issue with "cross join" fix for 'CrossJoinAndWithClause' :(
And if you need I can provide failing RIGHT join  test due to "cross join" fix which works on master.

IMHO this Cross join behavior should be configurable per query and be enabled by default only from linq as [HHH-5352](https://hibernate.atlassian.net/browse/HHH-5352) is a valid concern.

hazzik:
>I can provide failing RIGHT join test due to "cross join" fix which works on master.

Did it work in 5.2? ;-)

bahusoid:
>Did it work in 5.2? ;-)
Ok.. Maybe not :)

bahusoid:
> Did it work in 5.2? ;-)

Actually yes. You can mix cross join and right join on association in 5.2:
```SQL
select a, b, c
from a, b
right join b.c with ..
```

bahusoid:
Hm.. There is also interesting [commit ](https://github.com/hibernate/hibernate-orm/commit/753858ce735b2a4eb1a88e3dcd676a0f4e444c2b) which might fix my right join issue. Though not sure..

maca88:
As I am trying to add support for left joins for Linq (#864) and this issue is related, I want to give my thoughts about this. 

> Maybe we should completely get rid of theta-style joins in Linq?

I do agree. We shouldn't use comma syntax for doing a join, as it is an old way to join tables and it is not flexible, a simple explanation why can be found in the [mariadb documentation](https://mariadb.com/kb/en/comma-vs-join/). 

> What would be SQL for that query?

The query should look like this:

```SQl
select
    order0_.OrderId as col_0_0_,
    product1_.ProductId as col_1_0_,
    orderline2_.UnitPrice as col_2_0_
from
    Orders order0_
cross join Products product1_
inner join OrderLines orderline2_
    on (orderline2_.OrderId=order0_.OrderId and orderline2_.ProductId=product1_.ProductId)
```

Using a cross join allows us to reference the joined table in the later joins which is not possible when using the comma syntax.

> I'm not sure it's a good fix. It surely breaks mix of implicit and right joins in hql 

I do agree, I think that we shouldn't port the "fix" from Hibernate that replaces comma joins with cross joins as already mentioned, it is not the same when additional joins are added afterwards.

> There is still ordering issue with "cross join" fix for 'CrossJoinAndWithClause'

The query in that test is not valid as the left join references the table that is not available due to the JOIN precedence rules in SQL-99. IMO we shouldn't try to fix such queries, I think it is the user responsibility to know what are the limitations when using the comma syntax for joining tables.

>  There is also interesting commit which might fix my right join issue

I don't think that we need to fix anything when the user mixes comma joins with right/left/inner joins, for the above query we should convert it to sql as it is, without trying to convert the comma joins into cross joins.

bahusoid:
> The query in that test is not valid 

Yes we figured it out in the issue discussion. See pretty much the same thoughts here https://github.com/nhibernate/nhibernate-core/issues/2031#issuecomment-469297245

Also I did realize that entity join implementation in hql that I ported from hibernate have some flaws when mixed with associations. If I remember correctly entity joins are simply appended to the end disregarding it's actual usage in HQL. See example here: https://github.com/nhibernate/nhibernate-core/issues/2031#issuecomment-468643967

P.S. Sorry I don't really have time to actively participate now.

bahusoid:
Is this PR still valid? Isn't this functionality already implemented in https://github.com/nhibernate/nhibernate-core/pull/2327?

hazzik:
I don’t know now:)

maca88:
>  Isn't this functionality already implemented in #2327?

Yes it is, this PR can be closed.

PR #2039, opened 2019-03-05
Use generic CollectingNodeVisitor in hql parser

bahusoid:


src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -89,11 +89,12 @@ public IList List(ISessionImplementor session, QueryParameters queryParameters)
 			ErrorIfDML();
 			var query = ( QueryNode ) _sqlAst;
 			bool hasLimit = queryParameters.RowSelection != null && queryParameters.RowSelection.DefinesLimits;
-			bool needsDistincting = ( query.GetSelectClause().IsDistinct || hasLimit ) && ContainsCollectionFetches;
+			var containsCollectionFetches = ContainsCollectionFetches;
+			bool needsDistincting = ( query.GetSelectClause().IsDistinct || hasLimit ) && containsCollectionFetches;
```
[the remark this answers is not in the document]
bahusoid: yeah.. Better to undo

src/NHibernate/Hql/Ast/ANTLR/Tree/FromClause.cs
```diff
@@ -72,15 +72,27 @@ public FromClause ParentFromClause
...
+		
+		//6.0 TODO: Replace with Typed version below
 		public IList<IASTNode> GetExplicitFromElements()
 		{
-			return ASTUtil.CollectChildren(this, ExplicitFromPredicate);
+			return ASTUtil.CollectChildren<IASTNode>(this, ExplicitFromPredicate);
 		}
-
+		
+		internal IList<FromElement> GetExplicitFromElementsTyped()
```
hazzik: It feels like the return type shall be `IReadOnlyList<>`.

bahusoid: Why? That's not something we care about. Maybe user want to shuffle or slightly modify results in some way before further processing. I see no point in such restriction.

fredericDelaporte: I do not think we should restrict a method return type to be readonly when altering the returned value does not have any impact on the state of the object holding the method.

hazzik:
Ok

>

PR #2042, opened 2019-03-06
Fix missing cast when comparing a guid and string column in Sql Server

### LINKED ISSUE #2040: Incorrect SQL for cast inside a linq join (MS SQL)
Hello,

we are currently updating from an old NHibernate 4 version to the current NHibernate 5 (5.2.3) version and are facing some issues. 

One of our models contains a nullable string column which is joined on a column with a different type of an other table by using ToString() within the join, e.g:

```javascript
q =
from o in db.Orders
join s in db.Shippers
	on o.ShippedTo equals s.Reference.ToString()
where o.ShippedTo != null
select o.OrderId;
```
This worked in NHibernate 4 but fails in NHibernate 5. Here is a logged query example from our production code with 
[Join-Exception.txt](https://github.com/nhibernate/nhibernate-core/files/2931670/Join-Exception.txt)
NHibernate 4 and NHibernate 5.

```sql
-- NHibernate 4

select invruletem0_.ID as col_0_0_
from   InvRuleTemplate invruletem0_,
       InvRule invrule1_
where  (invruletem0_.Mandator & -1 /* @p0 */) != 0
       and (invrule1_.Mandator & -1 /* @p0 */) != 0
       and invrule1_.TargetID = cast(invruletem0_.ID as nvarchar(50))
       and (invrule1_.TargetID is not null)

-- NHibernate 5

select invruletem0_.ID as col_0_0_
from   InvRuleTemplate invruletem0_,
       InvRule invrule1_
where  (invruletem0_.Mandator & -1 /* @p0 */) != 0
       and (invrule1_.Mandator & -1 /* @p0 */) != 0
       and invrule1_.TargetID = invruletem0_.ID
       and (invrule1_.TargetID is not null)
```
It looks like that the .ToString() cast isn't translated into SQL anymore. This only happens when the string column is nullable. I guess the query translator recognizes that a cast isn't necessary for not nullable string columns.

This problem is reproducable with a unit test with the northwind domainmodel:
```javascript
[TestFixture]
public class LinqQuerySamples : LinqTestCase
{
	...

	[Category("JOIN")]
	[Test(Description = "This sample shows how to construct a join where one side is a nullable string and the other side a not nullable guid.")]
	public void DLinqJoin11()
	{
		var q =
			from o in db.Orders
			join s in db.Shippers
				on o.ShippedTo equals s.Reference.ToString()
			where o.ShippedTo != null
			select o.OrderId;

		ObjectDumper.Write(q);
	}
	
	....
}
```

I've also attached the unit test exception.

Thx in advance. Best regards!
[Join-Exception.txt](https://github.com/nhibernate/nhibernate-core/files/2931675/Join-Exception.txt)

maca88:
Fixes #2040

### DIFF
--- a/src/NHibernate.Test/Async/NHSpecificTest/NH3426/Fixture.cs
+++ b/src/NHibernate.Test/Async/NHSpecificTest/NH3426/Fixture.cs
@@ -84,5 +84,18 @@ namespace NHibernate.Test.NHSpecificTest.NH3426
 				Assert.That(list, Has.Count.EqualTo(1));
 			}
 		}
+
+		[Test]
+		public async Task CompareStringColumnWithGuidToStringAsync()
+		{
+			using (var session = OpenSession())
+			{
+				var list = await (session.Query<Entity>()
+				                  .Where(x => x.Id.ToString() == x.Name)
+				                  .ToListAsync());
+
+				Assert.That(list, Has.Count.EqualTo(0));
+			}
+		}
 	}
 }
--- a/src/NHibernate.Test/NHSpecificTest/NH3426/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3426/Fixture.cs
@@ -72,5 +72,18 @@ namespace NHibernate.Test.NHSpecificTest.NH3426
 				Assert.That(list, Has.Count.EqualTo(1));
 			}
 		}
+
+		[Test]
+		public void CompareStringColumnWithGuidToString()
+		{
+			using (var session = OpenSession())
+			{
+				var list = session.Query<Entity>()
+				                  .Where(x => x.Id.ToString() == x.Name)
+				                  .ToList();
+
+				Assert.That(list, Has.Count.EqualTo(0));
+			}
+		}
 	}
 }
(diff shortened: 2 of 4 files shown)


fredericDelaporte:
Since this is likely a 5.2 regression, it should target 5.2.x. But I have not checked yet it was working in 5.1.x.

maca88:
Changed the target to `5.2`. I've ran the tests on `5.1.x` branch and they all work.

fredericDelaporte:
Thanks Maca.

Since we have just released 5.2.4, and that this regression is not major, I intend to wait a bit before releasing a 5.2.5, in case some other things need a fix before the end of the month. So this PR will wait a bit, in case we just have to add the release commit here if there is nothing else to fix at the end of the month.

fredericDelaporte:
I think about it a bit too late, but it seems [SAP SQL Anywhere](https://teamcity.jetbrains.com/viewLog.html?buildId=2073687&buildTypeId=NHibernate_NHibernateSapSqlAnywhere) also needs a fix.

To spot it more easily while not failing PRs on 5.1x or earlier, I have enabled the build on master and 5.2.x. So we will still see troubles a bit too late, after PRs merges, but that will reduce a risk of releasing a regression on that database.

PR #2044, opened 2019-03-08
Fix transient detection for proxy

### LINKED ISSUE #2043: System.Reflection.TargetException when an interface is used as class mapping proxy definition
Hello,

This is an example model:
```javascript
{
}
````

```javascript
{
}
```

```javascript
```

```javascript
{
	return idGetter.Get(entity);
}
```
```javascript
entity.GetType().ToString()
"IEntityProxyProxy"
"System.Object"
{System.Type[3]}
idGetter
    Property: {System.Guid Id}
    PropertyName: "Id"
    @property: {System.Guid Id}
    propertyName: "Id"
```

> Fixes #1419

Thx, in adavance.

bahusoid:
Fixes #2043

### DIFF
--- a/src/NHibernate/Engine/ForeignKeys.cs
+++ b/src/NHibernate/Engine/ForeignKeys.cs
@@ -179,9 +179,17 @@ namespace NHibernate.Engine
 			}
 
 			// let the interceptor inspect the instance to decide
-			// let the persister inspect the instance to decide
-			return session.Interceptor.IsTransient(entity) ??
-			       session.GetEntityPersister(entityName, entity).IsTransient(entity, session);
+
+			if (session.Interceptor.IsTransient(entity) == true)
+				return true;
+
+			if (entity is INHibernateProxy proxy && proxy.HibernateLazyInitializer.IsUninitialized)
+			{
+				return false;
+			}
+
+			// let the persister inspect the instance to decide	
+			return session.GetEntityPersister(entityName, entity).IsTransient(entity, session);
 		}
 
 		/// <summary> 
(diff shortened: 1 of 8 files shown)


src/NHibernate/Engine/ForeignKeys.cs
```diff
@@ -179,9 +179,17 @@ public static bool IsNotTransientSlow(string entityName, object entity, ISession
...
 
 			// let the interceptor inspect the instance to decide
-			// let the persister inspect the instance to decide
-			return session.Interceptor.IsTransient(entity) ??
-			       session.GetEntityPersister(entityName, entity).IsTransient(entity, session);
+
+			if (session.Interceptor.IsTransient(entity) == true)
+				return true;
+
+			if (entity is INHibernateProxy proxy && proxy.HibernateLazyInitializer.IsUninitialized)
```
bahusoid: Can proxy be at all transient? Can't this check be simplified to `enitity is INHibernateProxy`?

bahusoid:
Oups.. I've accidentally merged it to upstream :(

bahusoid:
Is it OK to hard reset it? Or what should I do?

fredericDelaporte:
I have done it. The branch is protected and cannot be forced pushed to by default. I have also raised the protection.

bahusoid:
@fredericDelaporte Thanks.. I blame Ride for this (just decided to give it a shot) :) And somehow made this push without any confirmations dialog...

fredericDelaporte:
Rider? Never got this issue with it. But my default upstream is never nhibernate repository but mine instead. The additional constraint (review required) I have added for merging PR will not apply if merging locally then pushing with upstream set to nhibernate.

bahusoid:
> Rider? 

Yeah.. This treacherous setting bite me: Show PUSH dialog only when committing to protected branches 
IMHO It shouldn't be ON by default...

PR #2045, opened 2019-03-08
 Fix transient detection for proxy

### LINKED ISSUE #2043: System.Reflection.TargetException when an interface is used as class mapping proxy definition

bahusoid:
Fixes #2043


fredericDelaporte:
The change suspected in #2043 dates back to 5.0. Is this a regression of 5.0?

bahusoid:
>Is this a regression of 5.0?

Yes.  By adding call to method `ForeignKeys.IsTransientFast -> IEntityPersister.IsTransient -> IEntityTuplizer.GetIdentifier ` which doesn't work with proxies. 

So in fix I've added support for proxies (1st place as shortcut to `ForeignKeys.IsTransientFast`, and 2nd place to support proxies in `AbstractEntityTuplizer.GetIdentifier`)

src/NHibernate/Engine/ForeignKeys.cs
```diff
@@ -179,9 +179,16 @@ public static bool IsNotTransientSlow(string entityName, object entity, ISession
 			}
 
 			// let the interceptor inspect the instance to decide
-			// let the persister inspect the instance to decide
-			return session.Interceptor.IsTransient(entity) ??
-			       session.GetEntityPersister(entityName, entity).IsTransient(entity, session);
+			if (session.Interceptor.IsTransient(entity) == true)
+				return true;
+
+			if (entity is INHibernateProxy proxy && proxy.HibernateLazyInitializer.IsUninitialized)
```
bahusoid: Can proxy be at all transient? Can't this check be simplified to `enitity is INHibernateProxy`?

maca88: If I remember correctly it can be when the proxy is used in a different session.

bahusoid: I've also realized that user can force entity to be treated as transient by assigning Id to unsaved value. Not sure if it works but let's handle it here

src/NHibernate/Engine/ForeignKeys.cs
```diff
@@ -179,9 +179,16 @@ public static bool IsNotTransientSlow(string entityName, object entity, ISession
 			}
 
 			// let the interceptor inspect the instance to decide
+			if (session.Interceptor.IsTransient(entity) is bool transient)
```
fredericDelaporte: Surely it cannot have a significant performance impact in the middle of this processing, but still, using a cast here feels quite hackish to me. Better use a local var and check `HasValue`.
(I have checked the performance anyway : https://dotnetfiddle.net/6wcSBP.)

```suggestion
			var transient = session.Interceptor.IsTransient(entity);
			if (transient.HasValue)
```

src/NHibernate/Engine/ForeignKeys.cs
```diff
@@ -179,9 +179,16 @@ public static bool IsNotTransientSlow(string entityName, object entity, ISession
...
+				return transient;
```
fredericDelaporte:
```suggestion
				return transient.Value;
```

src/NHibernate/Engine/ForeignKeys.cs
bahusoid: Done. Hm.. Thanks for benchmarks. I was hoping that compiler would  be smart enough to make it equivalent for var

fredericDelaporte:
This one will have to be back-ported down to 5.0.

src/NHibernate/Tuple/Entity/AbstractEntityTuplizer.cs
```diff
@@ -114,6 +114,11 @@ public object Instantiate(object id)
 
 		public object GetIdentifier(object entity)
 		{
+			if (entity is INHibernateProxy proxy)
```
hazzik: In other places the caller checks for proxy before calling this method.

fredericDelaporte: Indeed, in `NHibernate.Type.EntityType`, the checks done there are no more needed with this PR changes.

There are other cases where the proxy is unproxied before hand, but that looks to me as a requirement of the implemented features there than due to `GetIdentifier` not supporting proxies.

bahusoid: > In other places the caller checks for proxy before calling this method.

And your point? You prefer to keep it that way? To me it looks better to support it here.
Or at least move this check to `AbstractEntityPersister.GetIdentifier`.

If it's about clean up other places - I think we can't remove checks in other places in this PR to avoid some custom implementations relying on it. I can add 6.0 TODO here for clean up.

hazzik: I would prefer to keep this check outside of this method as in other places.

bahusoid: So how about moving it to `AbstractEntityPersister.GetIdentifier`?  As  I don't see why `persister.GetIdentifier` shouldn't handle proxies.

Also I've just noticed the following code:
https://github.com/nhibernate/nhibernate-core/blob/8d13f2492ecd30a1fc451eee38208dbf95a2a614/src/NHibernate/Engine/ForeignKeys.cs#L156-L159
I think we need to use the same logic in both places. But not sure which.

src/NHibernate/Engine/ForeignKeys.cs
bahusoid: I've just noticed the following code:
https://github.com/nhibernate/nhibernate-core/blob/8d13f2492ecd30a1fc451eee38208dbf95a2a614/src/NHibernate/Engine/ForeignKeys.cs#L156-L159

I think we need to use the same logic in both places. But not sure which.

bahusoid: Also to decide if this check should be before `interceptor.IsTransient` as in `IsNotTransientSlow`. And should we also add shortcut `session.PersistenceContext.IsEntryFor(entity)`?

`IsNotTransientSlow` should really be implemented as `!IsTransientSlow` otherwise it's doesn't feel right to me that behavior is different.

bahusoid:
So for now  I've simply moved existing logic from `IsNotTransientSlow` to `IsTransientFast`. This change makes unnecessary changes in `GetIdentifier` so I've removed it from this PR as Alex don't like it in current state.

src/NHibernate/Tuple/Entity/AbstractEntityTuplizer.cs
fredericDelaporte: I am seeing more benefits in having the checks inside that method rather than outside. I do not really see why this method should not do the check itself. It does look very lightweight to me, so having the check done for cases where it can never be a proxy should not have a significant impact. What am I missing?

bahusoid: @fredericDelaporte I agree with you. But if you are OK with my latest changes I think this part is out of scope of this PR as now IsTransient logic never calls `GetIdentifier` for proxies

src/NHibernate/Async/Engine/ForeignKeys.cs
```diff
@@ -171,38 +167,25 @@ public static async Task<bool> IsNotTransientSlowAsync(string entityName, object
...
-				if (isTransient.HasValue)
-					return Task.FromResult<bool?>(isTransient);
+			if (entity.IsProxy())
+				return false;
 
-				if (entity is INHibernateProxy proxy && proxy.HibernateLazyInitializer.IsUninitialized)
-				{
-					return Task.FromResult<bool?>(false);
-				}
+			if (session.PersistenceContext.IsEntryFor(entity))
```
bahusoid: Actually as shows NH607 this check is wrong. PersistenceContext can contain entries for transient objects.
And hack was added to support wrong implementation of `IsNotTransientSlow`:
https://github.com/nhibernate/nhibernate-core/blob/ac39173567b31bcfad475ca32687c9faf0d37f87/src/NHibernate/Engine/ForeignKeys.cs#L249-L255

So in this PR I'm going to move this check back to `IsNotTransientSlow` and in new PR for 5.3 will try to remove it completly.

src/NHibernate/Engine/ForeignKeys.cs
```diff
@@ -171,6 +171,9 @@ public static bool IsNotTransientSlow(string entityName, object entity, ISession
 		/// </remarks>
 		public static bool? IsTransientFast(string entityName, object entity, ISessionImplementor session)
 		{
+			if (entity.IsProxy())
+				return false;
```
fredericDelaporte: So, despite some [hesitations](https://github.com/nhibernate/nhibernate-core/pull/2045#discussion_r263922575) on that subject, now proxies are treated as never transient.

What happens if we load a proxy of an entity with an assigned identifier, initialize it, delete it in session, flush, and save it again? In such case we would have a transient proxy, and it should be possible to save it, I guess. If that case does work with currently released NHibernate, then the logic from "not transient" should not be applied to "transient" too. (And we should consider removing it from the "not" case.)

bahusoid: This scenario still works (both direct and cascace saving) - as entity is unproxied along the way to `IsTransientFast`. I will commit tests for such cases later.

But from correctness standpoint you are right - my initial implementation is more correct. It correctly detects transient objects from any context. So I think I simply need to drop all my commits to the state you've approved initially. Do you agree? That means that `GetIdentifier` fix is also required here.

fredericDelaporte: Your initial implementation has my preference.

@hazzik,

> I would prefer to keep this check outside of this method as in other places.

Is-it because this is a patch of  5.2.x? So you consider that this kind of (even small) "semantic" change should be avoided?
(I do not actually see it as a semantic change. I see it more like fixing a bug in `GetIdentifier`, since its documentation does not state it does not support proxies.)

hazzik: Sorry, I did not have time to clarify my position on this subject. While this check is ok here, I think it should be performed outside of this method. First of all, `AbstractEntityPersister`/`AbstractEntityTuplizer` does not deal with any kind of instantiaded proxies currently (except a delegated method to instantiate one). Secondly, `ISessionImplementor.GetEntityPersister(...)` is a very heavy method and really should be left as a last resort. So, check for the proxy should happen outside of the persister to avoid its lookup.

bahusoid: >ISessionImplementor.GetEntityPersister(...) is a very heavy method and really should be left as a last resort. 

Why single dictionary lookup is considered heavy? And in our case it's `persister.IsTransient` method that calls `GetIdentifier`. So we really save nothing.
It's either put this check  in `persister.IsTransient` vs `GetIdentifier` (persister or tupilizer). To me the deeper the better - to cover all possible future cases (so in tupilizer).

> `AbstractEntityPersister/AbstractEntityTuplizer` does not deal with any kind of instantiaded proxies currently

I take it you meant - non-instantiated. As I already tell EntityPersister **must** be able to work with such proxy. For now it's only for `IsTransient` method.

hazzik: > Why single dictionary lookup is considered heavy?

You are confusing it with another method that is of `ISessionFactory`

bahusoid: Ok. It could be heavy for subclasses. But that wasn't really the point. My main point for proper `IsTransient` detection we need to support proxies. And it's better to do right in `GetIdentifier` method

bahusoid: I've pushed changes to the point I think is the right fix. Feel free to adjust it your way

src/NHibernate/Engine/ForeignKeys.cs
```diff
@@ -178,10 +178,29 @@ public static bool IsNotTransientSlow(string entityName, object entity, ISession
...
+			if (interceptorResult.HasValue)
+				return interceptorResult;
+
 			// let the persister inspect the instance to decide
-			return session.Interceptor.IsTransient(entity) ??
-			       session.GetEntityPersister(entityName, entity).IsTransient(entity, session);
+			if (proxy != null)
+			{
+				// The persister only deals with unproxied entities.
+				entity = proxy.HibernateLazyInitializer.GetImplementation();
```
bahusoid: I don't think it's a correct change. That leads to proxy initialization isn't it? Transient fast check should not do any such thing - it should return null if fast detection (without db hit) can't be made

bahusoid: OK I see.. It's about only initialized case. Well still it looks kinda hacky...

fredericDelaporte: This is the way to get the underlying entity, as used [here](https://github.com/nhibernate/nhibernate-core/blob/21846c017658e89af6c2d720b78624abf9acc064/src/NHibernate/Event/Default/DefaultLoadEventListener.cs#L186).
If the persister should not deal with proxies, I do not see any other way to achieve this.

fredericDelaporte: And I have just checked, fixing `GetIdentifier` is not enough. `IsTransient` may also call `GetVersion` which would also call a getter for the concrete class. And there, the only way to get it working would be to unproxy the object anyway. (Or the check on version would have to be given up with proxies, but that would change the behavior between proxy case and not-proxy case, which would not be a good move I think.)

bahusoid: That actually a good catch.

PR #2046, opened 2019-03-09
Nullable Guid ToString does not produce correct code on some dialects

hazzik:


hazzik:
I'll push a fix when tests will fail. Also, I'm not sure what branch shall I target to. Strictly speaking this is not a regression.

hazzik:
This is a follow-up to #2040

fredericDelaporte:
I think about it a bit too late, but it seems [SAP SQL Anywhere](https://teamcity.jetbrains.com/viewLog.html?buildId=2073687&buildTypeId=NHibernate_NHibernateSapSqlAnywhere) also needs a fix.

To spot it more easily while not failing PRs on 5.1x or earlier, I have enabled the build on master and 5.2.x. So we will still see troubles a bit too late, after PRs merges, but that will reduce a risk of releasing a regression on that database.

PR #2049, opened 2019-03-11
Fix property-ref ignoring not-found="exception" mapping

bahusoid:


src/NHibernate/Type/EntityType.cs
```diff
@@ -525,6 +529,10 @@ public object LoadByUniqueKey(string entityName, string uniqueKeyPropertyName, o
 				if (result == null)
 				{
 					result = persister.LoadByUniqueKey(uniqueKeyPropertyName, key, session);
+					if (result == null && !IsNullable)
+					{
+						factory.EntityNotFoundDelegate.HandleEntityNotFound(entityName, euk);
```
bahusoid: Not sure maybe it make sense to extend `HandleEntityNotFound` method along with `ObjectNotFoundException` to support `propertyName` property.
Now I simply supply `EntityUniqueKey` which I hope gives clear string representation for the issue.

hazzik: Usually it's `key`,

bahusoid: ? Are you suggesting to supply `key` instead of `euk` here? But we shouldn't loose `uniqueKeyPropertyName`   here - it's essential part of the `key`. Otherwise error message is wrong and misleading.

bahusoid: So to clarify: `LoadByUniqueKey` loads entity not by ID  but but by particular propertyName and propertyValue.
So exception also needs to include this information about both propertyName and propertyValue.
Supplying only key (propertyValue) is misleading as exception message suggests that there is no such ID with given value but query was executed not by ID but by propertyName.

As currently `HandleEntityNotFound` supports only id value I supply `EntityUniqueKey` which in string representation includes both propertyName and propertyValue:
>No row with the given identifier exists[NHibernate.Test.NHSpecificTest.NH1001.Department#EntityUniqueKey[NHibernate.Test.NHSpecificTest.NH1001.Department.PropRefId#99999]]

For more proper implementation I think `HandleEntityNotFound` and `ObjectNotFoundException`  can be extended to support propertyName. So we can throw exception that can look like:
No row with given property [properName] exists [propertyValue]

But not sure if it's worth it.

fredericDelaporte: Normally the user is supposed to know when the mapping is referencing through an unique key rather than through the primary key. NHibernate use the term `key` when it can be a property-ref (if it is not, then key contains the id), and `id` when it is always the primary key.

Most code is done with this assumption, so that is why it does not bother indicating the column name. That is left up to the user to check the mappings if he does not remember.

So, as long as the message states that the `key` is not found rather than stating that the `id` is not found, that is good enough. Sure it is more user friendly to tell explicitly the column, but this is going a bit beyond the PR primary subject in my opinion. And it may introduce a little inconsistency, if as I believe, all other current cases do not state the column.

bahusoid: >So, as long as the message states that the key is not found rather than stating that the id is not found, that is good enough

Error message that's thrown with this  PR in current state:
`No row with the given identifier exists [NHibernate.Test.NHSpecificTest.NH1001.Department#EntityUniqueKey[NHibernate.Test.NHSpecificTest.NH1001.Department.PropRefId#99999]]`

As you can see it's still explicitly states **identifier** rather than `key`. `EntityUniqueKey`  gives the following details: `EntityUniqueKey[NHibernate.Test.NHSpecificTest.NH1001.Department.PropRefId#99999]`

That's the only message that `ObjectNotFoundException` can provide. And I see no `key` related not found exceptions.
So I don't quite understand your position. Are you satisfied with current exception message?

>all other current cases do not state the column.

What cases are you talking about? Are you sure we have any other cases where exception is thrown for non ID column?

fredericDelaporte: I have not found any better than some dead code not even mentioning the key value.

Anyway, my point is, mentioning that the key is not found can be enough. I do not think the message is wrong and misleading without the column name, provided the `key` term is used, not `identifier`.

But now that I look at `HandleEntityNotFound`, it is just not the right method to call here. It is meant for the case where we have an id, as its parameter name and its interface documentation indicate. We should not use it with a key or anything else than an id. It would be unexpected by implementors and would be a contract violation. (This method is primarily meant for handling proxy loading failures.)

People can implement it in an interceptor, and they expect an id.

I think the code here has to throw directly an exception here. And neither an `ObjectNotFoundException` nor an `UnresolvableObjectException` since both are restricted to the id case too. (`ObjectNotFoundException` inherits `UnresolvableObjectException`, both have `summary`  stating they are specific to the lookup by id failure.) A new specific exception may be required here.

bahusoid: I've made this new exception to be interceptable in 6.0. But not sure any more.. So are you saying it's better just to throw directly?

fredericDelaporte: Yes, it seems to me NHibernate does not have currently a suitable mechanism for allowing intercepting a loading failure through a property-ref. So for the property-ref case, I think we need to throw directly, maybe with a 6.0 todo "have the property-ref case handled by HandleEntityNotFound too". (That could be adding an overload, or switching `id` to something like `idOrEuk`. Adding an overload implies more code, but a `idOrEuk` looks as a code smell to me.)

bahusoid: >maybe with a 6.0 todo "have the property-ref case handled by HandleEntityNotFound too"

Well if you think that interception might be useful than PR in current state is ready for review. As I've already added a method  to add in 6.0 [here](https://github.com/nhibernate/nhibernate-core/pull/2049/commits/8d9935eca288270cd883081298b187fc99f00758#diff-97a762863f6e5a44820f0c8f707a338eR13)

src/NHibernate/ObjectNotFoundByUniqueKeyException.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using NHibernate.Impl;
+
+namespace NHibernate
+{
+	/// <summary>
+	/// Thrown when entity can't be found by given unique key
+	/// </summary>
+	[Serializable]
+	public class ObjectNotFoundByUniqueKeyException : HibernateException
```
fredericDelaporte: It is not enough for an exception class to have the `[Serializable]` attribute. The `Exception` base class implements `ISerializable`, so, when new members are added in a derived class, `GetObjectData` must be overridden for serializing them. And in any case, the deserialization constructor must be re-declared.
See `NHibernate.ADOException` for an example.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -86,6 +86,11 @@ public void HandleEntityNotFound(string entityName, object id)
 				throw new ObjectNotFoundException(id, entityName);
 			}
 
+			public void HandleEntityNotFound(string entityName,string propertyName, object key)
```
fredericDelaporte: Lack of a space after a comma. `entityName,string`

src/NHibernate/ObjectNotFoundByUniqueKeyException.cs
bahusoid: Fixed

src/NHibernate/Impl/SessionFactoryImpl.cs
bahusoid: Fixed

PR #2055, opened 2019-03-13
WIP Make IEntityPersister.IsTransient not async

bahusoid:
By moving cache check from `IEntityPersister.IsTransient` to `ForeignKeys.IsTransientSlow`
Definitely for 6.0

I'm planning to split it in 2 PRs (extract DbTimestampType.UsePreparedStatement related changes to another PR). And it seems better to merge async regen changes as separate commit


src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -4230,14 +4230,7 @@ public virtual void AfterReassociate(object entity, ISessionImplementor session)
 				}
 			}
 
-			// check to see if it is in the second-level cache
```
hazzik: `session` argument is not used anymore. So, as we're talking about 6.0 it shall be removed.

PR #2056, opened 2019-03-13
Optimize GetOrphans and remove wrong checks from IsNotTransientSlow

bahusoid:
* Removed possibly slow transient check for current items (it's enough to have defined ID for orphan check). Made GetOrphans  method not async.

* Made use directly `persister.GetIdentifier` instead of additional session based checks via `GetEntityIdentifierIfNotUnsaved`. Removed hack that was introduced with NH607 (when `GetEntityIdentifierIfNotUnsaved` returns null as ID for not-null transient entity instead of throwing exception)

* Removed wrong checks from `IsNotTransientSlow` that it seems were added specifically to make old `GetOrphans` implementation work.


src/NHibernate/Engine/ForeignKeys.cs
```diff
@@ -246,16 +245,6 @@ public static object GetEntityIdentifierIfNotUnsaved(string entityName, object e
 
 					if (IsTransientFast(entityName, entity, session).GetValueOrDefault())
 					{
-						/***********************************************/
-						// TODO NH verify the behavior of NH607 test
-						// these lines are only to pass test NH607 during PersistenceContext porting
-						// i'm not secure that NH607 is a test for a right behavior
-						EntityEntry entry = session.PersistenceContext.GetEntry(entity);
-						if (entry != null)
-							return entry.Id;
```
bahusoid: PersistenceContext can contain transient objects in state Saving. This hack was added (that returns null ID for transient object) to make `GetOrphans` work. Removing it as `GetOrphans` is not relying on `GetEntityIdentifierIfNotUnsaved` anymore.

src/NHibernate/Collection/AbstractPersistentCollection.cs
```diff
@@ -716,29 +717,37 @@ protected virtual ICollection GetOrphans(ICollection oldElements, ICollection cu
...
 
-			// collect EntityIdentifier(s) of the *current* elements - add them into a HashSet for fast access
-			var currentIds = new HashSet<TypedValue>();
+			var currentObjects = new HashSet<object>(idType.GetComparer());
 			foreach (object current in currentElements)
 			{
-				if (current != null && ForeignKeys.IsNotTransientSlow(entityName, current, session))
+				if (current != null)
 				{
-					object currentId = ForeignKeys.GetEntityIdentifierIfNotUnsaved(entityName, current, session);
```
bahusoid: This call checks session entry dictionary and throw exception for transient object. This exception doesn't make any sense  for orphans detection - we should simply ignore transient objects. So let's call `persister.GetIdentifier` directly without any additional checks

src/NHibernate/Collection/AbstractPersistentCollection.cs
```diff
@@ -716,29 +717,37 @@ protected virtual ICollection GetOrphans(ICollection oldElements, ICollection cu
...
-			List<object> res = new List<object>();
+			var persister = session.Factory.GetEntityPersister(entityName);
+			IType idType = persister.IdentifierType;
 
-			// collect EntityIdentifier(s) of the *current* elements - add them into a HashSet for fast access
-			var currentIds = new HashSet<TypedValue>();
+			var currentObjects = new HashSet<object>(idType.GetComparer());
 			foreach (object current in currentElements)
 			{
-				if (current != null && ForeignKeys.IsNotTransientSlow(entityName, current, session))
```
bahusoid: Why to check all existing objects if they are transient . Let's check only objects detected as orphans

hazzik:
`GetOrphans` is still async in `PersistentArrayHolder`.

hazzik:
@bahusoid, could you please take a look at this:

> GetOrphans is still async in PersistentArrayHolder.

if you have time?

bahusoid:
@maca88 seems like a bug in AsyncGenerator. I think AsyncGenerator should add `Obsolete` attribute if it overrides abstract method with Obsolete attribute:
https://github.com/nhibernate/nhibernate-core/blob/1006665b0e066e94e05efc9112d1aadd2dafc00b/src/NHibernate/Async/Collection/AbstractPersistentCollection.cs#L131-L132
But it's not happening leading to warning (that treated as error):
https://github.com/nhibernate/nhibernate-core/blob/1006665b0e066e94e05efc9112d1aadd2dafc00b/src/NHibernate/Async/Collection/PersistentArrayHolder.cs#L30

maca88:
It is a bug, I've opened [an issue](https://github.com/maca88/AsyncGenerator/issues/127) for it.

src/NHibernate/Async/Engine/CollectionEntry.cs
```diff
@@ -70,18 +70,5 @@ public async Task PreFlushAsync(IPersistentCollection collection, CancellationTo
 			reached = false;
 			processed = false;
 		}
-
-		public Task<ICollection> GetOrphansAsync(string entityName, IPersistentCollection collection, CancellationToken cancellationToken)
```
hazzik: This is a breaking change.  Need to bring it back.

bahusoid: Done. I've noticed one more lost protected `GetOrphansAsync` method. I've also realized that after my latest changes `IdentityRemove` method is no more used anywhere. So I've removed all the changes in it and made it Obsolete

hazzik:
@fredericDelaporte could you please review this PR?

PR #2058, opened 2019-03-15
DB2 dialect enhancements

hazzik:
- Add schema update capability
- Fix generating parameter names
- Register bit operations
- Fix `substring` function registration 
- Fix `current_timestamp` function registration
- Fix `round` function registration
- Fix timestamp resolution in ticks
- Fix DbType.Guid type registration
- Fix DbType.Byte handling

Replaces #474

### DIFF
--- a/src/NHibernate/Driver/DB2DriverBase.cs
+++ b/src/NHibernate/Driver/DB2DriverBase.cs
@@ -2,6 +2,7 @@ using System;
 using System.Data;
 using System.Data.Common;
 using NHibernate.Engine;
+using NHibernate.SqlTypes;
 
 namespace NHibernate.Driver
 {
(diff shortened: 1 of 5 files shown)


src/NHibernate/Driver/DB2DriverBase.cs
```diff
@@ -7,6 +11,8 @@ namespace NHibernate.Driver
 	/// </summary>
 	public abstract class DB2DriverBase : ReflectionBasedDriver
 	{
+		private static readonly SqlType GuidSqlType = new SqlType(DbType.Binary, 16);
```
bahusoid: Is it used?

hazzik: No

src/NHibernate/Type/ByteType.cs
```diff
@@ -43,9 +43,10 @@ public override System.Type PrimitiveClass
 
 		public override void Set(DbCommand cmd, object value, int index, ISessionImplementor session)
 		{
-			cmd.Parameters[index].Value = Convert.ToByte(value);
-		}
+			var dp = cmd.Parameters[index];
+			dp.Value = dp.DbType == DbType.Int16 ? Convert.ToInt16(value) : Convert.ToByte(value);
```
hazzik: DB2 does not support byte native (as many RDBMS) and has a problem handling it. If the value is of type byte the provider remaps the parameter as Binary. Then it throws `InvalidCastException` trying to cast the value to `byte[]`.

hazzik:
>my local db2 installation is no more working

I'm using this docker image (gladly they have made it at least working from a first attempt): https://hub.docker.com/_/db2-developer-c-edition (need to "purchase" it). Also, it seems the image accepts only "testdb" as a database name.

Issue #2054, opened 2019-03-13
Incorrect SQL when using Contains in Select and Distinct


bacobart:
I'm using LINQ with the MSSQL2012 driver, NHibernate version 5.2.4. Executing below query results in the following exception:

> ORDER BY items must appear in the select list if SELECT DISTINCT is specified

```
public class TestEntity
{
	public virtual int Id { get; set; }
	public virtual string PropertyA { get; set; }
	public virtual string PropertyB { get; set; }
}

var result = session.Query<TestEntity>()
	   	    .Select(x => x.PropertyA.Contains("test") ? x.PropertyA : x.PropertyB)
		    .OrderBy(a => a)
		    .Distinct()
		    .ToList();

```

The generated SQL looks is listed below. NHibernate is adding PropertyA and PropertyB in the select, but not in the order by.

```
SELECT DISTINCT CASE 
		WHEN userquery_0_.PropertyA LIKE ('%' + @p0 + '%')
			THEN 1
		ELSE 0
		END AS col_0_0_
	,userquery_0_.PropertyA AS col_1_0_
	,userquery_0_.PropertyB AS col_2_0_
FROM [TestEntity] userquery_0_
ORDER BY (
		CASE 
			WHEN userquery_0_.PropertyA LIKE ('%' + @p0 + '%')
				THEN userquery_0_.PropertyA
			ELSE userquery_0_.PropertyB
			END
		) AS
```

fredericDelaporte:
If this is a regression, can you tell in which latest version it was working as expected?

The trouble comes from NHibernate choosing to evaluate part of the projection in-memory, after execution of the query, while it cannot do that for the order-by and so fully translate it to SQL.

But this causes the order-by to be based on an expression which is not part of the select, and since SQL-Server 2008, SQL-Server does no more support having a select distinct with an order-by using expressions which are not also selected.

And anyway, it also wrecks the query semantic, since the distinct will no more be applied only to the final expression but also to both of its underlying columns.

It looks like we should disable in memory evaluation of projections when a distinct is used.

bacobart:
Seems like this broke between 4.0.4.4000 and 4.1.1.4000. I thought it broke in 5.x because we are upgrading to 5.x (from 4.0.4.4000), but after some testing it seems this was already broken in 4.1.1.4000.

maca88:
[...]

PR #2061, opened 2019-03-15
Reduce cast usage for COUNT aggregate and add support for Mssql count_big

maca88:
This is a continuation of #2036, which extends the cast reduce usage for count aggregate function. In addition, this PR introduces `count_big` sql function that is used when using `LongCount` and `MsSql2000Dialect` overrides it to use the specific Sql Server `count_big` function (by default `count` is used).


src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -584,5 +601,118 @@ protected HqlTreeNode VisitNewArrayExpression(NewArrayExpression expression)
...
+			}
+
+			var persister = _parameters.SessionFactory.GetEntityPersister(entityName);
+			var index = persister.EntityMetamodel.GetPropertyIndexOrNull(memberExpression.Member.Name);
+			return !index.HasValue
+				? TypeFactory.GetDefaultTypeFor(expression.Type) // Not mapped
+				: persister.EntityMetamodel.PropertyTypes[index.Value];
+		}
+
+		private string TryGetEntityName(MemberExpression memberExpression)
```
hazzik: This is not entirely correct.

maca88: You are right, I forgot about custom entity names. This method was "enhanced" and moved to `ExpressionsHelper` class as it is also used by #1996, #2036, #2079, but was not updated here. Nonetheless, both methods do not work when having custom entity names. I realized that `IEntityNameProvider` exists and is implemented by the linq query which I will use to calculate the correct entity name instead of using `TryGetGuessEntityName` method. I will update the mentioned PRs when the method is fixed.

maca88: I took me more than expected as there are a lot mapping features that the method needs to be aware of in order to travese them, but as far as I know the method is now as accurate as it could be. Unfortunately it still needs to guess the entity name when casting is involved (like `((PizzaOrder)o.Order).Name`), but there is nothing we can do about it. I 've added a test for entity names in #2054. 
As this PR has an extension for `ISQLFunction` similar to what am I planning to do in #2061, I will update it after the other one is finished.

maca88:
Rebased and renamed to `ISQLFunction` extension interface as it will be used in all `ISQLFunction` implementations which I will address in a separate PR. Initially, I planned to deprecate `ISQLFunction.ReturnType` in favor of the newly added `ISQLFunctionExtended.GetEffectiveReturnType` method but it turned out that it would produce an unwanted breaking change. For example the following test would fail in Sql Server:
https://github.com/nhibernate/nhibernate-core/blob/d4206e24312b323e1da2f49affc0225d05866a3c/src/NHibernate.Test/QueryTest/CountFixture.cs#L23-L24
because `count` effective return type in Sql Server is an `int`. So the idea is that `ISQLFunction.ReturnType` should return the same type no matter what database is used, which is basically what it does today, except for some functions which IMO should be corrected:
- length: in Firebird and Oracle dialects returns `Int64`, SysbaseASE9 returns `string` (very suspicious) and in other dialect returns `Int32`.
- mod: in Firebird dialect returns `Double` where in other dialects returns `Int32`

The intention of `ISQLFunctionExtended.GetEffectiveReturnType` method is be used by higher level api like linq in order to determine whether a cast or a transparent cast is needed.

src/NHibernate/Dialect/MsSql2000Dialect.cs
```diff
@@ -286,7 +286,8 @@ protected virtual void RegisterKeywords()
 
 		protected virtual void RegisterFunctions()
 		{
-			RegisterFunction("count", new CountBigQueryFunction());
+			RegisterFunction("count", new CountQueryFunction());
+			RegisterFunction("count_big", new CountBigQueryFunction());
```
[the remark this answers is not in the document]
maca88: > Isn't this still causing the breaking change you were writing about?

For HQL nothing is changed, the new `GetEffectiveReturnType` method is used only for Linq queries. `CountQueryFunction.ReturnType` still returns `Int64`, which is what HQL uses to determine the type, so hql queries using `count` will always be casted to `Int64`.

src/NHibernate/Hql/Ast/ANTLR/SqlGenerator.g
```diff
@@ -150,7 +150,7 @@ selectExpr
 	;
 
 count
-	: ^(COUNT { Out("count("); }  ( distinctOrAll ) ? countExpr { Out(")"); } )
+	: ^(c=COUNT { OutAggregateFunctionName(c); Out("("); }  ( distinctOrAll ) ? countExpr { Out(")"); } )
```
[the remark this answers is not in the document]
maca88: >  I mean, was this causing the SQL-Server HQL count to count_big mapping to be ignored by this piece of code, causing NHibernate to anyway use SQL-Server int32 count?

Correct, `count` was hardcoded, which was the reason why changing from `count` to `count_big` in `CountBigQueryFunction` didn't work.

src/NHibernate/Dialect/Function/ISQLFunction.cs
```diff
@@ -41,4 +43,47 @@ public interface ISQLFunction
...
+			IMapping mapping,
+			bool throwOnError)
+		{
+			if (!(sqlFunction is ISQLFunctionExtended extendedSqlFunction))
+			{
+				try
+				{
+#pragma warning disable 618
+					return sqlFunction.ReturnType(argumentTypes.FirstOrDefault(), mapping);
+#pragma warning restore 618
```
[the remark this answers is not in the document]
maca88: Correct, they will be needed in #2316 instead. Rebased and removed the pragma warning.

src/NHibernate/Dialect/Function/ISQLFunctionExtended.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace NHibernate.Dialect.Function
+{
+	// 6.0 TODO: Merge into ISQLFunction
+	internal interface ISQLFunctionExtended : ISQLFunction
+	{
+		/// <summary>
+		/// The function name or <see langword="null"/> when multiple functions/operators/statements are used.
+		/// </summary>
+		string FunctionName { get; }
```
hazzik, line 13: I think this should be renamed to `Name` to reduce need of obsoleting things.

src/NHibernate/Hql/Ast/ANTLR/Tree/AggregateNode.cs
```diff
@@ -19,6 +20,19 @@ public AggregateNode(IToken token)
 		{
 		}
 
+		public string FunctionName
```
hazzik, line 23: I think this could be an overgeneralization and should be refactored into a node-specific code unless you see that it can be extended for functions other than `count` and `count_big`

src/NHibernate/Dialect/Function/ISQLFunctionExtended.cs
maca88: Why not, renamed in #2359.

src/NHibernate/Hql/Ast/ANTLR/Tree/AggregateNode.cs
maca88: This may be extended for string concatenation, which is supported by some databases ([Sql server](https://docs.microsoft.com/en-us/sql/t-sql/functions/string-agg-transact-sql?view=sql-server-ver15), [MySql](https://dev.mysql.com/doc/refman/8.0/en/group-by-functions.html#function_group-concat), [Postgresql](https://www.postgresql.org/docs/9.1/functions-aggregate.html) and [Oracle](https://docs.oracle.com/database/121/SQLRF/functions101.htm#SQLRF30030)), where different function names are used (`LISTAGG`, `GROUP_CONCAT`, `STRING_AGG`).

Nazgaul: @maca88 can you provide some sort of prototype of string_agg? I'm dying to use this with linq provider

maca88: To my knowledge it is not possible to implement it without modifying NHibernate source code. It is on my todo list for the next release, so unfortunately you will have to wait.

PR #2062, opened 2019-03-16
Fix invalid DML statements for static where in mapping

### LINKED ISSUE #2053: Dml Style Update fails with static where sql in mapping
Below query was working until 5.2.3. But with current master NH code it is no more working.
Fluent Mapping files includes a `Where("STATUS = 1")` filter. Normally NH adds the table alias name at the beginning of the filter, but in this case there is no alias added in the query, so it just adds the `.` character.

```c#
ISession.Query<TxnCycle>()
                .Where(x => x.StmtDate == null && accountNo.Contains(x.StmtAccountNo))
                .Update(u => new TxnCycle() { StmtDate = eodDate });
```
Output Sql:
```
NHibernate.SQL|
update OCN_ISS.TXN_CYCLE set STMT_DATE=:p0
  where ( .STATUS = 1) and (STMT_DATE is null)
    and (STMT_ACCOUNT_NO in (:p1));
:p0 = 20190310 [Type: Int32 (0:0:0)], :p1 = '1283' [Type: String (0:0:0)]
```

maca88:
The issue was introduced with #1931, which is not yet released. This PR uses a dedicated method that replaces the placeholder in the template, taking care of the dot when the alias is empty.

Fixes #2053


hazzik:
Could the test be incorporated into 1931?

Issue #2052, opened 2019-03-12
Possible wrong proxy determination which could lead to a sql exception


micmerchant:
Hello, 

this is a follow up to issue #2043.

The nhiberante [documentation ](https://nhibernate.info/doc/nhibernate-reference/performance.html) states in **21.1.3. Single-ended association proxies** that the id of proxies can be accessed without initialization

> Certain operations do not require proxy initialization
> - Equals(), if the persistent class does not override Equals()
> - GetHashCode(), if the persistent class does not override GetHashCode()
> - The identifier getter method

Let me explain one of our core concepts of our schema. We are heavily using union-subclassing inheritance with abstract base classes. Some of our models have a proxy definition some not. We are also mixing hbm mappings with fluent automapping.

Please don't blame me for our database schema (it's not mine), or how my (former) co-workers were/are misusing it :)

So let's say there are the following tables:
```javascript
public class EntityClassProxy : IEntity
{
	public virtual Guid Id { get; set; }
	public virtual string Name { get; set; }
}

class SubEntityInterfaceProxy : EntityClassProxy, ISubEntityProxy
{
	public virtual string AnotherName { get; set; }
}

class AnotherSubEntityInterfaceProxy : EntityClassProxy, IAnotherSubEntityProxy
{
	public virtual string AnotherName { get; set; }
}
```
as well the following table with a lookup:
```javascript
public class EntityWithSuperClassInterfaceLookup
{
	public virtual Guid Id { get; set; }
	public virtual string Name { get; set; }
	public virtual IEntity EntityLookup { get; set; }
}
```
with the following mapping: The base class proxy definition is the class itself. The subclasses proxy definitions point to interfaces.

```javascript
protected override HbmMapping GetMappings()
{
	var mapper = new ModelMapper();

	mapper.Class<EntityClassProxy>(rc =>
	{
		// calling the id getter of the proxy triggers a database query
		rc.Proxy(typeof(EntityClassProxy));

		// calling the id getter of the proxy doesn't trigger a database query
		//rc.Proxy(typeof(IEntity)); 

		rc.Id(x => x.Id);
		rc.Property(x => x.Name);
	});

	mapper.UnionSubclass<SubEntityInterfaceProxy>(rc =>
	{
		rc.Proxy(typeof(ISubEntityProxy));

		rc.Property(x => x.AnotherName);
	});

	mapper.UnionSubclass<AnotherSubEntityInterfaceProxy>(rc =>
	{
[...]

PR #2066, opened 2019-03-18
Tests for proxy interface handling

bahusoid:
Related to #2052 and test case for #2271

Added tests to clarify expected behavior.
In 5.2 both implicit and explicit interfaces are not leading to proxy initialization on interface `Id` access.

**Update:** After #2264 implicit interfaces are **not** leading to proxy initialization on `Id` access. And explicit interfaces lead to proxy initialization. Tests are adjusted according to this behavior.


hazzik:
Is this still valid?

bahusoid:
I believe yes. It's still unclear what's expected behavior and this tests depend on decisions made for #2085. In any case I think we should have tests that cover this behavior

bahusoid:
I've updated tests according to changes made in #2264. Also added test case for #2271

fredericDelaporte:
So do you consider my comment [here](https://github.com/nhibernate/nhibernate-core/issues/2184#issuecomment-518018943) is not valid?

>Your case will not be fixed. You should remove the explicit implementation and only use the implicit ones. Then it should work with NHibernate 5.3, but it may depend on how we fix #2085, if we fix it.
>
>With an explicit implementation of an interface, if the entity is mapped as the concrete class, there are no way NHibernate could tell if the Id property from the interface will actually match the Id from the mapped concrete class. So instead of doing some wild guess, it must trigger the proxy loading to simply call it and see what it yields.

For me, this #2066 PR and its issue should be rejected.

bahusoid:
> So do you consider my comment here is not valid?

I don't get it. It's not related to #2184. In #2184 - user wants to access identifier via interface without proxy initialization under different name in entity it's `EntityId` and in interface it's `Id`. So it's really no difference implicit or explicit implementation - proxy will be initialized anyway.

This PR just adds tests that shows how identifier column is handled when accessed via implicit/explicit interfaces in various scenarios. So why do you want to reject it? I see no such tests in existing codebase.

fredericDelaporte:
I had not noticed the name was not the same. And that is not the point anyway. Two interface members having the same name and signature have no reason to be the same. That would be quite a bold assumption.

bahusoid:
> Two interface members having the same name and signature have no reason to be the same.

Well in 5.2 both implicit/explicit identifier access doesn't trigger proxy initialization. So if name was the same - it would work in #2085. And from you comment it seems you expected proxy initialization even when name is the same.

In 5.3 you fixed explicit case. And this PR adds tests to make sure it's working  - so what's your reasons for rejecting.

fredericDelaporte:
The reason seem to me clearly written in above quote from myself. It is now re-staten [here](https://github.com/nhibernate/nhibernate-core/pull/2264#issuecomment-553116363) too:

> I consider this behavior of explicitly implemented `Id` not leading to proxy initialization by yielding the id value very debatable, because for correctness, an `Id` which is not implemented by the declared id in mapping cannot be assumed as being the entity id. It can be something else, and unless de-compiling and analyzing the concrete class, NHibernate cannot tell.

bahusoid:
And? Why do you think that those tests suggest otherwise? I agree with that. So why tests that make sure that proxy initialization actually occurs for explicit implementation and no proxy initialization for implicit should be rejected?

fredericDelaporte:
Sorry, I was assuming this PR was done in the context of #2271, still considering

> It should not.

(trigger lazy load).

So why not testing this, but it is in my mind already covered [here](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate.Test/StaticProxyTest/StaticProxyFactoryFixture.cs#L435). It does test that an explicitly implemented member is not "short-circuit" to its implicitly implemented "counter-part". That is precisely what causes an explicitly implemented "id" to trigger lazy-load.

bahusoid:
> It does test that an explicitly implemented member is not "short-circuit" to its implicitly implemented "counter-part". 

I wrote my tests a while ago - see no point dropping them now. Various scenarios for interface handling in one place. Also your tests seem to check some internal specifics of generated proxy. And my tests make no assumptions what should cause proxy initialization  - they just test actual initialization. So they more abstract from actual implementation.

Regarding #2271. Maybe it's just invalid test case and issue - will double check tomorrow.

hazzik:
Checked locally that there is no changes in async code.

PR #2069, opened 2019-03-19
Fix transient detection for proxy (#2045)

### LINKED ISSUE #2043: System.Reflection.TargetException when an interface is used as class mapping proxy definition

fredericDelaporte:
Fixes #2043 in v5.1.x


hazzik:
Was the async regenerated or just cherry-picked?

fredericDelaporte:
Yes I have run the generator, no changes occurred.

PR #2070, opened 2019-03-19
Fix determination of interface proxies

### LINKED ISSUE #2067: Wrong proxy built for base class with interfaced sub-classes
When a proxy for a base class is built (without mapping the proxy as an interface), a wrong proxy is built if the base class has sub-classes mapping their proxy as an interface.

The built proxy for the base class does not inherits from the base class. It should.

This causes a type cast exception when attempting a `Load<BaseClass>(id)`.

(This is extracted out of #2052, which suspect this bug as being the cause of the trouble in  #2052 (interface access causing a proxy load since v5.1). But the built proxy for the base class was also wrong in 4.x/5.0.x, it was also not inheriting from the base class. So I consider we need to handle separately the proxy building issue from the proxy loading one.)

hazzik:
The logic determining if we have an interface proxy or not moved to the tuplizer.

Fixes #2067


src/NHibernate/Proxy/StaticProxyFactory.cs
```diff
@@ -54,7 +54,21 @@ public override INHibernateProxy GetProxy(object id, ISessionImplementor session
...
 				ComponentIdType);
-			_cacheEntry = new ProxyCacheEntry(IsClassProxy ? PersistentClass : typeof(object), Interfaces);
+			_cacheEntry = new ProxyCacheEntry(GetProxyBaseType(PersistentClass), Interfaces);
+		}
+
+		private static System.Type GetProxyBaseType(System.Type type)
+		{
+			if (type.IsClass)
+			{
+				for (var t = type; t != null; t = t.BaseType)
```
bahusoid: It seems `session.Load<Cat>(catId)` test is going to fail If modify `Cat` by adding some unmapped base class.

hazzik: No she'll be fine.

hazzik: No, she'll be right.

hazzik:
Probably this would not work as `PersistentClass` of the `ILazyInitializer` has changed to `typeof(object)` in case of interface proxies, which is incorrect.

src/NHibernate.Test/NHSpecificTest/GH2067/BaseClassDerived1InterfaceDerived2ClassProxiesFixture.cs
```diff
@@ -0,0 +1,103 @@
...
+			{
+				using (var s = OpenSession())
+				{
+					var b = s.Load(typeof(Base), _id);
+					Assert.That(((Base) b).Id, Is.EqualTo(_id));
+				}
+			}
+
+			[Test]
+			public void ProxyForDerived1Interface()
```
bahusoid: This test is failing currently as generated proxy is not based on `Base` class

src/NHibernate.Test/NHSpecificTest/GH2067/BaseClassAndDerivedInterfaceProxiesFixture.cs
```diff
@@ -0,0 +1,113 @@
...
+			{
+				using (var s = OpenSession())
+				{
+					var b = s.Load<Base>(_id);
+					Assert.That(b.Id, Is.EqualTo(_id));
+				}
+			}
+
+			[Test]
+			public void ProxyForDerived1Interface()
```
bahusoid: This test is failing currently as generated proxy is not based on `Base` class

hazzik: I don't think that this test is valid. It would be extremely hard (and lead to inconsistencies) to find a _suitable_ base class for the proxy (there are several criteria: class shall be mapped, don't have an interface proxy, have all fields and methods overridable, etc). I don't think we should do it. As I said (or maybe I only thought(?)) in the issue discussion, the only concern we should fix is that the subclasses shall not affect the base class proxy. 

In this particular example if user wants a `Base` base class then they should call `Load<Base>(id)` and it will return a proxy that a base type `Base` and implements all subclass interfaces.

hazzik: Or at least not in this PR, as it is a separate issue to what @fredericDelaporte reported if I understand correctly.

bahusoid: Ok. It might be hard to implement and we can skip supporting this case for this matter. But I don't see any possible inconsistencies...

hazzik: I meant that user would not be able to predict if the code is correct or not as the result would be dependant on the hierarchy and mapping and get some unexpected result. So it would be inconsistent between similar but distinct hierarchies.

bahusoid: > there are several criteria: class shall be mapped, don't have an interface proxy, have all fields and methods overridable, etc

Hm. That's not the routine I have in mind.  I think we should take as a base first class specified as proxy in hierarchy and implement interfaces along the way to this class proxy.

hazzik: > I think we should take as a base first class specified as proxy in hierarchy and implement interfaces along the way to this class proxy.

I did not get it. 

Consider following hierarchy:

```csharp
//mapped
class A { public virtual int Id }
// not mapped
class B : A { public virtual int Something { get; set; } } 
// not mapped 
class C : B { public int Another {get; set;} }} 
interface ID { int Something { get; set; } int Another {get; set;} int MyProperty {get; set;} }
// mapped with proxy="ID"
class sealed D: C, ID { public int MyProperty {get; set;} } 
```

What shall the base class for proxy of `D` be?

bahusoid: `C` not mapped. So it's implicitly treated as `Proxy(typeof(C))`. And that's the first proxy class in hierarchy for `D`.
And till `C` class we have only one interface proxy `ID`. So generated class should look like:
`class DProxy: C, ID"

bahusoid: Was too fast :) Not mapped.. I see your point

bahusoid: So to conclude: my desired proxy hierarchy is still possible and I think it might be useful in some scenarios. But it is not related to this particular fix and requires some additional logic to analyze base type mappings so maybe sometime later in the future. I've adjusted my tests accordingly.

bahusoid:
I think tests for this PR better to place in StaticProxyTest folder/subfolder.

src/NHibernate/Proxy/AbstractProxyFactory.cs
```diff
@@ -19,11 +19,7 @@ public abstract class AbstractProxyFactory: IProxyFactory
 		protected virtual MethodInfo SetIdentifierMethod { get; private set; }
 		protected virtual IAbstractComponentType ComponentIdType { get; private set; }
 		protected virtual bool OverridesEquals { get; set; }
-
-		protected bool IsClassProxy
-		{
-			get { return Interfaces.Length == 1; }
-		}
+		protected internal bool IsClassProxy { get; internal set; }
```
bahusoid: Maybe should be true by default?

bahusoid: And why internal properties in public class? I think it should stay simply protected.

hazzik: > why internal properties in public class? I think it should stay simply protected.

The value is set outside the class so property is `protected internal`. I do not want consumers to change the value so setter is `internal`.

bahusoid: Ok. I just find it strange that we allow to create custom proxy based on this class but doesn't really allow to adjust this property in any way considering that this value must be set outside.

hazzik: I did not want to change the behaviour. It was calculated property, so I moved calculation outside.

src/NHibernate.Test/NHSpecificTest/GH2067/BaseClassAndDerivedInterfaceProxiesFixture.cs
hazzik: >but desired proxy class stays: `class DProxy: C, ID`

There is a trap: a field of `C` is not virtual and will always be null, which is undesirable.

hazzik:
I'm going to drop @bahusoid's tests as I believe the existing test cases do cover the required cases.

src/NHibernate/Proxy/Map/MapProxyFactory.cs
```diff
@@ -10,12 +10,22 @@ public class MapProxyFactory : IProxyFactory
...
+		{
+			this.entityName = entityName;
+		}
+
+		[Obsolete("Please use constructor accepting entityName instead.")]
 		public void PostInstantiate(string entityName, System.Type persistentClass, ISet<System.Type> interfaces,
 																MethodInfo getIdentifierMethod, MethodInfo setIdentifierMethod,
 																IAbstractComponentType componentIdType)
 		{
+			//6.0 TODO: throw NotSupportedException
```
bahusoid: Strange TODO inside obsolete method. Shouldn't we simply drop this method in 6.0? And when obsoleting something we also tend to add `//Since 5.3 comment`

hazzik: It’s not strange. This method would stay on the class (with additional argument) because of the interface, but shall not be supported.

Yeah, I’ve forgot about “since 5.3”

bahusoid:
Regarding dropping my tests. I'm fine wit it - I've added them with the wrong idea for "interface proxy" work. So they are at least excessive. But I still think that tests should be moved to StaticProxyTest

src/NHibernate/Proxy/Map/MapProxyFactory.cs
bahusoid: That's not so obvious that new override is meant. But why don't add override now?

src/NHibernate/Proxy/AbstractProxyFactory.cs
```diff
@@ -42,8 +43,28 @@ protected bool IsClassProxy
 			SetIdentifierMethod = setIdentifierMethod;
 			ComponentIdType = componentIdType;
 			OverridesEquals = ReflectHelper.OverridesEquals(persistentClass);
+			IsClassProxy = isClassProxy;
```
fredericDelaporte: To avoid breaking external implementations of `AbstractProxyFactory`, we should call the obsolete method here, which would do most of the initialization. Then the new method would override `IsClassProxy` value with the new parameter value.

src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
```diff
@@ -108,10 +107,9 @@ protected override IInstantiator BuildInstantiator(PersistentClass persistentCla
 			}
 		}
 
-		protected override IProxyFactory BuildProxyFactory(PersistentClass persistentClass, IGetter idGetter,
-		                                                            ISetter idSetter)
+		protected override IProxyFactory BuildProxyFactory(PersistentClass persistentClass, IGetter idGetter, ISetter idSetter)
 		{
-			bool needAccesorCheck = true; // NH specific (look the comment below)
+			bool isInterface = false; // NH specific (look the comment below)
```
fredericDelaporte: In my opinion, the comment is obsolete.

src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
```diff
@@ -120,27 +118,27 @@ protected override IInstantiator BuildInstantiator(PersistentClass persistentCla
...
 
-			if (_proxyInterface != null && !_mappedClass.Equals(_proxyInterface))
+			if (_proxyInterface != null && _mappedClass != _proxyInterface)
 			{
 				if (!_proxyInterface.IsInterface)
 				{
 					throw new MappingException("proxy must be either an interface, or the class itself: " + EntityName);
 				}
-				needAccesorCheck = false; // NH (the proxy is an interface all properties can be overridden)
+				isInterface = true; // NH (the proxy is an interface all properties can be overridden)
```
fredericDelaporte: The comment is no more needed here.

src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
```diff
@@ -120,27 +118,27 @@ protected override IInstantiator BuildInstantiator(PersistentClass persistentCla
...
 				proxyInterfaces.Add(_proxyInterface);
 			}
 
 			if (_mappedClass.IsInterface)
 			{
-				needAccesorCheck = false; // NH (the mapped class is an interface all properties can be overridden)
+				isInterface = true; // NH (the mapped class is an interface all properties can be overridden)
```
fredericDelaporte: The comment is no more needed here.

src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
hazzik: Removed

src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
hazzik: Removed

src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
hazzik: Removed

src/NHibernate/Proxy/AbstractProxyFactory.cs
hazzik: Done

PR #2074, opened 2019-03-20
Release 5.1.5

fredericDelaporte:
We may have a missing possible breaking change to add, depending on further investigations on #2052.

The missing possible breaking would be the fact that all interfaces of the proxy base class are now by default implemented and delegated to the loaded state. That is a change I was convinced the static proxy factory has done. But it seems some @bahusoid tests contradict this.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 5.1.4.{build}
+version: 5.1.5.{build}
 image: Visual Studio 2017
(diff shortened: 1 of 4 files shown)


fredericDelaporte:
No missing breaking change to add, due to #2085, which causes "duplicated" interface methods to be actually implemented by the base class method override, as with the dynamic proxy.

PR #2078, opened 2019-03-20
Avoid unnecessary join for entity comparisons in with clause

bahusoid:
Hql queries like:
`from Parent p inner join Child c with c.Parent = p`
Currently lead to unnecessary join for `c.Parent` in SQL. Issue is not entity join specific and can be reproduced for fetch joins also.


bahusoid:
Accidentally stumbled upon hibernate fix for this issue: https://github.com/hibernate/hibernate-orm/commit/17b8d2e294e51107d6c3af45bc306f53dc3dfe18
Hmm.. Different approach...

PR #2081, opened 2019-03-21
Force join for comparisons in hql when not null entity key can represent null entity

### LINKED ISSUE #1274: NH-3117 - Query on one-to-one property returns incorrect results
**Neal Groothuis** created an [issue](https://nhibernate.jira.com/browse/NH-3117) — <time datetime="2012-04-17T14:01:22.891Z" title="2012-04-17T14:01:22.891Z">17th April 2012, 14:01:22</time>:
> Given a pair of classes with a one-to-one relationship, attempting to query for items which do not have a corresponding item in the other class always returns an empty set.
> 
> For example, if I have Person and Employee tables a la section 5.1.1 of the manual (http://nhibernate.info/doc/nh/en/index.html#mapping-declaration-onetoone), the HQL query "from Person as person where person.Employee is null" returns an empty result set, even if there are entries in the Person table which do not have corresponding entries in the Employee table.
> 
> The attached project illustrates this.  It is configured to talk to a local SQL Server 2008 database called "nhibernate", and DDL to initialize the tables is included in the file DDL.sql.
> 
---
**Luke Horsley** added a comment — <time datetime="2014-02-25T11:06:09.111Z" title="2014-02-25T11:06:09.111Z">25th February 2014, 11:06:09</time>:
> This issue is a real major issue which appears to affect HQL, Criteria and LINQ. Rather than performing a join to the employees table it merely checks the person identifier is not null.
> 
---
**Frédéric Delaporte** added a comment — <time datetime="2015-04-17T09:53:21.682Z" title="2015-04-17T09:53:21.682Z">17th April 2015, 9:53:21</time>:
> If the example link was used as it is currently (<updated link>(http://nhibernate.info/doc/nh/en/index.html#mapping-declaration-onetoone)), the query is on the constrained side, meaning the other side forcibly exists.
> So this would not be a bug.
> By the way, in current example person id is declared as foreign. This means it is supposed to be generated on employee side, further enforcing the fact that in this model a person cannot exists without an employee entry.
> (Yes this looks functionally a bit inverted, but technically this is what this example does.)
> 
> But the attachment is indeed quite different. It is exactly like the case for which I was looking whether a bug was already reported or not.
> In the attachment case, both entities can live independently, having both their own identity generator. It is not the "true" one-to-one case, but the one relying on 'one-to-one property-ref' for matching on the side not holding the foreign-key, the other side using many-to-one.
> And yes, there is a querying bug in this case as described in description of this jira issue.
> 
> Neal, may you convert your attachment to a nunit case please? It would help for this issue to get solved. (See https://nhibernate.jira.com/browse/NH-2290 for an example nunit attachment, or see http://nhibernate.info/blog/2008/10/03/the-best-way-to-solve-nhibernate-bugs-submit-good-unit-test.html )
> 
---
**Neal Groothuis** added a comment — <time datetime="2015-04-18T15:00:02.145Z" title="2015-04-18T15:00:02.145Z">18th April 2015, 15:00:02</time>:
> Thanks for picking this up!  Unfortunately, I no longer have convenient access to a .NET development environment.
> 
---
**Frédéric Delaporte** added a comment — <time datetime="2015-09-28T13:48:07.860Z" title="2015-09-28T13:48:07.860Z">28th September 2015, 13:48:07</time>:
> It appears this bug occurs with "true" one-to-one cases too (at least with NH 4.0.3.4000).
> 
> "from Employee as e where e.Person is null" and "from Employee as e where e.Person.Id is null" get translated to a SQL query like "select e.* from EMPLOYEE as e where e.EMPLOYEE_ID is null".
> Furthermore, the similar bug occurs on "is not null" condition.
> 
> Workaround: testing nullity of a non-null-able property of Person other than its identifier (if we have such a property).
> 
> (I will try to take the time for uploading a test case soon.)
> 
---
**Frédéric Delaporte** added a comment — <time datetime="2017-01-23T18:44:39.954Z" title="2017-01-23T18:44:39.954Z">23rd January 2017, 18:44:39</time>:
> At last, here is the test case. Unfortunately, I have messed up with the drag & drop and put directly some files instead of just the zip. Sorry for the inconvenience. I am not allowed to delete them myself.
> 
---
**Frédéric Delaporte** added a comment — <time datetime="2017-01-23T22:25:32.654Z" title="2017-01-23T22:25:32.654Z">23rd January 2017, 22:25:32</time>:
> Now I have written this test case, it looks to me this issue is indeed a "work as intended" case for the `is null` test the opener reported.
> With NHibernate, navigating relationships without an explicit join is done with an inner join. The first two examples of <"select clause" documentation>(http://nhibernate.info/doc/nhibernate-reference/queryhql.html#queryhql-select) illustrates this.
> 
> So you can not test for non existence without an explicit join.
> 
> Your query should simply be written as:
> `from Employee as e left join e.Person as p where p is null`
> This one works.
> 
> Still, the "optimization" causing the test to occurs on parent side give some muddy result for
> `from Employee as e where e.Person is not null`
> This case should be equivalent to an inner join, but it is not, it does not filter out employee without person.
> A better workaround than my previous one is then:
> `from Employee as e join e.Person as p`
> 
> So now I will re-upload a test case with only the `not null` test case (hopefully without messing with the drag&drop again...).
>

bahusoid:
Fixes #1274 (criteria is not fixed, but criteria doesn't support implicit joins. So in criteria it's up to user to properly create joins  for such associations)

Add force join for associations with `EntityType.IsNullable == true` (like not constrained  one-to-one or associations with not-found="ignore" mapping):
https://github.com/nhibernate/nhibernate-core/blob/3d92bf4008c37b2aacc745c764450cd9c13c0947/src/NHibernate/Type/EntityType.cs#L294-L298

So when such property is used in comparisons like `where alias.OneToOne is null` we must add join and do comparison with joined table.


src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
```diff
@@ -409,12 +412,18 @@ private void DereferenceEntity(EntityType entityType, bool implicitJoin, string
...
+				joinIsNeeded = generateJoin || (Walker.IsInSelect && !Walker.IsInCase) || (Walker.IsInFrom && !Walker.IsComparativeExpressionClause)
+				               || comparisonWithNullableEntity;
 			}
 
 			if ( joinIsNeeded ) 
 			{
 				DereferenceEntityJoin( classAlias, entityType, implicitJoin, parent );
+				if (comparisonWithNullableEntity)
+				{
+					_columns = FromElement.GetIdentityColumns();
```
bahusoid: I'm not sure it's the best way to achieve required result.
This place detects entity comparisons like `where alias.OneToOne is null`. By default such comparisons are translated without join in to something like `where alias.OneToOneId is null`.
But instead we need join and do comparison with joined table `where oneToOneAlias.Id is null` so here we replace for this node columns with identity columns for joined table.

src/NHibernate/Hql/Ast/ANTLR/Tree/FromElement.cs
```diff
@@ -482,6 +482,17 @@ public IType GetPropertyType(string propertyName, string propertyPath)
 		}
 
 		public virtual string GetIdentityColumn()
+		{
+			var cols = GetIdentityColumns();
+			string result = StringHelper.Join(", ", cols);
```
hazzik: Any reason using `StringHelper.Join` instead of `string.Join`?

hazzik: Ok, got it now, just a method split.

bahusoid: Yep. It was there. I didn't pay attention. I think it can be replaced with `string.Join`...

hazzik:
Force pushed to resolve conflicts. The fixup commit got squashed.

R0nd:
Can we get this reviewed?

bahusoid:
> Adding joins for the general case, where foreign key are actual foreign keys, where a not-null foreign key can only mean the other side exists, is a bad move

Just to be clear `EntityType.IsNullable` has nothing to do with `not-null` attribute. It means not-null `id` can be treated as NULL entity for whatever reasons. And currently it's set for `true`  for exactly two cases: 
1) not constrained one-to-one or associations with not-found="ignore" mapping
2) associations with not-found="ignore" mapping

PR #2082, opened 2019-03-22
Use entities prepared by Loader in hql select projections

### LINKED ISSUE #2064: One-to-one properties not appearing in Select() projection result set
**Version:** 5.2.4

We're having an issue with creating projections (with Linq's `Select()`) of entities where one of the properties is a one-to-one map.

I've created a [repro](https://gist.github.com/jhinder/6c8ce3faf49b212ea7a514940ebe2fc5).

**Expected result:** `Results` is not null in both cases.

**Actual result:** `Results` is null in the first case (with the Select() projection), but in the second case (plain query) it's a correct, non-null object.

(We're using FluentNHibernate to generate the mappings, but the `<one-to-one />` mapping lines look fine in the generated XML.)

----

So far I've found out that `NHibernate.Type.OneToOneType.Hydrate()` is called twice each time, and the `owner` parameter has these values:

* First example (Select()):
  * First call: a `Results` object
  * Second call: `null`
* Second example (plain query):
  * First call: an `Owner` object
  * Second call: a `Results` object

Might a part of issue be that `owner` is null at one point, and thus `Hydrate`'s return value is null as well, leading a method further down the call stack to believe the object to be null?

bahusoid:
Was fixing #2064 
Also see this [comment](https://github.com/nhibernate/nhibernate-core/issues/2064#issuecomment-475253127) for details

**Update** Since #2064 is fixed by #2084 this PR can still be considered as an improvement that avoids additional entity lookup in session for projections.


src/NHibernate/Hql/Ast/ANTLR/Tree/SelectClause.cs
```diff
@@ -136,20 +138,22 @@ public void InitializeExplicitSelectClause(FromClause fromClause)
...
 
-						if (se != null && IsReturnableEntity(se))
+						if (IsReturnableEntity(se))
 						{
-							_fromElementsForLoad.Add(se.FromElement);
+							AddEntityToProjection(queryReturnTypeList.Count + j, se);
 						}
 					}
+
+					IList<IType> constructorArgumentTypeList = ConstructorNode.ToArgumentTypes(ctorSelectExpressions);
```
hazzik: This is just `_constructorNode.ConstructorArgumentTypeList`

PR #2084, opened 2019-03-23
Improve one-to-one handling in queries

### LINKED ISSUE #1125: NH-3178 - Exception when using one-to-one properties in a criteria projections
**Nathan Baulch** created an [issue](https://nhibernate.jira.com/browse/NH-3178) — <time datetime="2012-06-06T15:58:59.874Z" title="2012-06-06T15:58:59.874Z">6th June 2012, 15:58:59</time>:
> 
> I get the following exception when using a one-to-one property in a criteria projection:
> 
> NHibernate.Exceptions.GenericADOException: could not execute query
> ```
> [ SELECT this_.USERID as y0_ FROM USER this_ ]
> [SQL: SELECT this_.USERID as y0_ FROM USER this_] ---> System.IndexOutOfRangeException: Index was outside the bounds of the array.
>    at NHibernate.Loader.Criteria.CriteriaLoader.GetResultColumnOrRow(Object[] row, IResultTransformer customResultTransformer, IDataReader rs, ISessionImplementor session) in d:\GitHub\nhibernate\nhibernate-core\src\NHibernate\Loader\Criteria\CriteriaLoader.cs:line 101
>    at NHibernate.Loader.Loader.GetRowFromResultSet(IDataReader resultSet, ISessionImplementor session, QueryParameters queryParameters, LockMode<] lockModeArray, EntityKey optionalObjectKey, IList hydratedObjects, EntityKey[> keys, Boolean returnProxies) in d:\GitHub\nhibernate\nhibernate-core\src\NHibernate\Loader\Loader.cs:line 360
>    at NHibernate.Loader.Loader.DoQuery(ISessionImplementor session, QueryParameters queryParameters, Boolean returnProxies) in d:\GitHub\nhibernate\nhibernate-core\src\NHibernate\Loader\Loader.cs:line 468
>    at NHibernate.Loader.Loader.DoQueryAndInitializeNonLazyCollections(ISessionImplementor session, QueryParameters queryParameters, Boolean returnProxies) in d:\GitHub\nhibernate\nhibernate-core\src\NHibernate\Loader\Loader.cs:line 246
>    at NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters) in d:\GitHub\nhibernate\nhibernate-core\src\NHibernate\Loader\Loader.cs:line 1590
>    --- End of inner exception stack trace ---
>    at NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters) in d:\GitHub\nhibernate\nhibernate-core\src\NHibernate\Loader\Loader.cs:line 1599
> ```   
> Using many-to-one projections on the other hand does work. The difference seems to be the AbstractType.GetColumnSpan method which returns a hard coded zero in the OneToManyType override and a non-zero value in the ManyToOneType override.
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2012-06-06T16:44:57.349Z" title="2012-06-06T16:44:57.349Z">6th June 2012, 16:44:57</time>:
> Nathan Baulch, could you please provide an example or test case?
> 
---
**Nathan Baulch** added a comment — <time datetime="2012-06-06T17:29:28.534Z" title="2012-06-06T17:29:28.534Z">6th June 2012, 17:29:28</time>:
> The NH unit test project is a little overwhelming but I'm able to reproduce the issue by adding the following test to the MergeFixture class. I doubt this is an appropriate place for such a test to live permanently, I just needed a convenient place to use the one-to-one Person.Detail property.
> 
>     <Test>
>     public void OneToOneProjection()
>     {
>         using (var s = OpenSession())
>         {
>             var p = new Person {Name = "steve"};
>             new PersonalDetails {SomePersonalDetail = "I have big feet", Person = p};
>             s.Save(p);
>             s.Flush();
>             s.CreateCriteria<Person>().SetProjection(Projections.Property("Details")).List();
>         }
>     }
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2012-06-07T11:29:47.697Z" title="2012-06-07T11:29:47.697Z">7th June 2012, 11:29:47</time>:
> Nathan Baulch, it just not supported, see Oskar's explanation on NH-3176
> 
---
**Oskar Berggren** added a comment — <time datetime="2012-06-08T06:55:10.281Z" title="2012-06-08T06:55:10.281Z">8th June 2012, 6:55:10</time>:
> NH-3176 was about collections. one-to-one sounds more natural to support, technically there shouldn't be much difference from many-to-one. It might be a feature request rather then a bug though.
> 
---
**Nathan Baulch** added a comment — <time datetime="2012-06-11T07:22:50.800Z" title="2012-06-11T07:22:50.800Z">11th June 2012, 7:22:50</time>:
> My thoughts exactly Oskar, and if it can't be supported for technical reasons then a more meaningful error would be nice.
>

### LINKED ISSUE #1309: NH-3469 - Impossible to load one-to-one association with LINQ for composite-id
**Igor Kiselev** created an [issue](https://nhibernate.jira.com/browse/NH-3469) — <time datetime="2013-05-29T05:28:56.075Z" title="2013-05-29T05:28:56.075Z">29th May 2013, 5:28:56</time>:
> When you use one-to-one mapping for object, that have with LINQ provider it's impossible to load property that contains associated object if it is not explicitly loaded in select clause.
> 
> Example:
> ```xml
>  <class name="ExternalField">
>    <composite-id>
>      <key-property name="Id" />
>      <key-property name="Id2" />
>   </composite-id>
> </class>
> 
> <class name="Field">
>   <composite-id>
>     <key-property name="Id" />
>     <key-property name="Id2" />
>   </composite-id>        
>   <one-to-one name="ExternalField" lazy="false" fetch="join" />
> </class>
> ```
> ```csharp
> var fields = session.Query<Field>()
>   .ToList()
>   .Where(it => it.ExternalField != null).ToList();
> ```
> There will be no object in final list, even if DB have such. Same queries work when using NHibernate.Linq.
> 
> 
---
**Ricardo Peres** added a comment — <time datetime="2013-05-29T09:49:11.299Z" title="2013-05-29T09:49:11.299Z">29th May 2013, 9:49:11</time>:
> If you take out the first ToList() on your LINQ query, the problem appears:
> ```csharp
> var fields = session.Query<Field>().Where(it => it.ExternalField != null).ToList();
> ```
> ```sql
> select
>     field0_.Id as Id1_,
>     field0_.Id2 as Id2_1_
> from
>     Field field0_
> where
>     (
>         field0_.Id, field0_.Id2
>     ) is not null
> ```
> NHibernate is building a wrong SQL.
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2013-05-29T23:25:20.527Z" title="2013-05-29T23:25:20.527Z">29th May 2013, 23:25:20</time>:
> The problem in HQL
>

### LINKED ISSUE #1368: NH-3778 - Crash when performing a Linq query on a one-to-one mapped reference
**Craig Fowler** created an [issue](https://nhibernate.jira.com/browse/NH-3778) — <time datetime="2015-04-17T13:35:19.036Z" title="2015-04-17T13:35:19.036Z">17th April 2015, 13:35:19</time>:
> NHibernate fails to add the correct params to a SQL parameterised query when performing a Linq query upon an entity type which is mapped one-to-one.  This leads to a crash from the corresponding ADO driver when NHibernate attempts to execute the query.  A full reproduction solution is attached but in short, here's the problem:
> 
> 
> ```csharp
> var person = session.Load<Person>(1);
> 
> var results = session.Query<Employee>()
>   .Where(x => x.Person == person)
>   .ToList();
> ```
> 
> 
> That query - where the "Person" and "Employee" classes are mapped with a one-to-one relationship - crashes with the following exception:
> 
> 
> ```
> [ERROR] FATAL UNHANDLED EXCEPTION: NHibernate.Exceptions.GenericADOException: could not execute query
> [ select employee0*.Id as Id1_0_, person1_.Id as Id0_1_, employee0_.StartDate as StartDate1_0_, person1_.Name as Name0_1_ from Employee employee0_ left outer join Person person1_ on employee0_.Id=person1_.Id where employee0*.Id=@p0 ]
>   Name:p1 - Value:NHibernateBug.Person
> [SQL: select employee0*.Id as Id1_0_, person1_.Id as Id0_1_, employee0_.StartDate as StartDate1_0_, person1_.Name as Name0_1_ from Employee employee0_ left outer join Person person1_ on employee0_.Id=person1_.Id where employee0*.Id=@p0] ---> Mono.Data.Sqlite.SqliteException: SQLite error
> Insufficient parameters supplied to the command
>   at Mono.Data.Sqlite.SqliteStatement.BindParameter (Int32 index, Mono.Data.Sqlite.SqliteParameter param) <0x00000> in <filename unknown>:0 
>   at Mono.Data.Sqlite.SqliteStatement.BindParameters () <0x00000> in <filename unknown>:0 
>   at Mono.Data.Sqlite.SqliteCommand.BuildNextCommand () <0x00000> in <filename unknown>:0 
>   --- End of inner exception stack trace ---
>   at NHibernate.Loader.Loader.DoList (ISessionImplementor session, NHibernate.Engine.QueryParameters queryParameters, IResultTransformer forcedResultTransformer) <0x00000> in <filename unknown>:0 
>   at NHibernate.Loader.Loader.DoList (ISessionImplementor session, NHibernate.Engine.QueryParameters queryParameters) <0x00000> in <filename unknown>:0 
>   at NHibernate.Loader.Loader.ListIgnoreQueryCache (ISessionImplementor session, NHibernate.Engine.QueryParameters queryParameters) <0x00000> in <filename unknown>:0 
>   at NHibernate.Loader.Loader.List (ISessionImplementor session, NHibernate.Engine.QueryParameters queryParameters, ISet`1 querySpaces, NHibernate.Type.IType<] resultTypes) [0x00000> in <filename unknown>:0 
>   at NHibernate.Loader.Hql.QueryLoader.List (ISessionImplementor session, NHibernate.Engine.QueryParameters queryParameters) <0x00000> in <filename unknown>:0 
>   at NHibernate.Hql.Ast.ANTLR.QueryTranslatorImpl.List (ISessionImplementor session, NHibernate.Engine.QueryParameters queryParameters) <0x00000> in <filename unknown>:0 
>   at NHibernate.Engine.Query.HQLQueryPlan.PerformList (NHibernate.Engine.QueryParameters queryParameters, ISessionImplementor session, IList results) <0x00000> in <filename unknown>:0 
>   at NHibernate.Impl.SessionImpl.List (IQueryExpression queryExpression, NHibernate.Engine.QueryParameters queryParameters, IList results) <0x00000> in <filename unknown>:0
> ```
> 
> 
> The reproduction case attached uses a SQLite database, but I have also been able to reproduce it using an MS-SQL database.
> 
> Note that I wrote the reproduction case project using Mono, and thus it depends upon Mono.Data.Sqlite at present.  If you wish to execute the reproduction case on Windows/.NET, remove the reference to Mono.Data.Sqlite and download/install Sqlite from nuget.
> 
> To try the reproduction case (after making any changes for Sqlite), compile and execute the solution (there is only one project).  The ReproductionCase class contains the logic which triggers this crash.  The EntryPoint class just sets up and configures the session factory and db instance (including creating/exporting the schema).
> 
> Thanks
> 
---
**Craig Fowler** added a comment — <time datetime="2015-04-17T14:06:25.837Z" title="2015-04-17T14:06:25.837Z">17th April 2015, 14:06:25</time>:
> Oops, I forgot to mention that this issue also affects NH 3.3.3 (which is how we initially found it).
> 
> I am unsure if 3.x is still being supported with bugfixes but it would be nice if any fix were also backported as we are yet to make the upgrade to 4.0.
> 
---
**Victor Govorov** added a comment — <time datetime="2016-06-01T19:50:12.110Z" title="2016-06-01T19:50:12.110Z">1st June 2016, 19:50:12</time>:
> Person class doesn't have an equality operator implementation. In your case (in LinqToObject scenario) is better to compare objects via Objects.Equals method
> 
> ```csharp
> 
> var results = session.Query<Employee>()
>   .Where(x => Equals(x.Person, person))
>   .ToList();
> 
> ```
> 
> But Person.Equals method implementation may be quite complex and may not be converted to sql query. So in NH case the following code is more preferred and it produces the correct sql query
> 
> ```csharp
> 
> var results = session.Query<Employee>()
>   .Where(x => x.Person.Id == person.Id)
>   .ToList();
> 
> ```
> 
> 
> As my mind this is not a bug and issue may be closed.
> 
---
**Craig Fowler** added a comment — <time datetime="2016-06-01T20:12:27.389Z" title="2016-06-01T20:12:27.389Z">1st June 2016, 20:12:27</time>:
> I understand that.  However, in a many-to-one mapping scenario you can write equality operations just like that which are converted to SQL correctly without crashing. There is no need in that instance either to write an equality method implementation or operator overload. NH will correctly interpret it as a SQL equality comparison between the PK/FK values as applicable. NH only fails like this when the entities are mapped one-to-one.
> 
> If it is intentional that this does not work for one-to-one mappings and it cannot be implemented, surely a NotSupportedException would be preferable to producing and attempting to execute invalid SQL?
>

### LINKED ISSUE #2064: One-to-one properties not appearing in Select() projection result set

bahusoid:
Added proper column span calculation (mapping related checks extracted to  `GetOwnerColumnSpan`);
Adjusted `OneToOneType` to be properly handled as parameter and projection in queries;

1) Allows using one-to-one association in entity comparisons (fixes #1368)
2) Fixes one-to-one handling for composite keys (fixes #1309)
3) Fixes one-to-one projections (fixes #2064, fixes #1125)


src/NHibernate.Test/Associations/OneToOneFixture.cs
```diff
@@ -0,0 +1,167 @@
...
+	[TestFixture]
+	public class OneToOneFixture: TestCaseMappingByCode
+	{
+		[Test]
+		public void OneToOneCompositeQueryByEntityParam()
+		{
+			using (var session = OpenSession())
+			{
+				var e = session.Query<EntityWithCompositeId>().FirstOrDefault();
+				var loadedEntity = session.Query<Parent>().Where(p => p.OneToOneComp == e).FirstOrDefault();
```
bahusoid: Offtopic: Is there a way to see generated hql for given LINQ? I mean in some readable form not as node tree.

hazzik: No. The closest you can get is `To<Somehting>String` (I don’t remember the exact method name). It will create a human readable string for the tree.

bahusoid: Did you mean this?
`nhLinqExpression.ExpressionToHqlTranslationResults.Statement.AstNode.ToStringTree()`

It gives something like: 
`( query ( select_from ( from ( range NHibernate.Test.Associations.OneToOne.Parent p ) ) ( select p ) ) ( where ( == ( . p OneToOneComp ) ( : p1 ) ) ) ( take 1 ) )`
Is that the best we can get?

hazzik: Yes, that one. Do the same with desired HQL and compare the trees. This is what I usually do.

bahusoid: Ok. Thanks

hazzik:
Does it actually make sense to return `0` in `GetColumn` for `OneToOne` type?

hazzik:
What about criteria? Do the same queries work there?

bahusoid:
> Does it actually make sense to return 0 in GetColumn for OneToOne type?

Did you mean `GetColumnSpan`? Yeah otherwise owner entity thinks that one-to-one entity requires N columns in owner table. And it breaks some other checks in multiple places.

bahusoid:
> What about criteria? Do the same queries work there?

Will check...

src/NHibernate/Type/OneToOneType.cs
```diff
@@ -26,7 +26,7 @@ public override int GetColumnSpan(IMapping session)
 
 		public override SqlType[] SqlTypes(IMapping mapping)
 		{
-			return NoSqlTypes;
+			return GetIdentifierOrUniqueKeyType(mapping).SqlTypes(mapping);
```
hazzik: Is this change required? Hibernate did not change that.

src/NHibernate/Type/OneToOneType.cs
```diff
@@ -44,7 +44,8 @@ public override void NullSafeSet(DbCommand st, object value, int index, bool[] s
 
 		public override void NullSafeSet(DbCommand cmd, object value, int index, ISessionImplementor session)
 		{
-			//nothing to do
```
hazzik: Is this change required? Hibernate did not change that.

src/NHibernate/Type/OneToOneType.cs
bahusoid: Required to make work queries with parameters `.Where(x => x.OneToOne = param)`.
Seems it's not fixed in Hibernate.

src/NHibernate/Type/OneToOneType.cs
bahusoid: Same as this [comment](https://github.com/nhibernate/nhibernate-core/pull/2084/files#r269046031)

src/NHibernate/Hql/Ast/ANTLR/Tree/BinaryLogicOperatorNode.cs
```diff
@@ -88,8 +88,8 @@ protected void MutateRowValueConstructorSyntaxesIfNecessary(IType lhsType, IType
 
 			if (lhsType != null && rhsType != null)
 			{
-				int lhsColumnSpan = lhsType.GetColumnSpan(sessionFactory);
-				var rhsColumnSpan = rhsType.GetColumnSpan(sessionFactory);
+				int lhsColumnSpan = lhsType.GetParameterColumnSpan(sessionFactory);
+				var rhsColumnSpan = lhsType.GetParameterColumnSpan(sessionFactory);
```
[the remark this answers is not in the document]
bahusoid: So the only case they fixed is equality for joins with one-to-one `from A a, B b where a.b = b`

hazzik:
>Yeah otherwise owner entity thinks that one-to-one entity requires N columns in owner table. And it breaks some other checks in multiple places.

Can we instead find places where it expects `0`?

bahusoid:
> Can we instead find places where it expects 0?
Ok. Will check.

src/NHibernate/Type/OneToOneType.cs
hazzik: Do we need to make same change to the another overload of `NullSafeSet`?

bahusoid: No. It's about preparing command for owner table. At least from what I've seen in tests.

hazzik: Also, it seems that with this change the accused "hack" in hydrate is not necessary anymore.

bahusoid: Hm? Which accused hack?

src/NHibernate.Test/Associations/OneToOneFixture.cs
```diff
@@ -0,0 +1,167 @@
+using System;
+using System.Linq;
+using NHibernate.Cfg.MappingSchema;
+using NHibernate.Mapping.ByCode;
+using NUnit.Framework;
+
+namespace NHibernate.Test.Associations.OneToOne
```
hazzik: Namespace does not match the folder structure.

bahusoid: Intention is to hide entities from `NHibernate.Test.Associations` namespace. Not sure what's best way to proceed. Seems separate folder is excessive here.

src/NHibernate/Type/OneToOneType.cs
hazzik: This one: https://github.com/nhibernate/nhibernate-core/blob/55ffdb64d81dbc6a0b97d6c3b5a10939323ad8d6/src/NHibernate/Type/OneToOneType.cs#L98

src/NHibernate/Hql/Ast/ANTLR/Tree/BinaryLogicOperatorNode.cs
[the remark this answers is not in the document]
hazzik: Do we need to test this query?

src/NHibernate.Test/Associations/OneToOneFixture.cs
hazzik: You can make entities internal to the fixture class.

bahusoid: Not sure what you meant. I can't put them inside class. As that breaks db creation as our schema creator  tries to create tables that look something like `OuterClass+InnerClass`. And `+` doesn't work with  SQL Server.

I think I will just wrap entities in separate namespace.

hazzik:
How will this work with your other PR #2081?

bahusoid:
> How will this work with your other PR #2081?

It's unrelated. #2081 is about forcing join for association. Join for association is not affected by this PR. This pr fixes various explicit comparisons with one-to-one association (via parameter, null, other join)

src/NHibernate/Type/OneToOneType.cs
```diff
@@ -92,6 +96,16 @@ public override ForeignKeyDirection ForeignKeyDirection
 
 		public override object Hydrate(DbDataReader rs, string[] names, ISessionImplementor session, object owner)
 		{
+			if (owner == null && names.Length > 0)
```
bahusoid: This also fixes #2064. But #2082 still can be applied as it optimizes processing a bit by avoiding loaded entities "rehydrating".

src/NHibernate/Type/IType.cs
```diff
@@ -7,10 +7,10 @@ namespace NHibernate.Type
 {
 	internal static class TypeExtensions
 	{
-		public static int GetParameterColumnSpan(this IType type, IMapping sessionFactory)
+		public static int GetOwnerColumnSpan(this IType type, IMapping sessionFactory)
```
bahusoid: Not sure if it makes sense to add it to `IType` interface.

src/NHibernate/Type/OneToOneType.cs
bahusoid: No. This hack should stay in place (NH1611OneToOneIdentityFixture test for it) With embedded identifiers where entity is used as ID without this hack owner entity is sent to one-to-one `ComponentType.GetPropertyValues`. And exception is occurred as one-to-one expects different type for id.

Embedded identifier is an evil that should be obsoleted.

bahusoid:
> Can we instead find places where it expects 0

Done. Introduced `GetOwnerColumnSpan` for such places.

> What about criteria? Do the same queries work there?

Yes. It works with latest commit. Tests added.

src/NHibernate/Hql/Ast/ANTLR/Tree/BinaryLogicOperatorNode.cs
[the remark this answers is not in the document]
bahusoid: Test added.

src/NHibernate/Mapping/SimpleValue.cs
```diff
@@ -324,7 +324,7 @@ public bool IsSimpleValue
 
 		public virtual bool IsValid(IMapping mapping)
 		{
-			return ColumnSpan == Type.GetColumnSpan(mapping);
+			return ColumnSpan == Type.GetOwnerColumnSpan(mapping);
```
hazzik: This will break something called "SpecialOneToOneType".

hazzik: Which seems to be unsupported anyway.

bahusoid: Yeah. For the code below `ColumnSpan` is always 0. So it's never used. 
https://github.com/nhibernate/nhibernate-core/blob/c8b9ae6e7b4b839b21b22671e47adbba0707f0ad/src/NHibernate/Mapping/OneToOne.cs#L91-L103

Though I restored original behavior just in case...

hazzik: I think we should override logic for onetone.

src/NHibernate/Persister/Entity/AbstractPropertyMapping.cs
```diff
@@ -103,7 +103,7 @@ protected void AddPropertyPath(string path, IType type, string[] columns, string
 
 		protected internal void InitPropertyPaths( string path, IType type, string[] columns, string[] formulaTemplates, IMapping factory )
 		{
-			if (columns.Length != type.GetColumnSpan(factory))
+			if (columns.Length != type.GetOwnerColumnSpan(factory))
```
hazzik: Here again would not work for "special" one-to-one.

src/NHibernate/Mapping/SimpleValue.cs
bahusoid: I've added `GetOwnerColumnSpan` override to `SpecialOneToOne` in my latest commit. Or did you mean something else?

hazzik: I was thinking to implement override for `IsValid` in `OneToOne` to handle this logic.

PR #2088, opened 2019-03-26
Fix cacheable CreateSQLQuery throws on query with AddJoin

bahusoid:
Example of problem query:
```C#
s.CreateSQLQuery(query)
	.AddEntity("org", typeof(Organization))
	.AddJoin("emp", "org.employments")
	.SetCacheable(true)
	.List();
```
Thrown exception:
```
System.ArgumentException : If aliases is not null, then the length of aliases[] must equal the number of true elements in includeInTuple; aliases.Length=2; tuple 'true' length=1
Parameter name: aliases

   at NHibernate.Transform.CacheableResultTransformer.GetIncludeInTransform(IResultTransformer transformer, String[] aliases, Boolean[] includeInTuple) 
```


hazzik:
It seems similar fix can be used to fix #1125.

bahusoid:
> It seems similar fix can be used to fix #1125.

No. It's unrelated. #2084  is closer. The issue is due to `GetColumnSpan` returns 0 for one-to-one:
https://github.com/nhibernate/nhibernate-core/blob/c8b9ae6e7b4b839b21b22671e47adbba0707f0ad/src/NHibernate/Loader/Criteria/CriteriaLoader.cs#L131-L141

PR #2089, opened 2019-03-26
Bidirectional list fails if session only knows about child

RogerKratz:
When having a bidirectional inversed list, child seems to not fully "own" the collectionref when using session.Merge on child.

When running failing test, child will end up with order index -1 in db because when nhib's merge occur, the child is "unhooked" from its parent.

Probably closely related to #1531.


bahusoid:
I might be wrong  but from my understanding `inverse` instructs collection to be an owner. Not child entity itself. And it seems cascade mapping applies only for supplied entity to `Merge`. So you need to provide parent instead of child to make `Merge` work.

RogerKratz:
@bahusoid If I understand you correctly, no, I think you're wrong.
Owner: inverse=false
Non-owner: inverse=true
So in my test/example the child is the owner because the parent is marked with inverse=true.

bahusoid:
> inverse instructs collection to be an owner.

I meant owner/non-owner related logic is applied only for **collection**. Not for entity itself. So basically when flush is happened `parent.Children` handling behavior is changed accordingly (save children first or ...).

Child entity knows nothing about ownership.

RogerKratz:
> I meant owner/non-owner related logic is applied only for collection.

Sure, I'm not talking about cascading or similar. Still, it's a bidirectional reference, and child know it's parent (and orderindex) in the domain model, so why shouldn't it be possible to save child without session knowing about parent, if that's what you mean?

(Even with inverse=false it should work, but that's seldom/never used if ref is bidirectional I think?)

RogerKratz:
@bahusoid 
I just changed the title of this issue...
Inversed list fails if session only knows about child => Bidirectional list fails if session only knows about child
....maybe it makes it more clear?

bahusoid:
> and child know it's parent 

Child knows only what's in mapping. So it knows `Parent` but knows nothing that `Parent` has 'Children`.  `Merge`  only processes supplied entity and associations according to cascade settings. You have no cascade settings specified for `Parent`. So why do you expect it to be saved  as part of `Parent`collection?

RogerKratz:
> Child knows only what's in its mapping. So it knows Parent but knows nothing that Parent has Children. Merge only processes supplied entity and associations according to cascade settings. You have no cascade settings specified for Parent.
Agreed.

> So why do you expect it to be saved as part of Parent.Childrencollection?
I don't. Why do you think I expect that? ;) We must be misunderstanding each other... 
I expect "newChild" to be correctly persisted in the test.

bahusoid:
> Why do you think I expect that

Because you expect `Parent.Children` mapping logic be applied here (so index is filled correctly). I've just described you the way I think it works. And IMHO it works as expected.

RogerKratz:
Actually I dont. I expect **Child's** OrderIndex property and mapping to be used. 
Yes, the implementation of that property traverse the domain model (in this case to the parent). But from nhib's perspective (and the mapping), it should be irrelevant. It's just references in ram so to speak. And they "work" when Merge is called.

bahusoid:
I see. Than it's indeed misunderstanding. Sorry for all the fuss.

RogerKratz:
@bahusoid Thanks for your time and discussion. Glad I could finally make myself a bit clearer :)

hazzik:
https://stackoverflow.com/questions/1320966/nhibernate-mapping-list-doesnt-update-list-indexes

RogerKratz:
@hazzik IIRC, "bi-directional inverse list does not work at all" was true a couple of years ago. Nowadays it is/should be supported as of 6.3.3 in [documentation](https://nhibernate.info/doc/nhibernate-reference/collections.html).

The problem (if it is?) is probably the same as #1531. It only happens when merging entities. NHib then "detaches" the entities in the domain model from eachother and makes copies of the data. Due to this the references themself inside the model no longer points to the "correct" entities and the hack (if it is?) no longer works.

hazzik:
I was talking about bidirectional inverse list without a special column.

hazzik:
Thanks. This has now lives on master. Not sure though if we're going to fix this.

PR #2096, opened 2019-04-01
Avoid double proxy casting

bahusoid:
By adding `bool IsProxy(object entity, out INHibernateProxy proxy)` to `IProxyFactoryFactory`

### DIFF
--- a/src/NHibernate.Test/Bytecode/WrongProxyFactoryFactory.cs
+++ b/src/NHibernate.Test/Bytecode/WrongProxyFactoryFactory.cs
@@ -33,6 +33,8 @@ namespace NHibernate.Test.Bytecode
             return entity is INHibernateProxy;
         }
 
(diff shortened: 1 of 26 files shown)


hazzik:
I don't see how it avoids double casting at this moment. I can clearly see how that would avoid this in 6.0, but now its just a mess. Does it worth it now? 

Honestly, I would prefer abolish "IsProxy" method in its entirety. It was added as a some workaround for Spring.NET. And Spring.NET is dead now.

bahusoid:
Ok. Closing for now

bahusoid:
>  I would prefer abolish "IsProxy" method in its entirety.

Or should I obsolete `IsProxy` method now and replace its usages?

hazzik:
No, it would be too soon. I'll create an issue.

PR #2100, opened 2019-04-03
Allow to override default types with length or precision parameters

bahusoid:
It's possible since #702 to override default type with custom type. But current solution is limited as it doesn't handle precision or length based types. 

So this PR aims to fix this limitation and it allows:
1) Transparently replace any default type with custom type (with exception for places where explicitly `NHibernateUtil.SomeType` is used like `IQuery` set param methods `SetString`, `SetDateTime`,...)
2) Simplify custom type mapping by reusing existing `length`, `precision` and `scale` mapping attributes.

Example of `string` override:
```C#
//Somewhere before SessionFactory is created
TypeFactory.RegisterType(
	typeof(string),
	new CustomStringType(),
	new[] {"string"},
	configurator => configurator.SetTypeByLengthOrScale(length => new CustomStringType(length)));
```


hazzik:
Why not just like this?
```csharp
//Somewhere before SessionFactory is created
TypeFactory.RegisterType(
	typeof(string),
	new CustomStringType(),
	new[] {"string"},
	length => new CustomStringType(length));
```

```csharp
//Somewhere before SessionFactory is created
TypeFactory.RegisterType(
	typeof(string),
	new CustomStringType(),
	new[] {"string"},
	(scale, precision) => new CustomStringType(scale, precision));
```

Why do we need `TypeConfigurator`?

bahusoid:
To avoid public methods multiplications. In 6.0 it's going to be single public method which looks cleaner to me. Also it simplifies further extension - you can add any number of members to `TypeConfigurator` without breaking changes and still have only single public method exposed.

Well... If you are not convinced can make two additional methods as you suggested.

hazzik:
>and still have only single public method exposed.

I don't mind having a method overload.

bahusoid:
> I don't mind having a method overload.

Overloads with delgates tend to collide so I prefer avoid such overloads if possible. For instance  `length => new CustomStringType(length)` now is nicely resolved by compiler to `GetNullableTypeWithLengthOrScale`. But if you want to add some similar delgate with different input parameter like `NullableTypeCreatorDelegate` - you have to use new method name as overload will be ambigious for lambda calls.

So are you sure with overloads?

src/NHibernate/Type/TypeFactory.cs
```diff
@@ -221,39 +293,47 @@ private static void RegisterDefaultNetTypes()
 		{
 			// NOTE: each .NET type should appear only one time
 			RegisterType(typeof (Byte[]), NHibernateUtil.Binary, new[] {"binary"},
-						 l => GetType(NHibernateUtil.Binary, l, len => new BinaryType(SqlTypeFactory.GetBinary(len))));
+						 l => GetType(NHibernateUtil.Binary, l, len => new BinaryType(SqlTypeFactory.GetBinary(len))),
+						 true);
```
fredericDelaporte: Why `true` here ? A typo?

hazzik: Yes

PR #2101, opened 2019-04-03
Allow columns index be shared between base and subclass properties

### LINKED ISSUE #2099: "Composite Index" not working with inheritance
We want to create an index of multiple columns in our class. This seems not work, when inheritance is involved (in NH 5.2.1). 

The base class:
```
<?xml version="1.0" encoding="utf-8" ?>
<hibernate-mapping xmlns="urn:nhibernate-mapping-2.2"
                   auto-import="false"
                   assembly="RADFoundation, Version=2.1.0.0, Culture=neutral, PublicKeyToken=null"
                   namespace="RADFoundation.ProblemDomain">
  <class name="PersistentObject" table="RAD_persistentobject" abstract="true" dynamic-update="true" lazy="true">
	<property name="PDO_Deleted" column="PDO_Deleted" index="INDEX_PDO_DELETED, CF_BASE_COMPOSITE_IDX" access="backfield"/>
  </class>
</hibernate-mapping>
```

The other class:
```
<?xml version="1.0" encoding="utf-8" ?>
<hibernate-mapping xmlns="urn:nhibernate-mapping-2.2"
                   auto-import="false"
                   assembly="WorkflowEngine, Version=2.1.0.0, Culture=neutral, PublicKeyToken=null"
                   namespace="WorkflowEngine.Model">
  <union-subclass name="WorkflowInstance" table="WFE_workflowinstance" abstract="false" dynamic-update="true" lazy="true" extends="RADFoundation.ProblemDomain.PersistentObject">
	<property name="IsActive" column="IsActive" index="CF_BASE_COMPOSITE_IDX" access="nosetter.camelcase-underscore"/>
	<property name="IsWaiting" column="IsWaiting" index="CF_BASE_COMPOSITE_IDX" access="nosetter.camelcase-underscore"/>
  </union-subclass>
</hibernate-mapping>
```

The table WFE_workflowinstance contains the index **CF_BASE_COMPOSITE_IDX** and INDEX_PDO_DELETED as well, but **CF_BASE_COMPOSITE_IDX**  only contains the columns **IsActive** and **IsWaiting** and not the inheritated column **PDO_Deleted**.

bahusoid:
Fixes #2099

### DIFF
--- a/src/NHibernate/Mapping/DenormalizedTable.cs
+++ b/src/NHibernate/Mapping/DenormalizedTable.cs
@@ -43,6 +43,12 @@ namespace NHibernate.Mapping
 				IEnumerable<Index> includedIdxs = includedTable.IndexIterator;
 				foreach (Index parentIndex in includedIdxs)
 				{
+					var sharedIndex = GetIndex(parentIndex.Name);
+					if (sharedIndex != null)
+					{
+						sharedIndex.AddColumns(parentIndex.ColumnIterator);
+						continue;
+					}
 					Index index = new Index();
 					index.Name = Name + parentIndex.Name;
 					index.Table = this;
--- a/src/NHibernate/Mapping/DenormalizedTable.cs
+++ b/src/NHibernate/Mapping/DenormalizedTable.cs
@@ -43,17 +43,16 @@ namespace NHibernate.Mapping
 				IEnumerable<Index> includedIdxs = includedTable.IndexIterator;
 				foreach (Index parentIndex in includedIdxs)
 				{
-					//TODO: Change index name only for DB that require unique index name (like PostgreSQL)
-					var newName = Name + parentIndex.Name;
 					var sharedIndex = GetIndex(parentIndex.Name);
 					if (sharedIndex != null)
 					{
 						sharedIndex.AddColumns(parentIndex.ColumnIterator);
-						sharedIndex.Name = newName;
+						sharedIndex.IsInherited = true;
 						continue;
 					}
 					Index index = new Index();
-					index.Name = newName;
+					index.Name = parentIndex.Name;
+					index.IsInherited = true;
 					index.Table = this;
 					index.AddColumns(parentIndex.ColumnIterator);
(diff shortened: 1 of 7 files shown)


src/NHibernate/Mapping/DenormalizedTable.cs
```diff
@@ -43,6 +43,12 @@ public override IEnumerable<Index> IndexIterator
...
 				foreach (Index parentIndex in includedIdxs)
 				{
+					var sharedIndex = GetIndex(parentIndex.Name);
+					if (sharedIndex != null)
+					{
+						sharedIndex.AddColumns(parentIndex.ColumnIterator);
+						continue;
+					}
 					Index index = new Index();
 					index.Name = Name + parentIndex.Name;
```
bahusoid: With this fix it seems  this change is not required and we can keep existing index name intact. Should we?

bahusoid: Ok. I see why this change is needed. Some of DBs like PostgreSQL need unique index name...

src/NHibernate/Mapping/Index.cs
```diff
@@ -68,7 +68,8 @@ public static string BuildSqlDropIndexString(Dialect.Dialect dialect, Table tabl
 		/// </returns>
 		public string SqlCreateString(Dialect.Dialect dialect, IMapping p, string defaultCatalog, string defaultSchema)
 		{
-			return BuildSqlCreateIndexString(dialect, Name, Table, ColumnIterator, false, defaultCatalog, defaultSchema);
+			var indexName = IsInherited ? Table.Name + Name : Name;
```
bahusoid: Possible improvement to avoid changing index name for dialects where same index name is allowed for different tables.
According to my tests dialects that require unique index name in database:
Firebird
Oracle
SQLite
PostgreSQL

Support same index name:
SQL Server
SQL Server CE
MySql
SAP SQL Anywhere

ValResnick:
Can i do something here to close this PR ?

PR #2103, opened 2019-04-05
WIP Cascade entities loaded during flush

### LINKED ISSUE #1252: NH-2919 - Cascade update doesn't work for lazy loaded entities that are loaded by the flush
**Gen Cuevas** created an [issue](https://nhibernate.jira.com/browse/NH-2919) — <time datetime="2011-10-20T14:16:23.103Z" title="2011-10-20T14:16:23.103Z">20th October 2011, 14:16:23</time>:
> NHibernate appears to track entities clearly in the session as they are worked on and accessed.  As expected and desired, lazy loaded relations won't be brought into memory or acted upon during interactions with the session if they are never accessed, and hence are still NOT brought into memory during a SaveUpdate/Flush. HOWEVER, if the Flush of an entity that is already in the session causes the realization of a lazy loaded relation, that referenced entity is loaded by NHibernate and will have any changes to its properties persisted.  BUT, if it also as cascade relations, those are not followed because it is too late in the tracking process and all the cascading has already happened.  
> 
> We have decided to file this as a bug as we feel that this behavior of cascading through lazy loaded entities that are loaded in the flush, though an edge case, is behavior that we expected.  We were torn between filing this as a either a major or minor bug, but we decided to file this as a "major" as the only workaround we have at this point is to fully realize the whole domain model, in which case it kind of defeats the purpose of lazy loading.
> 
> I've attached a sample WPF standalone application that demonstrates this behavior.
> 
> In the project, I've created the following objects:
> - Parent: contains a list of persisted Child objects and a persisted ParentSummary object.  Both the Children list and the ParentSummary object have its cascade set to "all".
> - Child: contains a persisted Toy object called DynamicToy.  The Toy is a readonly property (i.e. it's nhibernate mapped property has access="readonly", and is generated and assigned in the Child class's constructor.  Thus, every time a new instance of Child is created, a new instance of a Toy is created and assigned to the child. This DynamicToy property has its cascade set to "all".
> - ParentSummary: contains a reference to its Parent, and a persisted readonly property (with a no-op setter), FirstChild, that is it's Parent's first child.  This FirstChild property has its cascade set to "all".
> - Toy: a simple object with an ID and Name.  It contains no references to other entities.
> 
> Note: This domain model may be a little contrived, but it stems from an edge case where, initially, the Child didn't have that toy property, but was added later, in which case, we wanted to ensure that if it already didn't have the Toy, then we would dynamically create it.
> 
> The data setup in the application (that is created when "Create Data" button is clicked) will setup the following objects:
> - Creates a Parent
> - Creates a Child, and adds it to the Parent's Children list.  A Toy is created as a result of the Child creation, and is assigned to the child's DynamicToy property.
> - Creates a ParentSummary for the Parent.
> 
> Once the data is setup, the following tests can then be run (via the "Run Test" button).  
> 			// TEST 1:
> 			// Will only load the parent - will only persist parent modifications
> 			using ( var session = _sessionFactory.OpenSession() ) {
> 				var parent = session.Get<Parent>( _loadedParentID );
> 
> 				session.Flush();
> 			}
> 
> 			// TEST 2:
> 			// Will only load the parent.  (Just getting the identifier will not load the entity)
> 			// Will only persist parent modifications.
> 			using ( var session = _sessionFactory.OpenSession() ) {
> 				var parent = session.Get<Parent>( _loadedParentID );
> 				parent.Name = "Changing Name";
> 				var id = parent.Summary.ID;
> 
> 				session.Flush();
> 			}
> 
> 			// TEST 3:
> 			// Will load the parent and the summary.  As a result, upon persistence, it will load 
> 			// the Children because of a property in summary that will force the Children to load.
> 			// However, it will not cascade through the Child object's properties.
> 			using ( var session = _sessionFactory.OpenSession() ) {
> 				var parent = session.Get<Parent>( _loadedParentID );
> 				parent.Summary.Parent = parent;
> 
> 				session.Flush();
> 			}
> 
> 			// TEST 4:
> 			// Will also load the parent and the summary.  As a result, upon persistence, it will load
> 			// the Children because of a property in summary that will force the Children to load.
> 			// However, it will not cascade through Child object's properties.
> 			using ( var session = _sessionFactory.OpenSession() ) {
> 				var summary = session.Get<ParentSummary>( _loadedSummaryID );
> 				var parent = session.Get<Parent>( _loadedParentID );
> 
> 				summary.Parent = parent;
> 
> 				session.Flush();
> 			}
> 
> When the option to "Instantiate child with a toy if none already exists when running the test" is not checked, the the tests above run fine.  However, if checked, then Test 3 & Test 4 both fail with a TransientObjectException, because the DynamicToy is created too late in the game (i.e. in the AbstractFlushingEventListener.PrepareEntityFlushes when cascading through the properties of ParentSummary), and thus is not made "persistent" by the cascade.
>

bahusoid:
This change allows to properly flush and cascade entities that were loaded and modified during flush (I know this is not recommended way of doing things but sometime it can be useful)

Fixes #1252


hazzik:
This is extremely weird use case. I spent too much time trying to understand what's going on.

bahusoid:
Yeah. I agree test case from #1252 is very weird. But in general this PR covers condition when entity is modified on loading event and this loading happens during flush for whatever reasons (non lazy association or something like that)

In test case it's not even loading it's proxy creation.

hazzik:
This is what it takes to make the test green (90ea46768ce5a3748005a64dc6b97f5f3c89f643):

https://github.com/nhibernate/nhibernate-core/blob/90ea46768ce5a3748005a64dc6b97f5f3c89f643/src/NHibernate.Test/NHSpecificTest/NH2919/Child.cs#L9-L13

hazzik:
The culprit is `access="readonly"`. This so called feature should have been never implemented.

hazzik:
The test becomes green when either: 

- DynamicToy is set using the property
- DynamicToy does not have `access="readonly"`

bahusoid:
>The test becomes green when either:
>DynamicToy is set using the property

That's because of stupid logic inside property setter. Yeah test is strange and I should have come up with something more simpler and clear instead of reusing it. But again the issue should be clear: entity is loaded and modified during flush. Cascade changes won't be applied for such entities. If you have some complex checks in interceptor - it's quite possible scenario (I've personally investigated it in some legacy app).

I will try to add some simple test demonstrating this behavior as I agree current test is hard to understand

bahusoid:
Actually I can't come up with anything that looks sane :) Maybe such scenarios are indeed should be considered too specific to be handled in nhibernate. So feel free to close this PR...

hazzik:
>DynamicToy does not have access="readonly"

This still holds.

hazzik:
These tests need a sensible asserts at least. They currently do not check anything (except implicit check for `TransientObjectException`).

PR #2104, opened 2019-04-05
Fix exception on session refresh for removed collection item

### LINKED ISSUE #1738: Refresh of locally removed collection item crashes with "instance was not in a valid state"
When removing an item from a mapped collection in one ISession, and refreshing the collection state in another session NHibernate crashes with "instance was not in a valid state".

Crash was reproduced in NHibernate 4.0.x
Code snippet to reproduce the crash:
```
	DataItemWithSubItems item;
	using (ISession session = sessionFactory.OpenSession()) {
		 item = session.Get<DataItemWithSubItems>(1);

		Assert.That(item.SubItems.Count, Is.EqualTo(3));
		Assert.That(item.SubItems[0].Name, Is.EqualTo("Name 101"));
		Assert.That(item.SubItems[1].Name, Is.EqualTo("Name 102"));
		Assert.That(item.SubItems[2].Name, Is.EqualTo("Name 103"));

		item.SubItems.RemoveAt(1);
		Assert.That(item.SubItems.Count, Is.EqualTo(2));
	}

	using (ISession session = sessionFactory.OpenSession()) {
		session.Update(item);

		session.Refresh(item);
		foreach (var rawItem in item.SubItems) {
			session.Refresh(rawItem); // <-- it crashes here.
		}

		Assert.That(item.SubItems.Count, Is.EqualTo(3));
		Assert.That(item.SubItems[0].Name, Is.EqualTo("Name 101"));
		Assert.That(item.SubItems[1].Name, Is.EqualTo("Name 102"));
		Assert.That(item.SubItems[2].Name, Is.EqualTo("Name 103"));
	}
```

Stacktrace:
```
NHibernate.HibernateException : instance was not in a valid state
   bei NHibernate.Engine.EntityEntry.get_IsReadOnly() in d:\Projects\NHibernate\nhibernate-core\src\NHibernate\Engine\EntityEntry.cs:Zeile 293.
   bei NHibernate.Event.Default.DefaultRefreshEventListener.OnRefresh(RefreshEvent event, IDictionary refreshedAlready) in d:\Projects\NHibernate\nhibernate-core\src\NHibernate\Event\Default\DefaultRefreshEventListener.cs:Zeile 115.
   bei NHibernate.Impl.SessionImpl.FireRefresh(RefreshEvent refreshEvent) in d:\Projects\NHibernate\nhibernate-core\src\NHibernate\Impl\SessionImpl.cs:Zeile 2492.
   bei NHibernate.Impl.SessionImpl.Refresh(Object obj) in d:\Projects\NHibernate\nhibernate-core\src\NHibernate\Impl\SessionImpl.cs:Zeile 1379.
   [User Code]
```

Workaround: use own IRefreshEventListener that does not evaluate EntityEntry.IsReadOnly for the locally deleted item.

mapped classes:
```
	public class DataItemWithSubItems {
		public int SID { get; private set; }
		public string Name { get; set; }
		public IList<SubItem> SubItems { get; private set; }
	}
	public class SubItem {
		public int SID { get; private set; }
		public string Name { get; set; }

		public DataItemWithSubItems Parent { get; private set; }
	}
```
Mapping files:
```
  <class name="NHibernateRefreshEventListener.DataItemWithSubItems, NHibernateRefreshEventListener" table="MAIN_TABLE" dynamic-update="true" lazy="false">
    <id name="SID" type="Int32" column="SID" unsaved-value="-1">
      <generator class="hilo" />
    </id>
    <property name="Name" column="NAME" length="50" not-null="false"/>

    <bag
      name="SubItems" inverse="true" lazy="false" cascade="all-delete-orphan">
      <key column="PARENT_SID"/>
      <one-to-many class="NHibernateRefreshEventListener.SubItem, NHibernateRefreshEventListener" />
    </bag>
  </class>
  
    <class name="NHibernateRefreshEventListener.SubItem, NHibernateRefreshEventListener" table="SUB_TABLE" dynamic-update="true" lazy="false">
    <id name="SID" type="Int32" column="SID" unsaved-value="-1">
      <generator class="hilo" />
    </id>
    <property name="Name" column="NAME" length="50" not-null="false"/>
    <many-to-one name="Parent" class="NHibernateRefreshEventListener.DataItemWithSubItems, NHibernateRefreshEventListener" column="PARENT_SID" />
  </class>

```

bahusoid:
Fixes #1738


bahusoid:
Actually I dislike this exception in `IsReadonly` property. It's quite unexpected to me. I got to thinking that maybe better fix would be to move exception to `SetReadOnly`. So the following code:
https://github.com/nhibernate/nhibernate-core/blob/6cb478fd7644507941140dd05b33e2fa9f66702e/src/NHibernate/Engine/EntityEntry.cs#L306-L326

should be changed to:
```C#
public bool IsModifiableEntity()
{
	return  !IsReadOnly && Persister.IsMutable;
}

public bool IsReadOnly => (status == Status.ReadOnly) || (status == Status.Deleted && previousStatus == Status.ReadOnly);

public void SetReadOnly(bool readOnly, object entity)
{
	if (status != Status.Loaded && status != Status.ReadOnly)
	{
		throw new HibernateException("instance was not in a valid state");
	}
 	if (readOnly == IsReadOnly) 
 		return; // simply return since the status is not being changed 
...
```

PR #2108, opened 2019-04-08
Multi-Tenancy: Implement tenant per Database strategy

### LINKED ISSUE #974: NH-3087 - Add support for multi tenancy
**Paul White** created an [issue](https://nhibernate.jira.com/browse/NH-3087) — <time datetime="2012-03-13T21:19:19.832Z" title="2012-03-13T21:19:19.832Z">13th March 2012, 21:19:19</time>:
> Hibernate now supports multi tenancy, documented here: http://docs.jboss.org/hibernate/core/4.1/devguide/en-US/html/ch16.html#d5e4642
> 
> The feature was tracked in Jira here: https://hibernate.atlassian.net/browse/HHH-5697
> Related commits [here](https://github.com/hibernate/hibernate-orm/search?q=HHH-5697&type=Commits)
> 
> Hopefully something similar can be implemented for NHibernate
> 
---
**Paul White** added a comment — <time datetime="2012-03-14T15:03:24.315Z" title="2012-03-14T15:03:24.315Z">14th March 2012, 15:03:24</time>:
> I have forked NHibernate Core to start work on this: https://github.com/IronsUK/nhibernate-core
> 
---
**Paul White** added a comment — <time datetime="2012-04-17T08:07:57.405Z" title="2012-04-17T08:07:57.405Z">17th April 2012, 8:07:57</time>:
> I created a pull request for some initial work around this: https://github.com/nhibernate/nhibernate-core/pull/91
> 
---
**Oskar Berggren** added a comment — <time datetime="2012-05-26T20:26:10.107Z" title="2012-05-26T20:26:10.107Z">26th May 2012, 20:26:10</time>:
> Pull request 91 partially merged. The remained cleaned up and available at https://github.com/oskarb/nhibernate-core/tree/NH-3087-prelim.
>

bahusoid:
Fixes #974

Based on hibernate implementation (related [commits](https://github.com/hibernate/hibernate-orm/search?q=HHH-5697&type=Commits)) with some modifications (omitted unused classes, interface members and modified tenant session configuration)

Example of tenant configuration:
1) Add `multi_tenancy.strategy` and `multi_tenancy.connection_provider`  NHibernate configuration properties (`hibernate-configuration\session-factory`):
```xml
<property name="multi_tenancy.strategy">Database</property>
<property name="multi_tenancy.connection_provider>AssemblyQualifiedNameForMyMultiTenancyConnectionProvider</property>
```

Or example with by code configuration:
```C#
Configuration configuration = GetNHConfiguration();
configuration.DataBaseIntegration(
	x =>
	{
		x.MultiTenancy = MultiTenancyStrategy.Database;
		x.MultiTenancyConnectionProvider<MyMultiTenancyConnectionProvider>();
	});
```

Where `MyMultiTenancyConnectionProvider` must implement `IMultiTenancyConnectionProvider`. Can be implemented based on provided `AbstractMultiTenancyConnectionProvider`. Example of implementation:
```C#
public class MyMultiTenancyConnectionProvider : AbstractMultiTenancyConnectionProvider
{
	private Dictionary<string, string> _dic = new Dictionary<string, string>
	{
		{"tenant1", "tenantConnectionString1"},
		{"tenant2", "tenantConnectionString2"},
	};

	protected override string GetTenantConnectionString(TenantConfiguration configuration)
	{
		return _dic[configuration.TenantIdentifier];
		//Or something like:
		//return ConfigurationProvider.Current.GetNamedConnectionString("tenantConnectionFor." + configuration.TenantIdentifier);
	}
}
```
2) Provide tenant configuration each time you open session. Something like:
```C#
sessionFactory.WithOptions()
.Tenant(new TenantConfiguration(TenantName))
.OpenSession()
```

You can create subclass of `TenantConfiguration` and provide some specific details about you tenant configuration and use it to obtain proper tenant connection inside your `IMultiTenancyConnectionProvider` implementation

Current tests are not exactly multi-tenant as executed on the same database. For proper multi-tenant tests we need additional db which I'm not sure how to properly add (and I think should be done as separate PR). Still current tests make sure that different connection string is used for different tenants (by injecting AppName  on SQL Server) and make sure that second-level/query cache is separated for different tenants. Let me know if I missed something to [test](https://github.com/bahusoid/nhibernate-core/blob/multiTenancy/src/NHibernate.Test/MultiTenancy/DatabaseStrategyNoDbSpecificFixture.cs).

Things to implement in future:
* Ability to clear tenant specific caches (maybe option to prefix cache region with tenantIdentifier?) (https://github.com/nhibernate/nhibernate-core/pull/2133)


src/NHibernate.Test/DebugSessionFactory.cs
```diff
@@ -400,7 +401,7 @@ public static ISessionCreationOptions GetCreationOptions(IStatelessSessionBuilde
 				(ISessionCreationOptions)sessionBuilder;
 		}
 
-		internal class SessionBuilder : ISessionBuilder
+		internal class SessionBuilder : ISessionBuilder, ISessionCreationOptionsWithMultiTenancy
```
hazzik: Seems like `ISessionBuilder` needs to be an abstract/virtual class. Can you please add a TODO for 6.0?

bahusoid: I don't get it. Where TODO should be added? This is test class wrapper. And `ISessionCreationOptionsWithMultiTenancy`  is a temp interface for supporting such custom interface implementations and it will be dropped in 6.0. My understanding no abstract class is required.

src/NHibernate/Cfg/Loquacious/IDbIntegrationConfigurationProperties.cs
```diff
@@ -41,5 +41,8 @@ public interface IDbIntegrationConfigurationProperties
 		SchemaAutoAction SchemaAction { set; }
 
 		void QueryModelRewriterFactory<TFactory>() where TFactory : IQueryModelRewriterFactory;
+
+		//TODO 6.0: Uncomment. Or even better consider dropping all configuration interfaces and use classes directly 
+		//MultiTenancy.MultiTenancyStrategy MultiTenancy { set; }
```
bahusoid: Is this even the right place to add? I'm a bit lost in those configuration classes.

maca88:
> tenantIdentifier is not added to EntityKey. It's not used by equality members (as session is scoped to single tenant) - so not sure why it was added

They removed it later with [HHH-8682](https://github.com/hibernate/hibernate-orm/commit/220a27af64cd1f7b2b6c3e12c670e5a1f8e49ec2).

bahusoid:
@maca88  Indeed. I've removed this part from description.

bahusoid:
Not sure if I added `MutiTenancy` "Loquacious" configuration to proper [place](https://github.com/nhibernate/nhibernate-core/blob/7b20d20e8472b3b563acddbcb73e890ce0ad1b21/src/NHibernate/Cfg/Loquacious/DbIntegrationConfigurationProperties.cs) but everything else seems to be ready.

src/NHibernate.Test/DebugSessionFactory.cs
hazzik: Nevermind

gliljas:
Sorry for being late here, but I personally think that the Hibernate structure is much cleaner. However, this version can certainly be wrapped away into extension methods or app unique "session providers", and made to look like the Hibernate version.

That's not saying that this is bad or that the Hibernate implementation is good. Like a lot of the Hibernate code it's quite bad.

What it boils down to is that NHibernate's extensibility should allow for something like this, without requiring very specific code changes. Injecting strategies, providers and decorators should do the trick.

bahusoid:
Could you be more specific. As it's unclear to me which part and why would you want to wrap in hibernate version. So could you describe some scenario that cleaner in hibernate?

gliljas:
Gladly. 

I just think that bringing things like a ConnectionProvider up to the topmost API feels a bit awkward. Multi tenancy is to me a more infrastructural concern. Just "tagging" the session, or even better, use the (I)CurrentTenantIdentifierResolver method that Hibernate does, would be more flexible, I think. It would also allow for different kinds of tenancy strategies, which may not necessarily use different connection providers (just as the Hibernate implementation does).

My objection really isn't with your implementation. It rather stems from a frustration with the old codebase in the configuration and execution pipeline, which is such a feast of bad practices (all inherited from Hibernate). With a properly extensible API, adding something like multi tenancy wouldn't require core changes. A similar extension, NHibernate.Shards, would also be able to inject itself with ease.

Maybe something for 7.0

bahusoid:
> I just think that bringing things like a ConnectionProvider up to the topmost API feels a bit awkward.

`MyTenantConnectionProvider` from example consists of two things: `tenantIdentifier` and `tenantConnectionString`. So I don't see it as something heavy and akward. It's essentialy `Database` specific configuration and you don't need to provide it for other tenancy strategies.

> Just "tagging" the session, or even better, use the (I)CurrentTenantIdentifierResolver method that Hibernate does, would be more flexible,

I don't see it that way. It just means that you split your tenant specific logic across different classes. And it means user needs to configure NHibernate in more places and implement more Nhibernate specific interfaces. That doesn't look cleaner to me. Now all that can be implemented in user code using convinient for user data models and be wrapped in whatever way he likes.

> It would also allow for different kinds of tenancy strategies, which may not necessarily use different connection providers 

As I alreayd said you don't need to provide tenant connection provider for other tenancy strategies - just add different ctor to `TenantConfiguration`.  But IMHO all other strategies should still support  different connection providers (for instance you want to store 100 small clients per database - means you need both schema/discriminator + database strategies).

gliljas:
> `MyTenantConnectionProvider` from example consists of two things: `tenantIdentifier` and `tenantConnectionString`. So I don't see it as something heavy and akward. It's essentialy `Database` specific configuration and you don't need to provide it for other tenancy strategies.

"Weight" is not the issue. It's placement of responsibilities. It _can_ be wrapped away. I just think that it should be by default.
 
> I don't see it that way. It just means that you split your tenant specific logic across different classes. And it means user needs to configure NHibernate in more places and implement more Nhibernate specific interfaces. That doesn't look cleaner to me. Now all that can be implemented in user code using convinient for user data models and be wrapped in whatever way he likes.
> 
OK, we're different. To me, it's cleaner.

> > It would also allow for different kinds of tenancy strategies, which may not necessarily use different connection providers
> 
> As I already said you don't need to provide tenant connection provider for other tenancy strategies - just add different ctor to `TenantConfiguration`. But IMHO all other strategies should still support different connection providers (for instance you want to store 100 small tenants per database - means you need both schema/discriminator + database strategies).

It may need that, but IMHO adding ctors isn't a convenient extensibility model.   

It's clearly a matter of taste, but to me:

1. Selecting "who" should be very lightweight or preferably completely transparent.
2. Selecting "how" belongs in a different layer.

grahambunce:
A bit late to this, but isn’t this handled perfectly well by the NHibernate Shards project? We use that perfectly well already to implement multi-tenancy across databases. Wouldn’t it be better to merge that into the base NHibernate code base, or add additional effort to that project?

bahusoid:
>  isn’t this handled perfectly well by the NHibernate Shards project?

All you need for multi-tenancy is to create separate session factory per tenant. Using NHibernate.Shards for this is very strange decision. But well - use whatever makes you happy.

And this PR allows to share the same session factory for all tenants.

src/NHibernate.Test/MultiTenancy/DatabaseStrategyNoDbSpecificFixture.cs
```diff
@@ -0,0 +1,348 @@
...
+		}
+
+		[Test]
+		public void DifferentConnectionStringForDifferentTenants()
+		{
+			if (!IsSqlServerDialect)
+				Assert.Ignore("MSSqlServer specific test");
+
+			using(var session1 = OpenTenantSession("tenant1"))
+			using (var session2 = OpenTenantSession("tenant2"))
```
fredericDelaporte, line 55: Whitespace glitch

src/NHibernate.Test/MultiTenancy/DatabaseStrategyNoDbSpecificFixture.cs
```diff
@@ -0,0 +1,348 @@
...
+		}
+
+		[Test]
+		public void StatelessSessionDifferentConnectionStringForDifferentTenants()
+		{
+			if (!IsSqlServerDialect)
+				Assert.Ignore("MSSqlServer specific test");
+
+			using(var session1 = OpenTenantStatelessSession("tenant1"))
+			using (var session2 = OpenTenantStatelessSession("tenant2"))
```
fredericDelaporte, line 88: Same here.

src/NHibernate.Test/MultiTenancy/DatabaseStrategyNoDbSpecificFixture.cs
```diff
@@ -0,0 +1,348 @@
...
+			}
+
+			Assert.That(Sfi.Statistics.PrepareStatementCount, Is.EqualTo(0));
+			Assert.That(Sfi.Statistics.SecondLevelCacheHitCount, Is.EqualTo(1));
+		}
+
+		private static string GetTenantId(ISession deserializedSession)
+		{
+			return deserializedSession.GetSessionImplementation().GetTenantIdentifier();
+		}
```
fredericDelaporte, line 250:
```suggestion
		private static string GetTenantId(ISession session)
		{
			return session.GetSessionImplementation().GetTenantIdentifier();
		}
```

This is not specific to deserialized sessions, isn't it?

src/NHibernate.Test/MultiTenancy/DatabaseStrategyNoDbSpecificFixture.cs
```diff
@@ -0,0 +1,348 @@
...
+		
+		private static string GetTenantId(IStatelessSession deserializedSession)
+		{
+			return deserializedSession.GetSessionImplementation().GetTenantIdentifier();
+		}
```
fredericDelaporte, line 255:
```suggestion
		private static string GetTenantId(IStatelessSession session)
		{
			return session.GetSessionImplementation().GetTenantIdentifier();
		}
```

bahusoid:
>  I think we should only provide the tenant id string for opening a session for a tenant. 

I don't know anything less convenient than single string param for whole configuration. That's some 
 legacy code smell to me. Even hibernate provides interface now (but again just for tenant id resolution) 

> Supplying the whole tenant configuration at each session opening looks like a code smell to me.

It's tenantIdentifier + tenantConnectionString. And it's wrapped in class to simplify further extensions. 

> I would rather have to add tenant configurations to the session factory first, which would have a (maybe concurrent) dictionary of them, then start opening session by just specifying their tenant id.

Introducing yet another dictionary when it can be avoided - again makes no sense to me and looks like code smell to me too.

src/NHibernate.Test/MultiTenancy/DatabaseStrategyNoDbSpecificFixture.cs
bahusoid: fixed

fredericDelaporte:
>But my big reluctance on using string key is based on personal experience where such design would lead to some awkward code to "deconstruct" string for retrieving tenant configuration.

For me, building some string based convention for supplying settings in a string to be parsed is not to be done. I am speaking about nothing else than a tenant identifier, with each tenant to be configured at the session factory level.

I do not understand the need for supplying all the tenant settings at each session opening. It looks unpractical to me.

>Sometime it's not something that can be/should be populated beforehand - tenant configuration can be dynamic, obtained on request and there is no reasons to keep it after end of request/session polluting session factory

Maybe then we should have both possibilities.

bahusoid:
> I am speaking about nothing else than a tenant identifier, with each tenant to be configured at the session factory level.

So OK it's about simple scenario when you have small static list of tenant configurations. It's not very useful for scenarios when tenant configurations are dynamic. 

> building some string based convention for supplying settings in a string to be parsed is not to be done

And still unless some part of code is bound to be executed for given tenant - you need some code that constructs required `tenantIdentifier` for session opening. And it means you have all tenant settings in this context .

> I do not understand the need for supplying all the tenant settings at each session opening.  It looks unpractical to me.

It's extensible. You can easily wrap this code to be executed the way you like. Via [string identifier](https://github.com/bahusoid/nhibernate-core/blob/fcc79c2d2ee3ecf3db3ff013df9c31fec4193a3d/src/NHibernate.Test/MultiTenancy/DatabaseStrategyNoDbSpecificFixture.cs#L109) or via some custom user class. That's how it supposed to work. And why to add additional dictionary lookup in our code on each session opening if we can avoid it. It can easily be added by user if needed.

> Maybe then we should have both possibilities.

As I already said I think it's trivial task that can be done in user code. Why to add additional dictionary lookup on session opening when we can avoid it and allow user to implement .

fredericDelaporte:
I remain unconvinced. This "supply the whole configuration at each session opening" pattern allows to supply inconsistent configurations, like supplying the same tenant identifier with different connections.
Dynamically adding tenant configurations to the session factory is possible. Of course this implies using a concurrent dictionary in order to be safe, but this would forbid mistakes like inconsistent configurations, provided we do not allow overriding a tenant configuration.
The session opening could have some `GetOrAdd` logic about tenant for allowing easier dynamical tenant configuration.

@hazzik, @maca88, may we have your thoughts about this PR in its current state?

bahusoid:
> like supplying the same tenant identifier with different connections

It could be valid scenario - same tenant db but different connections AppName in SqlServer (where AppName might contain some useful context info).

bahusoid:
In case of small amount of tenant databases known beforehand there is no real need in multitenancy solution. Just create multiple session factories and live with it. Or as suggested [here ](https://github.com/nhibernate/nhibernate-core/pull/2108#issuecomment-537699187) - use some projects that does it for you.  I don't understand why are you insisting on this use case.

I believe multitenancy is really useful in  SaaS like scenarios where tenants configurations are stored/retrieved externally. And in this scenario storing such configurations in dictionary in session factory is not useful at all. It only complicates things and pollutes session factory.

maca88:
> when tenant configuration is composite (like tenantName, region, store, ...) user needs to implement some kind of string parsing for recovering all this data from tenantIdentifier

I do understand that having a string for a tenant is not ideal when having a composite configuration (name, region, etc), but to me supplying the `IMultiTenantConnectionProvider` for each session is quite strange and IMO it would be better to have it configurable the way `IConnectionProvider` is, by setting `connection.provider` in the configuration. In order to support composite configuration, we could change the tenant identifier to object (`ToString` override would be required if the identifier is not a string), so that the user can supply whatever they want and then handle it in the `IMultiTenantConnectionProvider`.

> multiTenantConnectionProvider.GetConnection(tenantIdentifier);

I still need to check how Hibernate did this, but I do prefer a `IMultiTenantConnectionProvider` that has the `tenantIdentifier` as a parameter instead.

bahusoid:
> but to me supplying the IMultiTenantConnectionProvider for each session is quite strange

@maca88  NHibernate allows to open session with user supplied connection. I think it's essential to allow the same functionality for tenants session - it should be possible to open tenant session with any supplied connection.

`IMultiTenantConnectionProvider` is basically stripped to this functionality. It contains tenant identifier and opens connection via `IConnectionAccess`. That's it. It's very different from Hibernate version where it provides full `IConnectionProvider`.

I don't really have time right now to think of possible refactorings  and don't know when will be able to work on it. But if you disagree with current state - feel free to suggest or even adapt this changes further. Just keep in mind my intention - make it usable for dynamic tenants configurations and allow to open tenant session with supplied connection.

gliljas:
I will have a new look at this, later today.

bahusoid:
I actually have some ideas how to make everyone happy (I hope). So please wait for few days - I believe I can do it with not so many changes so it can still be squeezed in 5.3

bahusoid:
I believe it's done - `IMultiConnectionProvider` is moved to session factory settings. And `TenantConfigration` is transformed into simple DTO with `tenantIdentifier` and optional `tenantConnectionString`. `TenantConfiguration` might be extended in user code  as subclass to provide additional tenant information. And tenant connection information is retrieved via `SessionFactory.Settings.MultiTenancyConnectionProvider.GetConnectionAccess(tenantConfiguration)` as requested.

Also added default `DefaultMultiTenancyConnectionProvider` which expects connection string be provided on each session opening (otherwise it throws exception suggesting to implement own provider)

It's still doesn't have built in ability to specify static tenant configuration - but again it's trivial task. And hibernate also doesn't provide such ability. Example of static configuration :
```C#
	public class StaticMultiTenantConfiguration : DefaultMultiTenancyConnectionProvider
	{
		private Dictionary<string, string> _dic = new Dictionary<string, string>
		{
			{"tenant1", "tenantConnectionString1"},
			{"tenant2", "tenantConnectionString2"},
		};

		protected override string GetTenantConnectionString(TenantConfiguration configuration)
		{
			return _dic[configuration.TenantIdentifier];
			//Or something like:
			//return ConfigurationManager.ConnectionStrings["tenantConnectionFor." + configuration.TenantIdentifier];
		}
	}
```

maca88:
I revisited the PR and made [some refactoring](https://github.com/maca88/nhibernate-core/commit/e5e49b4c5ccb15bfbb65b4c6edc2efb6a152b285) that I would propose:

- Removed `ConnectionString` proprety from `TenantConfiguration`. IMO we should provide only the tenant indentifier and let the user decide whether to extend `TenantConfiguration` to add `ConnectionString` or handle it in `IMultiTenancyConnectionProvider`. Also in case we will support `MultiTenancyStrategy.DISCRIMINATOR` in the future, the connection string will be the same, so I thnik it is better to not have it.
- Added null check for tenant indentifier in `TenantConfiguration` constructor. I think we should "fail fast" and not wait until trying to open the session.
- Removed `ConnectionString` proprety from `IConnectionAccess` interface.
- Removed `IConnectionProvider` parameter from `IConnectionAccess.(Get|Close)Connection` methods. As the connection provider is always from the sesion factory, I've added the session factory parameter for `IMultiTenancyConnectionProvider.GetConnectionAccess` method instead.
- Removed `DefaultMultiTenancyConnectionProvider` and added `AbstractMultiTenancyConnectionProvider` instead. As mentined previously, let the user decide how to implement the connection string logic.
- Renamed configuration setting `multiTenancy` to `multi_tenancy.strategy` and `multiTenancy.connection_provider` to `multi_tenancy.connection_provider`. I think we should stick with the existing naming convention, rather than coping them from Hibernate.

bahusoid:
> Removed ConnectionString proprety from TenantConfiguration.

IMO it's useful to have it even for Discriminator case and it simplifies multi-tenant setup. But fine by me if others members agree it's better not to have it. Also then it seems it makes sense to add `Sfi.WithOptions().Tenant("tenantId")` shortcut (and maybe rename existing method too)

> Removed ConnectionString proprety from IConnectionAccess interface.

Why? It was added explicitly for convenience. It's been bugging me with `ConnectionProvider` that I can't check `ConnectionString` for it.

Agree with other changes.

bahusoid:
Actually scratch it - let's get rid of `TenantConfiguration.ConnectionString` - it can easily be added in user code. But please return `IConnectionAccess.ConnectionString` - it's not needed on NHibernate side, but it's useful information that can be used in user code. And feel free to commit your changes here.

maca88:
Pushed all changes except for removing `ConnectionString` property from `IConnectionAccess` interface.

gliljas:
Ok, now I'm taking a look... :)

bahusoid:
> Ok, now I'm taking a look... :)

Yeah about time. I'm done polishing it :)

src/NHibernate/Connection/ConnectionProvider.cs
```diff
@@ -203,5 +204,11 @@ protected virtual void Dispose(bool isDisposing)
 		}
 
 		#endregion
+
+		//TODO 6.0: Make abstract
+		public virtual DbConnection GetConnection(string connectionString)
```
fredericDelaporte, line 209: It should be move up to where `public abstract DbConnection GetConnection();` is declared, and I think we should now provide a default implementation for this `public abstract DbConnection GetConnection();`:
```c#
		public virtual DbConnection GetConnection()
		{
			return GetConnection(ConnectionString);
		}
```

src/NHibernate/Connection/DriverConnectionProvider.cs
```diff
@@ -31,12 +31,17 @@ public override void CloseConnection(DbConnection conn)
 		/// If there is any problem creating or opening the <see cref="DbConnection"/>.
 		/// </exception>
 		public override DbConnection GetConnection()
```
fredericDelaporte, line 33: With the previous suggestion, this member could be removed.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1032,7 +1033,12 @@ private CacheKey GenerateCacheKeyForEvict(object id, IType type, string entityOr
 					.GenerateCacheKey(id, type, entityOrRoleName);
 			}
 
-			return new CacheKey(id, type, entityOrRoleName, this);
+			if (settings.MultiTenancyStrategy != MultiTenancyStrategy.None && tenantIdentifier == null)
```
fredericDelaporte, line 1036: Shouldn't it be `tenantIdentifier != null`?

src/NHibernate/MultiTenancy/MultiTenancyStrategy.cs
```diff
@@ -0,0 +1,29 @@
...
+
+//		/// <summary>
+//		/// Multi-tenancy implemented by use of discriminator columns.
+//		/// </summary>
+//		Discriminator,
+//
+//		/// <summary>
+//		/// Multi-tenancy implemented as separate schemas.
+//		/// </summary>
+//		Schema,
```
fredericDelaporte, line 22: If you plan to introduce them later, it will shift the enumeration values, which will be a breaking change. So it is better to write down these values explicitly in order to freeze them.

src/NHibernate/Impl/SessionFactoryImpl.cs
bahusoid: No I meant that scenario calling `ISessionFactory.EvictEntity/EvictCollection` is not implemented at all for multi-tenancy. As I moved this part to my other PR https://github.com/nhibernate/nhibernate-core/pull/2133

But now I moved logic regarding deletion object from cache per ID to this PR (and other PR is about handling whole tenant cache). Error message is adjusted accordingly

src/NHibernate/Cache/CacheKey.cs
```diff
@@ -30,6 +31,20 @@ public class CacheKey : IDeserializationCallback
...
+			this.entityOrRoleName = entityOrRoleName;
+			_factory = factory;
+			_tenantIdentifier = tenantIdentifier;
+
+			_hashCode = GenerateHashCode();
+		}
+
+		//Since 5.3
+		[Obsolete("Use constructor with tenantIdentifier")]
 		public CacheKey(object id, IType type, string entityOrRoleName, ISessionFactoryImplementor factory)
```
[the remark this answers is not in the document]
bahusoid: Done

src/NHibernate/MultiTenancy/MultiTenancyStrategy.cs
bahusoid: I just moved comments to the end.

src/NHibernate/Connection/ConnectionProvider.cs
bahusoid: Done

bahusoid:
I believe I resolved all review suggestions and made last small change - exposed session `TenantConfiguration` as it might contain some useful context.

rodrigo-web-developer:
there is some example of this working? I'm trying to implement MultiTenancyConnectionProvider, but it seems to be more required configuration than just

```cs
config.DataBaseIntegration(x =>
            {
                x.MultiTenancy = MultiTenancyStrategy.Database;
                x.MultiTenancyConnectionProvider<MyMultiTenancyConnectionProvider>();
            });
```

It should be used? Or it is better to create a SessionFactory for each tenant, like some people said?

Issue #2109, opened 2019-04-09
Different GUID to string conversion when not explicitly referencing the id property


oskarb:
Tested on SQLite with NH 5.2.5.

With an entity where Id is Guid.

Consider these two queries:
```
List<string> guidDbStringsExplicitId =
    session.Query<Entity>().Select(e => e.Id.ToString()).ToList();

List<string> guidDbStringsImplicitId =
    session.Query<Entity>().Select(e => e.ToString()).ToList();
```

Both will aim to return the Id property, converted to string by the database engine, but the Id property is only explicitly referenced in one of them.

For guidDbStringsExplicitId, the generated SQL will be something like:
```
select
        substr(hex(entity0_.Id),
        7,
        2) || substr(hex(entity0_.Id),
        5,
        [...]
from
        Entity entity0_
```

While for guidDbStringsImplicitId is will be:
```
select
        cast(entity0_.Id as char) as col_0_0_ 
from
        Entity entity0_
```

The returned value is different. The expectation is that the SQL should be the same, or at least be functionally equivalent.

oskarb:
Related to PR #1856 and issue #1151.

oskarb:
PR #2111 contains a disabled test case that exposes this problem.

PR #2111, opened 2019-04-09
SQLiteDialect: Correct GUID to string conversion when BinaryGuid=False (regression) (fixes #2110)

### LINKED ISSUE #2110: SQLite GUID to string conversion wrong with BinaryGuid=False (regression)
Before NH 5.2, conversion from GUID to string (eg. entity.GuidProperty.ToString()) inside sqlite used a simple cast. This worked if BinaryGuid=False had been specified in the connection string.

Example SQL
```
    select
        cast(entity0_.Id as char) as col_0_0_ 
    from
        Entity entity0_
```

Following the fix for #1151 the conversion from GUID to string inside sqlite is now based on a sequence of  substr() and hex(). This gives the correct result if BinaryGuid=True (the default), but not if BinaryGuid=False. This is a regression in NH 5.2 (seems undocumented).

oskarb:
Attached are test cases for #2110 (and #2109), and a proposed fix for #2110.

The suggested fix teaches SQLiteDialect to detect the effective value of the BinaryGuid from the connection string and apply different expressions for the strguid function based on that. This fixes the first failure mentioned below.

In NH 5.2, without the fix, the result of the test cases are:

GH2110.ByCodeFixture(False).DbExplicitIdGuidToStringIsSameAsDotNet
Fails, because the expression given in the dialect for strguid is not valid with BinaryGuid=False. This is the regression caused by #1151. Fixed by this PR.

GH2110.ByCodeFixture(False).DbImplicitIdGuidToStringIsSameAsDotNet
Success, but only because NHibernate forgets to apply the strguid expression, opting for a straightforward `cast(entity0_.Id as char)` instead.

GH2110.ByCodeFixture(True).DbExplicitIdGuidToStringIsSameAsDotNet
Success. This is the case that was fixed by #1151.

GH2110.ByCodeFixture(True).DbImplicitIdGuidToStringIsSameAsDotNet
Fails (still), since the fix for #1151 is not applied here because NH doesn't apply the strguid method. Reported as #2109. Since it still fails, this case is ignored for now.

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3426/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3426/Fixture.cs
@@ -1,14 +1,57 @@
 ﻿using System;
 using System.Linq;
+using NHibernate.Cfg;
 using NHibernate.Cfg.MappingSchema;
+using NHibernate.Dialect;
 using NHibernate.Mapping.ByCode;
 using NUnit.Framework;
+using Environment = NHibernate.Cfg.Environment;
 
 namespace NHibernate.Test.NHSpecificTest.NH3426
 {
-	[TestFixture]
+	/// <summary>
+	/// Verify that we can convert a GUID column to a string in the standard GUID format inside
+	/// the database engine.
+	/// </summary>
+	[TestFixture(true)]
+	[TestFixture(false)]
 	public class Fixture : TestCaseMappingByCode
 	{
+		private readonly bool _useBinaryGuid;
+
+		public Fixture(bool useBinaryGuid)
+		{
+			_useBinaryGuid = useBinaryGuid;
+		}
+
+		protected override bool AppliesTo(Dialect.Dialect dialect)
+		{
+			// For SQLite, we run the tests for both storage modes (SQLite specific setting).
+			if (dialect is SQLiteDialect)
+				return true;
+
+			// For all other dialects, run the tests only once since the storage mode
+			// is not relevant. (We use the case of _useBinaryGuid==true since this is probably
+			// what most engines do internally.)
+			return _useBinaryGuid;
+		}
+
+		protected override void Configure(Configuration configuration)
+		{
+			base.Configure(configuration);
+
+			if (Dialect is SQLiteDialect)
+			{
+				var connStr = configuration.Properties[Environment.ConnectionString];
+
+				if (_useBinaryGuid)
+					connStr += "BinaryGuid=True;";
+				else
+					connStr += "BinaryGuid=False;";
+
+				configuration.Properties[Environment.ConnectionString] = connStr;
+			}
+		}
 
 		protected override HbmMapping GetMappings()
 		{
@@ -56,7 +99,7 @@ namespace NHibernate.Test.NHSpecificTest.NH3426
 					.Select(x => new { Id = x.Id.ToString() })
 					.ToList();
 
-				Assert.AreEqual(id.ToUpper(), list[0].Id.ToUpper());
+				Assert.That(list[0].Id.ToUpper(), Is.EqualTo(id.ToUpper()));
 			}
 		}
 		
@@ -94,6 +137,54 @@ namespace NHibernate.Test.NHSpecificTest.NH3426
 				var list = session.Query<Entity>()
 				                  .Where(x => ((Guid?) x.Id).ToString() == x.Id.ToString())
 				                  .ToList();
+
+				Assert.That(list, Has.Count.EqualTo(1));
+			}
+		}
+
+		[Test]
+		public void SelectGuidToStringImplicit()
+		{
+			if (Dialect is SQLiteDialect && _useBinaryGuid)
+				Assert.Ignore("Fails with BinaryGuid=True due to GH-2109. (2019-04-09).");
+
+			if (Dialect is FirebirdDialect || Dialect is MySQLDialect || Dialect is Oracle8iDialect)
+				Assert.Ignore("Since strguid() is not applied, it fails on Firebird, MySQL and Oracle " +
+							"because a simple cast cannot be used for GUID to string conversion on " +
+							"these dialects. See GH-2109.");
+
+			using (var session = OpenSession())
+			{
+				// Verify in-db GUID to string conversion when ToString() is applied to the entity that has
+				// a GUID id column (that is, we deliberately avoid mentioning the Id property). This
+				// exposes bug GH-2109.
+				var list = session.Query<Entity>()
+								.Select(x => new { Id = x.ToString() })
+								.ToList();
+
+				Assert.That(list[0].Id.ToUpper(), Is.EqualTo(id.ToUpper()));
+			}
+		}
+
+		[Test]
+		public void WhereGuidToStringImplicit()
+		{
+			if (Dialect is SQLiteDialect && _useBinaryGuid)
+				Assert.Ignore("Fails with BinaryGuid=True due to GH-2109. (2019-04-09).");
+
+			if (Dialect is FirebirdDialect || Dialect is MySQLDialect || Dialect is Oracle8iDialect)
+				Assert.Ignore("Since strguid() is not applied, it fails on Firebird, MySQL and Oracle " +
+							"because a simple cast cannot be used for GUID to string conversion on " +
+							"these dialects. See GH-2109.");
+
+			using (var session = OpenSession())
+			{
+				// Verify in-db GUID to string conversion when ToString() is applied to the entity that has
+				// a GUID id column (that is, we deliberately avoid mentioning the Id property). This
+				// exposes bug GH-2109.
+				var list = session.Query<Entity>()
+								.Where(x => x.ToString().ToUpper() == id)
+								.ToList();
 
 				Assert.That(list, Has.Count.EqualTo(1));
 			}
--- a/src/NHibernate/Dialect/SQLiteDialect.cs
+++ b/src/NHibernate/Dialect/SQLiteDialect.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
 using System.Text;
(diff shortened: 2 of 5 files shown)


hazzik:
Ok. This will go to 5.2.6.

fredericDelaporte:
Can you rebase this on 5.2.x branch?

oskarb:
Rebased, and PR retargeted.

I also figured it would be better to expand the existing test cases for the original issue NH-3426 instead of effectively duplicating.

src/NHibernate/Dialect/SQLiteDialect.cs
```diff
@@ -94,8 +102,34 @@ protected virtual void RegisterFunctions()
...
+			else
+				RegisterFunction("strguid", new SQLFunctionTemplate(NHibernateUtil.String, "cast(?1 as char)"));
+		}
+
+
+		public override void Configure(IDictionary<string, string> settings)
+		{
+			base.Configure(settings);
+
+			if (settings.TryGetValue(Cfg.Environment.ConnectionString, out string connectionString))
```
bahusoid: Shouldn't we also check for `Cfg.Environment.ConnectionStringName` as it's done in [`ConnectionProvider`](https://github.com/nhibernate/nhibernate-core/blob/27ea900b6ebaf5de5e9ca603ae57e6ca5197f968/src/NHibernate/Connection/ConnectionProvider.cs#L51-L62)?

I also wonder how to make it work with [multi-tenancy](https://github.com/nhibernate/nhibernate-core/pull/2108) where connection is provided externally... It seems default connection must be provided with proper configuration.

src/NHibernate.Test/NHSpecificTest/NH3426/Fixture.cs
```diff
@@ -1,14 +1,56 @@
...
+			return _useBinaryGuid;
+		}
+
+		protected override void Configure(Configuration configuration)
+		{
+			base.Configure(configuration);
+
+			if (Dialect is SQLiteDialect)
+			{
+				var connStr = configuration.Properties["connection.connection_string"];
```
bahusoid: Why not `Environment.ConnectionString`?

src/NHibernate/Dialect/SQLiteDialect.cs
oskarb: As for the first comment, it seems we should extract some code from ConnectionProvider to have something like Environment.ResolveConnectionString(...)

bahusoid: Maybe... Or just do copy/paste as you won't be able to reuse this logic in `ConnectionProvider` as `GetNamedConnectionString` can be overridden.

src/NHibernate/Dialect/SQLiteDialect.cs
```diff
@@ -94,8 +102,34 @@ protected virtual void RegisterFunctions()
...
-			RegisterFunction("strguid", new SQLFunctionTemplate(NHibernateUtil.String, "substr(hex(?1), 7, 2) || substr(hex(?1), 5, 2) || substr(hex(?1), 3, 2) || substr(hex(?1), 1, 2) || '-' || substr(hex(?1), 11, 2) || substr(hex(?1), 9, 2) || '-' || substr(hex(?1), 15, 2) || substr(hex(?1), 13, 2) || '-' || substr(hex(?1), 17, 4) || '-' || substr(hex(?1), 21) "));
+
+			if (_binaryGuid)
+				RegisterFunction("strguid", new SQLFunctionTemplate(NHibernateUtil.String, "substr(hex(?1), 7, 2) || substr(hex(?1), 5, 2) || substr(hex(?1), 3, 2) || substr(hex(?1), 1, 2) || '-' || substr(hex(?1), 11, 2) || substr(hex(?1), 9, 2) || '-' || substr(hex(?1), 15, 2) || substr(hex(?1), 13, 2) || '-' || substr(hex(?1), 17, 4) || '-' || substr(hex(?1), 21) "));
+			else
+				RegisterFunction("strguid", new SQLFunctionTemplate(NHibernateUtil.String, "cast(?1 as char)"));
+		}
+
+
+		public override void Configure(IDictionary<string, string> settings)
```
hazzik: This is driver specific settings, not dialect's

oskarb: I'm afraid I'm not sure exactly what you mean. What are you proposing?

oskarb:
It seems the test cases SelectGuidToStringImplicit and WhereGuidToStringImplicit that I added to showcase GH-2109 (affecting SQLite with BinaryGuid=true) also expose same problem on Firebird, Mysql and Oracle (any dialect where converting guid to string cannot be done using a regular cast in SQL).

src/NHibernate/Dialect/SQLiteDialect.cs
hazzik: What I'm saying is that BinaryGuid connection string option is specific to an ADO.NET provider implementation, not the dialect. For example, System.Data.SQLite and Mono.Data.Sqlite has this option, but Microsoft.Data.Sqlite does not. 

So, we need to handle this setting in a driver, not in a dialect.

src/NHibernate/Dialect/SQLiteDialect.cs
oskarb: It's a mess.

On a tangential note, Configuration.BuildSessionFactory() calls BuildSettings() which calls GetDerivedProperties(), which creates a temporary instance of the dialect to access any default properties set up by it, and combines those with the actual configuration. Then the whole thing gets sent to SettingsFactory.BuildSettings(), which, as the very first order of business, creates the real instance of the dialect _including_ reading its default properties and combining them with the actual configuration once more.

Anyway, back to ConnectionProvider. No idea why GetNamedConnectionString() is virtual. It is not overridden within NHibernate. You also can override both the top-level (public) Configure() and GetConnection(). One would think those options would be enough. I suspect someone just slapped virtual on everything. Would probably make more sense to combine reading both settings into a GetConnectionString() that might be virtual.

ConnectionProvider does only read the connection string once during Configure() and stores the effective value in its property `ConnectionString`. This is not available to the dialect because it doesn't have the ConnectionProvider and also the dialect is created earlier than the connection provider.

So, teaching the dialect about named connection strings is simpler but of course breaks down if anyone does in fact override GetNamedConnectionString() to get the configuration from somewhere else. And as @bahusoid says, it's still not enough for multitenancy.

I'll look into adding a dialect-specific setting to be used for scenarios where the connection string isn't enough.

src/NHibernate/Dialect/SQLiteDialect.cs
oskarb: Ah, that's an interesting point. One thing to keep in mind though is that the proposed configurability defaults to assume that BinaryGuid=true, which is the only case supported by Microsoft.Data.Sqlite.

So the proposed handling in the dialect should not be a problem at runtime, even though handling it in the driver might be considered more correct code wise. At the moment I'm not sure how to resolve this in the driver, since it still has to affect what the dialect does.

oskarb:
I've pushed an updated version with the following changes:

 * Also look at named connection string.

 * Introduce a dialect specific setting that can be used in cases where the automatic detection from connection string won't work.

 * Disabled the failing test cases for #2109 for more dialects that suffer from that problem.

src/NHibernate.Test/NHSpecificTest/NH3426/Fixture.cs
oskarb: Fixed.

hazzik:
Actually, I got into thinking... this applies not only to SQLite, but also to MySQL. The MySQL provider from Oracle has `OldGuids=true/false` connection string option. This kinda minor, so I would not bother with it now.

src/NHibernate/Dialect/SQLiteDialect.cs
fredericDelaporte: BinaryGuid connection string option is specific to the driver, but it still relates to how the database behaves about a specific type, which is a dialect business in my opinion.
That is just that some drivers do not support the non "binary guid" handling, so those drivers cannot be used with a database set to use non binary guid.
So I think this setting has to handled here in the dialect.

src/NHibernate/Dialect/SQLiteDialect.cs
```diff
@@ -94,8 +102,50 @@ protected virtual void RegisterFunctions()
...
+
+		private void ConfigureBinaryGuid(IDictionary<string, string> settings)
+		{
+			// We can use a SQLite specific setting to force it, but in common cases it
+			// should be detected automatically from the connection string below.
+			settings.TryGetValue(Cfg.Environment.SqliteBinaryGuid, out var strBinaryGuid);
+
+			if (string.IsNullOrWhiteSpace(strBinaryGuid))
+			{
+				string connectionString = Cfg.Environment.GetConfiguredConnectionString(settings);
```
fredericDelaporte: That part, detection of the setting from the connection string, is actually driver business. But it would require introducing a new driver interface (or wait for a 6.0 to allow adding a new member to the current one). Is it worth it?

fredericDelaporte:
So, to get this subject fixed, should this PR be changed for covering all the other databases case too? Or should they be handled in separated PRs?

The former should be preferred I think, but I have not checked if would expand a bit to much this PR.

src/NHibernate/Dialect/SQLiteDialect.cs
oskarb: I wonder if it is really worth creating and exposing additional interfaces for this, in order for the dialect to ask the driver (or be told) something which in practice can be handled as an internal matter for the dialect?

Also worth keeping in mind, this is intended to fix a regression.

oskarb:
This is intended to fix a regression I discovered, related to sqlite. Though I suppose there would be regressions for the other database where strguid was introduced/changed at the same time too.

I discovered this while working with unit tests that create the database on each run. Thinking about it, it's actually quite likely that I added BinaryGuid=false just to work around the problem with string conversions that has now been fixed in NH so maybe I don't need it anymore.

However, if someone have used BinaryGuid=false on a file-backed database that they can't easily just discard and recreate, they will be in more trouble.

fredericDelaporte:
I intend to merge this soon. I will likely prepare the 5.2.6 release the next week-end.

bahusoid:
@fredericDelaporte If you are going to merge changes to master please check [this comment](https://github.com/nhibernate/nhibernate-core/pull/2174#discussion_r293383930)

Issue #1402, opened 2017-10-24
Support Microsoft.Extensions.Configuration API


hazzik:

nickalbrecht:
Any progress on this? It still amazes me that there's a lot of code still hard coded to use `ConfigurationManager`

fredericDelaporte:
Feel free to grab it!

maulik-modi:
@fredericDelaporte , is it ready to be closed?

fredericDelaporte:
This is not done. But with 5.3 and #2116 there is a way to do it on user side.

maulik-modi:
> This is not done. But with 5.3 and #2116 there is a way to do it on user side.

@fredericDelaporte or @hazzik , Could you please highlight pending parts?

PR #2116, opened 2019-04-10
Ability to replace ConfigurationManager with a custom config provider

### LINKED ISSUE #2105: Xamarin .NET Standard 2.0.3 ConfigurationManager throws PlatformNotSupported exception in NHibernateLogger
i have a xamarin .net standard 2.0.3 project and try to use nhibernate with sqlite but get the error "System.PlatformNotSupportedException"

The error comes within this code from a .net standard 2.0.3 project:
```
    var cfg = new Configuration().DataBaseIntegration(db =>
      {
        db.Driver<SQLite20Driver>();
        db.Dialect<SQLiteDialect>();
        db.ConnectionString = "Data Source=test.db;Version=3;New=True";
      });
```
With just a (xunit) testproject and the nhibernate project this code works fine, but not if i want to start this code within an andorid system.

bahusoid:
Fixes #2105

Introduced abstract class [ConfigurationProvider](https://github.com/bahusoid/nhibernate-core/blob/configManager/src/NHibernate/Cfg/ConfigurationProvider.cs) exposed as `ConfigurationProvider.Current`.

To skip all configuration calls (fix for Xamarin) just set it to null:
```C#
ConfigurationProvider.Current = null;
```

Also implemented custom provider for tests under .NET Core (see [here](https://github.com/bahusoid/nhibernate-core/blob/configManager/src/NHibernate.Test/TestsContext.cs))


src/NHibernate.Test/TestsContext.cs
```diff
@@ -58,6 +49,35 @@ private static void ConfigureLog4Net()
...
 			hierarchy.Configured = true;
 		}
+
+		class NetCoreConfigurationManager : IConfigurationManager
+		{
+			private readonly System.Configuration.Configuration _configuration;
+
+			public NetCoreConfigurationManager()
+			{
+				var assemblyPath =
```
hazzik: I think this can become a parameter. And then the configuration manager moved to the main code base.

src/NHibernate/Cfg/IConfigurationManager.cs
```diff
@@ -0,0 +1,47 @@
...
+				return config.GetAppSetting(name);
+
+			var key = ConfigurationManager.AppSettings.Keys.Cast<string>().FirstOrDefault(k => name.Equals(k, StringComparison.OrdinalIgnoreCase));
+			return string.IsNullOrEmpty(key)
+				? null
+				: ConfigurationManager.AppSettings[key];
+		}
+	}
+
+	public interface IConfigurationManager
```
hazzik: I think it should be split into 2 or 3 different interfaces.

bahusoid: Why? It seems to me all or nothing is required as with our tests.

hazzik: Because the ConfigurationManager was an implementation detail, and abstracting it away one-to-one does not sound as a right move. It would make more sense if we have 3 distinct interfaces: connection string provider, logger provider, and configuration provider.

bahusoid: I disagree - this is abstraction for our exact "configuration from external source" needs. If user provides only one implementation - he can hit run-time `PlatformNotSupported` exception or something like that just because he didn't know that he must override three different interfaces at once.

User either uses some configuration file (app.config,xml,...) and then he must to override all current `IConfigurationManager` or don't need it at all  if he uses configuration by code.

I would think "why would I need to override Logger Provider" if I provide it explicitly by code? But in fact if "logger provider" is not overridden - runtime exception is thrown on Xamarin.

I was thinking to use `GetLoggerClass` instead  of `GetAppConfig` but decided to keep it just to be more flexible in future. But if you think we shouldn't provide general config access methods I can change it to `GetLoggerClass`. But I believe it should be single interface that covers "external configuration" scenario.

src/NHibernate.Test/TestsContext.cs
bahusoid: Done

src/NHibernate/Cfg/IConfigurationManager.cs
hazzik: I think the exception cases are orthogonal to the refactoring. I think we should deal with `ConfigurationErrorsException` as it is documented behaviour of `.AppSettings`/`.ConnectionStrings`/`GetSection`.

hazzik: These are 3 distinct cases. 

In case of logger we use an opportunistic approach (which I don't like) and it's OK if we just handle this exception. 

In case of connection string the user would not be able to use a named connection string and will need to use a provided connection string.

In case of configuration this is also configuration-over-default approach and we need to handle this exception.

bahusoid: I don't get how you want to deal with exception and where? You suggest to implement three interfaces and silently ignore this exception right inside of each default implementation.

So for instance with our tests  - if only one provider is overridden all others will be silently ignored? That doesn't sound right to me.

Again I don't consider this interface as different service providers - it's exactly configuration provider. Now we expect 3 things from configuration and so it's OK to have 3 members and not 3 different interfaces.

This splitting looks like unnecessary complication to me. Not just in our code but and for end user.

bahusoid: I've refactored code and got rid of App.config specific `GetAppConfig` method. Also instead of interface I introduced [abstract class](https://github.com/nhibernate/nhibernate-core/pull/2116/files#diff-db9ccbc60054326e75ad7f3a03212156R11-R20) (so new configuration settings can be added easily.)

So my main argument for this approach - all configuration related settings are present in one place and can be easily replaced by custom logic. IMHO that allows to fix both #1402 (not sure what you meant exactly there - but it's either already supported by using `SystemConfigurationManager` or can be easily implemented by end user) and #2105 (by making sure that no configuration calls are done)

hazzik:
>  not sure what you meant exactly there [#1402] - but it's either already supported by using SystemConfigurationManager or can be easily implemented by end user

There we need a way to implement a configuration from `IConfiguration` interface introduced in .NET Core 2.0

src/NHibernate/Cfg/ConfigurationManagerBase.cs
```diff
@@ -0,0 +1,109 @@
...
+using System.Configuration;
+using System.Linq;
+using NHibernate.Cfg.ConfigurationSchema;
+
+namespace NHibernate.Cfg
+{
+	/// <summary>
+	/// Base class for configuration settings that NHibernate might require.
+	/// </summary>
+	public abstract class ConfigurationManagerBase
```
hazzik: Lets rename this class to `ConfigurationProvider` and move each class into separate file.

src/NHibernate.Test/TestsContext.cs
```diff
@@ -15,7 +14,7 @@ namespace NHibernate.Test
 #endif
 	public class TestsContext
 	{
-		public static bool ExecutingWithVsTest { get; } =
+		private static bool ExecutingWithVsTest { get; } =
```
hazzik: Because this is now private this file applies only to .NET Core so it can be put into `#if NETCOREAPP2_0` as a whole.

src/NHibernate/Cfg/ConfigurationManagerBase.cs
bahusoid: Should all related classes/properties be renamed accordingly too?

src/NHibernate.Test/TestsContext.cs
bahusoid: Done

src/NHibernate/Cfg/ConfigurationManagerBase.cs
hazzik: Yes

src/NHibernate/Cfg/Settings.cs
```diff
@@ -18,6 +18,18 @@ namespace NHibernate.Cfg
 	/// </summary>
 	public sealed class Settings
 	{
+		private static ConfigurationProvider _configurationManager = new StaticSystemConfigurationManager();
+
+		/// <summary>
+		/// Provides ability to override default <see cref="System.Configuration.ConfigurationManager"/> with custom implementation.
+		/// Can be set to null if all configuration is specified by code 
+		/// </summary>
+		public static ConfigurationProvider ConfigurationManager
```
hazzik: Need to rename the field and property as well.

src/NHibernate/Connection/ConnectionProvider.cs
```diff
@@ -77,10 +76,9 @@ protected virtual string GetNamedConnectionString(IDictionary<string, string> se
 			if(!settings.TryGetValue(Environment.ConnectionStringName, out connStringName))
 				return null;
 
-			ConnectionStringSettings connectionStringSettings = ConfigurationManager.ConnectionStrings[connStringName];
-			if (connectionStringSettings == null)
-				throw new HibernateException(string.Format("Could not find named connection string {0}", connStringName));
-			return connectionStringSettings.ConnectionString;
+			var connectionString = Cfg.Settings.ConfigurationManager.GetNamedConnectionString(connStringName)
```
hazzik: There are some changes in 5.2.6 re here. Let's wait when it'll be merged first.

hazzik: Conflicted now.

src/NHibernate/Cfg/Settings.cs
```diff
@@ -18,6 +18,18 @@ namespace NHibernate.Cfg
 	/// </summary>
 	public sealed class Settings
 	{
+		private static ConfigurationProvider _configurationProvider = new StaticConfigurationManagerProvider();
+
+		/// <summary>
+		/// Provides ability to override default <see cref="System.Configuration.ConfigurationManager"/> with custom implementation.
+		/// Can be set to null if all configuration is specified by code 
+		/// </summary>
+		public static ConfigurationProvider ConfigurationProvider
```
hazzik: I do not like to have this property here. It seems only somehow related here.

However, I tried to put the property into Cfg.Environment or Cfg.SessionFactory classes as they seem to be more appropriate home for this property. Unfortunately, it did not work because these classes depend on logger and logger depends on the ConfigurationProvider. And so it creates a semi-circular dependency with undefined behaviour which resolves in TypeLoadException.

How about making this class fully independent by making a static property called `Current` in the `ConfigurationProvider` class itself? With the initial implementation (with interface) this was not really possible but without interface it is perfectly fine.

bahusoid: Ok

PR #2119, opened 2019-04-11
 Obsolete interfaces for Loquacios configuration and use config classes directly

bahusoid:
~Provide separate extension `ByCode` to avoid breaking changes and to improve discoverability for "by code" methods~ (made extensions as Configuration instance methods as was suggested)

IMHO there is no need in interfaces for such DTO like classes and they only clutter code and complicate adding new features.


src/NHibernate.Test/CfgTest/Loquacious/LambdaConfigurationFixture.cs
```diff
@@ -21,50 +19,57 @@ public class LambdaConfigurationFixture
...
+							m.DefaultSchema = "MySche";
+						});
+					x.DataBaseIntegration(
+						db =>
+						{
+							db.Dialect<MsSql2000Dialect>();
+							db.KeywordsAutoImport = Hbm2DDLKeyWords.AutoQuote;
 #if NETFX
-												db.Batcher<SqlClientBatchingBatcherFactory>();
+							db.Batcher<SqlClientBatchingBatcherFactory>();
```
bahusoid: Interesting why CI throws here 
`CfgTest/Loquacious/LambdaConfigurationFixture.cs(56,19): error CS0246: The type or namespace name 'SqlClientBatchingBatcherFactory' could not be found`

Seems `if NETFX ` is ignored for some reasons

hazzik:
>IMHO there is no need in interfaces for such DTO like classes and they only clutter code and complicate adding new features.

Agree

However, I propose to convert these methods to instance methods. This will allow to avoid breaking changes (the instance methods will supersede the extensions) while leaving interface similar and using the classes.

bahusoid:
> I propose to convert these methods to instance methods

Done

PR #2120, opened 2019-04-11
Prevent calling UpdateTimestampsCache if query spaces are not changed

### LINKED ISSUE #2115: Statefull Session commit performance issue when nothing changed and second level cache with query cache enabled
Hi,
I realised a perfomance issue, we are using second level cache so we need a session and transaction by default. After data read commit occours thats fine but it tries to do something with 
> UpdateTimestampsCache

```c#
[MethodImpl(MethodImplOptions.Synchronized)]
public virtual void Invalidate(IReadOnlyCollection<string> spaces)
{
	//TODO: to handle concurrent writes correctly, the client should pass in a Lock
	long ts = _updateTimestamps.NextTimestamp();
	//TODO: if lock.getTimestamp().equals(ts)
	if (log.IsDebugEnabled())
		log.Debug("Invalidating spaces [{0}]", StringHelper.CollectionToString(spaces));
	SetSpacesTimestamp(spaces, ts);
}
```

As you can see above 

> [MethodImpl(MethodImplOptions.Synchronized)] is a problem if spaces is empty. There are other sync. methods some of them check spaces before calling sync method some not.
I think all of them need to check spaces before calling sync method for concurrency performance issue?

Below usage checks before calling sync method.

> StandartQueryCache
```c#
var upToDates = spacesToCheck.Count > 0
				? _updateTimestampsCache.AreUpToDate(spacesToCheck.ToArray(), checkedSpacesTimestamp.ToArray())
				: Array.Empty<bool>();
```

hazzik:
Fixes #2115


src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -551,7 +551,7 @@ private static ICacheAssembler[] GuessTypes(IList cacheable)
 
 		protected virtual bool IsUpToDate(ISet<string> spaces, long timestamp)
 		{
-			return _updateTimestampsCache.IsUpToDate(spaces, timestamp);
+			return spaces.Count == 0 || _updateTimestampsCache.IsUpToDate(spaces, timestamp);
```
bahusoid: It feels like better fix would be to replace `Synchronized` attributes with locking on field in `UpdateTimestampsCache` and move `spaces.Count` check inside.

hazzik: Maybe. I'll change.

gokhanabatay: Hi @hazzik,
Also need to check spaces.count before calling Invalidate and PreInvalidate?
```c#
if (session.Factory.Settings.IsQueryCacheEnabled)
{
	session.Factory.UpdateTimestampsCache.Invalidate(executedSpaces);
}

if (session.Factory.Settings.IsQueryCacheEnabled)
{
	session.Factory.UpdateTimestampsCache.PreInvalidate(executedSpaces);
}
```

hazzik: There is a problem with hiding this logic into the method. These methods are virtual and if the logic is hidden there it brings a burden of implementation to the client. So, after thinking overnight I think the original approach is better.

I'll probably change the logic to be in both places. Inside and outside of these method calls.

bahusoid: > it brings a burden of implementation to the client

IMHO it's OK to make it client burden. We shouldn't try to foresee and optimize user implementations.

gokhanabatay: Hi @hazzik ,
I also realized that we are storing updatetimestampscache for non cache'ing entities.
Is this really necessary in our scenario we use query cache for just Cache.ReadWrite or Cache.Readonly entities?

We need high performance to do that we need to decrease Sync. and locking for common dml/select operations.

I think that this needs to be configurable(Store all entities updatetimestampscache or just cacheable entities) or we give us a way to implement our UpdateTimestampsCache?

In distributed cache scenario there is a lot of cache call(for cached entities) may be there would be a way to implement memory cache in front of second level cache?

src/NHibernate/Async/Cache/UpdateTimestampsCache.cs
```diff
@@ -55,20 +49,24 @@ public Task PreInvalidateAsync(object[] spaces, CancellationToken cancellationTo
...
 
-		[MethodImpl()]
 		public virtual async Task PreInvalidateAsync(IReadOnlyCollection<string> spaces, CancellationToken cancellationToken)
 		{
 			cancellationToken.ThrowIfCancellationRequested();
-			using (await _preInvalidate.LockAsync())
+			if (spaces.Count == 0)
+				return;
+
+			Lock.EnterUpgradeableReadLock();
```
hazzik: This is, most-likely, invalid.

maca88: Yes, it is invalid. I will try to update the async generator to support it. I think that it is better to use `EnterReadLock` instead `EnterUpgradeableReadLock` as here doesn't seem to be needed. Also, the async version of    `EnterUpgradeableReadLock` is harder to implement and it is also not available in the latest [AsyncEx library](https://github.com/StephenCleary/AsyncEx/issues/97).

hazzik: This lock upgrades to write lock a few lines below.

I highly disagree with the decision to drop upgradable locks support from AsyncEx and with the reasoning behind them.

To be honest I’m biased against author of AsyncEx after they claimed that AsyncLocal does not impact performance in any how; which turned to be completely false claim in practice. So I do not accept their reasoning for dropping the upgradable locks support as valid. 

I believe here is a perfect example of need of upgradable read locks hot high performance. Because we do not know the exact implementation of L2 Cache we assume the worst: it is slow. For example, L2 cache can be non local so there could be some latency in reading data. We want to allow people to read data while they are reading other data.

maca88: My bad, I missed that. Wouldn't be better to just use `(Enter\Exit)WriteLock` instead of `(Enter\Exit)UpgradeableReadLock` inside `PreInvalidate` and `Invalidate` methods and remove the lock from the `SetSpacesTimestamp` as in this case it will always be upgraded, except when canceling the cancellation token when `SetSpacesTimestampAsync` is called which is very hard to happen.

About `AyncEx`, I do agree that their reasoning for droping the upgradable locks is not valid. Anyway, we will have to write our own implementation in order to avoid adding external dependencies and if we are able to avoid upgradeable locks we would have an easier job for implementing our own `AsyncReadWriteLock`.

PR #2123, opened 2019-04-12
Small stylistic improvement in unit test

HerrAachen:
Using Assert.Throws is a more standard and less error-prone way to expect exceptions in unit test than try-catch-blocks

### DIFF
--- a/src/NHibernate.Test/Hql/SimpleFunctionsTest.cs
+++ b/src/NHibernate.Test/Hql/SimpleFunctionsTest.cs
@@ -23,15 +23,7 @@ namespace NHibernate.Test.Hql
 			Assert.AreEqual("noArgs", nf.Render(args, factoryImpl).ToString());
 
 			args.Add("aparam");
-			try
-			{
-				SqlString t = nf.Render(args, factoryImpl);
-				Assert.Fail("No exception if has argument");
-			}
-			catch (QueryException)
-			{
-				//correct
-			}
+			Assert.Throws<QueryException>(() => nf.Render(args, factoryImpl));
 		}
 
 		[Test]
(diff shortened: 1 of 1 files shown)


hazzik:
Thanks.

src/NHibernate.Test/Hql/SimpleFunctionsTest.cs
```diff
@@ -218,15 +178,7 @@ public void ClassicAvg()
 			args.Clear();
 			args.Add("va1");
 			args.Add("va2");
-			try
-			{
-				SqlString t = caf.Render(args, factoryImpl);
-				Assert.Fail("No exception 2 argument without <setquantifier>:" + t);
```
bahusoid: But why assert messages were removed? It was merged too fast

HerrAachen: I'm happy to create another PR. Do you think these messages added valuable context?

hazzik: Imo, not really

bahusoid: Now it's not clear at all why ClassicAvg test should fail with `QueryException`

bahusoid: Most test names here don't provide any context. So not sure why you think it's not valuable.

HerrAachen: @bahusoid I created another PR here: https://github.com/nhibernate/nhibernate-core/pull/2124
We can continue the conversation there.

PR #2128, opened 2019-04-15
Fix GetMany for ReadWriteCache

### LINKED ISSUE #2127: StackExchangeRedisCache with PreferMultipleGet = true calls GetMany multiple times
Hi @fredericDelaporte,  @maca88 , @hazzik 
I try to use NHibernate.Caches.StackExchangeRedis with PreferMultipleGet = true option but; it calls GetMany over an over record count.

> In ReadWriteCache result[i] = null;  makes cache GetMany call over an over why is that we already get the whole data just one call?

It has the same issue with GetManyAsync
Also is it really need to "lock" with Get operation, we already check timestamp cache? 
May be NonstrictReadWriteCache is the answer for less locking.

```c#
public object[] GetMany(CacheKey[] keys, long timestamp)
{
	if (log.IsDebugEnabled())
	{
		log.Debug("Cache lookup: {0}", string.Join(",", keys.AsEnumerable()));
	}
	var result = new object[keys.Length];
	lock (_lockObject)
	{
		var lockables = _cache.GetMany(keys.Select(o => (object) o).ToArray());
		for (var i = 0; i < lockables.Length; i++)
		{
			var lockable = (ILockable) lockables[i];
			var gettable = lockable != null && lockable.IsGettable(timestamp);

			if (gettable)
			{
				if (log.IsDebugEnabled())
				{
					log.Debug("Cache hit: {0}", keys[i]);
				}
				result[i] = ((CachedItem) lockable).Value;
			}

			if (log.IsDebugEnabled())
			{
				log.Debug(lockable == null ? "Cache miss: {0}" : "Cached item was locked: {0}", keys[i]);
			}

			result[i] = null; 
		}
	}
	return result;
}
```

> If PreferMultipleGet = true then StandartQueryCache Get operation calls GetMany when batching enabled. Then gets table updatetimestampscache from cache.
This code calls cache server again and again. With batching it could be done with single call?

```c#
public IList Get(QueryKey key, ICacheAssembler[] returnTypes, bool isNaturalKeyLookup, ISet<string> spaces, ISessionImplementor session)
{
	if (Log.IsDebugEnabled())
		Log.Debug("checking cached query results in region: '{0}'; {1}", _regionName, key);

	var cacheable = (IList) Cache.Get(key);
	if (cacheable == null)
	{
		Log.Debug("query results were not found in cache: {0}", key);
		return null;
	}

	var timestamp = (long) cacheable[0];

	if (Log.IsDebugEnabled())
		Log.Debug("Checking query spaces for up-to-dateness [{0}]", StringHelper.CollectionToString(spaces));

	if (!isNaturalKeyLookup && !IsUpToDate(spaces, timestamp))
	{
		Log.Debug("cached query results were not up to date for: {0}", key);
		return null;
	}

	return GetResultFromCacheable(key, returnTypes, isNaturalKeyLookup, session, cacheable);
}
```

hazzik:
- Combine cache hit/miss logs where possible

Fixes #2127

### DIFF
--- a/src/NHibernate/Cache/ReadWriteCache.cs
+++ b/src/NHibernate/Cache/ReadWriteCache.cs
@@ -111,35 +111,8 @@ namespace NHibernate.Cache
 				/*try
 				{
 					cache.Lock( key );*/
-
-				ILockable lockable = (ILockable) Cache.Get(key);
-
-				bool gettable = lockable != null && lockable.IsGettable(txTimestamp);
-
-				if (gettable)
-				{
-					if (log.IsDebugEnabled())
-					{
-						log.Debug("Cache hit: {0}", key);
-					}
-
-					return ((CachedItem) lockable).Value;
-				}
-				else
-				{
-					if (log.IsDebugEnabled())
-					{
-						if (lockable == null)
-						{
-							log.Debug("Cache miss: {0}", key);
-						}
-						else
-						{
-							log.Debug("Cached item was locked: {0}", key);
-						}
-					}
-					return null;
-				}
+				var lockable = (ILockable) Cache.Get(key);
+				return GetValue(txTimestamp, key, lockable);
 				/*}
 				finally
 				{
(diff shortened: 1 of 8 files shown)


src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -241,7 +220,7 @@ public ISoftLock Lock(CacheKey key, object version)
 				{
 					log.Debug("Caching: {0}", string.Join(",", keys.AsEnumerable()));
 				}
-				var keysArr = keys.Cast<object>().ToArray();
+				var keysArr = keys.ToArray<object>();
```
bahusoid: Why this conversion is required? You can call `_cache.LockMany(keys);` directly

src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -157,30 +130,36 @@ public object[] GetMany(CacheKey[] keys, long timestamp)
 			var result = new object[keys.Length];
 			lock (_lockObject)
 			{
-				var lockables = _cache.GetMany(keys.Select(o => (object) o).ToArray());
+				var lockables = _cache.GetMany(keys.ToArray<object>());
```
bahusoid: Why not `_cache.GetMany(keys)`? Same question for all other places with `ToArray` conversion

src/NHibernate/Cache/ReadWriteCache.cs
hazzik: I don't know, I just removed `Cast` which is not needed.

hazzik: I think this was done to prevent co-variant array conversion complaint by R#

src/NHibernate/Cache/ReadWriteCache.cs
hazzik: Same as above.

src/NHibernate/Cache/ReadWriteCache.cs
bahusoid: This array is not expected to be modified so it shouldn't be an issue. IMHO we should remove those conversions if R# complaint is the only reason.

hazzik: Done

PR #2132, opened 2019-04-15
Add GetMany for ReadWriteCache tests

maca88:
Tests for #2127.

### DIFF
--- a/src/NHibernate.Test/CacheTest/BatchableCacheFixture.cs
+++ b/src/NHibernate.Test/CacheTest/BatchableCacheFixture.cs
@@ -202,6 +202,351 @@ namespace NHibernate.Test.CacheTest
 		}
 
 		[Test]
+		public void GetManyReadWriteTest()
+		{
+			var persister = Sfi.GetEntityPersister(typeof(ReadWrite).FullName);
+			Assert.That(persister.Cache.Cache, Is.Not.Null);
+			Assert.That(persister.Cache.Cache, Is.TypeOf<BatchableCache>());
+			int[] getIds;
+			int[] loadIds;
+
+			using (var s = Sfi.OpenSession())
+			using (var tx = s.BeginTransaction())
+			{
+				var items = s.Query<ReadWrite>().ToList();
+				loadIds = getIds = items.OrderBy(o => o.Id).Select(o => o.Id).ToArray();
+				tx.Commit();
+			}
+
+			// Batch size 3
+			var parentTestCases = new List<Tuple<int[], int, int[][], int[], Func<int, bool>>>
+			{
+				// When the cache is empty, GetMany will be called three times. First time in type
+				// DefaultLoadEventListener, the second time in BatchingEntityLoader and third in ReadWriteCache.
+				new Tuple<int[], int, int[][], int[], Func<int, bool>>(
+					loadIds,
+					0,
+					new[]
+					{
+						new[] {0, 1, 2}, // Triggered by LoadFromSecondLevelCache method of DefaultLoadEventListener type
+						new[] {1, 2, 3}, // Triggered by Load method of BatchingEntityLoader type
+						new[] {0, 1, 2} // Triggered by PutMany method of ReadWriteCache type
+					},
+					new[] {0, 1, 2},
+					null
+				),
+				// When there are not enough uninitialized entities after the demanded one to fill the batch,
+				// the nearest before the demanded entity are added.
+				new Tuple<int[], int, int[][], int[], Func<int, bool>>(
+					loadIds,
+					4,
+					new[]
+					{
+						new[] {4, 5, 3},
+						new[] {5, 3, 2},
+						new[] {4, 5, 3},
+					},
+					new[] {3, 4, 5},
+					null
+				),
+				new Tuple<int[], int, int[][], int[], Func<int, bool>>(
+					loadIds,
+					5,
+					new[]
+					{
+						new[] {5, 4, 3},
+						new[] {4, 3, 2},
+						new[] {5, 4, 3},
+					},
+					new[] {3, 4, 5},
+					null
+				),
+				new Tuple<int[], int, int[][], int[], Func<int, bool>>(
+					loadIds,
+					0,
+					new[]
+					{
+						new[] {0, 1, 2} // 0 get assembled and no further processing is done
+					},
+					null,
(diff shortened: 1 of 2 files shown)


src/NHibernate.Test/CacheTest/BatchableCacheFixture.cs
```diff
@@ -201,6 +201,351 @@ public void MultipleGetReadOnlyCollectionTest()
...
+				var items = s.Query<ReadWrite>().ToList();
+				loadIds = getIds = items.OrderBy(o => o.Id).Select(o => o.Id).ToArray();
+				tx.Commit();
+			}
+
+			// Batch size 3
+			var parentTestCases = new List<Tuple<int[], int, int[][], int[], Func<int, bool>>>
+			{
+				// When the cache is empty, GetMany will be called three times. First time in type
+				// DefaultLoadEventListener, the second time in BatchingEntityLoader and third in ReadWriteCache.
```
bahusoid: Could you please clarify why multiple calls are required?

> When the cache is empty

And how many calls if cache is full?

maca88: > Could you please clarify why multiple calls are required?

The main reason why there are multiple calls when a cache miss occurs, is because `BatchingEntityLoader` does not know which keys `DefaultLoadEventListener` tried to get and the same applies for `ReadWriteCache`. If we would want to have only one call we would need to store the information about the current retrieved keys within an operation (in this case `ISession.Get`), which could be used by the mentioned classes.

> And how many calls if cache is full?

One.

bahusoid: >  If we would want to have only one call we would need to store the information about the current retrieved keys within an operation 

But why it knows when cache is full? Can't we add keys similar way it's done for full cache scenarios

maca88: > But why it knows when cache is full?

Because it is checked and retuned earlier if it exists. 

>  Can't we add keys similar way it's done for full cache scenarios

When the entity exists in the cache, it won't be readded into the cache.

In order to be more specific I will explain how the current logic works and what can be done to optimize it:
When retrieving a cacheable entity by calling `ISession.Get()` the `DefaultLoadEventListener.Load` is called which uses 3 steps to get the entity:
1. Tries to get it from the session persistence context (not relevant for this conversation)
2. Tries to get it from the second level cache
3. Tries to get it from the database

When the entity exists in the second level cache, it will be assembled and returned in the second step, which means that `BatchingEntityLoader` and `ReadWriteCache` won't be used. Only one cache call will be made in this case.
When the entity does not exist in the second level cache, the third step will be used which involves `BatchingEntityLoader` and afterwards `ReadWriteCache`. `BatchingEntityLoader` uses `BatchFetchQueue` to fill the batch of ids that will be retrieved from the database. In order to populate the batch of ids to be retrieved from the database, `BatchFetchQueue` calls the second level cache (`BatchFetchQueue.AreCached`) for uninitialized entities in order to populate the batch only with ids that are not in the cache. [Here](https://github.com/nhibernate/nhibernate-core/pull/2132/files#diff-f6f2c63dece44ec4a12f849d5103411fR243) we can see that the same ids (1 and 2) will be checked, which were already checked in the first step. After the entites are retreived for the calculated batch of ids (0,1 and 2) from the database `ReadWriteCache` again calls the cache to retrieve the entities in order to update them. In this case 3 `GetMany` call are triggered and after revisiting the logic we could optimize it to have 2 calls instead. So instead of having:
```
new[]
{
  new[] {0, 1, 2}, // Triggered by LoadFromSecondLevelCache method of DefaultLoadEventListener
  new[] {1, 2, 3}, // Triggered by Load method of BatchingEntityLoader
  new[] {0, 1, 2} // Triggered by PutMany method of ReadWriteCache
},
```
we would have:
```
new[]
{
  new[] {0, 1, 2}, // Triggered by LoadFromSecondLevelCache method of DefaultLoadEventListener
  new[] {0, 1, 2} // Triggered by PutMany method of ReadWriteCache
},
```
The call from `ReadWriteCache` is still required as the cache may be populated by another thread in the meantime. So in order to remove the second call, we would need to pass the `GetMany` result from `DefaultLoadEventListener.LoadFromSecondLevelCache` to the `BatchingEntityLoader` which in this case would have enough ids for populating the batch without triggering another `GetMany` call.

bahusoid: Thanks for taking the time to explain it!
> So in order to remove the second call, we would need to pass the GetMany result from DefaultLoadEventListener.LoadFromSecondLevelCache to the BatchingEntityLoader which in this case would have enough ids for populating the batch without triggering another GetMany call.

Can't we simply instruct `BatchingEntityLoader` to skip cache checks when it's called from `DefaultLoadEventListener` context?

>The call from ReadWriteCache is still required as the cache may be populated by another thread in the meantime

I'm not sure I understand the implication. Is this some kind of `if(!exists(key)) add(key,value)` check?

bahusoid: And shouldn't we do `CacheMode` check inside `BatchingEntityLoader` as it's done inside DefaultLoadEventListener:
https://github.com/nhibernate/nhibernate-core/blob/78ce59ed1ae3003f3f59c98d92f75ad3a26217d4/src/NHibernate/Event/Default/DefaultLoadEventListener.cs#L418-L424

maca88: > Can't we simply instruct BatchingEntityLoader to skip cache checks when it's called from DefaultLoadEventListener context?

I don't thnik so, as the first `GetMany` call could miss for the key that was requested from the caller and hit for the others. In this case `BatchingEntityLoader` would have to find two more keys to fill the batch, which are not in the cache.

> I'm not sure I understand the implication. Is this some kind of if(!exists(key)) add(key,value) check?

Some kind yes but it applies also for update. Another thread may update an existing entity and the cache between the time the entity was retrieved from the database and when `PutMany` is called. In order to avoid overriding the value that was set in the meantime we need to lock the keys and retrieve the current key values and update them only when the key was not modified. 

> And shouldn't we do CacheMode check inside BatchingEntityLoader

`BatchingEntityLoader` is used only when the batch size is greater that one, the same check has to be done when batching is disabled.

PR #2133, opened 2019-04-16
WIP Multi-Tenancy: Separate second-level cache per tenant (prototype for discussion)

bahusoid:
This is prototype just for discussion so please ignore any breaking or non-functional changes
Follow up to #2108

I think we should create separate cache per tenant for better data separation and to allow tenant specific cache clean up (must have feature from my multi-tenant experience) and other actions (like `ISessionFactory.EvictEntity`, `EvictCollection`, ....). Without this change `BulkOperationCleanupAction` cleans up cache for all tenants  when it's executed for some specific tenant. I don't see that it's somehow handled in hibernate so seems they clean cache for all tenants too (also found related complaints [here](https://stackoverflow.com/questions/46584489/hibernate-5-multitenancy-ehcache-configure-cache-for-each-tenant))

So my idea is to create cache dynamically by `tenantIdentifier` and provide to persisters cache creation delegate instead of already created instance of cache.

Some pending questions if you agree with this plan:
1) Should we still keep `tenantIdentifier` as part of `CacheKey`?
*me:* for now seems unnecessary.
2) Should `tenantIdentifier` be part of cache region name or stored as separate value and added as `CacheBase.TenantIdentifier`?
*me:* ~keep as separate value looks more cleaner solution.~  Made as part of cache region to minimize overall changes
3) Should we do something similar for `QueryCache`? For query cache user externally provides `regionName` and in this context `tenantIdentifier` is also available so we can leave it to user. 
*me:* should apply same logic

Any objections? Any other ideas how it should be implemented or suggestions?

### DIFF
--- a/src/NHibernate/Impl/SessionFactoryImpl.cs
+++ b/src/NHibernate/Impl/SessionFactoryImpl.cs
@@ -232,7 +232,7 @@ namespace NHibernate.Impl
 
 			#region Persisters
 
-			var caches = new Dictionary<Tuple<string, string>, ICacheConcurrencyStrategy>();
+			var caches = new ConcurrentDictionary<Tuple<string, string>, ICacheConcurrencyStrategy>();
 			entityPersisters = new Dictionary<string, IEntityPersister>();
 			implementorToEntityName = new Dictionary<System.Type, string>();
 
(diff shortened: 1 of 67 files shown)


src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -426,25 +427,26 @@ private IQueryCache BuildQueryCache(string queryCacheName)
...
 			if (strategy == null || !settings.IsSecondLevelCacheEnabled)
 				return null;
 
-			var cacheKey = new Tuple<string, string>(cacheRegion, strategy);
-			if (caches.TryGetValue(cacheKey, out var cache)) 
-				return cache;
+			return tenantIdentifier =>
+			{
+				var cache = caches.GetOrAdd(
+					new Tuple<string, string>(tenantIdentifier + "|" + cacheRegion, strategy),
```
bahusoid: Oups... I've pushed version with region name modification (and my current code is a mess so I won't update it now). But now `tenantIdentifier` is part of  cache key 
`new Tuple<string, string, string>(tenantIdentifier, cacheRegion, strategy)`

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -426,25 +427,26 @@ private IQueryCache BuildQueryCache(string queryCacheName)
 					properties);
 		}
 
-		private ICacheConcurrencyStrategy GetCacheConcurrencyStrategy(
+		private Func<string, ICacheConcurrencyStrategy> GetCacheConcurrencyStrategy(
```
bahusoid: Instead of cache instance returns delegate that will be used by persisters to obtain cache instance by `tenantIdentifier`.

PR #2136, opened 2019-04-19
Fix NullReferenceException in EntityEntry.GetLoadedValue while Update…

picbeats:
… on detached entity with null logical one-to-one reference
I had to apply this fix, to make 5.2.3 work with our system after migration from 4.0.0
The use case is rather exotic and triggered an exception on not so nice written test code, using entity, having a dependent one to many (effective one-to-one) child. In first session the entity is inserted (the child entity as a property is not set), than the entity is modified and updated in a second session (detached). Result was a NullReferenceException in EntityEntry.GetLoadedValue. The entity was never loaded from database, so loadedState is null. I added the check if the loaded state is null, and modified Cascade(Async).CascadeProperty to handle EntityEntry.GetLoadedValue returning a null safely.
With this fix it pass all tests and works fine on our system.  I didn't find out how to add a test though..

### DIFF
--- a/src/NHibernate/Engine/EntityEntry.cs
+++ b/src/NHibernate/Engine/EntityEntry.cs
@@ -205,6 +205,7 @@ namespace NHibernate.Engine
 
 		public object GetLoadedValue(string propertyName)
 		{
+			if (loadedState == null) return null;
 			int propertyIndex = ((IUniqueKeyLoadable) persister).GetPropertyIndex(propertyName);
 			return loadedState[propertyIndex];
 		}
--- a/src/NHibernate/Async/Engine/Cascade.cs
+++ b/src/NHibernate/Async/Engine/Cascade.cs
@@ -130,7 +130,11 @@ namespace NHibernate.Engine
 								// which association may be null) in the loadedState of the parent. The unwrap flag
 								// causes it to support having a null implementation, instead of throwing an entity
 								// not found error.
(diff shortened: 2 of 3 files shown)


src/NHibernate/Engine/EntityEntry.cs
```diff
@@ -205,6 +205,7 @@ public EntityKey EntityKey
 
 		public object GetLoadedValue(string propertyName)
 		{
+			if (loadedState == null) return null;
```
bahusoid: This doesn't look like correct fix. You should really investigate why `entityEntry.LoadedState` is null in your case. It looks like the real issue in some other place.

Maybe in throwing place value should be retrieved as:
```C#
entityEntry.LoadedState != null
				? entityEntry.GetLoadedValue(propertyName)
				: entityEntry.Persister.GetPropertyValue(owner, propertyName);
```

But it's really hard to say without unit test.

hazzik: LoadedState can be null for readonly entities. However I don’t know the exact reason why it is null here. We definitely need a test case.

picbeats: [](https://github.com/nhibernate/nhibernate-core/issues/2137)
I have now a test code, sorry for issue duplicate, first time contributing on github.

picbeats: https://github.com/nhibernate/nhibernate-core/issues/2137

picbeats: Created a new pull request, with unit test and fix as suggested by @bahusoid

hazzik:
Replaced by #2138

PR #2138, opened 2019-04-21
GH2137 NullReferenceException in EntityEntry.GetLoadedValue

picbeats:
Occures on update of inserted detached entity (see the unit test)
Tested to pass with Postgresql
The fix as suggested by @bahusoid


hazzik:
Hi @picbeats thanks for the contribution. There are a few issues with that:

- The test is failing on Firebird
- Need to run async generator

bahusoid:
> The fix as suggested by @bahusoid

This is not actually my suggested solution - as I didn't investigate the issue. And now it seems it's not correct fix. At least in this context (orphan deletion) - if `LoadedState` is null routine should either fail (and fix should be applied somewhere else) or be skipped (but not sure if it's a good idea).

It seems proper fix should somehow skip delete orphans routine altogether if called from `DefaultSaveOrUpdateEventListener.PerformUpdate` context - as it's intended for detached entities. And orphans are not possible for detached entities;

bahusoid:
Hmm.. From what I see `session.Update` actually doesn't create `LoadedState` for objects. Is this expected behavior? I personally didn't  use `Update` (as it issues update statement no matter dirty object or not, I'm using `Lock` instead) - but missing `LoadedState` means broken dirty check for such objects in sessions. Quite unexpected behavior to me.

So 
>skip delete orphans routine altogether if called from DefaultSaveOrUpdateEventListener.PerformUpdate context

will not fix this issue as null `LoadedState` will still fail on flush. And now it looks like the proper fix - we should create `LoadedState` for `Update`

bahusoid:
As this `session.Update` behavior (not adding LoadedState) is some old hibernate legacy  decided to go with simple workaround - skip orphan processing logic if `LoadedState` is null.

hazzik:
I've referenced an incorrect PR in the merged commit, so really had to force push. The correct commit is b734d00.

PR #2142, opened 2019-04-24
Fix properties mapped inside a <properties> group are empty when retrieving object

### LINKED ISSUE #1006: NH-2714 - Properties mapped inside a <properties> group are not set when retrieving object
**Dave Warry** created an [issue](https://nhibernate.jira.com/browse/NH-2714) — <time datetime="2011-05-19T01:07:17.299Z" title="2011-05-19T01:07:17.299Z">19th May 2011, 1:07:17</time>:
> See the attached failing test. 
> 
> Class A has two properties (Name, ExtraId) mapped within a <properties> group. Retrieving an instance of A returns an instance without these properties being set.
> 
> Some preliminary debugging shows that two instances of A are being created by the call to Session.Get(). The first is being correctly inititalised; the second (which is the instance actually returned) isn't.
>

lillo42:
Fixes #1006


src/NHibernate/Type/EmbeddedComponentType.cs
```diff
@@ -19,9 +19,12 @@ public override bool IsEmbedded
 
 		public override object Instantiate(object parent, ISessionImplementor session)
 		{
-			bool useParent= false;
-			// NH Different implementation : since we are not sure about why H3.2 use the "parent"
-			//useParent = parent != null && base.ReturnedClass.IsInstanceOfType(parent);
+			var useParent =
+				parent != null &&
+				//TODO: Yuck! This is not quite good enough, it's a quick
```
hazzik: The comment is from here: https://github.com/hibernate/hibernate-orm/blob/5c5de945e56b613b6ed8d0acc060bb124ca23584/hibernate-core/src/main/java/org/hibernate/type/EmbeddedComponentType.java#L34-L37

src/NHibernate/Type/EmbeddedComponentType.cs
```diff
@@ -19,9 +19,12 @@ public override bool IsEmbedded
 
 		public override object Instantiate(object parent, ISessionImplementor session)
 		{
-			bool useParent= false;
-			// NH Different implementation : since we are not sure about why H3.2 use the "parent"
```
hazzik: So, now we're sure why this code was written like this...

hazzik:
@lillo42 if you're ok. I'm going to force push the PR removing the merge commits.

hazzik:
Thanks @lillo42

deAtog:
I just spent a day debugging this issue on 4.0.4.GA only to find that this bug was reported over 8 years ago and was the result of changes made to NHibernate while porting code from Hibernate. Bug fixes like these ought to be back-ported/committed to earlier versions when possible, especially when the H3.2 code on which this fix is based was written 12 years ago.

PR #2143, opened 2019-04-25
Use reference equality in InsertActionSorter

### LINKED ISSUE #2141: Undefined call to Equals object in collection during flush just before commit
Since my upgrade from NH3.4 to 5.2.5, I discover a strange behavior.
Just before the commit on a my collection of entities, a strange call to the override Equals method is fired for each saved entities with empty object. This causes a crash (cast exception) because i don't check the type of object passed in argument for comparison.  
I tried to save my entities with a specific transaction and commit between each object (and clear NH session at each time) and it's ok in this case !

A workaround is to check the typeof of my object in my override Equals method but i need to recode all my entities (more than 400 entity). it's not normal.

I download source code for NH5.2.5 and this produce when the Flush() method is called in FlushBeforeTransactionCompletion method for auto FlushMode.  In this case, the  flushEventListener array contains something...
I can disable the autoFlush mode but it's not recommanded.

Help, please. Thanks

bahusoid:
Fixes #2141

### DIFF
--- a/src/NHibernate.Test/Async/Insertordering/AnimalModel/Fixture.cs
+++ b/src/NHibernate.Test/Async/Insertordering/AnimalModel/Fixture.cs
@@ -163,5 +163,29 @@ namespace NHibernate.Test.Insertordering.AnimalModel
 				// instead.)
 			}
 		}
+
+		// #2141
+		[Test]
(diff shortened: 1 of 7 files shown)


hazzik:
@bahusoid can you please retarget to 5.2.x?

bahusoid:
> can you please retarget to 5.2.x?

Done

PR #2146, opened 2019-04-27
Fix #1228

### LINKED ISSUE #1228: NH-2648 - HQL with joins in sub-select creates wrong SQL
**Harald M. Müller** created an [issue](https://nhibernate.jira.com/browse/NH-2648) — <time datetime="2011-04-13T02:51:24.728Z" title="2011-04-13T02:51:24.728Z">13th April 2011, 2:51:24</time>:
> The following query is translated to wrong SQL - the join is missing from the sub-select:
> 
> ```
> 
>     SELECT ROOT FROM NHibernate.Test.NHSpecificTest.HQL_MissingJoinInExists.Sheet AS ROOT 
>                             WHERE (EXISTS (FROM NHibernate.Test.NHSpecificTest.HQL_MissingJoinInExists.Shelf AS inv  
>                                            JOIN ROOT.Folder AS ROOT_Folder 
>                                            WHERE (((ROOT_Folder.Shelf) = (inv) AND inv.Id = 1)) )) 
>                                   AND ROOT.Name = 'SomeName'
> 
> ```
> 
> (The query was created by a HQL generator of ours - hence the superfluous parentheses). The exceotion is:
> 
> NHibernate.Exceptions.GenericADOException : could not execute query
> < select sheet0*.Id as Id2_, sheet0_.Name as Name2_, sheet0_.Folder as Folder2_ from Sheet sheet0_ where (exists (select shelf1_.Id from Shelf shelf1_ where folder2_.Shelf=shelf1_.Id and shelf1_.Id=1)) and sheet0*.Name='SomeName' >
> [SQL: select sheet0*.Id as Id2_, sheet0_.Name as Name2_, sheet0_.Folder as Folder2_ from Sheet sheet0_ where (exists (select shelf1_.Id from Shelf shelf1_ where folder2_.Shelf=shelf1_.Id and shelf1_.Id=1)) and sheet0*.Name='SomeName']
>   ----> System.Data.SqlClient.SqlException : The multi-part identifier "folder2_.Shelf" could not be bound.
> 
> 
> Since NHib 0.99 and in Hibernate, it has been possible to use joins inside subqueries. HQL allows it, and therefore, it should produce correct SQL (both syntactically and semantically). Actually, joins in subqueries are **necessary** if there are OR or NOT operators inside the subquery (some simpler queries can be rewritten with all Joins at top-level).
> 
> I consider this bug critical because there is no workaround for
> 
> * HQL generated for earlier NHib versions
> * manual HQL that needs OR or NOT inside a subquery
> * (I assume) the correct translation of Linq's .Any in the NHib.Linq provider
> 
> I have attached a test case that shows the behavior.
> 
> 
---
**Harald M. Müller** added a comment — <time datetime="2011-04-13T10:45:33.759Z" title="2011-04-13T10:45:33.759Z">13th April 2011, 10:45:33</time>:
> I looked a little bit in the code and found the following in HqlSqlWalker.g:
> 
>     fromElement! 
>     ...:
>        | je=joinElement 
>           -> //$je
> 
> i.e., the return $je of a joinElement is commented, so it returns the null tree. Probably, global joins are handled via some global variable machinery (there is a call to CreateFromJoinElement(...) inside joinElement's production), but this appears not to work for subqueries. Maybe this helps to solve this issue.
> 
> 
---
**Harald M. Müller** added a comment — <time datetime="2011-04-14T13:26:27.916Z" title="2011-04-14T13:26:27.916Z">14th April 2011, 13:26:27</time>:
> I have attached two images that show that the AST is in in inconsistent state: The "_fromElementsByTableAlias" collection for the subquery contains, as expected, two nodes (the second one is the joined one). But the AST itself contains only one child ...
> 
---
**Harald M. Müller** added a comment — <time datetime="2011-04-22T00:45:23.947Z" title="2011-04-22T00:45:23.947Z">22nd April 2011, 0:45:23</time>:
> Unfortunately, this bug is already in Hibernate. I have opened issue HHH-6151 there.
> 
---
**Harald M. Müller** added a comment — <time datetime="2011-04-27T00:25:16.839Z" title="2011-04-27T00:25:16.839Z">27th April 2011, 0:25:16</time>:
> I have attached a new test case NH2648Testcase.zip with (a) cleaner code; (b) a successful and a failing query to show that the problem is the join inside the exists.
> 
---
**Daniel A. Schilling** added a comment — <time datetime="2012-12-04T16:30:46.295Z" title="2012-12-04T16:30:46.295Z">4th December 2012, 16:30:46</time>:
> I believe the entities specified in an HQL join clause have to be related.  See http://nhforge.org/doc/nh/en/index.html#queryhql-joins.  All of the examples in that documentation are things like "from Eg.Cat as cat join cat.Mate" - there is a direct relationship between Cat and Cat.Mate, so it can easily figure out how to constrain the join (the "ON" part of the join).  In the query above - I'm having trouble figuring out what SQL you're expecting.  There is no clear relationship between Shelf and ROOT.Folder specified - at least not in the JOIN - there is later in the WHERE.  So are you expecting a CROSS JOIN?  HQL doesn't support cross joins.
> 
> I think the following HQL would probably work better (but I haven't tested this):
> 
> ```
> 
>     select root
>     from Sheet as root
>     where
>         exists (
>             from Shelf as inv
>             where
>                 root.Folder.Shelf = inv
>                 and inv.Id = 1)
>         and root.Name = 'SomeName'
> 
> ```
> 
> ... using an implicit join instead of an explicit one.  Doing so makes it clear that root.Folder.Shelf is directly related to root, not inv, so we could rewrite this as an explicit join like so:
> 
> ```
> 
>     select root
>     from Sheet as root
>         join root.Folder as fldr
>     where
>         exists (
>             from Shelf as inv
>             where
>                 fldr.Shelf = inv
>                 and inv.Id = 1)
>         and root.Name = 'SomeName'
> 
> ```
> 
> So the JOIN should be in the main query, not the sub query.
> 
---
**Daniel A. Schilling** added a comment — <time datetime="2012-12-04T18:21:29.862Z" title="2012-12-04T18:21:29.862Z">4th December 2012, 18:21:29</time>:
> The subquery is not needed at all here.  This simpler query should accomplish the same thing:
> 
> ```
>     select root
>     from Sheet as root
>         join root.Folder as fldr
>     where
>         fldr.Shelf.Id = 1
>         and root.Name = 'SomeName'
> ```
> 
> 
---
**Daniel A. Schilling** added a comment — <time datetime="2012-12-04T18:22:13.572Z" title="2012-12-04T18:22:13.572Z">4th December 2012, 18:22:13</time>:
> oops.  forgot the "select root" at the beginning.
>

csharper2010:
NH-2648 - HQL with joins in sub-select creates wrong SQL
Join in subselect based on alias for item in root select was not added
This fixes the specific case in the least risky way, all tests green

Signed-off-by: csharper2010 <csharper2010@googlemail.com>

### DIFF
--- a/src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
@@ -736,6 +736,11 @@ namespace NHibernate.Hql.Ast.ANTLR
 
 					HandleWithFragment(fromElement, with);
 				}
+
+				if (fromElement.Parent == null)
+				{
+					fromElement.FromClause.AddChild(fromElement);
+				}
 			}
 
 			if ( log.IsDebugEnabled() )
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,13 @@ insert_final_newline = true
 indent_style = tab
 dotnet_sort_system_directives_first = true
 csharp_space_after_cast = true
+csharp_new_line_before_open_brace = true
+csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
 
 [*.xsd]
 indent_style = tab
--- a/src/NHibernate.Test/NHSpecificTest/GH1228/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH1228/Fixture.cs
@@ -1,0 +1,133 @@
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.GH1228
+{
+	public class Fixture : BugTestCase
+	{
+		[Test]
+		public void TestOk()
+		{
+			using (ISession s = OpenSession())
+			{
+				using (ITransaction t = s.BeginTransaction())
+				{
+					try
+					{
+						{
+							var queryThatWorks = s.CreateQuery(@"
+							SELECT ROOT FROM NHibernate.Test.NHSpecificTest.GH1228.Sheet AS ROOT 
+							WHERE (EXISTS (FROM NHibernate.Test.NHSpecificTest.GH1228.Shelf AS inv  
+										   , ROOT.Folder AS ROOT_Folder 
+										   WHERE ROOT_Folder.Shelf = inv AND inv.Id = 1
+								  )	   ) 
+								  AND ROOT.Name = 'SomeName'");
+							queryThatWorks.List();
+						}
+						{
+							var queryThatWorks = s.CreateQuery(@"
+							SELECT ROOT FROM NHibernate.Test.NHSpecificTest.GH1228.Shelf AS ROOT 
+							WHERE (EXISTS (FROM NHibernate.Test.NHSpecificTest.GH1228.Sheet AS sheet
+										   , ROOT.Folders AS ROOT_Folder 
+										   WHERE ROOT_Folder = sheet.Folder AND sheet.Name = 'SomeName'
+								  )	   ) 
+								  AND ROOT.Id = 1");
+							queryThatWorks.List();
+						}
+					}
+					finally
+					{
+						s.Delete("from Sheet");
+						s.Delete("from Folder");
+						s.Delete("from Shelf");
+						t.Commit();
+					}
+				}
+			}
+		}
+
+		[Test]
+		public void TestWrongSql()
+		{
+			using (ISession s = OpenSession())
+			{
+				using (ITransaction t = s.BeginTransaction())
+				{
+					try
+					{
+						{
+							var queryThatCreatesWrongSQL = s.CreateQuery(@"
+							SELECT ROOT FROM NHibernate.Test.NHSpecificTest.GH1228.Sheet AS ROOT 
+							WHERE (EXISTS (FROM NHibernate.Test.NHSpecificTest.GH1228.Shelf AS inv  
+										   JOIN ROOT.Folder AS ROOT_Folder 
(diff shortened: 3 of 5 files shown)


src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
```diff
@@ -736,6 +736,11 @@ void PrepareFilterParameter()
 
 					HandleWithFragment(fromElement, with);
 				}
+
+				if (fromElement.Parent == null)
```
hazzik: Shouldn't it be done where the `fromElement` is created?

csharper2010: yes, but there are cases where the item is added to the tree explicitly in FromElement.SetOrigin but there is an else branch with comment
`				// Otherwise, the destination node was implied by the FROM clause and the FROM clause processor
				// will automatically add it in the right place.
`

But at least for the implied join in an explicit join no code took care of adding the item to the tree.

I tried to identify the case and go to the previous else if branch (where the comment HHH-276 is located) but that caused one legacy unit test to fail.

As there is no clear responsibility for how the element should be added to the tree, I found it most easy to add it where I did. It's some kind of catch all cases where the item is not added yet which one can find good or bad. But as I said, there is no clear responsibility so I tended towards the feeling: it's a good thing.

bahusoid:
NH-2648 is invalid issue.

```sql
FROM Entity e JOIN e.Parent
```

This query makes sense. It instructs `INNER JOIN Parent` table using association columns specified in `Entity.Parent` mapping.

Query from issue:
```sql 
FROM Entity e WHERE EXISTS ( FROM AnotherEntity JOIN e.Parent WHERE ...)
```
It doesn't make any sense. There is no association between `AnotherEntity` and `e.Parent` so inner join can't be generated.

This fix makes it generated as implicit joins. To me exception looks more appropriate here..

PR #2147, opened 2019-05-01
Improve async locking

maca88:
Currently, the async generator is generating separate fields for lock statements that contain an async invocation, which may cause troubles as two threads may simultaneously execute the same code, one for async and one in sync version of the method. Instead of using a lock statement, this PR uses `AsyncLock` for both sync and async locking, which prevents the mentioned problem. In addition, `AsyncReaderWriterLock` class was added and used for `ReadWriteCache` and `UpdateTimestampsCache` in order to avoid locking when only read operations are performed.


src/AsyncGenerator.yml
```diff
@@ -150,9 +150,6 @@
   transformation:
     configureAwaitArgument: false
     localFunctions: true
-    asyncLock:
```
maca88: I will most likely remove this option in the future as adding an additional field for async locking allows two threads to execute the "same code" simultaneously.

src/NHibernate/Util/AsyncReaderWriterLock.cs
```diff
@@ -0,0 +1,219 @@
...
+				{
+					return;
+				}
+
+				_waitingReadLockSemaphore.Dispose();
+				_waitingReadLockSemaphore = null;
+			}
+		}
+
+		private struct Releaser : IDisposable
```
hazzik: There is no point making this struct if it is a private. It will be boxed and leave in the heap.

maca88: You are right, I changed it to a sealed class.

hazzik: I did some perf testing. The fastest (and free of additional allocations) method is to return the struct. The second place is class (does not matter if it is returned as class or as `IDisposable`) and third place is a struct boxed to `IDisposable`.

maca88: Updated the code to use the fastest option. We could modify the struct to be `readonly`, but it requires C# langauge to be upgraded to `7.2` and I am not sure if it would perform any better. Also, I didn't apply the change for `AsyncLock`, as it would be a breaking change.

src/NHibernate/Util/AsyncReaderWriterLock.cs
```diff
@@ -0,0 +1,229 @@
...
+				if (_currentReaders == 0 && _writeLockSemaphore.CurrentCount == 0)
+				{
+					_readLockSemaphore.Release();
+				}
+			}
+		}
+
+		private void ReleaseWaitingReader()
+		{
+			lock (_writeLockSemaphore)
```
fredericDelaporte, line 187: This creates a synchronization point between all released readers. If think it can be avoided, but I am not sure it is worth it.

In the `ExitWriteLock`, we could store the `_waitingReadLockSemaphore` into a `_waitingDisposalSemaphore` member, release the readers, set `_readersWaiting` directly to 0 and `_waitingReadLockSemaphore` to `null`. `ReleaseWaitingReader` will be dropped.

In `WriteLock` and in `WriteLockAsync`, after the `if (waitForReadLocks) { }`, a non-null `_waitingDisposalSemaphore` would be the disposed and nullified. At that point, we are certain that all readers which were waiting in this semaphore have finished their work, and so it is safe to dispose it.
This is in case we want to do it at the soonest possible point without having to lock for all released readers. Otherwise this disposal could be done in `ExitWriteLock` before storing the current `_waitingReadLockSemaphore` into `_waitingDisposalSemaphore`.

fredericDelaporte:
Rebased for resolving conflicts

src/NHibernate/Util/AsyncReaderWriterLock.cs
```diff
@@ -0,0 +1,229 @@
...
+			{
+				if (_writersWaiting == 0 && _writeLockSemaphore.CurrentCount > 0)
+				{
+					_currentReaders++;
+
+					return true;
+				}
+
+				if (_waitingReadLockSemaphore == null)
+				{
```
fredericDelaporte, line 158: Haven't we a flaw here with current `ReleaseWaitingReader`?
What if an `ExitWriteLock` starts releasing the readers, but then a new write call causes new readers to hit this point while old waiting readers have not yet all already called `ReleaseWaitingReader`?
The new readers would then use the old semaphore, allowing some of them to run immediately. Or I am missing something?

src/NHibernate/Cache/UpdateTimestampsCache.cs
```diff
@@ -54,14 +55,16 @@ public void PreInvalidate(object[] spaces)
 			PreInvalidate(spaces.OfType<string>().ToList());
 		}
 
-		[MethodImpl(MethodImplOptions.Synchronized)]
 		public virtual void PreInvalidate(IReadOnlyCollection<string> spaces)
 		{
-			//TODO: to handle concurrent writes correctly, this should return a Lock to the client
```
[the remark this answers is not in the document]
maca88: Readded.

src/NHibernate/Cache/UpdateTimestampsCache.cs
```diff
@@ -72,22 +75,22 @@ public void Invalidate(object[] spaces)
 			Invalidate(spaces.OfType<string>().ToList());
 		}
 
-		[MethodImpl(MethodImplOptions.Synchronized)]
 		public virtual void Invalidate(IReadOnlyCollection<string> spaces)
 		{
-			//TODO: to handle concurrent writes correctly, the client should pass in a Lock
```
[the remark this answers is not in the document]
maca88: Readded.

src/NHibernate/Util/AsyncReaderWriterLock.cs
maca88: Nice finding! You are correct, unfortunately I wasn't able to make a test that would expose the issue, `_waitingReadLockSemaphore` was already disposed when adding new readers. In order to fix the issue I used the logic that you described in the [above comment](https://github.com/nhibernate/nhibernate-core/pull/2147/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.yml#r403746842).

src/NHibernate/Util/AsyncReaderWriterLock.cs
maca88: > If think it can be avoided, but I am not sure it is worth it.

It is worth it as it prevents from reusing the old waiting reader lock.

src/NHibernate/Util/AsyncReaderWriterLock.cs
fredericDelaporte: I think it is possible to reprod the issue by using thread starvation:

 - Start a writer task, lock the ReaderWriterLock with it.
 - Ensure there are more asynchronously waiting readers on the ReaderWriterLock than available threads (30 readers should be enough for that).
 - Before ending the first write blocking them, start a thread for a second write which will synchronously wait (to ensure it keeps its thread for running immediately once released) with a dedicated test semaphore before locking the ReaderWriterLock.
 - Also start some other "reader" threads which will synchronously wait with a second dedicated test semaphore before locking the ReaderWriterLock.
 - Release the first write lock. At that point, the previously readers task are all released by the semaphore, but they cannot run all due to lack of threads. Have a third test lock in their task for avoiding them to finish, and so ensuring the thread starvation remains.
 - Release the second write thread with the first test semaphore. Since it was synchronously waiting, it will still have its thread and will be able to enter the ReaderWriterLock for locking it.
 - Release the "next reader" threads with the second test semaphore. Same as for the write, they have their thread available since they were synchronously waiting. Here we have a concurrency risk for the test: maybe they will be able to reach the ReaderWriterLock and enter it before the second write. Maybe a crude Thread.Sleep before releasing them would be enough to ensure this does not happen.
 - Normally the bug should happen, those "next readers" will run through the lock despite the second writer having request the lock.
 - But also at that point, likely after another crude Thread.Sleep to ensure the "next readers" had time to reach the lock, the third test lock should be released to allow the test to finish by allowing the first reader to finish.

src/NHibernate/Util/AsyncReaderWriterLock.cs
```diff
@@ -182,19 +189,15 @@ private void ExitReadLock()
...
-				{
-					return;
-				}
-
-				_waitingReadLockSemaphore.Dispose();
-				_waitingReadLockSemaphore = null;
+				return;
 			}
+
+			_waitingDisposalSemaphore.Dispose();
```
fredericDelaporte, line 199: Using `_waitingDisposalSemaphore?.Dispose();` would be safer in principle (since we are in a threaded class), and simplifies the code by removing the `if`.

src/NHibernate/Util/AsyncReaderWriterLock.cs
maca88: Thanks for the instructions, but unfortunately I still wasn't able to create a failing test for the old version. The test that was added in the last commit reproduces the issue of reusing the `_waitingReadLockSemaphore` instance for the next readers, but due to `SemaphoreSlim.Release` always releasing the first readers the issue didn't shown up. With the old version `_waitingReadLockSemaphore` is used for the first and the next readers and disposed only after the next readers are finished reading.

src/NHibernate/Cache/UpdateTimestampsCache.cs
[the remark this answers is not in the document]
maca88: Added.

src/NHibernate/Cache/UpdateTimestampsCache.cs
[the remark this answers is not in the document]
maca88: Added.

src/NHibernate/Util/AsyncReaderWriterLock.cs
```diff
@@ -0,0 +1,232 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NHibernate.Util
+{
+	// Idea from:
+	// https://github.com/kpreisser/AsyncReaderWriterLockSlim
+	// https://devblogs.microsoft.com/pfxteam/building-async-coordination-primitives-part-7-asyncreaderwriterlock/
+	internal class AsyncReaderWriterLock
```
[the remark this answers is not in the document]
maca88: The way `SemaphoreSlim` is currently implemented it is not needed as we do not use `AvailableWaitHandle` property. As its implementation may change in the future I've made it disposable.

src/NHibernate/Util/AsyncReaderWriterLock.cs
maca88: True, fixed.

src/NHibernate/Util/AsyncReaderWriterLock.cs
```diff
@@ -191,13 +208,16 @@ private void ExitReadLock()
...
+			_waitingDisposalSemaphore?.Dispose();
+			_waitingDisposalSemaphore = null;
+		}
+
+		private void AssertNotDisposed()
+		{
+			if (_disposed)
 			{
-				return;
+				throw new InvalidOperationException("The instance is disposed.");
```
fredericDelaporte, line 219: Why not `throw new ObjectDisposedException(nameof(AsyncReaderWriterLock));`?

maca88: My bad, corrected.

PR #2149, opened 2019-05-05
Fix quoting <key /> columns containing reserved words

lillo42:
Fixes: https://github.com/nhibernate/nhibernate-core/issues/1100


src/NHibernate/Tool/hbm2ddl/SchemaMetadataUpdater.cs
```diff
@@ -58,9 +58,21 @@ public static void QuoteTableAndColumns(Configuration configuration, Dialect.Dia
...
 				QuoteTable(cm.Table, dialect);
 			}
+
 			foreach (var cm in configuration.CollectionMappings)
 			{
 				QuoteTable(cm.Table, dialect);
+
+				if(cm.Key?.ColumnIterator != null)
+				{	
+					foreach (Column column  in cm.Key.ColumnIterator)
```
bahusoid: I think it's better to extract it to `void QuoteColumns(IEnumerable<ISelectable> columns)` and reuse in `QuoteTable`. What about the code above? Shouldn't it also be done to for `ClassMapping.Key.ColumnIterator`?

bahusoid: And  are you sure it's  safe 'ISelectable` -> 'Column` conversions? It seems better to call `OfType<Column>()`  here

lillo42: I copy this part from:

https://github.com/nhibernate/nhibernate-core/blob/21975f6137738633fc6d99951360606e8c2298fc/src/NHibernate/Persister/Collection/AbstractCollectionPersister.cs#L233-L238


But I believe it's better user ```OfType<Column>()```

lillo42: Method extracted to ```void QuoteColumns(IEnumerable<Column> columns)``` a reuse in ```QuoteTable``` and now I'm using with ```ClassMapping.Key.ColumnIterator```

src/NHibernate/Tool/hbm2ddl/SchemaMetadataUpdater.cs
```diff
@@ -56,27 +57,45 @@ public static void QuoteTableAndColumns(Configuration configuration, Dialect.Dia
...
+		}
+
+		private static void QuoteTable(Table table, IEnumerable<Column> columns, Dialect.Dialect dialect)
 		{
 			if (!table.IsQuoted && dialect.IsKeyword(table.Name))
 			{
 				table.IsQuoted = true;
 			}
+
 			foreach (var column in table.ColumnIterator)
```
bahusoid: I meant to reuse this part as `QuoteColumns(table.ColumnIterator, dialect)`. I don't think that additonal `columns` parameter is really required here.

And have you actually checked if `ClassMapping.Key.ColumnIterator` needs explicit processing? It was just a question. I'm not sure why `table.Columns` doesn't include those columns. If it really doesn't - what about others "columns holders" properties like `classMapping.Identifier.ColumnIterator` or `collectionMapping.Element.ColumnIterator` or ...?

src/NHibernate/Tool/hbm2ddl/SchemaMetadataUpdater.cs
```diff
@@ -56,27 +57,45 @@ public static void QuoteTableAndColumns(Configuration configuration, Dialect.Dia
...
 			foreach (var cm in configuration.ClassMappings)
 			{
-				QuoteTable(cm.Table, dialect);
+				QuoteTable(cm.Table,cm.Key?.ColumnIterator?.OfType<Column>(), dialect);
 			}
+
 			foreach (var cm in configuration.CollectionMappings)
 			{
-				QuoteTable(cm.Table, dialect);
+				QuoteTable(cm.Table, cm.Key?.ColumnIterator?.OfType<Column>(), dialect);
```
bahusoid: What about `cm.CollectionTable`? Can you check if we have those columne there. Maybe whole fix should be as simple as:
```suggestion
				QuoteTable(cm.Table, dialect);
				QuoteTable(cm.CollectionTable, dialect);
```

hazzik: I've checked. It doesn't work.

bahusoid: Yeah I see different column instances in `cm,Keys` and inside `cm.CollectionTable`. And it seems reusing existing table columns would fix this issue without any additional changes. But not sure if it's a good idea. Maybe not.

src/NHibernate/Tool/hbm2ddl/SchemaMetadataUpdater.cs
bahusoid: >And have you actually checked if ClassMapping.Key.ColumnIterator needs explicit processing?

I've made a quick look - we don't need to touch any `ClassMapping` properties. They are all present in `table.ColumnIterator`.

Such changes only required for collection. So to existing `cm.Key.ColumnIterator` please also add call for `cm.Element.ColumnIterator`

hazzik:
hi @lillo42, can you please reset your master branch to be on par with the upstream? It's a little bit annoying to see all these merge commits for fresh PRs. Thanks in advance.

PR #2158, opened 2019-05-15
 Proper support for IN clause for composite values in Criteria

bahusoid:
Ported latest version of hibernate [InExpression](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/criterion/InExpression.java). As NHibernate version generates completely incorrect SQL for composite keys.


src/NHibernate/Criterion/InExpression.cs
```diff
@@ -62,41 +58,50 @@ public override SqlString ToSqlString(ICriteria criteria, ICriteriaQuery criteri
 				return new SqlString("1=0");
 			}
 
-			//TODO: add default capacity
-			SqlStringBuilder result = new SqlStringBuilder();
-			SqlString[] columnNames =
-				CriterionUtil.GetColumnNames(_propertyName, _projection, criteriaQuery, criteria);
-			
-			// Generate SqlString of the form:
-			// columnName1 in (values) and columnName2 in (values) and ...
```
bahusoid: This is wrong  as instead of  requested (1, 1), (2, 2), (3, 3) values it will find all combinations of those values as (1,2), (1,3), (2,1)...

src/NHibernate/Criterion/InExpression.cs
```diff
@@ -62,41 +58,50 @@ public override SqlString ToSqlString(ICriteria criteria, ICriteriaQuery criteri
...
+				var parens = columns.Length > 1 ? new[] {new SqlString("("), new SqlString(")"),} : null;
+				SqlString comaSeparator = new SqlString(", ");
+				var singleValueParam = SqlStringHelper.Repeat(new SqlString(bogusParam), columns.Length, comaSeparator, parens);
+
+				var parameters = SqlStringHelper.Repeat(singleValueParam, Values.Length, comaSeparator, null);
+
+				//single column: col1 in (?, ?)
+				//multi column:  (col1, col2) in ((?, ?), (?, ?))
+				return new SqlString(
+					parens?[0] ?? SqlString.Empty,
```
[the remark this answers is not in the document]
bahusoid: `parens` is either array containing open and closed parens `new[]{"(", ")"}` (only it's `SqlString` not string) or null if no parens are required.
So this is equivalent of:
```C#
bool needWrapInParens = columns.Length > 1;
return new SqlString(
	needWrapInParens ? new SqlString("(") : SqlString.Empty, //instead of parens?[0] ?? SqlString.Empty,
...
	needWrapInParens ? new SqlString(")") : SqlString.Empty,//instead of parens?[0] ?? SqlString.Empty,
```

Only those `parens` are also reused in `SqlStringHelper.Repeat`

hazzik:
@bahusoid I think you do not gain much from substituting "bogus" parameters because you have more string manipulations. I think it would be clearer (and probably more efficient) to split this code into 3 branches:
1. One column: super simple `col in (@...)` statement.
2. Multiple columns:
2a. When dialect supports multi-column IN statement.
2b. When dialect does not support multi-column IN statement.

bahusoid:
> I think you do not gain much from substituting "bogus" parameters

I'm not trying to gain anything. I need to split SQL template generation and parameters generation. This 
is due to NHibernate specific parameter backtrack functionality so you have to provide parameters with typed value specified right away `sqlString.Add(criteriaQuery.NewQueryParameter(typedValue)` .

This `bogus` parameter allows me to separate template generation and parameters generation routines.

bahusoid:
> I think it would be clearer (and probably more efficient) to split this code into 3 branches:

I need to use functions like `SqlStringHelper.Repeat` and `SqlStringHelper.Join` to effectively generate SQL and "bogus" parameter allows it. Otherwise as I explained above I need to inject  real parameters right in to generation process.

@hazzik  So please clarify how suggested splitting allows me to effectively generate SQL without `bogus` parameter.

src/NHibernate/SqlCommand/SqlString.cs
```diff
@@ -1024,6 +1024,19 @@ public SqlString GetSubselectString()
 			return new SubselectClauseExtractor(this).GetSqlString();
 		}
 
+		internal void SubstituteBogusParameters(IReadOnlyList<Parameter> actualParams, Parameter bogusParam)
+		{
+			int index = 0;
+			var keys = _parameters.Keys;
+			// ReSharper disable once ForCanBeConvertedToForeach
+			for (var i = 0; i < keys.Count; i++)
```
fredericDelaporte, line 1032: Why not using a `foreach` here?

bahusoid: Because of modifications in collection

fredericDelaporte: The code changes the value associated to the key, but it does not change the keys list, since it is only affecting existing keys.
Unfortunately the `SortedList` implementation, though not changing the underlying keys collection at all in this case, has a single "version" check logic for throwing on all iterations in case of any kind of modification.
So ok.

src/NHibernate/SqlCommand/SqlString.cs
```diff
@@ -1024,6 +1024,19 @@ public SqlString GetSubselectString()
 			return new SubselectClauseExtractor(this).GetSqlString();
 		}
 
+		internal void SubstituteBogusParameters(IReadOnlyList<Parameter> actualParams, Parameter bogusParam)
+		{
+			int index = 0;
+			var keys = _parameters.Keys;
+			// ReSharper disable once ForCanBeConvertedToForeach
```
fredericDelaporte, line 1031:
```suggestion
			// The loop below is technically not altering the keys collection on which we iterate, but
			// the underlying implementation still throws on foreach iterations over keys even if we
			// have only changed the associated value.
			// ReSharper disable once ForCanBeConvertedToForeach
```

hazzik:
I was reluctant to approve it because there were code changes which I think could be improved. I tried to do that, but never got time to finish.

fredericDelaporte:
You may have seen I usually do not merge directly after approval excepted for trivial changes, but usually do it the next day or later. We could decide to wait a bit more by default after an approval, to let time for others to react if needed. Like 2 days maybe.

bahusoid:
> were code changes which I think could be improved

I believe it's about `SubstituteBogusParameters` part. Well - it's all internal stuff - you can improve it anytime later. If you would be able to keep logic as close as possible to hibernate without this substitution stuff - I would love to see it.  And full logic rewrite just to avoid it - IMO not a good idea.

PR #2159, opened 2019-05-16
Unify handling of composite values in hql and Criteria

### LINKED ISSUE #1376: Composite id is incorrectly expanded in SQL
We have a table with a composite primary key mapped, as shown below:

`mapping.CompositeId().KeyReference(e => e.Node).KeyProperty(e => e.DtFr).UnsavedValue("any");`

We need to delete several rows from the table satisfying `n.Node.Contract.Code = "1234"` condition. We  use next HQL:

`delete from Entity n where n in (select c from Entity c where c.Node.Contract.Code = "1234")`

But in the generated SQL-query there are no parentheses around the primary key columns:

`select ... from ENTITY n where n.IDNODE, n.DTFR in (select c.IDNODE, c.DTFR ...`

We expect that n.IDNODE, n.DTFR columns should be surrounded with parentheses:

`select ... from ENTITY n where (n.IDNODE, n.DTFR) in (select c.IDNODE, c.DTFR ...`

Version of NHibernate is 4.1.1.4000.

bahusoid:
Generate composite values as row value constructor  `(col1, col2)` instead of failing or using only first column. That's how it's already handled [here](https://github.com/nhibernate/nhibernate-core/blob/b734d004ed7eb18198c11a40d9c9decce537ace8/src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs#L661-L670) for properties in hql. So let's  unify this logic.

Among other things fixes #1376


src/NHibernate.Test/CompositeId/ClassWithCompositeIdFixture.cs
```diff
@@ -228,5 +228,110 @@ public void Hql()
 				Assert.AreEqual(1, results.Count);
 			}
 		}
+
+		[Test]
+		public void HqlInClause()
+		{
+			//Need to port changes from InLogicOperatorNode.mutateRowValueConstructorSyntaxInInListSyntax
+			if (!Dialect.SupportsRowValueConstructorSyntaxInInList)
```
bahusoid: While this functionality is working for dialects where row value constructor is supported - this test is currently ignored for all dialects because `SupportsRowValueConstructorSyntaxInInList` is false for all dialects.  `SupportsRowValueConstructorSyntaxInInList` is set in my other earlier PR #2158

hazzik: Ok

PR #2160, opened 2019-05-17
Fix Order By for composite property projection in Criteria

### LINKED ISSUE #1103: NH-2926 - CriteriaQuery - Unable to sort by composite-id
**stanislav merovic** created an [issue](https://nhibernate.jira.com/browse/NH-2926) — <time datetime="2011-11-02T06:42:53.279Z" title="2011-11-02T06:42:53.279Z">2nd November 2011, 6:42:53</time>:
> When trying to add a order for a composite-id using criteria queries the alias is removed only for the last property and a invalid query is created.
> 
> Cause: Order.ToSqlString -> NHibernate.Util.StringHelper.RemoveAsAliasesFromSql(produced); -> sql.Substring(0, sql.LastIndexOfCaseInsensitive(" as "));
> 
> 
> 
---
**stanislav merovic** added a comment — <time datetime="2011-11-02T07:08:28.882Z" title="2011-11-02T07:08:28.882Z">2nd November 2011, 7:08:28</time>:
> Forgot to mention: This happens when using the identifier projection.
> 
> crit.AddOrder(Order.Asc(Projections.Id))
>

bahusoid:
Fixes #1103


hazzik:
I have another fix in mind. I tried to implement it a while back. The idea is to get rid of `StringHelper.RemoveAsAliasesFromSql` and instead generate projections without aliases when needed.

bahusoid:
@hazzik Yeah I've been thinking about such fix myself. It will allow to use multicolumn projections (not only `PropertyProjection`) in different scenarios. But it's not fully applicable to this scenario - `Order` class is special and it expects final aliases and not column names (see it calls method `GetColumnAliasesUsingProjection`).

And anyway as long as we have `IPropertyProjection` -  `IdentifierProjection` should be treated as such. So this fix is still applicable. It also removes one more usage of `RemoveAsAliasesFromSql` in `Order` so less to code with fix that removes it.

src/NHibernate/Criterion/CriterionUtil.cs
```diff
@@ -46,6 +46,11 @@ public static class CriterionUtil
 
 		internal static SqlString[] GetColumnNamesUsingProjection(IProjection projection, ICriteriaQuery criteriaQuery, ICriteria criteria)
 		{
+			if (projection is IPropertyProjection propertyProjection)
```
bahusoid: This change is actually unrelated to fix. But it fixes cases when `GetColumnNamesUsingProjection` is called for composite property projections (in case of any in the future)

bahusoid:
@hazzik I've edited recently [my answer ](https://github.com/nhibernate/nhibernate-core/pull/2160#issuecomment-511095322) to make it more clear.

PR #2161, opened 2019-05-18
Optimize ToArray conversions

bahusoid:
Optimized `collection.Select().ToArray()` calls with the following [extension](https://github.com/bahusoid/nhibernate-core/blob/2e6b9b8b333c59a167aba2bff82ed359162d68ad/src/NHibernate/Util/EnumerableExtensions.cs#L42-L53)


hazzik:
Hi, @bahusoid. Nice optimization. I actually planed this for a few thousand years (figurally) now. I made a synthetic [benchmark](https://gist.github.com/hazzik/13e6c5be40d3e8424136c0315b9c27fb) for this. 

A few take aways from the benchmark: 

- The boxing of `this` to `ICollection<T>` kills the intention. Better to use framework's `Array.ConvertAll` / `List<T>.ConvertAll`. The manual implementation is actually slower than both `Select(x=>...).ToArray()` and `Array.ConvertAll(...)`
- Method groups are slow.

bahusoid:
> The manual implementation is actually slower than both

It seems that's true only for .NET Core. On full .NET framework `Select(x=>...).ToArray()` is slower

hazzik:
I've updated benchmark. The boxing to `ICollection<>` is the culprit. I did not run on the Full .NET as I don't have access to a windows machine atm. You can run it for yourself. You'll need to add `[ClrJob]` attribute to the class.

hazzik:
Ok, I've run on full .NET. The version with boxing is indeed faster than Select (about 25% faster), however, it is still 3 times slower than version without boxing.

src/NHibernate/SqlCommand/SqlInsertBuilder.cs
```diff
@@ -221,7 +220,7 @@ public SqlCommandInfo ToSqlCommandInfo()
 
 		public SqlType[] GetParametersTypeArray()
 		{
-			return (new List<SqlType>(new SafetyEnumerable<SqlType>(columns.Values))).ToArray();
```
hazzik: There is a subtle difference between `SafetyEnumerable` and `OfType`. `SafetyEnumerable` will not filter out nulls, while `OfType` will. Not sure how this is relevant here.

bahusoid: Good catch.. Better to restore original logic

hazzik:
I propose to rename these extensions to `ConvertAll` to be on par with the framework. What do you think?

hazzik:
Or maybe not.

bahusoid:
>I propose to rename these extensions to ConvertAll

I'm not sure it's a good fit here. `ConvertAll` name doesn't imply that returning type might be changed.

PR #2166, opened 2019-05-28
Optimize usages of SqlString.Append

bahusoid:
Each `SqlString.Append` call leads to new `SqlString` creation. And called `SqlString` [ctor ](https://github.com/nhibernate/nhibernate-core/blob/f13368cf30fe016edad4846ce1d2bd5b5457141d/src/NHibernate/SqlCommand/SqlString.cs#L202-L218) doesn't look lightweight to me. So let's minimize subsequent Append calls.


src/NHibernate/Dialect/PostgreSQLDialect.cs
```diff
@@ -177,7 +177,7 @@ public override string GetDropSequenceString(string sequenceName)
 
 		public override SqlString AddIdentifierOutParameterToInsert(SqlString insertString, string identifierColumnName, string parameterName)
 		{
-			return insertString.Append(" returning ").Append(identifierColumnName);
+			return insertString.Append(" returning " +  identifierColumnName);
```
hazzik: Probably should be `,` instead of `+`

src/NHibernate/Dialect/PostgreSQL81Dialect.cs
```diff
@@ -111,7 +111,7 @@ public override SqlString AppendIdentitySelectToInsert(SqlString insertSql)
 
 		public override SqlString AppendIdentitySelectToInsert(SqlString insertString, string identifierColumnName)
 		{
-			return insertString.Append(" returning ").Append(identifierColumnName);
+			return insertString.Append(" returning " + identifierColumnName);
```
hazzik: Probably should be `,` instead of `+`

src/NHibernate/Hql/Ast/ANTLR/Tree/EntityJoinJoinSequenceImpl.cs
```diff
@@ -30,12 +30,14 @@ public EntityJoinJoinSequenceImpl(ISessionFactoryImplementor factory, EntityType
...
-				on.Append(" and ").Append(filters);
-			}
+			var on = withClauseFragment ?? SqlString.Empty;
+			//Note: filters logic commented as GetOnCondition always returns empty string for entity join (as IsReferenceToPrimaryKey is always true)
+			//TODO: If filters for entity join really make sense we need to inline GetOnCondition part that retrieves filters
+//			var filters = _entityType.GetOnCondition(_tableAlias, Factory, enabledFilters);
+//			if (!string.IsNullOrEmpty(filters))
+//			{
+//				on = on.Append(filters);
+//			}
```
hazzik: Unrelated change, should be reverted.

bahusoid: Original code is non functional and `SqlString.Append` related:
```C#
if (!string.IsNullOrEmpty(filters))
{
	on.Append(" and ").Append(filters);
}
```

as `SqlString` is immutable all those appends make no sense. So I figure better to comment non functional code with some details how to fix it in the future.

src/NHibernate/Dialect/PostgreSQL81Dialect.cs
bahusoid: Ok. Done

src/NHibernate/Hql/Ast/ANTLR/Tree/EntityJoinJoinSequenceImpl.cs
ggeurts: I agree that this change should be reverted to keep changes purely focused on SqlString usage changes.

src/NHibernate/SqlCommand/SqlString.cs
```diff
@@ -347,7 +347,23 @@ public SqlString Append(string text)
 		{
 			if (string.IsNullOrEmpty(text)) return this;
 			if (_length == 0) return new SqlString(text);
-			return new SqlString(new object[] { this, text });
+			return new SqlString(this, text);
+		}
+
+		public SqlString Append(params object[] parts)
```
ggeurts: I would consider introducing a new `private SqlString(SqlString other, params object[] parts)` constructor which can optimise the parts copying and append operations.

bahusoid: Actually I've considered such ctor. But looking at existing ctors `SqlString(SqlString x)` and `SqlString(params object[] parts)` - I didn't find a way to merge them nicely. So IMHO it's not worth it. But if you've come up with implementation  - feel free to add it.

src/NHibernate/Hql/Ast/ANTLR/Tree/EntityJoinJoinSequenceImpl.cs
bahusoid: As I already said it's exactly only about `SqlString.Append`. Though yeah added comment does look like something unrelated - I will adjust comment to make it clear.

PR #2170, opened 2019-06-09
Fixes NH 2229

lillo42:
Fixes https://github.com/nhibernate/nhibernate-core/issues/994


hazzik:
Hi @lillo42, could you please reset your master branch to the current master to avoid these merge commits in the PR? Or, alternatively, next time you create a PR - please make sure that you branch off the our master. Thanks.

src/NHibernate/Action/CollectionRecreateAction.cs
```diff
@@ -67,15 +67,32 @@ private void PreRecreate()
 
 		public override void ExecuteAfterTransactionCompletion(bool success)
 		{
-			base.ExecuteAfterTransactionCompletion(success);
+			var cacheKey = new CacheKey(GetKey(), Persister.KeyType, Persister.Role, Session.Factory);
 			
-			if (success)
+			base.ExecuteAfterTransactionCompletion(success);
+			if (success && Persister.HasCache)
```
hazzik: `Persister.HasCache` would always be true, because otherwise the `ExecuteAfterTransactionCompletion` is not called.

src/NHibernate/Cache/Entry/CollectionCacheEntry.cs
```diff
@@ -31,6 +31,24 @@ public static CollectionCacheEntry Create(IPersistentCollection collection, ICol
...
+			try
+			{
+				entry = new CollectionCacheEntry
+				{
+					state = collection.Disassemble(persister)
+				};
+
+				return true;
+			}
+			catch
```
hazzik: Having a catch-all here is really bad.

Ideally we should not have a catch here at all.

PR #2171, opened 2019-06-11
Upgrade NUnit

hazzik:


src/NHibernate.Test/NHSpecificTest/GH1963/Fixture.cs
```diff
@@ -50,7 +45,7 @@ where e.Flag
 							 select e;
 
 				Assert.That(
-					result.ToList,
+					() => result.ToList(),
```
hazzik: This was throwing exception that delegate is no valid. Changed this from a method group to a delegate and it triggered an async test generation.

fredericDelaporte:
The showBuildMenu.bat has been forgotten, it does still (try) to use NUnit.ConsoleRunner.3.7.0.

Fixing it to use 3.10.0 is not enough, many tests fail as if the test dll was no more able to find its configuration file.

fredericDelaporte:
This is an issue in NUnit since v3.8: nunit/nunit-console#418.
The build menu runs the test through `NHibernate.nunit` project file, which uses the Multiple processModel option with which loading .config files does no more work.
The teamcity build runs test dll directly as separated commands, without using that option.

PR #2180, opened 2019-06-14
Upgrade AsyncGenerator to 0.8.2.11 for 5.2

bahusoid:
To make it work with VS 2019


src/NHibernate.Test/Async/SessionBuilder/Fixture.cs
```diff
@@ -189,51 +189,6 @@ public async Task CanSetConnectionAsync()
 				FlushMode.Always, FlushMode.Auto, FlushMode.Commit, FlushMode.Manual);
 		}
 
-		[Test]
-		public async Task CanSetInterceptorAsync()
```
bahusoid: Removed as it's not async (see https://github.com/maca88/AsyncGenerator/issues/116)

PR #2181, opened 2019-06-17
Skip null entities when bag is populated

bahusoid:
That's the way it's done in hibernate so I see no good reasons why we do it differently. And also null elements break filtering for many-to-many collections.

### DIFF
--- a/src/NHibernate.Test/Async/NHSpecificTest/GH1994/Fixture.cs
+++ b/src/NHibernate.Test/Async/NHSpecificTest/GH1994/Fixture.cs
@@ -117,6 +117,23 @@ namespace NHibernate.Test.NHSpecificTest.GH1994
 			}
 		}
 
+		[Test]
+		public async Task TestFilteredBagQueryOverAsync()
+		{
+			using (var s = OpenSession())
+			{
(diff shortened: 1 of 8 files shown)


hazzik:
We were discussing the behaviour of "not-found" somewhere, but I'm not sure where. Do you remember?

hazzik:
Found it: https://github.com/nhibernate/nhibernate-core/pull/1995#issuecomment-460052660

bahusoid:
Yeah. The main concern about this change made Frederic about `not-found="ignore"` mapping:
>I agree it does not look very useful to have those null elements in the collection. But we have to be careful about the flush behavior. Not having them could only mean they would be deleted from the database. We have to check how it behaves currently. If currently, these not-found elements are left untouched by flushes (unless explicitly removed from the collection by the application), changing this could be a serious breaking change for some applications.

But such elements can't be deleted anyway - NHibernate doesn't keep identifiers for them. So null or missing element won't make a difference for flush. I will add test to verify it.

bahusoid:
I've added tests.. Turns out for many-to-many collections all broken records are purged anyway on flush if collection is changed. Because collection is fully removed and then recreated. So I've run those tests on master (with adjustments to null presence) and this branch and have verified that flushing behavior is not changed with this PR.

PR #2182, opened 2019-06-18
Upgrade AsyncGenerator to 0.17.1

maca88:
The version contains the fix for the issue discovered in #2056.

It also adds .Net Core support, and this PR switches the generation target framework to netcoreapp for better cross-platform development support.


src/AsyncGenerator.yml
```diff
@@ -3,6 +3,8 @@
   targetFramework: net461
   concurrentRun: true
   applyChanges: true
+  suppressDiagnosticFailures:
+  - pattern: ^.*(Hql\.g).*$
```
maca88: The new version of Roslyn reports warnings for Hql.g, which have to be omitted.

src/NHibernate.Test/Async/NHSpecificTest/GH1547/Fixture.cs
```diff
@@ -103,12 +103,12 @@ private async Task BenchmarkAsync<T>(string test, Func<ISession, IQueryable<T>>
 				{
 					var query = queryFactory(session);
 					// Warm up.
-					await (RunBenchmarkUnitAsync(cache, query));
+					await (RunBenchmarkUnitAsync(cache, query, cancellationToken));
```
maca88: This change is due maca88/AsyncGenerator/issues/126

default.build
```diff
@@ -65,7 +65,7 @@
 	<target name="find-async-generator-console">
 		<property name="nuget-package-id" value="CSharpAsyncGenerator.CommandLine" />
 		<call target="get-nuget-package-path" />
-		<property name="async-generator-console"  value="${nuget-package-path}/tools/AsyncGenerator.CommandLine.exe" />
+		<property name="async-generator-console"  value="${nuget-package-path}/tools/net461/AsyncGenerator.CommandLine.exe" />
```
maca88: `net461` is now deprecated as the new version of Roslyn supports only `net472` and `netcoreapp2.1`. Currently the generator does not support any C# 8 features yet, so there are no difference between them in terms of supported features.

bahusoid:
It seems .NET Core support is also added in this version. Shouldn't we start using it right away? ) It's my understanding it would allow to skip installing mono for development process on Unix.

maca88:
> Shouldn't we start using it right away?

Yes we could, the only reason I didn't use it is because the minimal .NET Core version is higher than what NHibernate has (`netcoreapp2.1` vs `netcoreapp2.0`). If you are fine with that I will change it to .NET Core.

> It's my understanding it would allow to skip installing mono for development process on Unix.

.NET Core would need to be installed instead.

bahusoid:
> because the minimal .NET Core version is higher than what NHibernate has (netcoreapp2.1 vs netcoreapp2.0). If you are fine with that I will change it to .NET Core.

I personally see no issues with that. But maybe I don't understand all the implications.

maca88:
The only implication is that a developer with .NET Core 2.0 installed would be able to build the project but not able to generate the async code. I've changed the nant build to use .NET Core, `ShowBuildMenu.sh` still needs to be changed, which I plan to do it tomorrow.

src/AsyncGenerator.yml
```diff
@@ -3,6 +3,8 @@
   targetFramework: net461
```
bahusoid: What this setting is about? Shouldn't it also be changed to .net core? If yes can it still be netcoreapp2.0?

I've noticed that compilation and debugging under .net core is a little bit faster - so my main desire is to avoid net461 compilation altogether in development process. So with .net core migration I'm hoping to get rid of the following AsyncGenerator exception:
>  [exec] Msbuild failed when processing the file 'E:\Git\nhibernate-core\src\NHibernate\NHibernate.csproj' with message: C:\Program Files\dotnet\sdk\2.1.602\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets: (198, 5): Assets file 'E:\Roman\Git\nhibernate-core\src\NHibernate\obj\project.assets.json' doesn't have a target for '.NETFramework,Version=v4.6.1'. Ensure that restore has run and that you have included 'net461' in the TargetFrameworks for your project.

So is this setting related to this error?

hazzik:
@maca88, I was thinking about having a dotnet global tool. Is it feasible?

hazzik:
Hi, I'm having issue on MacOS:

```
hazzik-mac:src hazzik$ dotnet ../Tools/CSharpAsyncGenerator.CommandLine.0.16.0/tools/netcoreapp2.1/AsyncGenerator.CommandLine.dll
21:59:58,517 INFO  CommandLine: AsyncGenerator

21:59:59,150 INFO  AsyncCodeGenerator: Generating async code started
21:59:59,559 INFO  AsyncCodeGenerator: Opening project '/Users/hazzik/Projects/nhibernate-core/src/NHibernate/NHibernate.csproj' started
22:00:02,021 FATAL CommandLine: System.InvalidOperationException: One or more errors occurred while opening the project:
Msbuild failed when processing the file '/Users/hazzik/Projects/nhibernate-core/src/NHibernate/NHibernate.csproj' with message: /usr/local/share/dotnet/sdk/3.0.100-preview4-011223/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.TargetingPackResolution.targets: (84, 5): The "ResolveFrameworkReferences" task failed unexpectedly.
System.IO.FileNotFoundException: Could not load file or assembly 'NuGet.Frameworks, Version=5.0.0.4, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.

File name: 'NuGet.Frameworks, Version=5.0.0.4, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
   at Microsoft.NET.Build.Tasks.ResolveFrameworkReferences.ExecuteCore()
   at Microsoft.NET.Build.Tasks.TaskBase.Execute() in /_/src/Tasks/Common/TaskBase.cs:line 35
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.ExecuteInstantiatedTask(ITaskExecutionHost taskExecutionHost, TaskLoggingContext taskLoggingContext, TaskHost taskHost, ItemBucket bucket, TaskExecutionMode howToExecuteTask)


Hint: For suppressing irrelevant errors use SuppressDiagnosticFailures option.
   at AsyncGenerator.AsyncCodeGenerator.CheckForErrors(MSBuildWorkspace workspace, String itemType, ImmutableArray`1 supressFailuresPredicates, ILogger logger)
   at AsyncGenerator.AsyncCodeGenerator.OpenProject(MSBuildWorkspace workspace, String filePath, ImmutableArray`1 supressFailuresPredicates, ILogger logger, CancellationToken cancellationToken)
   at AsyncGenerator.AsyncCodeGenerator.GenerateAsync(AsyncCodeConfiguration configuration, CancellationToken cancellationToken)
   at AsyncGenerator.CommandLine.Program.Main(String[] args)
```

hazzik:
These work fine:

```
mono ../Tools/CSharpAsyncGenerator.CommandLine.0.16.0/tools/net461/AsyncGenerator.CommandLine.exe
```
```
mono ../Tools/CSharpAsyncGenerator.CommandLine.0.16.0/tools/net472/AsyncGenerator.CommandLine.exe
```

hazzik:
ok, this seems to be an issue with the dotnet sdk version. I wonder how can we fix it?

hazzik:
Works for SDK 2.1.700.

Does not work for SDK 2.2.300:

```
hazzik-mac:src hazzik$ dotnet new globaljson --sdk-version 2.2.300 --force
The template "global.json file" was created successfully.
hazzik-mac:src hazzik$ dotnet ../Tools/CSharpAsyncGenerator.CommandLine.0.16.0/tools/netcoreapp2.1/AsyncGenerator.CommandLine.dll
22:22:57,813 INFO  CommandLine: AsyncGenerator

22:22:58,321 INFO  AsyncCodeGenerator: Generating async code started
22:22:58,689 INFO  AsyncCodeGenerator: Opening project '/Users/hazzik/Projects/nhibernate-core/src/NHibernate/NHibernate.csproj' started
22:22:59,978 FATAL CommandLine: System.InvalidOperationException: One or more errors occurred while opening the project:
Msbuild failed when processing the file '/Users/hazzik/Projects/nhibernate-core/src/NHibernate/NHibernate.csproj' with message: /usr/local/share/dotnet/sdk/2.2.300/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets: (208, 5): Error reading assets file: Error loading lock file '/Users/hazzik/Projects/nhibernate-core/src/NHibernate/obj/project.assets.json' : Could not load file or assembly 'Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'. The system cannot find the file specified.

Hint: For suppressing irrelevant errors use SuppressDiagnosticFailures option.
   at AsyncGenerator.AsyncCodeGenerator.CheckForErrors(MSBuildWorkspace workspace, String itemType, ImmutableArray`1 supressFailuresPredicates, ILogger logger)
   at AsyncGenerator.AsyncCodeGenerator.OpenProject(MSBuildWorkspace workspace, String filePath, ImmutableArray`1 supressFailuresPredicates, ILogger logger, CancellationToken cancellationToken)
   at AsyncGenerator.AsyncCodeGenerator.GenerateAsync(AsyncCodeConfiguration configuration, CancellationToken cancellationToken)
   at AsyncGenerator.CommandLine.Program.Main(String[] args)
```

src/AsyncGenerator.yml
maca88: > What this setting is about?  If yes can it still be netcoreapp2.0?

This setting set the target framework which is used to compile the project. When having multiple target frameworks it is required to set one for which the compilation will be done.

> Shouldn't it also be changed to .net core?

I forgot to change it, yes it should be changed to .net core.

>  If yes can it still be netcoreapp2.0?

It has to be one of the frameworks that the project has in the `<TargetFrameworks>` tag, so yes it can be `netcoreapp2.0`.

> So is this setting related to this error?

Yes it is related, you probably set in `NHibernate.dev.props` to only `netcoreapp2.0`.

maca88:
> I was thinking about having a dotnet global tool. Is it feasible?

I also thought about it but then I realized that it would not work as a specific version may be required for a repository (`NHibernate.Caches`) or branch (`5.2.x`) if installed as a global tool. Recently, I found out that in .NET Core 3 will be possible to have a tool installed locally ([source](https://www.dotnetcurry.com/dotnet/1495/dotnet-core-global-tools) search for "Upcoming changes in .NET Core 3.0"), which is ideal for this tool. I will check what has to be done and try to add support for it.

maca88:
> this seems to be an issue with the dotnet sdk version. I wonder how can we fix it?

In .NET Core 2.2 `Json.Newtonsoft` was updated from v9 to v10 and for some reason there is no redirect to the new version. I've made a fix by adding a redirect (maca88/AsyncGenerator/pull/128).

src/NHibernate/AdoNet/SqlClientBatchingBatcher.cs
```diff
@@ -70,6 +70,47 @@ public override void AddToBatch(IExpectation expectation)
 			}
 		}
 
+		public override Task AddToBatchAsync(IExpectation expectation, CancellationToken cancellationToken)
```
maca88: As the target framework was changed to `netcoreapp2.0`, `SqlClientBatchingBatcher` is not compiled anymore and we have to manually add the async methods.

maca88:
@hazzik can you please verify if the current code works on MacOS?

maca88:
The current code does not compile on `net461` and to resolve the problem maca88/AsyncGenerator/issues/129 must be resolved, which I plan to do it tomorrow.

ShowBuildMenu.sh
```diff
@@ -190,28 +190,7 @@ getAsyncGeneratorPath(){
...
-	if [ ! -f nuget.exe ]
-	then
-		wget https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
-	fi
-
-	async_generator_path="CSharpAsyncGenerator.CommandLine.$(cat packages.config | grep id=\"CSharpAsyncGenerator.CommandLine | cut -d\" -f4)/tools"
-
-	if [ ! -d $async_generator_path ]
-	then
-		mono nuget.exe install
```
hazzik: Still need `nuget install` part, I believe.

maca88: You are right, I reverted it. Unfortunately .NET Core does not support packages.config, so mono is still required here.

fredericDelaporte: Maybe should we work around this by replacing this package.config by a "package.csproj"?

maca88: I've added `packages.csproj` and used it only for the generator as I don't quite understand where other tools like `vswhere` are used.

Tools/packages.config
```diff
@@ -7,7 +7,6 @@
   <package id="NUnit.Extension.NUnitV2ResultWriter" version="3.6.0" targetFramework="net461" />
   <package id="NUnit.Extension.TeamCityEventListener" version="1.0.6" targetFramework="net461" />
   <package id="NUnit.Extension.VSProjectLoader" version="3.8.0" targetFramework="net461" />
-  <package id="CSharpAsyncGenerator.CommandLine" version="0.14.0" targetFramework="net461" />
   <package id="vswhere" version="2.1.4" targetFramework="net461" />
   <package id="GitReleaseManager" version="0.7.0" targetFramework="net461" />
```
fredericDelaporte: This one is referenced [here](https://github.com/nhibernate/nhibernate-core/blob/master/ReleaseProcedure.txt#L6):

> Create a draft release in Github with GitReleaseManager. If you have used the NHibernate build menu, it should be available in Tools\gitreleasemanage.x.x.x\

So if running the build menu does no more trigger a fetch of this package, this sentence above will no more hold true.

Tools/packages.config
```diff
@@ -7,7 +7,6 @@
   <package id="NUnit.Extension.NUnitV2ResultWriter" version="3.6.0" targetFramework="net461" />
   <package id="NUnit.Extension.TeamCityEventListener" version="1.0.6" targetFramework="net461" />
   <package id="NUnit.Extension.VSProjectLoader" version="3.8.0" targetFramework="net461" />
-  <package id="CSharpAsyncGenerator.CommandLine" version="0.14.0" targetFramework="net461" />
   <package id="vswhere" version="2.1.4" targetFramework="net461" />
```
fredericDelaporte: `vswhere` is used by [actual_msbuild.cmd](https://github.com/nhibernate/nhibernate-core/blob/master/Tools/actual_msbuild.cmd#L3), used for generating the documentation, see [here](https://github.com/nhibernate/nhibernate-core/blob/master/doc/documentation.build#L68).

Tools/packages.config
```diff
@@ -7,7 +7,6 @@
   <package id="NUnit.Extension.NUnitV2ResultWriter" version="3.6.0" targetFramework="net461" />
```
fredericDelaporte: NUnit ConsoleRunner is directly launched by the build menu, and I guess other NUnit packages are dependencies.

Tools/packages.config
maca88: Thanks for the references, I've moved it to `packages.csproj` and updated the release procedure.

Tools/packages.config
maca88: Thanks, now it is moved to `packages.csproj`.

Tools/packages.config
maca88: `NUnit.ConsoleRunner` does not have any dependencies, `NUnit.Console` has all those dependencies. As only `NUnit.ConsoleRunner` seems to be used I've only moved it into `packages.csproj`.

ShowBuildMenu.bat
```diff
@@ -1,12 +1,16 @@
...
+set NUNIT_VERSION=%NUNIT_VERSION:~57%
+set NUNIT_VERSION=%NUNIT_VERSION:" />=%
+
 set NANT="%~dp0Tools\nant\bin\NAnt.exe" -t:net-4.0
 set BUILD_TOOL_PATH=%~dp0Tools\BuildTool\bin\BuildTool.dll
 set BUILDTOOL=dotnet %BUILD_TOOL_PATH%
 set AVAILABLE_CONFIGURATIONS=%~dp0available-test-configurations
 set CURRENT_CONFIGURATION=%~dp0current-test-configuration
-set NUNIT="%~dp0Tools\NUnit.ConsoleRunner.3.7.0\tools\nunit3-console.exe"
+set NUNIT="%userprofile%\.nuget\packages\NUnit.ConsoleRunner\%NUNIT_VERSION%\tools\nunit3-console.exe"
```
bahusoid: Wouldn't it be better to keep those tools inside Tools folder?
I think it should be possible by adding "nuget.config" file to Tools folder with the following content:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <config>
    <add key="repositoryPath" value="./packages" />
  </config>
</configuration>
```

src/NHibernate.Test/Async/Tools/hbm2ddl/SchemaValidator/SchemaValidateFixture.cs
```diff
@@ -53,34 +53,6 @@ public async Task ShouldVerifySameTableAsync()
 		}
 
 #if NETFX
-		[Test, SetCulture("tr-TR"), SetUICulture("tr-TR")]
-		public async Task ShouldVerifySameTableTurkishAsync()
```
bahusoid: Could you please clarify why async code is not generated here? Is it some limitation of Roslyn  parser - so disabled code is not visible?

ShowBuildMenu.bat
maca88: `nuget.config` didn't work for `dotnet restore`, but it worked with `--packages` option. Now the packages are restored into Tools folder as before.

src/NHibernate.Test/Async/Tools/hbm2ddl/SchemaValidator/SchemaValidateFixture.cs
maca88: By changing the `targetFramework` to `netcoreapp2.0`, the parser treats directives like `NETFX` as disabled and no symbols and syntax nodes are produced for members inside those directives, which means that code is available only as a plain text.

bahusoid: So is it required for parser to have compilable code to produce such symbols/syntax nodes? I mean if for instance we supply NETFX directive under netcoreapp2.0 -  parser still won't be able to process it due to compilation errors?

Tools/packages.config
maca88: My bad, the dependencies are still needed otherwise the tests won't run (with error: `Unknown result format: nunit2`). Adding `NUnit.Console` as the dependency fixed the issue.

src/NHibernate.Test/Async/Tools/hbm2ddl/SchemaValidator/SchemaValidateFixture.cs
maca88: > So is it required for parser to have compilable code to produce such symbols/syntax nodes?

Yes, the generator uses symbols to find where the method is referenced, if the containing type has any async counterparts, if the method has any overrides, if implements any interfaces... Without all this information is not possible to determine what should be converted and what not.

> I mean if for instance we supply NETFX directive under netcoreapp2.0 - parser still won't be able to process it due to compilation errors?

In case when a type or member exists only in NETFX, the parser will create a syntax node for the type/member (as its a valid C# code), but it won't contain the information about the symbol, if target framework would be set to `netcoreapp2.0`, which mean that the code will be partially evaluated.
For this specific test, where the code is valid in both NETFX and NETCore, I've added a special attribute `NetFxOnlyAttribute` which will run only in NETFX, but the code be compiled in both frameworks, which will allow the generator to generate the async code when possible.

maca88:
I tried to re-generate this PR and #2205 on a machine with only VS 2017, but the mentioned class does not change. Does this change happen on each re-generation or it is random?

fredericDelaporte:
On each regeneration of this PR, even twice in a row. I have not checked the pending merge of 5.2.x. I have just tried again this evening, the trouble is still there. Maybe that is bound to the dotnet sdk. dotnet --version outputs 2.1.509 on my setup.
I do not have the trouble on master.

maca88:
That's strange, I retried with VS 2017 version `15.9.16` and dotnet sdk version `2.1.509`, but I am still unable to reproduce the issue. I also tried to clean the working directory with `git clean -d -x -f` command prior generation, but without success. Do you have `build-common\NHibernate.dev.props` file and if you have it, could you paste its content?

fredericDelaporte:
I do not have a `NHibernate.dev.props` file.

default.build
```diff
@@ -46,32 +46,32 @@
...
+		<property name="tool.path" value="${tools.dir}/${tool.id}/${tool.version}/" />
 	</target>
 
 	<target name="find-async-generator-console">
-		<property name="nuget-package-id" value="CSharpAsyncGenerator.CommandLine" />
-		<call target="get-nuget-package-path" />
-		<property name="async-generator-console"  value="${nuget-package-path}/tools/AsyncGenerator.CommandLine.exe" />
+		<property name="tool.id" value="CSharpAsyncGenerator.CommandLine" />
+		<call target="get-tool-info" />
+		<property name="async-generator-console"  value="${tool.path}/tools/netcoreapp2.1/AsyncGenerator.CommandLine.dll" />
```
fredericDelaporte: Small glitch here: `tool.path` always includes a final `/`, so adding one right after it causes a double `//`.

fredericDelaporte:
I have tried again after installing dotnet SDK 2.2.109, the trouble is still here. I have also added .Net Framework 4.8 dev-pack, trouble still here.

There is nothing significant in the logs:
```
solution-restore:

     [exec] Starting 'dotnet (msbuild /t:Restore ./src/NHibernate.sln)' in 'E:\Projets\nhibernate\nhibernate-core'
     [exec] Microsoft (R) Build Engine version 15.9.20+g88f5fadfbe pour .NET Core
     [exec] Copyright (C) Microsoft Corporation. Tous droits réservés.
     [exec]
     [exec]   Restauration effectuée dans 54,16 ms pour E:\Projets\nhibernate\nhibernate-core\src\NHibernate\NHibernate.csproj.
     [exec]   Restauration effectuée dans 54,17 ms pour E:\Projets\nhibernate\nhibernate-core\src\NHibernate.DomainModel\NHibernate.DomainModel.csproj.
     [exec]   Restauration effectuée dans 61,92 ms pour E:\Projets\nhibernate\nhibernate-core\src\NHibernate.TestDatabaseSetup\NHibernate.TestDatabaseSetup.csproj.
     [exec]   Restauration effectuée dans 61,92 ms pour E:\Projets\nhibernate\nhibernate-core\src\NHibernate.Test\NHibernate.Test.csproj.
     [exec]   Restauration effectuée dans 61,89 ms pour E:\Projets\nhibernate\nhibernate-core\src\NHibernate.Test.VisualBasic\NHibernate.Test.VisualBasic.vbproj.

find-async-generator-console:


common.init:


common.tools-restore:

     [exec] Starting 'dotnet (restore ./packages.csproj --packages .)' in 'E:\Projets\nhibernate\nhibernate-core\Tools'
     [exec]   Restauration effectuée dans 40,29 ms pour E:\Projets\nhibernate\nhibernate-core\Tools\packages.csproj.

get-tool-info:


generate-async:

     [exec] Starting 'dotnet (../Tools/CSharpAsyncGenerator.CommandLine/0.17.0//tools/netcoreapp2.1/AsyncGenerator.CommandLine.dll)' in 'E:\Projets\nhibernate\nhibernate-core\src'
     [exec] 22:30:40,607 INFO  CommandLine: AsyncGenerator
     [exec]
     [exec] 22:30:41,087 INFO  AsyncCodeGenerator: Generating async code started
     [exec] 22:30:41,402 INFO  AsyncCodeGenerator: Opening project 'E:\Projets\nhibernate\nhibernate-core\src\NHibernate\NHibernate.csproj' started
     [exec] 22:30:42,614 INFO  AsyncCodeGenerator: Opening project 'E:\Projets\nhibernate\nhibernate-core\src\NHibernate\NHibernate.csproj' completed
     [exec] 22:30:42,614 INFO  AsyncCodeGenerator: Configuring project prior analyzation started
     [exec] 22:30:42,641 INFO  AsyncCodeGenerator: Configuring project prior analyzation completed
     [exec] 22:30:42,654 INFO  AsyncCodeGenerator: Compiling project 'NHibernate' started
     [exec] 22:30:44,612 INFO  AsyncCodeGenerator: Compiling project 'NHibernate' completed
     [exec] 22:30:44,612 INFO  AsyncCodeGenerator: Initializing registered plugins for project 'NHibernate' started
     [exec] 22:30:45,187 INFO  AsyncCodeGenerator: Initializing registered plugins for project 'NHibernate' completed
     [exec] 22:30:45,187 INFO  AsyncCodeGenerator: Analyzing project 'NHibernate' started
     [exec] 22:30:45,204 INFO  ProjectAnalyzer: Parsing documents started
     [exec] 22:30:45,274 INFO  ProjectAnalyzer: Parsing documents completed
     [exec] 22:30:45,274 INFO  ProjectAnalyzer: Pre-analyzing documents started
     [exec] 22:30:46,839 INFO  ProjectAnalyzer: Pre-analyzing documents completed
     [exec] 22:30:46,839 INFO  ProjectAnalyzer: Scanning references started
     [exec] 22:30:53,265 INFO  ProjectAnalyzer: Scanning references completed
     [exec] 22:30:53,266 INFO  ProjectAnalyzer: Analyzing documents started
     [exec] 22:30:53,350 INFO  ProjectAnalyzer: Analyzing documents completed
     [exec] 22:30:53,350 INFO  ProjectAnalyzer: Post-analyzing documents started
     [exec] 22:30:58,593 INFO  ProjectAnalyzer: Post-analyzing documents completed
     [exec] 22:30:58,593 INFO  ProjectAnalyzer: Diagnose of documents started
     [exec] 22:30:58,625 WARN  Diagnostics: Diagnostics for document: Collection\Generic\PersistentGenericSet.cs
     [exec] 22:30:58,625 WARN  Diagnostics: Method Add (306,3)-(326,4) ignored reason: Implements an external interface System.Collections.Generic.ISet<T>.Add(T) that has not an async counterpart
     [exec] 22:30:58,625 WARN  Diagnostics: Method reference Initialize (311,5)-(311,15): Has an async counterpart that could't been used
     [exec] 22:30:58,625 WARN  Diagnostics: Method UnionWith (328,3)-(343,4) ignored reason: Implements an external interface System.Collections.Generic.ISet<T>.UnionWith(System.Collections.Generic.IEnumerable<T>) that has not an async counterpart
     [exec] 22:30:58,625 WARN  Diagnostics: Method reference Initialize (334,4)-(334,14): Has an async counterpart that could't been used
     [exec] 22:30:58,625 WARN  Diagnostics: Method IntersectWith (345,3)-(356,4) ignored reason: Implements an external interface System.Collections.Generic.ISet<T>.IntersectWith(System.Collections.Generic.IEnumerable<T>) that has not an async counterpart
     [exec] 22:30:58,625 WARN  Diagnostics: Method reference Initialize (347,4)-(347,14): Has an async counterpart that could't been used
     [exec] 22:30:58,625 WARN  Diagnostics: Method ExceptWith (358,3)-(373,4) ignored reason: Implements an external interface System.Collections.Generic.ISet<T>.ExceptWith(System.Collections.Generic.IEnumerable<T>) that has not an async counterpart
     [exec] 22:30:58,625 WARN  Diagnostics: Method reference Initialize (364,4)-(364,14): Has an async counterpart that could't been used
     [exec] 22:30:58,625 WARN  Diagnostics: Method SymmetricExceptWith (375,3)-(388,4) ignored reason: Implements an external interface System.Collections.Generic.ISet<T>.SymmetricExceptWith(System.Collections.Generic.IEnumerable<T>) that has not an async counterpart
     [exec] 22:30:58,625 WARN  Diagnostics: Method reference Initialize (381,4)-(381,14): Has an async counterpart that could't been used
     [exec] 22:30:58,625 WARN  Diagnostics: Method Remove (426,3)-(446,4) ignored reason: Implements an external interface System.Collections.Generic.ICollection<T>.Remove(T) that has not an async counterpart
     [exec] 22:30:58,625 WARN  Diagnostics: Method reference Initialize (431,5)-(431,15): Has an async counterpart that could't been used
     [exec] 22:30:58,625 WARN  Diagnostics: Method Clear (448,3)-(463,4) ignored reason: Implements an external interface System.Collections.Generic.ICollection<T>.Clear() that has not an async counterpart
     [exec] 22:30:58,625 WARN  Diagnostics: Method reference Initialize (456,5)-(456,15): Has an async counterpart that could't been used
     [exec] 22:30:58,632 INFO  Diagnostics: Diagnostics for document: Criterion\QueryOver.cs
     [exec] 22:30:58,632 INFO  Diagnostics: Method NHibernate.IQueryOver<TRoot>.List (223,3)-(224,21): Invalid ParameterGeneration option 'Optional' for method 'NHibernate.Criterion.QueryOver<TRoot>.NHibernate.IQueryOver<TRoot>.List()'. Explicit implementor can have only 'Required' option combined with 'Forw
ardNone' or 'SealedForwardNone' option. The ParameterGeneration will be set to 'Required'
     [exec] 22:30:58,632 INFO  Diagnostics: Method NHibernate.IQueryOver<TRoot>.List (226,3)-(227,24): Invalid ParameterGeneration option 'Optional' for method 'NHibernate.Criterion.QueryOver<TRoot>.NHibernate.IQueryOver<TRoot>.List<U>()'. Explicit implementor can have only 'Required' option combined with 'F
orwardNone' or 'SealedForwardNone' option. The ParameterGeneration will be set to 'Required'
     [exec] 22:30:58,632 INFO  Diagnostics: Method NHibernate.IQueryOver<TRoot>.RowCount (235,3)-(236,55): Invalid ParameterGeneration option 'Optional' for method 'NHibernate.Criterion.QueryOver<TRoot>.NHibernate.IQueryOver<TRoot>.RowCount()'. Explicit implementor can have only 'Required' option combined wi
th 'ForwardNone' or 'SealedForwardNone' option. The ParameterGeneration will be set to 'Required'
     [exec] 22:30:58,632 INFO  Diagnostics: Method NHibernate.IQueryOver<TRoot>.RowCountInt64 (238,3)-(239,61): Invalid ParameterGeneration option 'Optional' for method 'NHibernate.Criterion.QueryOver<TRoot>.NHibernate.IQueryOver<TRoot>.RowCountInt64()'. Explicit implementor can have only 'Required' option c
ombined with 'ForwardNone' or 'SealedForwardNone' option. The ParameterGeneration will be set to 'Required'
     [exec] 22:30:58,632 INFO  Diagnostics: Method NHibernate.IQueryOver<TRoot>.SingleOrDefault (241,3)-(242,32): Invalid ParameterGeneration option 'Optional' for method 'NHibernate.Criterion.QueryOver<TRoot>.NHibernate.IQueryOver<TRoot>.SingleOrDefault()'. Explicit implementor can have only 'Required' opti
on combined with 'ForwardNone' or 'SealedForwardNone' option. The ParameterGeneration will be set to 'Required'
     [exec] 22:30:58,632 INFO  Diagnostics: Method NHibernate.IQueryOver<TRoot>.SingleOrDefault (244,3)-(245,35): Invalid ParameterGeneration option 'Optional' for method 'NHibernate.Criterion.QueryOver<TRoot>.NHibernate.IQueryOver<TRoot>.SingleOrDefault<U>()'. Explicit implementor can have only 'Required' o
ption combined with 'ForwardNone' or 'SealedForwardNone' option. The ParameterGeneration will be set to 'Required'
     [exec] 22:30:58,633 WARN  Diagnostics: Diagnostics for document: Action\BulkOperationCleanupAction.cs
     [exec] 22:30:58,633 WARN  Diagnostics: Property accessor get_AfterTransactionCompletionProcess (119,86)-(120,37) ignored reason: Ignored by PropertyConversion option
     [exec] 22:30:58,634 WARN  Diagnostics: Diagnostics for document: Linq\DefaultQueryProvider.cs
     [exec] 22:30:58,634 WARN  Diagnostics: Method Execute (91,3)-(96,4) ignored reason: Implements an external interface System.Linq.IQueryProvider.Execute(System.Linq.Expressions.Expression) that has not an async counterpart
     [exec] 22:30:58,634 WARN  Diagnostics: Method reference ExecuteQuery (95,11)-(95,23): Has an async counterpart that could't been used
     [exec] 22:30:58,634 WARN  Diagnostics: Method PrepareQuery (185,3)-(203,4) ignored reason: Has out parameters
     [exec] 22:30:58,634 WARN  Diagnostics: Method reference CreateFilter (195,21)-(195,33): Has an async counterpart that could't been used
     [exec] 22:30:58,645 WARN  Diagnostics: Diagnostics for document: Impl\AbstractSessionImpl.cs
     [exec] 22:30:58,645 WARN  Diagnostics: Property accessor get_Connection (280,42)-(280,78) ignored reason: Ignored by PropertyConversion option
     [exec] 22:30:58,657 INFO  Diagnostics: Diagnostics for document: Engine\Loading\CollectionLoadContext.cs
     [exec] 22:30:58,657 INFO  Diagnostics: Method EndLoadingCollections (235,3)-(271,4): Invalid ParameterGeneration option 'Required' for method 'NHibernate.Engine.Loading.CollectionLoadContext.EndLoadingCollections(NHibernate.Persister.Collection.ICollectionPersister, System.Collections.Generic.IList<NHib
ernate.Engine.Loading.LoadingCollectionEntry>, bool, NHibernate.Cache.CacheBatcher)'. Method can not have 'Required' when the last parameter is optional. The ParameterGeneration will be set to 'Optional'
     [exec] 22:30:58,665 WARN  Diagnostics: Diagnostics for document: Impl\SessionFactoryImpl.cs
     [exec] 22:30:58,665 WARN  Diagnostics: Method Dispose (840,3)-(849,4) ignored reason: Implements an external interface System.IDisposable.Dispose() that has not an async counterpart
     [exec] 22:30:58,665 WARN  Diagnostics: Method reference Close (848,4)-(848,9): Has an async counterpart that could't been used
     [exec] 22:30:58,681 WARN  Diagnostics: Diagnostics for document: Proxy\Map\MapLazyInitializer.cs
     [exec] 22:30:58,681 WARN  Diagnostics: Property accessor get_GenericMap (20,49)-(20,101) ignored reason: Ignored by PropertyConversion option
     [exec] 22:30:58,681 WARN  Diagnostics: Property accessor get_Map (17,4)-(17,53) ignored reason: Ignored by PropertyConversion option
     [exec] 22:30:58,688 WARN  Diagnostics: Diagnostics for document: Impl\StatelessSessionImpl.cs
     [exec] 22:30:58,688 WARN  Diagnostics: Method BestGuessEntityName (354,3)-(365,4) ignored reason: Implicitly ignored because of the explictly ignored method NHibernate.Engine.ISessionImplementor.BestGuessEntityName(object)
     [exec] 22:30:58,688 WARN  Diagnostics: Method reference GetImplementation (361,46)-(361,63) ignored reason: Cascade ignored
     [exec] 22:30:58,688 WARN  Diagnostics: Method reference GetImplementation (361,46)-(361,63): Has an async counterpart that could't been used
     [exec] 22:30:58,690 WARN  Diagnostics: Diagnostics for document: Collection\Generic\PersistentGenericBag.cs
     [exec] 22:30:58,690 WARN  Diagnostics: Method Clear (188,3)-(203,4) ignored reason: Implements an external interface System.Collections.Generic.ICollection<T>.Clear() that has not an async counterpart
     [exec] 22:30:58,690 WARN  Diagnostics: Method reference Initialize (196,5)-(196,15): Has an async counterpart that could't been used
     [exec] 22:30:58,690 WARN  Diagnostics: Method Remove (216,3)-(225,4) ignored reason: Implements an external interface System.Collections.Generic.ICollection<T>.Remove(T) that has not an async counterpart
     [exec] 22:30:58,690 WARN  Diagnostics: Method reference Initialize (218,4)-(218,14): Has an async counterpart that could't been used
     [exec] 22:30:58,694 WARN  Diagnostics: Diagnostics for document: Collection\Generic\PersistentGenericList.cs
     [exec] 22:30:58,694 WARN  Diagnostics: Method Clear (268,3)-(283,4) ignored reason: Implements an external interface System.Collections.Generic.ICollection<T>.Clear() that has not an async counterpart
     [exec] 22:30:58,694 WARN  Diagnostics: Method reference Initialize (276,5)-(276,15): Has an async counterpart that could't been used
     [exec] 22:30:58,694 WARN  Diagnostics: Method Remove (472,3)-(491,4) ignored reason: Implements an external interface System.Collections.Generic.ICollection<T>.Remove(T) that has not an async counterpart
     [exec] 22:30:58,694 WARN  Diagnostics: Method reference Initialize (477,5)-(477,15): Has an async counterpart that could't been used
     [exec] 22:30:58,704 WARN  Diagnostics: Diagnostics for document: Transaction\AdoTransaction.cs
     [exec] 22:30:58,704 WARN  Diagnostics: Method Dispose (350,3)-(353,4) ignored reason: Implements an external interface System.IDisposable.Dispose() that has not an async counterpart
     [exec] 22:30:58,704 WARN  Diagnostics: Method reference Dispose (352,4)-(352,11): Has an async counterpart that could't been used
     [exec] 22:30:58,711 WARN  Diagnostics: Diagnostics for document: Driver\BasicResultSetsCommand.cs
     [exec] 22:30:58,711 WARN  Diagnostics: Method BindParameters (63,3)-(67,4) ignored reason: Never used
     [exec] 22:30:58,711 WARN  Diagnostics: Function  (66,22)-(66,125) ignored reason: Cascade ignored
     [exec] 22:30:58,711 WARN  Diagnostics: Method reference Bind (66,69)-(66,73) ignored reason: Method is copied
     [exec] 22:30:58,711 WARN  Diagnostics: Method reference Bind (66,69)-(66,73): Has an async counterpart that could't been used
     [exec] 22:30:58,718 WARN  Diagnostics: Diagnostics for document: Action\EntityAction.cs
     [exec] 22:30:58,719 WARN  Diagnostics: Property accessor get_BeforeTransactionCompletionProcess (112,96)-(115,58) ignored reason: Ignored by PropertyConversion option
     [exec] 22:30:58,719 WARN  Diagnostics: Property accessor get_AfterTransactionCompletionProcess (119,94)-(122,57) ignored reason: Ignored by PropertyConversion option
     [exec] 22:30:58,735 WARN  Diagnostics: Diagnostics for document: Collection\Generic\PersistentGenericIdentifierBag.cs
     [exec] 22:30:58,735 WARN  Diagnostics: Method Clear (330,3)-(339,4) ignored reason: Implements an external interface System.Collections.Generic.ICollection<T>.Clear() that has not an async counterpart
     [exec] 22:30:58,735 WARN  Diagnostics: Method reference Initialize (332,4)-(332,14): Has an async counterpart that could't been used
     [exec] 22:30:58,735 WARN  Diagnostics: Method Remove (463,3)-(475,4) ignored reason: Implements an external interface System.Collections.Generic.ICollection<T>.Remove(T) that has not an async counterpart
     [exec] 22:30:58,735 WARN  Diagnostics: Method reference Initialize (465,4)-(465,14): Has an async counterpart that could't been used
     [exec] 22:30:58,740 WARN  Diagnostics: Diagnostics for document: Multi\QueryBatchExtensions.cs
     [exec] 22:30:58,740 WARN  Diagnostics: Method System.Collections.Generic.IEnumerable<TResult>.GetEnumerator (594,4)-(597,5) ignored reason: Explicity implements an external interface System.Collections.Generic.IEnumerable<TResult>.GetEnumerator() that has not an async counterpart
     [exec] 22:30:58,740 WARN  Diagnostics: Method reference GetEnumerable (596,12)-(596,25): Has an async counterpart that could't been used
     [exec] 22:30:58,740 WARN  Diagnostics: Method System.Collections.IEnumerable.GetEnumerator (599,4)-(602,5) ignored reason: Explicity implements an external interface System.Collections.IEnumerable.GetEnumerator() that has not an async counterpart
     [exec] 22:30:58,740 WARN  Diagnostics: Method reference GetEnumerable (601,12)-(601,25): Has an async counterpart that could't been used
     [exec] 22:30:58,741 WARN  Diagnostics: Diagnostics for document: Impl\DelayedEnumerator.cs
     [exec] 22:30:58,741 WARN  Diagnostics: Method System.Collections.IEnumerable.GetEnumerator (36,3)-(39,4) ignored reason: Explicity implements an external interface System.Collections.IEnumerable.GetEnumerator() that has not an async counterpart
     [exec] 22:30:58,741 WARN  Diagnostics: Method reference GetEnumerable (38,25)-(38,38): Has an async counterpart that could't been used
     [exec] 22:30:58,744 WARN  Diagnostics: Diagnostics for document: Collection\Generic\PersistentGenericMap.cs
     [exec] 22:30:58,744 WARN  Diagnostics: Method Add (250,3)-(268,4) ignored reason: Implements an external interface System.Collections.Generic.IDictionary<TKey, TValue>.Add(TKey, TValue) that has not an async counterpart
     [exec] 22:30:58,744 WARN  Diagnostics: Method reference Initialize (265,4)-(265,14): Has an async counterpart that could't been used
     [exec] 22:30:58,744 WARN  Diagnostics: Method Remove (270,3)-(286,4) ignored reason: Implements an external interface System.Collections.Generic.IDictionary<TKey, TValue>.Remove(TKey) that has not an async counterpart
     [exec] 22:30:58,744 WARN  Diagnostics: Method reference Initialize (275,5)-(275,15): Has an async counterpart that could't been used
     [exec] 22:30:58,744 WARN  Diagnostics: Method Clear (374,3)-(389,4) ignored reason: Implements an external interface System.Collections.Generic.ICollection<System.Collections.Generic.KeyValuePair<TKey, TValue>>.Clear() that has not an async counterpart
     [exec] 22:30:58,744 WARN  Diagnostics: Method reference Initialize (382,5)-(382,15): Has an async counterpart that could't been used
     [exec] 22:30:58,747 WARN  Diagnostics: Diagnostics for document: Action\CollectionAction.cs
     [exec] 22:30:58,747 WARN  Diagnostics: Property accessor get_AfterTransactionCompletionProcess (122,94)-(123,111) ignored reason: Ignored by PropertyConversion option
     [exec] 22:30:58,747 WARN  Diagnostics: Property accessor get_Key (55,33)-(55,44) ignored reason: Ignored by PropertyConversion option
     [exec] 22:30:58,747 INFO  ProjectAnalyzer: Diagnose of documents completed
     [exec] 22:30:58,747 INFO  AsyncCodeGenerator: Analyzing project 'NHibernate' completed
     [exec] 22:30:58,747 INFO  AsyncCodeGenerator: Transforming project 'NHibernate' started
     [exec] 22:30:58,751 INFO  ProjectTransformer: Generating documents started
     [exec] 22:31:01,290 INFO  ProjectTransformer: Generating documents completed
     [exec] 22:31:01,290 INFO  ProjectTransformer: Adding generated documents to the project started
     [exec] 22:31:01,352 INFO  ProjectTransformer: Adding generated documents to the project completed
     [exec] 22:31:01,353 INFO  AsyncCodeGenerator: Transforming project 'NHibernate' completed
     [exec] 22:31:01,365 INFO  AsyncCodeGenerator: Applying project 'E:\Projets\nhibernate\nhibernate-core\src\NHibernate\NHibernate.csproj' changes started
     [exec] 22:31:01,393 INFO  AsyncCodeGenerator: Applying project 'E:\Projets\nhibernate\nhibernate-core\src\NHibernate\NHibernate.csproj' changes completed
```

fredericDelaporte:
If I am alone having this, let's assume it is a trouble with my setup. @hazzik, @bahusoid, may you please confirm you have no async generation glitch when regenerating this PR please?

fredericDelaporte:
I have now just tried on my Debian box: same trouble! Rider 2019.2.2 (same on my Windows laptop), dotnet 2.2.204, mono 5.16.0.179.

hazzik:
For me ShowBuildMenu.bat does not work at all.

Tools/.gitignore
```diff
@@ -1,6 +1,8 @@
 nuget.exe
 NUnit.*
```
fredericDelaporte: Ii should be changed to `nunit.*`. `NUnit` was the right case prior to this PR, but now the NUnit directories are in lower case. It is not an issue under Windows, but it is one under Linux. It causes all those NUnit packages to be seen as untracked.

fredericDelaporte:
> For me ShowBuildMenu.bat does not work at all.

I had the trouble at some point, but purging the binaries in the build tool folder caused it to be gone, and I have not investigated further. Maybe that is something like that for you too Alexander.

hazzik:
I'm facing what @lillo42 has experienced in #2208 but on Windows.

hazzik:
For me AsyncGenerator works fine. But I have only VS 2019 (16.2.2 atm)

bahusoid:
It also works fine for me on Windows. I also have only VS 2019:
```
   [exec] Microsoft (R) Build Engine version 16.0.450+ga8dc7f1d34 for .NET Core
```
and .NET SDK version 2.1.602.

So may be it would make sense to split this PR in two to make sure it's not something .net core specific ( "Upgrade version" and "Migration to .NET Core")?

default.build
maca88: Fixed.

Tools/.gitignore
maca88: Changed and also removed `SQLitePCLRaw.core.*` as it is not needed anymore.

fredericDelaporte:
I will test this.
I was also intending to test the trouble on my employer PC, but I forgot. On that PC, there is also no VS 2019 but a VS 2017. But there is no Rider either contrary to my home PCs, and maybe that is the culprit.

fredericDelaporte:
The AsyncGeneratorNet472 branch exhibits the same trouble on my PCs (Windows & Linux), so that is not a "switch to core" trouble. It seems bound to the current async generator version, and some other specificity of both of mine PCs.
There is still no reference to `FutureBatch` in diagnostics, although it does generate a file for it, but with one method lacking.

hazzik:
@fredericDelaporte could it be the same/similar to last time when you had partially installed VS?

fredericDelaporte:
There is no VS installed on my Linux PC, so definitely not. Anyway I had already checked my VS installation under Windows, too.
Now I have re-check all this on my employer PC, so a third PC, this one without Rider, and I still get the same trouble.
I cannot get what causes this trouble to occur only with me, whatever the device I use. French regional settings interfering in some way? That would be quite strange.
I guess I should take time debug the generator, although I am not knowing its internals and it may take me some time to figure where I should debug.

Anyway as long as I am the only one to have that issue, let ignore it.
May someone else validate this PR? (It needs a rebase I think, by the way.)

maca88:
> French regional settings interfering in some way?

It shouldn't. I tried to changed the regional setting to French and it still works, so the issue must be something else.
At this point we know that the issue is not related to .NET Core and it works with version `0.14.0` which is currently used on the master branch. I've analized the commtits from version `0.14.0` onwards and none of the commits changed the method scanning logic, which I suspect is the issue. The only thing that may change it, is the upgrade of Roslyn api from `2.10.0` to `3.0.0` which was released with version `0.16.0`. In order to test that I've made the following branches:
- [AsyncGenerator17](https://github.com/maca88/nhibernate-core/tree/AsyncGenerator17) version `0.17.0`
- [AsyncGenerator16](https://github.com/maca88/nhibernate-core/tree/AsyncGenerator16) version `0.16.0`
- [AsyncGenerator15](https://github.com/maca88/nhibernate-core/tree/AsyncGenerator15) version `0.15.0`

all branches uses the old Roslyn api version `2.10.0` and the generation is restricted to NHibernate project only. @fredericDelaporte could you test the above branches on one of your computers and report on which branch the issue appears. Could you also include the whole log in debug mode by changing it in `AsyncGenerator.CommandLine.exe.config` inside the folder `net461`, for the first version that has the issue:
```
<root>
  <priority value="INFO" /> <!-- Change this to DEBUG -->
      <appender-ref ref="Console" /> <!-- Change this to RollingFile as the generated log is more than 3MB -->
</root>
```

fredericDelaporte:
No trouble with the 15, trouble occurs with the 16 (checked under Windows, home PC with Rider/VS2017).
Seen diagnostics in logs for that file are:
```
Diagnostics for document: Impl\FutureBatch.cs
Method Add (33,3)-(44,4) ignored reason: Never used
Method Add (46,3)-(49,4) ignored reason: Never used
Method GetFutureValue (51,3)-(60,4) ignored reason: Never used
Function  (56,5)-(56,89) ignored reason: Cascade ignored
Function  (56,5)-(56,89): Is passed as an argument to a ObjectCreationExpression which is currently not supported
Function  (55,5)-(55,50) ignored reason: Cascade ignored
Function  (55,5)-(55,50): Is passed as an argument to a ObjectCreationExpression which is currently not supported
Method reference GetCurrentResult (55,11)-(55,27) ignored reason: Method is copied
Method GetEnumerator (62,3)-(71,4) ignored reason: Never used
Function  (67,5)-(67,89) ignored reason: Cascade ignored
Function  (67,5)-(67,89): Is passed as an argument to a ObjectCreationExpression which is currently not supported
Function  (66,5)-(66,50) ignored reason: Cascade ignored
Function  (66,5)-(66,50): Is passed as an argument to a ObjectCreationExpression which is currently not supported
Method reference GetCurrentResult (66,11)-(66,27) ignored reason: Method is copied
Method GetResults (73,3)-(112,4) ignored reason: Never used
Method reference List (83,19)-(83,23) ignored reason: Cascade ignored
Method reference GetResultsFrom (107,15)-(107,29) ignored reason: Cascade ignored
Method GetCurrentResult (114,3)-(117,4) ignored reason: Never used
Method reference GetResults (116,20)-(116,30) ignored reason: Cascade ignored
Method CreateMultiApproach (125,3)-(125,95) ignored reason: Never used
Method AddTo (126,3)-(126,109) ignored reason: Never used
Method ClearCurrentFutureBatch (128,3)-(128,53) ignored reason: Never used
Method IsQueryCacheable (129,3)-(129,66) ignored reason: Never used
Method CacheRegion (130,3)-(130,63) ignored reason: Never used
Method AddResultTransformer (132,3)-(138,4) ignored reason: Never used
Method .ctor (28,3)-(31,4) ignored reason: Unsupported method kind Constructor
Method TransformTuple (154,4)-(157,5) ignored reason: Never used
Method TransformList (159,4)-(169,5) ignored reason: Never used
Method Equals (172,4)-(175,5) ignored reason: Overrides an external method object.Equals(object) that has not an async counterpart
Method GetHashCode (177,4)-(180,5) ignored reason: Overrides an external method object.GetHashCode() that has not an async counterpart
Method .ctor (149,4)-(152,5) ignored reason: Unsupported method kind Constructor
Property accessor get_Future (17,34)-(17,38) ignored reason: Ignored by PropertyConversion option
Property accessor set_Future (17,39)-(17,43) ignored reason: Ignored by PropertyConversion option
Property accessor get_Query (15,34)-(15,38) ignored reason: Ignored by PropertyConversion option
Property accessor set_Query (15,39)-(15,43) ignored reason: Ignored by PropertyConversion option
Property accessor get_ResultType (16,36)-(16,40) ignored reason: Ignored by PropertyConversion option
Property accessor set_ResultType (16,41)-(16,45) ignored reason: Ignored by PropertyConversion option
```

With 15, if that helps, I have:
```
Diagnostics for document: Impl\FutureBatch.cs
Method Add (33,3)-(44,4) ignored reason: Never used
Method Add (46,3)-(49,4) ignored reason: Never used
Method GetFutureValue (51,3)-(60,4) ignored reason: Never used
Function  (56,5)-(56,89) ignored reason: Cascade ignored
Function  (56,5)-(56,89): Is passed as an argument to a ObjectCreationExpression which is currently not supported
Function  (55,5)-(55,50) ignored reason: Cascade ignored
Function  (55,5)-(55,50): Is passed as an argument to a ObjectCreationExpression which is currently not supported
Method reference GetCurrentResult (55,11)-(55,27) ignored reason: Method is copied
Method reference GetCurrentResult (55,11)-(55,27): Has an async counterpart that could't been used
Method GetEnumerator (62,3)-(71,4) ignored reason: Never used
Function  (66,5)-(66,50) ignored reason: Cascade ignored
Function  (66,5)-(66,50): Is passed as an argument to a ObjectCreationExpression which is currently not supported
Method reference GetCurrentResult (66,11)-(66,27) ignored reason: Method is copied
Method reference GetCurrentResult (66,11)-(66,27): Has an async counterpart that could't been used
Function  (67,5)-(67,89) ignored reason: Cascade ignored
Function  (67,5)-(67,89): Is passed as an argument to a ObjectCreationExpression which is currently not supported
Method CreateMultiApproach (125,3)-(125,95) ignored reason: Never used
Method AddTo (126,3)-(126,109) ignored reason: Never used
Method ClearCurrentFutureBatch (128,3)-(128,53) ignored reason: Never used
Method IsQueryCacheable (129,3)-(129,66) ignored reason: Never used
Method CacheRegion (130,3)-(130,63) ignored reason: Never used
Method AddResultTransformer (132,3)-(138,4) ignored reason: Never used
Method .ctor (28,3)-(31,4) ignored reason: Unsupported method kind Constructor
Property accessor get_Query (15,34)-(15,38) ignored reason: Ignored by PropertyConversion option
Property accessor set_Query (15,39)-(15,43) ignored reason: Ignored by PropertyConversion option
Property accessor get_Future (17,34)-(17,38) ignored reason: Ignored by PropertyConversion option
Property accessor set_Future (17,39)-(17,43) ignored reason: Ignored by PropertyConversion option
Property accessor get_ResultType (16,36)-(16,40) ignored reason: Ignored by PropertyConversion option
Property accessor set_ResultType (16,41)-(16,45) ignored reason: Ignored by PropertyConversion option
Method TransformTuple (154,4)-(157,5) ignored reason: Never used
Method TransformList (159,4)-(169,5) ignored reason: Never used
Method Equals (172,4)-(175,5) ignored reason: Overrides an external method object.Equals(object) that has not an async counterpart
Method GetHashCode (177,4)-(180,5) ignored reason: Overrides an external method object.GetHashCode() that has not an async counterpart
Method .ctor (149,4)-(152,5) ignored reason: Unsupported method kind Constructor
```

maca88:
@fredericDelaporte thanks for the logs. From the logs of version `0.16.0`, the `GetResult` method is ignored because it is only used in method `GetCurrentResult` which is also ignored. Unfortunately, when scanning for `GetCurrentResult` method usages Roslyn will not return any as the only usage is when instantiating `FutureValue`, which has a compilation error due to the not yet generated `GetCurrentResultAsync` method. The generator in such case scans through the compilation errors and if it finds an error `CS0103` which contains the method name, the method won't be ignored even if Roslyn returns no usages for it:
https://github.com/maca88/AsyncGenerator/blob/824a63fd50142885cb8ef350f27bd0310a23123c/Source/AsyncGenerator/Analyzation/Internal/ProjectAnalyzer.PostAnalyze.cs#L680-L686

It seems like that in your case the error `CS0103` is not present and that is the reason why `GetCurrentResult` is ignored. I've made an isolated test maca88/AsyncGenerator/pull/139, but I am still unable to reproduce the issue. @fredericDelaporte could you please try to run the test and check if it works for you and if not could you paste the errors from the `GetDiagnostics` method at https://github.com/maca88/AsyncGenerator/blob/824a63fd50142885cb8ef350f27bd0310a23123c/Source/AsyncGenerator/Analyzation/Internal/ProjectAnalyzer.PostAnalyze.cs#L680. The expected error should be: 
```
error CS0103: The name 'GetCurrentResultAsync' does not exist in the current context
```

bahusoid:
So regex ` '(.+)'` is used for extracting method name from error message and my understanding errors are localized in french for Frederic.
So my guess error message in french contains `'` elsewhere and your regex grabs some different text. It may help to change regex to  `'(.+?)'` (use non-greedy quantifier). But it depends on where additional `'` occurs in error message.

I think you should try to force English localization in AsyncGenerator. Add something like this on app init:
```C#
CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
CultureInfo.CurrentUICulture = CultureInfo.InvariantCulture;
CultureInfo.DefaultThreadCurrentCulture = CultureInfo.InvariantCulture;
CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.InvariantCulture;
```

maca88:
Thanks @bahusoid, by setting the `XXXUICulture` I am finally able to reproduce the issue. 

> So my guess error message in French contains ' elsewhere

Thats correct. The reason why the version `0.15.0` and below aren't affected by this problem is because they are using Roslyn version `2.7.0` that don't include localizations, which is something that I overlooked.

> I think you should try to force English localization in AsyncGenerator

As `AsyncGenerator` may teoretically be used as a library I will fix the issue so that it works with all localizations.

@fredericDelaporte sorry for bothering you, I could detect the issue earlier.

bahusoid:
> As AsyncGenerator may teoretically be used as a library I will fix the issue so that it works with all localizations.

My point it's hard to do such parsing for translations. What if some translation uses some different wrapping around method? Like `"methodName"` instead of `'mehtodName'`.  So I think before compilation starts you should change localization to English. You can change it back when your are done.  So it shouldn't be an issue for library case.

maca88:
> What if some translation uses some different wrapping around method?

[I've removed the wrapping](https://github.com/maca88/AsyncGenerator/pull/139/commits/1c6020644ae5b1101e60468d530ffaeb0b55a4c3#diff-cde142aa39c6709fd150881f3b6faa0aR681) and replaced the regex search by using `Contains`, which I think is good enough.

maca88:
Upgraded to version `0.17.1`, which fixes the issue with `FutureBatch` and rebased to master.

fredericDelaporte:
It is great you have found the cause, @maca88 and @bahusoid. Now I do not have issuez regenrating on my setup. 

So this bug depends on the language. Switching the thread to English may not be enough, depending on how English localization is embedded within Roslyn. If it is the neutral language resource (which is very likely to be), it should be ok. Otherwise some distribution of Roslyn may omit the English localization and then the switch would not be honored.

So your solution, just using a simple `Contain` without delimiters, which should be language independent, is likely preferable to a language switch.

Still we may have the reverse trouble: imagine some code using an `exist` method. For honoring convention it should be named `Exist`, but still, this `exist` could get an undue async counterpart if the code also generate a CS0103 in the same file. And this bug would depend on localization too. To avoid it, maybe the `Contain` should be changed to something like `Regex.IsMatch(o.GetMessage(), "[^\s\w]" + Regex.Escape(methodOrAccessor.AsyncCounterpartName) + "[^\s\w]")`. This assumes that in all localization, the method name will be delimited by something else than a white-space or a word character. It seems a reasonable assumption to me.

And we may also have the trouble for methods matching a CS0103 of the same file but for a method with a longer name.

I assume the code checks only the CS0103 of the same file, so the risk is limited, but not unlikely. Checking for some kind of delimiters would considerably reduce the risk. It would not eliminate it, particularly in case the file contains many classes. But then that could be worked around by splitting the file, which is anyway a good thing to do in such case.

> @fredericDelaporte sorry for bothering you, I could detect the issue earlier.

Me too I could, no point being sorry for me. Thanks for having found a solution.

bahusoid:
> Otherwise some distribution of Roslyn may omit the English localization and then the switch would not be honored.

English resource is embedded into project (see [resource](https://github.com/dotnet/roslyn/blob/768f9327148d18a30952c5bd61427def1337d0eb/src/Features/CSharp/Portable/CSharpFeaturesResources.resx#L458) and [project](https://github.com/dotnet/roslyn/blob/768f9327148d18a30952c5bd61427def1337d0eb/src/Features/CSharp/Portable/Microsoft.CodeAnalysis.CSharp.Features.csproj#L61)). And translations are provided as satellite assemblies. So I don't think that English localization can be omitted somehow.

maca88:
> should be changed to something like Regex.IsMatch(o.GetMessage(), "[^\s\w]" + Regex.Escape(methodOrAccessor.AsyncCounterpartName) + "[^\s\w]")

I tried this but it does not work for the following cultures:
cs: `Název GetCurrentResultAsync v aktuálním kontextu neexistuje.`
ko: `GetCurrentResultAsync' 이름이 현재 컨텍스트에 없습니다.`
tr: `GetCurrentResultAsync' adı geçerli bağlamda yok`

> Checking for some kind of delimiters would considerably reduce the risk. It would not eliminate it, particularly in case the file contains many classes.

Probably the most robust approach would be as @bahusoid suggested to always use invariant culture to be able to use the delimiters and as you suggested, searching the error only for lines where the class is defined. I've made maca88/AsyncGenerator/issues/140 for this, but I don't think that it is a showstopper for this PR.

src/NHibernate.Test/Linq/MathFTests.cs
```diff
@@ -7,7 +7,6 @@
 
 namespace NHibernate.Test.Linq
 {
-	[TestFixture]
 	public class MathFTests : LinqTestCase
```
fredericDelaporte: I guess this class was not having async test generated due to being netcoreapp only, when we were generating for netFx.

Have you removed the attribute to avoid having its asyn tests generated? Why not having them?

src/NHibernate/AdoNet/SqlClientBatchingBatcher.cs
fredericDelaporte: So we should remove the partial keyword on this class.

src/NHibernate.Test/Linq/MathFTests.cs
maca88: > I guess this class was not having async test generated due to being netcoreapp only, when we were generating for netFx.

Correct.

> Have you removed the attribute to avoid having its asyn tests generated?

Yes, otherwise the generated tests would fail as `ToListAsync` would be called on a queryable `List<>` which is populated on set up:
https://github.com/nhibernate/nhibernate-core/blob/8f765cf6b18fe5fb5eb934387677c97a3370dee0/src/NHibernate.Test/Linq/MathFTests.cs#L17-L19

src/NHibernate/AdoNet/SqlClientBatchingBatcher.cs
maca88: Removed.

fredericDelaporte: Off-topic considerations: Wondering on whether the type causing this to be NetFx only was still not available in core, I found out it seems it is available in netcoreapp2.x, but bugged and fixed only in 3.0. See dotnet/corefx#29391.
But for 5.0, we may have a batching API available, which would allow to remove most of our batcher implementations in favor of using that API. See dotnet/corefx#35135. (Assuming that targeting coreFx 5.0 implies dropping NetFx, as coreFx 5.0 is supposed to obsolete netFx.)

src/NHibernate.Test/Linq/MathFTests.cs
fredericDelaporte: I had to test it to understand. So the trouble is, the async generator has no way to understand on which queryable using the NHibernate `ToListAsync` is valid or not.
This means the async generator cannot be used in an application which would use NHibernate and some other queryables.
That is off-topic for this PR, but maybe we should then relax the `ToListAsync` implementation for defaulting to `ToList` on not handled queryables, with some warning, or with an option to enable not throwing.

PR #2188, opened 2019-07-10
Add BenchmarkDotnet

lillo42:
Hi folks,

I'm studying performance on NHibernate and what I could do to improve, and I realize we don't have a Benchmark project, so I create the project.

What do you think about add this project? Should I change something? Feel free to change anything

We need to add more test


fredericDelaporte:
I have not looked into your PR, but just to let you know:
We have sometime used FransBouma benchmark. I have also suggested an additional case in his benchmark, see FransBouma/RawDataAccessBencher#39.

hazzik:
It appears there is no much interest in this.

PR #2195, opened 2019-08-18
Fix nuget symbol push

fredericDelaporte:
They are pushed only with the v3 api url, see
https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg#publishing-a-symbol-package

Fix the #2019 PR which is in the same milestone.

### DIFF
--- a/default.build
+++ b/default.build
@@ -323,7 +323,7 @@
 				</items>
 			</in>
 			<do>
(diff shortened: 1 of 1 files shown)


hazzik:
Should this go to 5.2.x?

fredericDelaporte:
This would require back-porting #2019 to 5.2.x. 5.2.x does not support snupkg currently. Maybe were you willing to ask this on #2196, although #2196 does already target 5.2.x.

hazzik:
Never mind then

PR #2196, opened 2019-08-18
Fix session factory deserialization with Prevalence Cache

### LINKED ISSUE #2190: Cannot instantiate a SessionFactory using Prevalence cache
When using the Prevalence cache provider, a `NullReferenceException` is thrown from `SessionFactoryImpl.GetRealObject` during `Configuration.BuildSessionFactory`. This exception occurs because the Prevalence cache provider attempts to unserialize a session factory upon start. The constructor for SessonFactorImpl calls `settings.CacheProvider.Start` prior to calling `SessionFactoryObjectFactory.AddInstance`. Thus the call to `SessionFactoryObjectFactory.GetNamedInstance` in `SessionFactoryImpl.GetRealObject` returns `null` despite the fact that the SessionFactory that is being deserialized is the one that started the cache provider.

There are a couple of ways to resolve this issue:
1. Do not serialize the SessionFactory as part of the cache key. This is the approach taken by Hibernate ORM.
2. Add an additional check to `SessionFactoryImpl.GetRealObject` that compares the name of the requested session factory against that of the current session factory prior to calling `SessionFactoryObjectFactory.GetNamedInstance`.

Personally, I prefer option 1 over option 2. However, option 1 may not be possible due to changes made in #1891.

fredericDelaporte:
Fix #2190 

The Bamboo Prevalence cache does serialize entities to disk, with their keys, using binary serialization. NH 5.2 has included the session factory in the serialized data of keys.

This causes a crash when instantiating a session factory using a Prevalence cache having entities serialized, because their deserialization is triggered by the cache provider initialization, which occurs during session factory instantiation, before it registers its name for deserialization resolving. So it cannot deserialize keys due to them failing to deserialize their session factory.


src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -214,6 +214,22 @@ public SessionFactoryImpl(Configuration cfg, IMapping mapping, Settings settings
...
+				uuid = (string)UuidGenerator.Generate(null, null);
+			}
+			catch (Exception ex)
+			{
+				throw new AssertionFailure("Could not generate UUID", ex);
+			}
+
+			SessionFactoryObjectFactory.AddInstance(uuid, name, this, properties);
+
+			#endregion
```
fredericDelaporte: This fix, moving name/uuid registration at the beginning of the session factory construction, seems to work and does not cause any old test failure, but is quite debatable.

It allows deserialization to resolve a session factory not yet fully built. And this occurs in this Prevalence cache case, meaning it will recompute its cached entities keys hashcodes with a session factory not yet fully built.

This case is quite wicked. The persisters creation creates the caches, so we have to start their provider before creating the persisters.

PR #2197, opened 2019-08-24
Fix dependent transaction failure

### LINKED ISSUE #2172: Using DependentTransaction fails
When I use `Transaction.Current.DependentClone(DependentCloneOption.RollbackIfNotComplete)`, Nhibernate went wrong. The error message is `HibernateException: Synchronization timeout for transaction completion. Either raise {Cfg.Environment.SystemTransactionCompletionLockTimeout}, or this may be a bug in NHibernate.`. This error may occur in NHibernate 5.x.x
Here is the code:
```c#
    public class TestNHibernate
    {
        public static void Main(string[] arg)
        {
            new TestNHibernate().Do1();
        }
        public virtual void Do1()
        {
            Transaction.Current = new CommittableTransaction();
            var current = Transaction.Current;
            Do2();
            (current as CommittableTransaction)?.Commit();
        }
        public virtual void Do2()
        {
            var current1 = Transaction.Current;
            Transaction.Current = Transaction.Current.DependentClone(DependentCloneOption.RollbackIfNotComplete);
            var builtSessionFactory =
                Fluently.Configure()
                    .Database(
                        MsSqlConfiguration.MsSql2012.ConnectionString(
                            @"Server=localhost;Database=test;user=sa;password=123;Trusted_Connection=True;"))
                    .Mappings(m => m.FluentMappings.Add<PM>())
                    .BuildSessionFactory();
            var session = builtSessionFactory.OpenSession();
            session.Save(new Person { });
            (Transaction.Current as DependentTransaction).Complete();
            Transaction.Current.Dispose();
            Transaction.Current = current1;
        }
        public class Person
        {
        }
        public class PM : ClassMap<Person>
        {
            public PM()
            {
                Id(person => person.Id);
                Map(u => u.Name);
            }
        }
    }
```

fredericDelaporte:
Fix #2172

This is a regression of NH5.0, causing dependent transactions to be unusable.
The trouble with dependent transactions is, they are usually disposed of before completing the actual transaction.
This causes a bug in NHibernate transaction context when checking for the transaction status. It fortunately uses a clone for this, but due to a bug where the status may still be active while the transaction is rollbacked, it does also check the original transaction, and currently assumes that if it is disposed, the transaction is no more active. 
This is not true with dependent transactions.
(Well actually the check method just yield null to tell it does not know, but then the code calling it do the "not active" path.)

Since this PR targets 5.2.x, I will delay releasing 5.2.6.
If merged, this PR should be back-ported down to 5.0.x


src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -285,16 +285,17 @@ protected virtual void Unlock()
 				{
 					// Cloned transaction is not disposed "unexpectedly", its status is accessible till context disposal.
 					var status = EnlistedTransaction.TransactionInformation.Status;
-					if (status != TransactionStatus.Active)
+					if (status != TransactionStatus.Active || _preparing)
```
fredericDelaporte: The trouble with dependent transactions occurs in the prepare phase, where we are sure the transaction is not (yet) roll-backed.
This way of dodging that trouble looks like one more crutch in this dreaded system transaction management. But I have seen no better.

(Checking the original transaction concrete type is unreliable, because a `DependentTransaction` own `Clone` method will not yield another `DependentTransaction`, but a `Transaction`. Moreover, knowing it is a dependent transaction will anyway not help knowing whether the transaction is still actually active or not.)

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -236,7 +236,7 @@ public virtual void Wait()
 					// Remove the block then throw.
 					Unlock();
 					throw new HibernateException(
-						"Synchronization timeout for transaction completion. Either raise {Cfg.Environment.SystemTransactionCompletionLockTimeout}, or this may be a bug in NHibernate.");
+						$"Synchronization timeout for transaction completion. Either raise {Cfg.Environment.SystemTransactionCompletionLockTimeout}, or this may be a bug in NHibernate.");
```
fredericDelaporte: Trivial bug fix by the way. I do not think it is even worth a mention in commit.

fredericDelaporte:
WIP: it will still surely fail in case a session is used with many dependent transaction clones of the same transaction. Going to add test for that.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -285,16 +285,44 @@ protected virtual void Unlock()
...
+					// to the current transaction being already disposed.)
+					// The current transaction may not be reachable during 2PC phases and transaction completion events,
+					// but in such cases the context transaction is either no more active or in prepare phase, which is
+					// already covered by _preparing test.
+					try
+					{
+						var currentTransaction = System.Transactions.Transaction.Current;
+						if (!ReferenceEquals(currentTransaction, _originalTransaction) &&
+							currentTransaction == EnlistedTransaction)
+							return currentTransaction.TransactionInformation.Status;
```
fredericDelaporte: This allows chaining multiple dependent transaction usage of the same original transaction on the same session. This means this use case will works with some exception handling, which hurts performances.
But dodging theses exceptions for this dependent case would require to check the current transaction first (instead of trying the original one), which would introduce another exception handling for more regular cases than dependent transaction or rollback.

fredericDelaporte:
It seems we are getting an async test trouble. Fiddling with `Transaction.Current` in an async context along with distributed transactions (which introduces actual thread changes) does not work well. Especially, it looks like the finally cannot set to `null` `Current` for the original execution context, causing teardown and subsequent tests to fail.
So I am going to commit an async ignore change for these tests.

src/NHibernate.Test/Async/SystemTransactions/DistributedSystemTransactionFixture.cs
```diff
@@ -13,13 +13,14 @@
...
 using System.Transactions;
 using log4net;
-using log4net.Repository.Hierarchy;
 using NHibernate.Cfg;
 using NHibernate.Engine;
-using NHibernate.Linq;
 using NHibernate.Test.TransactionTest;
 using NUnit.Framework;
 
+using sysTran = System.Transactions;
```
hazzik: I dont' like this much.

fredericDelaporte: Ok, I will revert.
I did that due to otherwise having to specify the full namespace everywhere before "sysTran" `Transaction` because of the conflict with NHibernate `Transaction` namespace, and I then rather shorten the namespace prefixing.

PR #2174, opened 2019-06-12
Fix future queries with non lazy associations

bahusoid:
Fixes #2173 (some technical details behind this issue can be  found [here](https://github.com/nhibernate/nhibernate-core/issues/2173#issuecomment-501019937))

Implemented the following fix:
* Moved actual collection initialization logic after all queries are processed (as it's done in legacy `MultiCriteriaImpl`/`MultiQueryImpl`)
* Introduced `LoadingCollectionEntry.StopLoading` flag that indicates that collection should not be populated. This allows to delay actual collection initialization till all other batch queries are processed (so it's a fix for #1293)
 
Issue is introduced in 5.2 with new `QueryBatch` implementation.

PR #2200, opened 2019-08-25
Release 5.2.6

fredericDelaporte:
Preparing the release. It currently accounts for pending PRs, async update and dependent transaction fix. If some of them are discarded or delayed to another release, this PR will have to be updated accordingly.

As written by @bahusoid [here](https://github.com/nhibernate/nhibernate-core/pull/2174#discussion_r293383930), some care should be taken when merging to master after the release:

> When merging to master this change should be ignored. The following code is expected in master (no need to wrap in `new List`):
> ```C#
> queryInfo.ResultToCache = queryCacheBuilder.Result;
> ```

### DIFF
--- a/releasenotes.txt
+++ b/releasenotes.txt
@@ -1,3 +1,30 @@
+Build 5.2.6
+=============================
+
+Release notes - NHibernate - Version 5.2.6
+
+11 issues were resolved in this release.
+
+** Bug
+
+   * #2190 Cannot instantiate a SessionFactory using Prevalence cache
+   * #2177 New Fetch() method in QueryOver returns IQueryOver<> instead of QueryOver<>
+   * #2172 Using DependentTransaction fails
+   * #2175 Subcriteria on component collection generates incorrect join alias 
+   * #2173 Futures not batching correctly in NH 5.2.x
+   * #2141 Undefined call to Equals object in collection during flush just before commit 
+   * #2127 StackExchangeRedisCache with PreferMultipleGet = true calls GetMany multiple times
+   * #2110 Wrong GUID to string conversion with SQLite  BinaryGuid=False
+
+** Task
+
+   * #2200 Release 5.2.6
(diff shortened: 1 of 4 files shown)


releasenotes.txt
```diff
@@ -1,3 +1,31 @@
...
+   * #2173 Futures not batching correctly in NH 5.2.x
+   * #2141 Undefined call to Equals object in collection during flush just before commit 
+   * #2127 StackExchangeRedisCache with PreferMultipleGet = true calls GetMany multiple times
+   * #2110 Wrong GUID to string conversion with SQLite  BinaryGuid=False
+
+** Task
+
+   * #2200 Release 5.2.6
+   * #2199 Upgrade AsyncGenerator to 0.8.2.12
+   * #2180 Upgrade AsyncGenerator to 0.8.2.11 for 5.2
```
hazzik: Shouldn't this be superseded by 2199?

fredericDelaporte: You mean, just dropped out of the release notes? Why not.

fredericDelaporte:
Rebased, I will likely release it next Sunday.

fredericDelaporte:
#2111, in this release, has updated the documentation. But I do not intend to update it on nhibernate.info yet, because it is about a new feature, while we only publish minor version of the documentation, not release version. So updating the 5.2 documentation could be confusing for those attempting to use the feature with a 5.2.5 or below version of NHibernate. I think it should wait 5.3.

oskarb:
> #2111, in this release, has updated the documentation. But I do not intend to update it on nhibernate.info yet, because it is about a new feature, while we only publish minor version of the documentation, not release version. So updating the 5.2 documentation could be confusing for those attempting to use the feature with a 5.2.5 or below version of NHibernate. I think it should wait 5.3.

I agree - especially given that it was most likely for a really narrow use case anyway.

Issue #2202, opened 2019-09-02
SetSession(ISession session) is not invoked on an ISessionFactory-scoped interceptors


YevhenLukomskyi:
`OpenSession()` doesn't invoke `SetSession(ISession session)` on an ISessionFactory-scoped interceptors. It works when an interceptor is registered on a session level,
It is reproducible in NHibernate 5.2.5

fredericDelaporte:
There is a reason for this: a session factory can be used in multi-threaded context, it has to be thread safe.
So a session factory interceptor cannot have the session, it will cause it to be not thread safe.

PR #2203, opened 2019-09-02
Set default interceptor in SessionBuilderImpl

YevhenLukomskyi:
Resolves https://github.com/nhibernate/nhibernate-core/issues/2202


src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1460,7 +1461,7 @@ public virtual ISession OpenSession()
 				if (_interceptor != null)
 				{
 					// NH specific feature
-					// _interceptor may be the shared accros threads EmptyInterceptor.Instance, but that is
+					// _interceptor may be the shared across threads EmptyInterceptor.Instance, but that is
```
fredericDelaporte: This comment is now violated. With this PR, it may be another shared interceptor than the empty one which does nothing of the session and so is harmless.

This introduces threading bugs. This is a no go.

PR #2205, opened 2019-09-08
Merge 5.2.6

fredericDelaporte:
This PR is not meant to be merged. For having a clean merge, it will be better to directly push to master.


doc/reference/modules/configuration.xml
```diff
@@ -1081,6 +1081,20 @@ in the parameter binding.</programlisting>
                             </para>
                         </entry>
                     </row>
+                    <row>
+                        <entry>
+                            <literal>sqlite.binaryguid</literal>
```
fredericDelaporte: This new setting lacks its declaration in nhibernate-configuration.xsd file. Thus, it cannot be used in configuration files, and can only be used programmatically. Fixing this is just one line to add into nhibernate-configuration.xsd.
Ideally, it should be done on 5.2.x. But I do not think it deserves a release on its own. So I propose to handle it with a PR targeting master, and waiting for 5.3. If we have a 5.2.x to do for some other reasons before, we could then retarget 5.2x. Otherwise just fix it for 5.3. @oskarb

fredericDelaporte:
The failures are expected, due to the tests which were not ignored for async generation. Those tests directly manipulate the current system transaction, which does not work well mixed with async operations.

src/AsyncGenerator.yml
```diff
@@ -183,6 +183,18 @@
...
+      containingTypeName: DistributedSystemTransactionFixture
+    - conversion: Ignore
+      name: CanUseSessionWithManyDependentTransaction
+      containingTypeName: DistributedSystemTransactionFixture
+    - conversion: Ignore
+      name: CanUseDependentTransaction
+      containingTypeName: SystemTransactionFixture
+    - conversion: Ignore
+      name: CanUseSessionWithManyDependentTransaction
+      containingTypeName: SystemTransactionFixture
```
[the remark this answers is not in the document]
maca88: The issue was indeed since 0.14.0 when the `NUnitPlugin` was introduced. `NUnitPlugin` rules currently have a higher priority that the rules defined by the user which is something that I overlooked. By just swaping the priorities will not work for NHibernate, because of the following rule:
```
- conversion: Ignore
  anyBaseTypeRule: IsTestCase
```
which is supposed to run last (after `NUnitPlugin` rules), otherwise most of the tests would not be generated. In order to support that I've made maca88/AsyncGenerator/pull/138, which allows to define when the rule should be executed by using a new option `executionPhase`:
```
- conversion: Ignore
  anyBaseTypeRule: IsTestCase
  executionPhase: PostProviders
``` 
Should I create a new PR for fixing this issue or update #2182?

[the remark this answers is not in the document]
fredericDelaporte: Ok, thanks for looking into this. Updating #2182 is fine I think.

releasenotes.txt
```diff
@@ -342,6 +369,14 @@ Release notes - NHibernate - Version 5.2.0
 As part of releasing 5.2.0, a misnamed setting in 5.0.0 release notes has been fixed:
 transaction.use_connection_on_system_events correct name is transaction.use_connection_on_system_prepare
 
+Build 5.1.6
```
fredericDelaporte: I have included the backports release notes (5.1.6 and 5.0.8) in the async regen commit.

src/NHibernate/Cache/ReadOnlyCache.cs
```diff
@@ -87,7 +87,7 @@ public ISoftLock Lock(CacheKey key, object version)
 				return result;
 			}
 
-			var checkKeys = new List<CacheKey>();
+			var checkKeys = new List<object>();
```
bahusoid: This change seems unnecessary (same for async version)

fredericDelaporte: This change results from the merge, not from a manual change during conflict resolution.
Reverting it would mean reverting a part of a change done in 5.2.6.
This change comes from [here](https://github.com/nhibernate/nhibernate-core/commit/c9ab94d315b56dc913745592ff052481f034f2a8#diff-cc4750cddf1d3379d1c9ea40b4256323R94), it has been done to avoid a cast or an array type coercion. It is not unnecessary, and already validated by #2128 review.

bahusoid: As you can see there is no `Select(o => (object) o)` removal on master because it's already removed [here](https://github.com/nhibernate/nhibernate-core/commit/5ee4401702a4f8eeddc05c7f09646bf14ea3ecdd#diff-cc4750cddf1d3379d1c9ea40b4256323R58) in #2161 without this change. So this change is not necessary on master and could be removed by squashing it in merge commit.

releasenotes.txt
hazzik: It seems that async regen should be embedded to the merge commit itself. And release notes to be on own commit.

fredericDelaporte: I am not used to edit files not having any conflicts in a merge commit. I was even not thinking this possible. I will check.
(Or maybe add in my local 5.2.x a commit for this prior to the merge, then drop it from my local 5.2.x ? Looks too hackish, and even worst.)

hazzik: You edit files before merge (there was a conflict or --no-commit option used) and then you `git add <path>` the manually edited files; then `git commit`.

src/NHibernate/Cache/ReadOnlyCache.cs
fredericDelaporte: Ok, done.

releasenotes.txt
fredericDelaporte: Ok, thanks for the hint.

fredericDelaporte:
Review comments have been addressed.

PR #2209, opened 2019-09-12
IN clause support in hql for composite keys on databases without row value constructor support

bahusoid:
Port hibernate [`mutateRowValueConstructorSyntaxInInListSyntax`](https://github.com/hibernate/hibernate-orm/blob/e5dc635a52362f69b69acb8d5b166b69b165dbbd/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/InLogicOperatorNode.java#L119).

Allows to use `IN (compositeId1, compositeId2)`  in hql  (and `Contains` queries in LINQ) for DBs like SQL Server.

~Depends on #2211 (without it "mutate" logic applies for all dialects)~

### DIFF
--- a/src/NHibernate/Hql/Ast/ANTLR/Tree/BinaryLogicOperatorNode.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/Tree/BinaryLogicOperatorNode.cs
@@ -200,8 +200,7 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 			return embeddedParameters.ToArray();
 		}
 
-		//TODO 6.0: Change to private protected (C#7.2 feature)
-		internal string Translate(int valueElements, string comparisonText, string[] lhsElementTexts, string[] rhsElementTexts)
+		private protected string Translate(int valueElements, string comparisonText, string[] lhsElementTexts, string[] rhsElementTexts)
 		{
 			var multicolumnComparisonClauses = new string[valueElements];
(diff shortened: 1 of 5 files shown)


hazzik:
>Depends on #2158 ( SupportsRowValueConstructorSyntaxInInList part)

Can you please unwind this dependency?

src/NHibernate/Hql/Ast/ANTLR/Tree/BinaryLogicOperatorNode.cs
```diff
@@ -200,17 +200,19 @@ public IParameterSpecification[] GetEmbeddedParameters()
 			return embeddedParameters.ToArray();
 		}
 
-		private string Translate(int valueElements, string comparisonText, string[] lhsElementTexts, string[] rhsElementTexts)
+		//TODO 6.0: Change to private protected (C#7.2 feature)
```
hazzik: You can change the project settings. To enable 7.2. I think the build agent should handle it.

bahusoid:
> Can you please unwind this dependency?
I created separate PR #2211

src/NHibernate/Hql/Ast/ANTLR/Tree/BinaryLogicOperatorNode.cs
bahusoid: Done. Will see if all is OK with CI builds

PR #2210, opened 2019-09-12
Support LINQ GroupBy subquery

### LINKED ISSUE #1124: NH-3155 - Linq subquery with group is not supported
**Alexander Zaytsev** created an [issue](https://nhibernate.jira.com/browse/NH-3155) — <time datetime="2012-05-20T06:35:08.793Z" title="2012-05-20T06:35:08.793Z">20th May 2012, 6:35:08</time>:
> Following query throws an "NotImplementedException".
> 
> 
> ```csharp
> var sq = db.Orders
>     .GroupBy(x => x.ShippingDate)
>     .Select(x => x.Max(o => o.OrderId));
> ```
> 
> 
> 
> ```csharp
> var result = db.Orders
>     .Where(x => sq.Contains(x.OrderId))
>     .Select(x => x.OrderId)
>     .Count();
> ```
> 
> 
> While identical HQL query does the work:
> 
> 
> 
> ```csharp
> var q = session.CreateQuery("select count(o.OrderId) from Order o where o.OrderId in (select max(o1.OrderId) from Order o1 group by o1.ShippingDate)").UniqueResult();
> ```
> 
> 
> 
---
**Oskar Berggren** added a comment — <time datetime="2012-11-19T21:13:30.399Z" title="2012-11-19T21:13:30.399Z">19th November 2012, 21:13:30</time>:
> The NotImplementedException is from AggregatingGroupByRewriter that intends to flatten the subquery generated by the group by inside the "explicit" group by. This fails because Contains() is handled as a result operator inside the query model of the subquery, which the AggregatingGroupByRewriter doesn't recognize and so it throws.
> 
> Whitelisting the CointainsResultOperator is not enough, because it will now be preceded by the GroupBy result operator. This throws from inside relinq at a later state - I think this is because the GroupBy generates a different type than the Contains operator expects. The intervening Max operator has already been converted to an expression in the select clause.
> 
> I have a candidate fix that involves rewriting the Contains operator from the SubQueryExpression to an NhContainsExpression (new) in the outer query.
> 
---
**Alexander Zaytsev** added a comment — <time datetime="2012-11-19T21:33:24.927Z" title="2012-11-19T21:33:24.927Z">19th November 2012, 21:33:24</time>:
> I have a semi-fixed version somewhere, but I'm not sure. The problem that with Contains we have to create a new copy of that, because it is not preserving the type of query.
> 
---
**Oskar Berggren** added a comment — <time datetime="2012-11-19T23:27:38.319Z" title="2012-11-19T23:27:38.319Z">19th November 2012, 23:27:38</time>:
> The exception I get if I just white-list ContainsResultOperator in AggregatingGroupByRewriter is from ContainsResultOperator.GetOutputDataInfo(), since the type output from the inserted GroupResultOperator() (i.e. IGrouping<something>) isn't compatible with the type of the item in the ContainsResultOperator (e.g. Int64).
> 
> It seems we must either inject some custom result operator in between that will coerce the type but which shouldn't need to affect the generated HQL, or remove the ContainsResultOperator entirely by rewriting it to a custom expression in the predicate of the parent query's body clauses. An attempt at the latter is available at https://github.com/oskarb/nhibernate-core/tree/NH-3155, but it now causes trouble with other types of Contains() found in WhereTests.cs.
> 
---
**Oskar Berggren** added a comment — <time datetime="2012-11-24T14:11:37.860Z" title="2012-11-24T14:11:37.860Z">24th November 2012, 14:11:37</time>:
> At https://github.com/oskarb/nhibernate-core/tree/NH-3155-alt2 is an alternative attempt at a solution, by overriding ContainsResultOperator to remove the type check. It does feel very hackish.
> 
---
**dthompson** added a comment — <time datetime="2012-11-28T15:38:45.807Z" title="2012-11-28T15:38:45.807Z">28th November 2012, 15:38:45</time>:
> The fix branch https://github.com/oskarb/nhibernate-core/tree/NH-3155-alt2 doesn't work for the linked issue https://nhibernate.jira.com/browse/NH-3154 not that it was supposed to but just an fyi. I ran into 3154 as well which would be nice to be able to do. Thanks for all the hard work.
>

bahusoid:
Fixes #1124

Yet another workaround to make LINQ GroupBy subquery works. IMHO it's less intrusive than last @oskarb  attempt in https://github.com/oskarb/nhibernate-core/tree/NH-3155-alt2. And doesn't look very hackish to me - allows type mismatch and doesn't fail in specific places.

### DIFF
--- a/src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
+++ b/src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
@@ -365,7 +365,7 @@ namespace NHibernate.Linq.Visitors
 		public override void VisitResultOperator(ResultOperatorBase resultOperator, QueryModel queryModel, int index)
 		{
 			PreviousEvaluationType = CurrentEvaluationType;
-			CurrentEvaluationType = resultOperator.GetOutputDataInfo(PreviousEvaluationType);
+			CurrentEvaluationType = GetOutputDataInfo(resultOperator, PreviousEvaluationType);
 
 			if (resultOperator is ClientSideTransformOperator)
 			{
@@ -380,6 +380,26 @@ namespace NHibernate.Linq.Visitors
 			}
 
 			ResultOperatorMap.Process(resultOperator, this, _hqlTree);
+		}
+
+		private static IStreamedDataInfo GetOutputDataInfo(ResultOperatorBase resultOperator, IStreamedDataInfo evaluationType)
+		{
+			//ContainsResultOperator contains data integrity check so for `values.Contains(x)` it checks that  'x' is proper type to be used inside 'values.Contains()'
+			//Due to some reasons (possibly NH expression rewritings) those types might be incompatible (known case NH-3155 - group by subquery). So resultOperator.GetOutputDataInfo throws something like: 
+			//System.ArgumentException : The items of the input sequence of type 'System.Linq.IGrouping`2[System.Object[],EntityType]' are not compatible with the item expression of type 'System.Int32'.
+			//Parameter name: inputInfo
+			//at Remotion.Linq.Clauses.ResultOperators.ContainsResultOperator.GetOutputDataInfo(StreamedSequenceInfo inputInfo)
+			//But in this place we don't really care about types involving inside expression, all we need to know is operation result which is bool for Contains
+			//So let's skip possible type exception mismatch if it allows to generate proper SQL
+			switch (resultOperator)
+			{
+				case ContainsResultOperator _:
+				case AnyResultOperator _:
+				case AllResultOperator _:
+					return new StreamedScalarValueInfo(typeof(bool));
+			}
+
+			return resultOperator.GetOutputDataInfo(evaluationType);
 		}
 
 		public override void VisitSelectClause(SelectClause selectClause, QueryModel queryModel)
--- a/src/NHibernate.Test/Async/Linq/NestedSelectsTests.cs
+++ b/src/NHibernate.Test/Async/Linq/NestedSelectsTests.cs
@@ -362,5 +362,38 @@ namespace NHibernate.Test.Linq
 			Assert.That(orders.Count, Is.EqualTo(830));
 			Assert.That(orders[0].OrderLinesIds, Is.Empty);
 		}
+
+		[Test]
+		public async Task NoNestedSelects_AnyOnGroupBySubqueryAsync()
+		{
+			var subQuery = from vms in db.Animals
+							group vms by vms.Father
+							into vmReqs
+							select vmReqs.Max(x => x.Id);
+
+			var outerQuery = from vm in db.Animals
+							where subQuery.Any(x => vm.Id == x)
+							select vm;
+			var animals = await (outerQuery.ToListAsync());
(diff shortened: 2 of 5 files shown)


src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -382,6 +382,26 @@ public override void VisitResultOperator(ResultOperatorBase resultOperator, Quer
...
+			//Due to some reasons (possibly NH expression rewritings) those types might be incompatible (known case NH-3155 - group by subquery). So resultOperator.GetOutputDataInfo throws something like: 
+			//System.ArgumentException : The items of the input sequence of type 'System.Linq.IGrouping`2[System.Object[],EntityType]' are not compatible with the item expression of type 'System.Int32'.
+			//Parameter name: inputInfo
+			//at Remotion.Linq.Clauses.ResultOperators.ContainsResultOperator.GetOutputDataInfo(StreamedSequenceInfo inputInfo)
+			//But in this place we don't really care about types involving inside expression, all we need to know is operation result which is bool for Contains
+			//So let's skip possible type exception mismatch if it allows to generate proper SQL
+			switch (resultOperator)
+			{
+				case ContainsResultOperator _:
+				case AnyResultOperator _:
```
bahusoid: `AnyResultOperator ` actually currently doesn't throw on type mismatch. But it looks like omission in ReLinq as I see no reasons why it should behave any differently. So added it just in case if it's changed in ReLinq.

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -382,6 +382,26 @@ public override void VisitResultOperator(ResultOperatorBase resultOperator, Quer
...
+		private static IStreamedDataInfo GetOutputDataInfo(ResultOperatorBase resultOperator, IStreamedDataInfo evaluationType)
+		{
+			//ContainsResultOperator contains data integrity check so for `values.Contains(x)` it checks that  'x' is proper type to be used inside 'values.Contains()'
+			//Due to some reasons (possibly NH expression rewritings) those types might be incompatible (known case NH-3155 - group by subquery). So resultOperator.GetOutputDataInfo throws something like: 
+			//System.ArgumentException : The items of the input sequence of type 'System.Linq.IGrouping`2[System.Object[],EntityType]' are not compatible with the item expression of type 'System.Int32'.
+			//Parameter name: inputInfo
+			//at Remotion.Linq.Clauses.ResultOperators.ContainsResultOperator.GetOutputDataInfo(StreamedSequenceInfo inputInfo)
+			//But in this place we don't really care about types involving inside expression, all we need to know is operation result which is bool for Contains
+			//So let's skip possible type exception mismatch if it allows to generate proper SQL
+			switch (resultOperator)
```
hazzik: Instead of doing it here, can we fix the condition at the source?

bahusoid: I don't really understand the real source of the issue. You should know this parts much better than me. And as this issue hangs since 2012 I think it's worth to implement workaround.

PR #2212, opened 2019-09-15
Fix build menu for running NUnit tests with 3.10.0

fredericDelaporte:
#2171 has updated nunit to its 3.10 version. But the build menu has the nunit console runner version hard-coded in it, and it was not updated.

This PR updates it, and also dodges an NUnit issue appeared in 3.8: nunit/nunit-console#418.


ShowBuildMenu.bat
```diff
@@ -182,7 +182,7 @@ SET NUNITPLATFORM=
 goto test-run
 
 :test-run
-start "nunit3-console" cmd /K %NUNIT% %NUNITPLATFORM% --agents=1 --process=separate NHibernate.nunit
+start "nunit3-console" cmd /K %NUNIT% %NUNITPLATFORM% --agents=1 NHibernate.nunit
```
fredericDelaporte: Removing that option default to multiple, which by default runs assemblies in parallel. But along with the `agents=1`, they will still be run sequentially, so it is not an issue to remove the `process` option.

PR #2213, opened 2019-09-15
Add a shortcut to reduce Transaction.Current reads

fredericDelaporte:
Since some users complain of a reduction performance with transactions and NHibernate v5.x, I have a look in which changes may have an impact. See [here](https://groups.google.com/forum/#!topic/nhusers/7C6z352w7ro) and [here](https://groups.google.com/forum/#!topic/nhusers/c6TUQwj3Ze0) for the reports on performance losses.

In v4.x and previous versions, there was a shortcut avoiding to read `Transaction.Current` when the session was already having a transaction context.

So I have made a bench for comparing accessing the session transaction context vs the current transaction. The later is around twenty times slower on my setup.

As a consequence I propose re-adding that shortcut.

(Still the first thread states the negative impact occurs also with regular transactions, while upgrading from NHv3.3. But with regular transactions, there is never a transaction context on the session, so the shortcut is not taken, and the current transaction is read. And in v3.3, the default transaction factory was already the system transaction one. The only explanation I could have would be that the performance of reading the current transaction itself has also dropped with .Net Framework v4, those old projects upgrades having probably also upgraded from an old framework.)

### DIFF
--- a/src/NHibernate.Test/Async/SystemTransactions/SystemTransactionFixture.cs
+++ b/src/NHibernate.Test/Async/SystemTransactions/SystemTransactionFixture.cs
@@ -10,15 +10,16 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Transactions;
 using NHibernate.Cfg;
 using NHibernate.Driver;
 using NHibernate.Engine;
-using NHibernate.Linq;
 using NHibernate.Test.TransactionTest;
 using NUnit.Framework;
+using NHibernate.Linq;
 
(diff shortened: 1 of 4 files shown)


fredericDelaporte:
CodeFactor:

> Insert parentheses within the conditional AND and OR expressions to declare the operator precedence.

Seriously? That is beginner level basic knowledge... I am reluctant to add parenthesis for making explicit such a trivial operator precedence.

hazzik:
It's really hard to read because the logic is inverted (gorge of negations and conjunctions). Adding parenthesis will not improve readability though. Either invert if or extract a method. I would prefer later one.

PR #2214, opened 2019-09-15
Allow configuring auto-join transaction globally

fredericDelaporte:
For application willing to enlist in current system transaction only explicitly, it would be better to be able to configure the auto-join transaction option globally rather than at each session opening.

This PR add the option as a configuration option.

Disabling auto-join may improve performance for applications which are not using transaction scopes, or which are using them seldom and takes care of re-enabling auto-join or joining explicitly in those cases.

### DIFF
--- a/src/NHibernate.Test/SystemTransactions/AutoJoinSettingFixture.cs
+++ b/src/NHibernate.Test/SystemTransactions/AutoJoinSettingFixture.cs
@@ -1,0 +1,55 @@
+using System.Transactions;
+using NHibernate.Cfg;
+using NUnit.Framework;
+
+namespace NHibernate.Test.SystemTransactions
+{
+	[TestFixture(true)]
+	[TestFixture(false)]
+	[TestFixture(new object[] { null })]
+	public class AutoJoinSettingFixture : TestCase
+	{
+		private readonly bool? _autoJoinTransaction;
+
+		public AutoJoinSettingFixture(bool? autoJoinTransaction)
+		{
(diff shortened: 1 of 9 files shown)


src/NHibernate.Test/SystemTransactions/AutoJoinSettingFixture.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace NHibernate.Test.SystemTransactions
+{
+	[TestFixture(true)]
+	[TestFixture(false)]
+	[TestFixture(new object[] { null })]
+	public class AutoJoinSettingFixture : TestCase
+	{
+		private readonly bool? _autoJoinTransaction;
+
+		public AutoJoinSettingFixture(bool? autoJoinTransaction)
```
hazzik: You can use `[Values]` here: https://github.com/nunit/docs/wiki/Values-Attribute

src/NHibernate.Test/SystemTransactions/AutoJoinSettingFixture.cs
```diff
@@ -0,0 +1,55 @@
...
+
+		[Test]
+		public void CheckTransactionJoined()
+		{
+			using (new TransactionScope())
+			using (var s = OpenSession())
+			{
+				Assert.That(
+					s.GetSessionImplementation().TransactionContext,
+					_autoJoinTransaction != false ? Is.Not.Null : Is.Null);
```
fredericDelaporte: Granted, I should care more about avoiding negating.

PR #2215, opened 2019-09-15
Fix ShowBuildMenu.sh

hazzik:


hazzik:
@lillo42 please test.

PR #2224, opened 2019-09-23
Skip lookup for non cachable LINQ query plan

bahusoid:
This lookup looks [quite heavy](https://github.com/nhibernate/nhibernate-core/blob/48b3bb9d433e411344bd2b7e50c88b4f57e607ec/src/NHibernate/Util/SoftLimitMRUCache.cs#L73-L88) and involves locking ( actually looks like it contains unnecessary double locking - I will create additional PR for this).


bahusoid:
I've just realized it kills all the benefits for the case when plan is actually cached :)

PR #2226, opened 2019-09-23
Use DateTime.UtcNow for timestamps

bahusoid:
as it's quicker to access than `DateTime.Now`


bahusoid:
> Is it worth it?

It actually just feels more proper to me to use  Utc in those places as it doesn't have unexpected time shifts (like due to daylight saving time). It seems mentioned places should be affected by time shifts anyway.

PR #2229, opened 2019-09-25
Cache query plan for DML LINQ queries

### LINKED ISSUE #2222: NHibernate Internal QueryPlanCache Dml Style Expression is not cached
The query plan cache is never hit for linq DML queries.

```c#
ISession.Query<StmtAccountStat>().Where(x => x.AccountNo == "1311").Update(x => new StmtAccountStat() { FirstDelayDate = DateTime.Now });
```
NHLinqExpression.cs
image

**p2 is Entity Class we should ignore in ExpressionParameterVisitor.cs VisitConstant
p5 is the property name that we want to update should be ignored
DmlExpressionRewriter.cs ConvertAssignmentsToDictionaryExpression method below code adds extra 2 parameter p3=1, p4=1 should be ignored**

```c#
var listInit = Expression.ListInit(
				Expression.New(
					DictionaryConstructorInfo,
					Expression.Condition(
						Expression.Equal(param, Expression.Constant(null, typeof(TSource))),
						Expression.Constant(assignments.Count),
						Expression.Constant(assignments.Count))),
				inits);
```
```
IQueryExpression.Key
UPDATE .Select[**StmtAccountStat**, Issuing.Data, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null,**System.Collections.Generic.Dictionary**`2[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Object, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e](.Where[StmtAccountStat, Issuing.Data, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null](NHibernate.Linq.NhQueryable`1[StmtAccountStat], Quote((x, ) => (String.op_Equality(x.AccountNo, p1))), ), Quote((, ) => (new System.Collections.Generic.Dictionary`2[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Object, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e(**Equal(, NULL) ? p3 : p4, )p5**p6)), )
```

bahusoid:
Fixes #2222

Maybe I'm wrong but it seems cached constants are not possible with DML queries (as no result transformers are supported). So let's just ignore `CanCachePlan` check for DML queries.

Regarding  changes in #1544. I think of another fix for cached constants - we should "parametrize" them. So that such queries should be cachable. I mean  for query like:
```C#
from c in db.Customers
where c.CustomerId == "ALFKI"
select new { c.CustomerId, c.ContactName, Constant = 1 }
```
Currently transformer is generated with embedded constants inside. Which for query above should generate something like:
```C#
public object ItemTransfomer(object[] row)
{
return new{CustomerId = row[0], ContactName = row[1], Constant = 1};
}
```
Instead we should generate something like this:
```C#
public object ItemTransfomer(object[] row, IDictionoray<string, object> parameters)
{
return new{CustomerId = row[0], ContactName = row[1], Constant = (int)parameters["p1"]};
}
```
And provide `NhLinqExpression.ParameterValuesByName` for `parameters` parameter.

### DIFF
--- a/src/NHibernate.Test/Linq/ConstantTest.cs
+++ b/src/NHibernate.Test/Linq/ConstantTest.cs
@@ -271,17 +271,33 @@ namespace NHibernate.Test.Linq
 
 			using (session.BeginTransaction())
 			{
+				db.Customers.Where(c => c.CustomerId == "UNKNOWN").Update(x => new Customer {CompanyName = "Constant1"});
 				db.Customers.Where(c => c.CustomerId == "ALFKI").Update(x => new Customer {CompanyName = x.CompanyName});
 				Assert.That(
 					cache,
-					Has.Count.EqualTo(1),
-					"First query plan should be cached.");
+					Has.Count.EqualTo(2),
+					"Query plans should be cached.");
 
 				using (var spy = new LogSpy(queryPlanCacheType))
 				{
-					// Should hit plan cache.
-					db.Customers.Where(c => c.CustomerId == "ANATR").Update(x => new Customer {CompanyName = x.CompanyName});
-					Assert.That(cache, Has.Count.EqualTo(1), "Second query should not cause a plan to be cached.");
+					//Queries below should hit plan cache.
+					using (var sqlSpy = new SqlLogSpy())
+					{
+						db.Customers.Where(c => c.CustomerId == "ANATR").Update(x => new Customer {CompanyName = x.CompanyName});
+						db.Customers.Where(c => c.CustomerId == "UNKNOWN").Update(x => new Customer {CompanyName = "Constant2"});
+
+						var sqlEvents = sqlSpy.Appender.GetEvents();
+						Assert.That(
+							sqlEvents[0].RenderedMessage,
+							Does.Contain("ANATR").And.Not.Contain("UNKNOWN").And.Not.Contain("Constant1"),
+							"Unexpected constant parameter value");
+						Assert.That(
+							sqlEvents[1].RenderedMessage,
+							Does.Contain("UNKNOWN").And.Contain("Constant2").And.Not.Contain("Constant1"),
+							"Unexpected constant parameter value");
+					}
+
+					Assert.That(cache, Has.Count.EqualTo(2), "Additional queries should not cause a plan to be cached.");
 					Assert.That(
 						spy.GetWholeLog(),
 						Does
@@ -289,7 +305,7 @@ namespace NHibernate.Test.Linq
 							.And.Not.Contain("unable to locate HQL query plan in cache"));
 
 					db.Customers.Where(c => c.CustomerId == "ANATR").Update(x => new Customer {ContactName = x.ContactName});
-					Assert.That(cache, Has.Count.EqualTo(2), "Third query should be cached");
+					Assert.That(cache, Has.Count.EqualTo(3), "Query should be cached");
 				}
 			}
 		}
--- a/src/NHibernate.Test/Linq/ConstantTest.cs
+++ b/src/NHibernate.Test/Linq/ConstantTest.cs
@@ -269,22 +269,28 @@ namespace NHibernate.Test.Linq
 					.GetValue(Sfi.QueryPlanCache);
 			cache.Clear();
 
-			db.Customers.Where(c => c.CustomerId == "ALFKI").Update(x => new Customer {CompanyName = x.CompanyName});
-			Assert.That(
-				cache,
-				Has.Count.EqualTo(1),
-				"First query plan should be cached.");
-
-			using (var spy = new LogSpy(queryPlanCacheType))
+			using (session.BeginTransaction())
 			{
-				// Should hit plan cache.
 				db.Customers.Where(c => c.CustomerId == "ALFKI").Update(x => new Customer {CompanyName = x.CompanyName});
-				Assert.That(cache, Has.Count.EqualTo(1), "Second query should not cause a plan to be cache.");
 				Assert.That(
-					spy.GetWholeLog(),
-					Does
-						.Contain("located HQL query plan in cache")
-						.And.Not.Contain("unable to locate HQL query plan in cache"));
+					cache,
+					Has.Count.EqualTo(1),
+					"First query plan should be cached.");
+
+				using (var spy = new LogSpy(queryPlanCacheType))
+				{
+					// Should hit plan cache.
+					db.Customers.Where(c => c.CustomerId == "ANATR").Update(x => new Customer {CompanyName = x.CompanyName});
+					Assert.That(cache, Has.Count.EqualTo(1), "Second query should not cause a plan to be cached.");
+					Assert.That(
+						spy.GetWholeLog(),
+						Does
+							.Contain("located HQL query plan in cache")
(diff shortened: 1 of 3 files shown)


hazzik:
Not sure if it is a good idea to cache a DML query.

hazzik:
Or is it about query plan cache?

bahusoid:
Yes. It's exactly about query plan cache.

src/NHibernate/Linq/NhLinqExpression.cs
```diff
@@ -88,13 +88,16 @@ public IASTNode Translate(ISessionFactoryImplementor sessionFactory, bool filter
...
+			if (QueryMode == QueryMode.Select && CanCachePlan)
+			{
+				CanCachePlan =
+					// If some constants do not have matching HQL parameters, their values from first query will
+					// be embedded in the plan and reused for subsequent queries: do not cache the plan.
+					!ParameterValuesByName
 					.Keys
 					.Except(requiredHqlParameters.Select(p => p.Name))
 					.Any();
+			}
```
hazzik: Shouldn't CanCachePlan become false here?

bahusoid: When? Fix simply ignores logic that was introduced #1544 and DML queries are **always** cached.

My understanding that "non-parameterized constants issue" is not relevant for DML queries as no result transformers with embedded constants are possible for DML.

hazzik: It is relevant. The original tests do not cover all cases of the issue and just focusing on the constants in select statement.

bahusoid: Can you provide an example?

bahusoid: If you are sure  that DML queries can be affected you can close this PR. But currently DML query plans are never cached as `requiredHqlParameters` contain only parameters for Select part (seems [intentional](https://github.com/nhibernate/nhibernate-core/blob/a41010f0611e93fc13c56af615b86b4bb90ad01d/src/NHibernate/Linq/DmlExpressionRewriter.cs#L103-L105) behavior)

hazzik: I tried to come up with a query, but I could not.

gokhanabatay: Cached linq has better than non cached, I don't figure out why but its quite slow than hql.

https://github.com/nhibernate/nhibernate-core/issues/2222#issuecomment-534267789

gokhanabatay:
Hi @bahusoid, @hazzik,
I think the issue with linq is reflection there is a lot of reflection in NHibernate code, it could be more faster if we use cached delegates.
Also there is a lot of MakeGenericMethod call that could be cached via method info and generic method types using **ConcurrentDictionary<MethodInfo, ConcurrentDictionary<Type, Delegate>>**
What do you think?

https://www.nuget.org/packages/DelegatesFactory/

Think about below code;
```c#
public virtual IQueryable CreateQuery(Expression expression)
{
	MethodInfo m = CreateQueryMethodDefinition.MakeGenericMethod(expression.Type.GetGenericArguments()[0]);

	return (IQueryable)m.Invoke(this, new object[] { expression });
}

//Without reflection
public virtual IQueryable CreateQuery(Expression expression)
{
        //ReflectionHelper.Get returns cached delegate for input method info and generic type parameters
	var query = (Func<INhQueryProvider, Expression, IQueryable>)ReflectionHelper.Get(CreateQueryMethodDefinition, expression.Type.GetGenericArguments()[0]);

	return query(this, expression);
}
```
https://internetexception.com/post/2016/09/02/Faster-than-Reflection-Delegates-Part-3

image

src/NHibernate.Test/Linq/ConstantTest.cs
```diff
@@ -257,6 +258,42 @@ public void PlansAreCached()
...
+					// Should hit plan cache.
+					db.Customers.Where(c => c.CustomerId == "ANATR").Update(x => new Customer {CompanyName = x.CompanyName});
+					Assert.That(cache, Has.Count.EqualTo(1), "Second query should not cause a plan to be cached.");
+					Assert.That(
+						spy.GetWholeLog(),
+						Does
+							.Contain("located HQL query plan in cache")
+							.And.Not.Contain("unable to locate HQL query plan in cache"));
+
+					db.Customers.Where(c => c.CustomerId == "ANATR").Update(x => new Customer {ContactName = x.ContactName});
```
fredericDelaporte: Since the removed check is about constant which may be cached in the query plan, the test should also check queries executing for the same plan do what they are supposed to do : executing with their own values rather than with the one from the previous query.
Can you please add some check reading back altered entities to ensure expected entities have been changed with expected values, while entities matching the previous query were not re-changed?

src/NHibernate/Linq/NhLinqExpression.cs
```diff
@@ -88,13 +88,16 @@ public IASTNode Translate(ISessionFactoryImplementor sessionFactory, bool filter
 
 			ParameterDescriptors = requiredHqlParameters.AsReadOnly();
 
-			CanCachePlan = CanCachePlan &&
-				// If some constants do not have matching HQL parameters, their values from first query will
-				// be embedded in the plan and reused for subsequent queries: do not cache the plan.
-				!ParameterValuesByName
+			if (QueryMode == QueryMode.Select && CanCachePlan)
```
fredericDelaporte: I know the cached constant bug was concerning projections, which are irrelevant to DML queries. Still this is a bit strange we need to ignored non parameterized constants: we should not have them in the first place I think. Maybe this fix dodges the real issue here: why do we have non-parameterized constants for DML queries? It is actually normal, or a bug which need to be fixed, instead of dodging the trouble?

src/NHibernate.Test/Linq/ConstantTest.cs
bahusoid: Added check on generated SQL instead

src/NHibernate/Linq/NhLinqExpression.cs
bahusoid: To me skipping unrelated logic looks more effective.

As I said here https://github.com/nhibernate/nhibernate-core/pull/2229#discussion_r328567660 it looks like intentional behavior with responsible code [here](https://github.com/nhibernate/nhibernate-core/blob/a41010f0611e93fc13c56af615b86b4bb90ad01d/src/NHibernate/Linq/DmlExpressionRewriter.cs#L103-L105). But I don't have time to investigate it thoroughly. IMHO it's not worth it.

PR #2233, opened 2019-09-30
Fix possible issue with async code for delayed entity inserts

bahusoid:
`ThreadStatic` shouldn't be used in async code.
I believe it should fix intermittent and annoying `MergeDeliveryNodeAsync` test fail on AppVeyor CI.

### DIFF
--- a/src/NHibernate/Action/DelayedPostInsertIdentifier.cs
+++ b/src/NHibernate/Action/DelayedPostInsertIdentifier.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 
 namespace NHibernate.Action
 {
@@ -15,22 +16,27 @@ namespace NHibernate.Action
 	[Serializable]
 	public class DelayedPostInsertIdentifier
 	{
-		private static long GloablSequence = 0;
-		private static readonly object SyncRoot = new object();
-
+		private static long GlobalSequence = 0;
 		private readonly long sequence;
 
 		public DelayedPostInsertIdentifier()
 		{
-			lock (SyncRoot)
-			{
-				if (GloablSequence == long.MaxValue)
-				{
-					GloablSequence = 0;
-				}
+			sequence = GetNextSequence();
+		}
 
-				sequence = GloablSequence++;
-			}
+		private static long GetNextSequence()
+		{
+			long initialValue, computedValue;
+			do
+			{
+				initialValue = GlobalSequence;
+				computedValue =
+					initialValue == long.MaxValue
+						? 1
+						: initialValue + 1;
+			} while (Interlocked.CompareExchange(ref GlobalSequence, computedValue, initialValue) != initialValue);
+
+			return computedValue;
 		}
 
 		public override bool Equals(object obj)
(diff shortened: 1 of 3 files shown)


src/NHibernate/Action/EntityIdentityInsertAction.cs
```diff
@@ -32,17 +33,6 @@ public EntityKey DelayedEntityKey
 			get { return delayedEntityKey; }
 		}
 
-		private EntityKey GenerateDelayedEntityKey()
-		{
-			lock (lockObject)
```
bahusoid: No need to lock inside ctor

src/NHibernate/Action/DelayedPostInsertIdentifier.cs
```diff
@@ -15,20 +16,27 @@ namespace NHibernate.Action
 	[Serializable]
 	public class DelayedPostInsertIdentifier
 	{
-		[ThreadStatic]
-		private static long _Sequence = 0;
+		private static long GlobalSequence = 0;
```
bahusoid: Initially I was going to replace `ThreadStatic` with `AsyncLocal`. But decided it's not really necessary here - let's make it global counter across all threads

hazzik: This is what I was going to ask.

src/NHibernate/Action/DelayedPostInsertIdentifier.cs
```diff
@@ -15,20 +16,27 @@ namespace NHibernate.Action
...
-				sequence = _Sequence++;
-			}
+				initialValue = GlobalSequence;
+				computedValue =
+					initialValue == long.MaxValue
+						? 1
+						: initialValue + 1;
+			} while (Interlocked.CompareExchange(ref GlobalSequence, computedValue, initialValue) != initialValue);
+
+			return computedValue;
```
hazzik: Why not just `Interlocked.Increase(ref GlobalSequence)`?

bahusoid: To avoid negative and zero values in keys.

bahusoid: I don't want how to implement counter reset with `Interlocked.Increase`'. I mean  to make counter after `long.MaxValue` generate 1.
Negative and zero values might not be the issue  -  but as it's public class I decided to keep old behavior.

bahusoid: On the other hand it's only exposed in `ToString`. So let me know if you think it's OK to have negative/zero sequence field value and I will change it to `Interlocked.Increment`

hazzik: I don't mind negative "sequences". If you don't like negative numbers in this context we can make the `sequence` of type `uint` or `ulong`.

In fact, I don't think we need _any_ sequence there: we could remove all fields from this class and use reference comparison only.

bahusoid: There is no `Interlocked.Increment` for unsigned types. And as this type is serialize I don't want to run into some strange issues due to inequality of deserialized keys because of reference comparison.
So let's simply use `Interlocked.Increment`

hazzik: Ok.

PR #2237, opened 2019-10-04
WIP Disable Pooling for Firebird

bahusoid:


bahusoid:
Was checking if it improves processing time. It doesn't

PR #2239, opened 2019-10-06
Simplify handling of lazy property groups

hazzik:
By moving out the decision of when include the property or not I was able to remove overloads which were added in #1922.

### DIFF
--- a/src/NHibernate/Hql/Ast/ANTLR/Tree/FromElement.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/Tree/FromElement.cs
@@ -337,9 +337,10 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 		/// <returns>the property select SQL fragment.</returns>
 		public string RenderPropertySelect(int size, int k)
 		{
-			return IsAllPropertyFetch
-				? _elementType.RenderPropertySelect(size, k, IsAllPropertyFetch)
-				: _elementType.RenderPropertySelect(size, k, _fetchLazyProperties);
+			return _elementType.RenderPropertySelect(
+				size,
+				k,
+				IsAllPropertyFetch || _fetchLazyProperties?.Contains(TableAlias) == true);
 		}
 
 		public override SqlString RenderText(Engine.ISessionFactoryImplementor sessionFactory)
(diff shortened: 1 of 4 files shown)


src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -1636,56 +1636,21 @@ public virtual string IdentifierSelectFragment(string name, string suffix)
...
-				.SetSuffix(suffix)
-				.SetUsedAliases(IdentifierAliases);
+			var select = new SelectFragment(Factory.Dialect)
+			              .SetSuffix(suffix)
+			              .SetUsedAliases(IdentifierAliases);
 
 			int[] columnTableNumbers = SubclassColumnTableNumberClosure;
 			string[] columnAliases = SubclassColumnAliasClosure;
 			string[] columns = SubclassColumnClosure;
-			HashSet<string> fetchColumnsAndFormulas = null;
```
hazzik: In my mind this code is equivalent to `fetchProperties?.Contains(name)`.

src/NHibernate/Hql/Ast/ANTLR/Tree/FromElement.cs
```diff
@@ -337,9 +337,10 @@ public string RenderIdentifierSelect(int size, int k)
 		/// <returns>the property select SQL fragment.</returns>
 		public string RenderPropertySelect(int size, int k)
 		{
-			return IsAllPropertyFetch
-				? _elementType.RenderPropertySelect(size, k, IsAllPropertyFetch)
-				: _elementType.RenderPropertySelect(size, k, _fetchLazyProperties);
+			return _elementType.RenderPropertySelect(
+				size,
+				k,
+				IsAllPropertyFetch || _fetchLazyProperties?.Contains(TableAlias) == true);
```
bahusoid: Didn't dig in to changes but on a first look it looks wrong.  Table alias and property names? I see no connection..

hazzik: `TableAlias` is a parameter which will go to `PropertySelectFragment` as "name". I just moved the decision to the caller.

hazzik: You're right. This is just nonsense.

hazzik: I've mistaken `name` for the "property" name, but it is a table name/alias.

PR #2242, opened 2019-10-09
Test case for NH-3972 - SQL error when selecting a column of a subclass when sibling classes have a column of the same name

lillo42:
Test case for  https://github.com/nhibernate/nhibernate-core/issues/1189


src/NHibernate/Persister/Entity/AbstractPropertyMapping.cs
```diff
@@ -94,11 +93,16 @@ public virtual string[] ToColumns(string propertyName)
 
 		protected void AddPropertyPath(string path, IType type, string[] columns, string[] formulaTemplates)
 		{
-			typesByPropertyPath[path] = type;
-			columnsByPropertyPath[path] = columns;
+			if (!typesByPropertyPath.ContainsKey(path))
```
lillo42: Would be better if we copy from is  Hibernate project?

https://github.com/hibernate/hibernate-orm/blob/d6376834a723f5147ba4152c1d5576fb500b3acb/hibernate-core/src/main/java/org/hibernate/persister/entity/AbstractPropertyMapping.java#L162-L242

bahusoid: I would say yes.
According to https://hibernate.atlassian.net/browse/HHH-11544 [(commit](https://github.com/hibernate/hibernate-orm/commit/865c159bff91c55bf9dfb708eada681b25dcf5e1)  your fix doesn't cover all cases:
> The reason for this, is that during the metamodel building hibernate currently resolves p.base to the first type it encounters i.e. either PolymorphicSub1 or PolymorphicSub2. This will make it work for one type, but fail for the other.

lillo42: @bahusoid Do we have interface ```MetadataImplementor```? 
If not, I believe we need change a lot of file and maybe it's better if we open other PR, or for this case will be better we fixes on NH 6?

bahusoid: > Do we have interface MetadataImplementor

I have no idea. 

> maybe it's better if we open other PR

Maybe I just don't understand your fix. Old logic - last path occurrence is used; your fix - first path occurrence is used. So doesn't it fix one case but breaks another (as I quoted in comment above)?
I mean will your work for all sub-classes:
```C#
var result = session.Query<DataRecord>().Select(x => new
				{
					x.Subject,
					State1 = ((Incident)x).State.Description,
					State2 = ((Change)x).State.Description,
				}).ToArray();
```

bahusoid (approved):
So suggested fix is not valid so let's commit test case only.

PR #2246, opened 2019-10-14
Fix QueryOver Coalesce projection in Select

### LINKED ISSUE #1107: NH-2983 - Coalesce in projection doesn't work if there is more than 1 Coalesce
**phillip.haydon** created an [issue](https://nhibernate.jira.com/browse/NH-2983) — <time datetime="2011-12-21T01:02:04.279Z" title="2011-12-21T01:02:04.279Z">21st December 2011, 1:02:04</time>:
> Originally posted: http://groups.google.com/group/nhusers/browse_thread/thread/868177e01816a5fb
> 
> ----------------------------
> 
> I'll write a test case tonight and add it.
> 
> ----------------------------
> 
> 
> I have a query here: 
>```c#
> var netSales = uow.Session 
>                  .QueryOver<TransactionSummary>(() => tsAlias) 
>                  .Where(x => x.EntityId == entityId) 
>                  .And(x => x.TransactionDate >= dayOpen) 
>                  .And(x => x.TransactionDate < dayClose) 
>                  .SelectList(x => x 
>                      .SelectSum(xx => xx.SalesExclGstExcl).WithAlias(()  => tsAlias.SalesExclGstExcl) 
>                      .SelectSum(xx => xx.TransactionCount).WithAlias(()  => tsAlias.TransactionCount)) 
>                  .TransformUsing(Transformers.AliasToBean(typeof(TransactionSummary))) 
>                  .Take(1) 
>                  .FutureValue(); 
>```
> This generates the following SQL 
> ```sql
> SELECT TOP (1 /** @p0 **/) sum(this*.Sales_Excl_GST_EXCL) as y0*, 
>                   sum(this*.TransactionCount)    as y1* 
>  FROM   dbo.tbTransactionSummary this_ 
>  WHERE  this_.EntityId = 678 /** @p1 **/ 
>         and this_.TransactionDate >= '2010-10-06T05:00:00.00' /** @p2 **/ 
>         and this_.TransactionDate < '2010-10-07T05:00:00.00' /** @p3 **/; 
> ```
> Which works perfectly. No issues. 
> 
> ------------------------ 
> 
> The problem is that if I'm looking at a date range that has no data,  it returns NULL and breaks. So far so good... 
> 
> If I add a Coalesce onto the Summed projections: 
> ```c#
> .SelectList(x => x 
>      .SelectSum(xx => xx.SalesExclGstExcl.Coalesce(0m)).WithAlias(() =>  tsAlias.SalesExclGstExcl) 
>      .SelectSum(xx => xx.TransactionCount.Coalesce(0)).WithAlias(() =>  tsAlias.TransactionCount)) 
>```
> This also generates the correct SQL: 
>```sql
> SELECT TOP (1 /** @p0 */) sum(coalesce(this_.Sales_Excl_GST_EXCL, 0 /*  @p1 **/)) as y0_, 
>                   sum(coalesce(this*.TransactionCount, 0 /* @p2 */))  as y2* 
>  FROM   dbo.tbTransactionSummary this_ 
>  WHERE  this_.EntityId = 678 /** @p3 **/ 
>         and this_.TransactionDate >= '2010-10-06T05:00:00.00' /** @p4 **/ 
>         and this_.TransactionDate < '2010-10-07T05:00:00.00' /** @p5 **/ 
>```
> Except if you look at the returned columns, the second one is named  'y2*' instead of 'y1*' 
> 
> So when the result gets back to NH it throws an exception: 
>```
> could not execute query 
> [ SELECT TOP (@p0)  sum(coalesce(this_.Sales_Excl_GST_EXCL, @p1)) as
> y0_, sum(coalesce(this_.TransactionCount, @p2)) as y2_ FROM
> dbo.tbTransactionSummary this_ WHERE this_.EntityId = @p3 and
> this_.TransactionDate >= @p4 and this_.TransactionDate < @p5 ]
>  Name:cp0 - Value:0  Name:cp1 - Value:0  Name:cp2 - Value:678
> Name:cp3 - Value:6/10/2010 5:00:00 AM  Name:cp4 - Value:7/10/2010
> 5:00:00 AM
> [SQL: SELECT TOP (@p0)  sum(coalesce(this_.Sales_Excl_GST_EXCL, @p1))
> as y0_, sum(coalesce(this_.TransactionCount, @p2)) as y2_ FROM
> dbo.tbTransactionSummary this_ WHERE this_.EntityId = @p3 and
> this_.TransactionDate >= @p4 and this_.TransactionDate < @p5]
>```

bahusoid:
Fixes #1107 

In fact simple `Select(x => x.Value.Coalesce(0))` was broken in QueryOver due to using `NHibernateUtil.Object` as return type for `coalesce` function which is something completely unrelated:
> Handles "any" mappings and the old deprecated "object" type.

### DIFF
--- a/src/NHibernate/Criterion/ProjectionsExtensions.cs
+++ b/src/NHibernate/Criterion/ProjectionsExtensions.cs
@@ -316,7 +316,7 @@ namespace NHibernate.Criterion
 		{
 			IProjection property = ExpressionProcessor.FindMemberProjection(methodCallExpression.Arguments[0]).AsProjection();
 			var replaceValueIfIsNull = ExpressionProcessor.FindMemberProjection(methodCallExpression.Arguments[1]);
-			return Projections.SqlFunction("coalesce", NHibernateUtil.Object, property, replaceValueIfIsNull.AsProjection());
+			return new SqlFunctionProjection("coalesce", returnTypeProjection: property, property, replaceValueIfIsNull.AsProjection());
 		}
 
 		/// <summary>
--- a/src/NHibernate/Criterion/SqlFunctionProjection.cs
+++ b/src/NHibernate/Criterion/SqlFunctionProjection.cs
@@ -115,7 +115,8 @@ namespace NHibernate.Criterion
 
 		public override IType[] GetTypes(ICriteria criteria, ICriteriaQuery criteriaQuery)
 		{
-			return new IType[] {GetReturnType(criteria, criteriaQuery)};
+			var type = GetReturnType(criteria, criteriaQuery);
+			return type != null ? new[] {type} : Array.Empty<IType>();
 		}
 
 		private IType GetReturnType(ICriteria criteria, ICriteriaQuery criteriaQuery)
--- a/src/NHibernate/Criterion/SqlFunctionProjection.cs
+++ b/src/NHibernate/Criterion/SqlFunctionProjection.cs
@@ -1,11 +1,10 @@
 using System;
-using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using NHibernate.Dialect.Function;
 using NHibernate.Engine;
 using NHibernate.SqlCommand;
 using NHibernate.Type;
-using NHibernate.Util;
 
 namespace NHibernate.Criterion
 {
@@ -16,6 +15,7 @@ namespace NHibernate.Criterion
 		private readonly ISQLFunction function;
 		private readonly string functionName;
 		private readonly IType returnType;
+		private readonly IProjection returnTypeProjection;
 
 		public SqlFunctionProjection(string functionName, IType returnType, params IProjection[] args)
 		{
@@ -28,6 +28,13 @@ namespace NHibernate.Criterion
 		{
 			this.function = function;
 			this.returnType = returnType;
+			this.args = args;
+		}
+
+		public SqlFunctionProjection(string functionName, IProjection returnTypeProjection, params IProjection[] args)
+		{
+			this.functionName = functionName;
+			this.returnTypeProjection = returnTypeProjection;
 			this.args = args;
 		}
 
@@ -108,9 +115,16 @@ namespace NHibernate.Criterion
 
 		public override IType[] GetTypes(ICriteria criteria, ICriteriaQuery criteriaQuery)
 		{
+			return new IType[] {GetReturnType(criteria, criteriaQuery)};
+		}
+
+		private IType GetReturnType(ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
 			ISQLFunction sqlFunction = GetFunction(criteriaQuery);
-			IType type = sqlFunction.ReturnType(returnType, criteriaQuery.Factory);
-			return new IType[] {type};
+
+			var resultType = returnType ?? returnTypeProjection?.GetTypes(criteria, criteriaQuery).FirstOrDefault();
+
+			return sqlFunction.ReturnType(resultType, criteriaQuery.Factory);
 		}
 
 		public override TypedValue[] GetTypedValues(ICriteria criteria, ICriteriaQuery criteriaQuery)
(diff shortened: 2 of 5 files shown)


src/NHibernate/Criterion/ProjectionsExtensions.cs
```diff
@@ -316,7 +316,7 @@ internal static IProjection ProcessCoalesce(MethodCallExpression methodCallExpre
 		{
 			IProjection property = ExpressionProcessor.FindMemberProjection(methodCallExpression.Arguments[0]).AsProjection();
 			var replaceValueIfIsNull = ExpressionProcessor.FindMemberProjection(methodCallExpression.Arguments[1]);
-			return Projections.SqlFunction("coalesce", NHibernateUtil.Object, property, replaceValueIfIsNull.AsProjection());
+			return new SqlFunctionProjection("coalesce", returnTypeProjection: property, property, replaceValueIfIsNull.AsProjection());
```
bahusoid: Other option is to use `GuessType` here  but current fix looks more proper to me (though it requires changes in `SqlFunctionProjection`). Let me know if you think `GuessType`  is enough here.

src/NHibernate/Criterion/SqlFunctionProjection.cs
```diff
@@ -107,10 +114,17 @@ private static SqlString GetProjectionArgument(ICriteriaQuery criteriaQuery, ICr
...
+			return new IType[] {GetReturnType(criteria, criteriaQuery)};
+		}
+
+		private IType GetReturnType(ICriteria criteria, ICriteriaQuery criteriaQuery)
 		{
 			ISQLFunction sqlFunction = GetFunction(criteriaQuery);
-			IType type = sqlFunction.ReturnType(returnType, criteriaQuery.Factory);
-			return new IType[] {type};
+
+			var resultType = returnType ?? returnTypeProjection?.GetTypes(criteria, criteriaQuery).FirstOrDefault();
```
fredericDelaporte: If there are more than one type returned, should it be silently ignored? (Shouldn't `SingleOrDefault()`be called here?)

And all the null coalesce, OrDefault stuff combined with yielding an `IType[]` of one element which may as a result be `null` does not look good. Digging on the `GetTypes` usage, it seems it ends up being used by calling methods on the elements it contains without checking if they are `null` or not. (See `CriteriaLoader.GetResultRow`.)
So either it should yield an empty type array when no type are determined if such an empty array can be a valid return value, or throw for the sake of failing earlier.
This defect was somehow already there, since nothing was checking the sqlFunction return type was not `null`. But it looks worsen by the change.

bahusoid: > If there are more than one type returned, should it be silently ignored? (Shouldn't SingleOrDefault() be called here?)

Well I don't know. I made it the way I saw somewhere else in Criteria handling. IMHO it's not important until multi-column support is added to `ISQLFunction.ReturnType`

> And all the null coalesce, OrDefault stuff combined with yielding an IType[] of one element which may as a result be null does not look good. ... it seems it ends up being used by calling methods

Not sure I properly understand it. But this logic ends up in `sqlFunction.ReturnType` which expects single `IType` and returns single `IType`.  And I see no restrictions why `null` should be forbidden for `sqlFunction.ReturnType` - function return type is known beforehand in most cases . So both `returnType` and `returnTypeProjection` should be nullable.

fredericDelaporte: The trouble is, `GetTypes` returned array is used as if its content could never be null, without null checks, see `CriteriaLoader.GetResultRow`. So potentially yielding  `new IType [] { null }` looks plain wrong.
The PR introduces more possibilities for this to happen.
I think is should either yield an empty array, or throw. It should not yield an array containing null. Or is there something I have missed which this returned value to be actually supported?

bahusoid: No but empty type collection also doesn't make much sense for using in Select projection... Ok - I've added null handling just in case.

PR #2249, opened 2019-10-15
Improve handling of SqlCeParameter.SqlDbType

hazzik:


src/NHibernate/Driver/SqlServerCeDriver.cs
```diff
@@ -12,6 +11,11 @@ namespace NHibernate.Driver
 	/// </summary>
 	public class SqlServerCeDriver : ReflectionBasedDriver
 	{
+		private static readonly Action<object, SqlDbType> SetSqlDbType =
+			DelegateHelper.BuildPropertySetter<SqlDbType>(
+				System.Type.GetType("System.Data.SqlServerCe.SqlCeParameter, System.Data.SqlServerCe", true),
```
bahusoid: Inside `ReflectionBasedDriver` we use `ReflectHelper.TypeFromAssembly` which seems contain some additional logic. I think it should be used here too.

hazzik: Ok

hazzik: Done

Issue #2251, opened 2019-10-16
Publish development nightly builds on nuget


bahusoid:
**Update** It's [live](https://www.myget.org/gallery/nhibernate). See how to start using development builds [here](https://github.com/nhibernate/nhibernate-core/issues/2251#issuecomment-626687676) 

With so long development cycle I think we should publish development versions on nuget as [pre-release packages](https://docs.microsoft.com/en-us/nuget/create-packages/prerelease-packages). Maybe once in two months just to make latest changes available for public testing.

Without release notes preparation or anything to minimize the burden. I guess just a link that compares previous branch version with dev version should suffice. Something like: https://github.com/nhibernate/nhibernate-core/compare/5.2.x...b845938db54388fdb4242e38cf393f70c2d5b4be

Suggested version naming:
5.3-dev.01

maca88:
Alternatively, we could use the new [github package registry](https://github.com/features/package-registry) for "nightly builds", which could be automatically published when a PR is merged into master by using a webhook to trigger a CI job that would publish the package. Something similar what Microsoft is doing for their libraries like [Roslyn on myget](https://dotnet.myget.org/feed/roslyn/package/vsix/7922692f-f018-45e7-8f3f-d3b7c0262841), without having to create a release on github.

fredericDelaporte:
In my quite stalled personal todo-list, I was having this too. I was thinking of doing it from AppVeyor builds, but if maca88 solution is easier to put in place, why not using it.

bahusoid:
I'm all in for automated nightly builds publishing if it's easy achievable. But what using github means for users? Am I right that they won't see such packages until they manually add github as package source?  I don't see it as big issue - just want to understand implications.

I also have zero experience with nuget publishing. So it's not quite clear to me how much time needed for manual publishing if we just publish manually version that's prepared by teamcity release build.
It seems the following steps are required:
1) Change NHibernate version to 5.3-dev (at least for TeamCity release build). I see it as one time operation - no need to change it incrementally for all builds
2) Publish pre-release on github (with name like `5.3-dev.01`) - just to tag source code. 
3) Publish manually nuget package prepared by TeamCity.

If it's only 2  steps (not counting one time first step) I think we can live with manual publishing. Or did I miss something?

hazzik:
[...]

PR #2254, opened 2019-10-19
Add dev build version suffix

bahusoid:
Related to #2251

With current PR state prepared [nuget package](https://teamcity.jetbrains.com/viewLog.html?buildId=2597421&buildTypeId=NHibernate_NHibernateReleasePackage&tab=artifacts&branch_NHibernate=pull%2F2254#%2FNHibernate-5.3%2Fnuget_gallery) looks like:
NHibernate.5.3.0-dev.2645.nupkg


hazzik:
I think we need to have different suffixes for PRs and master branch. The "NHibernate (Release)" build runs for all branches.

bahusoid:
> I think we need to have different suffixes for PRs and master branch. 

I'm not sure I understand the whole plan. Why? Are we going to publish PR binaries too?

hazzik:
>Are we going to publish PR binaries too?

Not sure, but why not?

hazzik:
Also, the publish step in TC could not be run only on specific branch.

bahusoid:
```xml
<arg value="/p:BuildNumber=&quot;${build.counter}&quot;" if="${property::exists('build.counter')}" />
```

The check above seems not to work. Any idea why? Team City still tries to get this property
https://teamcity.jetbrains.com/viewLog.html?buildId=2597215&buildTypeId=NHibernate_NHibernateReleasePackage&tab=buildLog&branch_NHibernate=pull%2F2254

bahusoid:
> Not sure, but why not?

Wouldn't it be excessive? All those temporary in progress pr binaries... I don't think publishing them is a good idea.

Also I'm currently not sure how to detect if it's PR build or not.

bahusoid:
So with current PR state prepared [nuget package](https://teamcity.jetbrains.com/viewLog.html?buildId=2597219&buildTypeId=NHibernate_NHibernateReleasePackage&tab=artifacts&branch_NHibernate=pull%2F2254#%2FNHibernate-5.3%2Fnuget_gallery) looks like:
NHibernate.5.3.0-dev.2642.nupkg

build-common/common.xml
```diff
@@ -11,15 +11,19 @@
 	<!-- Path to the folder that contain the external assemblies -->
 	<property name="lib.dir" value="lib" dynamic="true" />
 
-	<!-- This is used only for build folder -->
-	<!-- TODO: Either remove or refactor to use NHibernate.props -->
-	<property name="project.version" value="5.2.6" overwrite="false" />
-	<property name="project.version.numeric" value="5.2.6" overwrite="false" />
```
hazzik: `project.version.numeric` is used to build the documentation.

default.build
```diff
@@ -39,6 +40,8 @@
 			<arg value="/p:IncludeSymbols=&quot;True&quot;" />
 			<arg value="/p:IncludeSource=&quot;True&quot;" />
 			<arg value="/p:PackageOutputPath=&quot;${path::get-full-path(nuget.nupackages.dir)}&quot;" />
+			<!-- build.counter is not available by default. It needs to be exposed in Team City as system parameter system.build.counter with value %build.counter% -->
+			<arg value="/p:BuildNumber=&quot;${build.counter}&quot;" />
```
hazzik: I think it's better to add `if="${property::exists('build.counter')}"` element

bahusoid: It doesn't work. I've removed it in my last commit. See https://github.com/nhibernate/nhibernate-core/pull/2254#issuecomment-544115119

bahusoid:
> Also, the publish step in TC could not be run only on specific branch.

I guess we can  create copy of existing release build that will be run only on master branch with nuget publishing. And existing release build should be excluded for master branch.

build-common/common.xml
bahusoid: What's use of it? I see it's used [here](https://github.com/nhibernate/nhibernate-core/blob/ac39173567b31bcfad475ca32687c9faf0d37f87/doc/NHibernate.shfbproj.template#L26). But I don't see any representation of it in generated help files on master. And current help files also seems fine (except 5.2 version is stated in docs but it's hardcoded in xml).

default.build
bahusoid: I added skip if it's empty

maca88:
> Wouldn't it be excessive? All those temporary in progress pr binaries... I don't think publishing them is a good idea.

I have the same thinking, I think that we should start only with publishing master branch for now and later if we see the need to have also PRs included, we could do that later.

build-common/common.xml
bahusoid: I nuked it - everything seems OK. Are you sure we need it?

bahusoid: OK I see it's about "NHibernate API" docs which is not part of any build. But still I see no point in additional property - let's use `project.version` there too.

fredericDelaporte:
I am agree on not publishing packages for PRs, at least for now. But if possible, also include release fix branches (starting from 5.3.x).
Once we see how it fares with master and release branches "nightly builds", maybe we could reconsider whether it should also build PR, but for a start I think it is better to not build them.

bahusoid:
Guys I won't be able to do any coding in the following 3 weeks. So here some thoughts on how we can proceed further if anyone wants to step in.

*  It seems this PR is ready - it provides uniquely versioned binaries. And I think publishing can be implemented in another PR

* It seems publishing should be implemented inside nant build script but publishing parameters should be supplied from Team City.  We can create parametrized version of [nugetpush](https://github.com/nhibernate/nhibernate-core/blob/01effc3019eebb6c950b954e6e702e79ac00d013/default.build#L331) and use it for publishing.

For now I think of the following parameters that required on Team City (I've actuality already played with Team City parameters and added some of them to [realease build](https://teamcity.jetbrains.com/admin/editBuildParams.html?id=buildType:NHibernate_NHibernateReleasePackage))
`system.nuget.publishing`  - Allows to disable publishing for build. I think It also would be good if publishing is disabled by default in case manual build triggering (I tried to implement this behavior in param settings but not sure if it actually works as I intended). So user should explicitly enable publishing in custom build options
`system.nuget.branchNameRegex` - Regex matching if publishing is required for this branch. Currently supplied value most probably wrong
`system.build.branchName` - Branch name. Actually not sure what exactly is supplied by Team City
... (params for publishing like publishUrl, credentials and so on)

Params are available in nant as properties without `system.` prefix - `$(nuget.publishing)`
Parameters need to be set to some default value before using - otherwise other builds will fail. Something like:
```xml
<property name="nuget.publishing" value="false" overwrite="false" />
```

fredericDelaporte:
I do not get how we generate an actual release from this. I mean, NuGet package without the dev prefix, for actually releasing a new version.
Checking the release artifacts, I see the assembly has a 5.3-dev product version, while the NuGet is versioned 5.3.0-dev. Is this expected?

bahusoid:
> I do not get how we generate an actual release from this.

This pr should not affect release publishing. Just delete the suffix in release PR  and do your release procedure as always. Maybe we should also add logic to skip dev publishing if suffix is missed.

> I see the assembly has a 5.3-dev product version, while the NuGet is versioned 5.3.0-dev. Is this expected?

That's just default behavior. Yeah - I guess it's better to use 3 digit version for consistency.

bahusoid:
And after release with the next PR or as separate commit we should add dev suffix back and specify next version number. That's the plan.

hazzik:
>And after release with the next PR or as separate commit we should add dev suffix back and specify next version number. That's the plan.

Too much friction in my mind.

bahusoid:
I don't see it as to much friction. As I said above I think without dev prefix - dev publishing should be disabled. So there is no rush in making those changes.  After release PRs can be merged - they just won't be published.
And when we feel that it's time to publish and know next version  we add it along with dev suffix. IMHO it's the  right way to do it - published dev builds should have correct version.

default.build
```diff
@@ -39,6 +41,10 @@
 			<arg value="/p:IncludeSymbols=&quot;True&quot;" />
 			<arg value="/p:IncludeSource=&quot;True&quot;" />
 			<arg value="/p:PackageOutputPath=&quot;${path::get-full-path(nuget.nupackages.dir)}&quot;" />
+			<!-- build.counter is not available by default. It needs to be exposed in Team City as system parameter system.build.counter with value %build.counter% -->
```
hazzik: As the `build.counter` and `build.number` are not available in the TC. Maybe it's better to set `VersionSuffix` directly instead?

bahusoid: So release build will always generate dev versions? So how then generate actual release for actual releasing?
And then again dev builds may be published with wrong version  after release (it seems you don't see it as an issue though).

bahusoid: Maybe as alternative build should be done without dev suffix. And then as additional step with dev suffix if dev publishing is enabled. But I don't like the idea that after release both branches will publish dev builds under the same version

fredericDelaporte:
To reduce the friction, we may, in the same release package, generate both nuget with dev prefix and without.
The one with dev prefix would be pushed to MyGet, the other one not pushed anywhere excepted manually.
The binaries for the release zip should be generated without the dev prefix then.
We will still have to commit a version change after publishing a release in order to adjust the dev version. But at least we will no more need to changed the version in the last release PR release.
(I still like the current situation where we only change it in release PR, without any need to change it afterward, excepted that it causes the dev version to be currently misleading.)

bahusoid:
Your suggestions looks similar to [this](https://github.com/nhibernate/nhibernate-core/pull/2254#discussion_r336934379). And it adds additional full project recompilation + nuget packaging on each release build execution.

> We will still have to commit a version change after publishing a release in order to adjust the dev version.

So all it saves is one edit in release PR - dev suffix removal. IMHO not worth it.

Regarding updating two versions in release PR - you still need to manually adjust version in release branch. And you must know next version beforehand - I see it as additional inconvenience.

Bumping version commit is something that could be done easily form github UI without PR. I don't see why it's considered so much trouble and worth adding additional 30+ secs on each release build execution.

bahusoid:
No matter what we decide regarding publishing I think this PR is useful anyway. It minimizes frictions that are required during release -  version in common.xml is no longer need to be updated manually

And let's start with something - I believe it's a minimal changes that allow us to publish dev builds.

doc/NHibernate.shfbproj.template
```diff
@@ -23,7 +23,7 @@
     <FeedbackEMailLinkText>nhibernate team</FeedbackEMailLinkText>
     <HelpTitle>NHibernate API Reference</HelpTitle>
     <IndentHtml>False</IndentHtml>
-    <HelpFileVersion>${project.version.numeric}</HelpFileVersion>
+    <HelpFileVersion>${project.version}</HelpFileVersion>
```
fredericDelaporte, line 26: What does this change? It seems to me this help file version is hard to see anywhere. I have generated the api chm just to check, but I have not seen it anywhere when opening the file or checking its properties.

bahusoid: It's not about documentation. See discussion here https://github.com/nhibernate/nhibernate-core/pull/2254#discussion_r336746720. If I remember correctly it generates specification for API - takes methods declarations, comments from code and generates html from it. Something like it. It's not part of the build now.

bahusoid:
> ReleaseProcedure.txt needs an update. Some of its instructions are no more valid with this PR. And it should contain additions to tell how to generate the release packages (zip and NuGet, without dev suffix).

Ok will check it tomorrow.

> There is also a version into doc\reference\master.xml, which is still 5.2 while this PR bumps the version to 5.3. This causes the reference documentation generated by build to still be labeled 5.2, while the binaries are labelled 5.3.

I don't think we should care about generated documentation as long it's not published. So I think it's OK to update it only with release - just to minimize the hassle with bumping dev version.

doc/NHibernate.shfbproj.template
fredericDelaporte: Ok I have forgot about that property removal, and replacing it here is fine.

> I have generated the api chm

I know what this SHFB file is about. It is for generating the API chm, and that is what I have done to seek where this version was appearing. (Run `nant api` into `doc` folder. It takes a very long time.) So far I have not found it, but well, I now have the reason for the change, so it is ok for me.

Yes

fredericDelaporte:
> I don't think we should care about generated documentation as long it's not published. So I think it's OK to update it only with release - just to minimize the hassle with bumping dev version.

Why not, provided we have proper instructions into ReleaseProcedure.txt.

doc/NHibernate.shfbproj.template
bahusoid: Can you share it? Interesting to take a look bus was lazy to install some required apps.

appveyor.yml
```diff
@@ -1,4 +1,4 @@
-version: 5.2.7.{build}
+version: '{build}'
```
bahusoid, line 1: In attempt to minimize the frictions. I see no point in this - appveyour shows branch or tag if available. IMHO that's quite enough:
image

bahusoid:
> Why not, provided we have proper instructions into ReleaseProcedure.txt.

I have nothing to add except clearing VersionSuffix before release. Do you have something else in mind?

doc/NHibernate.shfbproj.template
fredericDelaporte: For some reason, the upload fails on GitHub. SO I have used some sharing service, which delete the file after 30 days without download. It is [here](http://dl.free.fr/ohxC4Q4SK).

build-common/NHibernate.props
```diff
@@ -2,14 +2,11 @@
   <Import Condition="Exists('NHibernate.dev.props')" Project="NHibernate.dev.props"/>
 
   <PropertyGroup>
-    <VersionMajor Condition="'$(VersionMajor)' == ''">5</VersionMajor>
-    <VersionMinor Condition="'$(VersionMinor)' == ''">2</VersionMinor>
-    <VersionPatch Condition="'$(VersionPatch)' == ''">7</VersionPatch>
-    <VersionSuffix Condition="'$(VersionSuffix)' == ''"></VersionSuffix>
+    <VersionPrefix Condition="'$(VersionPrefix)' == ''">5.3.0</VersionPrefix>
```
hazzik, line 5: This would require to manually support assembly version pinning: it should be 5.3.0 for entire 5.3.x branch.

bahusoid: Sorry - I don't get what it means? Could you please clarify? What would require it?

bahusoid: Got it.. )

hazzik: Updated comment. Hopefully it is clearer now.

bahusoid: Regex replace can handle it.

doc/NHibernate.shfbproj.template
bahusoid: Thanks!

build-common/NHibernate.props
hazzik: https://xkcd.com/1171/

build-common/common.xml
```diff
@@ -11,15 +11,26 @@
...
+    <xmlpeek
+      file="${root.dir}/build-common/NHibernate.props"
+      xpath="/Project/PropertyGroup/VersionPrefix"
+      property="version.prefix" />
+    <property name="project.version" value="${version.prefix}" />
+  </target>
+  
+  <target name="set-doc-version">
+    <xmlpoke
+      file="${root.dir}/doc/reference/master.xml"
```
fredericDelaporte, line 25: This causes the git "workspace" to get "dirty", as this file is not ignored by git. Could you add so final steps to reference build file targets in order to clean it after builds?

ReleaseProcedure.txt
```diff
@@ -25,9 +25,9 @@ Tools\gitreleasemanager\0.7.0\tools\GitReleaseManager.exe create -o nhibernate -
...
     formalism, the NuGet package will have an invalid link to release notes.
 
-  * Check/update version number in common.xml and NHibernate.props under
-    build-common folder, in master.xml under doc\reference folder, and in
-    appveyor.yml in the NHibernate root.
+  * Check/update version number in NHibernate.props under build-common folder.
+    Clear VersionSuffix if present.
+    NOTE: This paragraph is changed for 5.3 - if you are releasing earlier version please check instruction from appropriate branch   
 
   * Don't forget to commit the above.
```
fredericDelaporte, line 33: I think a final step should be added to finish the release : preparing the master branch for next version builds. And by the way I propose another missing step:

```
  * If this was the master branch, create the stable branch for the new release, named
    according to its version with "x" instead of its patch number. By example, "5.3.x".
  * Increment the version number minor in NHibernate.props under build-common folder
    and set VersionSuffix to "dev", and commit it directly (without going through a PR).
```

ReleaseProcedure.txt
```diff
@@ -25,9 +25,9 @@ Tools\gitreleasemanager\0.7.0\tools\GitReleaseManager.exe create -o nhibernate -
     If the release tag does not match the released version with major.minor.release
     formalism, the NuGet package will have an invalid link to release notes.
 
-  * Check/update version number in common.xml and NHibernate.props under
-    build-common folder, in master.xml under doc\reference folder, and in
-    appveyor.yml in the NHibernate root.
+  * Check/update version number in NHibernate.props under build-common folder.
+    Clear VersionSuffix if present.
+    NOTE: This paragraph is changed for 5.3 - if you are releasing earlier version please check instruction from appropriate branch
```
fredericDelaporte, line 30: I do not think we really need that note, since normally, one would read this file from the branch on which he is issuing a release PR.

build-common/NHibernate.props
bahusoid: I don't get hazzik point at all. And why overkill? It's single line change. With your suggested change we have to do additional modifications in common.xml to properly construct `${project.version}`. For what?

build-common/NHibernate.props
```diff
@@ -2,15 +2,15 @@
...
+    <!-- Clear VersionSuffix for making release and set it to dev for making development builds -->
+    <VersionSuffix Condition="'$(VersionSuffix)' == ''">dev</VersionSuffix>  
 
-    <VersionPrefix>$(VersionMajor).$(VersionMinor).$(VersionPatch)</VersionPrefix>
-    <AssemblyVersion>$(VersionMajor).$(VersionMinor).0.0</AssemblyVersion>
-    <FileVersion>$(VersionPrefix).0</FileVersion>
```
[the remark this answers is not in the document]
bahusoid: Maybe we need it back - I don't remember why I removed it in first place. Do you remember what kind of problems it causes? As now prepared file version contains also git commit - which I find pretty useful.

build-common/common.xml
bahusoid: git ignore wouldn't help here  as it only works for new files. Maybe we should rename this file (`common.xml.template?`) and copy it after modification? That way common.xml can be added to git ignore

fredericDelaporte: > Maybe we should rename this file (common.xml.template?) and copy it after modification? That way common.xml can be added to git ignore

The altered file is `doc/reference/master.xml`, so that is not `common.xml` which should have an ignored version with a template, but `master.xml`.
And I propose another solution : add a step after the build to change back the file to its initial value, avoiding it to be considered altered.

build-common/NHibernate.props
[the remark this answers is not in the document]
fredericDelaporte: Read more [here](https://stackoverflow.com/q/64602/1178314).

TLDR, we need to guarantee the file version changes for each release, so it must not be the assembly version, which does not change for patches. According to one answer of above SO link:

> The AssemblyFileVersion is optional. If not given, the AssemblyVersion is used.

But according to you, removing it caused it to instead contain more information. Better check what is going on here, in order to be sure we can rely on the behavior you have seen.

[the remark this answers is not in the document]
bahusoid: Sorry my mistake - it wasn't file version. It's product version contains git commit. I restored `FileVersion`

build-common/common.xml
bahusoid: I decided to go with additional file:
1) To keep all logic in one step;
2) To avoid dirty workspace in case of errors or build interruption;

ReleaseProcedure.txt
bahusoid: Added with some adjustments.  But I don't think that's last step is mandatory and should be always done during release (as next version might be yet undecided). Feel free to adjust it further directly.

ReleaseProcedure.txt
bahusoid: Removed

build-common/NHibernate.props
bahusoid: regex has been removed

ReleaseProcedure.txt
```diff
@@ -55,3 +54,10 @@ Tools\gitreleasemanager\0.7.0\tools\GitReleaseManager.exe create -o nhibernate -
   * If this was a stable branch, merge it forward to master. Perhaps some
     changes need to be edited out in the merge, but this will reduce the
     risk of some issue being left unfixed on master.
+ 
+  * If this was the master branch, create the stable branch for the new release, named
+    according to its version with "x" instead of its patch number. By example, "5.3.x".
+    
+  * If next version is decided set it in NHibernate.props under build-common folder
+    (in most cases - increment minor version) and set VersionSuffix to "dev".
+    Commit it directly (without going through a PR).
```
fredericDelaporte, line 63: I think there no harm to always bump to the "next" minor, even if we end up releasing a major and not that minor. This would avoid some confusion about whether the "dev" builds are a previous pre-release or already a "vNext".
And we always need to put the "dev" sufffix back in place, in my opinion.

build-common/NHibernate.props
```diff
@@ -2,13 +2,14 @@
...
+    <VersionPatch  Condition="'$(VersionPatch)'  == ''">0</VersionPatch>
+    <!-- Clear VersionSuffix for making release and set it to dev for making development builds -->
+    <VersionSuffix Condition="'$(VersionSuffix)' == ''">dev</VersionSuffix>
 
-    <VersionPrefix>$(VersionMajor).$(VersionMinor).$(VersionPatch)</VersionPrefix>
-    <AssemblyVersion>$(VersionMajor).$(VersionMinor).0.0</AssemblyVersion>
+    <VersionPrefix Condition="'$(VersionPrefix)' == ''">$(NhVersion).$(VersionPatch)</VersionPrefix>
+    <VersionSuffix Condition="'$(VersionSuffix)' != '' AND '$(BuildNumber)' != ''">$(VersionSuffix).$(BuildNumber)</VersionSuffix>
+    <AssemblyVersion>$(NhVersion).0.0</AssemblyVersion>
     <FileVersion>$(VersionPrefix).0</FileVersion>
```
fredericDelaporte, line 13: We are back to previous `FileVersion` handling, so ok on principle.

Nonetheless this PR is also for preparing the ability to push pre-release packages. So I wonder, shouldn't we also take care of releasing a unique `FileVersion` for these pre-releases? Otherwise, someone distributing a pre-release through an installer may end up not being able to upgrade it until we change at least the patch version.

So maybe change `.0` by `.*` if that works here, or put in fourth number the TC build number. Or maybe just stay like this, not supporting packing pre-releases. What do you think?

bahusoid: I've checked how it's done by microsoft on `Microsoft.Data.SqlClient` example:
https://www.nuget.org/packages/Microsoft.Data.SqlClient/2.0.0-preview3.20122.2

Yeah build number is part of file version - so yes let's add it too. Though they also don't bother with keeping assembly version stable for minor releases...

ReleaseProcedure.txt
bahusoid: >  bump to the "next" minor, even if we end up releasing a major and not that minor.

I'm not against it. But I think It might be confusing for users. They might still expect that minor version would be released along with major one.

> This would avoid some confusion about whether the "dev" builds 

As long those versions are not publicly published I think it's OK to have version mismatch for compiled libs. So 
* we shouldn't publish dev builds without "dev" suffix; 
* "dev" suffix must be set only with proper next version. 

So that way there is shouldn't be confusion with published versions. That was my initial plan.

But I'm fine with always bumping version commit plan too.

build-common/NHibernate.props
fredericDelaporte: Keeping assembly version stable is important with strong naming, which NHibernate still issues, while it seems many do no more bother with strong naming.
But ceasing to be strongly named is a breaking change, so we stay strongly named.

PR #2264, opened 2019-11-03
Fix duplicated methods generated in proxies

### LINKED ISSUE #2085: Duplicated methods generated in proxies
The static proxy factory may generate duplicated methods in proxies. This does not cause a crash, but instead, the first defined method is executed, ignoring the other ones.

This happen when the proxified type has many methods with the same signature, but coming from interfaces. In such case, these methods are all generated as regular class methods, instead of being generated as explicit interface implementation.

By example, the proxy generated for: https://github.com/nhibernate/nhibernate-core/blob/22cd7605f3031cd4cef8298554e02a6437867427/src/NHibernate.Test/StaticProxyTest/Model.cs#L28-L33

(`IEntity` defines the same members.)

Is currently:

```c#
[Serializable]
[StructLayout(LayoutKind.Auto, CharSet = CharSet.Auto)]
public class InterfacedEntityProxy : InterfacedEntity, ISerializable, INHibernateProxy, IEntity
{
	private ILazyInitializer __lazyInitializer;

	private NHibernateProxyFactoryInfo __proxyInfo;

	public InterfacedEntityProxy(ILazyInitializer _lazyInitializer, NHibernateProxyFactoryInfo _proxyInfo)
	{
		this.__lazyInitializer = _lazyInitializer;
		this.__proxyInfo = _proxyInfo;
	}

	public override Guid get_Id()
	{
		if (this.__lazyInitializer == null)
		{
			return base.Id;
		}
		return (Guid)this.__lazyInitializer.get_Identifier();
	}

	public override void set_Id(Guid guid)
	{
		if (this.__lazyInitializer == null)
		{
			base.Id = guid;
			return;
		}
		this.__lazyInitializer.Initialize();
		this.__lazyInitializer.set_Identifier(guid);
		((InterfacedEntity)this.__lazyInitializer.GetImplementation()).Id = guid;
	}

	public override string get_Name()
	{
		if (this.__lazyInitializer == null)
		{
			return base.Name;
		}
		return ((InterfacedEntity)this.__lazyInitializer.GetImplementation()).Name;
	}

	public override void set_Name(string name)
	{
		if (this.__lazyInitializer == null)
		{
			base.Name = name;
			return;
		}
		((InterfacedEntity)this.__lazyInitializer.GetImplementation()).Name = name;
	}

	public override string get_Text()
	{
		if (this.__lazyInitializer == null)
		{
			return base.Text;
		}
		return ((InterfacedEntity)this.__lazyInitializer.GetImplementation()).Text;
	}

	public override void set_Text(string text)
	{
		if (this.__lazyInitializer == null)
		{
			base.Text = text;
			return;
		}
		((InterfacedEntity)this.__lazyInitializer.GetImplementation()).Text = text;
	}

	public override string ToString()
	{
		if (this.__lazyInitializer == null)
		{
			return base.ToString();
		}
		return ((object)this.__lazyInitializer.GetImplementation()).ToString();
	}

	ILazyInitializer INHibernateProxy.get_HibernateLazyInitializer()
	{
		return this.__lazyInitializer;
	}

	public override Guid get_Id()
	{
		if (this.__lazyInitializer == null)
		{
			return base.Id;
		}
		return ((IEntity)this.__lazyInitializer.GetImplementation()).Id;
	}

	public override void set_Id(Guid id)
	{
		if (this.__lazyInitializer == null)
		{
			base.Id = id;
			return;
		}
		((IEntity)this.__lazyInitializer.GetImplementation()).Id = id;
	}

	public override string get_Name()
	{
		if (this.__lazyInitializer == null)
		{
			return base.Name;
		}
		return ((IEntity)this.__lazyInitializer.GetImplementation()).Name;
	}

	public override void set_Name(string name)
	{
		if (this.__lazyInitializer == null)
		{
			base.Name = name;
			return;
		}
		((IEntity)this.__lazyInitializer.GetImplementation()).Name = name;
	}

	public override string get_Text()
	{
		if (this.__lazyInitializer == null)
		{
			return base.Text;
		}
		return ((IEntity)this.__lazyInitializer.GetImplementation()).Text;
	}

	public override void set_Text(string text)
	{
		if (this.__lazyInitializer == null)
		{
			base.Text = text;
			return;
		}
		((IEntity)this.__lazyInitializer.GetImplementation()).Text = text;
	}

	public InterfacedEntityProxy(SerializationInfo serializationInfo, StreamingContext streamingContext)
	{
	}

	[SecurityCritical]
	public override void GetObjectData(SerializationInfo serializationInfo, StreamingContext context)
	{
		serializationInfo.SetType(typeof(NHibernateProxyObjectReference));
		new NHibernateProxyObjectReference(this.__proxyInfo, this.__lazyInitializer.get_Identifier(), this.__lazyInitializer.get_IsUninitialized() ? null : this.__lazyInitializer.GetImplementation()).GetObjectData(serializationInfo, context);
	}
}
```

It has two `get_Id`, but only the first one will be called when getting the `Id`, even if we do that through `IEntity`. The second one has a different implementation which would cause a lazy initialization.

Fixing this will be a possible breaking changes for cases where the user is relying on this bug, as in #2052.

fredericDelaporte:
Fixes #2085

### DIFF
--- a/src/NHibernate/Proxy/ProxyBuilderHelper.cs
+++ b/src/NHibernate/Proxy/ProxyBuilderHelper.cs
@@ -102,6 +102,13 @@ namespace NHibernate.Proxy
 
 		internal static IEnumerable<MethodInfo> GetProxiableMethods(System.Type type, IEnumerable<System.Type> interfaces)
 		{
+			if (type.IsInterface || type == typeof(object) || type.GetInterfaces().Length == 0)
+			{
+				return GetProxiableMethods(type)
+					.Concat(interfaces.SelectMany(i => i.GetMethods()))
+					.Distinct();
+			}
+
 			var proxiableMethods = new HashSet<MethodInfo>(GetProxiableMethods(type), new MethodInfoComparer(type));
 			foreach (var interfaceMethod in interfaces.SelectMany(i => i.GetMethods()))
 			{
@@ -295,24 +302,33 @@ namespace NHibernate.Proxy
 		/// </summary>
 		private class MethodInfoComparer : IEqualityComparer<MethodInfo>
 		{
-			private readonly System.Type _baseType;
-			private readonly HashSet<System.Type> _baseTypeInterfaces;
 			private readonly Dictionary<System.Type, Dictionary<MethodInfo, MethodInfo>> _interfacesMap;
-			private readonly bool _baseIsObjectOrInterface;
 
 			public MethodInfoComparer(System.Type baseType)
 			{
-				_baseType = baseType;
-				_baseIsObjectOrInterface = _baseType.IsInterface || _baseType == typeof(object);
-				_baseTypeInterfaces = new HashSet<System.Type>(baseType.GetInterfaces());
-				_interfacesMap = new Dictionary<System.Type, Dictionary<MethodInfo, MethodInfo>>(_baseTypeInterfaces.Count);
+				_interfacesMap = BuildInterfacesMap(baseType);
+			}
+
+			private static Dictionary<System.Type, Dictionary<MethodInfo, MethodInfo>> BuildInterfacesMap(System.Type type)
+			{
+				return type.GetInterfaces()
+					.Distinct()
+					.ToDictionary(i => i, i => ToDictionary(type.GetInterfaceMap(i)));
+			}
+
+			private static Dictionary<MethodInfo, MethodInfo> ToDictionary(InterfaceMapping interfaceMap)
+			{
+				var map = new Dictionary<MethodInfo, MethodInfo>(interfaceMap.InterfaceMethods.Length);
+				for (var i = 0; i < interfaceMap.InterfaceMethods.Length; i++)
+				{
+					map.Add(interfaceMap.InterfaceMethods[i], interfaceMap.TargetMethods[i]);
+				}
+
+				return map;
 			}
 
 			public bool Equals(MethodInfo x, MethodInfo y)
 			{
-				if (_baseIsObjectOrInterface)
-					return x == y;
-
 				if (x == y)
 					return true;
 				if (x == null || y == null)
@@ -327,22 +343,13 @@ namespace NHibernate.Proxy
 				// implementation of the other.
 				if (x.DeclaringType.IsInterface == y.DeclaringType.IsInterface)
 					return false;
+
 				var interfaceMethod = x.DeclaringType.IsInterface ? x : y;
 				// If the interface is not implemented by the base type, the method cannot be implemented by the
 				// base type method.
-				if (!_baseTypeInterfaces.Contains(interfaceMethod.DeclaringType))
+				if (!_interfacesMap.TryGetValue(interfaceMethod.DeclaringType, out var map))
 					return false;
 
-				if (!_interfacesMap.TryGetValue(interfaceMethod.DeclaringType, out var map))
-				{
-					var interfaceMap = _baseType.GetInterfaceMap(interfaceMethod.DeclaringType);
-					map = new Dictionary<MethodInfo, MethodInfo>(interfaceMap.InterfaceMethods.Length);
-					for (var i = 0; i < interfaceMap.InterfaceMethods.Length; i++)
-					{
-						map.Add(interfaceMap.InterfaceMethods[i], interfaceMap.TargetMethods[i]);
-					}
-					_interfacesMap.Add(interfaceMethod.DeclaringType, map);
-				}
 				var baseMethod = x.DeclaringType.IsInterface ? y : x;
 				return map[interfaceMethod] == baseMethod;
 			}
--- a/src/NHibernate/Proxy/ProxyBuilderHelper.cs
+++ b/src/NHibernate/Proxy/ProxyBuilderHelper.cs
@@ -136,8 +136,6 @@ namespace NHibernate.Proxy
 			return methodBuilder;
 		}
 
-		private static readonly System.Type[] _equalsParametersTypes = { typeof(object) };
-
 		internal static MethodBuilder GenerateMethodSignature(string name, MethodInfo method, TypeBuilder typeBuilder)
 		{
 			var explicitImplementation = method.DeclaringType.IsInterface;
@@ -146,8 +144,7 @@ namespace NHibernate.Proxy
 #pragma warning disable 618
 					typeBuilder.BaseType == typeof(ProxyDummy)) &&
 #pragma warning restore 618
-				(method.Name == "Equals" && method.GetParameters().Select(p => p.ParameterType).SequenceEqual(_equalsParametersTypes) ||
-					method.Name == "GetHashCode" && method.GetParameters().Length == 0))
+				(IsEquals(method) || IsGetHashCode(method)))
 			{
 				// If we are building a proxy for an interface, and it defines an Equals or GetHashCode, they must
 				// be implicitly implemented for overriding object methods.
@@ -205,6 +202,18 @@ namespace NHibernate.Proxy
 			}
 
 			return methodBuilder;
+		}
+
+		private static bool IsGetHashCode(MethodBase method)
+		{
+			return method.Name == "GetHashCode" && method.GetParameters().Length == 0;
+		}
+
+		private static bool IsEquals(MethodBase method)
+		{
+			if (method.Name != "Equals") return false;
+			var parameters = method.GetParameters();
+			return parameters.Length == 1 && parameters[0].ParameterType == typeof(object);
 		}
 
 		internal static void GenerateInstanceOfIgnoresAccessChecksToAttribute(
--- a/src/NHibernate/Proxy/ProxyBuilderHelper.cs
+++ b/src/NHibernate/Proxy/ProxyBuilderHelper.cs
@@ -14,6 +14,7 @@ using System.Reflection.Emit;
 using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
 using System.Security;
+using NHibernate.Proxy.DynamicProxy;
 using NHibernate.Util;
 
 namespace NHibernate.Proxy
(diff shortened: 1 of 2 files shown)


src/NHibernate/Proxy/ProxyBuilderHelper.cs
```diff
@@ -101,11 +102,12 @@ internal static IEnumerable<MethodInfo> GetProxiableMethods(System.Type type)
...
-			var proxiableMethods =
-				GetProxiableMethods(type)
-					.Concat(interfaces.SelectMany(i => i.GetMethods()))
-					.Distinct();
-			
+			var proxiableMethods = new HashSet<MethodInfo>(GetProxiableMethods(type), new MethodInfoComparer(type));
+			foreach (var interfaceMethod in interfaces.SelectMany(i => i.GetMethods()))
+			{
+				proxiableMethods.Add(interfaceMethod);
+			}
```
fredericDelaporte: I could have kept the previous code, just giving the comparer to `Distinct`. But its contract does not guarantee we will takes the base type methods rather than "duplicated" methods from the interfaces. (Its current implementation will keep the methods from the base type, but we should not rely on this.)
By explicitly adding the base methods to the hashset first, it is guaranteed that this is dups from interfaces which will be ignored.

src/NHibernate/Proxy/ProxyBuilderHelper.cs
```diff
@@ -134,25 +136,48 @@ internal static MethodBuilder GetObjectDataMethodBuilder(TypeBuilder typeBuilder
...
+
 		internal static MethodBuilder GenerateMethodSignature(string name, MethodInfo method, TypeBuilder typeBuilder)
 		{
-			//TODO: Should we use attributes of base method?
-			var methodAttributes = MethodAttributes.Public | MethodAttributes.HideBySig | MethodAttributes.Virtual;
+			var explicitImplementation = method.DeclaringType.IsInterface;
+			if (explicitImplementation &&
+				(typeBuilder.BaseType == typeof(object) ||
+#pragma warning disable 618
+					typeBuilder.BaseType == typeof(ProxyDummy)) &&
```
fredericDelaporte: Old dynamic proxy legacy. Not handling this causes its `InterfaceWithEqualsGethashcodeTests` test to fail.

src/NHibernate/Proxy/ProxyBuilderHelper.cs
```diff
@@ -134,25 +136,48 @@ internal static MethodBuilder GetObjectDataMethodBuilder(TypeBuilder typeBuilder
...
+				: name;
+			var methodBuilder =
+				typeBuilder.DefineMethod(
+					implementationName,
+					methodAttributes,
+					CallingConventions.HasThis,
+					method.ReturnType,
+					parameters.ToArray(param => param.ParameterType));
+			if (explicitImplementation)
+				methodBuilder.SetImplementationFlags(MethodImplAttributes.Managed | MethodImplAttributes.IL);
```
fredericDelaporte: Copied from `NHibernateProxyBuilder.ImplementGetLazyInitializer` explicit implementation, I do not actually know if it is needed.

src/NHibernate/Proxy/ProxyBuilderHelper.cs
```diff
@@ -134,25 +136,48 @@ internal static MethodBuilder GetObjectDataMethodBuilder(TypeBuilder typeBuilder
...
-			//TODO: Should we use attributes of base method?
-			var methodAttributes = MethodAttributes.Public | MethodAttributes.HideBySig | MethodAttributes.Virtual;
+			var explicitImplementation = method.DeclaringType.IsInterface;
+			if (explicitImplementation &&
+				(typeBuilder.BaseType == typeof(object) ||
+#pragma warning disable 618
+					typeBuilder.BaseType == typeof(ProxyDummy)) &&
+#pragma warning restore 618
+				(method.Name == "Equals" && method.GetParameters().Select(p => p.ParameterType).SequenceEqual(_equalsParametersTypes) ||
+					method.Name == "GetHashCode" && method.GetParameters().Length == 0))
```
fredericDelaporte: CodeFactor wants additional redundant parens here while ReSharper does not want them. I am on ReSharper side on this.

hazzik: Please extract 2 methods “IsEquals” and “IsGetHashCode” this way would everyone (CodeFactor, ReSharper and me) would be happy.

fredericDelaporte: We could then use them for the other place they are detected and specially handled.

src/NHibernate/Proxy/ProxyBuilderHelper.cs
```diff
@@ -253,5 +287,72 @@ private static string[] GenerateTypeNames(int count)
 
 			return result;
 		}
+
+		/// <summary>
+		/// Method equality for the proxy building purpose: we want to equate an interface method to a base type
+		/// method which implements it. This implies the base type method has the same signature and there is no
+		/// explicit implementation of the interface method in the base type.
+		/// </summary>
+		private class MethodInfoComparer : IEqualityComparer<MethodInfo>
```
hazzik: This comparer is just an optimization to generate only one method if the interface method is implemented implicitly. Right?

fredericDelaporte: Initially I wanted to handle it fully in the method using that comparer, but it was too complex.

Yes it is for generating only one method if the interface is implemented implicitly. No it is not an optimization from my viewpoint.

Not doing so, while implementing the base method and the interface method both publicly, is #2085 bug. Granted, the other changes in the method generation is enough to fix it alone.

But then, not doing so would trigger again #2052. This bug was re-qualified as just a "possible breaking change" bound to the switch to the static proxy, but if we can avoid that breaking change, that would be better.

fredericDelaporte:
Your change seems to me a good change in principle, but I have not yet thoroughly reviewed it, just to let you know.

hazzik:
> Your change seems to me a good change in principle, but I have not yet thoroughly reviewed it, just to let you know.

I've just moved the logic out of the comparer.

hazzik:
Would it be fair to say that the `GetProxiableMethods` method is doing following: "getting all `proxiable` methods from a base class plus all interface methods which are not implicitly implemented by the base class"?

fredericDelaporte:
Yes, that is it.

fredericDelaporte:
> I've just moved the logic out of the comparer.

Indeed, I agree with the change.

src/NHibernate/Proxy/ProxyBuilderHelper.cs
```diff
@@ -101,11 +102,19 @@ internal static IEnumerable<MethodInfo> GetProxiableMethods(System.Type type)
 
 		internal static IEnumerable<MethodInfo> GetProxiableMethods(System.Type type, IEnumerable<System.Type> interfaces)
 		{
-			var proxiableMethods =
-				GetProxiableMethods(type)
+			if (type.IsInterface || type == typeof(object) || type.GetInterfaces().Length == 0)
```
bahusoid: Are you sure it's valid shortcut?  I think `type.IsInterface`  part should be removed. In theory interface can contain multiple interfaces with the same members:
```C#
interface IReadOnlyId
{
	int Id { get; }
}

interface IId
{
	int Id { get; set; }
}

//IFoo contains both IReadOnlyId.Id and IId.Id getters. It's ambiguous but valid code
interface IFoo : IReadOnlyId, IId
{
}

public class Foo : IFoo
{
}
```
So for `IFoo` proxy (not sure about `Foo` proxy) duplicated methods are still generated.
Also I wonder when  interface `IFoo` is used as proxy will both `IId.Id` and `IReadOnlyId.Id` access for entity  work without proxy initialization.

hazzik: The type there is "Foo" (a class), so this check does not apply.

hazzik: If you use "IFoo" as a base for proxy. It would generate duplicate methods because all interface method are generated explicitly unless they are implemented implicitly in the base class.

In short - add tests to see the behaviour.

bahusoid: >The type there is "Foo

My concern is mainly for interface proxy - `IFoo`. I thought it was clear .

> unless they are implemented implicitly in the base class.

So it seems for interface proxy we should generate all members implicitly. That will solve possible duplicates and unnecessary proxy initialization.

bahusoid: Added test case in #2066. Current behavior leads to unnecessary proxy initialization. But as it's different interface members it's not really duplicate methods. So in the end seems unrelated to this PR.

bahusoid:
This PR has changed behavior for `Id` access via explicit interface (see #2066)
In 5.2: `Id` access via both implicit/explicit interfaces are not leading to proxy initialization.
After this PR: `Id` access via explicit interface leads to proxy initialization.

Shouldn't it be labeled as possible breaking change?

fredericDelaporte:
This is a bit special : this possible breaking change should have already occurred with 5.1.0 and the introduction of the static proxy factory. But some unrelated troubles and fixes have led to this behavior being still supporting at least at some point in 5.3. (According to this [comment](https://github.com/nhibernate/nhibernate-core/issues/2052#issuecomment-478344983), 5.3 was again supporting this debatable case since #2067.)
I do not know what is actually the status of this behavior in 5.1.x and 5.2x.

I consider this behavior of explicitly implemented `Id` not leading to proxy initialization be yielding the id value very debatable, because for correctness, an `Id` which is not implemented by the declared id in mapping cannot be assumed as being the entity id. It can be something else, and unless de-compiling and analyzing the concrete class, NHibernate cannot tell.

bahusoid:
> But some unrelated troubles and fixes have led to this behavior being still supported at least at some point in 5.3. 

No. Those changes are not related to this behavior (at least for class proxy). We made fixes for interface proxy. Handling of class proxy is not affected by those fixes at least regarding identifier property handling. I've just run #2066 tests for 5.2 branch - `ProxyClassIdAccessExplicitInterface` is failing. No proxy initialization for implicit interface.

bahusoid:
Just to be clear I'm not advocating for explicit interface support without proxy initialization. I'm just telling how it works in 5.2. And I've added #2066 tests specifically to avoid possible confusion in future.

fredericDelaporte:
It is no added on #2085 description (it was indeed already mentioned in it, but not formally):

> Possible breaking change: a class proxy for a class having an explicitly implemented interface declaring a member with the same name than the class id will now trigger a lazy load if the interface "id" is accessed.

If you have a simpler wording, feel free to edit.

bahusoid:
I would express it that way: Identifier property access via explicitly implemented interface now triggers proxy initialization.
Seems simpler to me (but maybe just for me :))

fredericDelaporte:
The thing is, it may not be the identifier. So it looks misleading to me.

bahusoid:
IMHO you have one identifier property in entity - the one which is mapped as identifier. For other names it's just internal implementation detail. But enough offtopic.

PR #2274, opened 2019-11-16
WIP - Add support for IAsyncEnumerable

### LINKED ISSUE #2267: Add support for IAsyncEnumerable
C# 8.0 and .NET Standard 2.1 brought us [asynchronous streams](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#asynchronous-streams) (horrible name, I'll call them "async enumerables" from now on). It would be nice if NHibernate could add new methods that return an [`IAsyncEnumerable<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.iasyncenumerable-1?view=netstandard-2.1). This would allow us to iterate over the results with `await foreach`.

I know this won't be a simple task and will require a few preparation steps, like adding .NET Standard 2.1 (and maybe .NET Core 3.0) targets to the build process. (Maybe this would even be the first time that NHibernate has a different API across target frameworks?) But I thought that this issue could at least be a tracking issue for the overall progress.

maca88:
Limitations: The added Linq extension method `AsAsyncEnumerable` do not support `PostExecuteTransformer` as the transformer operates with `IEnumerable<>`.

Fixes: #2267

Added WIP as the async generator has some issues generating obsolete overloads and tests with `IAsyncEnumerable`.


src/NHibernate/IQuery.cs
```diff
@@ -664,4 +696,27 @@ public partial interface IQuery
...
+		/// <typeparam name="T">The element type.</typeparam>
+		/// <param name="query">The query.</param>
+		public static IAsyncEnumerable<T> AsyncEnumerable<T>(this IQuery query)
+		{
+			if (query is AbstractQueryImpl abstractQuery)
+			{
+				return abstractQuery.AsyncEnumerable<T>();
+			}
+
+			throw new NotImplementedException();
```
hazzik: Should be `NotSupportedException`. Also there is a helper method to do upcast.

src/NHibernate/Hql/IQueryTranslator.cs
```diff
@@ -119,4 +127,25 @@ public partial interface IQueryTranslator
...
+		/// <param name="queryParameters">The query parameters.</param>
+		/// <param name="session">The session.</param>
+		public static IAsyncEnumerable<T> GetAsyncEnumerable<T>(this IQueryTranslator queryTranslator, QueryParameters queryParameters, IEventSource session)
+		{
+			if (queryTranslator is QueryTranslatorImpl queryTranslatorImpl)
+			{
+				return queryTranslatorImpl.GetAsyncEnumerable<T>(queryParameters, session);
+			}
+
+			throw new NotImplementedException();
```
hazzik: Same as `IQuery`

src/NHibernate/Engine/Query/HQLQueryPlan.cs
```diff
@@ -20,7 +21,34 @@ public partial interface IQueryPlan
...
+		/// <param name="queryParameters">The query parameters.</param>
+		/// <param name="session">The session.</param>
+		public static IAsyncEnumerable<T> PerformAsyncIterate<T>(this IQueryPlan queryPlan, QueryParameters queryParameters, IEventSource session)
+		{
+			if (queryPlan is HQLQueryPlan hQLQueryPlan)
+			{
+				return hQLQueryPlan.PerformAsyncIterate<T>(queryParameters, session);
+			}
+
+			throw new NotImplementedException();
```
hazzik: Same as IQuery

src/NHibernate/Engine/ISessionImplementor.cs
```diff
@@ -62,6 +64,41 @@ internal static void AutoFlushIfRequired(this ISessionImplementor implementor, I
...
+		/// <param name="filter">The filter to apply.</param>
+		/// <param name="parameters">The query parameters.</param>
+		public static IAsyncEnumerable<T> AsyncEnumerableFilter<T>(this ISessionImplementor session, object collection, string filter, QueryParameters parameters)
+		{
+			if (session is AbstractSessionImpl abstractSession)
+			{
+				return abstractSession.AsyncEnumerableFilter<T>(collection, filter, parameters);
+			}
+
+			throw new NotImplementedException();
```
hazzik: Same

src/NHibernate/Loader/Hql/QueryLoader.cs
```diff
@@ -495,6 +497,44 @@ internal IEnumerable GetEnumerable(QueryParameters queryParameters, IEventSource
...
+				InitializeAsync,
+				queryParameters.IsReadOnly(session),
+				_queryTranslator.ReturnTypes,
+				_queryTranslator.GetColumnNames(),
+				queryParameters.RowSelection,
+				_selectNewTransformer ?? queryParameters.ResultTransformer,
+				_queryReturnAliases,
+				session);
+
+			async Task<InitializeResult> InitializeAsync(CancellationToken cancellationToken)
```
hazzik: With this the async enumerable becomes a "future"-like version. Which is not how the Enumerable behaves. I think we still need to execute query upfront, not on `MoveNextAsync()`

src/NHibernate/Loader/Hql/QueryLoader.cs
```diff
@@ -495,6 +497,44 @@ internal IEnumerable GetEnumerable(QueryParameters queryParameters, IEventSource
...
+				}
+
+				var command = await PrepareQueryCommandAsync(queryParameters, false, session, cancellationToken).ConfigureAwait(false);
+				var dataReader = await GetResultSetAsync(command, queryParameters, session, null, cancellationToken).ConfigureAwait(false);
+
+				if (stopWatch != null)
+				{
+					stopWatch.Stop();
+					session.Factory.StatisticsImplementor.QueryExecuted("HQL: " + _queryTranslator.QueryString, 0, stopWatch.Elapsed);
+					// NH: Different behavior (H3.2 use QueryLoader in AST parser) we need statistic for orginal query too.
```
hazzik: Don't need this comment here.

src/NHibernate/Util/JoinedAsyncEnumerable.cs
```diff
@@ -0,0 +1,77 @@
...
+
+		public JoinedAsyncEnumerable(IAsyncEnumerable<T>[] asyncEnumerables)
+		{
+			_asyncEnumerables = asyncEnumerables;
+		}
+
+		public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+		{
+			return new JoinedAsyncEnumerator(
+				_asyncEnumerables.Select(o => o.GetAsyncEnumerator()).ToArray(),
```
hazzik: There is .ToArray extension with a selector.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -594,6 +594,21 @@ public override IEnumerable<T> Enumerable<T>(IQueryExpression queryExpression, Q
...
+		{
+			using (BeginProcess())
+			{
+				queryParameters.ValidateParameters();
+				var plan = GetHQLQueryPlan(queryExpression, true);
+				AutoFlushIfRequired(plan.QuerySpaces);
+
+				using (SuspendAutoFlush()) //stops flush being called multiple times if this method is recursively called
+				{
+					return plan.PerformAsyncIterate<T>(queryParameters, this);
```
hazzik: Not sure how this will behave....

src/NHibernate/Impl/AsyncEnumerableImpl.cs
```diff
@@ -0,0 +1,258 @@
...
+				}
+				finally
+				{
+					_session.DefaultReadOnly = sessionDefaultReadOnlyOrig;
+				}
+
+				// In order to simulate what SafetyEnumerable does, we have to skip elements of different type
+				if (!isCurrentValid)
+				{
+					await MoveNextAsync().ConfigureAwait(false);
```
hazzik: I get it that you love recursions, but can we please remove it?

hazzik:
Looks good so far.

src/NHibernate/IQuery.cs
```diff
@@ -119,6 +123,35 @@ public partial interface IQuery
...
+
+		/// <summary>
+		/// Strongly-typed version of <see cref="Enumerable()"/>.
+		/// </summary>
+		/// <param name="cancellationToken">A cancellation token that can be used to cancel the work</param>
+		/// <typeparam name="T"></typeparam>
+		/// <returns></returns>
+		// Since v5.3
+		[Obsolete("This method has no more usages and will be removed in a future version")]
+		Task<IEnumerable<T>> EnumerableAsync<T>(CancellationToken cancellationToken = default(CancellationToken));
```
bahusoid: Shouldn't we suggest what should be used instead?

src/NHibernate/Engine/Query/HQLQueryPlan.cs
```diff
@@ -174,6 +198,69 @@ public IEnumerable PerformIterate(QueryParameters queryParameters, IEventSource
...
+				return new CollectionHelper.EmptyAsyncEnumerable<T>();
+			}
+
+			if (Translators.Length == 1)
+			{
+				return Translators[0].GetAsyncEnumerable<T>(queryParameters, session);
+			}
+
+			var results = new IAsyncEnumerable<T>[Translators.Length];
+			for (int i = 0; i < Translators.Length; i++)
```
bahusoid: Non-async version in case multiple translators opens all DB readers before processing. In case multiple open readers are not supported (like MS SQL Server with MARS disabled by default) - all data populated at once. It makes `PerformIterate` quite useless for such queries (see #2017).

Just want to clarify async behavior. So how async version behaves in this regard? Does it also opens all readers at once?

hazzik: Currently as it is implemented it opens the connection and starts reading only when you call MoveNextAsync first time.

src/NHibernate/Impl/AsyncEnumerableImpl.cs
```diff
@@ -0,0 +1,258 @@
...
+				try
+				{
+					isCurrentValid = await PostMoveNextAsync().ConfigureAwait(false);
+				}
+				finally
+				{
+					_session.DefaultReadOnly = sessionDefaultReadOnlyOrig;
+				}
+
+				// In order to simulate what SafetyEnumerable does, we have to skip elements of different type
```
bahusoid: I actually quite dislike this safe enumerable part. It mostly means some user code error which is silently ignored. May be we shouldn't do it in async part?

hazzik: Same

bahusoid:
> The added Linq extension method AsAsyncEnumerable

I'm pretty sure that `Enumerable.AsEnumerable` LINQ extension leads to `IQuery.List`. That makes sync counterpart quite different from async version.

We either need to make `AsEnumerable` to use `PerformIterate` or maybe use different methods. Like `ToEnumerable/ToAsyncEnumerable`.

src/NHibernate/Loader/Hql/QueryLoader.cs
maca88: >  I think we still need to execute query upfront, not on MoveNextAsync()

Initially I tried to copy `QueryLoader.GetEnumerable` behavior, but the problem is that in order to do that we would need to call `PrepareQueryCommand` and `GetResultSet` synchronously, otherwise we would have a similar issue as with the current `EnumerableAsync`, where the enumerable is wrapped in a task. As I am not sure what is the "correct" behavior, I've checked what Microsoft did with EF and they seem to execute the query on the first `MoveNextAsync`, calling `AsAsyncEnumerable` does not trigger any query (checked with SQL Serever Profiler). My current opinion on this is that `EnumerableImpl` is kind of broken as it does not support multiple iterations, which is also stated in the documentation:
https://github.com/nhibernate/nhibernate-core/blob/e8288a58c50a2ecf7ee1e47ee0e0f55c9a1213e5/src/NHibernate/Impl/EnumerableImpl.cs#L20-L23
and by executing the query upon calling `AsAsyncEnumerable`, means that we would only be able to iterate once.

maca88:
> I'm pretty sure that Enumerable.AsEnumerable LINQ extension leads to IQuery.List on iteration

I test it and calling `AsEnumerable` does not execute any query, which means that it is possible to iterate over the same enumerable multiple times. I've mistakenly added such test for it instead of `IAsyncEnumerable`:
https://github.com/nhibernate/nhibernate-core/pull/2274/files#diff-64ae2936b1d2c700144c9d334ef6d9a6R12

bahusoid:
> I tested it and calling AsEnumerable does not execute any query,

Yeah that's expected (`As` suggests that it only casts inside )  I meant on actual enumeration `IQuery.List` will be called.  `AsAsyncEnumerable` name suggests that it behaves as `AsEnumerable` but in async way. But it's not the case. It's completely different behavior.

It seems we currently don't expose LINQ method executing `IQuery.Enumerable`. But if we add one I don't think it should be executed on standard `AsEnumerable` extension  - it should be some separate method (maybe `ToEnumerable`? and then async counterpart should be `ToAsyncEnumerable`).

src/NHibernate/Loader/Hql/QueryLoader.cs
bahusoid: > My current opinion on this is that EnumerableImpl is kind of broken as it does not support multiple iterations, which is also stated in the documentation:

Agree. That's something I've been trying to fix in #2017.

maca88:
> AsAsyncEnumerable name suggests that it behaves as AsEnumerable but in async way. But it's not the case. It's completely different behavior.

In the current state of this PR they are not completely different, the only difference is that `AsEnumerable` will read all rows from the data reader when the first `MoveNext` is called, where `AsAsyncEnumerable` will read just one. Basically `AsEnumerable` works similar as a future query which executes the query and read all rows when first accessed, which kills the meaning of `AsEnumerable`. The issue lies in `DefaultQueryProvider.ExecuteQuery` method where `IQuery.List` is called instead of `IQuery.Enumerable`:
https://github.com/nhibernate/nhibernate-core/blob/d8931ac97e97f56d811ea8b5383c08e39a96df40/src/NHibernate/Linq/DefaultQueryProvider.cs#L209
Using `IQuery.List` means that the rows will be traversed twice even when `Enumerable.ToList()` is called, first inside `IQuery.List` and the second time inside `Enumerable.ToList()`, which basically passes the result into the `List` constructor:
https://github.com/microsoft/referencesource/blob/17b97365645da62cf8a49444d979f94a59bbb155/System.Core/System/Linq/Enumerable.cs#L947-L950
> It seems we currently don't expose LINQ method executing IQuery.Enumerable

I think that we should instead change the current method (`DefaultQueryProvider.ExecuteQuery`) to use it instead of `IQuery.List`. By using `IQuery.Enumerable`, `AsEnumerable` would still not work as expected because `IQuery.Enumerable` does not support multiple iterations, so we would also need to change than. Do really `IQuery.Enumerable` needs to execute the query when called, is there a reason behind it?

bahusoid:
> Do really IQuery.Enumerable needs to execute the query when called, is there a reason behind it?

Did you mean `IQuery.List`? Well as far as I know `IQuery.Enumerable` is not implemented for stateless session (don't know reasons)

maca88:
> Did you mean IQuery.List?

No, I meant the following code that is executed when `IQuery.Enumerable` is called:
https://github.com/nhibernate/nhibernate-core/blob/d8931ac97e97f56d811ea8b5383c08e39a96df40/src/NHibernate/Loader/Hql/QueryLoader.cs#L478-L481
which execute the query, preventing it from iterating the enumerable multiple times:
```
var enumerable = s.CreateQuery("from Customer").Enumerable<Customer>();
enumerable.ToList();
enumerable.ToList();
```

> Well as far as I know ISession.Enumerable is not implemented for stateless session (don't know why)

Good point, that was probably the reason why `List` was used as it supports more features than `Enumerable` method. It was probably not implemented because `IQueryPlan.PerformIterate` has an `IEventSource` parameter and stateless session does not implement it. The `IEventSource` is used by `EnumerableImpl` which uses `ISession.DefaultReadOnly` (`IEventSource` extends `ISession`) property that is not available with a stateless session. In order to support it for the stateless session we would need to change the parameter type to `ISessionImplementor` and then inside `EnumerableImpl` use the `DefaultReadOnly` only when the given session is castable to `ISession`.

src/NHibernate/Loader/Hql/QueryLoader.cs
maca88: Unfurtunately #2017 does not fix multiple iterations as in order to support that the following code should be executed inside `EnumerableImpl` when first `MoveNext` is called, similar what was done for `AsyncEnumerableImpl` in this PR:
https://github.com/nhibernate/nhibernate-core/blob/d8931ac97e97f56d811ea8b5383c08e39a96df40/src/NHibernate/Loader/Hql/QueryLoader.cs#L478-L481

bahusoid: > does not fix multiple iterations ... to support that the following code should be executed inside EnumerableImpl when first MoveNext is called,

Actually all I need to fix it is to comment these [2 lines](https://github.com/bahusoid/nhibernate-core/blob/0a796d5e85ce5a5d15bcf1119d6235c9782abbbd/src/NHibernate/Engine/Query/HQLQueryPlan.cs#L163-L164) in my PR. And case with multiple translators already works properly there.

bahusoid: It got me thinking. Wouldn't it be less changes to implement async part similarly.
I mean keep `Translators.GetAsyncEnumerable` do query execution. And delay and make it reusable  using iterator
 similar to this [`GetEnumerable`](https://github.com/bahusoid/nhibernate-core/blob/0a796d5e85ce5a5d15bcf1119d6235c9782abbbd/src/NHibernate/Engine/Query/HQLQueryPlan.cs#L169-L178). Something like:
```C#
private IEnumerable<IAsyncEnumerable<object>> GetAsyncEnumerables(QueryParameters queryParameters, IEventSource session)
{
	foreach (var t in Translators)
	{
		foreach (object obj in t.GetAsyncEnumerable(queryParameters, session))
		{
			yield return obj;
		}
	}
}
```

bahusoid:
> The IEventSource is used by EnumerableImpl which uses ISession.DefaultReadOnly

`QueryParameters.IsReadOnly()` should be enough. That's what `List` uses

src/NHibernate/Loader/Hql/QueryLoader.cs
maca88: > Wouldn't it be less changes to implement async part similarly

This is what I was trying to achieve initially, but the problem is that without C#8 we are not able to use `await foreach`, which means we cannot make the following method:
```C#
private async IAsyncEnumerable<T> GetAsyncEnumerables<T>(QueryParameters queryParameters, IEventSource session)
{
  foreach (var t in Translators)
  {
    await foreach (var element in t.GetAsyncEnumerable<T>(queryParameters, session))
    {
      yield return element;
    }
  }
}
```
Your code would not compile because `IAsyncEnumerable` does not implement `IEnumerable` which is needed by `foreach` statement.

bahusoid: That's why I suggest returning `IEnumerable<IAsyncEnumerable<T>` and not `IAsyncEnumerable<T>`. And "flattening" part should be done inside `JoinedAsyncEnumerable`

src/NHibernate/IQuery.cs
bahusoid: How about adding support for conversion rule `Obsolete` in AsyncGenerator.yml to avoid adding manually those methods.
Something like: 
```yml
#Since v5.3
    - conversion: Obsolete
      name: EnumerableAsync
      description: "Use AsyncEnumerable instead"
```

src/NHibernate/Engine/Query/HQLQueryPlan.cs
```diff
@@ -174,6 +198,69 @@ public IEnumerable PerformIterate(QueryParameters queryParameters, IEventSource
...
+			{
+				var result = await (Translators[i].GetEnumerableAsync(queryParameters, session, cancellationToken)).ConfigureAwait(false);
+				results[i] = result;
+			}
+			return new JoinedEnumerable(results);
+		}
+
+		// Since v5.3
+		[Obsolete("This method has no more usages and will be removed in a future version")]
```
bahusoid: It seems enough to have async methods in interface marked as obsolete. Generator seems smart enough to find obsolete async methods in interface and keep them generated. That would make less change noise.

src/NHibernate/Engine/Query/HQLQueryPlan.cs
```diff
@@ -174,6 +198,69 @@ public IEnumerable PerformIterate(QueryParameters queryParameters, IEventSource
...
+		{
+			if (Log.IsDebugEnabled())
+			{
+				Log.Debug("enumerable: {0}", _sourceQuery);
+				queryParameters.LogParameters(session.Factory);
+			}
+
+			if (Translators.Length == 0)
+			{
+				return new CollectionHelper.EmptyAsyncEnumerable<T>();
```
bahusoid: I don't think we need this optimization along with `CollectionHelper.EmptyAsyncEnumerable` (at least not in this PR). This case means query is executed  on not mapped entity. If it's not throwing already before - I see no point in optimizing broken case.

src/NHibernate/Loader/Hql/QueryLoader.cs
maca88: > That's why I suggest returning IEnumerable<IAsyncEnumerable<T> and not IAsyncEnumerable<T>.

Now I understod what you meant and indeed it simplified `HQLQueryPlan.PerformAsyncIterate`.

> Actually all I need to fix it is to comment these 2 lines in my PR

That correct, but calling `IQueryTranslator.GetEnumerable` will still trigger the query to be executed, where `IQueryTranslator.GetAsyncEnumerable` in this PR does not. In order to make those two method consistent I deprecated `EnumerableImpl`, added `IEnumerable<>` to `AsyncEnumerableImpl` and used it instead.

src/NHibernate/Loader/Hql/QueryLoader.cs
maca88: Now this method is shared for `IEnumerable` and `IAsyncEnumerable` and the comment is removed, but I am not sure whether it is still relevant.

src/NHibernate/Impl/AsyncEnumerableImpl.cs
maca88: I will try to avoid them. This code is not needed anymore as `SafetyEnumerable` behavior is not simulated anymore.

src/NHibernate/IQuery.cs
maca88: > Shouldn't we suggest what should be used instead?

Description updated.

> How about adding support for conversion rule Obsolete

Is seems a good idea, I will try to implement it in the near future.

src/NHibernate/Impl/AsyncEnumerableImpl.cs
maca88: Now an exception will be thrown instead.

src/NHibernate/Engine/Query/HQLQueryPlan.cs
maca88: I overlooked that, now it is generated instead.

src/NHibernate/Engine/Query/HQLQueryPlan.cs
maca88: `CollectionHelper.EmptyAsyncEnumerable` is now removed thanks to your proposal for yielding `IAsyncEnumerable`.

src/NHibernate/Impl/SessionImpl.cs
bahusoid: What are you talking about? Is it about  `using (SuspendAutoFlush())`? Yeah this code seems suspicious  - if this suspend is really required  ( as this code looks broken to me even for sync case - it's surely not scoped for items iterations) it needs to be done inside iterator.

src/NHibernate/Impl/AsyncEnumerableImpl.cs
```diff
@@ -0,0 +1,327 @@
...
+
+			private void PostRead()
+			{
+				if (_selection != null && _selection.MaxRows != RowSelection.NoValue)
+				{
+					_hasNext = _hasNext && (_currentRow < _selection.MaxRows);
+				}
+
+				bool sessionDefaultReadOnlyOrig = _session.DefaultReadOnly;
+				_session.DefaultReadOnly = _readOnly;
```
bahusoid: Does it really make sense to do it for each processed item? Maybe it should be done once for iterations (so this logic should be moved to `GetAsyncEnumerables` or even to `JoinedAsyncEnumerable`

src/NHibernate/Loader/Hql/QueryLoader.cs
bahusoid: It doesn't add any big overheads? Then OK. So please confirm it makes #2017 obsolete

maca88: > It doesn't add any big overheads for sync case?

Not really, the reading logic remained the same.

> So please confirm it makes #2017 obsolete

It does, I obsoleted `JoinedEnumerable` in this PR and adopted `PerformIterate` from your PR, but I kept the `if (Translators.Length == 1)` condition in order to avoid nested enumeration.

src/NHibernate/Impl/AsyncEnumerableImpl.cs
```diff
@@ -0,0 +1,327 @@
...
+
+			private readonly InitializeEnumerableAsync _initializeAsync;
+			private readonly InitializeEnumerable _initialize;
+			private readonly bool _readOnly;
+			private readonly IType[] _types;
+			private readonly string[][] _columnNames;
+			private readonly RowSelection _selection;
+			private readonly IResultTransformer _resultTransformer;
+			private readonly string[] _returnAliases;
+			private readonly IEventSource _session;
```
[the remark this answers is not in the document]
maca88: Done

src/NHibernate/NHibernateUtil.cs
```diff
@@ -532,18 +532,20 @@ public static Clob CreateClob(string str)
 		/// </summary>
 		public static void Close(IEnumerator enumerator)
 		{
-			EnumerableImpl hibernateEnumerator = enumerator as EnumerableImpl;
-			if (hibernateEnumerator == null)
+			if (enumerator?.GetType().GetGenericTypeDefinition() != typeof(AsyncEnumerableImpl<>.AsyncEnumeratorImpl))
```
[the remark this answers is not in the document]
maca88: You are right, changed to your proposal.

src/NHibernate/Linq/DefaultQueryProvider.cs
```diff
@@ -299,6 +315,17 @@ public int ExecuteDml<T>(QueryMode queryMode, Expression expression)
 			return query.ExecuteUpdate();
 		}
 
+		public IAsyncEnumerable<T> GetAsyncEnumerable<T>(Expression expression)
+		{
+			var nhLinqExpression = PrepareQuery(expression, out var query);
+			if (nhLinqExpression.ExpressionToHqlTranslationResults?.PostExecuteTransformer != null)
+			{
+				throw new NotSupportedException("AsyncEnumerable is not supported when PostExecuteTransformer is set.");
```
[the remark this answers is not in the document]
maca88: > Is this transformation for whole list?

Yes, the `PostExecuteTransformer` takes the whole list as an argument and it can be used for aggregate methods like `Enumerable.Count` and also for `Enumerable.All` to support polymorphic queries:
https://github.com/nhibernate/nhibernate-core/blob/ac39173567b31bcfad475ca32687c9faf0d37f87/src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessAll.cs#L25-L27

As `AsAsyncEnumerable` does not have an overload like `ToFutureValue` does, where you can do something like `q.ToFutureValue(o => o.Count())`, this code will never be reached when using the current `AsAsyncEnumerable` extension method. We could add an overload for it and cast the result to `IEnumerable<T>` as `AsyncEnumerableImpl` now implements it, but it kills the purpose of `IAsyncEnumerable` as the iteration will be done synchronously using `IEnumerable` instead.

src/NHibernate/Impl/AsyncEnumerableImpl.cs
maca88: > Does it really make sense to do it for each processed item?

The user may change the `DefaultReadOnly` value while iterating, which may cause a readonly query to load entites as writable, if `DefaultReadOnly` is not set for each item.

src/NHibernate/Impl/SessionImpl.cs
hazzik: Yes, I'm talking about that.

hazzik:
So, with the latest iteration of changes this has become a close resemblance of "Future" feature, right?

maca88:
The last iteration of changes applied the same logic in terms of when the query is executed to the current `Enumerable` methods, which is similar how "Future" feature work, right.

maca88:
Rebased and re-generated with AsyncGenerator `0.18.1`

src/NHibernate/Impl/SessionImpl.cs
maca88: As now `PerformAsyncIterate` and `PerformIterate` do not execute any query, the auto flush logic was moved into the initialization of the enumerator, so that the auto flush logic will be triggered before the query is executed.

maca88:
Only now I realized that `PerformIterate` and now `PerformAsyncIterate` trigger N+1 queries. For example the following query:
```
s.CreateQuery("from Simple").Enumerable<Simple>()
``` 
will trigger one query that will retrieve all ids and then on each `MoveNext` call a separate query will be triggered for the current id, if the entity is not in the cache ([hibenate does](https://stackoverflow.com/a/15589125) the same). Hibernate has also [scroll method](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java#L1533) that does what I thought `Enumerable` does, which executes only one query. For me, the behavior of `scroll` method is more suitable for `Enumerable` and `AsyncEnumerable` but it would require to port the feature from hibernate, which is a topic for a separate PR.

maca88:
Rebased.

bahusoid:
> Only now I realized that PerformIterate and now PerformAsyncIterate trigger N+1 queries. 

That sucks :( It's not very useful for entities but I guess it's still can be useful for scalar values...

> For me, the behavior of scroll method is more suitable for Enumerable and AsyncEnumerable but it would require to port the feature from hibernate, which is a topic for a separate PR.

Yeah. And I would just replace Enumerable implementation with scroll.  Current enumerable entity handling is not intuitive at all.

maca88:
> Yeah. And I would just replace Enumerable implementation with scroll

One issue that I didn't checked, is how `scroll` behave when the query has collection fetches. Hibernate uses [FetchingScrollableResultsImpl](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/internal/FetchingScrollableResultsImpl.java), but I didn't tested it yet to understand how it works.

fredericDelaporte:
> So, with the latest iteration of changes this has become a close resemblance of "Future" feature, right?

> The last iteration of changes applied the same logic in terms of when the query is executed to the current `Enumerable` methods, which is similar how "Future" feature work, right.

The main purpose of the Future feature is not deferred execution but batching together queries. (Future also features a very loose coupling between code parts defining the queries, contrary to the recent query batcher which requires referring to the same batcher reference for adding queries into it, but as such also helps knowing exactly what is batched together.)
Deferred execution is just a side effect of how Future works.

This `IAsyncEnumerable` feature does nothing to batch queries together, so it is still two quite different features.

Anyway, it appears [this "enumerable" feature is not very attractive with its current performances issues](https://github.com/nhibernate/nhibernate-core/pull/2274#issuecomment-593123787).

Is it the reason for this PR to be stalled? Or should its conflicts be resolved for allowing merging it?

maca88:
> Is it the reason for this PR to be stalled? 

Yes, the `scroll` feature from Hibernate should be ported first, so that it can be used in this PR to avoid N+1 issue. I haven't started working on it yet as my current priority is the `Include` method. Changed to WIP until the `scroll` feature is ported.

zlsmith86:
Is there an issue to track the porting of the `scroll` feature?

Harpush:
Any progress here with the scroll feature?

Issue #2177, opened 2019-06-13
New Fetch() method in QueryOver returns IQueryOver<> instead of QueryOver<>


jrauber:
In 

> Full control of entities fetching in Criteria (#1599)

 the new introduced method
 
`Fetch(SelectMode mode, Expression<Func<TSubType, object>> path)`

in `Criterion/QueryOver.cs` returns the `IQueryOvery<>` interface instead of the previously returned `QueryOver<>`. As a result, queries that are not already linked to the session cannot be transformed by calling `query.GetExecutableQueryOver(session)` because the interface does not provide the `GetExecutableQueryOver(session)` method.

bahusoid:
Actually there is already extension method returning `QueryOver`. But it's hidden by interface implementation. So
as a workaround you can tell compiler to use extension by specifying param name explicitly:
```C#
queryOver.Fetch(SelectMode.Fetch, associationPaths: x => x.Child1).GetExecutableQueryOver()
```

PR #2275, opened 2019-11-24
Add missing options to QueryOver

### LINKED ISSUE #2270: IQueryOver<T> is lacking some options
Compared to criteria, QueryOver is lacking methods for setting the timeout, comments and fetch size. They are available through the underlying criteria, but a direct mean for setting them would be better.

Original Title: Documentation does not detail how to set timeout on IQueryOver<T>
Original description: Title should be self explanatory. If [this](https://gist.github.com/wullemsb/4966132) is still relevant, please include it in the documentation of "**QueryOver Queries**".

fredericDelaporte:
Fix #2270

Fix it by adding the missing options as extension methods, rather than documenting getting them through the underlying criteria. If that is accepted, then #2270 should be rewritten a bit.

Adds `SetTimeout`, `SetFetchSize`, `SetComment`.

It seems the `IQueryOver` interface tends to drop the `Set` prefix. Should theses extension methods do the same? It does not look to me as a proper convention, since methods are supposed to start with a verb. But if it is preferred to keep dropping the verb for the interface consistency, I will do.

### DIFF
--- a/src/NHibernate/QueryOverExtensions.cs
+++ b/src/NHibernate/QueryOverExtensions.cs
@@ -1,0 +1,42 @@
+namespace NHibernate
+{
+	// 6.0 TODO: merge into IQueryOver<TRoot>
+	public static class QueryOverExtensions
+	{
+		/// <summary>
+		/// Set a timeout for the underlying ADO.NET query.
+		/// </summary>
+		/// <param name="queryOver">The query on which to set the timeout.</param>
+		/// <param name="timeout">The timeout in seconds.</param>
+		/// <returns><see langword="this" /> (for method chaining).</returns>
+		public static IQueryOver<TRoot> SetTimeout<TRoot>(this IQueryOver<TRoot> queryOver, int timeout)
+		{
+			queryOver.RootCriteria.SetTimeout(timeout);
+			return queryOver;
+		}
+		
+		/// <summary>
+		/// Set a fetch size for the underlying ADO query.
+		/// </summary>
+		/// <param name="queryOver">The query on which to set the timeout.</param>
+		/// <param name="fetchSize">The fetch size.</param>
+		/// <returns><see langword="this" /> (for method chaining).</returns>
+		public static IQueryOver<TRoot> SetFetchSize<TRoot>(this IQueryOver<TRoot> queryOver, int fetchSize)
+		{
+			queryOver.RootCriteria.SetFetchSize(fetchSize);
+			return queryOver;
+		}
+
+		/// <summary>
+		/// Add a comment to the generated SQL.
+		/// </summary>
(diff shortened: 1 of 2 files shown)


src/NHibernate/QueryOverExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+			return queryOver;
+		}
+
+		/// <summary>
+		/// Add a comment to the generated SQL.
+		/// </summary>
+		/// <param name="queryOver">The query on which to set the timeout.</param>
+		/// <param name="comment">A human-readable string.</param>
+		/// <returns><see langword="this" /> (for method chaining).</returns>
+		public static IQueryOver<TRoot> SetComment<TRoot>(this IQueryOver<TRoot> queryOver, string comment)
```
bahusoid: I would prefer it that way:
```C#
public static TQueryOver SetComment(this TQueryOver queryOver, string comment) where TQueryOver: IQueryOver
```

It avoids losing initial type (so also covers `DetachedCriteria`).

src/NHibernate/QueryOverExtensions.cs
```diff
@@ -0,0 +1,42 @@
+namespace NHibernate
+{
+	// 6.0 TODO: merge into IQueryOver<TRoot>
```
bahusoid: Why?

fredericDelaporte: For consistency with the few other cases already put in the interface which just delegate to the criteria/root criteria.
We could put instead the converse todo: moving those other options to extension methods.

bahusoid: Well interface thing won't work with generic `TQueryOver`.
If you want to stick to interface this change should be reverted and you need also to add public methods to `QueryOver` returning exact type -  to cover detached queryover (see issues like #2177, maybe you should add a test to make sure detached QueryOver actually covered).

But IMHO we should keep them as extensions  - much cleaner implementation.

fredericDelaporte: I mean, instead of this PR current todo, the option methods already in the interface could be flagged with the converse todo, like `6.0 TODO: move to extension method`.

bahusoid: Yeah I got it and I have no opinion on that... I'm just saying that current TODO is not applicable. Maybe just update this TODO to analyze and move suitable QueryOver interface methods to extensions.

fredericDelaporte: Todo changed.

bahusoid: > they are quite hard to override.

Yeah but QueryOver is hardwired to criteria and basically wrapper around it. So I don't see it as big drawback here...

Issue #2278, opened 2019-11-29
IInterceptor.OnPrepareStatement results not used in insert/update commands


shiznit013:
I have a project where the database names are dynamic. The tables in each database are the same. We use a custom interceptor where the ID of the database is replaced with a placeholder. This code worked until we switched to the .NET Core version of NHibernate.

**Exception**
```
could not execute batch command.[SQL: SQL not available]

at NHibernate.AdoNet.GenericBatchingBatcher.DoExecuteBatchAsync(DbCommand ps, CancellationToken cancellationToken)
at NHibernate.AdoNet.AbstractBatcher.ExecuteBatchWithTimingAsync(DbCommand ps, CancellationToken cancellationToken)
at NHibernate.AdoNet.AbstractBatcher.ExecuteBatchAsync(CancellationToken cancellationToken)
at NHibernate.Engine.ActionQueue.ExecuteActionsAsync(IList list, CancellationToken cancellationToken)
at NHibernate.Engine.ActionQueue.ExecuteActionsAsync(CancellationToken cancellationToken)
at NHibernate.Engine.ActionQueue.ExecuteActionsAsync(CancellationToken cancellationToken)
at NHibernate.Event.Default.AbstractFlushingEventListener.PerformExecutionsAsync(IEventSource session, CancellationToken cancellationToken)
at NHibernate.Event.Default.DefaultFlushEventListener.OnFlushAsync(FlushEvent event, CancellationToken cancellationToken)
at NHibernate.Impl.SessionImpl.FlushAsync(CancellationToken cancellationToken)
at NHibernate.Impl.SessionImpl.BeforeTransactionCompletionAsync(ITransaction tx, CancellationToken cancellationToken)
at NHibernate.Transaction.AdoTransaction.CommitAsync(CancellationToken cancellationToken)
at UserQuery.Main(), line 15
at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__139_1(Object state)
at System.Threading.QueueUserWorkItemCallback.<>c.<.cctor>b__6_0(QueueUserWorkItemCallback quwi)
at System.Threading.ExecutionContext.RunForThreadPoolUnsafe[TState](ExecutionContext executionContext, Action`1 callback, TState& state)
at System.Threading.QueueUserWorkItemCallback.Execute()
at System.Threading.ThreadPoolWorkQueue.Dispatch()
at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback()
```
[...]

PR #2285, opened 2019-12-15
Fix GenericBatchingBatcher to call IInterceptor.OnPrepareStatement

maca88:
Fixes: #2278

### DIFF
--- a/src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
+++ b/src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
@@ -120,11 +120,9 @@ namespace NHibernate.Test.Ado
 
 				tx.Commit();
 
-				// Called for each batch and once for generating AbstractBatcher.CurrentCommand
-				Assert.That(interceptor.TotalCalls, Is.EqualTo(5));
+				// Called for each insert statement and once for generating AbstractBatcher.CurrentCommand
+				Assert.That(interceptor.TotalCalls, Is.EqualTo(21));
 				var log = sqlLog.GetWholeLog();
-				// In reality only 4 comments are added (one for each batch), but because AbstractBatcher.CurrentCommand is logged for each added command
-				// 20 will be logged instead.
 				Assert.That(FindAllOccurrences(log, "/* TEST */"), Is.EqualTo(20));
 			}
 
@@ -143,9 +141,10 @@ namespace NHibernate.Test.Ado
 
 				tx.Commit();
 
-				Assert.That(interceptor.TotalCalls, Is.EqualTo(1));
+				var totalCalls = Sfi.ConnectionProvider.Driver.SupportsMultipleQueries ? 1 : 2;
+				Assert.That(interceptor.TotalCalls, Is.EqualTo(totalCalls));
 				var log = sqlLog.GetWholeLog();
-				Assert.That(FindAllOccurrences(log, "/* TEST */"), Is.EqualTo(1));
+				Assert.That(FindAllOccurrences(log, "/* TEST */"), Is.EqualTo(totalCalls));
 			}
 
 			Cleanup();
--- a/src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
+++ b/src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
@@ -120,8 +120,7 @@ namespace NHibernate.Test.Ado
 
 				tx.Commit();
 
-				// Called for each insert statement and once for generating AbstractBatcher.CurrentCommand
-				Assert.That(interceptor.TotalCalls, Is.EqualTo(21));
+				Assert.That(interceptor.TotalCalls, Is.EqualTo(1));
 				var log = sqlLog.GetWholeLog();
 				Assert.That(FindAllOccurrences(log, "/* TEST */"), Is.EqualTo(20));
 			}
--- a/src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
+++ b/src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
@@ -5,6 +5,8 @@ using System.Linq;
 using NHibernate.AdoNet;
 using NHibernate.Cfg;
 using NHibernate.Dialect;
+using NHibernate.Linq;
+using NHibernate.SqlCommand;
 using NUnit.Framework;
 using Environment = NHibernate.Cfg.Environment;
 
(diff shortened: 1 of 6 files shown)


src/NHibernate/AdoNet/GenericBatchingBatcher.cs
```diff
@@ -207,9 +207,10 @@ public int ExecuteNonQuery()
 				{
 					return 0;
 				}
+
 				var batcherCommand = _batcher.Driver.GenerateCommand(
 					_commandType,
-					_sql.ToSqlString(),
+					_batcher.GetSQL(_sql.ToSqlString()),
```
bahusoid: But this is different behavior from other batchers. In other batchers interceptor is executed per each generated command in batch. And this one is executed once for full batch. This difference might be sensible...

maca88: My understanding is that `IInterceptor.OnPrepareStatement` should be called before calling `DbCommand.Prepare`, that is why I added it here. Other batchers like `MySqlClientBatchingBatcher` or `SqlClientBatchingBatcher` are using a command set class (`XXXClientSqlCommandSet`) where each command is prepared separately and added to the set. `GenericBatchingBatcher` does not prepare each command separately, instead `DbCommand.Prepare` is called for the whole batch (only one `DbCommand` is used), so it already differs from other batchers. Calling `IInterceptor.OnPrepareStatement` for a statement that will not be prepared doesn't seem right to me.

bahusoid: `IInterceptor.OnPrepareStatement` can contain some parsing logic and that logic might not expect that it's executed on batch. So this looks like breaking change to me...
That's my main concern.
Also I prefer old way as it allows to safely enable/disable batching and make sure your interceptor will work the same...

bahusoid: > IInterceptor.OnPrepareStatement should be called before calling DbCommand.Prepare... Calling IInterceptor.OnPrepareStatement for a statement that will not be prepared doesn't seem right to me.

I don't get it. To me call to `DbCommand.Prepare` is some internal implementation detail. What importance of it? Comment for `OnPrepareStatement` says `Called when sql string is being prepared.`  - nothing about command preparation... Why do you think `OnPrepareStatement` is bound to be called before `DbCommand.Prepare`?

maca88: > Why do you think OnPrepareStatement is bound to be called before DbCommand.Prepare?

By the name and how it is currently used in NHibernate. Prepared statement IMO is a quite known database feature, which can be also seen when searching it on google. The main reason why I called it only before calling `DbCommand.Prepare` is because I was trying to imitate other batchers, which are calling `OnPrepareStatement` for every generated command prior calling `DbCommand.Prepare` (called only when enabled). Do note that also `BasicResultSetsCommand` calls `OnPrepareStatement` for the whole batch of queries (see [InterceptorOnPrepareStatementTest](https://github.com/nhibernate/nhibernate-core/pull/2285/files#diff-cc4c115a1dff727c1516b41e876dfb4aR145)) and not for each query. 

> What importance of it?

Not much, as `OnPrepareStatement` is IMO mostly used for sql statement modifications and not for some kind of statistics, I updated the PR to call `OnPrepareStatement` for each statement separately.

src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
```diff
@@ -102,6 +104,63 @@ public void MassivePerformanceTest(bool batched)
...
+				s.SetBatchSize(5);
+				for (var i = 0; i < 20; i++)
+				{
+					s.Save(new VerySimple { Id = 1 + i, Name = $"Fabio{i}", Weight = 1.45 + i });
+				}
+
+				tx.Commit();
+
+				// Called for each insert statement and once for generating AbstractBatcher.CurrentCommand
+				Assert.That(interceptor.TotalCalls, Is.EqualTo(21));
```
maca88: I tried to remove the call for `AbstractBatcher.CurrentCommand`, but I didn't found a clean solution to do it. The issue is that without calling `OnPrepareStatement` for the `AbstractBatcher.CurrentCommand` the log message will not contain the changes made by the method.

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
```diff
@@ -183,7 +183,7 @@ public void Append(DbParameterCollection parameters)
 					_commandType = _batcher.CurrentCommand.CommandType;
 				}
 				
-				_sql.Add(_batcher.CurrentCommandSql.Copy());
+				_sql.Add(_batcher.GetSQL(_batcher.CurrentCommandSql.Copy()));
```
bahusoid: Why `SqlString.Copy` call is required?

maca88: In order to avoid modifying its parameters, which are exposed by `SqlString.GetParameters` method. The `Parameter` class is not immutable as `ParameterPosition` and `BackTrack` properties can be modified.

bahusoid: Can't we call it once when final `SqlString` is created instead (`_sql.ToSqlString().Copy()`)?

src/NHibernate.Test/Ado/GenericBatchingBatcherFixture.cs
bahusoid: Are you saying that now `DbCommand` is still generated for each unique statement ( I didn't realize it). So now interceptor call is duplicated for unique commands.

Also it seems old batcher would optimize this to single interceptor call. And with new batcher it's 21 intercpetor calls. Not nice...

maca88: > Are you saying that now DbCommand is still generated for each unique statement

No, only two commands are generated. One is created for `AbstractBatcher.CurrentCommand`, only because it is needed for `AbstractBatcher` in order to work and the other command that is actually executed.

> Also it seems old batcher would optimize this to single interceptor call

You are right, I've modified the code to have only a single interceptor call.

src/NHibernate/AdoNet/GenericBatchingBatcher.cs
maca88: > Can't we call it once when final SqlString is created instead

As other batchers do not copy the `SqlString` prior calling the interceptor method, I've removed the `Copy` call to be consistent with them.

PR #2294, opened 2019-12-22
Add StackExchangeRedis cache provider documentation

fredericDelaporte:
@maca88, may you please add some more stuff to this minimal documentation for Nhibernate.Caches.StackExchangeRedis?

I have just duplicated the RtMemoryCache section, and checked its content was relevant for this cache too. But I guess some more directions on how to set up this cache would be welcome, if you can give some more.


maca88:
I have documented every setting that currently exists.

PR #2295, opened 2019-12-23
Optimize filter applying logic

bahusoid:
Avoid property map preparation logic (`GetColumnsToTableAliasMap`) if filters are disabled or there are no subclasses.
And few other related refactorings

### DIFF
--- a/src/NHibernate/Util/FilterHelper.cs
+++ b/src/NHibernate/Util/FilterHelper.cs
@@ -60,22 +60,15 @@ namespace NHibernate.Util
 
 		public void Render(StringBuilder buffer, string defaultAlias, IDictionary<string, string> propMap, IDictionary<string, IFilter> enabledFilters)
 		{
-			if (filterNames != null)
+			for (int i = 0; i < filterNames.Length; i++)
 			{
-				int max = filterNames.Length;
-				if (max > 0)
+				if (enabledFilters.ContainsKey(filterNames[i]))
 				{
-					for (int i = 0; i < max; i++)
+					string condition = filterConditions[i];
+					if (!string.IsNullOrEmpty(condition))
 					{
-						if (enabledFilters.ContainsKey(filterNames[i]))
-						{
-							string condition = filterConditions[i];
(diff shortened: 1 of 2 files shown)


fredericDelaporte:
CodeFactor seems to have a bug on this one.

bahusoid:
> CodeFactor seems to have a bug on this one.


Yeah I've noticed it long ago. In previous PR it incorrectly claims that 1 issue is fixed and thus tolerates multiple classes in single file (we should really allow it for Test project):
https://www.codefactor.io/repository/github/nhibernate/nhibernate-core/pull/fixed/2288

src/NHibernate/Util/FilterHelper.cs
```diff
@@ -55,50 +55,44 @@ public string Render(String alias, IDictionary<string, IFilter> enabledFilters)
...
 				int startOfProperty = i + FilterImpl.MARKER.Length + 1;
 
-				upTo = condition.IndexOf(" ", startOfProperty);
+				upTo = condition.IndexOf(' ', startOfProperty);
 				upTo = upTo >= 0 ? upTo : condition.Length;
 				string property = condition.Substring(startOfProperty, upTo - startOfProperty);
 
-				string fullColumn = propMap.ContainsKey(property) ? propMap[property] : string.Format("{0}.{1}", defaultAlias, property);
+				if (!propMap.TryGetValue(property, out var fullColumn))
+					fullColumn = string.IsNullOrEmpty(defaultAlias)
```
fredericDelaporte: Robustness improvement by the way, in case some caller may omit the alias some days? Or an actual bug fix?

bahusoid: It's logic from [`GetColumnsToTableAliasMap`](https://github.com/nhibernate/nhibernate-core/blob/3eb16b0ecf6d532bbe64d3d80b48bf8e830ba70d/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs#L3685-L3693)

But now this part is not executed for entities without sub-classes - so I added it here just in case.

PR #2299, opened 2020-01-02
Proper query plan caching for DML LINQ queries

bahusoid:
Properly fixes #2222  by including modified property names in to query key (see https://github.com/nhibernate/nhibernate-core/pull/2300#issuecomment-570933604 for details)

PR #2300, opened 2020-01-03
Do not cache ExpandedQueryExpression query plan if it's based on not cacheable query expression 

### LINKED ISSUE #2298: Dml Linq Update Produce Wrong Sql
Hi,
We have a major issue that produce wrong update statement at production environment.
Linq Dml style update queries creates same sql with different Linq queries please check, i think the issue is query expression plan cache, it does not includes update fields to cache key.

UnitTest: [UpdateRetrieve_TxnIssuer_Fails](https://github.com/gokhanabatay/NHibernateBugTest/blob/master/NHibernateBugTest/TxnIssuerTest.cs)
```c#
 using (ISession session = SessionProvider.ISessionFactory
                                        .WithOptions()
                                        .Interceptor(new ContextInterceptor())
                                        .OpenSession())
{
    using (ITransaction transaction = session.BeginTransaction())
    {
        var listOfGuids = session.Query<TxnIssuer>().Select(x => x.Guid).ToList();

        //Below query updates MrcDailyMoved and MrcDailyMovedDate with success
        session.Query<TxnIssuer>()
                   .Where(x => listOfGuids.Contains(x.Guid))
                   .Update(x => new TxnIssuer
                   {
                        MrcDailyMoved = "Y",
                        MrcDailyMovedDate = DateTime.Now
                   });

        Assert.That(session.Query<TxnIssuer>()
            .Where(x => x.MrcDailyMoved == "Y")
            .ToList().Count == 10);

        //This Update Statement Creates wrong Sql, I think expression cache does not include Update fields to statement so above and below linq creates or use the same statement.
        //Below query updates MrcDailyMoved and MrcDailyMovedDate instead of CycleMoved,CycleMovedDate
       session.Query<TxnIssuer>()
                  .Where(x => listOfGuids.Contains(x.Guid))
                  .Update(x => new TxnIssuer
                  {
                       CycleMoved = "Y",
                       CycleMovedDate = DateTime.Now
                  });

        Assert.That(session.Query<TxnIssuer>()
            .Where(x => x.CycleMoved == "Y")
            .ToList().Count == 10);
    }
}
```
Linq1 resulting SQL:
```sql
update ISSUER set
    MRC_DAILY_MOVED=:p0,
    MRC_DAILY_MOVED_DATE=:p1
  where GUID in (:p2 , :p3 , :p4 , :p5 , :p6 , :p7 , :p8 , :p9 , :p10 , :p11);
:p0 = 'Y' [Type: String (0:0:0)],
:p1 = 20200102 [Type: Int32 (0:0:0)],
:p2 = 1 [Type: Int64 (0:0:0)],
:p3 = 2 [Type: Int64 (0:0:0)],
:p4 = 3 [Type: Int64 (0:0:0)],
:p5 = 4 [Type: Int64 (0:0:0)],
:p6 = 5 [Type: Int64 (0:0:0)],
:p7 = 6 [Type: Int64 (0:0:0)],
:p8 = 7 [Type: Int64 (0:0:0)],
:p9 = 8 [Type: Int64 (0:0:0)],
:p10 = 9 [Type: Int64 (0:0:0)],
:p11 = 10 [Type: Int64 (0:0:0)]
```
Linq2:
```sql
update ISSUER set
    MRC_DAILY_MOVED=:p0,
    MRC_DAILY_MOVED_DATE=:p1
  where GUID in (:p2 , :p3 , :p4 , :p5 , :p6 , :p7 , :p8 , :p9 , :p10 , :p11);
:p0 = 'Y' [Type: String (0:0:0)],
:p1 = 20200102 [Type: Int32 (0:0:0)],
:p2 = 1 [Type: Int64 (0:0:0)],
:p3 = 2 [Type: Int64 (0:0:0)],
:p4 = 3 [Type: Int64 (0:0:0)],
:p5 = 4 [Type: Int64 (0:0:0)],
:p6 = 5 [Type: Int64 (0:0:0)],
:p7 = 6 [Type: Int64 (0:0:0)],
:p8 = 7 [Type: Int64 (0:0:0)],
:p9 = 8 [Type: Int64 (0:0:0)],
:p10 = 9 [Type: Int64 (0:0:0)],
:p11 = 10 [Type: Int64 (0:0:0)]
```

bahusoid:
Fixes #2298. Do not cache `ExpandedQueryExpression` query plan  if it's based on not cacheable query expression

It fixes regression since 5.1 with possible execution of wrong DML query for queries with parameters list (details [here](https://github.com/nhibernate/nhibernate-core/issues/2298#issuecomment-571547485))

In master branch it depends on #2299 (already merged)

And async code should be re-genarated when merging to master (as test for UpdateAsync is not generated in  5.2)

### DIFF
--- a/src/NHibernate.Test/Linq/ConstantTest.cs
+++ b/src/NHibernate.Test/Linq/ConstantTest.cs
@@ -1,8 +1,10 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using NHibernate.Criterion;
 using NHibernate.DomainModel.Northwind.Entities;
 using NHibernate.Engine.Query;
+using NHibernate.Linq;
 using NHibernate.Linq.Visitors;
 using NHibernate.Util;
 using NUnit.Framework;
@@ -275,6 +277,63 @@ namespace NHibernate.Test.Linq
 				cache,
 				Has.Count.EqualTo(0),
 				"Query plan should not be cached.");
+		}
+
+		[Test]
+		public void PlansWithNonParameterizedConstantsAreNotCachedForExpandedQuery()
+		{
+			var queryPlanCacheType = typeof(QueryPlanCache);
+
+			var cache = (SoftLimitMRUCache)
+				queryPlanCacheType
+					.GetField("planCache", BindingFlags.Instance | BindingFlags.NonPublic)
+					.GetValue(Sfi.QueryPlanCache);
+			cache.Clear();
+
+			var ids = new[] {"ANATR", "UNKNOWN"}.ToList();
+			db.Customers.Where(x => ids.Contains(x.CustomerId)).Select(
+				c => new {c.CustomerId, c.ContactName, Constant = 1}).First();
+
+			Assert.That(
+				cache,
+				Has.Count.EqualTo(0),
+				"Query plan should not be cached.");
+		}
+
+		//GH-2298 - Different Update queries - same query cache plan
+		[Test]
+		public void DmlPlansForExpandedQuery()
+		{
+			var queryPlanCacheType = typeof(QueryPlanCache);
+
+			var cache = (SoftLimitMRUCache)
+				queryPlanCacheType
+					.GetField("planCache", BindingFlags.Instance | BindingFlags.NonPublic)
+					.GetValue(Sfi.QueryPlanCache);
+			cache.Clear();
+
+			using (session.BeginTransaction())
+			{
+				var list = new[] {"UNKNOWN", "UNKNOWN2"}.ToList();
+				db.Customers.Where(x => list.Contains(x.CustomerId)).Update(
+					x => new Customer
+					{
+						CompanyName = "Constant1"
+					});
+
+				db.Customers.Where(x => list.Contains(x.CustomerId))
+				.Update(
+					x => new Customer
+					{
+						ContactName = "Constant1"
+					});
+
+				Assert.That(
+					cache.Count,
+					//2 original queries + 2 expanded queries are expected in cache
+					Is.EqualTo(0).Or.EqualTo(4),
+					"Query plans should either be cached separately or not cached at all.");
+			}
 		}
 	}
 }
--- a/src/NHibernate/Impl/ExpressionQueryImpl.cs
+++ b/src/NHibernate/Impl/ExpressionQueryImpl.cs
@@ -150,9 +150,10 @@ namespace NHibernate.Impl
 		}
 	}
 
-	internal class ExpandedQueryExpression : IQueryExpression
+	internal class ExpandedQueryExpression : IQueryExpression, ICacheableQueryExpression
 	{
 		private readonly IASTNode _tree;
+		private ICacheableQueryExpression _cacheableExpression;
 
 		public ExpandedQueryExpression(IQueryExpression queryExpression, IASTNode tree, string key)
 		{
@@ -160,6 +161,7 @@ namespace NHibernate.Impl
 			Key = key;
 			Type = queryExpression.Type;
 			ParameterDescriptors = queryExpression.ParameterDescriptors;
+			 _cacheableExpression = queryExpression as ICacheableQueryExpression;
 		}
 
 		#region IQueryExpression Members
@@ -176,6 +178,8 @@ namespace NHibernate.Impl
 		public IList<NamedParameterDescriptor> ParameterDescriptors { get; private set; }
 
 		#endregion
+
+		public bool CanCachePlan => _cacheableExpression?.CanCachePlan ?? true;
 	}
 
 	internal class ParameterExpander
--- a/src/NHibernate.Test/Async/Linq/ConstantTest.cs
+++ b/src/NHibernate.Test/Async/Linq/ConstantTest.cs
@@ -11,12 +11,13 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using NHibernate.Criterion;
 using NHibernate.DomainModel.Northwind.Entities;
 using NHibernate.Engine.Query;
+using NHibernate.Linq;
 using NHibernate.Linq.Visitors;
 using NHibernate.Util;
 using NUnit.Framework;
-using NHibernate.Linq;
 
 namespace NHibernate.Test.Linq
 {
@@ -264,6 +265,27 @@ namespace NHibernate.Test.Linq
 			await ((from c in db.Customers
 			 where c.CustomerId == "ALFKI"
 			 select new { c.CustomerId, c.ContactName, Constant = 1 }).FirstAsync());
+			Assert.That(
+				cache,
+				Has.Count.EqualTo(0),
+				"Query plan should not be cached.");
+		}
+
+		[Test]
+		public async Task PlansWithNonParameterizedConstantsAreNotCachedForExpandedQueryAsync()
+		{
+			var queryPlanCacheType = typeof(QueryPlanCache);
+
+			var cache = (SoftLimitMRUCache)
+				queryPlanCacheType
+					.GetField("planCache", BindingFlags.Instance | BindingFlags.NonPublic)
+					.GetValue(Sfi.QueryPlanCache);
+			cache.Clear();
+
+			var ids = new[] {"ANATR", "UNKNOWN"}.ToList();
+			await (db.Customers.Where(x => ids.Contains(x.CustomerId)).Select(
+				c => new {c.CustomerId, c.ContactName, Constant = 1}).FirstAsync());
+
 			Assert.That(
 				cache,
 				Has.Count.EqualTo(0),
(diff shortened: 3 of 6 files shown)


src/NHibernate/Linq/NhLinqDmlExpression.cs
```diff
@@ -17,6 +17,10 @@ public NhLinqDmlExpression(QueryMode queryMode, Expression expression, ISessionF
 		{
 			Key = $"{queryMode.ToString().ToUpperInvariant()} {Key}";
 			QueryMode = queryMode;
+
+			//For DML queries parameters check doesn't really work (queries are either not cached or cached incorrectly - see GH-2222 and GH-2298)
+			//So for now disable caching for DML queries
+			CanCachePlan = false;
```
bahusoid: The real fix for #2298 is in `ExpandedQueryExpression`. But as long DML CanCachePlan check is not functional it's better to disable it completely.

src/NHibernate/Impl/ExpressionQueryImpl.cs
```diff
@@ -160,6 +160,7 @@ public ExpandedQueryExpression(IQueryExpression queryExpression, IASTNode tree,
 			Key = key;
 			Type = queryExpression.Type;
 			ParameterDescriptors = queryExpression.ParameterDescriptors;
+			CanCachePlan = (queryExpression as ICacheableQueryExpression)?.CanCachePlan ?? true;
```
bahusoid: Defaults to `true` for consistency with existing checks in `GetHQLQueryPlan`/`GetFilterQueryPlan`

fredericDelaporte:
> It might be considered as missing case for #1544

I do not think so, as #1544 was about non parameterized constants. Here the bug seems to me quite another trouble: the query cache key seems to ignore projections, they would not be part of the query key, causing different "update projections" to be considered equal.

If think this is "just" a bug of the initial feature, Linq DML queries (#391), introduced in 5.0. In the select case, projections should be ignored by the query cache key because they are not translated to SQL, and that is likely why they are ignored. But this seems to have been overlooked when they have been used for expressing DML queries.

> But looks worth fixing in 5.2 branch anyway.

And back-ported down to 5.0 indeed...

fredericDelaporte:
Or is it something like `ExpandedQueryExpression` hiding away the projection, which would not occur with non expanded expressions? I need to take a deeper look.

bahusoid:
So the current issue is due to the way UPDATE query is prepared in `DmlExpressionRewriter.ConvertAssignmentsToDictionaryExpression`. It adds the following `Select` to query:
```C#
selectPartOfUpdateQuery.Select(x => new Dictionary<string, object>()
{
{"propertyPath1", updateValue1},
....
};
```

The problem here is that property path is stored as string constant and our `ExpressionParameterVisitor` detects it as query parameter and replaces it with named parameter like `p1`. So cache key is generated without actual info about modified properties  - same key will be generated for all properties.

bahusoid:
I have an idea how to fix it (see https://github.com/bahusoid/nhibernate-core/tree/todo_gh2222).

Instead of dictionary I generate a block of variable assignments:
```C#
var propertyPath1 = propValue1;
var propertyPath2 = propValue2;
```
So in cache key it's added as `(Assign(propertyPath1, p1)propertyPath1)` with modified property names in key. Maybe we should make this fix in 5.2 instead?

src/NHibernate/Linq/NhLinqDmlExpression.cs
fredericDelaporte: I still have to do more checks, but I now think we have a combination of bugs:
 - `NhLinqExpression` may be wrapped inside `ExpandedQueryExpression`, causing their "custom" cache disabling logic to be defeated in `QueryPlanCache`. This PR fix this point, in a good way I think
 - Projections seem not included in the query plan cache key at least in the DML case, causing different projections to be considered equal. This change I am commenting is a work-around for this (including the private to protected change). I think it should be handled in a dedicated PR.
 - "Thanks" to #2222, the previous trouble is hidden when expression are not expanded.

Since the trouble only occurs with queries using parameter lists, and is not a regression, I am not convinced we should issue patches for it. A fix in 5.3 would be enough IMO.
If fixing in 5.3, better fix it with two PR, one for properly handling the cache disabling logic as this PR does, another one for handling the projection cache key trouble.
Otherwise, if still fixing in 5.2 and backporting all the way down to 5.0, ok for keeping the workaround of disabling the cache here.

src/NHibernate.Test/Linq/ConstantTest.cs
```diff
@@ -257,6 +259,41 @@ public void PlansAreCached()
 			}
 		}
 
+		//GH-2298 - Different Update queries - same query cache plan
+		[Test]
+		public void DmlPlansAreProperlyHandled()
```
fredericDelaporte: This test would be better put into `LinqBulkManipulation` namespace I think.

fredericDelaporte:
So, overall, I am more for fixing this in 5.3, in two PR: one for fixing `ExpandedQueryExpression` hiding away `CanCachePlan`, another one for fixing the query cache plan generation for DML queries.
Your idea for this later case looks good.
@gliljas, you may be interested in this conversation.

bahusoid:
> So, overall, I am more for fixing this in 5.3, in two PR:

OK

src/NHibernate/Impl/ExpressionQueryImpl.cs
fredericDelaporte: Resolving `CanCachePan` in constructor could lead to some future bug, by example if someday the query translation, which mutates `CanCachePlan`, is moved after constructing this. (Granted, it is very unlikely since the constructor need the ast tree.)

But from a robustness standpoint, it would be better to have this evaluated in the property call.

gokhanabatay:
> > So, overall, I am more for fixing this in 5.3, in two PR:
> 
> OK

@bahusoid,  will you fix the issue in 5.2.x or release 5.3 I don't understand?
When will you plan to release if you decide to fix?
If you decide not to fix, then what will we do? We have a lot of linq update expressions

src/NHibernate.Test/Linq/ConstantTest.cs
bahusoid: I prefer to keep it here. IMHO it's very similar to `PlansWithNonParameterizedConstantsAreNotCachedForExpandedQuery` only for DML query. Also I think it's better to keep tests with a bit hacky reflection logic in one place (part about retrieving query plan cache).

src/NHibernate/Impl/ExpressionQueryImpl.cs
bahusoid: Ok. Moved to property getter

bahusoid:
> When will you plan to release if you decide to fix?

@gokhanabatay I'm not doing releases - it's not really up to me. Yes for now it's planned to be fixed in 5.3 only. Though this PR can be merged in 5.2 too.

bahusoid:
DML test fails due to changes done #2229. After #2299 is merged - it's going to pass.

fredericDelaporte:
For some reason I do not understand, the Travis Firebird build fails to start Firebird repeatedly on this PR. I consider force-merging this, likely tomorrow or on Tuesday, in case someone wants to react before I do it.

(Also for some other reason I guess, Travis was no more seeing I am belonging to the NHibernate organization, forbidding me to relaunch builds. On Github side, Travis had access granted to my organizations, but it was still seeing none. Revoking Travis access on Github and signing up again in Travis solved the trouble.)

src/NHibernate.Test/Linq/ConstantTest.cs
```diff
@@ -4,6 +4,7 @@
 using NHibernate.Criterion;
 using NHibernate.DomainModel.Northwind.Entities;
 using NHibernate.Engine.Query;
+using NHibernate.Linq;
```
fredericDelaporte: This causes an async regen change. Going to push it.

bahusoid:
> For some reason I do not understand, the Travis Firebird build fails to start Firebird repeatedly on this PR.

Seems https://github.com/nhibernate/nhibernate-core/commit/8a7375a9f17bcf689643cc67b285879bf954b9e0 should fix it

hazzik:
Yes it should fix it. They updated the default distro si it started failing.

hazzik:
> Also for some other reason I guess, Travis was no more seeing I am belonging to the NHibernate organization, forbidding me to relaunch builds.

Try to log out and login again. This fixed similar issue for me.

hazzik:
Also try to press “sync account” here: https://travis-ci.org/account/preferences

hazzik:
There is no settings in Travis to restrict access to the repos, so I assume it depends solely on the permissions fetched from GitHub

fredericDelaporte:
I had tried the sync: no effect; checked the GitHub settings, nothing wrong. So I had go for revoking GitHub authorizations and signing-up again into Travis.

>>Seems 8a7375a should fix it
>
>Yes it should fix it. They updated the default distro so started failing.

So, shall we ignore the failure, or cherry-pick this commit directly into 5.2.x and 5.1.x (as it seems it was a direct trivial commit to master anyway)?

hazzik:
> So, shall we ignore the failure, or cherry-pick this commit directly into 5.2.x and 5.1.x (as it seems it was a direct trivial commit to master anyway)?

Yes, please cherry-pick

fredericDelaporte:
I have cherry-picked it directly to 5.2.x and 5.1.x branches. I do not think it should be squashed in this PR.
(But no need to remove it or rebase, squashing will do it.)

fredericDelaporte:
Well, the "files changed" was still showing the travis one, so I have rebased and squashed the PR just in case.

PR #2303, opened 2020-01-12
Release 5.2.7

fredericDelaporte:
Release ready.


fredericDelaporte:
@gokhanabatay, may you check that the TeamCity release build artifacts (which should be available in less than one hour) has no more the issue you were reporting for your application in #2298?

PR #2304, opened 2020-01-13
Add logging for QueryBatch similar to Loader.DoQuery

### LINKED ISSUE #2296: Missing Row Count in Debug Log for Future queries
Using this (simplified) code (running with Dotnet Core):
```
public void Execute(ISessionFactory factory)
{
        using(var session = factory.OpenSession())
        using(var tx = session.BeginTransaction())
        {
            // Total row number is not written to log
             var future = session.CreateQuery("select Id As Id from Post")
     		                 .SetResultTransformer(Transformers.AliasToBean<PostId>())
		                 .Future<PostId>();
	      future.ToList(); 

               // This is ok - total row number is written to log   
               // session.CreateCriteria<Blog>().List<Blog>();    

               tx.Commit();
        }
}

public class PostId
{
     public int Id { get; set; }
}
```

1. With **NHibernate 5.2.6**, The total 'row count' log message is **Not** written to the log
    I only receive the following: 
   ` "result row: "` 
   (from **_NHibernate.Loader.Loader:_**) 
    with No row count information! 

2. With the previous **Nhibernate 5.1.3**, the total was written and we had (for example):   
  `"result set row: 1"` 
    (from _**NHibernate.Impl.MultiQueryImpl**_)

    AND we also got:

    `"done processing result set (10 rows)"` 
    (from _**NHibernate.Impl.MultiQueryImpl**_)

  
3. I see that  _MultiQueryImpl_ is obsolete
https://github.com/nhibernate/nhibernate-core/blob/4e54de2f82fda24b04f6f432facdb5a96ef625a2/src/NHibernate/Impl/MultiQueryImpl.cs#L22

4. **BUT**, for example, when the following code snippet is run (with **NHibernate 5.2.6**):
` session.CreateCriteria<Blog>().List<Blog>();`
we do get the expected msg in the log from Loader.cs
`"done processing result set (5 rows)"`
 (from **_NHibernate.Loader.Loader:_**)

5. So, can you please provide this missing functionality?

bahusoid:
Fixes #2296 
Added logging from [`Loader.DoQuery`](https://github.com/nhibernate/nhibernate-core/blob/df6cbbe12bb00f22fab1fc0ad024a1b1a2f13a8c/src/NHibernate/Loader/Loader.cs#L505-L534) to QueryBatch

We might also fix this one in 5.2.

It's not exactly a regression as debug logging is not a public API.  And this logging is not present for criteria `MultiCriteriaImpl` but only in `MultiQueryImpl`.
But I agree logging for `List()` and for `Future().List()`  should be the same if possible.
And this logging is missing for hql Future queries in 5.2 - so might be considered as regression.

What do you think?

### DIFF
--- a/src/NHibernate/Multi/QueryBatchItemBase.cs
+++ b/src/NHibernate/Multi/QueryBatchItemBase.cs
@@ -16,6 +16,8 @@ namespace NHibernate.Multi
 	/// </summary>
 	public abstract partial class QueryBatchItemBase<TResult> : IQueryBatchItem<TResult>, IQueryBatchItemWithAsyncProcessResults
 	{
+		private static readonly INHibernateLogger Log = NHibernateLogger.For(typeof(QueryBatch));
+
 		protected ISessionImplementor Session;
 		private List<EntityKey[]>[] _subselectResultKeys;
 		private List<QueryInfo> _queryInfos;
@@ -175,6 +177,7 @@ namespace NHibernate.Multi
 
 			var dialect = Session.Factory.Dialect;
 			var hydratedObjects = new List<object>[_queryInfos.Count];
+			var isDebugLog = Log.IsDebugEnabled();
 
 			using (Session.SwitchCacheMode(_cacheMode))
 			{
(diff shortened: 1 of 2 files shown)


src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -16,6 +16,8 @@ namespace NHibernate.Multi
 	/// </summary>
 	public abstract partial class QueryBatchItemBase<TResult> : IQueryBatchItem<TResult>, IQueryBatchItemWithAsyncProcessResults
 	{
+		private static readonly INHibernateLogger Log = NHibernateLogger.For(typeof(QueryBatch));
```
bahusoid: `QueryBatch` type here is intentional.. Don't want to create separate logger for each result type

fredericDelaporte:
Rebased (and squashed) to have the Travis fix commit in this PR branch.

PR #2310, opened 2020-02-02
Fix transaction leak

### LINKED ISSUE #2144: AdoTransaction memory leak (5.2.5)
I'm investigation memory leak in my project. I found one of memory leak reasons inside NHibernate assembly. AdoTransaction objects created in NHibernate.AdoNet.AbstractBatcher.Prepare() via accessing property NHibernate.AdoNet.ConnectionManager.Transaction are not disposing.

image

So i tried to fix it.

First change in NHibernate\Transaction\AdoTransaction.cs
image

Second change in NHibernate\AdoNet\ConnectionManager.cs 
image

I'm not sure that thouse changes are correct but now all work fine. Can you check it?

fredericDelaporte:
Fixes #2144

Instead of disposing `ConnectionManager._transaction` in `AfterTransaction` as proposed in #2144, this obsoletes `ConnectionManager.Transaction` in favor of a new `ConnectionManager.CurrentTransaction` property which does not create a transaction when none are already created, but yields `null`.

So now any existing transaction should be the result of a call to `BeginTransaction`, and its disposal is always a responsibility of the caller.

### DIFF
--- a/src/NHibernate.Test/Ado/BatcherFixture.cs
+++ b/src/NHibernate.Test/Ado/BatcherFixture.cs
@@ -1,4 +1,3 @@
-using System.Collections;
 using NHibernate.AdoNet;
 using NHibernate.Cfg;
 using NUnit.Framework;
@@ -44,11 +43,11 @@ namespace NHibernate.Test.Ado
 		private void Cleanup()
 		{
 			using (ISession s = Sfi.OpenSession())
-			using (s.BeginTransaction())
+			using (var t = s.BeginTransaction())
 			{
 				s.CreateQuery("delete from VerySimple").ExecuteUpdate();
 				s.CreateQuery("delete from AlmostSimple").ExecuteUpdate();
-				s.Transaction.Commit();
+				t.Commit();
 			}
 		}
 
--- a/src/NHibernate.Test/Async/TransactionTest/TransactionFixture.cs
+++ b/src/NHibernate.Test/Async/TransactionTest/TransactionFixture.cs
@@ -145,7 +145,8 @@ namespace NHibernate.Test.TransactionTest
 
 				using (var s1 = builder.Interceptor(new TestInterceptor(1, flushOrder)).OpenSession())
 				using (var s2 = builder.Interceptor(new TestInterceptor(2, flushOrder)).OpenSession())
-				using (var s3 = s1.SessionWithOptions().Connection().Interceptor(new TestInterceptor(3, flushOrder)).OpenSession())
+				using (var s3 = s1.SessionWithOptions().Connection().Interceptor(new TestInterceptor(3, flushOrder))
+				                  .OpenSession())
 				using (var t = s.BeginTransaction())
 				{
 					var p1 = new Person();
@@ -234,8 +235,114 @@ namespace NHibernate.Test.TransactionTest
 			using (var s = Sfi.WithOptions().Connection(connection).OpenSession())
 			{
 				CacheablePerson person = null;
-				Assert.DoesNotThrowAsync(async () => person = await (s.LoadAsync<CacheablePerson>(id)), "Failed loading entity from second level cache.");
+				Assert.DoesNotThrowAsync(
+					async () => person = await (s.LoadAsync<CacheablePerson>(id)),
+					"Failed loading entity from second level cache.");
 				Assert.That(person.NotNullData, Is.EqualTo(notNullData));
+			}
+		}
+
+		[Test]
+		public async Task CanCommitFromSessionTransactionAsync()
+		{
+			int id;
+			using (var s = OpenSession())
+			{
+				Assert.That(s.GetCurrentTransaction(), Is.Null);
+				using (s.BeginTransaction())
+				{
+					var person = new Person();
+					await (s.SaveAsync(person));
+					id = person.Id;
+
+					await (s.GetCurrentTransaction().CommitAsync());
+				}
+				Assert.That(s.GetCurrentTransaction(), Is.Null);
+			}
+
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				var person = await (s.GetAsync<Person>(id));
+				Assert.That(person, Is.Not.Null);
+				await (t.CommitAsync());
+			}
+		}
+
+		[Test]
+		public async Task CanRollbackFromSessionTransactionAsync()
+		{
+			int id;
+			using (var s = OpenSession())
+			{
+				Assert.That(s.GetCurrentTransaction(), Is.Null);
+				using (s.BeginTransaction())
+				{
+					var person = new Person();
+					await (s.SaveAsync(person));
+					id = person.Id;
+
+					await (s.GetCurrentTransaction().RollbackAsync());
+
+					// Need to check before leaving the current using, otherwise the rollback could be the result of the
+					// disposing.
+					using (var s2 = OpenSession())
+					using (var t2 = s2.BeginTransaction())
+					{
+						person = await (s2.GetAsync<Person>(id));
+						Assert.That(person, Is.Null);
+						await (t2.CommitAsync());
+					}
+				}
+				Assert.That(s.GetCurrentTransaction(), Is.Null);
+			}
+		}
+
+		[Test, Obsolete]
+		public async Task CanCommitFromSessionObsoleteTransactionAsync()
+		{
+			int id;
+			using (var s = OpenSession())
+			using (s.BeginTransaction())
+			{
+				var person = new Person();
+				await (s.SaveAsync(person));
+				id = person.Id;
+
+				await (s.Transaction.CommitAsync());
+			}
+
+			using (var s = OpenSession())
+			using (var t = s.BeginTransaction())
+			{
+				var person = await (s.GetAsync<Person>(id));
+				Assert.That(person, Is.Not.Null);
+				await (t.CommitAsync());
+			}
+		}
+
+		[Test, Obsolete]
+		public async Task CanRollbackFromSessionObsoleteTransactionAsync()
+		{
+			int id;
+			using (var s = OpenSession())
+			using (s.BeginTransaction())
+			{
+				var person = new Person();
+				await (s.SaveAsync(person));
+				id = person.Id;
+
+				await (s.Transaction.RollbackAsync());
+
+				// Need to check before leaving the current using, otherwise the rollback could be the result of the
+				// disposing.
+				using (var s2 = OpenSession())
+				using (var t2 = s2.BeginTransaction())
+				{
+					person = await (s2.GetAsync<Person>(id));
+					Assert.That(person, Is.Null);
+					await (t2.CommitAsync());
+				}
 			}
 		}
 	}
--- a/src/NHibernate.Test/Async/TransactionTest/TransactionFixture.cs
+++ b/src/NHibernate.Test/Async/TransactionTest/TransactionFixture.cs
@@ -9,11 +9,10 @@
 
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using NHibernate.Linq;
 using NUnit.Framework;
+using NHibernate.Linq;
 
 namespace NHibernate.Test.TransactionTest
 {
@@ -96,41 +95,30 @@ namespace NHibernate.Test.TransactionTest
 			{
 				using (ITransaction t = s.BeginTransaction())
 				{
-					Assert.AreSame(t, s.Transaction);
-					Assert.IsFalse(s.Transaction.WasCommitted);
-					Assert.IsFalse(s.Transaction.WasRolledBack);
+					Assert.AreSame(t, s.GetCurrentTransaction());
+					Assert.IsFalse(s.GetCurrentTransaction().WasCommitted);
+					Assert.IsFalse(s.GetCurrentTransaction().WasRolledBack);
 					await (t.CommitAsync());
 
-					// ISession.Transaction returns a new transaction
-					// if the previous one completed!
-					Assert.IsNotNull(s.Transaction);
-					Assert.IsFalse(t == s.Transaction);
+					// ISession.GetCurrentTransaction() returns null if the transaction is completed.
+					Assert.IsNull(s.GetCurrentTransaction());
 
 					Assert.IsTrue(t.WasCommitted);
 					Assert.IsFalse(t.WasRolledBack);
-					Assert.IsFalse(s.Transaction.WasCommitted);
-					Assert.IsFalse(s.Transaction.WasRolledBack);
 					Assert.IsFalse(t.IsActive);
-					Assert.IsFalse(s.Transaction.IsActive);
 				}
 
 				using (ITransaction t = s.BeginTransaction())
 				{
 					await (t.RollbackAsync());
 
-					// ISession.Transaction returns a new transaction
-					// if the previous one completed!
-					Assert.IsNotNull(s.Transaction);
-					Assert.IsFalse(t == s.Transaction);
+					// ISession.GetCurrentTransaction() returns null if the transaction is completed.
+					Assert.IsNull(s.GetCurrentTransaction());
 
 					Assert.IsTrue(t.WasRolledBack);
 					Assert.IsFalse(t.WasCommitted);
 
-					Assert.IsFalse(s.Transaction.WasCommitted);
-					Assert.IsFalse(s.Transaction.WasRolledBack);
-
 					Assert.IsFalse(t.IsActive);
-					Assert.IsFalse(s.Transaction.IsActive);
 				}
 			}
 		}
--- a/src/NHibernate.Test/TransactionTest/TransactionFixture.cs
+++ b/src/NHibernate.Test/TransactionTest/TransactionFixture.cs
@@ -1,8 +1,6 @@
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using NHibernate.Linq;
 using NUnit.Framework;
 
 namespace NHibernate.Test.TransactionTest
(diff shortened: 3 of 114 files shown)


src/NHibernate/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
```diff
@@ -188,9 +188,9 @@ protected virtual void DropTemporaryTableIfNecessary(IQueryable persister, ISess
 			{
 				IIsolatedWork work = new TmpIdTableDropIsolatedWork(persister, log, session);
 
-				if (ShouldIsolateTemporaryTableDDL())
+				if (ShouldIsolateTemporaryTableDDL() && session.ConnectionManager.CurrentTransaction != null)
 				{
-					session.ConnectionManager.Transaction.RegisterSynchronization(
+					session.ConnectionManager.CurrentTransaction.RegisterSynchronization(
```
fredericDelaporte: This is a "side" bug-fix indeed. The temporary table would never have been dropped previously, in case the operation occurred outside of a transaction. The NHibernate transaction would have received the drop to do "after transaction", but the synchs are executed only on actual commit, not upon `ConnectionManager.AfterTransaction` call.

src/NHibernate/Transaction/AdoTransaction.cs
```diff
@@ -370,6 +370,7 @@ protected virtual void Dispose(bool isDisposing)
 					// don't dispose of multiple times.
 					return;
 				}
+				_isAlreadyDisposed = true;
```
fredericDelaporte: Change kept from #2144: flagging it at the end of this method instead of as soon as possible cause the method to be vulnerable to an endless loop, in case a `Dispose` is performed inside `AfterTransaction`, called later by this method.

This change was required by #2144 since it was calling `Dispose` in `AfterTransaction`. This PR here does not do it and so does not actually need that change, but still it is safer to have it.

src/NHibernate.Test/Ado/BatcherFixture.cs
```diff
@@ -44,11 +43,11 @@ public void OneRoundTripInserts()
 		private void Cleanup()
 		{
 			using (ISession s = Sfi.OpenSession())
-			using (s.BeginTransaction())
+			using (var t = s.BeginTransaction())
 			{
 				s.CreateQuery("delete from VerySimple").ExecuteUpdate();
 				s.CreateQuery("delete from AlmostSimple").ExecuteUpdate();
-				s.Transaction.Commit();
+				t.Commit();
```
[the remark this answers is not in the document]
bahusoid: I think it's better to isolate such changes in tests in separate commit to simplify review (at least in PR).
And you should add some Obsolete test that still uses `s.Transaction.Commit();` to make sure this case still works.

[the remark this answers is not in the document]
fredericDelaporte: That would mean a commit that does not even compile, since we treat warnings as errors. It is troublesome.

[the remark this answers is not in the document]
bahusoid: Just mark test as `[Obsolete]` and it should be OK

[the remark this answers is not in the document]
fredericDelaporte: What? These tests are not obsolete, they need to be updated. We have to keep them, not drop them at next major release.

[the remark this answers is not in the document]
bahusoid: You:
> That would mean a commit that does not even compile, since we treat warnings as errors. It is troublesome.

Me:
> Just mark test as [Obsolete] and it should be OK

And you've just committed tests marked as `[Obsolete]` for legacy `s.Transaction`... That's all I meant.. So not sure what's you `What?` about..

src/NHibernate/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
bahusoid: This use case got me thinking if it's useful option for someone. If I get it right in old implementation you can supply some after transaction events before actual DB transaction is created.

src/NHibernate.Test/Ado/BatcherFixture.cs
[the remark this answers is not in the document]
fredericDelaporte: Misunderstanding all the way long:
The following:

> That would mean a commit that does not even compile, since we treat warnings as errors. It is troublesome.

Was refeting to:

> I think it's better to isolate such changes in tests in separate commit to simplify review (at least in PR).

So when you replied:

> Just mark new test as `[Obsolete]` and it should be OK

I thought you were willing me to flag as obsolete all the changed tests to avoid altering them an "ease" the review. Thus my reaction.

So the trouble remain: how do you want to split in separated commits the fix of the trouble and the fix of the tests, without issuing a commit that cannot compile?

src/NHibernate.Test/TransactionTest/TransactionFixture.cs
```diff
@@ -230,5 +230,97 @@ public void WhenCommittingItemsWillAddThemTo2ndLevelCache()
...
+		{
+			int id;
+			using (var s = OpenSession())
+			using (s.BeginTransaction())
+			{
+				var person = new Person();
+				s.Save(person);
+				id = person.Id;
+
+				s.CurrentTransaction.Commit();
```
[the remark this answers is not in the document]
fredericDelaporte: why not

src/NHibernate/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
fredericDelaporte: This sounds as quite a corner case.
You mainly could provide "after transaction" events that would have been happily dropped without ever being executed on next non-transacted statement run against the session...
Anyway, the property being just obsoleted, and kept with all its (anti) features, there are no breaking changes here.
Those adapting to use the new property we will have to adapt for starting the transaction first. I do not think there is any actual need for setting after transaction events before starting the transaction, simply because if you want to do anything out of the transaction on the session after setting them, you will lose them.

src/NHibernate.Test/Ado/BatcherFixture.cs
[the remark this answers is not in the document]
bahusoid: I don't understand why non-compilable commit is an issue here. When you push multiple commits at once - only last one will be compiled.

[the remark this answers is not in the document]
fredericDelaporte: Just a principle I think should be followedto (and I do not think I am alone): do not commit something that does not compile.
A separated commit should always compile. It can be cherry-picked, we can branch from it, ... It is not really "nice" to leave a commit which does not compile, even in a tree just meant for PR review.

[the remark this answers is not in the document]
fredericDelaporte: Done

src/NHibernate.Test/TransactionTest/TransactionFixture.cs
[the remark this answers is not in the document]
fredericDelaporte: And done

src/NHibernate/ISession.cs
```diff
@@ -764,6 +764,11 @@ public partial interface ISession : IDisposable
 		/// </summary>
 		ITransaction Transaction { get; }
 
+		/// <summary>
+		/// The current transaction if any, else <see langword="null" />.
+		/// </summary>
+		ITransaction CurrentTransaction { get; }
```
bahusoid: With 2 commits breaking changes are spotted quite easily :) Do we really need it?

src/NHibernate/Async/Context/ThreadLocalSessionContext.cs
```diff
@@ -32,18 +32,19 @@ private static async Task CleanupAnyOrphanedSessionAsync(ISessionFactory factory
 
 				try
 				{
-					if (orphan.Transaction != null && orphan.Transaction.IsActive)
+					try
 					{
-						try
+						var rollbackTask = orphan.CurrentTransaction?.RollbackAsync(cancellationToken);
```
bahusoid: Are you sure `IsActive` check is no more needed here?

src/NHibernate/ISession.cs
fredericDelaporte: Indeed I forgot, have to add it as an extension method, ...

We need it for obsoleting `ISession.Transaction`, unless we consider the return value from `BeginTransaction` is enough and that is the user concern to pass around the transaction if he needs it / needs to check if a transaction is ongoing.
Otherwise we could keep that `Transaction` property non-obsoleted, and just do a breaking change at 6.0. But it is not developer friendly to do breaking changes without warning in previous versions allowing to anticipate them.

src/NHibernate/Async/Context/ThreadLocalSessionContext.cs
fredericDelaporte: It was previously needed because the `Transaction` was anyway never null, and was an inactive one if no transaction was ongoing.
Due to the `Transaction` property being not yet removed, it can still be created without being started altogether, so in fact the `IsActive` check should remain for now.

bahusoid: > It was previously needed because the Transaction was anyway never null, and was an inactive one if no transaction was ongoing.

`ITransaction.IsActive` might contain some custom logic for determination. That's my main concern here.. To me it seems `IsActive` check should remain even after  `Transaction` removal.

fredericDelaporte: It seems highly unlikely to me that such an implementation would be supported by NHibernate. It is not expected that a transaction might go inactive without an explicit commit or rollback call and without still supporting such a call if it goes inactive "by itslef". Most code just rollback or commit transactions it has explicitly started without checking that flag.

fredericDelaporte: But anyway, we can leave that for further reconsideration when `Transaction` properties would be actually removed.

src/NHibernate/Async/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
```diff
@@ -68,9 +68,9 @@ protected virtual async Task DropTemporaryTableIfNecessaryAsync(IQueryable persi
 			{
 				IIsolatedWork work = new TmpIdTableDropIsolatedWork(persister, log, session);
 
-				if (ShouldIsolateTemporaryTableDDL())
+				if (ShouldIsolateTemporaryTableDDL() && session.ConnectionManager.CurrentTransaction != null)
 				{
-					session.ConnectionManager.Transaction.RegisterSynchronization(
+					session.ConnectionManager.CurrentTransaction.RegisterSynchronization(
```
bahusoid: If you think that's it's worth adding `CurrentTransaction` to interface than we should use it instead  everywhere to make sure implementation is not ovveriden.

src/NHibernate/ISession.cs
```diff
@@ -763,14 +771,9 @@ public partial interface ISession : IDisposable
 		/// Get the current Unit of Work and return the associated <c>ITransaction</c> object.
 		/// </summary>
 		// Since v5.3
-		[Obsolete("Use CurrentTransaction instead, and check for nulls.")]
+		[Obsolete("Use GetCurrentTransaction extension method instead, and check for nulls.")]
```
bahusoid: Why `nulls`. Shouldn't it be simply `null`? Typo?

src/NHibernate.Test/Async/TransactionTest/TransactionFixture.cs
```diff
@@ -96,41 +95,32 @@ public async Task WasCommittedOrRolledBackAsync()
...
+					Assert.IsFalse(s.GetCurrentTransaction().WasRolledBack);
 					await (t.CommitAsync());
 
-					// ISession.Transaction returns a new transaction
-					// if the previous one completed!
-					Assert.IsNotNull(s.Transaction);
-					Assert.IsFalse(t == s.Transaction);
+					// ISession.GetCurrentTransaction() does not return a new transaction
+					// if the previous one is completed.
+					Assert.IsNull(s.GetCurrentTransaction());
```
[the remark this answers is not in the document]
fredericDelaporte: Yes, `Transaction` property's backing field is set to `null` in `ConnectionManager.AfterTransaction`. So if you read it after commit/rollback, you have a new instance, not `begun`, not committed or rollbacked.
I think people using the property are used to that, and that the xml comment `Get the current transaction if any, else <see langword="null" />` will be enough for them to understand they will get null after commit or rollback.
The properties `WasCommitted` / `Rollacked` serve for those who keep their own reference to the transaction object (and serve likely in after transaction events too).

src/NHibernate/Async/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
fredericDelaporte: Previous `Transaction` property implementation was not either (not `virtual`). So only a custom `ISession` may benefit of this change. (And of course only after we have put `CurrentTransaction` into the interface.)

fredericDelaporte: I am not committing your suggestion as it needs more changes to be effective. I will add another commit.

src/NHibernate/ISession.cs
fredericDelaporte: Just a kind of contraction of `null values`, not really good language anyway. Will adjust.

src/NHibernate/Async/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
fredericDelaporte: In fact this change is invalid.
`session` is not an `ISession`, but an `ISessionImplementor`. It has never had a transaction property, and so the old code was just right going through the connection manager.

src/NHibernate.Test/Async/TransactionTest/TransactionFixture.cs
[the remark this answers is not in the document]
bahusoid: > I think people using the property are used to that

What about people who didn't use this property before. Note should be added for them.

Actually your new comment is not helpful. Previous one explains strange unexpected behavior - we still in transaction scope but `s.Transaction` is already assigned to new instance. And after refactoring comment simply makes no sense  - why would anyone expect here new transaction. It's null that should be explained here.
So it should be something like:
`// ISession.GetCurrentTransaction() returns null if previous transaction is completed!`

src/NHibernate/Async/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
bahusoid: > So only a custom ISession may benefit of this change. (And of course only after we have put CurrentTransaction into the interface.)

I don't get this point. If we add this to interface we can expect custom implementations. No?
So does it make sense to add `GetCurrentTransaction` to interface if it may results in different behavior - and here custom logic won't be used. Maybe it's enough to have it as extension methods?

> I am not committing your suggestion as it needs

I'm using suggested change just for illustration. I don''t expect it to be committed.

fredericDelaporte: We cannot do what you ask anyway. It was not supported by the previous `Transaction` property because it does not exist on the type this `session` variable is. And the change you are asking here is unrelated to what this PR aims to fix.

bahusoid: My main point here -  it seems `ISession/IStatelessSession` is a wrong place for `Transaction/CurrentTransaction` property. It seems we should keep it simply as extension, so if you agree just remove TODO about adding new method to interface. And if someone needs extensibility - `CurrentTransaction` should be added directly to `ISessionImplementor` (with different PR)

P.S. This PR still contains breaking changes in `IStatelessSession`.

src/NHibernate.Test/Async/TransactionTest/TransactionFixture.cs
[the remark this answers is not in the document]
fredericDelaporte: > What about people who didn't use this property before. Note should be added for them.

This:

> the xml comment `Get the current transaction if any, else <see langword="null" />`

Should also tell them. Anyway I am rewording it to make it more explicit.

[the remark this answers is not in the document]
bahusoid: > `Get the current transaction if any, else <see langword="null" />`

I would change it to something like:
`Get active current tranasaction if any, else <see langword="null">. Please note it returns null if current transaction is completed (commited or rolledback)`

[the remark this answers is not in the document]
fredericDelaporte: I have put:

> Get the current transaction if any is ongoing, else <see langword="null" />.

Clear enough I think, and concise.

src/NHibernate/Async/Hql/Ast/ANTLR/Exec/AbstractStatementExecutor.cs
fredericDelaporte: Fixed, previous commits squashed, new changes are in the fixup commit.
Some glitches in previous fixup commits have been adjusted while squashing.

fredericDelaporte:
Rebased and fixup squashed.

fredericDelaporte:
Rebased for checking the generator update was not requiring a regen for this PR.

hazzik:
I did a merge as 2 commits have better representation of the work done.

PR #2316, opened 2020-02-22
Add multiple arguments support for ISQLFunction

maca88:
This is a continuation of #2061 where the code was extracted from #2079 in order to reduce the PR scope for easier review. Unfortunately, `ModulusFunctionTypeDetector` cannot be tested in this PR due to the current Linq behavior, which always executes `mod` on the client side, so tests were not ported from #2079.

### DIFF
--- a/src/NHibernate/Dialect/Function/ClassicCountFunction.cs
+++ b/src/NHibernate/Dialect/Function/ClassicCountFunction.cs
@@ -14,6 +14,8 @@ namespace NHibernate.Dialect.Function
 		{
 		}
 
+		// Since v5.3
+		[Obsolete("Use GetReturnType method instead.")]
 		public override IType ReturnType(IType columnType, IMapping mapping)
 		{
 			return NHibernateUtil.Int32;
--- a/src/NHibernate/Dialect/Function/ClassicCountFunction.cs
+++ b/src/NHibernate/Dialect/Function/ClassicCountFunction.cs
@@ -20,5 +20,13 @@ namespace NHibernate.Dialect.Function
 		{
 			return NHibernateUtil.Int32;
 		}
+
+		public override IType GetReturnType(IEnumerable<IType> argumentTypes, IMapping mapping, bool throwOnError)
+		{
+			// 6.0 TODO: return NHibernateUtil.Int32;
+#pragma warning disable 618
+			return ReturnType(argumentTypes.FirstOrDefault(), mapping);
+#pragma warning restore 618
+		}
 	}
 }
--- a/src/NHibernate/Dialect/Function/ClassicCountFunction.cs
+++ b/src/NHibernate/Dialect/Function/ClassicCountFunction.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using NHibernate.Engine;
 using NHibernate.Type;
 
--- a/src/NHibernate/Dialect/Function/CountQueryFunctionInfo.cs
+++ b/src/NHibernate/Dialect/Function/CountQueryFunctionInfo.cs
@@ -9,6 +9,8 @@ namespace NHibernate.Dialect.Function
 	{
 		public CountQueryFunctionInfo() : base("count", true) { }
 
+		// Since v5.3
+		[Obsolete("Use GetReturnType method instead.")]
 		public override IType ReturnType(IType columnType, IMapping mapping)
 		{
 			return NHibernateUtil.Int64;
--- a/src/NHibernate/Dialect/Function/CountQueryFunctionInfo.cs
+++ b/src/NHibernate/Dialect/Function/CountQueryFunctionInfo.cs
@@ -15,5 +15,13 @@ namespace NHibernate.Dialect.Function
 		{
 			return NHibernateUtil.Int64;
 		}
+
+		public override IType GetReturnType(IEnumerable<IType> argumentTypes, IMapping mapping, bool throwOnError)
+		{
+			// 6.0 TODO: return NHibernateUtil.Int64;
+#pragma warning disable 618
+			return ReturnType(argumentTypes.FirstOrDefault(), mapping);
+#pragma warning restore 618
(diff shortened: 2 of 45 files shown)


maca88:
Rebased.

src/NHibernate/Dialect/Function/ClassicCountFunction.cs
```diff
@@ -14,9 +14,11 @@ public ClassicCountFunction() : base("count", true)
 		{
 		}
 
+		// Since v5.3
+		[Obsolete("Use GetReturnType method instead.")]
```
fredericDelaporte, line 18: This PR still preserves `ClassicCountFunction` behavior but creates a risk of losing it.

I think we need to have some non-obsolete override already added in this class for ensuring dropping obsolete members will not result in losing the behavior of this class.

src/NHibernate/Dialect/Function/ClassicCountFunction.cs
```diff
@@ -14,9 +14,11 @@ public ClassicCountFunction() : base("count", true)
...
 		public override IType ReturnType(IType columnType, IMapping mapping)
 		{
 			return NHibernateUtil.Int32;
 		}
 	}
```
fredericDelaporte, line 23:
```suggestion

		public override IType GetReturnType(IEnumerable<IType> argumentTypes, IMapping mapping, bool throwOnError)
		{
			return NHibernateUtil.Int32;
		}
	}
```

src/NHibernate/Dialect/Function/CountQueryFunctionInfo.cs
```diff
@@ -9,9 +9,11 @@ class CountQueryFunctionInfo : ClassicAggregateFunction
 	{
 		public CountQueryFunctionInfo() : base("count", true) { }
 
+		// Since v5.3
+		[Obsolete("Use GetReturnType method instead.")]
```
fredericDelaporte, line 13: Same here.

src/NHibernate/Dialect/Function/CountQueryFunctionInfo.cs
```diff
@@ -9,9 +9,11 @@ class CountQueryFunctionInfo : ClassicAggregateFunction
...
 		public override IType ReturnType(IType columnType, IMapping mapping)
 		{
 			return NHibernateUtil.Int64;
 		}
 	}
```
fredericDelaporte, line 18:
```suggestion

		public override IType GetReturnType(IEnumerable<IType> argumentTypes, IMapping mapping, bool throwOnError)
		{
			return NHibernateUtil.Int64;
		}
	}
```

src/NHibernate/Dialect/SQLiteDialect.cs
```diff
@@ -490,8 +490,10 @@ protected class SQLiteCastFunction : CastFunction
 		{
 			protected override bool CastingIsRequired(string sqlType)
 			{
-				// SQLite doesn't support casting to datetime types.  It assumes you want an integer and destroys the date string.
-				if (StringHelper.ContainsCaseInsensitive(sqlType, "date") || StringHelper.ContainsCaseInsensitive(sqlType, "time"))
+				// SQLite doesn't support casting to datetime or uniqueidentifier types. It assumes you want an integer and destroys the date or uniqueidentifier string.
```
fredericDelaporte, line 493:
```suggestion
				// SQLite does not have types for datetime, date, time, and uniqueidentifier, they are to be handled as text. Trying to cast to them causes the SQLite lax type inference from unknown type to interpret the cast as casting to numeric, destroying the value. See https://www.sqlite.org/datatype3.html#determination_of_column_affinity.
```

bahusoid: > Trying to cast to them causes the SQLite lax type inference from unknown type to interpret the cast as casting to numeric, destroying the value

That's some tough sentence )  Original version seems easier to comprehend...

fredericDelaporte: The trouble for me is, the previous sentence is somewhat wrong. It somewhat implies SQLite has a datetime or uniqueidentifier type. But no, it has not.

It seems there are some misunderstanding in NHibernate about SQLite type system, with invalid types used in dialect, like:
https://github.com/nhibernate/nhibernate-core/blob/708ae88ab544523ab9fa8d277483a7131619d7b7/src/NHibernate/Dialect/SQLiteDialect.cs#L39-L67
This method is quite rubbish, since the only actual types handled by SQLite are `text`, `real`, `integer`, `blob` and `numeric`. Using any other type name does not cause any exception but causes SQLite to infer the type with some brittle logic:

> 1. If the declared type contains the string "INT" then it is assigned INTEGER affinity.
> 2. If the declared type of the column contains any of the strings "CHAR", "CLOB", or "TEXT" then that column has TEXT affinity. Notice that the type VARCHAR contains the string "CHAR" and is thus assigned TEXT affinity.
> 3. If the declared type for a column contains the string "BLOB" or if no type is specified then the column has affinity BLOB.
> 4. If the declared type for a column contains any of the strings "REAL", "FLOA", or "DOUB" then the column has REAL affinity.
> 5. Otherwise, the affinity is NUMERIC.

This means the last mappings, for date, time, bool, uniqueidentifier, all ends-up being treated as `numeric`. Which generally does not fail but is inefficient: SQLite will attempt to convert each value in such case, and falls back to storing them as text if not convertible.

Back to the comment: in fact I should even have written it this way:

> NHibernate dialect type bindings for SQLite are invalid and cause convert to have unexpected results when attempting to convert date, time and uniqueidentifier.

src/NHibernate/Dialect/Function/ClassicCountFunction.cs
maca88: Agree, I forgot to add it.

maca88:
Rebased and removed SqLite changes for cast, as it was fixed by #2346.

PR #2320, opened 2020-03-01
Support setting parameters with a dynamic object

### LINKED ISSUE #2009: Add support for IDictionary<string, object> to IQuery.SetProperties
Hello, everyone,

Recently I've been trying to use IQuery.SetProperties with ExpandoObject and, of course, failed.
Support for IDictionary<string, object> (and similar interfaces) would be benefical as it would also handle other use cases, such as use of Dictionary itself. Also it is relatively easy to implement.

Thanks.

fredericDelaporte:
Fixes #2009

### DIFF
--- a/src/NHibernate/Impl/AbstractQueryImpl.cs
+++ b/src/NHibernate/Impl/AbstractQueryImpl.cs
@@ -651,6 +651,8 @@ namespace NHibernate.Impl
 			return this;
 		}
 
+		// Since 5.3
+		[Obsolete("This method was never surfaced to a query interface. Use the overload or extension method taking a generic dictionary instead.")]
 		public IQuery SetProperties(IDictionary map)
 		{
 			string[] @params = NamedParameters;
@@ -669,6 +671,26 @@ namespace NHibernate.Impl
 				else
 				{
 					SetParameter(namedParam, obj, DetermineType(namedParam, obj.GetType()));
+				}
+			}
+			return this;
+		}
+
+		public IQuery SetParameters(IDictionary<string, object> map)
+		{
+			foreach (var namedParam in NamedParameters)
+			{
+				var obj = map[namedParam];
+				switch (obj)
+				{
+					case null:
+						break;
+					case IEnumerable enumerable when !(enumerable is string):
+						SetParameterList(namedParam, enumerable);
+						break;
+					default:
+						SetParameter(namedParam, obj, DetermineType(namedParam, obj.GetType()));
+						break;
 				}
 			}
 			return this;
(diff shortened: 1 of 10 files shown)


src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -651,6 +651,8 @@ public IQuery SetEnum(string name, Enum val)
 			return this;
 		}
 
+		// Since 5.3
+		[Obsolete("This method was never surfaced to a query interface. Use the overload or extension method taking a generic dictionary instead.")]
 		public IQuery SetProperties(IDictionary map)
```
fredericDelaporte: I have considered adding an extension method for it too, but this was causing an ambiguous method call when using a `Dictionary<string, object>`, since it does implement both `IDictionary` and `IDictionary<string, object>`. So it looks to me not worth keeping this method.

src/NHibernate/QueryExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+	public static class QueryExtensions
+	{
+		/// <summary>
+		/// Bind the keys values of the given dictionary to named parameters of the query,
+		/// matching keys with parameter names and mapping value types to
+		/// NHibernate types using heuristics.
+		/// </summary>
+		/// <param name="query">The query on which to set the parameters.</param>
+		/// <param name="map">A dictionary of parameters values by names.</param>
+		public static IQuery SetParameters(this IQuery query, IDictionary<string, object> map)
```
fredericDelaporte: I had to change the name, since non extension methods takes precedence. Moreover the name `SetProperties` does not make much sense in my view, this one is more semantically accurate I think.

So I have added a commit to obsolete `SetProperties` in favor of a `SetParameters` overload.

src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -674,7 +676,34 @@ public IQuery SetProperties(IDictionary map)
...
+						SetParameter(namedParam, obj, DetermineType(namedParam, obj.GetType()));
+						break;
+				}
+			}
+			return this;
+		}
+
+		// Since 5.3
+		[Obsolete("Use SetParameters method instead.")]
 		public IQuery SetProperties(object bean)
```
hazzik: What if we use this method to support the `IDictionary<string, object>`? 
```suggestion
		public IQuery SetProperties(object bean)
		{
			if (bean is IDictionary<string, object> map)
			{
			// TODO:
			}
		}
```

fredericDelaporte: It would be less discoverable, but it would support `dynamic` without having to do an explicit static call on some extension class.

PR #2321, opened 2020-03-03
Support setting parameters with a dynamic object

### LINKED ISSUE #2009: Add support for IDictionary<string, object> to IQuery.SetProperties

fredericDelaporte:
Fixes #2009

Alternate to #2320, as per @hazzik suggestion.

### DIFF
--- a/src/NHibernate.Test/Legacy/SQLLoaderTest.cs
+++ b/src/NHibernate.Test/Legacy/SQLLoaderTest.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Dynamic;
 using NHibernate.Dialect;
 using NHibernate.DomainModel;
 using NUnit.Framework;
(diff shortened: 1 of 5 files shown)


src/NHibernate.Test/Legacy/SQLLoaderTest.cs
```diff
@@ -140,6 +141,64 @@ public void FindBySQLProperties()
...
+				session.Flush();
+
+				var query =
+					session.CreateSQLQuery("select {category.*} from Category {category} where {category}.Name = :Name")
+					       .AddEntity("category", typeof(Category));
+				dynamic parameters = new ExpandoObject();
+				parameters.Name = s.Name;
+				// dynamic does not work on inherited interface method calls. https://stackoverflow.com/q/3071634
+				IQuery q = query;
+				q.SetProperties(parameters);
```
fredericDelaporte: A bit ugly, but there is not much we can do about it. Maybe "duplicate" the method in `ISqlQuery` in the next major release, but I do not think we should do such a thing to work around this `dynamic` short-coming.

src/NHibernate/Impl/AbstractQueryImpl.cs
```diff
@@ -674,8 +676,33 @@ public IQuery SetProperties(IDictionary map)
...
+		}
+
 		public IQuery SetProperties(object bean)
 		{
+			if (bean is IDictionary<string, object> map)
+			{
+				return SetParameters(map);
+			}
+
 			System.Type clazz = bean.GetType();
```
hazzik: Prob add call to non generic IDictionary as well

fredericDelaporte: Do you want me to add support for the non generic dictionary?

It could be done, but I do not think this will be actually useful for most users. I think only very old pieces of code may benefit of it, and well, we could leave it to the user to migrate or convert to more modern classes than the `HashTable` and the like.

I am on the side of dropping non-generic support whenever it may reduce the quantity of code we have, so I do not tend to add support for it.

I am wrong, is there still modern needs for non-generic support?

Maybe it could help those who use a generic dictionary with the "wrong" key/value types, but should we care for this case?

hazzik: IMO, generics do not make much sense when `T` is an `object`.

fredericDelaporte: They have less added value in such case, but still have quite some:
 - more consistent code: `Dictionary<keyType, valueType>` / `Dictionary<object, object>` vs `Dictionary<keyType, valueType>` / `HashTable`, `List<valueType>` / `List<object>`vs `List<valueType>` / `ArrayList`.
 - disposable `IEnumerator<valueType>` vs non-disposable `IEnumerator` (with some special handling in `for each` in order to still dispose them when the underlying concrete implementation is disposable).
 - non generic types APIs seem quite frozen (no non-generic set were added when the generic one was added, for example).
 - "object generic" collections can be fed to methods taking generic interfaces arguments, while non-generic collections require dedicated methods: using `IList<object>` / `IList<valueType>` can be handled by a single `DoSomething(IList<T>)`, while using `IList` / `IList<object>` would require an additional `DoSomething(IList)` overload.

The only added value of non-generic types vs "object generic" I see, it is more concise to declare non-generic. But their somewhat inconsistent naming nullifies this point in my view. And the potential for having to duplicate methods is a major pain.

fredericDelaporte: Marking as resolved since support for non generic `IDictionary` has been added, even though I am on the side of dropping non-generic support.

PR #2322, opened 2020-03-06
Add support for OData group by queries

### LINKED ISSUE #2135: OData GroupBy/Aggregate query not working in ASP.Net Web Api with Nhibernate
Versions:
Nhibernate.3.3.1.4000
Microsoft.AspNet.OData.7.0.1
.NetFramework.4.5.1

Querying a model (with Nhibernate mapping) with OData.

**Note: All other query options like $expand, $orderby, $filter, $select etc are working as expected except GroupBy and Aggregate**

Using OData GroupBy query as below :
https://localhost:65000/api/Model?$apply=groupby((userName))

Gives below exception:
```json
{
  "message": "An error has occurred.",
  "exceptionMessage": "MemberInitExpression",
  "exceptionType": "System.NotSupportedException",
  "stackTrace": "   at NHibernate.Linq.Visitors.HqlGeneratorExpressionTreeVisitor.VisitExpression(Expression expression)\r\n   at NHibernate.Linq.Visitors.ResultOperatorProcessors.ProcessGroupBy.<>c__DisplayClass1.<Process>b__0(Expression k)\r\n   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()\r\n   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)\r\n   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)\r\n   at NHibernate.Linq.Visitors.ResultOperatorProcessors.ProcessGroupBy.Process(GroupResultOperator resultOperator, QueryModelVisitor queryModelVisitor, IntermediateHqlTree tree)\r\n   at NHibernate.Linq.Visitors.ResultOperatorProcessors.ResultOperatorProcessor`1.Process(ResultOperatorBase resultOperator, QueryModelVisitor queryModel, IntermediateHqlTree tree)\r\n   at NHibernate.Linq.Visitors.ResultOperatorProcessors.ResultOperatorMap.Process(ResultOperatorBase resultOperator, QueryModelVisitor queryModel, IntermediateHqlTree tree)\r\n   at NHibernate.Linq.Visitors.QueryModelVisitor.VisitResultOperator(ResultOperatorBase resultOperator, QueryModel queryModel, Int32 index)\r\n   at Remotion.Linq.Clauses.ResultOperatorBase.Accept(IQueryModelVisitor visitor, QueryModel queryModel, Int32 index)\r\n   at Remotion.Linq.QueryModelVisitorBase.VisitResultOperators(ObservableCollection`1 resultOperators, QueryModel queryModel)\r\n   at Remotion.Linq.QueryModelVisitorBase.VisitQueryModel(QueryModel queryModel)\r\n   at NHibernate.Linq.Visitors.QueryModelVisitor.GenerateHqlQuery(QueryModel queryModel, VisitorParameters parameters, Boolean root)\r\n   at NHibernate.Linq.NhLinqExpression.Translate(ISessionFactoryImplementor sessionFactory)\r\n   at NHibernate.Hql.Ast.ANTLR.ASTQueryTranslatorFactory.CreateQueryTranslators(String queryIdentifier, IQueryExpression queryExpression, String collectionRole, Boolean shallow, IDictionary`2 filters, ISessionFactoryImplementor factory)\r\n   at NHibernate.Engine.Query.QueryPlanCache.GetHQLQueryPlan(IQueryExpression queryExpression, Boolean shallow, IDictionary`2 enabledFilters)\r\n   at NHibernate.Impl.AbstractSessionImpl.GetHQLQueryPlan(IQueryExpression queryExpression, Boolean shallow)\r\n   at NHibernate.Impl.AbstractSessionImpl.CreateQuery(IQueryExpression queryExpression)\r\n   at NHibernate.Linq.DefaultQueryProvider.PrepareQuery(Expression expression, IQuery& query, NhLinqExpression& nhQuery)\r\n   at NHibernate.Linq.DefaultQueryProvider.Execute(Expression expression)\r\n   at NHibernate.Linq.DefaultQueryProvider.Execute[TResult](Expression expression)\r\n   at Remotion.Linq.QueryableBase`1.System.Collections.IEnumerable.GetEnumerator()\r\n   at System.Linq.Enumerable.<OfTypeIterator>d__95`1.MoveNext()\r\n   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)\r\n   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)\r\n   at MatrixGroup.Bmx.Server.Filters.UnitOfWorkFilterAttribute.OnActionExecuted(HttpActionExecutedContext actionExecutedContext)\r\n   at System.Web.Http.Filters.ActionFilterAttribute.OnActionExecutedAsync(HttpActionExecutedContext actionExecutedContext, CancellationToken cancellationToken)\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Filters.ActionFilterAttribute.<CallOnActionExecutedAsync>d__5.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Filters.ActionFilterAttribute.<ExecuteActionFilterAsyncCore>d__0.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Controllers.ActionFilterResult.<ExecuteAsync>d__2.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Filters.AuthorizationFilterAttribute.<ExecuteAuthorizationFilterAsyncCore>d__2.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Controllers.ExceptionFilterResult.<ExecuteAsync>d__0.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Web.Http.Controllers.ExceptionFilterResult.<ExecuteAsync>d__0.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Dispatcher.HttpControllerDispatcher.<SendAsync>d__1.MoveNext()"
}
```

Using OData Aggregate query as below :
https://localhost:65000/api/Model?$apply=aggregate(Rank%20with%20max%20as%20MaxRank)

Gives below exception:
```json
{
  "message": "An error has occurred.",
  "exceptionMessage": "Could not determine a type for class: Microsoft.AspNet.OData.Query.Expressions.NoGroupByWrapper, Microsoft.AspNet.OData, Version=7.0.1.20718, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
  "exceptionType": "NHibernate.HibernateException",
  "stackTrace": "   at NHibernate.Impl.AbstractQueryImpl.GuessType(Type clazz)\r\n   at NHibernate.Impl.AbstractQueryImpl.GuessType(Object param)\r\n   at NHibernate.Impl.AbstractQueryImpl.DetermineType(String paramName, Object paramValue)\r\n   at NHibernate.Impl.AbstractQueryImpl.SetParameter(String name, Object val)\r\n   at NHibernate.Linq.DefaultQueryProvider.SetParameters(IQuery query, IDictionary`2 parameters)\r\n   at NHibernate.Linq.DefaultQueryProvider.PrepareQuery(Expression expression, IQuery& query, NhLinqExpression& nhQuery)\r\n   at NHibernate.Linq.DefaultQueryProvider.Execute(Expression expression)\r\n   at NHibernate.Linq.DefaultQueryProvider.Execute[TResult](Expression expression)\r\n   at Remotion.Linq.QueryableBase`1.System.Collections.IEnumerable.GetEnumerator()\r\n   at System.Linq.Enumerable.<OfTypeIterator>d__95`1.MoveNext()\r\n   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)\r\n   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)\r\n   at MatrixGroup.Bmx.Server.Filters.UnitOfWorkFilterAttribute.OnActionExecuted(HttpActionExecutedContext actionExecutedContext)\r\n   at System.Web.Http.Filters.ActionFilterAttribute.OnActionExecutedAsync(HttpActionExecutedContext actionExecutedContext, CancellationToken cancellationToken)\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Filters.ActionFilterAttribute.<CallOnActionExecutedAsync>d__5.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Filters.ActionFilterAttribute.<ExecuteActionFilterAsyncCore>d__0.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Controllers.ActionFilterResult.<ExecuteAsync>d__2.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Filters.AuthorizationFilterAttribute.<ExecuteAuthorizationFilterAsyncCore>d__2.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Controllers.ExceptionFilterResult.<ExecuteAsync>d__0.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Web.Http.Controllers.ExceptionFilterResult.<ExecuteAsync>d__0.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.Dispatcher.HttpControllerDispatcher.<SendAsync>d__1.MoveNext()"
```

### LINKED ISSUE #2221: GroupBySelectClauseRewriter NotImplementedException
Execute this query gives the following exception:

```csharp
public class Company
{
    public int Id { get; set; }
    public string Description { get; set; }
}

public class CompanyDto
{
    public int Id { get; set; }
    public string Description { get; set; }
}

var query = session.Query<Company>()
    .Select(x =>
        new CompanyDto()
	{
	    Id = x.Id,
	    Description = x.Description,
	})
    .GroupBy(x => x.Id)
    .Select(x =>
        new CompanyDto()
	{
	    Id = x.Key,
	    Description = x.Max(y => y.Description),
	});

query.ToList()
```

```
The method or operation is not implemented
[StackTrace]
in NHibernate.Linq.GroupBy.GroupBySelectClauseRewriter.VisitMember(MemberExpression expression)
in NHibernate.Linq.Expressions.NhAggregatedExpression.VisitChildren(ExpressionVisitor visitor)
in NHibernate.Linq.Expressions.NhExpression.Accept(ExpressionVisitor visitor)
in NHibernate.Linq.GroupBy.GroupBySelectClauseRewriter.ReWrite(Expression expression, GroupResultOperator groupBy, QueryModel model)
in NHibernate.Linq.GroupBy.GroupBySelectClauseRewriter.VisitSubQuery(SubQueryExpression expression)
in System.Linq.Expressions.ExpressionVisitor.VisitMemberAssignment(MemberAssignment node)
in System.Linq.Expressions.ExpressionVisitor.Visit[T](ReadOnlyCollection`1 nodes, Func`2 elementVisitor)
in System.Linq.Expressions.ExpressionVisitor.VisitMemberInit(MemberInitExpression node)
in NHibernate.Linq.GroupBy.GroupBySelectClauseRewriter.ReWrite(Expression expression, GroupResultOperator groupBy, QueryModel model)
in NHibernate.Linq.GroupBy.AggregatingGroupByRewriter.<>c__DisplayClass2_0.<FlattenSubQuery>b__0(Expression s)
in Remotion.Linq.Clauses.SelectClause.TransformExpressions(Func`2 transformation)
in NHibernate.Linq.GroupBy.AggregatingGroupByRewriter.FlattenSubQuery(QueryModel queryModel, QueryModel subQueryModel, GroupResultOperator groupBy)
in NHibernate.Linq.GroupBy.AggregatingGroupByRewriter.ReWrite(QueryModel queryModel)
in NHibernate.Linq.Visitors.QueryModelVisitor.GenerateHqlQuery(QueryModel queryModel, VisitorParameters parameters, Boolean root, Nullable`1 rootReturnType)
in NHibernate.Linq.NhLinqExpression.Translate(ISessionFactoryImplementor sessionFactory, Boolean filter)
in NHibernate.Hql.Ast.ANTLR.ASTQueryTranslatorFactory.CreateQueryTranslators(IQueryExpression queryExpression, String collectionRole, Boolean shallow, IDictionary`2 filters, ISessionFactoryImplementor factory)
in NHibernate.Engine.Query.QueryExpressionPlan.CreateTranslators(IQueryExpression queryExpression, String collectionRole, Boolean shallow, IDictionary`2 enabledFilters, ISessionFactoryImplementor factory)
in NHibernate.Engine.Query.QueryExpressionPlan..ctor(IQueryExpression queryExpression, Boolean shallow, IDictionary`2 enabledFilters, ISessionFactoryImplementor factory)
in NHibernate.Engine.Query.QueryPlanCache.GetHQLQueryPlan(IQueryExpression queryExpression, Boolean shallow, IDictionary`2 enabledFilters)
in NHibernate.Impl.AbstractSessionImpl.GetHQLQueryPlan(IQueryExpression queryExpression, Boolean shallow)
in NHibernate.Impl.AbstractSessionImpl.CreateQuery(IQueryExpression queryExpression)
in NHibernate.Linq.DefaultQueryProvider.PrepareQuery(Expression expression, IQuery& query)
in NHibernate.Linq.DefaultQueryProvider.Execute(Expression expression)
in NHibernate.Linq.DefaultQueryProvider.Execute[TResult](Expression expression)
in Remotion.Linq.QueryableBase`1.GetEnumerator()
in System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
in System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
```

maca88:
Fixes #2221
Fixes part of #2334
Fixes #2135


src/NHibernate/Linq/GroupBy/GroupBySelectClauseRewriter.cs
```diff
@@ -67,8 +75,8 @@ protected override Expression VisitMember(MemberExpression expression)
 				return base.VisitMember(expression);
 			}
 
-			if ((elementSelector is NewExpression || elementSelector.NodeType == ExpressionType.Convert)
```
maca88: Instead of checking whether it is a `ExpressionType.Convert` expression, its operand is checked in order to verify whether it is actually supported. The `ExpressionType.Convert` was used for `ExpressionType.ArrayIndex` expression which was added by #433.

maca88:
The last commit resolves the issue with `group by` reported in #2334.

src/NHibernate/Linq/Visitors/TransparentIdentifierRemovingExpressionVisitor.cs
```diff
@@ -0,0 +1,100 @@
...
+			if (memberBindings == null)
+				return base.VisitMember(memberExpression);
+
+			var matchingAssignment = memberBindings
+			                         .Where(binding => binding.MatchesReadAccess(memberExpression.Member))
+			                         .LastOrDefault();
+
+			// Added logic: In some cases (e.g OData), the member can be from a different derived class, in such case
+			// we need to check the member DeclaringType instead of ReflectedType
+			if (matchingAssignment == null && memberExpression.Expression.NodeType == ExpressionType.MemberInit)
```
maca88: The only way to test this was to include and use `Microsoft.AspNetCore.OData` package as this cannot be reproduced by creating classes in code.

src/NHibernate/Linq/Visitors/TransparentIdentifierRemovingExpressionVisitor.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using Remotion.Linq;
+using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Parsing;
+using MemberBinding = Remotion.Linq.Parsing.ExpressionVisitors.MemberBindings.MemberBinding;
+
+namespace NHibernate.Linq.Visitors
+{
+	// Copied from Relinq and added a fallback for comparing two member info by DeclaringType and Name
```
hazzik: Need license and copyright at the top/bottom

maca88: Added.

src/NHibernate/Linq/Visitors/TransparentIdentifierRemovingExpressionVisitor.cs
```diff
@@ -0,0 +1,117 @@
...
+			var memberBindings = GetMemberBindingsCreatedByExpression(memberExpression.Expression);
+			if (memberBindings == null)
+				return base.VisitMember(memberExpression);
+
+			var matchingAssignment = memberBindings
+			                         .Where(binding => binding.MatchesReadAccess(memberExpression.Member))
+			                         .LastOrDefault();
+
+			// Added logic: In some cases (e.g OData), the member can be from a different derived class, in such case
+			// we need to check the member DeclaringType instead of ReflectedType
```
hazzik: It's all good for now, but probably one of 2 things should happen:

1. This fixed in OData provider so it generates the correct expression.
2. This fixed in ReLinq so we do not copy code from there and just use fixed version.

hazzik:
@maca88 what about #2135?

src/NHibernate/Linq/GroupResultOperatorExtensions.cs
```diff
@@ -19,12 +19,31 @@ private static IEnumerable<Expression> ExtractKeyExpressions(this Expression exp
 		{
 			// Recursively extract key expressions from nested initializers 
 			// --> new object[] { ((object)new object[] { x.A, x.B }), x.C }
-			// --> x.A, x.B, x.C
-			if (expr is NewExpression)
-				return (expr as NewExpression).Arguments.SelectMany(ExtractKeyExpressions);
-			if (expr is NewArrayExpression)
-				return (expr as NewArrayExpression).Expressions.SelectMany(ExtractKeyExpressions);
+			// --> new List<string> { x.A, x.B }
+			// --> new CustomType(x.A, x.B) { C = x.C }
```
hazzik: Not sure if this is correct. There are examples when this assumption is not working.

maca88:
> @maca88 what about #2135?

This PR solves it as well, added it in the description.

src/NHibernate/Linq/Visitors/TransparentIdentifierRemovingExpressionVisitor.cs
maca88: > but probably one of 2 things should happen:

I do agree and [created an issue](https://github.com/OData/WebApi/issues/2108) for ODAta, which I thnik is where the issue should be fixed.

src/NHibernate/Linq/GroupResultOperatorExtensions.cs
maca88: It would be incorrect for Linq to Objects, but in our scenario the group by is done on the database, so the constructor of `CustomType2` will never be called. In our case the following expressions will produce the same result:
```C#
GroupBy(x => new List<string> { x.A, x.B, x.C })
GroupBy(x => new CustomType(x.A, x.B) { C = x.C })
GroupBy(x => new { x.A x.B, x.C })
```
The idea of this method is to collect all mapped properties inside the expression and use them as keys for grouping.

hazzik: I think this might be restricted to some known types (potentially a configurable option) and some heuristics (like looking for [CompilerGeneratedAttribute]).

bahusoid: > so the constructor of CustomType2 will never be called. 

What about case when group key is selected?
```C#
GroupBy(x => new CustomType(x.A, x.B) { C = x.C }).Select(x => x.Key);
```
In any case I think it's better keep the code simpler and allow it without any added heuristics (and I've seen multiple requests specifically to support ctors for user DTOs for whatever legacy reasons).

> There are examples when this assumption is not working.

Well it applies for all cases with user types - user can override equality methods or getters with some strange logic and that's it.

fredericDelaporte: > There are examples when this assumption is not working.

This concern already holds true for current master state. I do not think we should consider this as added or worsen by this PR. So I think it should be addressed in another PR, if it actually needs to be addressed.

src/NHibernate/Linq/Visitors/TransparentIdentifierRemovingExpressionVisitor.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using Remotion.Linq;
+using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Parsing;
+using MemberBinding = Remotion.Linq.Parsing.ExpressionVisitors.MemberBindings.MemberBinding;
+
+namespace NHibernate.Linq.Visitors
+{
+	// Copied from Relinq and added a fallback for comparing two member info by DeclaringType and Name
```
fredericDelaporte, line 29:
```suggestion
	// Copied from Relinq and added a fallback for comparing two member info by DeclaringType and Name
	// 6.0 TODO: drop if https://github.com/OData/WebApi/issues/2108 is fixed and add a possible breaking
	// change requiring to upgrade OData. (See https://github.com/nhibernate/nhibernate-core/pull/2322#discussion_r401215456 )
```

src/NHibernate/Linq/GroupResultOperatorExtensions.cs
maca88: > I think this might be restricted to some known types (potentially a configurable option) and some heuristics (like looking for [CompilerGeneratedAttribute])

What would happen when the restriction is not met? For example the following query:
```C#
_session.Query<X>().GroupBy(x => new CustomType(x.A, x.B) { C = x.C }).ToList();
```
if the restriction is not met we could either throw an exception or execute `_session.Query<X>()` and do the rest on the client, which I don't thnik we should do. 

> like looking for [CompilerGeneratedAttribute]

In OData, the [wrappers that they use](https://github.com/OData/WebApi/blob/279f60dabac9c6de2bb8f560886ffa499940bc89/src/Microsoft.AspNet.OData.Shared/Query/Expressions/DynamicTypeWrapper.cs#L112) have only `JsonConverter` attribute which I don't think we should check. The only restriction that we may add, is to check whether the type overriders `Equals` and `GetHashCode` methods and throw when it does. But even that restriction does not make much sense as grouping in Linq To Objects by a type that does not override those methods won't group any items together, where in our case it may group from zero to all items together. In my opinion we should by default allow all types and optionally having a configurable option that would throw when the given restriction is not met.

> What about case when group by key is selected?

In that case the constructor will be called for each group. The following query produces the same result:
```C#
GroupBy(x => new {x.A, x.B, x.C }).Select(x => new CustomType(x.Key.A, x.Key.B) { C = x.Key.C }); 
```
When the user wants to return all groups as a list of `CustomType` then IMO your query is better as it is shorter to write.

bahusoid: > to check whether the type overrides Equals and GetHashCode methods and throw when it **does**

I don't get it.  Wouldn't it throw for all anonymous types? Did you mean "and throw when it **doesn't"**? If I got it right the main concern was about difference between Linq to Objects and NHibernate results. And from this context user type must override equality members to have the same results.
But there is so many ways that user can break the difference in results so I wouldn't implement any additional restrictions at all.

fredericDelaporte:
Either we squash it all for merging, or the last three commits should be squashed together for then merging the two remaining commits.

bahusoid:
> Either we squash it all for merging, or the last three commits should be squashed together for then merging the two remaining commits.

I think changes that would be no longer necessary after https://github.com/OData/WebApi/issues/2108 is fixed should go as separate commit. And I think we should also add TODO in code with link to opened ODATA issue and suggestion to drop this code when it's fixed (if it's really only a hack for ODATA)

fredericDelaporte:
> Anbd I think we should also add TODO in code with link to opened ODATA issue and suggestion to drop this code when it's fixed (if it's really only a hack for ODATA)

Like 90b3316?...

bahusoid:
Oh.. Somehow missed your commit. So yes I'm for your suggestion with separate commits.

src/NHibernate/Linq/GroupResultOperatorExtensions.cs
maca88: > I don't get it. Why? 

Whether we would want to prevent the user from thinking that the `Equals` and `GetHashCode` would be used for grouping.

> Did you mean and throw when it **doesn't"**?

No, I ment when it does override those methods it would throw.

> But there is so many ways that user can break the difference in results so I wouldn't implement any additional restrictions at all.

True, that is why I also agree that we should not add any restriction by default.

Also, I have to correct myself as in the following case the constructor will also be called:
```C#
_session.Query<X>().GroupBy(x => new CustomType(x.A, x.B) { C = x.C }).Select(o => new { o.Key.A, o.Key.B, o.Key.C }).ToList();
```
beacuse there is no explicit link in the expression between `o.Key.A\o.Key.B` and `x.A\x.B`, due to the fact that `x.A` and `x.B` are passed to the constructor.
Which means that it will work correctly even when `x.A` is set in the constructor to property `CustomType.B`.

maca88:
Squashed into two commits and preserved @fredericDelaporte as a co-author. I didn't tag #2334 in the commit as the issue is also about selection which is not fixed by this PR.

fredericDelaporte:
So I have updated the description for avoiding closing #2334 when merging this PR.

PR #2324, opened 2020-03-06
Update IIsEntityDecider to use ExpressionsHelper.TryGetMappedType

maca88:
Extracted from #2316 and will allow the following query to work:
```C#
db.Query<Animal>().Select(o => o.FatherOrMother.SerialNumber).ToList();

// Not mapped property
public Animal FatherOrMother => Father ?? Mother;
```
The current logic detects `FatherOrMother` as a mapped property and a join is added, which prevents #2316 from support the above query. By using `ExpressionsHelper.TryGetMappedType` prevents `FatherOrMother` to be detected as a mapped property.

### DIFF
--- a/src/NHibernate/Linq/Visitors/MemberExpressionJoinDetector.cs
+++ b/src/NHibernate/Linq/Visitors/MemberExpressionJoinDetector.cs
@@ -19,21 +19,16 @@ namespace NHibernate.Linq.Visitors
 	{
 		private readonly IIsEntityDecider _isEntityDecider;
 		private readonly IJoiner _joiner;
-		private readonly ISessionFactoryImplementor _sessionFactory;
 
 		private bool _requiresJoinForNonIdentifier;
 		private bool _preventJoinsInConditionalTest;
 		private bool _hasIdentifier;
 		private int _memberExpressionDepth;
 
-		public MemberExpressionJoinDetector(
-			IIsEntityDecider isEntityDecider,
-			IJoiner joiner,
-			ISessionFactoryImplementor sessionFactory)
+		public MemberExpressionJoinDetector(IIsEntityDecider isEntityDecider, IJoiner joiner)
 		{
 			_isEntityDecider = isEntityDecider;
 			_joiner = joiner;
-			_sessionFactory = sessionFactory;
 		}
 
 		protected override Expression VisitMember(MemberExpression expression)
--- a/src/NHibernate/Linq/Visitors/MemberExpressionJoinDetector.cs
+++ b/src/NHibernate/Linq/Visitors/MemberExpressionJoinDetector.cs
@@ -1,6 +1,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq.Expressions;
+using NHibernate.Engine;
 using NHibernate.Linq.Expressions;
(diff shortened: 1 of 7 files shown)


src/NHibernate/Linq/Visitors/MemberExpressionJoinDetector.cs
```diff
@@ -18,21 +19,26 @@ internal class MemberExpressionJoinDetector : RelinqExpressionVisitor
...
 		private bool _requiresJoinForNonIdentifier;
 		private bool _preventJoinsInConditionalTest;
 		private bool _hasIdentifier;
 		private int _memberExpressionDepth;
 
-		public MemberExpressionJoinDetector(IIsEntityDecider isEntityDecider, IJoiner joiner)
+		public MemberExpressionJoinDetector(
+			IIsEntityDecider isEntityDecider,
+			IJoiner joiner,
+			ISessionFactoryImplementor sessionFactory)
```
fredericDelaporte, line 32: This addition seems to have no use, is this some forgotten "trash change"?

src/NHibernate/Linq/Visitors/WhereJoinDetector.cs
```diff
@@ -71,10 +73,14 @@ internal class WhereJoinDetector : RelinqExpressionVisitor
 		// The following is used for member expressions traversal.
 		private int _memberExpressionDepth;
 
-		internal WhereJoinDetector(IIsEntityDecider isEntityDecider, IJoiner joiner)
+		internal WhereJoinDetector(
+			IIsEntityDecider isEntityDecider,
+			IJoiner joiner,
+			ISessionFactoryImplementor sessionFactory)
```
fredericDelaporte, line 79: Same here.

src/NHibernate/Linq/Visitors/MemberExpressionJoinDetector.cs
maca88: My bad, `sessionFactory` is used in #2079 for `ExpressionKeyVisitor.Visit` method which is not included in this PR. Removed the unused parameter and field.

maca88:
> Can you add a test case for the kind of query this change allow support?

By just updating `IIsEntityDecider` the mentioned query in the first post does not execute, because `SelectClauseHqlNominator` will still nominate the not mapped property as evaluatable. With the last commit I added a simple fix for `SelectClauseHqlNominator` issue and a test case.

src/NHibernate/Linq/Visitors/SelectClauseNominator.cs
```diff
@@ -168,8 +171,10 @@ private bool CanBeEvaluatedInHqlSelectStatement(Expression expression, bool proj
 				return projectConstantsInHql;
 			}
 
-			// Assume all is good
-			return true;
+			return !(expression is MemberExpression memberExpression) || // Assume all is good
+			       // Evaluate only expressions that represent a mapped property
```
fredericDelaporte, line 175: Shouldn't it be "Nominate" rather than "Evaluate"?
```suggestion
			       // Nominate only expressions that represent a mapped property or a translatable method call
```

maca88: Yes, it should.

PR #2327, opened 2020-03-11
Add cross join support for Hql and Linq query provider

### LINKED ISSUE #1060: NH-1014 - NHibernate Cross Join Syntax Causes Issues With SQL Server 2000/2005
**John Chapman** created an [issue](https://nhibernate.jira.com/browse/NH-1014) — <time datetime="2007-05-17T10:20:15.639Z" title="2007-05-17T10:20:15.639Z">17th May 2007, 10:20:15</time>:
> See forum post: http://forum.hibernate.org/viewtopic.php?t=974101&highlight=
> 
> When using a cross join in SQL Server 2000/2005 with the comma syntax SQL Server 2005 requires that all inner joins (inner join syntax) occur directly below the table being joined to.  If a comma join occurs between the table and an inner join sql server will provide a "unable to bind multi-part alias..." error.
> 
> Changing the dialect to provide ANSI compliant " CROSS JOIN " instead of " , " would resolve this issue.
> 
> Patch to come soon...
> 
---
**John Chapman** added a comment — <time datetime="2007-05-17T17:09:35.374Z" title="2007-05-17T17:09:35.374Z">17th May 2007, 17:09:35</time>:
> zip file includes a patch for the ANSIJoinFragment which should resolve this issue.
> 
---
**justme84** added a comment — <time datetime="2007-07-21T04:16:01.624Z" title="2007-07-21T04:16:01.624Z">21st July 2007, 4:16:01</time>:
> I'm not sure if this change won't break support for other databases (Google suggests DB2 may not support cross join syntax).
> 
---
**justme84** added a comment — <time datetime="2007-07-21T05:41:47.405Z" title="2007-07-21T05:41:47.405Z">21st July 2007, 5:41:47</time>:
> forgot to add - the way to go would be to create a special JoinFragment/ANSIJoinFragment subclass and override Dialect.CreateOuterJoinFragment() in MsSql2000Dialect to return the subclass. If you change your patch accordingly, I will apply it.
> 
---
**John Chapman** added a comment — <time datetime="2007-07-25T18:47:11.561Z" title="2007-07-25T18:47:11.561Z">25th July 2007, 18:47:11</time>:
> I attached a file containing the patch as you recommended.  Note that I don't really like what I did since I changed the buffer to be protected on the ANSIJoinFragment.  I wanted to preserve the existing ANSI functionality, but that required me to make the buffer protected to make it available to the new join fragment subclass.  I don't believe the existing naming strategy matches the .NET standards since protected is technically aware to other assemblies which happen to inherit from that class.
> 
> I can modify the code if you would like, or we could leave it if you prefer.  I also had to create a new virtual method to create an instance of itself.  This is to support the Copy method so that Copy would create the correct subclass.  I could have made Copy virtual and duplicated the copy code, but preferred to leave the copy code alone and make a new virtual method to create the instance of itself.  This could also be changed if you wish.
> 
---
**R. Muti** added a comment — <time datetime="2007-11-03T23:51:59.687Z" title="2007-11-03T23:51:59.687Z">3rd November 2007, 23:51:59</time>:
> This is still an issue with HQL, which appears to use QueryJoinFragment instead of ANSIJoinFragment.
> 
> I tried to make a similar change to QueryJoinFragment (I just replaced ", " with " cross join "), but for some reason it translated the HQL "from ObjectA a, ObjectB" into the SQL "from cross join TableA a cross join TableB b".  I'm not very familiar with the source code, so I'm not sure why this is happening.
> 
> 
---
**Fabio Maulo** added a comment — <time datetime="2009-06-15T20:15:06.241Z" title="2009-06-15T20:15:06.241Z">15th June 2009, 20:15:06</time>:
> We can't apply the patch without some specific tests.
>

maca88:
Fixes: #1128
Closes #1060

Hql example:
```Sql
select s from EntityComplex s cross join EntityComplex q where s.SameTypeChild.Id = q.SameTypeChild.Id
```

Linq example:
```C#
from o in db.Orders
from o2 in db.Orders
where (o.OrderId == o2.OrderId + 1) || (o.OrderId == o2.OrderId - 1)
select new { o.OrderId, OrderId2 = o2.OrderId }
```
Linq query provider will use cross join when using multiple `from` statements and will fallback to an inner join with `on 1=1`, when the dialect does not support it.

 Possible breaking change:
- Custom dialects used for databases that do not support cross join, will have to override `SupportsCrossJoin` property and set it to `false`.

### DIFF
--- a/src/NHibernate.Test/Linq/LinqQuerySamples.cs
+++ b/src/NHibernate.Test/Linq/LinqQuerySamples.cs
@@ -1570,7 +1570,9 @@ namespace NHibernate.Test.Linq
 
 				var sql = sqlSpy.GetWholeLog();
 				Assert.That(sql, Does.Contain(useCrossJoin ? "cross join" : "inner join"));
-				Assert.That(GetTotalOccurrences(sql, "left outer join"), Is.EqualTo(1));
+				Assert.That(GetTotalOccurrences(sql, "left outer join"), Is.EqualTo(0));
+				Assert.That(GetTotalOccurrences(sql, "inner join"), Is.EqualTo(useCrossJoin ? 1 : 2));
+				Assert.That(GetTotalOccurrences(sql, "cross join"), Is.EqualTo(useCrossJoin ? 1 : 0));
 			}
 		}
 
--- a/src/NHibernate.Test/Linq/LinqQuerySamples.cs
+++ b/src/NHibernate.Test/Linq/LinqQuerySamples.cs
@@ -1,7 +1,9 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using NHibernate.DomainModel.Northwind.Entities;
+using NSubstitute;
 using NUnit.Framework;
 
 namespace NHibernate.Test.Linq
@@ -1301,7 +1303,13 @@ namespace NHibernate.Test.Linq
 				where c.Address.City == "London"
 				select o;
 
-			ObjectDumper.Write(q);
+			using (var sqlSpy = new SqlLogSpy())
+			{
+				ObjectDumper.Write(q);
+
+				var sql = sqlSpy.GetWholeLog();
+				Assert.That(GetTotalOccurrences(sql, "inner join"), Is.EqualTo(1));
+			}
 		}
 
 		[Category("JOIN")]
@@ -1407,7 +1415,13 @@ namespace NHibernate.Test.Linq
 				where p.Supplier.Address.Country == "USA" && p.UnitsInStock == 0
 				select p;
 
-			ObjectDumper.Write(q);
+			using (var sqlSpy = new SqlLogSpy())
+			{
+				ObjectDumper.Write(q);
+
+				var sql = sqlSpy.GetWholeLog();
+				Assert.That(GetTotalOccurrences(sql, "inner join"), Is.EqualTo(1));
+			}
 		}
 
 		[Category("JOIN")]
@@ -1423,7 +1437,16 @@ namespace NHibernate.Test.Linq
 				where e.Address.City == "Seattle"
 				select new {e.FirstName, e.LastName, et.Region.Description};
 
-			ObjectDumper.Write(q);
+			using (var sqlSpy = new SqlLogSpy())
+			{
+				ObjectDumper.Write(q);
+
+				var sql = sqlSpy.GetWholeLog();
+				// EmployeeTerritories and Territories
+				Assert.That(GetTotalOccurrences(sql, "inner join"), Is.EqualTo(2));
+				// Region
+				Assert.That(GetTotalOccurrences(sql, "left outer join"), Is.EqualTo(1));
+			}
 		}
 
 		[Category("JOIN")]
@@ -1447,7 +1470,13 @@ namespace NHibernate.Test.Linq
 							   e1.Address.City
 						   };
 
-			ObjectDumper.Write(q);
+			using (var sqlSpy = new SqlLogSpy())
+			{
+				ObjectDumper.Write(q);
+
+				var sql = sqlSpy.GetWholeLog();
+				Assert.That(GetTotalOccurrences(sql, "inner join"), Is.EqualTo(1));
+			}
 		}
 
 		[Category("JOIN")]
@@ -1462,7 +1491,13 @@ namespace NHibernate.Test.Linq
 				join o in db.Orders on c.CustomerId equals o.Customer.CustomerId into orders
 				select new {c.ContactName, OrderCount = orders.Average(x => x.Freight)};
 
-			ObjectDumper.Write(q);
+			using (var sqlSpy = new SqlLogSpy())
+			{
+				ObjectDumper.Write(q);
+
+				var sql = sqlSpy.GetWholeLog();
+				Assert.That(GetTotalOccurrences(sql, "join"), Is.EqualTo(0));
+			}
 		}
 
 		[Category("JOIN")]
@@ -1503,15 +1538,32 @@ namespace NHibernate.Test.Linq
 		}
 
 		[Category("JOIN")]
-		[Test(Description = "This sample explictly joins two tables with a composite key and projects results from both tables.")]
-		public void DLinqJoin5d()
+		[TestCase(true, Description = "This sample explictly joins two tables with a composite key and projects results from both tables.")]
+		[TestCase(false, Description = "This sample explictly joins two tables with a composite key and projects results from both tables.")]
+		public void DLinqJoin5d(bool useCrossJoin)
 		{
+			if (useCrossJoin && !Dialect.SupportsCrossJoin)
+			{
+				Assert.Ignore("Dialect does not support cross join.");
+			}
+
 			var q =
 				from c in db.Customers
 				join o in db.Orders on new {c.CustomerId, HasContractTitle = c.ContactTitle != null} equals new {o.Customer.CustomerId, HasContractTitle = o.Customer.ContactTitle != null }
 				select new { c.ContactName, o.OrderId };
 
-			ObjectDumper.Write(q);
+			using (var substitute = SubstituteDialect())
+			using (var sqlSpy = new SqlLogSpy())
+			{
+				ClearQueryPlanCache();
+				substitute.Value.SupportsCrossJoin.Returns(useCrossJoin);
+
+				ObjectDumper.Write(q);
+
+				var sql = sqlSpy.GetWholeLog();
+				Assert.That(sql, Does.Contain(useCrossJoin ? "cross join" : "inner join"));
+				Assert.That(GetTotalOccurrences(sql, "left outer join"), Is.EqualTo(1));
+			}
 		}
 
 		[Category("JOIN")]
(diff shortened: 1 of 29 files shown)


src/NHibernate/Dialect/Dialect.cs
```diff
@@ -1337,6 +1337,11 @@ public virtual JoinFragment CreateOuterJoinFragment()
 			return new ANSIJoinFragment();
 		}
 
+		/// <summary>
+		/// Does this dialect support CROSS JOIN?
+		/// </summary>
+		public virtual bool SupportsCrossJoin => true;
```
bahusoid: Shouldn't it be false by default? Otherwise it can be breaking change for some user custom dialects

maca88: I took the same approach Hibernate did with [getCrossJoinSeparator](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java#L2368) which enables cross join by default. If you think that is better to have it disabled by default I will do it, otherwise we can add it as a possible breaking change.

bahusoid: I'm fine with 
>  add it as a possible breaking change.

bahusoid:
What about  joins generated for associations used in expressions (something like `where o.Association.PropertyId = 1`). If I remember correctly they are also generated as implicit joins. Shouldn't we also use cross joins instead  in Linq queries?

maca88:
> Is this case tested by any CI configurations?

No, I will try to update the tests to test both scenarios.

> Maybe we should simulate cross join as inner join with bogus condition like 1 = 1

Why not, I will add this fallback.

> What about joins generated for associations used in expressions (something like where o.Association.PropertyId == 1). If I remember correctly they are also generated as implicit joins. Shouldn't we also use cross joins instead in Linq queries?

They are still generated with an implicit join. I tried to use cross join for such scenarios but the code is quite messy and it requires a lot more work as I expected. There are preconditions that such joins always starts with a comma, for example:
https://github.com/nhibernate/nhibernate-core/blob/bf1e11647e63d9220a369b224f4a2a1545d439d2/src/NHibernate/Hql/Ast/ANTLR/Util/JoinProcessor.cs#L152-L155

https://github.com/nhibernate/nhibernate-core/blob/ac39173567b31bcfad475ca32687c9faf0d37f87/src/NHibernate/SqlCommand/QuerySelect.cs#L217-L225

https://github.com/nhibernate/nhibernate-core/blob/5417dbc34cfb3e4377dd489a956a0702a6f24092/src/NHibernate/Hql/CollectionSubqueryFactory.cs#L24

which are also the root issue for the problem described in #1060:

> I tried to make a similar change to QueryJoinFragment (I just replaced ", " with " cross join "), but for some reason it translated the HQL "from ObjectA a, ObjectB" into the SQL "from cross join TableA a cross join TableB b"

As this is not trivial to change, I decided to leave it for a separate PR.

bahusoid:
> As this is not trivial to change, I decided to leave it for a separate PR.
Ok. Got it. So to clarify - did you try to do it globally (both for LINQ and hql)? Won't it be simpler to do just for Linq as a start?

maca88:
> did you try to do it globally (both for LINQ and hql)?
 
Yes, the issues that were mentioned are for implementing it globally.

> Won't it be simpler to do just for Linq as a start?

First I have to correct myself, Linq already supports expression like `where o.Association.PropertyId == 1`, `WhereJoinDetector` detects all associations and add joins for them. Implicit joins are only added for expression that are within an on statement:
```C#
from c in db.Customers join o in db.Orders on c.ContactTitle equals o.Customer.ContactTitle
```
where an implicit join is added for `o.Customer.ContactTitle`. With the last commit, `WhereJoinDetector` is also used for on statements and a cross join or inner join (fallback) is generated instead.

> Is this case tested by any CI configurations?

Now all cross join tests that I found in `LinqQuerySamples` and `JoinTests` are tested for both scenarios.

> Maybe we should simulate cross join as inner join with bogus condition like 1 = 1

The fallback is now added.

maca88:
The recent commit also fixes #1128, added a test for that and updated the description.

src/NHibernate/Hql/Ast/ANTLR/Hql.g
```diff
@@ -255,6 +256,7 @@ fromClause
 fromJoin
 	: ( ( ( LEFT | RIGHT ) (OUTER)? ) | FULL | INNER )? JOIN^ (FETCH)? path (asAlias)? (propertyFetch)? (withClause)?
 	| ( ( ( LEFT | RIGHT ) (OUTER)? ) | FULL | INNER )? JOIN^ (FETCH)? ELEMENTS! OPEN! path CLOSE! (asAlias)? (propertyFetch)? (withClause)?
+	| CROSS JOIN^ { WeakKeywords(); } path (asAlias)?
```
bahusoid: Shouldn't `propertyFetch` be supported here?

maca88: Yes it should, added.

src/NHibernate/Linq/ReWriters/AddJoinsReWriter.cs
```diff
@@ -59,6 +62,29 @@ public override void VisitNhHavingClause(NhHavingClause havingClause, QueryModel
...
+		{
+			VisitJoinClause(joinClause, queryModel, joinClause);
+		}
+
+		private void VisitJoinClause(JoinClause joinClause, QueryModel queryModel, IBodyClause bodyClause)
+		{
+			joinClause.InnerSequence = _whereJoinDetector.Transform(joinClause.InnerSequence);
+
+			// When associations are located in the outer key (e.g. from a in A join b in B b on a.C.D.Id equals b.Id),
+			// we have to insert the association join before the current join in order to produce a valid query.
```
bahusoid: Does this mean that such queries supported in LINQ but in HQL similar queries will fail:
`from A a inner join B b with a.C.D.Id = b.Id`
`from A a inner join B b with b.C.D.Id = a.Id`

maca88: > from A a inner join B b with a.C.D.Id = b.Id

This case is supported by HQL and for each association an inner join is added where in this PR a left outer join is added instead. After a second thought about what type of join would need to be used, I removed the logic for the outer key and leave it to HQL for adding inner joins. I had the following query in mind: 
```c#
from A a inner join B b with a.C.D.NullableProp = b.NullableProp
```
Linq to Objects would throw a NRE when `a.C` or `a.C.D` is null, but what should we do when generating the query? There are two options:
- Use inner joins for associations as HQL does today
- Use left joins for associations and use the logic that was done in #1996 in order to simulate what Linq to Objects would do for: `a?.C?.D?.NullableProp == b?.NullableProp`

My conclusion was that the second option is not practical in real world scenarios as it would join rows with `null` values.

> from A a inner join B b with b.C.D.Id = a.Id

This case is not supported by HQL.

bahusoid: > would join rows with null values.

But NULL values won't be joined anyway in DB unless you explicitly add `is null` check. null = null evaluates to false. So yes - left join seems unnecessary.

maca88: > But NULL values won't be joined anyway in DB unless you explicitly add is null check

By using the second option, the logic from #1996 would add the null checks

maca88: The last commit switches left to inner join also for `from A a inner join B b with b.C.D.Id = a.Id` case.

bahusoid: > `from A a inner join B b with b.C.D.Id = a.Id`
> This case is not supported by HQL

Ideally this case should be fixed in HQL but I guess it's out of scope for this PR.

Regarding switch to inner join for associations. What about left join?
`from A a left join B b with b.C.D.Id = a.Id`

That seems to be the case when left joins need to be added for associations. Again this PR is not about left joins but could you please clarify how it will work when #2328 is merged?

maca88: > could you please clarify how it will work when #2328 is merged?

The [last commit](https://github.com/nhibernate/nhibernate-core/pull/2328/commits/777701eaa268c173f2d4b730766baa37022cf0a7) in #2328 addresses the issue that the inner join is added only when the "parent join" is an inner join. By defualt `_whereJoinDetector` add joins as a left outer join, so we have to use `joinClause.MakeInner()` when we want an inner join. In #2328 `NhOuterJoinClause` class was added in order to distinguish whether is an outer join or not.

src/NHibernate.Test/Linq/LinqQuerySamples.cs
```diff
@@ -1558,14 +1672,27 @@ public void DLinqJoin8()
...
+		public void DLinqJoin9(bool useCrossJoin)
 		{
-			var expected =
+			if (useCrossJoin && !Dialect.SupportsCrossJoin)
+			{
+				Assert.Ignore("Dialect does not support cross join.");
+			}
+
+			ICollection expected, actual;
+			expected =
```
bahusoid: If I'm not mistaken - this query loads a big bunch or records in multiple db roundtrips just to check total count of expected records. Not good. How about commenting this part (just in case test data is changed) and leave in test only check with hardcoded expected count (as you already did ) and  leave comment that this value can be obtained from this query.

maca88: I agree, the query is now commented.

src/NHibernate/Linq/ReWriters/AddJoinsReWriter.cs
bahusoid: Got it. Thanks

PR #2328, opened 2020-03-15
Add left join support for Linq query provider

maca88:
Fixes: #864

This PR provides two additional ways to do a left join:
- Query syntax
```C#
from c in db.Customers
  join o in db.Orders on c.CustomerId equals o.Customer.CustomerId into orders
  from o in orders.DefaultIfEmpty()
  select o;
```
- Extension method
```C#
db.Orders
  .LeftJoin(db.Orders, x => x.OrderId, x => x.OrderId - 1, (order, order1) => new { order, order1 })
  .Select(x => new { First = x.order, Second = x.order1 })
  .ToList();
```

### DIFF
--- a/src/NHibernate/Linq/GroupJoin/NonAggregatingGroupJoinRewriter.cs
+++ b/src/NHibernate/Linq/GroupJoin/NonAggregatingGroupJoinRewriter.cs
@@ -168,7 +168,6 @@ namespace NHibernate.Linq.Visitors
 	{
 		private readonly IQuerySource _querySource;
 		private bool _references;
-		private readonly List<IQuerySource> _usages = new List<IQuerySource>();
 
 		public QuerySourceUsageLocator(IQuerySource querySource)
 		{
@@ -177,10 +176,7 @@ namespace NHibernate.Linq.Visitors
 
 		internal bool LeftJoin { get; private set; }
 
-		public IList<IQuerySource> Usages
-		{
-			get { return _usages.AsReadOnly(); }
-		}
+		public IList<IQuerySource> Usages { get; } = new List<IQuerySource>();
 
 		public void Search(IBodyClause clause)
 		{
@@ -195,7 +191,7 @@ namespace NHibernate.Linq.Visitors
 
 			if (_references)
 			{
-				_usages.Add(querySource);
+				Usages.Add(querySource);
 			}
 		}
 
--- a/src/NHibernate/Linq/GroupJoin/NonAggregatingGroupJoinRewriter.cs
+++ b/src/NHibernate/Linq/GroupJoin/NonAggregatingGroupJoinRewriter.cs
@@ -2,10 +2,12 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using NHibernate.Linq.Clauses;
 using NHibernate.Linq.GroupJoin;
 using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Clauses.ResultOperators;
 using Remotion.Linq.Parsing;
 
 namespace NHibernate.Linq.Visitors
@@ -14,7 +16,6 @@ namespace NHibernate.Linq.Visitors
 	{
 		private readonly QueryModel _model;
 		private readonly IEnumerable<GroupJoinClause> _groupJoinClauses;
-		private QuerySourceUsageLocator _locator;
 
 		private NonAggregatingGroupJoinRewriter(QueryModel model, IEnumerable<GroupJoinClause> groupJoinClauses)
 		{
@@ -67,19 +68,19 @@ namespace NHibernate.Linq.Visitors
 				//            This is used to repesent an outer join, and again the "from" is removing the hierarchy. So
 				//            simply change the group join to an outer join
 
-				_locator = new QuerySourceUsageLocator(nonAggregatingJoin);
+				var locator = new QuerySourceUsageLocator(nonAggregatingJoin);
 
 				foreach (var bodyClause in _model.BodyClauses)
 				{
-					_locator.Search(bodyClause);
+					locator.Search(bodyClause);
 				}
 
-				if (IsHierarchicalJoin(nonAggregatingJoin))
+				if (IsHierarchicalJoin(nonAggregatingJoin, locator))
 				{
 				}
-				else if (IsFlattenedJoin(nonAggregatingJoin))
+				else if (IsFlattenedJoin(nonAggregatingJoin, locator))
 				{
-					ProcessFlattenedJoin(nonAggregatingJoin);
+					ProcessFlattenedJoin(nonAggregatingJoin, locator);
 				}
 				else if (IsOuterJoin(nonAggregatingJoin))
 				{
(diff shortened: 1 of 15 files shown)


hazzik:
Please resolve conflicts

maca88:
Rebased and commented the Linq to Objects query.

bahusoid:
> Extension method

I wonder if it would be more useful to have extension method `Join` that takes `JoinType`. That should allow to make `right` join from LINQ query.

What about additional restrictions (`with` clause)? They seems to be not supported by extension.

bahusoid:
Just ignore my last comment (deleted it) - I peeked in to code and decided new extension is not required...

src/NHibernate/Linq/LinqExtensionMethods.cs
```diff
@@ -2463,6 +2464,97 @@ public static IFutureValue<TSource> ToFutureValue<TSource>(this IQueryable<TSour
...
+				Expression.Call(selectMany, exprGroupJoin, selectManyCollectionSelector, selectManyResultSelector));
+		}
+
+		private class LeftJoinIntermediate<TOuter, TInner>
+		{
+			public TOuter OneOuter { get; set; }
+			public IEnumerable<TInner> ManyInners { get; set; }
+		}
+
+		private class Replacer : ExpressionVisitor
```
hazzik: Re-linq already has a class which does this. `ReplacingSomethingVisitor`.

maca88: I didn't spot that before. Replaced with Relinq `ReplacingExpressionVisitor`.

src/NHibernate/Linq/GroupJoin/NonAggregatingGroupJoinRewriter.cs
```diff
@@ -156,27 +168,34 @@ internal class QuerySourceUsageLocator : RelinqExpressionVisitor
...
 			_querySource = querySource;
 		}
 
-		public IList<IBodyClause> Clauses
+		internal bool LeftJoin { get; private set; }
+
+		public IList<IQuerySource> Usages
 		{
-			get { return _clauses.AsReadOnly(); }
+			get { return _usages.AsReadOnly(); }
```
bahusoid: What's the point of wrapping collection as readonly on each property access in internal class?

src/NHibernate/Linq/GroupJoin/NonAggregatingGroupJoinRewriter.cs
```diff
@@ -92,19 +93,30 @@ private void ReWrite()
...
+				? new NhOuterJoinClause(nonAggregatingJoin.JoinClause)
+				: (IQuerySource) nonAggregatingJoin.JoinClause;
+
 			// Need to:
 			// 1. Remove the group join and replace it with a join
 			// 2. Remove the corresponding "from" clause (the thing that was doing the flattening)
-			// 3. Rewrite the selector to reference the "join" rather than the "from" clause
-			SwapClause(nonAggregatingJoin, nonAggregatingJoin.JoinClause);
+			// 3. Rewrite the query model to reference the "join" rather than the "from" clause
+			SwapClause(nonAggregatingJoin, (IBodyClause) nhJoin);
```
bahusoid: A bit unrelated.. But do you know any reasons why `SwapClause` method is implemented as `RemoveAt` and `Insert` instead of simply `bodyClauses[i] = newClause`

maca88: The only difference is in how the collection changed event is triggered, when using an indexer to set a value the event handler will be invoked once by using [NotifyCollectionChangedAction.Replace](https://referencesource.microsoft.com/#system/compmod/system/collections/objectmodel/observablecollection.cs,231),  and with `RemoveAt/Insert` it will be invoked two times. The same is done in `LeftJoinRewriter`:
https://github.com/nhibernate/nhibernate-core/blob/ac39173567b31bcfad475ca32687c9faf0d37f87/src/NHibernate/Linq/Visitors/LeftJoinRewriter.cs#L46-L47
and in `AggregatingGroupByRewriter`:
https://github.com/nhibernate/nhibernate-core/blob/4d47799230686aef2e3377bce42c8b35f7bd3674/src/NHibernate/Linq/GroupBy/AggregatingGroupByRewriter.cs#L86-L87

so maybe we should address this in a seprate PR if needed.

src/NHibernate/Linq/GroupJoin/NonAggregatingGroupJoinRewriter.cs
maca88: Dropped.

maca88:
Rebased.

src/NHibernate/Linq/GroupJoin/NonAggregatingGroupJoinRewriter.cs
bahusoid: Ok.. Yeah I figured this part. I thought you know if such difference is important somehow... Yeah can be fixed in other PR later..

maca88: >  I thought you know if such difference is important somehow.

To my knowledge it is not important and it should work the same.

src/NHibernate/Linq/QuerySourceNamer.cs
```diff
@@ -15,11 +15,16 @@ public class QuerySourceNamer
 		private int _differentiator = 1;
 
 		public void Add(IQuerySource querySource)
+		{
+			Add(querySource, CreateUniqueName(querySource.ItemName));
```
fredericDelaporte, line 19: This creates unused unique names when the `_map` already contains the `querySource`. It looks to me as a bad move.
I think this method should not call its overload but do all its little job by itself, allowing to keep an effective shortcut without evaluating twice the shortcut.

src/NHibernate/Util/ReflectionCache.cs
```diff
@@ -67,6 +67,11 @@ internal static class QueryableMethods
 		{
 			internal static readonly MethodInfo SelectDefinition =
 				ReflectHelper.GetMethodDefinition(() => Queryable.Select(null, default(Expression<Func<object, object>>)));
+			internal static readonly MethodInfo SelectManyDefinition =
+				ReflectHelper.GetMethodDefinition(() => Queryable.SelectMany<object, object, object>(null, (Expression<Func<object, IEnumerable<object>>>) null, null));
+
+			internal static readonly MethodInfo GroupJoinDefinition =
+				ReflectHelper.GetMethodDefinition(() => Queryable.GroupJoin<object, object, object, object>(null, null, null, null, null));
```
fredericDelaporte, line 74: These would have to be updated if #2352 gets merged before, or conversely, #2352 would have to be updated on these if #2328 gets merged before.

src/NHibernate/Linq/QuerySourceNamer.cs
maca88: I overlooked that, fixed.

src/NHibernate/Util/ReflectionCache.cs
maca88: Rebased and updated.

hazzik:
@maca88 I'm trying to understand why we need booth `NhJoin` (which has a flag `IsInner`) and `NhOuterJoin`?

maca88:
At first I tried to use `NhJoin` but it didn't worked out as it implements `IFromClause`, which requires `fromExpression` parameter that we don't have in this case. Ideally, if `JoinClause` would have a property whether it is an inner or outer join we wouldn't need `NhOuterJoinClause`, which would simplify the code and avoid registering both joins in `QuerySourceIdentifier`:
https://github.com/nhibernate/nhibernate-core/blob/0132e4cf6e6cb06bdcbf4b2fc70439f44a8ff36f/src/NHibernate/Linq/Visitors/QuerySourceIdentifier.cs#L63-L64
I was also thinking about tracking which `JoinClause` is an outer join in order to avoid having a wrapper, but the problem is that `JoinClause` can be cloned when rewriting the query model, which would probably ends up with a more complex solution.

bahusoid:
@hazzik  Do you still have some pending questions/plans with reviewing this PR?

bahusoid:
Oups.. Something is not right... LINQ tests are failing on master after merge

hazzik:
There is some interference with #2324.

Issue #2094, opened 2019-03-30
Query on subclass entity ("Table-per-subclass with discriminator" mapping) uses subclass' table alias with base class' key column name in join to collection in a dynamic component


gmcelhanon:
We have the following mappings for an abstract "EducationOrganziation" with a derived "School".

> Note that the PK column for _EducationOrganization_ is _EducationOrganizationId_, but in the derived entities, the key column name changes to match the derived entity name -- _SchoolId_ for _School_, _LocalEducationAgencyId_ for _LocalEducationAgency_, etc.

```xml
  <!-- Class definition -->
  <class name="EducationOrganization" table="EducationOrganization" lazy="false" abstract="true" discriminator-value="NULL">
    <id name="EducationOrganizationId" column="EducationOrganizationId" type="int">
      <generator class="assigned" />
    </id>

    <!-- Discriminator -->
    <discriminator column="Discriminator" type="string" />
    ...

    <!-- Table-per-subclass, using discriminator -->
    <subclass name="School" lazy="false" discriminator-value="edfi.School" >
      <join table="School" >
          <key>
            <column name="SchoolId" />
          </key>

          <!-- PK properties -->
          <property name="SchoolId" column="SchoolId" type="int" not-null="true" insert="false" update="false"/>

          <!-- Properties -->
          ...

          <!-- Collections -->
          <set name="SchoolCategories" cascade="all-delete-orphan" inverse="true" lazy="false">
            <key>
              <column name="SchoolId" />
            </key>
            <one-to-many class="SchoolCategory" />
          </set>

          <dynamic-component name="Extensions">
            <set name="Sample" cascade="all-delete-orphan" inverse="true" lazy="false">
              <key>
                <column name="SchoolId" />
              </key>
              <one-to-many class="SchoolExtension" />
            </set>
          </dynamic-component>
      </join>
    </subclass>
  </class>  
```

When we execute an HQL query like the following:
```sql
from School a left join fetch a.SchoolCategories b where a.Id IN (:ids)
```
... the following `FROM` clause is generated:
```sql
from edfi.EducationOrganization school0_ 
    inner join edfi.School school0_1_ 
        on school0_.EducationOrganizationId=school0_1_.SchoolId 
    left outer join edfi.SchoolCategory schoolcate1_ 
        on school0_.EducationOrganizationId=schoolcate1_.SchoolId
```
[...]

PR #2331, opened 2020-03-19
override of GetSubclassPropertyTableNumber in JoinedSubclassEntityPer…

### LINKED ISSUE #2330: join on multiple conditions
hi,

I've run into the following problem when trying to do a join on multiple columns:

```sql
select Id, col, fk_t2
from dbo.table1 t1
left outer join dbo.table2 t2 on t1.fk_t2 = t2.Id and t2.col = t1.col
where <filtering>
```

but the output is:

```sql
select Id, col, fk_t2
from dbo.table1 t1
left outer join dbo.table2 t2 on t1.fk_t2 = t2.Id and t2.col = t1_1.col
where <filtering>
```

The problem is that when there's inheritance join with a withClause and the column which is used in the withClause is present in the base and the inherited table the condition is not generated correctly. The table that is considered in the with clause is from the base table which is not yet joined.

I've checked the code and I think I've found a solution. Can you give me privileges to create a topic branch?

georgi-yakimov:
Fixes #2330


gliljas:
Thanks! Please add tests too.

georgi-yakimov:
hi @gliljas what do I need in order to run the tests locally? I have both nunit 2 and 3 installed. For some reason the tests are not getting started. Where I can read more about the unit tests practices you're following?

gliljas:
NUnit 3 is what is used. Personally I use the ReSharper test runner, but the one built into VS should work too. You should run the ShowBuildMenu.bat file in the project root.

georgi-yakimov:
The issue is not yet finished. I have to add unit tests.

src/NHibernate/Persister/Entity/IPropertyMapping.cs
```diff
@@ -29,6 +29,8 @@ public interface IPropertyMapping
 		/// <returns>true if a type was found, false if not</returns>
 		bool TryToType(string propertyName, out IType type);
 
+		string[] ToColumns(ICriteria pathCriteria, string propertyName, System.Func<ICriteria, string> getSQLAlias);
```
bahusoid: I didn't look in to actual implementation but this change looks wrong. `IPropertyMapping` interface shouldn't depend on any query API (`ICriteria` in your case)

georgi-yakimov:
Hi can someone please check what's going on with the travis? It's been in waiting status since the weekend? Thanks in advance

fredericDelaporte:
This sometimes happens, and usually pushing new commit solves it. As anyway you have a conflict to resolve and new issues in CodeFactor, you have some commits to add (eventually with a rebase), so let's see if the issue remains after that.

georgi-yakimov:
Hi. `continuous-integration/appveyor/pr` is failing and I can't figure out the reason why with looking at the log. Is it possible that the failure is not due to my changes?

hazzik:
@georgi-yakimov in your issue which you have originally opened you are talking about HQL, but it seems this PR concerns only Criteria API. Can you please clarify?

georgi-yakimov:
hi @hazzik the problem is that when there's inheritance `join` with a withClause and the column which is used in the withClause is present in the base and the inherited table the condition is not generated correctly. The table that is considered in the with clause is from the base table which is not yet joined. So instead of

```sql
select t1.*
from SomeTable t1
join Child c_1 on c_1.FK_Root_Id = t1.FK_Child_Id and c_1.IsActive = 1
join Root c_1_1 on c_1_1.Id = c_1.FK_Root_Id
```

the result is

```sql
select t1.*
from SomeTable t1
join Child c_1 on c_1.FK_Root_Id = t1.FK_Child_Id and c_1_1.IsActive = 1
join Root c_1_1 on c_1_1.Id = c_1.FK_Root_Id
```

You can check the unit tests I've added for more details

bahusoid:
This fix with use first/last index looks more like workaround for some real core issue. Can you explain more what's behind this fix?

Also I tried to convert your query to `QueryOver` which is basically a typed wrapper around Criteria and for some reasons your fix is no longer working:
```C#
PersonBase f = null;
var results =
	session.QueryOver<UserEntityVisit>()
			.JoinAlias(
				x => x.PersonBase,
				() => f,
				JoinType.LeftOuterJoin,
				Restrictions.Where(() => f.Deleted == false))
			.List();
```

bahusoid:
Also I might be wrong but it seems some of findings described here https://github.com/nhibernate/nhibernate-core/issues/2094 might be related to your issue

georgi-yakimov:
hey, could somebody please take a look why a required check is hanging for like a week by now? Thanks in advance

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -2110,13 +2110,16 @@ public virtual int GetSubclassPropertyTableNumber(string propertyPath)
 					return getSubclassColumnTableNumberClosure()[idx];
 				}
 			}*/
-			int index = Array.IndexOf(SubclassPropertyNameClosure, rootPropertyName); //TODO: optimize this better!
-			return index == -1 ? 0 : GetSubclassPropertyTableNumber(index);
+			int index = useLastIndex
```
hazzik, line 2113: what would be the case when `useLastIndex` is required to be `false`?

hazzik:
This really looks like a workaround. It is easy to break this logic.

If you add a subclass of `PersonBase` with `Deleted` property, then it will be broken.

The issue here is much deeper: only properties of the class itself are supported in the "with" clause.

bahusoid:
#2330 is actually also fixed by #2545 (for Criteria) and #2361 (hql). So it seems this PR can be closed as replaced

fredericDelaporte:
After having undone the fix in this PR, its tests work locally. So I have pushed the undo to check this with the CI.

fredericDelaporte:
So the test case alone without any change to NHibernate current master branch succeeds.

Issue #2241, opened 2019-10-07
Duplicate parameters in query


aleshchev:
Hi all,

Can you please remove duplicate parameters in query.

Code Example
```c#
var ids = new List<Guid> { Guid.NewGuid(), Guid.NewGuid() };
var products = session.Query<Product>().Where(z => ids.Contains(z.Id) && ids.Contains(z.Id)).ToList();
```

SQL Query
```sql
select product0_.Id as id1_0_, product0_.Name as name2_0_ 
from Product product0_ 
where (product0_.Id in (@p0 , @p1)) and (product0_.Id in (@p2 , @p3));
@p0 = dda54ba7-7407-4a29-a041-b88e4d2168a9 [Type: Guid (0:0:0)],
@p1 = bf65e0d9-76ca-4e11-87e9-e7557abc12dd [Type: Guid (0:0:0)],
@p2 = dda54ba7-7407-4a29-a041-b88e4d2168a9 [Type: Guid (0:0:0)],
@p3 = bf65e0d9-76ca-4e11-87e9-e7557abc12dd [Type: Guid (0:0:0)]
```

dmlittledm:
Same problem. Query with 2 same-value parameters sometimes causes MS SQL to build very bad execution plan that works hundreds times worse than same query with one parameter.

PR #2335, opened 2020-03-21
Minimize Linq query parameters

maca88:
Fixes: #2241

With this PR, only one sql parameter will be generated for each variable, except for ODBC drivers that do not support named parameters.

Possible breaking changes:
- `VisitorParameters.ConstantToParameterMap` may contain the same parameter for multiple constant expressions.


src/NHibernate.Test/Linq/ParameterTests.cs
```diff
@@ -0,0 +1,120 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using NUnit.Framework;
+
+namespace NHibernate.Test.Linq
+{
+	[TestFixture]
+	public class ParameterTests : LinqTestCase
```
hazzik: All these test would need to have a test demonstrating that the parameters are not over-cached.

src/NHibernate.Test/Linq/ParameterTests.cs
```diff
@@ -0,0 +1,120 @@
...
+			AssertTotalParameters(
+				db.Orders.Where(o => o.OrderId == value && o.OrderId != value),
+				1);
+		}
+
+		[Test]
+		public void UsingDifferentValueTypeParameters()
+		{
+			var value = 1;
+			var value2 = 2;
```
bahusoid: What about case `var value2 = 1;`? Will it be still 2  parameters or 1? I think it should still be 2 parameters (at least by default). As this query modifications due to parameter values may affect performance on DBs with cached execution plan.

hazzik:
There are complictions with this approach - it would mean that the same queries but different parameter values would have different query cache key (otherwise there will be an "overcaching issue"). This would lead to the re-translation of the linq query.

maca88:
> it would mean that the same queries but different parameter values would have different query cache key

Correct, in the current state this PR may create different query plans when there are parameters with the same type, which is not ideal.
What do you thnik if we would have the following logic for calculating parameters:
- A inline constant value would not be translated to a sql parameter
```C#
db.Products.Where(o => o.Name == "teest" && o.Active == true) // Zero sql parameters, both values would be translated to sql
```
- Each variable would be translated to a single sql parameter
```C#
var value = "test";
db.Products.Where(o => o.Name == value && o.Name != value) // One sql parameter
```
- Each inline non constant value would be translated to a separate sql parameter (remains the same)
```C#
db.Products.Where(o => o.Name == GetName(true) && o.Name != GetName(true)) // Two sql parameter

string GetName(bool value) 
{
	return value ? "test" : "test2";
}
```
I am not sure that it would take to implement such logic, but I would like to hear you opinions.

src/NHibernate.Test/Linq/ParameterTests.cs
maca88: > What about case var value2 = 1;? Will it still be 2 parameters or 1?

In the current state of this PR, only one parameter would be generated. Please check my last [comment](https://github.com/nhibernate/nhibernate-core/pull/2335#issuecomment-602035664).

maca88:
I forgot the case when an external libary can dynamically create a linq expression (e.g. OData) which means that the first rule (inline constant) cannot be applied. I will try to implement only the second rule (parameter for each variable).

bahusoid:
> Each variable would be translated to a single sql parameter

That's IMHO the only change that needs to be implemented in scope of this PR. And keep existing behavior for other cases. (I assume that existing behavior - new param generation for all cases)

> A inline constant value would not be translated to a sql parameter
I'm not sure that this behavior should be done as default. But as optional it would be actually very useful for query optimization purposes

maca88:
With the last commit only the second rule was implemented and the value matching logic was removed. Also, now the following queries will produce only one parameter:
```
var value = new Product {Name = "test"};
db.Products.Where(o => o.Name == value.Name && o.Name != value2.Name).ToList();
db.Orders.Where(o => o.OrderId == -value.ProductId && o.OrderId != -value.ProductId).ToList();
```

src/NHibernate/Linq/NhRelinqQueryParser.cs
```diff
@@ -50,12 +53,36 @@ static NhRelinqQueryParser()
...
 			return PreProcessor.Process(partiallyEvaluatedExpression);
 		}
 
+		/// <summary>
+		/// Applies the minimal transformations required before parameterization,
+		/// expression key computing and parsing.
+		/// </summary>
+		/// <param name="expression">The expression to transform.</param>
+		/// <returns><see cref="PartialEvaluationResult"/> that contains the transformed expression.</returns>
+		public static PartialEvaluationResult PartialEvaluate(Expression expression)
```
maca88: I had to change the method name as it differs only in the result type.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -1,38 +1,212 @@
...
+	/// subtrees that can be pre-evaluated before actually generating the query. Examples for evaluatable subtrees are operations on constant
+	/// values (constant folding), access to closure variables (variables used by the LINQ query that are defined in an outer scope), or method
+	/// calls on known objects or their members. In a second step, it replaces all of the evaluatable subtrees (top-down and non-recursive) by 
+	/// their evaluated counterparts.
+	/// </summary>
+	/// <remarks>
+	/// This visitor visits each tree node at most twice: once via the <see cref="EvaluatableTreeFindingExpressionVisitor"/> for analysis and once
+	/// again to replace nodes if possible (unless the parent node has already been replaced).
+	/// </remarks>
+	internal sealed class NhPartialEvaluatingExpressionVisitor : RelinqExpressionVisitor
```
maca88: Unfortunatelly, Relinq `PartialEvaluatingExpressionVisitor` class is sealed and the only way to link the evaluated expression with the variable was to copy the class and add the required logic.

src/NHibernate/Linq/Visitors/PartialEvaluationResult.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Collections.Generic;
+using System.Linq.Expressions;
+
+namespace NHibernate.Linq.Visitors
+{
+	/// <summary>
+	/// The result of <see cref="NhRelinqQueryParser.PartialEvaluate"/> method.
+	/// </summary>
+	public class PartialEvaluationResult
```
maca88: I created a separate class in order to be able to add additional data in the future.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -1,38 +1,212 @@
...
+			string currentPath;
+			switch (expression)
+			{
+				case MemberExpression memberExpression:
+					childExpression = memberExpression.Expression;
+					currentPath = memberExpression.Member.Name;
+					break;
+				case ConstantExpression constantExpression:
+					path = null;
+					if (constantExpression.Type.Attributes.HasFlag(TypeAttributes.NestedPrivate) &&
```
maca88: Used the same condition as Microsoft uses in [their ORM](https://github.com/dotnet/efcore/blob/e400969114d7a671b46bd9ba7b903bd8e7d63329/src/EFCore/Query/Internal/ParameterExtractingExpressionVisitor.cs#L604) in order to detect the compiler generated class containing the variables.

src/NHibernate.Test/Linq/ParameterTests.cs
maca88: Do we still need such test with the recent commit?

src/NHibernate.Test/Linq/ParameterTests.cs
```diff
@@ -0,0 +1,217 @@
...
+				ids.Count);
+		}
+
+		[Test]
+		public void UsingTwoListParameters()
+		{
+			var ids = new List<int> {11008, 11019, 11039};
+			var ids2 = new List<int> {11008, 11019, 11039};
+			AssertTotalParameters(
+				db.Orders.Where(o => ids.Contains(o.OrderId) && ids2.Contains(o.OrderId)),
```
bahusoid: Just curious how will it handle `new[]{ value, 2}.Contains(o.OrderId) && new []{value, 3}.Contains(o.OrderId)`.

src/NHibernate/Linq/NhRelinqQueryParser.cs
maca88: We can change it back to `PreTransform` when #1868 is merged as it adds session factory parameter.

src/NHibernate.Test/Linq/ParameterTests.cs
maca88: With the current logic 4 parameters will be added. This is quite a special case as `new[]{ value, 2}` is treated as one parameter and expanded later when executing the query.

bahusoid: Maybe cover it with test? And could you also add test for DML LINQ query.
Will this logic work for queries like (I'm actually not sure if it should be done for DML linq queries):
```C#
db.Customers.Where(c => c.CustomerId == "UNKNOWN").Update(x => new Customer {ContactName = value, CompanyName = value});
```

src/NHibernate.Test/Linq/ParameterTests.cs
bahusoid: Seems we need test to make sure that different cache key is generated for 
`db.Products.Where(o => o.Name == value  || o.Name == value).ToList();`
and
`db.Products.Where(o => o.Name == value1  || o.Name == value2).ToList();`

src/NHibernate.Test/Linq/ParameterTests.cs
maca88: > Maybe cover it with test?

Added.

> And could you also add test for DML LINQ query

I forgot about them. After some testing I found out that HQL does not support reusing parameters for DML queries, so now the new logic is not used for them anymore. Also, tests were added.

src/NHibernate.Test/Linq/ParameterTests.cs
maca88: > Seems we need test to make sure that different cache key is generated

Added `ValidateMixingTwoParametersCacheKeys` test for that, where the line expression key is checked which should be good enough I thnik.
Additionaly, added a check for all tests that the actual parameters number matches with the linq expression parameters.

src/NHibernate/Linq/Visitors/PartialEvaluationResult.cs
bahusoid: Maybe `QueryMode` should be added here?

bahusoid: Or maybe not... I see it's not very convenient..

maca88: > Maybe QueryMode should be added here?

If we add it here, then we would need to add it as internal otherwise would be confusing for users as it is not populated by the `PreTransform` method. But if we add it as internal, then users (also in our tests) cannot simulate what nhibernate does as `QueryMode` would be set internally by the `NhLinqExpression`. That is why I added it as an additional parameter. Also, with #2079 `ExpressionParameterVisitor` will be obsolete.

maca88:
Rebased.

src/NHibernate/Linq/NhRelinqQueryParser.cs
maca88: Renamed back to `PreTransform`.

src/NHibernate/Linq/Visitors/PartialEvaluationResult.cs
bahusoid: Well `PartialEvaluate` can take `QueryMode` as additional parameter... But OK - maybe it's not worth it.

> Also, with #2079 `ExpressionParameterVisitor` will be obsolete.

Could you please clarify how it correlates with this PR. Will you need to modify `ExpressionKeyVisitor` or what's used instead in similar way as you modified `ExpressionParameterVisitor`?

maca88: > Well PartialEvaluate can take QueryMode as additional parameter

That is another option which I think it is even better. By passing the information to the `PreTransform` method we could skip the variable detection logic for DML queries and avoid passing the `QueryMode` to `ExpressionParameterVisitor` as `QueryVariables` dictionary will be empty. With the last commit the `QueryMode` is passed into `PreTransform` method instead.

> Will you need to modify ExpressionKeyVisitor or what's used instead in similar way as you modified ExpressionParameterVisitor?

The logic of `ExpressionParameterVisitor` that detects whether an expression is a parameter or not, was moved into `ExpressionKeyVisitor`, other stuff like removing `MappedAs` calls and detecting parameter types was moved in the translation phase. So yes, the `ExpressionKeyVisitor` will need to be modified similar than `ExpressionParameterVisitor` is modified here.

src/NHibernate/Linq/Visitors/PreTransformationParameters.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Linq.Expressions;
+using NHibernate.Engine;
+using Remotion.Linq.Parsing.ExpressionVisitors.TreeEvaluation;
+
+namespace NHibernate.Linq.Visitors
+{
+	/// <summary>
+	/// Contains the information needed by <see cref="NhRelinqQueryParser.PreTransform"/> to perform an early transformation.
+	/// </summary>
+	public class PreTransformationParameters
```
maca88: Added a separate class, in order to be able to add addtitional information to pre-transformation phase in the future. One example would be to add support for:
- A inline constant value would not be translated to a sql parameter, [described here](https://github.com/nhibernate/nhibernate-core/pull/2335#issuecomment-602035664).

in order to add such feature we would need to extend `NhLinqExpression` constructor with `NhQueryableOptions` parameter that would be passed to `PreTransform` method, which would contain the information whether the parameters needs to be inlined or not.

src/NHibernate/Linq/NhRelinqQueryParser.cs
```diff
@@ -53,24 +54,33 @@ static NhRelinqQueryParser()
...
 		/// Applies the minimal transformations required before parametrization,
 		/// expression key computing and parsing.
 		/// </summary>
 		/// <param name="expression">The expression to transform.</param>
-		/// <param name="sessionFactory">The session factory.</param>
-		/// <returns>The transformed expression.</returns>
-		public static Expression PreTransform(Expression expression, ISessionFactoryImplementor sessionFactory)
+		/// <param name="parameters">The parameters used in the transformation process.</param>
+		/// <returns><see cref="PreTransformationResult"/> that contains the transformed expression.</returns>
+		public static PreTransformationResult PreTransform(Expression expression, PreTransformationParameters parameters)
```
maca88: Not a breaking change as the overload was added in #1868, which is not yet released.

src/NHibernate.Test/Linq/ParameterTests.cs
maca88: Added also the mentioned check for the query plan cache to verify whether queries are actually cached. The linq expression key is not enough as `CanCachePlan` is calculated in the `Translate` method.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
maca88: Added the license and copyright from Relinq.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -1,51 +1,231 @@
...
+			else
+			{
+				Expression<Func<object>> lambdaWithoutParameters = Expression.Lambda<Func<object>>(Expression.Convert(subtree, typeof(object)));
+				var compiledLambda = lambdaWithoutParameters.Compile();
+
+				object value = compiledLambda();
+				return Expression.Constant(value, subtree.Type);
+			}
+		}
+
```
fredericDelaporte, line 159:
```suggestion

		#endregion
		
		protected override Expression VisitConstant(ConstantExpression expression)
		{
			if (expression.Value is Expression value)
			{
				return EvaluateIndependentSubtrees(value, _preTransformationParameters);
			}
			return base.VisitConstant(expression);
		}

```

src/NHibernate.Test/Linq/ParameterTests.cs
```diff
@@ -0,0 +1,459 @@
...
+				1);
+		}
+
+		[Test]
+		public void UsingNegateValueTypeParameter()
+		{
+			var value = 1;
+			AssertTotalParameters(
+				db.Orders.Where(o => o.OrderId == value && o.OrderId != -value),
+				2);
```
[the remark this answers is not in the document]
maca88: With the last commit only one parameter will be used.

maca88:
Rebased.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -1,51 +1,195 @@
...
+			// When having multiple level closure, we have to evaluate each closure independently
+			if (evaluatedExpression is ConstantExpression constantExpression)
+			{
+				evaluatedExpression = VisitConstant(constantExpression);
+			}
+
+			// Variables in expressions are never a constant, they are encapsulated as fields of a compiler generated class.
+			// Skip detecting variables for DML queries as HQL does not support reusing parameters for them.
+			if (expression.NodeType != ExpressionType.Constant &&
+			    _preTransformationParameters.QueryMode == QueryMode.Select &&
```
bahusoid, line 112: I wonder maybe it's better to introduce separate property for controlling parameters minimization (something like `_preTransformationParameters.MinimizeParameters` ) which by default is initialized based on `QueryMode` value.

maca88: Probably better in case the information would be used elsewhere in the future, done.

bahusoid: Good. But let's make it public. I see no point in hiding DTO like properties and it might be useful in some custom user code implementation. And comments regarding `Skip detecting variables for DML`  should be moved to `MinimizeParameters` initialization code (also should we move it right inside `PreTransformationParameters` ctor?)

maca88: Made it public and moved the initialization inside ctor so that it can be set by the user.

PR #2341, opened 2020-03-31
Fetch requests do not work with collection projections

LodewijkSioen:
Added tests to show #1298 is fixed on the current master.

### DIFF
--- a/src/NHibernate.Test/Linq/ProjectionsTests.cs
+++ b/src/NHibernate.Test/Linq/ProjectionsTests.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using NHibernate.DomainModel.Northwind.Entities;
+using NHibernate.Linq;
 using NUnit.Framework;
 
 namespace NHibernate.Test.Linq
@@ -211,11 +212,27 @@ namespace NHibernate.Test.Linq
 			var result = query.ToList();
 			Assert.That(result.Count, Is.EqualTo(2155));
 		}
+		
+		[Test]
+		public void CanProjectManyCollectionsWithFetch()
+		{
+			var query = db.Orders.FetchMany(o => o.OrderLines).SelectMany(o => o.OrderLines);
+			var result = query.ToList();
+			Assert.That(result.Count, Is.EqualTo(2155));
+		}
 
 		[Test]
 		public void CanProjectCollections()
 		{
 			var query = db.Orders.Select(o => o.OrderLines);
+			var result = query.ToList();
+			Assert.That(result.Count, Is.EqualTo(830));
+		}
+
+		[Test]
+		public void CanProjectCollectionsWithFetch()
+		{
+			var query = db.Orders.Fetch(o => o.OrderLines).Select(o => o.OrderLines);
 			var result = query.ToList();
 			Assert.That(result.Count, Is.EqualTo(830));
 		}
@@ -229,11 +246,32 @@ namespace NHibernate.Test.Linq
 		}
 
 		[Test]
+		public void CanProjectCollectionsInsideAnonymousTypeWithFetch()
+		{
+			var query = db.Orders.Fetch(o => o.OrderLines).Select(o => new { o.OrderId, o.OrderLines });
+			var result = query.ToList();
+			Assert.That(result.Count, Is.EqualTo(830));
+		}
+
+		[Test]
 		public void ProjectAnonymousTypeWithCollection()
 		{
 			// NH-3333
 			// done by WCF DS: context.Orders.Expand(o => o.OrderLines) from the client 
 			var query = from o in db.Orders
+						select new { o, o.OrderLines };
+
+			var result = query.ToList();
+			Assert.That(result.Count, Is.EqualTo(830));
+			Assert.That(result[0].o.OrderLines, Is.EquivalentTo(result[0].OrderLines));
+		}
+
+		[Test]
+		public void ProjectAnonymousTypeWithCollectionWithFetch()
+		{
+			// NH-3333
+			// done by WCF DS: context.Orders.Expand(o => o.OrderLines) from the client 
+			var query = from o in db.Orders.Fetch(o => o.OrderLines)
 						select new { o, o.OrderLines };
 
 			var result = query.ToList();
@@ -255,6 +293,19 @@ namespace NHibernate.Test.Linq
 		}
 
 		[Test]
+		public void ProjectAnonymousTypeWithCollection1WithFetch()
+		{
+			// NH-3333
+			// done by WCF DS: context.Orders.Expand(o => o.OrderLines) from the client 
+			var query = from o in db.Orders.Fetch(o => o.OrderLines)
+						select new { o.OrderLines, o };
+
+			var result = query.ToList();
+			Assert.That(result.Count, Is.EqualTo(830));
+			Assert.That(result[0].o.OrderLines, Is.EquivalentTo(result[0].OrderLines));
+		}
+
+		[Test]
 		public void ProjectAnonymousTypeWithCollection2()
 		{
 			// NH-3333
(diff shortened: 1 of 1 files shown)


src/NHibernate.Test/Linq/ProjectionsTests.cs
```diff
@@ -241,6 +266,19 @@ public void ProjectAnonymousTypeWithCollection()
 			Assert.That(result[0].o.OrderLines, Is.EquivalentTo(result[0].OrderLines));
 		}
 
+		[Test]
+		public void ProjectAnonymousTypeWithCollectionWithFetch()
+		{
+			// NH-3333
+			// done by WCF DS: context.Orders.Expand(o => o.OrderLines) from the client
```
bahusoid: Above comments seems irrelevant for added test cases (applies to other added tests too)

src/NHibernate.Test/Linq/ProjectionsTests.cs
```diff
@@ -294,7 +369,26 @@ public void ProjectKnownTypeWithCollection()
...
+				            ExpandedElement = o,
+				            ProjectedProperty0 = o.OrderLines,
+				            Description = "OrderLine",
+				            ReferenceDescription = "OrderLine"
+			            };
+
+			var result = query.ToList();
+			Assert.That(result.Count, Is.EqualTo(830));
+			Assert.That(NHibernateUtil.IsInitialized(result[0].ExpandedElement.OrderLines), Is.True);
+			Assert.That(NHibernateUtil.IsInitialized(result[0].ProjectedProperty0), Is.True);
```
bahusoid: Add these checks to other tests you've added above

src/NHibernate.Test/Linq/ProjectionsTests.cs
```diff
@@ -294,7 +369,26 @@ public void ProjectKnownTypeWithCollection()
 			Assert.That(result.Count, Is.EqualTo(830));
 			Assert.That(result[0].ExpandedElement.OrderLines, Is.EquivalentTo(result[0].ProjectedProperty0));
 		}
-		
+
```
bahusoid: Add comment with issue number. Something like:
```C#
// NH-3396
```

bahusoid (dismissed):
Async code is missing. It can be generated via ShowBuildMenu.bat option H. Please note after 
 any modifications it needs to be regenerated again. And see my comments below

src/NHibernate.Test/Linq/ProjectionsTests.cs
LodewijkSioen: There is only one test that uses the `ExpandedWrapper`, but I can add the IsInitialized check to the other too.

LodewijkSioen:
I've run the 'H' option in the buildmenu, but no code was generated...

LodewijkSioen:
Ah, I did have an error, but it wasn't obvious:
```
     [exec] Starting 'dotnet (../Tools/CSharpAsyncGenerator.CommandLine/0.18.1/tools/netcoreapp2.1/AsyncGenerator.CommandLine.dll)' in 'C:\Projects\External\NHibernate\nhibernate-core\src'
     [exec] Could not execute because the specified command or file was not found.
     [exec] Possible reasons for this include:
     [exec]   * You misspelled a built-in dotnet command.
     [exec]   * You intended to execute a .NET Core program, but dotnet-../Tools/CSharpAsyncGenerator.CommandLine/0.18.1/tools/netcoreapp2.1/AsyncGenerator.CommandLine.dll does not exist.
     [exec]   * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
```
I don't have a netcoreapp2.1-folder. Only net472...

bahusoid:
> I don't have a netcoreapp2.1

Well yeah - .net core 2.1 is required to be installed on PC to make async generator work

LodewijkSioen:
No, I don't have this folder: "../Tools/CSharpAsyncGenerator.CommandLine/0.18.1/tools/netcoreapp2.1/" in my git checkout. only "../Tools/CSharpAsyncGenerator.CommandLine/0.18.1/tools/net472/".

src/NHibernate.Test/Linq/ProjectionsTests.cs
```diff
@@ -275,6 +340,18 @@ public void ProjectAnonymousTypeWithCollection3()
...
 			var result = query.ToList();
+			Assert.That(NHibernateUtil.IsInitialized(result[0].OrderLines), Is.True);
+			Assert.That(result.Count, Is.EqualTo(830));
+		}
+
+		[Test]
+		public void ProjectAnonymousTypeWithCollection3WithFetch()
+		{
+			var query = from o in db.Orders.Fetch(o => o.OrderLines)
+						select new { OrderLines = o.OrderLines.ToList() };
```
maca88, line 351: The sql created from this query returns more than 7000 rows, I think that it would be better to use the `PatientRecords` collection of `Patient` which has less data:
```
var query = from o in db.Patients.Fetch(o => o.PatientRecords)
						select new { PatientRecords = o.PatientRecords.ToList() };
```
the same applies for other queries that were added by this PR.

LodewijkSioen: Perhaps I should just add the test from the original bug report. Since doing a Fetch when projecting a collection is actually redundant. Adding that one test will guard against a regression for this code.

LodewijkSioen:
I'm sorry, but I'm unable to build the project using ShowBuildMenu. I emptied my .nuget folder, did a git-clean but I still get this exception: [buildlog.txt](https://github.com/nhibernate/nhibernate-core/files/4419332/buildlog.txt)

Buildding in Visual Studio works, but generating the async code after building in Studio gives me the following exception: [asynclog.txt](https://github.com/nhibernate/nhibernate-core/files/4419341/asynclog.txt)

src/NHibernate.Test/Linq/ProjectionsTests.cs
bahusoid: > Perhaps I should just add the test from the original bug report

Fine by me

maca88: I am also ok with that.

hazzik:
The workaround would be to run following in the root of the repository:

> dotnet new globaljson --sdk-version 2.1.700 --force

(Please check what 2.1 SDK version you have first)

@maca88 can you investigate support for 3.1?

hazzik:
Which PR has fixed the issue, anyone?

LodewijkSioen:
I had tried adding a global.json, but I probably used the wrong sdk-version. Anyhow: I changed the test as discussed. Only the test from the original Issue (#1298) is added and the async version is committed.

maca88:
> @maca88 can you investigate support for .net 3.1?

I am using .NET Core version 3.1.101 which works fine, but when I tried to install version `3.1.201`, I got the same error. With version `3.1.200+` the sdk is shipped with Roslyn version `3.5.0`, which I wasn't been able to make it work by using Roslyn `3.0.0` as a dependency. I suspect that [Microsoft.CodeAnalysis.Workspaces.MSBuild](https://www.nuget.org/packages/Microsoft.CodeAnalysis.Workspaces.MSBuild/3.5.0) is the problem as with `3.5.0` they dropped support for .NET Core 2.1. [Currently](https://github.com/maca88/AsyncGenerator/pull/146), the only way I was able to make it work with .NET Core version `3.1.201`, was to upgrade Roslyn dependencies to `3.5.0`.

maca88:
I've concluded that the only way to support `.NET Core 3.1` is to upgrade to Roslyn `3.6.0` which is not yet released. With version `3.5.0` the generator fails to generate code on linux and osx. By just upgrading Roslyn we lose support for .NET Core 2.1, so the new release will contain `netcoreapp2.1` and `netcoreapp3.1` versions, which means that if we want to generate async code with .NET Core `2.1` and `3.1` we will need to detect which version to use. [Here](https://github.com/maca88/nhibernate-core/commit/6c8c320dc069af36a9e4785983d02be0c31c46a3) is how nant configuration needs to be changed in order to support both versions. I am planning to release a new version when Roslyn `3.6.0` is released.

PR #2346, opened 2020-04-13
Fix SQLite typing

fredericDelaporte:
Invalid types were declared in SQLite dialect, causing cast errors, and having led to an undue hack disabling casting for some types.

Possible breaking change: in order to avoid a floating point division bug losing the fractional part, decimal are now stored as `REAL` instead of `NUMERIC`. Both are binary floating point types, excepted that `NUMERIC` stores integral values as `INTEGER`. This change may cause big integral decimal values to lose more precision in SQLite.

### DIFF
--- a/src/NHibernate/Dialect/SQLiteDialect.cs
+++ b/src/NHibernate/Dialect/SQLiteDialect.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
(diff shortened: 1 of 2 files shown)


src/NHibernate/Type/TimeAsTimeSpanType.cs
```diff
@@ -43,10 +43,11 @@ public override object Get(DbDataReader rs, int index, ISessionImplementor sessi
 			try
 			{
 				var value = rs[index];
-				if(value is TimeSpan time) //For those dialects where DbType.Time means TimeSpan.
+				if (value is TimeSpan time) //For those dialects where DbType.Time means TimeSpan.
 					return time;
-                
-				return ((DateTime)value).TimeOfDay;
+
+				var dbValue = Convert.ToDateTime(value);
```
bahusoid, line 49: Is this value really culture specific? Shouldn't invariant culture be provided here?

fredericDelaporte, line 49: The direct cast was not failing thanks to having previously used the "internal use only" `DATETIME` type, causing the data provider to do the convert itself. But this was also wrecking casting, and as documented in their [chm](https://system.data.sqlite.org/index.html/doc/trunk/Doc/SQLite.NET.chm?mimetype=application/x-chm), this type is meant for being used outside of the provider itself.

So we need know to do the convert, as are already doing other NHibernate types dealing with `DateTime`, see [`TimeType`](https://github.com/nhibernate/nhibernate-core/blob/25bc518461b71809d5701851468be0b9a550f832/src/NHibernate/Type/TimeType.cs#L60) by example.

src/NHibernate/Dialect/SQLiteDialect.cs
```diff
@@ -38,32 +37,46 @@ public SQLiteDialect()
...
 			RegisterColumnType(DbType.SByte, "INTEGER");
 			RegisterColumnType(DbType.UInt16, "INTEGER");
 			RegisterColumnType(DbType.UInt32, "INTEGER");
 			RegisterColumnType(DbType.UInt64, "INTEGER");
+
+			// NUMERIC and REAL are almost the same, they are binary floating point numbers. There is only a slight difference
+			// for values without a floating part. They will be represented as integers with numeric, but still as floating
+			// values with real. The side-effect of this is numeric being able of storing exactly bigger integers than real.
 			RegisterColumnType(DbType.Currency, "NUMERIC");
 			RegisterColumnType(DbType.Decimal, "NUMERIC");
```
maca88, line 57: In #2079 I [had to change](https://github.com/nhibernate/nhibernate-core/blob/d3557db6ff12539087f103d46872ca987c06a95b/src/NHibernate/Dialect/SQLiteDialect.cs#L53) `Currency` and `Decimal` to REAL in order to correct division operator from prevent returing an integer when two non decimal values are divided (e.g. 19/2 would return 9 if NUMERIC is used). Maybe we can fix the issue in this PR instead.

src/NHibernate/Type/TimeAsTimeSpanType.cs
fredericDelaporte: It is actually a variant of ISO8601, see SQLite type doc, followed by the System.Data.SQLite provider (using the variant without `T` but a space instead), so it is neither culture specific, nor culture sensitive.
Ideally this change should be done in every type doing this convert, so also in `AbstractDateTimeType` and `TimeType`. For me, this is out of the scope of this PR, and we should keep it consistent with other types.
(And we would also have to investigate whether the other converts should be switched to invariant culture or not, like those in `NDataReader`.)

bahusoid: Not all types suffer from culture specific conversions. DateTime can suffer very badly due to  month and day switch in different cultures. So I don't think introducing new code involving possible DateTime conversion from string should be consistent with other incorrect places.

src/NHibernate/Dialect/SQLiteDialect.cs
fredericDelaporte: This will result in a loss of precision for big integer numbers stored into a decimal, since with current mapping, they are handled as int by SQLite, while they will be handled a binary float with the `REAL` typing. So this would be a possible breaking change for some corner cases. But it seems better than having that divide bug.

src/NHibernate/Type/TimeAsTimeSpanType.cs
bahusoid: > Ideally this change should be done in every type doing this convert, so also in AbstractDateTimeType

I see two places with different datetime conversions also doesn't look good.. So at least TODO should be added to cover it in next pr

fredericDelaporte: Done.

bahusoid: https://github.com/nhibernate/nhibernate-core/issues/1173 might be related to this disscussion

src/NHibernate/Dialect/SQLiteDialect.cs
```diff
@@ -484,17 +498,5 @@ public override bool SupportsForeignKeyConstraintInAlterTable
 		// Said to be unlimited. http://sqlite.1065341.n5.nabble.com/Max-limits-on-the-following-td37859.html
 		/// <inheritdoc />
 		public override int MaxAliasLength => 128;
-
-		[Serializable]
-		protected class SQLiteCastFunction : CastFunction
```
hazzik, line 489: It seems this need to be obsoleted as it was `protected` for a while.

fredericDelaporte: Your are right, I have been a bit quick on this one.

src/NHibernate/Type/TimeAsTimeSpanType.cs
fredericDelaporte: The inverted case is highly provider dependent, I believe. In most cases, that is the provider which would do the conversion, not NHibernate.

mihaicodrean:
Was this really a good call when it comes to Decimal? There's no controlling of the decimals count now, when further casting to string, for example, making reporting a pain, for example. Moreover, the SQLite data types manual specifies the affinity of DECIMAL(10,5) to NUMERIC, not REAL.

fredericDelaporte:
There was never any actual control of decimal count with SQLite, it has no decimal support. It only has binary floats.

mihaicodrean:
You're right, I worded it ambiguously. I meant it more like decimals vs. no-decimals, because one now ends up with decimals where there were none previously, due to the replaced NUMERIC:
```
select sum(cast(FileSize as NUMERIC)) / 1024 / 1024 as MB from Files
=> 14
```
with REAL:
```
select sum(cast(FileSize as REAL)) / 1024 / 1024 as MB from Files
=> 14.9871826171875
```
This is a breaking change, with the workaround to add an extra cast:
```
select cast(sum(cast(FileSize as REAL)) / 1024 / 1024 as INTEGER) as MB from Files
=> 14
```

fredericDelaporte:
This case exploits some very specific behavior of SQLite, where a type that can handle a fractional part behaves as an integer regarding divisions if the provided number is integral. This looks to me as an extreme corner case. Better explicitly use integers where integer behavior is required.

PR #2097, opened 2019-04-01
Add support for fetching an individual lazy property with Criteria

bahusoid:
Follow up to #1922 
To avoid accidental misuse with `SelectMode.FetchLazyProperties` new enum is called `SelectMode.FetchLazyPropertyGroup` (and not `SelectMode.FetchLazyProperty`)

Example of usage:
```C#
var entity = session.QueryOver<Entity>()
	.Fetch(
		SelectMode.FetchLazyPropertyGroup,
		 ec => ec.LazyProp1, ec => ec.LazyProp2, ec => ec.OtherEntity.LazyProp3)
	.Where(ec => ec.Id == id)
	.SingleOrDefault();
```

Fixes #1861

PR #2347, opened 2020-04-14
Support fetching individual lazy properties for Criteria EntityProjection

bahusoid:
Related to #2097 (I forgot about `Criteria` `EntityProjection`) and #1922

### DIFF
--- a/src/NHibernate/Loader/OuterJoinableAssociation.cs
+++ b/src/NHibernate/Loader/OuterJoinableAssociation.cs
@@ -237,8 +237,8 @@ namespace NHibernate.Loader
 						                    new EntityLoadInfo(entitySuffix)
 						                    {
 							                    LazyProperties = EntityFetchLazyProperties,
-												IncludeLazyProps = SelectMode == SelectMode.FetchLazyProperties,
-											});
+							                    IncludeLazyProps = SelectMode == SelectMode.FetchLazyProperties,
+						                    });
 				case SelectMode.ChildFetch:
 					return ReflectHelper.CastOrThrow<ISupportSelectModeJoinable>(Joinable, "child fetch select mode")
 					                    .IdentifierSelectFragment(RHSAlias, entitySuffix);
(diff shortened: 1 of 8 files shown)


src/NHibernate/Persister/Entity/IQueryable.cs
```diff
@@ -16,7 +17,7 @@ internal static class AbstractEntityPersisterExtensions
 		/// Given a query alias and an identifying suffix, render the property select fragment.
 		/// </summary>
 		//6.0 TODO: Merge into IQueryable
-		public static string PropertySelectFragment(this IQueryable query, string alias, string suffix, string[] fetchProperties)
+		public static string PropertySelectFragment(this IQueryable query, string alias, string suffix, ICollection<string> fetchProperties)
```
bahusoid, line 20: Not a breaking change (added in #1952)

src/NHibernate/SelectMode.cs
```diff
@@ -33,12 +33,12 @@ public enum SelectMode
 		JoinOnly,
 
 		/// <summary>
-		/// Skips fetching for eagerly mapped association (no-op for lazy association).
```
bahusoid, line 36: It won't skip  `lazy="false"` association from loading - so better say it explicitly which case is handled here

src/NHibernate/Loader/OuterJoinableAssociation.cs
```diff
@@ -228,16 +229,17 @@ internal string GetSelectFragment(string entitySuffix, string collectionSuffix,
...
+											next?.Joinable,
+											next?.RHSAlias,
+											RHSAlias,
+											collectionSuffix,
+											ShouldFetchCollectionPersister(),
+											new EntityLoadInfo(entitySuffix)
+											{
+												LazyProperties = EntityFetchLazyProperties, 
+												IncludeLazyProps = SelectMode == SelectMode.FetchLazyProperties,
+											});
```
fredericDelaporte, line 242: Wider code change than expected, due to some bad interaction between VS formatting and JetBrains formatting while using tabs, I think.
For this reason, I tend to prefer writing:
```
return aThing
	.DoSomething()
	.DoSomethingElse();
```
Rather than:
```
return aThing.DoSomething()
             .DoSomethingElse();
```
The former does not change on reformat, while the later tends to get wrecked on reformat.

Anyway here, may you either revert to previous formatting, or if changing it, change for a more stable one?

bahusoid: restored previous formatting

PR #2349, opened 2020-04-14
Add support for Oracle binary floating point types

maca88:
Extracted from #2079 and aligned the registration with #1868.

### DIFF
--- a/src/NHibernate/Dialect/Oracle10gDialect.cs
+++ b/src/NHibernate/Dialect/Oracle10gDialect.cs
@@ -1,5 +1,10 @@
+using System.Collections.Generic;
+using System.Data;
+using NHibernate.Cfg;
 using NHibernate.Dialect.Function;
 using NHibernate.SqlCommand;
+using NHibernate.SqlTypes;
+using NHibernate.Util;
 
 namespace NHibernate.Dialect
 {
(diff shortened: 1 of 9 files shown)


src/NHibernate/Dialect/Oracle10gDialect.cs
```diff
@@ -29,19 +64,6 @@ protected override void RegisterFunctions()
 			RegisterFunction("random", new SQLFunctionTemplate(NHibernateUtil.Double, "cast(DBMS_RANDOM.VALUE() as binary_double)"));
 		}
 
-		/* 6.0 TODO: consider redefining float and double registrations
-		protected override void RegisterNumericTypeMappings()
```
maca88, line 33: Here _useBinaryFloatingPointTypes is not yet populated as `RegisterNumericTypeMappings` is called in the constructor.

src/NHibernate/Driver/OracleClientDriver.cs
```diff
@@ -8,6 +9,9 @@ namespace NHibernate.Driver
 	/// <summary>
 	/// A NHibernate Driver for using the Oracle DataProvider.
 	/// </summary>
+	// Since v5.3
+	// Deprecated by Microsoft: https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/oracle-and-adonet
+	[Obsolete("Use OracleManagedDataClientDriver or OracleDataClientDriver driver instead.")]
 	public class OracleClientDriver : ReflectionBasedDriver
```
hazzik, line 15: Let's not deprecate it yet.

fredericDelaporte: If we do not deprecate it before a major release which would switch to .Net Framework 5, that would mean a drop of this without prior warning through a deprecation.

What is the trouble with deprecating it? Forcing people to add directives for disabling deprecation warning? It should be very infrequent, since this is usually loaded through configuration, and so not seen by the compiler.

At "worst", a NHibernate warning log will be emitted. (I know in some cases NHibernate does detect and warn about obsolete classes used through configuration, but I have not checked if this is the case for drivers.)

But as this is not the subject of this PR, why not doing this in another PR anyway.

hazzik: Ok

PR #2350, opened 2020-04-16
Optimize LINQ batch item processing for queries with overridden result type

bahusoid:


src/NHibernate/Multi/LinqBatchItem.cs
```diff
@@ -9,8 +10,15 @@
...
 namespace NHibernate.Multi
 {
+	interface ILinqBatchItem
+	{
+		List<TResult> GetTypedResults<TResult>();
+	}
+
 	public static class LinqBatchItem
 	{
+		internal static ConcurrentDictionary<System.Type, Func<ILinqBatchItem, IList>> GetResultsForTypeDic = new ConcurrentDictionary<System.Type, Func<ILinqBatchItem, IList>>();
```
bahusoid, line 20: Let's generate compiled delegate returning typed results and cache it.

It's OK to have static cache as it will contain only few items (and for now it's only for nullable types)

src/NHibernate/Multi/LinqBatchItem.cs
```diff
@@ -90,27 +99,24 @@ private List<T> GetTransformedResults(IList transformerList)
...
 
-		private System.Type GetResultTypeIfChanged()
+		private IList GetTypedResults(System.Type type)
 		{
-			if (_postExecuteTransformer == null)
-			{
-				return null;
-			}
-			var elementType = _postExecuteTransformer.Method.GetParameters()[1].ParameterType.GetGenericArguments()[0];
-			if (typeof(T).IsAssignableFrom(elementType))
```
bahusoid, line 100: There is no need to do such analysis - if type is overridden its type provided in `ExpressionToHqlTranslationResults.ExecuteResultTypeOverride`

hazzik:
What is the performance improvement?

bahusoid:
>What is the performance improvement? 

Negligible to investigate. That's some old work that's been dusting in my local repos. If you think that compiled expression generation looks ugly or may worsen the perf - I will undo this part and just leave it as clean up PR that removes unnecessary  `GetResultTypeIfChanged`

hazzik:
I've replaced lambda compilation with `MethodInfo.CreateDelegate` hopefully it works.

bahusoid:
Looks cleaner  :)

bahusoid:
But it doesn't work.. 
```
System.NotSupportedException: Specified method is not supported.
   at System.Delegate.BindToMethodInfo(Object target, IRuntimeMethodInfo method, RuntimeType methodType, DelegateBindingFlags flags)
```

src/NHibernate/Multi/LinqBatchItem.cs
```diff
@@ -20,6 +20,10 @@ public static class LinqBatchItem
 	{
 		internal static ConcurrentDictionary<System.Type, Func<ILinqBatchItem, IList>> GetResultsForTypeDic = new ConcurrentDictionary<System.Type, Func<ILinqBatchItem, IList>>();
 
+		internal static MethodInfo GetTypedResultsMethod = ReflectHelper
+			.GetMethod((ILinqBatchItem i) => i.GetTypedResults<object>())
```
bahusoid, line 24: Did you test it? I think it would still fail. It wasn't working because it's not generic delegate

hazzik:
Yeah, I saw :-(

src/NHibernate/Multi/LinqBatchItem.cs
hazzik: It should work. It is (almost) the same code as you had.

bahusoid: That's delegate you create here:
```C#
public IList MyMethod(ILinqBatchItem i) => i.GetTypedResults<object>();
```
It's not generic. I will fix it soon

bahusoid: Hm.. Tests are green...  But delegate always returns `List<object>`. Don't you agree? It's just seems not important for affected queries that return aggregate results.. I'm still going to push fixed changes with your first solution to avoid boxing...

bahusoid: Ok.. I'm wrong. I tested it and it works as expected. I misunderstood how `ReflectHelper.GetMethod`.. But strange why `CreateDelegate` didn't work with it

hazzik: I’m calling ``GetMethod works in reverse of `Expression.Call` — it assumes that lambda’s body consists of only `MethodCalExpression` and extracts it’s `MethodInfo`.

In this case the MethodInfo is already closed generic with “object” generic argument, so I’m calling `GetGenericMethodDefinition()`

PR #2354, opened 2020-04-19
Do not require Mono to build on not Windows 

hazzik:
Use official `Microsoft.NETFramework.ReferenceAssemblies` package to cross-build .NET 4.6.1 instead of mono binaries.


.travis.yml
```diff
@@ -1,6 +1,6 @@
 dist: trusty
 language: csharp
-mono: latest
+mono: none
```
hazzik, line 3: This shreds around 5 minutes off build

build-common/NHibernate.props
```diff
@@ -38,4 +37,10 @@
     
     <DisableImplicitPackageTargetFallback>True</DisableImplicitPackageTargetFallback>
   </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
```
fredericDelaporte, line 43: What is the reason for this detailed configuration, over just using:
```
  <ItemGroup>
    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
  </ItemGroup>
```
?

hazzik: No reason, it is what was generated by `dotnet add package`

hazzik: @fredericDelaporte done

bahusoid:
I'm a bit confused. Please confirm we don't run any tests for Full .NET framework (mono) on linux. Why this mono stuff was there? Just to test that Full .NET framework can be compiled on linux?

hazzik:
To build by `dotnet` otherwise we would need to cross-build on mono; or exclude full .net targets from build on Linux.

But personally this is for me to support development on MacOS.

PR #2358, opened 2020-04-22
Support criteria aliases in SqlProjection and SQLCriterion

### LINKED ISSUE #896: NH-1432 - Expression.Sql should support aliases other than {alias}
**Jerome Haltom** created an [issue ](https://nhibernate.jira.com/browse/NH-1432)— <time datetime="2008-08-07T08:04:22.094Z" title="2008-08-07T08:04:22.094Z">7th August 2008, 8:04:22</time>:
> Expression.Sql("") only allows use of the root alias using {alias}. It should allow usage of any subcriteria alias. Attached is a patch which implements this.
> 
> The patch is non-ideal in that it pokes a whole through to CriteriaImpl, which it probably should not do. I'm seeking advice on the best way to obtain the required information.
> 
---
**Fabio Maulo** added a comment — <time datetime="2008-08-07T09:33:20.604Z" title="2008-08-07T09:33:20.604Z">7th August 2008, 9:33:20</time>:
> This is part of the documentation of Expression class.
> 	/// This class is semi-deprecated. Use <see cref="Restrictions"/>. 
> 
> For me this issue is candidate to "won't fix".
> 
> 
---
**Jerome Haltom** added a comment — <time datetime="2008-08-18T09:22:31.717Z" title="2008-08-18T09:22:31.717Z">18th August 2008, 9:22:31</time>:
> Hmm. I meant to set it as effecting 1.2.1. I can't seem to figure out how to do that though.
> 
---
**Jesse** added a comment — <time datetime="2008-10-22T18:17:55.689Z" title="2008-10-22T18:17:55.689Z">22nd October 2008, 18:17:55</time>:
> Restrictions don't contain an operation to  Create a Sql filter except for projections. To add a where clause filter we still have to use Expression.Sql.  I have attached a patch which uses a regular expression to find the aliases and properly apply the table alias used in the query.  This makes it so that you can have multiple aliases like this  
> 
> "dbo.MinDate({alias}.dtEndDate, {type}.dtEndDate) > ?" 
> 
> where {alias} still refers to the root entity and {type} refers to another alias in the criteria.
> 
> 
> 
---
**Jesse** added a comment — <time datetime="2008-10-22T18:18:44.926Z" title="2008-10-22T18:18:44.926Z">22nd October 2008, 18:18:44</time>:
> Adding Patch as mentioned in previous comment
> 
---
**luis cornejo** added a comment — <time datetime="2008-10-28T23:30:23.404Z" title="2008-10-28T23:30:23.404Z">28th October 2008, 23:30:23</time>:
> I second this, the aliases references should be the same as if one id using createSQL
> 
---
**luis cornejo** added a comment — <time datetime="2008-10-28T23:34:28.539Z" title="2008-10-28T23:34:28.539Z">28th October 2008, 23:34:28</time>:
> Also, 
> 
> if it won't be implemented, can anyone suggest a way to get the right alias of subcriterias in there for custom sql use with criterias?
> 
---
**lcorneliussen** added a comment — <time datetime="2009-11-11T05:17:32.220Z" title="2009-11-11T05:17:32.220Z">11th November 2009, 5:17:32</time>:
> +1 Especially when working with legacy-databases it is often necessary to create custom sql-statements. It would be great if we could reuse the aliases there.
> 
---
**Boris Drajer** added a comment — <time datetime="2012-04-09T10:46:54.329Z" title="2012-04-09T10:46:54.329Z">9th April 2012, 10:46:54</time>:
> +1. This is important as a last-resort solution for all database-specific features that aren't supported in another way. Supporting a single alias here means supporting only the most trivial cases.
> 
> Note that the previous patches don't seem to resolve certain types of subqueries - if I'm not mistaken, it's the detached queries that don't work. For example, a query like
> 
> mainQuery.Where(Subqueries.Exists(detachedCriteria.Where(Expression.Sql(sql_expression)))
> 
> will fail to find mainQuery aliases referenced by sql_expression. Looking through the source, I have been able to find the way to traverse from the inner expression to the outer, but it involved a bit of reflection and a casting of interfaces to real types, so it's a bit of a hack. I'm including a patch as a proof of concept, but it needs to be worked on by someone who knows more about NH.
> 
---
**Josh Berke** added a comment — <time datetime="2013-01-16T15:11:08.911Z" title="2013-01-16T15:11:08.911Z">16th January 2013, 15:11:08</time>:
> I ran into this issue as well...aliases should work the same way across the system.
>

bahusoid:
Fixes #896

### DIFF
--- a/src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
+++ b/src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
@@ -1069,7 +1069,7 @@ namespace NHibernate.Test.Criteria
 				var subquery = QueryOver.Of(() => studentSubquery)
 				         .And(
 					         Restrictions.Sql("{e}.studentId = 667 and {studentSubquery}.studentId = 667")
-					                   .AddCriteriaAliases("e", "studentSubquery")).Select(Projections.Id());
+					                   .AddAliases("e", "studentSubquery")).Select(Projections.Id());
 				         
 				var uniqueResult = s.CreateCriteria(typeof(Student))
 				                    .Add(Subqueries.Exists(subquery.DetachedCriteria))
@@ -1090,7 +1090,7 @@ namespace NHibernate.Test.Criteria
 							                    TypeFactory.HeuristicType(typeof(short)),
 							                    TypeFactory.HeuristicType(typeof(string)),
 							                    TypeFactory.HeuristicType(typeof(string)),
-						                    }).AddCriteriaAliases("e", "c"))
+						                    }).AddAliases("e", "c"))
 				                    .UniqueResult();
 
 				Assert.That(uniqueResult, Is.Not.Null);
--- a/src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
+++ b/src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
@@ -1065,8 +1065,14 @@ namespace NHibernate.Test.Criteria
 
 			using (var s = OpenSession())
 			{
+				Student studentSubquery = null;
+				var subquery = QueryOver.Of(() => studentSubquery)
+				         .And(
+					         Expression.Sql("{e}.studentId = 667 and {studentSubquery}.studentId = 667")
+					                   .AddCriteriaAliases("e", "studentSubquery")).Select(Projections.Id());
+				         
 				var uniqueResult = s.CreateCriteria(typeof(Student))
-				                    .Add(Expression.Sql("{e}.studentId = 667").AddCriteriaAliases("e"))
+				                    .Add(Subqueries.Exists(subquery.DetachedCriteria))
 				                    .AddOrder(Order.Asc("Name"))
 				                    .CreateCriteria("Enrolments", "e")
 				                    .AddOrder(Order.Desc("Year"))
--- a/src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
+++ b/src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using NHibernate.Dialect;
 using NHibernate.Criterion;
+using NHibernate.Linq;
 using NHibernate.SqlCommand;
 using NHibernate.Transform;
 using NHibernate.Type;
@@ -1017,6 +1018,87 @@ namespace NHibernate.Test.Criteria
 
 			t.Commit();
 			s.Close();
+		}
+
+		[Test]
+		public void TestSQLProjectionWithAliases()
+		{
+			using(ISession s = OpenSession())
+			using(ITransaction t = s.BeginTransaction())
+			{
+				Course course = new Course();
+				course.CourseCode = "HIB";
+				course.Description = "Hibernate Training";
+				s.Save(course);
+
+				Student gavin = new Student();
+				gavin.Name = "Gavin King";
+				gavin.StudentNumber = 667;
+				s.Save(gavin);
+
+				Student xam = new Student();
+				xam.Name = "Max Rydahl Andersen";
+				xam.StudentNumber = 101;
+				s.Save(xam);
+
+				Enrolment enrolment = new Enrolment();
+				enrolment.Course = course;
+				enrolment.CourseCode = course.CourseCode;
+				enrolment.Semester = 1;
+				enrolment.Year = 1999;
+				enrolment.Student = xam;
+				enrolment.StudentNumber = xam.StudentNumber;
+				xam.Enrolments.Add(enrolment);
+				s.Save(enrolment);
+
+				enrolment = new Enrolment();
+				enrolment.Course = course;
+				enrolment.CourseCode = course.CourseCode;
+				enrolment.Semester = 3;
+				enrolment.Year = 1998;
+				enrolment.Student = gavin;
+				enrolment.StudentNumber = gavin.StudentNumber;
+				gavin.Enrolments.Add(enrolment);
+				s.Save(enrolment);
+				t.Commit();
+			}
+
+			using (var s = OpenSession())
+			{
+				var uniqueResult = s.CreateCriteria(typeof(Student))
+				                    .Add(Expression.Like("Name", "Gavi", MatchMode.Start))
+				                    .AddOrder(Order.Asc("Name"))
+				                    .CreateCriteria("Enrolments", "e")
+				                    .AddOrder(Order.Desc("Year"))
+				                    .AddOrder(Order.Desc("Semester"))
+				                    .CreateCriteria("Course", "c")
+				                    .AddOrder(Order.Asc("Description"))
+				                    .SetProjection(
+					                    Projections.SqlProjection(
+						                    "{alias}.studentId as studentNumber, {e}.Year as year, {e}.Semester as semester,"
+						                    + " {c}.CourseCode as courseCode, {c}.Description as descr",
+						                    new string[] {"studentNumber", "year", "semester", "courseCode", "descr"},
+						                    new[]
+						                    {
+							                    TypeFactory.HeuristicType(typeof(long)),
+							                    TypeFactory.HeuristicType(typeof(short)),
+							                    TypeFactory.HeuristicType(typeof(short)),
+							                    TypeFactory.HeuristicType(typeof(string)),
+							                    TypeFactory.HeuristicType(typeof(string)),
+						                    }).AddCriteriaAliases("e", "c"))
+				                    .UniqueResult();
+
+				Assert.That(uniqueResult, Is.Not.Null);
+			}
+
+			using (var s = OpenSession())
+			using (s.BeginTransaction())
+			{
+				s.Query<Enrolment>().Delete();
+				s.Query<Student>().Delete();
+				s.Query<Course>().Delete();
+				s.Transaction.Commit();
+			}
 		}
 
 		[Test]
--- a/src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
+++ b/src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
@@ -971,7 +971,7 @@ namespace NHibernate.Test.Criteria
 			ProjectionList p2 = Projections.ProjectionList()
 				.Add(Projections.Min("StudentNumber"))
 				.Add(Projections.Avg("StudentNumber"))
-				.Add(Projections.SqlProjection(
+				.Add(Projections.Sql(
 						"1 as constOne, count(*) as countStar",
 						new String[] { "constOne", "countStar" },
 						new IType[] { NHibernateUtil.Int32, NHibernateUtil.Int32 }
@@ -1068,7 +1068,7 @@ namespace NHibernate.Test.Criteria
 				Student studentSubquery = null;
 				var subquery = QueryOver.Of(() => studentSubquery)
 				         .And(
-					         Expression.Sql("{e}.studentId = 667 and {studentSubquery}.studentId = 667")
+					         Restrictions.Sql("{e}.studentId = 667 and {studentSubquery}.studentId = 667")
 					                   .AddCriteriaAliases("e", "studentSubquery")).Select(Projections.Id());
 				         
 				var uniqueResult = s.CreateCriteria(typeof(Student))
@@ -1080,7 +1080,7 @@ namespace NHibernate.Test.Criteria
 				                    .CreateCriteria("Course", "c")
 				                    .AddOrder(Order.Asc("Description"))
 				                    .SetProjection(
-					                    Projections.SqlProjection(
+					                    Projections.Sql(
 						                    "{alias}.studentId as studentNumber, {e}.Semester as semester,"
 						                    + " {c}.CourseCode as courseCode, {c}.Description as descr",
 						                    new string[] {"studentNumber", "semester", "courseCode", "descr"},
@@ -1250,7 +1250,7 @@ namespace NHibernate.Test.Criteria
 			ProjectionList p2 = Projections.ProjectionList()
 				.Add(Projections.Min("StudentNumber"))
 				.Add(Projections.Avg("StudentNumber"))
-				.Add(Projections.SqlProjection(
+				.Add(Projections.Sql(
 						"1 as constOne, count(*) as countStar",
 						new String[] { "constOne", "countStar" },
 						new IType[] { NHibernateUtil.Int32, NHibernateUtil.Int32 }
@@ -1550,7 +1550,7 @@ namespace NHibernate.Test.Criteria
 			ProjectionList p2 = Projections.ProjectionList()
 				.Add(Property.ForName("StudentNumber").Min())
 				.Add(Property.ForName("StudentNumber").Avg())
-				.Add(Projections.SqlProjection(
+				.Add(Projections.Sql(
 					"1 as constOne, count(*) as countStar",
 					new String[] { "constOne", "countStar" },
 					new IType[] { NHibernateUtil.Int32, NHibernateUtil.Int32 }
@@ -2043,7 +2043,7 @@ namespace NHibernate.Test.Criteria
 			ProjectionList p2 = Projections.ProjectionList()
 				.Add(Property.ForName("StudentNumber").Min())
 				.Add(Property.ForName("StudentNumber").Avg())
-				.Add(Projections.SqlProjection(
+				.Add(Projections.Sql(
 						"1 as constOne, count(*) as countStar",
 						new String[] { "constOne", "countStar" },
 						new IType[] { NHibernateUtil.Int32, NHibernateUtil.Int32 }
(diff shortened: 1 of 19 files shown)


src/NHibernate/Criterion/Projections.cs
```diff
@@ -214,26 +214,21 @@ public static AggregateProjection Sum(IProjection projection)
...
+		/// A SQL projection, a typed select clause fragment.
+		/// The string {alias} will be replaced by the alias of the root entity.
+		/// Criteria aliases can also be used: "{a}.Value + {bc}.Value". Such aliases need to be registered via call to AddCriteriaAliases("a", "bc")
 		/// </summary>
-		/// <param name="sql"></param>
-		/// <param name="columnAliases"></param>
-		/// <param name="types"></param>
-		/// <returns></returns>
-		public static IProjection SqlProjection(string sql, string[] columnAliases, IType[] types)
+		public static SQLProjection SqlProjection(string sql, string[] columnAliases, IType[] types)
```
fredericDelaporte, line 221: Same here.

src/NHibernate/Criterion/Expression.cs
```diff
@@ -18,57 +18,72 @@ private Expression()
...
 		/// Apply a constraint expressed in SQL, with the given SQL parameters
+		/// The string {alias} will be replaced by the alias of the root entity.
+		/// Criteria aliases can also be used: "{a}.Value + {bc}.Value". Such aliases need to be registered via call to AddCriteriaAliases("a", "bc")
 		/// </summary>
 		/// <param name="sql"></param>
 		/// <param name="values"></param>
 		/// <param name="types"></param>
 		/// <returns></returns>
-		public static AbstractCriterion Sql(SqlString sql, object[] values, IType[] types)
+		public static SQLCriterion Sql(SqlString sql, object[] values, IType[] types)
```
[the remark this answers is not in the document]
bahusoid: I don't know... How about deprecating this class altogether and create `Restrictions.Sql` instead?

Considering class summary  `This class is semi-deprecated. Use <see cref="Restrictions"/>. ` might be a good idea..

src/NHibernate/Criterion/Projections.cs
bahusoid: Deprecate and add `Projections.Sql` instead?

src/NHibernate/Criterion/Expression.cs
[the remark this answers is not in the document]
bahusoid: Ok.. Might not be so good idea of deprecating full `Expression` class. At least not in scope of this                           PR. It's used quite extensively across code instead of recommended `Restrictions`. So how about deprecate only `Expression.Sql` methods?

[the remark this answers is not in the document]
fredericDelaporte: Most usages are in tests, only 6 in NHibernate. It sound a bit strange to deprecate all its methods but not the class itself.
But well, if you prefer to deprecate the class in another PR, why not.

src/NHibernate/Criterion/Projections.cs
fredericDelaporte: Ok. Maybe for consistency do the same with `SqlGroupProjection` and `GroupProjection`.

fredericDelaporte: Well, it may conflict with a `Group` method, but it seems to me to be a bit the same thing. I do not really make sense of these variations here...

src/NHibernate/Criterion/Expression.cs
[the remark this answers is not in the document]
bahusoid: It's actively used in documentation too. So I think it's better to keep this class be empty for legacy reasons..

src/NHibernate/Criterion/Projections.cs
bahusoid: Done ( I didn't touch `GroupProjection`)

src/NHibernate/Criterion/Expression.cs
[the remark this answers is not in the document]
bahusoid: Done

src/NHibernate/Criterion/ICriteriaQuery.cs
```diff
@@ -6,6 +6,15 @@
 
 namespace NHibernate.Criterion
 {
+	//TODO 6.0: Add to ICriteriaQuery interface
+	internal interface ICriteriaQueryNextVer
```
hazzik, line 10: nit: I don't like the name

src/NHibernate/Criterion/SQLCriterion.cs
```diff
@@ -59,5 +62,19 @@ public override string ToString()
 		{
 			return _sql.ToString();
 		}
+
+		/// <summary>
+		/// Provide list of criteria aliases that's used in SQL projection.
+		/// To be replaced with SQL aliases.
+		/// </summary>
+		public SQLCriterion AddCriteriaAliases(params string[] criteriaAliases)
```
hazzik, line 70: nit: `AddAliases`?

src/NHibernate/Criterion/SQLProjection.cs
```diff
@@ -37,15 +39,52 @@ internal SQLProjection(string sql, string groupBy, string[] columnAliases, IType
...
-			//return result;
-			return new SqlString(sql?.Replace("{alias}", criteriaQuery.GetSQLAlias(criteria)));
+			return GetSqlString(criteria, criteriaQuery, sql);
+		}
+
+		/// <summary>
+		/// Provide list of criteria aliases that's used in SQL projection.
+		/// To be replaced with SQL aliases.
+		/// </summary>
+		public SQLProjection AddCriteriaAliases(params string[] criteriaAliases)
```
hazzik, line 49: nit: `AddAliases`

src/NHibernate/Criterion/ICriteriaQuery.cs
bahusoid: It's internal and name explicitly suggests that it will be merged to main interface. What's not to like? )  I like it! Your suggestion?

bahusoid: I got rid of interface

src/NHibernate/Criterion/SQLCriterion.cs
```diff
@@ -59,5 +62,19 @@ public override string ToString()
 		{
 			return _sql.ToString();
 		}
+
+		/// <summary>
+		/// Provide list of criteria aliases that's used in SQL projection.
+		/// To be replaced with SQL aliases.
+		/// </summary>
+		public SQLCriterion AddAliases(params string[] criteriaAliases)
```
hazzik, line 70: After a close inspection of the changes I concluded that this method is unnecessary. All information has already presented in the query.

We have 2 ways forward:
- Extract aliases from SQLCriterion/SQLProjection with Regex; This approach requires just small adjustments, but I don't quite like it as it requires regex processing, which could be expensive. This can be slightly optimized if aliases are extracted on SQLCriterion/SQLProjection is construction.
- Get aliases from Criteria.

bahusoid: > Extract aliases from SQLCriterion/SQLProjection with Regex

I don't like any automatic parsing routines and intentionally made it manual. All automatic parsing logic should at least be optional - as in most cases it's not required. So I don't like your PR which now adds mandatory Replace calls for all criteria aliases in query.

And initially I wanted to support not only aliases but any projections in this template:
```C#
Projections.SQL(" {value1} + {x}.Value2")
.AddAliases("x")
.AddProjection("value1", Projections.Property("Value1"))
```

But projections support requires `SqlString.Replace(sqlString, sqlString)` which is missing. And I left implementation for later.

So I'm for manual explicit alias registration (similar how it's done in hql SQL - `ISQLQuery.AddEntity`). As convenience we can add method (like `ResolveAliases`) which would parse string via regex and populate `_criteriaAliases` but don't see necessity in it (user can implement it himself).

> Get aliases from Criteria.

`ICriteriaQuery` contains this data so why can't it be exposed? It feels just like proper place for retrieving such information. And I'm not sure I understand how it can be retrieved from `ICriteria`. Could you clarify?

hazzik: >And in some cases it's harmful (like when user wants `{alias}` be present in SQL projection without any transformation)
> So I'm for manual explicit alias registration
> PR which now adds mandatory Replace calls for all criteria aliases in query.

I don't understand why would anyone want to leave untranslated `{something}` in the SQL. So, I don't see a way when there is an alias in SQLCriterion/SQLProjection which would not be registered in `AddAliases(...)` call.

For me it feels more natural and convenient to translate all aliases rather than manually register them for translation.

If you are talking about performance here, it is negligible in my view as CriteriaTranslator already does some unnecessary heavy lifting "just in case".

> I'm for manual explicit alias registration (similar how it's done in hql SQL - `ISQLQuery.AddEntity`)

`AddEntity` is a little different than this. It add a necessary information to the query which it would not have otherwise (entity type for example). On the other hand, aliases are already known and readily available to use, so `AddAliases` not add real value.

> > Get aliases from Criteria.

> `ICriteriaQuery `contains this data so why can't it be exposed? It feels just like proper place for retrieving such information. 

This is exactly what I meant and how I implemented it. 

> And I'm not sure I understand how it can be retrieved from `ICriteria`. Could you clarify?

I did not say `ICriteria`, I meant somewhere in the Criteria. There are aliases in `ICriteria`, but they would need to be passed to the `ICriteriaQuery` to get SQLAliases. I tried to implement it this way, but was more complex.

> And initially I wanted to support not only aliases but any projections in this template

Nothing in your PR or my addition stops you to implement this, and the solution would be orthogonal to the resolving aliases. The only similarity would be that it would be happening at the same place.

hazzik: I think we would need to invoke @fredericDelaporte so he can express his view and resolve the arguments.

bahusoid: > I don't understand why would anyone want to leave untranslated {something} 

Because it's part of some unrelated string in this projection. 

> Nothing in your PR or my addition stops you to implement this

Mostly user would use either projection or criteria alias but not both - so unnecessary additional work. And it could clash with user aliases for projections. With explicit registering it's less probable.

> In the other hand, aliases are already known and readily available to use, so AddAliases not add real value

Explicity is added value for me.

> If you are talking about performance here, ... CriteriaQueryTranslator already does some unnecessary heavy lifting "just in case"

That's a not good point for adding additionally unnecessary heavy lifting. We should minimize it instead. 
And yes that's my main concern - I really don't like that for each SQL projection/criterion we generate dictionary with existing aliases, then generate possible aliases to replace in template and call Replace for them.

In any way your approach is different implementation so if @fredericDelaporte also think it's better just create own PR and I will close mine.

hazzik:
Let's probably split this PR into 2 parts: 
1) obsolete methods returning `AbstractCriterion` and addition of methods returining `SQLCriterion`
2) implementation of #896

Because we don't have disagreement on the first part.

src/NHibernate/Criterion/SQLCriterion.cs
hazzik: >it's better just create own PR and I will close mine.

Created

src/NHibernate/Criterion/Expression.cs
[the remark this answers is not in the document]
fredericDelaporte: We can still obsolete it, but we will likely have to keep it in next major version. Not every obsoleted members are meant to be dropped, for easing upgrades, especially those not costly to keep and potentially massively used. (That is also the version for telling in code since which version the member is obsoleted, helping to decide whether to drop it or not. Otherwise if we drop them all on each major version, the comment would not be useful.)

From [Microsoft guidelines](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes#binary-breaking-change):

>CONSIDER keeping types and methods with the `ObsoleteAttribute` indefinitely in low and middle-level libraries.

(They classify an ORM as a low or middle-level library.)

src/NHibernate/Criterion/ICriteriaQuery.cs
fredericDelaporte: Well, I thought also about "niting" here, but have let it due to being internal.

src/NHibernate/Criterion/SQLCriterion.cs
fredericDelaporte: As I am mostly a HQL and Linq "guy", having not used a lot Criteria, I do not think my view should be taken with a lot of weight.

This said, I am more in favor of Alexander viewpoint, at least about its rationales. But I have not thoroughly reviewed again this PR, and not yet Alexander's one either. I think I will do this tomorrow.

src/NHibernate/Criterion/ICriteriaQuery.cs
bahusoid: So what's wrong with 'NextVer' suffix? I just want to stop thinking how to name my next temp internal interface that should be merged. I thought it's a good plan to use some name convention for such interfaces. If  `ISomethingNextVer` is not good enough can you suggest something better?

hazzik: I intentionally prefixed my comment with [‘nit:’](https://stackoverflow.com/q/27810522)

fredericDelaporte: I understand the rationale here, and I do not have a better idea, that is was I had not written anything initially. I would not oppose this practice, I was just sharing a feeling.

src/NHibernate/Criterion/SQLProjection.cs
```diff
@@ -37,15 +39,48 @@ internal SQLProjection(string sql, string groupBy, string[] columnAliases, IType
...
-			//result.Replace(sql, "{alias}");
-			//return result;
-			return new SqlString(sql?.Replace("{alias}", criteriaQuery.GetSQLAlias(criteria)));
+			return GetSqlString(criteria, criteriaQuery, sql);
+		}
+
+		/// <summary>
+		/// Provide list of criteria aliases that's used in SQL projection.
+		/// To be replaced with SQL aliases.
+		/// </summary>
```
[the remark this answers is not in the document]
bahusoid: I don't not really understand where it's going. If other PR is preferred - I see no point in members obsoletion. All those obsolete members were driven by breaking changes introduced in this PR.  And why are you suggesting to comment members that are not present in other version?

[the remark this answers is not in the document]
fredericDelaporte: Because it is not yet the end of it. If this PR if finally chosen, public API meant to be used by user code should be fully commented.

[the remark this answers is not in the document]
fredericDelaporte: About member obsoletion, the subject still exists, since `Expression` is commented as "semi-obsolete". At some point, something should be done about it, don't you think?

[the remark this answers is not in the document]
bahusoid: Ok. But unless it's chosen I see no point on working on obosletion here. It's just easier to make those changes again - they won't merge nicely with other PR.

src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
```diff
@@ -1019,6 +1020,92 @@ public void ProjectionsTest()
...
+				t.Commit();
+			}
+
+			using (var s = OpenSession())
+			{
+				Student studentSubquery = null;
+				var subquery = QueryOver.Of(() => studentSubquery)
+				         .And(
+					         Restrictions.Sql("{e}.studentId = 667 and {studentSubquery}.studentId = 667")
+					                   .AddAliases("e", "studentSubquery")).Select(Projections.Id());
```
hazzik, line 1072: Will there be cases when user will want to `.AddAliases(...)` later? If not I think it is better to add `params[] aliases` parameter to the `Restrictions.Sql` method instead of introducing a new method.

src/NHibernate.Test/Criteria/CriteriaQueryTest.cs
```diff
@@ -1019,6 +1020,92 @@ public void ProjectionsTest()
...
+					                    Projections.Sql(
+						                    "{alias}.studentId as studentNumber, {e}.Semester as semester,"
+						                    + " {c}.CourseCode as courseCode, {c}.Description as descr",
+						                    new string[] {"studentNumber", "semester", "courseCode", "descr"},
+						                    new[]
+						                    {
+							                    TypeFactory.HeuristicType(typeof(long)),
+							                    TypeFactory.HeuristicType(typeof(short)),
+							                    TypeFactory.HeuristicType(typeof(string)),
+							                    TypeFactory.HeuristicType(typeof(string)),
```
hazzik, line 1092: A little side-track rant: I know, it is a copy-paste from other test, but `columns` and `types` have [Connascence of Position](https://en.wikipedia.org/wiki/Connascence) and CoV which is a sign of a bad design. (NHibernate has this across all code-base, but I think we should start progressing to remove that.

bahusoid: Yeah I agree we should fight it. Though not sure about this place.. This is dictated by method signature.  No? To avoid it we need to provide method where all data is supplied in single parameter. Such wrapping in tests seems excessive...

hazzik: But this is not only in tests. The test reflect real-like use cases.

PR #2365, opened 2020-04-27
Add Linq parameter type detection

maca88:
This is a continuation of #2335 where the code was extracted from #2079. The main purpose of this PR is to gain the ability to have parameter types calculated before rewriting the query model, which we can use when building the HQL tree. Today parameter types are calculated by HQL logic which is unable to correctly calculate types is certain cases (Example: mod function discussed [here](https://github.com/nhibernate/nhibernate-core/pull/2079#discussion_r330745052)).


src/NHibernate/Linq/Visitors/ExpressionKeyVisitor.cs
```diff
@@ -86,49 +112,70 @@ protected override Expression VisitConstant(ConstantExpression expression)
...
+			{
+				_string.Append("{");
+				_string.Append(string.Join(",", enumerable.Cast<object>()));
+				_string.Append("}");
+				return;
+			}
+
+			// When MappedAs is used we have to put all sql types information in the key in order to
+			// distinct when different precisions/sizes are used.
+			if (_sessionFactory != null && value is IType type)
```
maca88, line 143: With this PR `ExpressionKeyVisitor` generates different keys when `MappedAs` type is changed in order to avoid using the same query plan as now parameter types are always retrieved from the `ParameterMetadata`.

src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -125,11 +116,14 @@ protected override Expression VisitConstant(ConstantExpression expression)
...
+				// TODO 6.0: Remove
 				if (expression.Value == null)
 					type = NHibernateUtil.GuessType(expression.Type);
 
 				// Constant characters should be sent as strings
-				if (expression.Type == typeof(char))
+				// TODO 6.0: Remove
+				if (_queryVariables == null && expression.Type == typeof(char))
 				{
 					value = value.ToString();
```
maca88, line 128: This to me looks as a workaroud to fix `CharIndexFunction` and `CharIndexOffsetNegativeFunction` tests on Sql Server and Sql CE databases. Instead of converting it to string, this PR set the correct parameter size for drivers that fail without it.

src/NHibernate/Util/ParameterHelper.cs
```diff
@@ -0,0 +1,166 @@
+﻿using System;
+using System.Collections;
+using System.Linq;
+using NHibernate.Engine;
+using NHibernate.Proxy;
+using NHibernate.Type;
+
+namespace NHibernate.Util
+{
+	internal static class ParameterHelper
```
maca88, line 10: The methods were extracted from `AbstractQueryImpl` in order to be used by `ConstantTypeLocator`.

src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -59,22 +53,19 @@ public ExpressionParameterVisitor(PreTransformationResult preTransformationResul
 			return visitor._parameters;
 		}
 
-		public static Expression Visit(
-			PreTransformationResult preTransformationResult,
-			out IDictionary<ConstantExpression, NamedParameter> parameters)
+		public static IDictionary<ConstantExpression, NamedParameter> Visit(PreTransformationResult preTransformationResult)
```
maca88, line 56: Not a breaking change as the method is not yet released.

fredericDelaporte:
There is a failing new test.

src/NHibernate/Driver/SqlClientDriver.cs
```diff
@@ -283,6 +287,18 @@ protected static bool IsBlob(DbParameter dbParam, SqlType sqlType)
...
 
+		/// <summary>
+		/// Interprets if a parameter is a character (for the purposes of setting its default size)
+		/// </summary>
+		/// <param name="dbParam">The parameter</param>
+		/// <param name="sqlType">The <see cref="SqlType" /> of the parameter</param>
+		/// <returns>True, if the parameter should be interpreted as a character, otherwise False</returns>
+		protected static bool IsChar(DbParameter dbParam, SqlType sqlType)
+		{
+			return (DbType.StringFixedLength == dbParam.DbType || DbType.StringFixedLength == dbParam.DbType) &&
```
fredericDelaporte, line 298: I guess one of these should be `Ansi` prefixed, shouldn't it?

src/NHibernate.Test/Linq/ParameterTypeLocatorTests.cs
```diff
@@ -0,0 +1,427 @@
...
+					{"2", o => o is EnumStoredAsStringType},
+					{"\"test\"", o => o is AnsiStringType},
+				},
+				db.Users.Where(o => (o.Name == "test" ? o.NotMappedUser : o).Enum1 == EnumStoredAsString.Medium),
+				db.Users.Where(o => EnumStoredAsString.Medium == (o.Name == "test" ? o : o.NotMappedUser).Enum1)
+			);
+		}
+
+		[Test]
+		public void DynamicMemberTest()
```
fredericDelaporte, line 231: This test is failing.

```
Invalid type, actual type: p1
  Expected: True
  But was:  False

   at NHibernate.Test.Linq.ParameterTypeLocatorTests.AssertResult(Dictionary`2 expectedResults, QueryMode queryMode, Expression expression, Type targetType) in D:\BuildAgent\work\30546188361a242\src\NHibernate.Test\Linq\ParameterTypeLocatorTests.cs:line 418
   at NHibernate.Test.Linq.ParameterTypeLocatorTests.AssertResult(Dictionary`2 expectedResults, IQueryable query) in D:\BuildAgent\work\30546188361a242\src\NHibernate.Test\Linq\ParameterTypeLocatorTests.cs:line 369
   at NHibernate.Test.Linq.ParameterTypeLocatorTests.AssertResults(Dictionary`2 expectedResults, IQueryable[] queries) in D:\BuildAgent\work\30546188361a242\src\NHibernate.Test\Linq\ParameterTypeLocatorTests.cs:line 361
   at NHibernate.Test.Linq.ParameterTypeLocatorTests.DynamicMemberTest() in D:\BuildAgent\work\30546188361a242\src\NHibernate.Test\Linq\ParameterTypeLocatorTests.cs:line 233
```

And it seems the assert message does surface the expected information, yielding the parameter name instead of its type.

src/NHibernate/Driver/OdbcDriver.cs
```diff
@@ -78,10 +78,18 @@ private void SetVariableLengthParameterSize(DbParameter dbParam, SqlType sqlType
...
 				{
-					case DbType.AnsiString:
+					case DbType.StringFixedLength:
 					case DbType.AnsiStringFixedLength:
+						// For types that are using one character (CharType, AnsiCharType, TrueFalseType, YesNoType and EnumCharType),
+						// we have to specify the length otherwise sql function like charindex won't work as expected.
+						if (sqlType.Length == 1)
+						{
+							dbParam.Size = sqlType.Length;
+						}
```
fredericDelaporte, line 88: This may be SQL-Server specific, and maybe could create issues for people using ODBC to access another kind of database. Well, I guess no, but I am not sure.

Unless there is a way to detect the target database, maybe we need to introduce a setting for allowing to enable or disable length setting logic in the Odbc driver.

src/NHibernate/Hql/Ast/ANTLR/Tree/BetweenOperatorNode.cs
```diff
@@ -63,25 +63,27 @@ private IASTNode GetHighOperand()
 
 		private static void Check(IASTNode check, IASTNode first, IASTNode second)
 		{
-			var expectedTypeAwareNode = check as IExpectedTypeAwareNode;
-			if (expectedTypeAwareNode != null)
+			if (!(check is IExpectedTypeAwareNode expectedTypeAwareNode) ||
+			    expectedTypeAwareNode.ExpectedType != null)
```
fredericDelaporte, line 67: So the actual change here is the `ExpectedType != null` check, the other changes are code cleanup.

src/NHibernate/Hql/Ast/ANTLR/Tree/BinaryArithmeticOperatorNode.cs
```diff
@@ -32,50 +32,50 @@ public void Initialize()
 			IType lhType = (lhs is SqlNode) ? ((SqlNode)lhs).DataType : null;
 			IType rhType = (rhs is SqlNode) ? ((SqlNode)rhs).DataType : null;
 
-			if (lhs is IExpectedTypeAwareNode && rhType != null)
+			TrySetExpectedType(lhs, rhType, true);
+			TrySetExpectedType(rhs, lhType, false);
```
fredericDelaporte, line 36: The previous code would have set rhs expected type only if lhs was not a type aware node or rhType was null.
Was this a bug fixed by the way? Or is this an overlooked change?

gliljas:
Interesting. While developing the NodaTime integration I came upon a case where a type (DatePeriod) implements IEnumerable. It works just fine in terms of persistence, but when querying, a DatePeriod parameter will be added using SetParameterList, which falls apart completely. It's not a list. It's a composite type. The "implements IEnumerable but is not a string" check is just a bit too naïve, but given its context, it's the best that can be done. 

Will this PR or its siblings open up a way for this to be remedied?

src/NHibernate/Driver/SqlClientDriver.cs
maca88: Yes, corrected.

src/NHibernate.Test/Linq/ParameterTypeLocatorTests.cs
maca88: > This test is failing.

I realized that visiting a dynamic expression in .NET Framework does not reduce it where on .NET Core it does, unless `DynamicExpressionVisitor` is used.

> And it seems the assert message does surface the expected information

Fixed.

src/NHibernate/Driver/OdbcDriver.cs
maca88: From what I see before NH-4083 `Size` was always set and except NH-4083 I didn't found other issues related to odbc drivers and parameters. So I think we should be fine by setting the size only for char types without adding an additional setting.

src/NHibernate/Hql/Ast/ANTLR/Tree/BetweenOperatorNode.cs
maca88: Correct.

src/NHibernate/Hql/Ast/ANTLR/Tree/BinaryArithmeticOperatorNode.cs
maca88: Only a `ExpectedType != null` check was added and the `rhs` check was modified to check for `IExpectedTypeAwareNode` instead `ParameterNode`, other changes are code cleanup. 
`ParameterNode` implements `IExpectedTypeAwareNode` and checking for `ParameterNode` instead of `IExpectedTypeAwareNode` would ommit `BooleanLiteralNode`, which doesn't seem right to me. If it is copied for `lhs` it should be also copied for `rhs`. So it was intentional, but I forgot to mention it.

maca88:
> Will this PR or its siblings open up a way for this to be remedied?

Originally not, but with the last commit it will no longer call `SetParameterList` for such types. `SetParameterList` method is only used for `IN` operator, which will only be generated when using `Contains` method in a linq expression. So instead of checking "is IEnumerable but is not a string" now checks for `Contains` method.

maca88:
Rebased.

hazzik:
Please hold off merging till weekend. I want to review this.

src/NHibernate/Hql/Ast/ANTLR/Tree/BetweenOperatorNode.cs
```diff
@@ -63,25 +63,27 @@ private IASTNode GetHighOperand()
...
-					if (secondNode != null)
-					{
-						expectedType = secondNode.DataType;
-					}
+					expectedType = secondNode.DataType;
 				}
-				expectedTypeAwareNode.ExpectedType = expectedType;
 			}
+
+			expectedTypeAwareNode.ExpectedType = expectedType;
```
hazzik, line 86: This is probably readable:

```suggestion
			expectedTypeAwareNode.ExpectedType = (first as SqlNode)?.DataType ?? (second as SqlNode)?.DataType;
```

gliljas:
By pure coincidence, I'm creating something very similar in my NodaTime project (required to prevent ambiguous Linq-to-Hql processes). I wouldn't mind using this instead, but it's internal and I may need to avoid taking the shortcut of just returning the first candidate type when there's a coalesce or a conditional.

src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -107,6 +102,17 @@ protected override Expression VisitMethodCall(MethodCallExpression expression)
 				return Expression.Call(null, expression.Method, query, arg);
 			}
 
+			if (_functionRegistry != null &&
+				_functionRegistry.TryGetGenerator(method, out var generator) &&
+				generator is CollectionContainsGenerator)
```
hazzik, line 107: What about CollectionContainsRuntimeHqlGenerator?

hazzik: Ok, I've understood how it works. CollectionContainsRuntimeHqlGenerator returns CollectionContainsGenerator for a mehtod. So this code works.

src/NHibernate/Param/NamedParameter.cs
```diff
@@ -15,6 +15,8 @@ public NamedParameter(string name, object value, IType type)
 		public object Value { get; internal set; }
 		public IType Type { get; internal set; }
 
+		public virtual bool IsCollection => false;
```
hazzik, line 18: Probably just create a constructor to set this value instead of a new type.

src/NHibernate/Hql/Ast/ANTLR/Tree/BetweenOperatorNode.cs
maca88: Changed, as it is readable to me. I couldn't commit the suggestion due to `Applying suggestions on deleted lines is not supported`.

src/NHibernate/Param/NamedParameter.cs
maca88: Done.

maca88:
> I wouldn't mind using this instead, but it's internal and I may need to avoid taking the shortcut of just returning the first candidate type when there's a coalesce or a conditional.

I am not sure if we should make `ExpressionsHelper.TryGetMappedType` public as I think that in the future we should get rid of it and instead introduce custom expressions like `NhPropertyExpression`, `NhCaseExpression`... that would have the information of the mapped type and persister/component type. These expressions would replace expressions like `MemberExpression` and `ConditionalExpression` and be available to all rewriters including the one provided by the `IQueryModelRewriterFactory` interface.

maca88:
Rebased due to conflict with #2387.

gliljas:
I couldn't agree more. Something that carries mapping metadata within the expression tree (which doesn't have to be Linq based. It could cause confusion). Hibernate is working on the Semantic Query Model approach. I'm not sure how that will turn out, but the idea seems right. Preferably, the SQL would also be expressed as a visitable tree.

src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -107,6 +102,17 @@ protected override Expression VisitMethodCall(MethodCallExpression expression)
 				return Expression.Call(null, expression.Method, query, arg);
 			}
 
+			if (_functionRegistry != null &&
+				_functionRegistry.TryGetGenerator(method, out var generator) &&
```
hazzik, line 106: This is potentially expensive operation. Could it be replaced with something lighter?

maca88: Added an additional check for method name in order to avoid calling `TryGetGenerator` too often.

hazzik: Thanks.

src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -18,23 +20,18 @@ public class ExpressionParameterVisitor : RelinqExpressionVisitor
...
-			ReflectHelper.FastGetMethodDefinition(Enumerable.Skip, default(IEnumerable<object>), 0);
-		private static readonly MethodInfo EnumerableTakeDefinition =
-			ReflectHelper.FastGetMethodDefinition(Enumerable.Take, default(IEnumerable<object>), 0);
-
-		private readonly ICollection<MethodBase> _pagingMethods = new HashSet<MethodBase>
-			{
-				QueryableSkipDefinition, QueryableTakeDefinition,
-				EnumerableSkipDefinition, EnumerableTakeDefinition
-			};
+		private static readonly ISet<MethodBase> PagingMethods = new HashSet<MethodBase>
```
hazzik, line 28:
```suggestion
		private static readonly HashSet<MethodBase> PagingMethods = new HashSet<MethodBase>
```

src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -107,6 +102,18 @@ protected override Expression VisitMethodCall(MethodCallExpression expression)
 				return Expression.Call(null, expression.Method, query, arg);
 			}
 
+			if (_functionRegistry != null &&
```
hazzik, line 105: While we do not have other functions that accept collection arguments it would be nice if we can generalize this logic. I was trying to implement support for `string.Concat(...)` which has an overload with `params`. I had stuck on the over-caching issue for different parameter types.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -226,18 +226,14 @@ private HqlTreeNode VisitNhNominated(NhNominatedExpression nhNominatedExpression
 
 		private HqlTreeNode VisitInvocationExpression(InvocationExpression expression)
 		{
-			//This is an ugly workaround for dynamic expressions.
-			//Unfortunately we can not tap into the expression tree earlier to intercept the dynamic expression
-			if (expression.Arguments.Count == 2 &&
-			    expression.Arguments[0] is ConstantExpression constant &&
-			    constant.Value is CallSite site &&
-			    site.Binder is GetMemberBinder binder)
+#if NETCOREAPP2_0
```
hazzik, line 229: Why NETCOREAPP2_0 only?

src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
maca88: I agree and added `TryGetCollectionParameter` method to `IHqlGeneratorForMethodExtended` interface. Unfortunately the method name check optimization is now gone, but given that `DefaultLinqToHqlGeneratorsRegistry.TryGetGenerator` caches the result, I think it should be fine.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
maca88: Because in .NET Framework a dynamic expression [will never be reduced](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/DynamicExpression.cs#L147) with an `ExpressionVisitor`, where in .NET Core [it will be reduced](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/DynamicExpression.cs#L149) when the visitor is not a `DynamicExpressionVisitor`.

src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -161,6 +180,15 @@ protected override Expression VisitConstant(ConstantExpression expression)
 			return base.VisitConstant(expression);
 		}
 
+		private NamedParameter CreateParameter(ConstantExpression expression, object value, IType type)
+		{
+			return new NamedParameter(
+				"p" + (_parameters.Count + 1),
+				value,
+				type,
+				_collectionParameters.Contains(expression));
```
hazzik, line 189: I don't really understand. Why we cannot simply use the same check as it was in ExpressionKeyVisitor.VisitParameter?

maca88: In order to support types that implements `IEnumerable`, which should not be treated as collections. One example is the NodaTime integration that @gliljas is working on (see [above comment](https://github.com/nhibernate/nhibernate-core/pull/2365#issuecomment-626370794)).

fredericDelaporte:
Rebased for resolving a conflict.

PR #2367, opened 2020-04-27
Replace generic LinkedHashMap with Dictionary where applicable

bahusoid:
The only place where it seems matters is [`BatchFetchQueue`](https://github.com/nhibernate/nhibernate-core/blob/cc6baef36425e5f4a72689e3281560955835d68a/src/NHibernate/Engine/BatchFetchQueue.cs#L34) but not 100% sure.

### DIFF
--- a/src/NHibernate/Mapping/Table.cs
+++ b/src/NHibernate/Mapping/Table.cs
@@ -27,7 +27,7 @@ namespace NHibernate.Mapping
 	public class Table : IRelationalModel
 	{
 		private readonly List<string> checkConstraints = new List<string>();
-		private readonly LinkedHashMap<string, Column> columns = new LinkedHashMap<string, Column>();
+		private readonly Dictionary<string, Column> columns = new Dictionary<string, Column>();
 		private readonly Dictionary<ForeignKeyKey, ForeignKey> foreignKeys = new Dictionary<ForeignKeyKey, ForeignKey>();
 		private readonly Dictionary<string, Index> indexes = new Dictionary<string, Index>();
 		private int? uniqueInteger;
--- a/src/NHibernate/SqlCommand/SqlInsertBuilder.cs
+++ b/src/NHibernate/SqlCommand/SqlInsertBuilder.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using NHibernate.Engine;
 using NHibernate.SqlTypes;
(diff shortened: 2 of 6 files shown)


src/NHibernate/Mapping/Table.cs
```diff
@@ -27,7 +27,7 @@ public enum SchemaAction
 	public class Table : IRelationalModel
 	{
 		private readonly List<string> checkConstraints = new List<string>();
-		private readonly LinkedHashMap<string, Column> columns = new LinkedHashMap<string, Column>();
+		private readonly Dictionary<string, Column> columns = new Dictionary<string, Column>();
```
hazzik, line 30: I think it does make sense to have a linked hash map here: it ensures the order of elements. And the application does depend on the order of columns.

src/NHibernate/SqlCommand/SqlInsertBuilder.cs
```diff
@@ -19,7 +20,7 @@ public class SqlInsertBuilder : ISqlStringBuilder
 		private string comment;
 
 		// columns-> (ColumnName, Value) or (ColumnName, SqlType) for parametrized column
-		private readonly LinkedHashMap<string, object> columns = new LinkedHashMap<string, object>();
+		private readonly Dictionary<string, object> columns = new Dictionary<string, object>();
```
hazzik, line 23: Again, order of columns is important.

src/NHibernate/SqlCommand/SqlUpdateBuilder.cs
```diff
@@ -19,7 +19,7 @@ public class SqlUpdateBuilder : SqlBaseBuilder, ISqlStringBuilder
 		private string comment;
 
 		// columns-> (ColumnName, Value) or (ColumnName, SqlType) for parametrized column
-		private readonly LinkedHashMap<string, object> columns = new LinkedHashMap<string, object>();
+		private readonly Dictionary<string, object> columns = new Dictionary<string, object>();
```
hazzik, line 22: Order of columns is important here.

hazzik:
Order of columns in a table definitely does matter. In the insert/update sql builders it might matter, depending on how the parameters get bound to the command.

bahusoid:
Yeah. But in Dictionary order is the same too as long elements from the dictionary are not deleted. It's a known behavior too so I see no problem here

hazzik:
It is confirmed, but not guaranteed. I feel like it's better to be safe than sorry.

bahusoid:
> but not guaranteed

I don't believe it would ever change and it's better to avoid using some custom baked classes. But Ok my belief is not very strong argument :)

fredericDelaporte:
I am agree it is better to not rely on this long known behavior of dictionary, even though I have already relied on it in some circumstances in other projects. We already have the code for avoiding relying on it in NHibernate, better keep it that way.
We can not be sure all implementations of the standard will always have this behavior.

PR #2368, opened 2020-04-27
Replace SafetyEnumerable<T> with OfType<T> where applicable

bahusoid:
Obsolete `SafetyEnumerable<T>` by replacing with `OfType<T>` extension.
Exception  [`HQLQueryPlan`](https://github.com/nhibernate/nhibernate-core/blob/b25f9227249a34663bee56269ddf0946306184a2/src/NHibernate/Engine/Query/HQLQueryPlan.cs#L179)  where `Cast` is applied (as more appropriate behavior in this context)

### DIFF
--- a/src/NHibernate/Mapping/Constraint.cs
+++ b/src/NHibernate/Mapping/Constraint.cs
@@ -91,9 +91,13 @@ namespace NHibernate.Mapping
 		{
 			foreach (Column col in columnIterator)
 			{
-				if (!col.IsFormula)
-					AddColumn(col);
+				AddColumn(col);
 			}
+		}
+
+		public void AddColumns(IEnumerable<ISelectable> columnIterator)
+		{
+			AddColumns(columnIterator.OfType<Column>());
 		}
 
 		/// <summary>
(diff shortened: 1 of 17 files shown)


src/NHibernate/Mapping/Constraint.cs
```diff
@@ -91,11 +91,15 @@ public void AddColumns(IEnumerable<Column> columnIterator)
 		{
 			foreach (Column col in columnIterator)
 			{
-				if (!col.IsFormula)
```
bahusoid, line 94: `IsFormula` is not virtual and always false for `Column`

bahusoid:
> This usage looks to me as a bug "hider", silently ignoring invalid result rows (not having the expected type). Shouldn't it be a Cast<T> here instead?

Yeah.. Indeed.. Only with special null handling for structs. Something like:
`Select( obj => obj  == null ? (default(T)) : (T)obj);` 
But  I think such fix better to do in separate PR as possible breaking change.

>But maybe that is more a trouble of missing null checks where we add things to these collections.

I would say yes to this. And regarding columns handling - null values didn't flow nicely to consumer code before. They stopped pretty quickly in `AddColumns`: https://github.com/nhibernate/nhibernate-core/blob/2f8f31772e176e54dff55b896a8c93d0429a2999/src/NHibernate/Mapping/Constraint.cs#L94

bahusoid:
>This usage looks to me as a bug "hider", silently ignoring invalid result rows (not having the expected type). Shouldn't it be a Cast<T> here instead?

After another thought I made this change in this PR.

PR #2377, opened 2020-05-11
Use dotnet to push packages to nuget

hazzik:


default.build
```diff
@@ -313,11 +312,9 @@
 		
 	</target>
 	
-	<target name="nugetpushbat" depends="init binaries common.download-nuget nuget.set-properties"
+	<target name="nugetpushbat" depends="init binaries nuget.set-properties"
 					description="Creates files for the release on nuget gallery.">
 
-		<copy file="${tools.dir}/NuGet.exe" todir="${nuget.nupackages.dir}"/>
-
 		<echo message="rem In order to use this bat you have to be sure you have executed 'nuget SetApiKey' ${environment::newline()}" file="${nuget.nupackages.pushbatfile}" append="false"/>
```
bahusoid, line 318: This comment  'nuget SetApiKey'  should also be updated accordingly

hazzik: There is no [`dotnet nuget SetApiKey`](https://github.com/NuGet/Home/issues/6437). 

> As a workaround you could use nuget.exe setApiKey (or set the values manually in your nuget.config and then use dotnet nuget push. dotnet.exe should read the nuget.config and use the api keys found there.

hazzik: Feel free to change as you see fit.

bahusoid: It seems workaround doesn't work for [everyone ](https://github.com/NuGet/Home/issues/6437#issuecomment-436806547) (linux specific issue?)

Well have nothing to add - if you see no possible issues. Then fine. May be just add link to nuget.exe (https://dist.nuget.org/win-x86-commandline/latest/nuget.exe) in this comment

fredericDelaporte: I do not think we need to explain more. Anyway, only most active contributors do the push currently. I guess we will all find what to do in case of troubles.

PR #2378, opened 2020-05-14
Add missing ISession.Get(entityName, id, lockMode)

### LINKED ISSUE #2313: Add missing overloads to ISession.Get methods
`ISession.Get` methods have no overloads taking both an `entityName` and a `lockMode`, which, as a result, forces to lock in a separated query when getting an entity by entity name.

In high concurrency environments, it may cause issues.

Generic `Get<T>` methods also lack overloads with an `entityName` argument, which are required when the class is not the entity name.

bahusoid:
Fixes #2313 

`IStatelessSession.Get(entityName, id, lockMode)` is already present

### DIFF
--- a/src/NHibernate/ISession.cs
+++ b/src/NHibernate/ISession.cs
@@ -15,7 +15,7 @@ using NHibernate.Util;
 namespace NHibernate
 {
 	// 6.0 TODO: Convert most of these extensions to interface methods
-	public static class SessionExtensions
+	public static partial class SessionExtensions
 	{
 		/// <summary>
 		/// Obtain a <see cref="IStatelessSession"/> builder with the ability to grab certain information from
@@ -48,6 +48,30 @@ namespace NHibernate
 		/// <returns>The current transaction or <see langword="null" />..</returns>
 		public static ITransaction GetCurrentTransaction(this ISession session)
 			=> session.GetSessionImplementation().ConnectionManager.CurrentTransaction;
+
+		/// <summary>
+		/// Return the persistent instance of the given entity class with the given identifier, or null
+		/// if there is no such persistent instance. (If the instance, or a proxy for the instance, is
+		/// already associated with the session, return that instance or proxy.)
+		/// </summary>
+		public static object Get(this ISession session, string entityName, object id, LockMode lockMode)
+		{
+			return
+				ReflectHelper
+					.CastOrThrow<SessionImpl>(session, "Get with entityName and lockMode")
+					.Get(entityName, id, lockMode);
+		}
+
+		//NOTE: Keep it as extension
+		/// <summary>
+		/// Return the persistent instance of the given entity name with the given identifier, or null
+		/// if there is no such persistent instance. (If the instance, or a proxy for the instance, is
+		/// already associated with the session, return that instance or proxy.)
+		/// </summary>
+		public static T Get<T>(this ISession session, string entityName, object id, LockMode lockMode = null)
+		{
+			return (T) session.Get(entityName, id, lockMode);
+		}
 	}
 
 	/// <summary>
--- a/src/NHibernate/ISession.cs
+++ b/src/NHibernate/ISession.cs
@@ -54,6 +54,11 @@ namespace NHibernate
 		/// if there is no such persistent instance. (If the instance, or a proxy for the instance, is
 		/// already associated with the session, return that instance or proxy.)
 		/// </summary>
+		/// <param name="session">The session.</param>
+		/// <param name="entityName">The entity name.</param>
+		/// <param name="id">The entity identifier.</param>
+		/// <param name="lockMode">The lock mode to use for getting the entity.</param>
+		/// <returns>A persistent instance, or <see langword="null" />.</returns>
 		public static object Get(this ISession session, string entityName, object id, LockMode lockMode)
 		{
 			return
@@ -68,6 +73,12 @@ namespace NHibernate
 		/// if there is no such persistent instance. (If the instance, or a proxy for the instance, is
 		/// already associated with the session, return that instance or proxy.)
 		/// </summary>
+		/// <typeparam name="T">The entity class.</typeparam>
+		/// <param name="session">The session.</param>
+		/// <param name="entityName">The entity name.</param>
+		/// <param name="id">The entity identifier.</param>
+		/// <param name="lockMode">The lock mode to use for getting the entity.</param>
+		/// <returns>A persistent instance, or <see langword="null" />.</returns>
 		public static T Get<T>(this ISession session, string entityName, object id, LockMode lockMode = null)
 		{
 			return (T) session.Get(entityName, id, lockMode);
--- a/src/NHibernate/ISession.cs
+++ b/src/NHibernate/ISession.cs
@@ -79,9 +79,25 @@ namespace NHibernate
 		/// <param name="id">The entity identifier.</param>
 		/// <param name="lockMode">The lock mode to use for getting the entity.</param>
 		/// <returns>A persistent instance, or <see langword="null" />.</returns>
-		public static T Get<T>(this ISession session, string entityName, object id, LockMode lockMode = null)
+		public static T Get<T>(this ISession session, string entityName, object id, LockMode lockMode)
 		{
 			return (T) session.Get(entityName, id, lockMode);
+		}
+
+		//NOTE: Keep it as extension
+		/// <summary>
+		/// Return the persistent instance of the given entity name with the given identifier, or null
+		/// if there is no such persistent instance. (If the instance, or a proxy for the instance, is
+		/// already associated with the session, return that instance or proxy.)
+		/// </summary>
+		/// <typeparam name="T">The entity class.</typeparam>
+		/// <param name="session">The session.</param>
+		/// <param name="entityName">The entity name.</param>
+		/// <param name="id">The entity identifier.</param>
+		/// <returns>A persistent instance, or <see langword="null" />.</returns>
+		public static T Get<T>(this ISession session, string entityName, object id)
+		{
+			return (T) session.Get(entityName, id);
 		}
 	}
 
--- a/src/NHibernate/IStatelessSession.cs
+++ b/src/NHibernate/IStatelessSession.cs
@@ -11,7 +11,7 @@ using NHibernate.Util;
 namespace NHibernate
 {
 	// 6.0 TODO: Convert most of these extensions to interface methods
-	public static class StatelessSessionExtensions
+	public static partial class StatelessSessionExtensions
 	{
 		/// <summary>
 		/// Creates a <see cref="IQueryBatch"/> for the session.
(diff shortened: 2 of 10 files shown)


src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1189,12 +1183,31 @@ public object Get(System.Type entityClass, object id)
...
 		public object Get(System.Type clazz, object id, LockMode lockMode)
+		{
+			return Get(clazz.FullName, id, lockMode);
+		}
+
+		/// <summary>
+		/// Return the persistent instance of the given entity name with the given identifier, or null
+		/// if there is no such persistent instance. (If the instance, or a proxy for the instance, is
+		/// already associated with the session, return that instance or proxy.)
+		/// </summary>
```
fredericDelaporte, line 1194:
```suggestion
		/// <inheritdoc />
```

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1189,12 +1183,31 @@ public object Get(System.Type entityClass, object id)
...
+				try
+				{
+					FireLoad(loadEvent, LoadEventListener.Get);
+					success = true;
+					return loadEvent.Result;
+				}
+				finally
+				{
+					AfterOperation(success);
+				}
```
fredericDelaporte, line 1210: This turns the PR into a non trivial bug fix by the way, causing the connection to be released after each Get provided with a lock mode in default release mode, if calling these Get(s) outside of a transaction. (Not having the connection released is a bug.) I think it should be mentioned in the PR description.

src/NHibernate/ISession.cs
```diff
@@ -48,6 +48,30 @@ public static IQueryBatch CreateQueryBatch(this ISession session)
...
+					.Get(entityName, id, lockMode);
+		}
+
+		//NOTE: Keep it as extension
+		/// <summary>
+		/// Return the persistent instance of the given entity name with the given identifier, or null
+		/// if there is no such persistent instance. (If the instance, or a proxy for the instance, is
+		/// already associated with the session, return that instance or proxy.)
+		/// </summary>
+		public static T Get<T>(this ISession session, string entityName, object id, LockMode lockMode = null)
```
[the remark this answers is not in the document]
bahusoid: But it's not consistent with session implementation. It's null provided there and decision is made inside `LoadEvent` (`LoadEvent.DefaultLockMode` is used).  I actually have no preference of how it should be implemented (well null is less code :) ) - it's just need to be consistent then with session implementation.

src/NHibernate/Impl/SessionImpl.cs
bahusoid: Can you update the PR description the way you think is better?

src/NHibernate/Impl/SessionImpl.cs
bahusoid: Hm.. I wonder if it would cause xml doc error as it's not yet in interface.. Will see...

src/NHibernate/ISession.cs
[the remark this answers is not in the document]
bahusoid: So I would keep it for now null to keep default value in one place - `LoadEvent.DefaultLockMode`

src/NHibernate/Impl/SessionImpl.cs
fredericDelaporte: I do not think it will not cause an error, but will cause the member to have no documentation. I have overlooked it was of course not yet in the interface. Anyway as you wish here. I consider `Impl` namespace is not belonging to public APIs meant to be directly used by users.

src/NHibernate/ISession.cs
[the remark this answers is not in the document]
fredericDelaporte: Well, `null` is slightly less explicit as it leaves open the question of whether some default would then be applied or not, but we can consider it is still good enough.

src/NHibernate/Impl/SessionImpl.cs
fredericDelaporte: I have checked with a test, and in fact the `AfterStatement` logic is already closing the connection, even without this `AfterOperation` call. 

But it causes quite another change to have that `AfterOperation` call: it will release the lock if we are outside of a transaction.
So this change switches a somewhat working call to a non-working call. The lock will be released at the end of the Get.

It can be argued that acquiring a lock outside a transaction is anyway a mistake, especially since many other operations on the session would cause the `AfterOperation` to release the lock.
I think the only appropriate action here would be to throw when a lock is asked for (other than `None`) while outside of a transaction.
But it is a breaking change, and I think it should be the subject of another PR, probably for 6.0.

Here the change should be reverted because it is still a possible breaking change and unrelated to the PR scope. Likely a comment should be added to warn about this pitfall. I thing the lack of `AfterOperation` was done on purpose, but that is a debatable choice and at least a comment should explain why it is not there.
Something like `AfterOperation call is skipped to avoid releasing the lock when outside of a transaction.`.

But then, it is troublesome to have refactored `Get`s without a lock mode to use the `Get` with a lock mode, because they were not behaving the same and cannot behave the same regarding `AfterOperation`, unless accepting the possible breaking change this PR currently introduces.

bahusoid: Could you please clarify what it would break if we just skip `AfterOperation` if `lockMode` is null? As it's just silly that `session.Get<T>(id)` and `session.Get<T>(id, lockMode: null)` behaves differently

fredericDelaporte: It may still breaks users not using a transaction (at least not one NHibernate is aware of, like using a transaction scope with a non-scope aware transaction factory) for doing a get with an actual lock mode then using more get with null or None lock mode in order to still keep the previously acquired lock of the first get.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1156,48 +1156,62 @@ public object Load(System.Type entityClass, object id)
...
-		/// using "for update", if supported. A new key will be assigned to the object.
-		/// This should return an existing proxy where appropriate.
-		///
-		/// If the object does not exist in the database, null is returned.
-		/// </summary>
-		/// <param name="clazz"></param>
-		/// <param name="id"></param>
-		/// <param name="lockMode"></param>
-		/// <returns></returns>
+		/// <inheritdoc />
```
hazzik, line 1177: `<inheritdoc />` is not valid here because there is no parent.
```suggestion
		/// <inheritdoc cref="SessionExtensions.Get" />
```

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1156,48 +1156,62 @@ public object Load(System.Type entityClass, object id)
...
 		public object Get(System.Type clazz, object id, LockMode lockMode)
+		{
+			return Get(clazz.FullName, id, lockMode);
+		}
+
+		/// <inheritdoc />
+		public object Get(string entityName, object id, LockMode lockMode)
```
hazzik, line 1184: `<inheritdoc />` is not valid here because there is no parent.
```suggestion
		/// <inheritdoc cref="SessionExtensions.Get" />
```

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1156,48 +1156,62 @@ public object Load(System.Type entityClass, object id)
...
-		/// <param name="id"></param>
-		/// <param name="lockMode"></param>
-		/// <returns></returns>
+		/// <inheritdoc />
 		public object Get(System.Type clazz, object id, LockMode lockMode)
+		{
+			return Get(clazz.FullName, id, lockMode);
+		}
+
+		/// <inheritdoc />
```
hazzik, line 1183: `<inheritdoc />` is not valid here because there is no parent.
```suggestion
		/// <inheritdoc cref="SessionExtensions.Get" />
```

hazzik: I should probably go sleeping. It's my third attempt to put this comment. 🤦

src/NHibernate/IStatelessSession.cs
```diff
@@ -31,6 +31,23 @@ public static IQueryBatch CreateQueryBatch(this IStatelessSession session)
...
+		/// if there is no such persistent instance. (If the instance, or a proxy for the instance, is
+		/// already associated with the session, return that instance or proxy.)
+		/// </summary>
+		/// <typeparam name="T">The entity class.</typeparam>
+		/// <param name="session">The session.</param>
+		/// <param name="entityName">The entity name.</param>
+		/// <param name="id">The entity identifier.</param>
+		/// <param name="lockMode">The lock mode to use for getting the entity.</param>
+		/// <returns>A persistent instance, or <see langword="null" />.</returns>
+		public static T Get<T>(this IStatelessSession session, string entityName, object id, LockMode lockMode = null)
```
hazzik, line 47: Probably split this into 2 methods to be consistent with ISession.

src/NHibernate/Impl/SessionImpl.cs
bahusoid: > <inheritdoc /> is not valid here because there is no parent.

Yeah we know it's been discussed here: https://github.com/nhibernate/nhibernate-core/pull/2378#discussion_r426177054

As this class is not intended to be used by end user I think it's OK to have it as is. It will become valid when we merge this method to interface. With you suggestion it's still invalid as multiple Get extensions are present. And we need to add to TODO to remove `cref`.  So maybe leave it be...

src/NHibernate/Impl/StatelessSessionImpl.cs
```diff
@@ -539,40 +539,32 @@ public void Delete(string entityName, object entity)
...
-		/// Retrieve a entity, obtaining the specified lock mode.
+		/// Retrieve an entity, obtaining the specified lock mode.
 		/// </summary>
 		/// <returns> a detached entity instance </returns>
 		public object Get(string entityName, object id, LockMode lockMode)
 		{
 			using (BeginProcess())
 			{
-				object result = Factory.GetEntityPersister(entityName).Load(id, null, lockMode, this);
+				object result = Factory.GetEntityPersister(entityName).Load(id, null, lockMode ?? LockMode.None, this);
```
bahusoid, line 567: For consistency with `session.Get(entityName, id, lockMode)`

src/NHibernate/IStatelessSession.cs
bahusoid: Done

src/NHibernate/Impl/StatelessSessionImpl.cs
fredericDelaporte: Did you mean `Get(string entityName, object id)`? Otherwise I do not get it. Well I do not get it either, I see no consistency issue here.

bahusoid: For consistency with call `session.Get(entityName, id, null)` which uses `LockMode.None` by default. Stateless session for my surprise was throwing exception

fredericDelaporte: Ok.

PR #2381, opened 2020-05-15
Use optimized Dictionary.Remove(key, out value) in .NET Core 

bahusoid:


src/NHibernate/Engine/StatefulPersistenceContext.cs
```diff
@@ -447,8 +443,9 @@ public bool ContainsEntity(EntityKey key)
 		/// </summary>
 		public object RemoveEntity(EntityKey key)
 		{
-			object tempObject = entitiesByKey[key];
-			entitiesByKey.Remove(key);
+			if (!entitiesByKey.Remove(key, out var tempObject))
+				throw new KeyNotFoundException(key.ToString());
```
bahusoid, line 447: Java version doesn't throw here ([`HashMap.remove`](https://www.geeksforgeeks.org/hashmap-remove-method-in-java/) is [used](https://github.com/hibernate/hibernate-orm/blob/2bcb1b0a6d8600ba3a60eae6460dcb52bf5628e7/hibernate-core/src/main/java/org/hibernate/engine/internal/StatefulPersistenceContext.java#L418) which returns null instead). Maybe we shouldn't either?

hazzik: This is actually reported as [NH-3478](https://github.com/nhibernate/nhibernate-core/issues/1310) and [NH-1845](https://nhibernate.jira.com/browse/NH-1845)

PR #2382, opened 2020-05-16
Refactor debug logging in AbstractBatcher

bahusoid:
Minimize clutter of `Log.IsDebugEnabled` calls and get rid of typing logic in `GetReaderStopwatch`

### DIFF
--- a/src/NHibernate/AdoNet/AbstractBatcher.cs
+++ b/src/NHibernate/AdoNet/AbstractBatcher.cs
@@ -93,11 +93,7 @@ namespace NHibernate.AdoNet
 			}
 
 			var cmd = _factory.ConnectionProvider.Driver.GenerateCommand(type, sql, parameterTypes);
-			LogOpenPreparedCommand();
-			if (Log.IsDebugEnabled())
-			{
-				Log.Debug("Building an DbCommand object for the SqlString: {0}", sql);
-			}
+			LogOpenPreparedCommand(sql);
 			_commandsToClose.Add(cmd);
 			return cmd;
 		}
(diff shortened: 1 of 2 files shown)


src/NHibernate/AdoNet/AbstractBatcher.cs
```diff
@@ -232,7 +228,7 @@ public int ExecuteNonQuery(DbCommand cmd)
 			}
 			finally
 			{
-				if (Log.IsDebugEnabled() && duration != null)
```
bahusoid, line 235: `duration` is created only under Debug logging. I see no point in additional code cluttering with  `Log.IsDebugEnabled()`

src/NHibernate/AdoNet/AbstractBatcher.cs
```diff
@@ -242,37 +238,31 @@ public virtual DbDataReader ExecuteReader(DbCommand cmd)
...
+					? reader
+					: NHybridDataReader.Create(reader);
 			}
 			catch (Exception e)
 			{
 				e.Data["actual-sql-query"] = cmd.CommandText;
 				Log.Error(e, "Could not execute query: {0}", cmd.CommandText);
 				throw;
 			}
-			finally
```
bahusoid, line 259: Useless finally block. Moved logic to `LogOpenReader` (as in case exception `reader` is null)

src/NHibernate/AdoNet/AbstractBatcher.cs
```diff
@@ -242,37 +238,31 @@ public virtual DbDataReader ExecuteReader(DbCommand cmd)
...
+
+		private DbDataReader DoExecuteReader(DbCommand cmd)
+		{
 			try
 			{
-				reader = cmd.ExecuteReader();
+				var reader = cmd.ExecuteReader();
+				return _factory.ConnectionProvider.Driver.SupportsMultipleOpenReaders
+					? reader
+					: NHybridDataReader.Create(reader);
```
bahusoid, line 258: Now `NHybridDataReader.Create` creation is also measured for `ExecuteReader took`. But why shouldn't it? It also allows to drop ugly `GetReaderStopwatch` with `NHybridDataReader` typing logic.

PR #2383, opened 2020-05-16
Support for criteria aliases in SqlProjection and SQLCriterion

### LINKED ISSUE #896: NH-1432 - Expression.Sql should support aliases other than {alias}

hazzik:
Fixes #896

This is an alternative approach to #2358. Here we replace all registered aliases without need of explicitly registering them.

### DIFF
--- a/src/NHibernate/Criterion/SQLCriterion.cs
+++ b/src/NHibernate/Criterion/SQLCriterion.cs
@@ -43,7 +43,7 @@ namespace NHibernate.Criterion
 					parameters[paramPos++].BackTrack = parameter.BackTrack;
 				}
 			}
-			return SQLProjection.GetSqlString(criteria, criteriaQuery, _sql);
+			return criteriaQuery.RenderSQLAliases(_sql).Replace("{alias}", criteriaQuery.GetSQLAlias(criteria));
 		}
 
 		public override TypedValue[] GetTypedValues(ICriteria criteria, ICriteriaQuery criteriaQuery)
--- a/src/NHibernate/Criterion/SQLCriterion.cs
+++ b/src/NHibernate/Criterion/SQLCriterion.cs
@@ -9,6 +9,7 @@ namespace NHibernate.Criterion
 	/// <summary>
 	/// An <see cref="ICriterion"/> that creates a SQLExpression.
 	/// The string {alias} will be replaced by the alias of the root entity.
+	/// Criteria aliases can also be used: "{a}.Value + {bc}.Value".
 	/// </summary>
 	/// <remarks>
 	/// This allows for database specific Expressions at the cost of needing to 
@@ -42,7 +43,7 @@ namespace NHibernate.Criterion
 					parameters[paramPos++].BackTrack = parameter.BackTrack;
 				}
 			}
-			return _sql.Replace("{alias}", criteriaQuery.GetSQLAlias(criteria));
+			return SQLProjection.GetSqlString(criteria, criteriaQuery, _sql);
 		}
 
 		public override TypedValue[] GetTypedValues(ICriteria criteria, ICriteriaQuery criteriaQuery)
--- a/src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
+++ b/src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
@@ -865,12 +865,16 @@ namespace NHibernate.Loader.Criteria
 
 		internal IDictionary<string, string> GetCriteriaSQLAliasMap()
 		{
-			var result = criteriaSQLAliasMap.Where(p => !string.IsNullOrEmpty(p.Key.Alias)).ToDictionary(p => p.Key.Alias, p => p.Value);
-			if (outerQueryTranslator != null)
+			var result = outerQueryTranslator != null
+				? new Dictionary<string, string>(outerQueryTranslator.GetCriteriaSQLAliasMap())
+				: new Dictionary<string, string>();
+
+			//NOTE: we SHOULD override the outer aliases
+			foreach (var p in criteriaSQLAliasMap)
 			{
-				foreach (var p in outerQueryTranslator.GetCriteriaSQLAliasMap())
+				if (!string.IsNullOrEmpty(p.Key.Alias))
 				{
-					result.Add(p.Key, p.Value);
+					result[p.Key.Alias] = p.Value;
 				}
 			}
 
--- a/src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
+++ b/src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
@@ -863,6 +863,20 @@ namespace NHibernate.Loader.Criteria
 			return GetEntityName(subcriteria);
 		}
 
+		internal IDictionary<string, string> GetCriteriaSQLAliasMap()
+		{
+			var result = criteriaSQLAliasMap.Where(p => !string.IsNullOrEmpty(p.Key.Alias)).ToDictionary(p => p.Key.Alias, p => p.Value);
+			if (outerQueryTranslator != null)
+			{
+				foreach (var p in outerQueryTranslator.GetCriteriaSQLAliasMap())
+				{
+					result.Add(p.Key, p.Value);
+				}
+			}
+
+			return result;
+		}
+
 		public string GetSQLAlias(ICriteria criteria, string propertyName)
 		{
 			if (StringHelper.IsNotRoot(propertyName, out var root))
--- a/src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
+++ b/src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
@@ -863,19 +863,18 @@ namespace NHibernate.Loader.Criteria
 			return GetEntityName(subcriteria);
 		}
 
-		internal IDictionary<string, string> GetCriteriaSQLAliasMap()
+		/// <summary> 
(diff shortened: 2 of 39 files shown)


src/NHibernate/Criterion/SQLCriterion.cs
```diff
@@ -42,7 +43,7 @@ public override SqlString ToSqlString(ICriteria criteria, ICriteriaQuery criteri
 					parameters[paramPos++].BackTrack = parameter.BackTrack;
 				}
 			}
-			return _sql.Replace("{alias}", criteriaQuery.GetSQLAlias(criteria));
+			return SQLProjection.GetSqlString(criteria, criteriaQuery, _sql);
```
fredericDelaporte, line 46: Slight code smell, but well, not worth introducing a common helper class for this. So let it be.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
```diff
@@ -863,6 +863,20 @@ public string GetEntityName(ICriteria subcriteria, string propertyName)
...
 		}
 
+		internal IDictionary<string, string> GetCriteriaSQLAliasMap()
+		{
+			var result = criteriaSQLAliasMap.Where(p => !string.IsNullOrEmpty(p.Key.Alias)).ToDictionary(p => p.Key.Alias, p => p.Value);
+			if (outerQueryTranslator != null)
+			{
+				foreach (var p in outerQueryTranslator.GetCriteriaSQLAliasMap())
+				{
+					result.Add(p.Key, p.Value);
```
fredericDelaporte, line 873: Possible breaking change. At least in SQL, it is supported to have a subquery alias hiding an outer query alias. If it is also supported with Criteria, then this code needs to be adjusted.
```suggestion
					if (!result.ContainsKey(p.Key))
						result.Add(p.Key, p.Value);
```

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
```diff
@@ -863,6 +863,20 @@ public string GetEntityName(ICriteria subcriteria, string propertyName)
...
+				}
+			}
```
fredericDelaporte, line 875: Or maybe something like:
```suggestion
			var result = (outerQueryTranslator.GetCriteriaSQLAliasMap() ?? criteriaSQLAliasMap.Where(p => !string.IsNullOrEmpty(p.Key.Alias))).ToDictionary(p => p.Key.Alias, p => p.Value);
			if (outerQueryTranslator != null)
			{
				foreach (var p in criteriaSQLAliasMap.Where(p => !string.IsNullOrEmpty(p.Key.Alias)))
				{
					result[p.Key] = p.Value;
				}
			}
```

hazzik: This is how I coded it initially.

hazzik: We should override the `outerQueryTranslator `'s aliases.

hazzik: I'll add test

hazzik:
> At least in SQL, it is supported to have a subquery alias hiding an outer query alias. If it is also supported with Criteria, then this code needs to be adjusted.

Yes, this is supported. For the actual SQL it will generate unique aliases.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
hazzik: Done

src/NHibernate/Criterion/SQLCriterion.cs
hazzik: This was copied from the other PR as is. The initial change is just a reset of my changes on top of master.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
hazzik: Fixed

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
```diff
@@ -863,6 +863,24 @@ public string GetEntityName(ICriteria subcriteria, string propertyName)
 			return GetEntityName(subcriteria);
 		}
 
+		internal IDictionary<string, string> GetCriteriaSQLAliasMap()
+		{
+			var result = outerQueryTranslator != null
+				? new Dictionary<string, string>(outerQueryTranslator.GetCriteriaSQLAliasMap())
```
hazzik, line 869: Creating a new dictionary because `outerQueryTranslator.GetCriteriaSQLAliasMap()` could be an `EmptyDictionary` which is read-only.

src/NHibernate/Loader/Criteria/CriteriaQueryTranslator.cs
```diff
@@ -863,6 +863,24 @@ public string GetEntityName(ICriteria subcriteria, string propertyName)
...
+				: new Dictionary<string, string>();
```
hazzik, line 870: We can avoid creating a new dictionary with in 2 ways:

1. Replace it with a `.Concat`. The downside is that the combined collection would contain duplicate elements. So it will depend on the order in which the aliases are rendered.
2. [Encapsulate](https://github.com/hazzik/nhibernate-core/pull/109) logic of rendering of SQL aliases here. This solves both creating a new dictionary and order of rendering aliases as it would be explicitly coded.

PR #2385, opened 2020-05-16
Fix detecting entities of left join by IEntityDecider

hazzik:
Fixes combination of #2324 and #2328


maca88:
Thanks for fixing the issue, I completely forgot about this cast.

PR #2386, opened 2020-05-17
Fix OData string comparisons (lt, gt, ge, le)

### LINKED ISSUE #2362: Not all oData operators are supported for strings
OData allows to use operators like >, <, <= and >= with strings, e.g.
$filter=TerritoryDescription ge 'South'

But attempt to use them with nHibernate cause System.NotSupportedException.

Is it possible to add support for these operators and stings?

 ```json
{
	"error": {
		"code": "",
		"message": "An error has occurred.",
		"innererror": {
			"message": "The 'ObjectContent`1' type failed to serialize the response body for content type 'application/json; odata.metadata=minimal'.",
			"type": "System.InvalidOperationException",
			"stacktrace": "",
			"internalexception": {
				"message": "Int32 Compare(System.String, System.String, System.StringComparison)",
				"type": "System.NotSupportedException",
				"stacktrace": "   at NHibernate.Linq.Visitors.HqlGeneratorExpressionVisitor.VisitMethodCallExpression(MethodCallExpression expression)\r\n   at NHibernate.Linq.Visitors.HqlGeneratorExpressionVisitor.VisitExpression(Expression expression)\r\n   at NHibernate.Linq.Visitors.HqlGeneratorExpressionVisitor.VisitBinaryExpression(BinaryExpression expression)\r\n   at NHibernate.Linq.Visitors.HqlGeneratorExpressionVisitor.VisitExpression(Expression expression)\r\n   at NHibernate.Linq.Visitors.QueryModelVisitor.VisitWhereClause(WhereClause whereClause, QueryModel queryModel, Int32 index)\r\n   at Remotion.Linq.QueryModelVisitorBase.VisitBodyClauses(ObservableCollection`1 bodyClauses, QueryModel queryModel)\r\n   at Remotion.Linq.QueryModelVisitorBase.VisitQueryModel(QueryModel queryModel)\r\n   at NHibernate.Linq.Visitors.QueryModelVisitor.Visit()\r\n   at NHibernate.Linq.Visitors.QueryModelVisitor.GenerateHqlQuery(QueryModel queryModel, VisitorParameters parameters, Boolean root, Nullable`1 rootReturnType)\r\n   at NHibernate.Linq.NhLinqExpression.Translate(ISessionFactoryImplementor sessionFactory, Boolean filter)\r\n   at NHibernate.Hql.Ast.ANTLR.ASTQueryTranslatorFactory.CreateQueryTranslators(IQueryExpression queryExpression, String collectionRole, Boolean shallow, IDictionary`2 filters, ISessionFactoryImplementor factory)\r\n   at NHibernate.Engine.Query.QueryPlanCache.GetHQLQueryPlan(IQueryExpression queryExpression, Boolean shallow, IDictionary`2 enabledFilters)\r\n   at NHibernate.Impl.AbstractSessionImpl.GetHQLQueryPlan(IQueryExpression queryExpression, Boolean shallow)\r\n   at NHibernate.Impl.AbstractSessionImpl.CreateQuery(IQueryExpression queryExpression)\r\n   at NHibernate.Linq.DefaultQueryProvider.PrepareQuery(Expression expression, IQuery& query)\r\n   at NHibernate.Linq.DefaultQueryProvider.Execute(Expression expression)\r\n   at NHibernate.Linq.DefaultQueryProvider.Execute[TResult](Expression expression)\r\n   at Remotion.Linq.QueryableBase`1.System.Collections.IEnumerable.GetEnumerator()\r\n   at Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSetSerializer.WriteResourceSet(IEnumerable enumerable, IEdmTypeReference resourceSetType, ODataWriter writer, ODataSerializerContext writeContext)\r\n   at Microsoft.AspNet.OData.Formatter.ODataOutputFormatterHelper.WriteToStream(Type type, Object value, IEdmModel model, ODataVersion version, Uri baseAddress, MediaTypeHeaderValue contentType, IWebApiUrlHelper internaUrlHelper, IWebApiRequestMessage internalRequest, IWebApiHeaders internalRequestHeaders, Func`2 getODataMessageWrapper, Func`2 getEdmTypeSerializer, Func`2 getODataPayloadSerializer, Func`1 getODataSerializerContext)\r\n   at Microsoft.AspNet.OData.Formatter.ODataMediaTypeFormatter.WriteToStreamAsync(Type type, Object value, Stream writeStream, HttpContent content, TransportContext transportContext, CancellationToken cancellationToken)\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()\r\n   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n   at System.Web.Http.WebHost.HttpControllerHandler.<WriteBufferedResponseContentAsync>d__1b.MoveNext()"
			}
		}
	}
}
```

### PR DESCRIPTION
**Pull request by bahusoid**
Fixes #2362

Support StringComparison parameter with warning for string  methods (StartsWith, EndsWith, IndexOf, Replace)

### DIFF
--- a/src/NHibernate.Test/Linq/ODataTests.cs
+++ b/src/NHibernate.Test/Linq/ODataTests.cs
@@ -75,6 +75,7 @@ namespace NHibernate.Test.Linq
 			Assert.That(results, Has.Count.EqualTo(expectedRows));
 		}
 
+		//GH-2362
 		[TestCase("$filter=CustomerId le 'ANATR'",2 )]
 		[TestCase("$filter=startswith(CustomerId, 'ANATR')", 1)]
 		[TestCase("$filter=endswith(CustomerId, 'ANATR')", 1)]
--- a/src/NHibernate.Test/Linq/ODataTests.cs
+++ b/src/NHibernate.Test/Linq/ODataTests.cs
@@ -79,9 +79,12 @@ namespace NHibernate.Test.Linq
 		[TestCase("$filter=CustomerId le 'ANATR'",2 )]
 		[TestCase("$filter=startswith(CustomerId, 'ANATR')", 1)]
 		[TestCase("$filter=endswith(CustomerId, 'ANATR')", 1)]
-		[TestCase("$filter=indexof(CustomerId, 'ANATR') eq 0", 1)]
-		public void StringFilter(string queryString, int expectedCount)
+		[TestCase("$filter=indexof(CustomerId, 'ANATR') eq 0", 1, true)]
+		public void StringFilter(string queryString, int expectedCount, bool locateFunction = false)
 		{
+			if(locateFunction && !TestDialect.SupportsLocate)
+				Assert.Ignore("Locate function is not supported.");
+
 			Assert.That(
 				ApplyFilter(session.Query<Customer>(), queryString).Cast<Customer>().ToList(),
 				Has.Count.EqualTo(expectedCount));
--- a/src/NHibernate.Test/Linq/ODataTests.cs
+++ b/src/NHibernate.Test/Linq/ODataTests.cs
@@ -75,6 +75,17 @@ namespace NHibernate.Test.Linq
 			Assert.That(results, Has.Count.EqualTo(expectedRows));
 		}
 
+		[TestCase("$filter=CustomerId le 'ANATR'",2 )]
+		[TestCase("$filter=startswith(CustomerId, 'ANATR')", 1)]
+		[TestCase("$filter=endswith(CustomerId, 'ANATR')", 1)]
+		[TestCase("$filter=indexof(CustomerId, 'ANATR') eq 0", 1)]
+		public void StringFilter(string queryString, int expectedCount)
+		{
+			Assert.That(
+				ApplyFilter(session.Query<Customer>(), queryString).Cast<Customer>().ToList(),
+				Has.Count.EqualTo(expectedCount));
+		}
+
 		private IQueryable ApplyFilter<T>(IQueryable<T> query, string queryString)
 		{
 			var context = new ODataQueryContext(CreatEdmModel(), typeof(T), null) { };
--- a/src/NHibernate.Test/Linq/ODataTests.cs
+++ b/src/NHibernate.Test/Linq/ODataTests.cs
@@ -76,15 +76,12 @@ namespace NHibernate.Test.Linq
 		}
 
 		//GH-2362
-		[TestCase("$filter=CustomerId le 'ANATR'",2 )]
+		[TestCase("$filter=CustomerId le 'ANATR'", 2)]
 		[TestCase("$filter=startswith(CustomerId, 'ANATR')", 1)]
 		[TestCase("$filter=endswith(CustomerId, 'ANATR')", 1)]
-		[TestCase("$filter=indexof(CustomerId, 'ANATR') eq 0", 1, true)]
-		public void StringFilter(string queryString, int expectedCount, bool locateFunction = false)
+		[TestCase("$filter=indexof(CustomerId, 'ANATR') eq 0", 1)]
+		public void StringFilter(string queryString, int expectedCount)
 		{
-			if (locateFunction && !TestDialect.SupportsLocate)
-				Assert.Ignore("Locate function is not supported.");
-
 			Assert.That(
 				ApplyFilter(session.Query<Customer>(), queryString).Cast<Customer>().ToList(),
 				Has.Count.EqualTo(expectedCount));
--- a/src/NHibernate.Test/Linq/ODataTests.cs
+++ b/src/NHibernate.Test/Linq/ODataTests.cs
@@ -82,7 +82,7 @@ namespace NHibernate.Test.Linq
 		[TestCase("$filter=indexof(CustomerId, 'ANATR') eq 0", 1, true)]
 		public void StringFilter(string queryString, int expectedCount, bool locateFunction = false)
 		{
-			if(locateFunction && !TestDialect.SupportsLocate)
+			if (locateFunction && !TestDialect.SupportsLocate)
 				Assert.Ignore("Locate function is not supported.");
 
 			Assert.That(
--- a/src/NHibernate/Linq/Functions/BaseHqlGeneratorForMethod.cs
+++ b/src/NHibernate/Linq/Functions/BaseHqlGeneratorForMethod.cs
@@ -21,13 +21,13 @@ namespace NHibernate.Linq.Functions
 
 		private protected static void LogIgnoredParameter(MethodInfo method, string paramType)
 		{
-			if(Log.IsWarnEnabled())
+			if (Log.IsWarnEnabled())
 				Log.Warn("Method parameter of type '{0}' is ignored when converting to hql the following method: {1}", paramType, method);
 		}
 
 		private protected static void LogIgnoredStringComparisonParameter(MethodInfo actualMethod, MethodInfo methodWithStringComparison)
 		{
-			if(actualMethod == methodWithStringComparison)
+			if (actualMethod == methodWithStringComparison)
 				LogIgnoredParameter(actualMethod, nameof(StringComparison));
 		}
 
--- a/src/NHibernate/Linq/Functions/BaseHqlGeneratorForMethod.cs
+++ b/src/NHibernate/Linq/Functions/BaseHqlGeneratorForMethod.cs
@@ -1,5 +1,7 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using NHibernate.Hql.Ast;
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/NHibernate.Test/Linq/ODataTests.cs [hunk 1]
```diff
@@ -53,6 +53,17 @@ public void OrderGroupBy(string queryString, int expectedRows)
 			Assert.That(results, Has.Count.EqualTo(expectedRows));
 		}
 
+		[TestCase("$filter=CustomerId le 'ANATR'",2 )]
+		[TestCase("$filter=startswith(CustomerId, 'ANATR')", 1)]
+		[TestCase("$filter=endswith(CustomerId, 'ANATR')", 1)]
+		[TestCase("$filter=indexof(CustomerId, 'ANATR') eq 0", 1)]
```
fredericDelaporte, line 59: Since we do not register `EndsWith`, `StartsWith` or `IndexOf` overloads with `StringComparison`, does this means there is an inconsistency in OData implementation here? It would emit `Compare` calls with that argument, but it would omit it for these three other cases?

So I see three options:
- only register what is currently required by OData, as does this PR currently.
- also register other cases overloads in case OData changes on this, or in case another Linq generator library requires them.
- reject this as an external issue, to be fixed on OData side.

@maca88, @hazzik

bahusoid: > does this means there is an inconsistency in OData implementation here?

Yeap. Seems to be.

> also register other cases overloads in case OData changes on this, or in case another Linq generator library requires them.

I would choose this one as I see no difference with call to default `string.Compare`. It's not corelated in any way to with how DB is configured to make string comparisons. In fact `StringComparison` might be used to make LINQ to objects queries behave the same as DB (as I already mentioned in case of default SQL Server  case insensitive collation)

fredericDelaporte: Ok.

maca88: By just adding the missing overload we introduce another issue where we generate the same query when using `StringComparison.XXX` and `StringComparison.XXXIgnoreCase`, which would cause an unexpected result on a case sensitive collation. IMO we should at least try to mimic .NET behavior when using `XXXIgnoreCase` by using `lower` function on each side.
> does this means there is an inconsistency in OData implementation here?

Yes, from the OData [source code](https://github.com/OData/WebApi/blob/813349a77669b5cd7c677538119b77f0ef82d755/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ExpressionBinderBase.cs#L200) only `StringComparison.Ordinal` option is used for `string.Compare`, where for `EndsWith`, `StartsWith` and other methods the `StringComparison` overload [is not used](https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNet.OData.Shared/Query/Expressions/ClrCanonicalFunctions.cs#L48).

fredericDelaporte: > By just adding the missing overload we introduce another issue where we generate the same query when using `StringComparison.XXX` and `StringComparison.XXXIgnoreCase`, which would cause an unexpected result on a case sensitive collation.

The thing is, we do not try to support different `StringComparison` values, we just ignore them. The trouble is the same conversely, on case insensitive collations, when using a non ignoring case `StringComparison`. And for this converse case, there is no ".Net" side workaround.

The current state is, NHibernate bluntly states these overloads are not supported. That is fine for user generated expression. But when the user cannot control the generated expression, it is a blocker.

So I think accepting these overloads, while not actually supporting the `StringComparison` argument, is an acceptable trade-of. Maybe we should emit a NHibernate warning or information about the argument being ignored, though.

hazzik:
Maybe we should check that the argument is `StringComparison.Ordinal` and if it is not - throw an exception?

src/NHibernate.Test/Linq/ODataTests.cs
maca88: > And for this converse case, there is no ".Net" side workaround

Some databases supports the COLLATE clause ([SqlLite](https://www.sqlite.org/datatype3.html#collation), [SqlServer](https://blog.sqlauthority.com/2007/04/30/case-sensitive-sql-query-search/), [Oracle](https://oracle-base.com/articles/12c/column-level-collation-and-case-insensitive-database-12cr2#statement-level), [PostgreSql](https://www.postgresql.org/docs/current/collation.html), [MySql](https://dev.mysql.com/doc/refman/8.0/en/charset-collate.html)) within the query, which could be used to do case sensitive/insensitive comparison. But thats a matter for a separate PR.

> So I think accepting these overloads, while not actually supporting the StringComparison argument, is an acceptable trade-of

It is fine by me.

maca88:
> if it is not - throw an exception?

I would lean more towards throwing an exception than just ignoring the argument, but I fine with both approaches.

bahusoid:
I don't understand why `Ordinal` should be supported but not `CurrentCulture` or  `InvariantCulture`.  Specific fix for ODATA behavior? Again for SQL server equivalent would be something like `OrdinalIgnoreCase`. Supporting only `Ordinal` makes it look more like a strange hack to me... So I agree with Frederic about warn and simply ignoring this param.

src/NHibernate/Linq/Functions/StringGenerator.cs
```diff
@@ -210,18 +216,28 @@ public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject,
...
 										ReflectHelper.GetMethodDefinition<string>(s => s.IndexOf(' ', 0)),
-										ReflectHelper.GetMethodDefinition<string>(s => s.IndexOf(" ", 0))
+										ReflectHelper.GetMethodDefinition<string>(s => s.IndexOf(" ", 0)),
+										MethodWithComparer1,
+										MethodWithComparer2,
 									};
 		}
 		public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject, ReadOnlyCollection<Expression> arguments, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor)
 		{
+			if (LogIgnoredStringComparisonParameter(method, MethodWithComparer1, MethodWithComparer2))
```
hazzik, line 236: Easier to check that function has 3 arguments.

src/NHibernate/Linq/Functions/StringGenerator.cs
```diff
@@ -95,15 +98,18 @@ public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject,
...
 		{
-			SupportedMethods = new[] { ReflectHelper.GetMethodDefinition<string>(x => x.EndsWith(null)) };
+			SupportedMethods = new[] {ReflectHelper.GetMethodDefinition<string>(x => x.EndsWith(null)), MethodWithComparer,};
 		}
 
 		public override bool AllowsNullableReturnType(MethodInfo method) => false;
 
 		public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject, ReadOnlyCollection<Expression> arguments, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor)
 		{
+			LogIgnoredStringComparisonParameter(method, MethodWithComparer);
```
hazzik, line 112: Easier to check that function has 2 arguments

src/NHibernate/Linq/Functions/StringGenerator.cs
```diff
@@ -244,21 +260,32 @@ public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject,
...
 		}
 
 		public override HqlTreeNode BuildHql(MethodInfo method, Expression targetObject, ReadOnlyCollection<Expression> arguments, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor)
 		{
-			return treeBuilder.MethodCall("replace",
-																		visitor.Visit(targetObject).AsExpression(),
-																		visitor.Visit(arguments[0]).AsExpression(),
-																		visitor.Visit(arguments[1]).AsExpression());
+#if NETCOREAPP2_0
+			LogIgnoredStringComparisonParameter(method, MethodWithComparer);
```
hazzik, line 282: Easier to check number of arguments

src/NHibernate/Linq/Functions/StringGenerator.cs
bahusoid: Not in case ambiguity. Like here 2 arguments for "s.IndexOf(" ", 0)" and `s.IndexOf("", StringComparison.Oridnal)`

And what you mean be easier? This method is easier to reuse in multiple places. And argument count is error prone approach when doing changes in multiple places?

src/NHibernate.Test/Linq/ODataTests.cs
```diff
@@ -75,6 +75,21 @@ public void BasePropertyFilter(string queryString, int expectedRows)
...
 
+		//GH-2362
+		[TestCase("$filter=CustomerId le 'ANATR'",2 )]
+		[TestCase("$filter=startswith(CustomerId, 'ANATR')", 1)]
+		[TestCase("$filter=endswith(CustomerId, 'ANATR')", 1)]
+		[TestCase("$filter=indexof(CustomerId, 'ANATR') eq 0", 1, true)]
+		public void StringFilter(string queryString, int expectedCount, bool locateFunction = false)
+		{
+			if(locateFunction && !TestDialect.SupportsLocate)
+				Assert.Ignore("Locate function is not supported.");
```
fredericDelaporte, line 86: Whitespace glitch (missing space after if), but anyway should be removed after #2392 merging.

src/NHibernate/Linq/Functions/BaseHqlGeneratorForMethod.cs
```diff
@@ -9,10 +11,33 @@ namespace NHibernate.Linq.Functions
...
+
 		public IEnumerable<MethodInfo> SupportedMethods { get; protected set; }
 
 		public abstract HqlTreeNode BuildHql(MethodInfo method, Expression targetObject, ReadOnlyCollection<Expression> arguments, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor);
 
 		public virtual bool AllowsNullableReturnType(MethodInfo method) => true;
+
+		private protected static void LogIgnoredParameter(MethodInfo method, string paramType)
+		{
+			if(Log.IsWarnEnabled())
```
fredericDelaporte, line 24: You do not like having to put a space after an `if`, it seems!

src/NHibernate/Linq/Functions/BaseHqlGeneratorForMethod.cs
```diff
@@ -9,10 +11,33 @@ namespace NHibernate.Linq.Functions
...
+				Log.Warn("Method parameter of type '{0}' is ignored when converting to hql the following method: {1}", paramType, method);
+		}
+
+		private protected static void LogIgnoredStringComparisonParameter(MethodInfo actualMethod, MethodInfo methodWithStringComparison)
+		{
+			if(actualMethod == methodWithStringComparison)
```
fredericDelaporte, line 30: Same here.

src/NHibernate/Linq/Functions/StringGenerator.cs
fredericDelaporte: >Easier to check that function has 2 arguments

But less "future-proof", we cannot assume another overload with 2 arguments may never be added into .Net.

src/NHibernate.Test/Linq/ODataTests.cs
bahusoid: I trusted Rider for doing such things... Need to check why it's not happening anymore..

PR #2391, opened 2020-05-20
Use latest Firebird for appveyour and travis

bahusoid:
Any idea why firebird sometimes fails on appveyor? Let's upgrade just in case :)


bahusoid:
Nope... Didn't help... Seems hack with clearing pool sometimes doesn't work on appveyour:
https://github.com/nhibernate/nhibernate-core/blob/bd3fa290e86bd479d7a2659773d54642d8c8caec/src/NHibernate.Test/TestCase.cs#L295-L304

PR #2393, opened 2020-05-21
Improve dirty check performance of set 

ssimek:
I'm surprised that this trivial issue caused by an O(N^2) dirty check comparison went unfixed for so long. In my particular real-life scenario, this fix improved the performance by about an order of magnitude, almost all of which was spent in the dirty check according to profiler.

Also, is there a possiblity to include this in the 4.0.x branch and create a 4.0.5 release? We cannot move to a newer version so easily.

Thanks!

Related to #857

### DIFF
--- a/src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
+++ b/src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
@@ -8,6 +8,22 @@ namespace NHibernate.Collection.Generic.SetHelpers
 	internal class SetSnapShot<T> : ISetSnapshot<T>
 	{
 		private readonly List<T> _elements;
+
+		/// <summary>
+		/// Lazy cache for TryGetValue calls
+		/// </summary>
+		private Dictionary<T, T> _lookupCache;
+
+		/// <summary>
+		/// Index of the first item that is not yet in the lookup cache
+		/// </summary>
+		private int _lookupCacheUnprocessed;
+
+		/// <summary>
+		/// <see langword="true" /> if the snapshot contains a <see langword="null" /> element
+		/// </summary>
+		private bool _containsNull;
+
 		public SetSnapShot()
 		{
 			_elements = new List<T>();
@@ -36,6 +52,16 @@ namespace NHibernate.Collection.Generic.SetHelpers
 		public void Add(T item)
 		{
 			_elements.Add(item);
+			// we don't need to invalidate the cache if it was already created,
+			// as items are always added at the end and will be caught by
+			// _lookupCacheUnprocessed
+
+			if (item == null)
+			{
+				// keep track if the collection contains a null item
+				// these are processed separately in TryGetValue
+				_containsNull = true;
+			}
 		}
 
 		public void Clear()
@@ -45,7 +71,9 @@ namespace NHibernate.Collection.Generic.SetHelpers
 
 		public bool Contains(T item)
 		{
-			return _elements.Contains(item);
+			// use the caching implementation in TryGetValue
+			T discard;
+			return TryGetValue(item, out discard);
 		}
 
 		public void CopyTo(T[] array, int arrayIndex)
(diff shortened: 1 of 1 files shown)


src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -8,6 +8,22 @@ namespace NHibernate.Collection.Generic.SetHelpers
 	internal class SetSnapShot<T> : ISetSnapshot<T>
 	{
 		private readonly List<T> _elements;
```
fredericDelaporte, line 10: Changing this into a `LinkedHashMap<T, T>` (from `NHibernate.Util` namespace) could do most of the job, and that would be way simpler, also avoiding having two collections instead of one.
But it does not support a `null` key. It could be modified to have an option for supporting it, maybe. This option would cause it to no more respect `IDictionary<TKey, TValue>` contract which explicitly states `null` keys are not supported, so it is debatable.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -95,13 +123,47 @@ public bool IsReadOnly
...
+			{
+				T snapshotElement = _elements[_lookupCacheUnprocessed++];
+				if (snapshotElement != null)
+				{
+					// be careful not to replace cache entries to preserve
+					// original semantics and return the first matching object
+					if (!_lookupCache.ContainsKey(snapshotElement))
+					{
+						_lookupCache.Add(snapshotElement, snapshotElement);
+					}
```
fredericDelaporte, line 155: This is accidental semantic. Having the snapshot containing two times the same element makes no sense, and would anyway cause bugs if this was actually happening. So it is unneeded.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
fredericDelaporte: In fact we do not need the ordering preserving capability either for the `SetSnapshot`. Set are not order sensitive, included in dirty check logics.

So we could use a regular `Dictionary<T, T>` instead of the `List`, with your additional field for handling the `null` case. The enumerator would have to be overloaded for also yielding the null if there is some, the count too for taking the null element into account, ...

ssimek: My implementation focuses primarily on preventing any possible regressions as I am not so familiar with NH to anticipate what can and cannot appear in the snapshot. Other than being significantly more complex, LinkedHashMap would also not support multiple entries which would claim to be Equal. I assumed there was a reason for the List<T> based implementation of the SetSnapShot, it just failed terribly due to the way TryGetValue is used in a loop against a similarly sized collection.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
ssimek: I was not willing to risk that, see above comment.

ssimek:
As for this not being a fix for the original issue - the report says that dirty comparison is "slow", not that it cannot scale to huge sets. I don't care if it assigned as a fix or an improvement, just wanted to share what I consider an useful contribution.
In the very real case for which this fix was designed, the entire process which caught my attention took roughly 21 minutes before and 3 minutes after applying the fix. Note that those 3 minutes were spent doing actual work, 99.6% of the rest was spent inside SetSnapShot.TryGetValue calls. So more than two orders of magnitude improvement. I do not have the exact numbers, but the collections were not unimaginably large. Quadratic complexity doesn't need a lot to get out of hand.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
fredericDelaporte: This snapshot helper class ends up being consumed only in `PersistentGenericSet`. It may by passed around in NHibernate code but only for being used inside some `PersistentGenericSet` instance.
So you can check what features it actually needs.
And as it is an internal class, it can be freely changed.

fredericDelaporte:
I am giving my rationale for classifying it, it should not be taken as any kind of reproach. Your contribution is anyway welcome, it pinpoints a point that can be improved.
Even if your proposed fix is changed for something else, it is still valuable.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
ssimek: There is no strong guarantee of input. Also, an instance of SetSnapShot gets passed to and returned from GetOrphans, used in SequenceEquals, etc. Scenarios where the lookup by key does not even come into play, that's why I am creating the lookup dictionary very lazily.
I would need extremely deep knowledge of NH to feel confident changing the (even undocumented) contract of a class that's been there unchanged since 2.0.

ssimek:
As for the other question - is there a chance of getting this (if accepted) into a 4.0.5 release, or do we have to keep our own fork?

fredericDelaporte:
Well, the last 4.x release was 4.1.2GA, April 2018, for a regression appeared in 4.1. It is highly unlikely we will back-port a performance improvement all the way down to 4.0.x

fredericDelaporte:
Replaced by #2394.

PR #2394, opened 2020-05-21
Optimize PersistentGenericSet snapshot

maca88:
An alternative approach of #2393, which uses `HashSet<T>.TryGetValue` for .Net Core and a dictionary for .NET Framework


src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -1,109 +1,177 @@
...
-			var idx = _elements.IndexOf(element);
-			if (idx >= 0)
+			get
 			{
-				value = _elements[idx];
-				return true;
-			}
+				if (_syncRoot == null)
+				{
+					Interlocked.CompareExchange<object>(ref _syncRoot, new object(), null);
```
maca88, line 169: Used the same logic that as `List<>` has.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -1,109 +1,178 @@
...
+#endif
 
 namespace NHibernate.Collection.Generic.SetHelpers
 {
+#if NETFX || NETSTANDARD2_0
+	// TODO 6.0: Consider removing this class in case we upgrade to .NET 4.7.2 and NET Standard 2.1,
+	// which have HashSet<T>.TryGetValue
 	[Serializable]
-	internal class SetSnapShot<T> : ISetSnapshot<T>
+	internal class SetSnapShot<T> : ICollection<T>, IReadOnlyCollection<T>, ICollection
```
hazzik, line 15: Deserialization callback is missing which is required to properly deserialize Dictionary. I think this was a part of a reason why it was implemented using `List<>` in a first place.

src/NHibernate/Async/Collection/Generic/PersistentGenericSet.cs
```diff
@@ -54,7 +54,7 @@ public override async Task<bool> EqualsSnapshotAsync(ICollectionPersister persis
 		{
 			cancellationToken.ThrowIfCancellationRequested();
 			var elementType = persister.ElementType;
-			var snapshot = (ISetSnapshot<T>)GetSnapshot();
+			var snapshot = (SetSnapShot<T>)GetSnapshot();
```
bahusoid, line 57: Can't we add `GetSnapshotTyped()` returning `SetSnapShot<T>` and avoid typing in all methods

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
maca88: Maybe I am missing something as I don't understand why we need a deserialization callback to properly deserialize the dictionary and what should the callback do. I've [made a fiddle](https://dotnetfiddle.net/evzfmL) to test the serialization of `SetSnapShot` and it's working without the callback. Also `PersistentGenericMap` has a dictionary field and it doesn't implement `IDeserializationCallback`, so if we need the callback for `SetSnapShot`, then we probably would need to add it for `PersistentGenericMap`.

src/NHibernate/Async/Collection/Generic/PersistentGenericSet.cs
maca88: Added.

bahusoid: That's not exactly what I meant. Typing is still there - it's just moved to separate method. I meant extract content of `GetSnapshot` to `GetSnapshotTyped`

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
bahusoid: Is it about #1891? I believe it is only the case if hashcodes are cached for dictionary keys

fredericDelaporte: As far as I remember, the issue with dictionaries happens when the type may be deserialized from a graph of entities which own deserialization requires having the deserialized dictionaries populated early.
The concrete example of this need is a cache key class which "value" depends on an internal dictionary, while the key has been serialized as part of another dictionary. In such case we must ensure the cache key own internal dictionary is deserialized and ready for key computing before the key is added to its containing dictionary.
Implementing `IDeserializationCallback` is then required because we must then serialize the dictionary as a simpler type (an array), available sooner in the deserialization process.

For the case here, I do not think we need to do this. This `SetSnapShot` class does not need to have its state ready early for being correctly deserialized inside a graph of entity, as its state is not required during the deserialization process of the entities it may belong to.

src/NHibernate/Async/Collection/Generic/PersistentGenericSet.cs
fredericDelaporte: I do not get what or how @bahusoid. We have `GetSnapshot() => session.PersistenceContext.GetSnapshot(this)`, we anyway have to put the typing somewhere, unless rewriting a lot of things to be generic, but that would not fit in this PR scope.

bahusoid: We have `PersistentGenericSet.GetSnapshot`. Why can't content of it be extracted to separate method:
```C#
public override object GetSnapshot(ICollectionPersister persister)
{
	return GetSnapshotSet(persister);
}

private SetSnapShot<T> GetSnapshotSet(ICollectionPersister persister)
{
	var clonedSet = new SetSnapShot<T>(WrappedSet.Count);
	var enumerable = from object current in WrappedSet
					select persister.ElementType.DeepCopy(current, persister.Factory);
	foreach (var copied in enumerable)
	{
		clonedSet.Add((T) copied);
	}

	return clonedSet;
}
```

And use it inside this class to avoid typing only inside `PersistentGenericSet`

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -1,109 +1,178 @@
...
 		public bool Contains(T item)
 		{
-			return _elements.Contains(item);
+			return item == null ? _hasNull : _values.ContainsKey(item);
 		}
 
 		public void CopyTo(T[] array, int arrayIndex)
 		{
-			_elements.CopyTo(array, arrayIndex);
+			_values.Keys.CopyTo(array, arrayIndex);
```
fredericDelaporte, line 89: We are lacking the `_hasNull` handling here, for correctness. For consistency with current `GetEnumerator()`, it should be considered as the first element to copy.
```suggestion
			if (_hasNull)
				array[arrayIndex] = default(T);
			_values.Keys.CopyTo(array, arrayIndex + (_hasNull ? 1 : 0));
```

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -1,109 +1,178 @@
...
+			if (!(array is T[] typedArray))
+			{
+				throw new ArgumentException($"Array must be of type {typeof(T[])}.", nameof(array));
+			}
+
+			CopyTo(typedArray, index);
 		}
 
-		public bool IsSynchronized
+		public int Count => _values.Count;
```
fredericDelaporte, line 123:
```suggestion
		public int Count => _values.Count + (_hasNull ? 1 : 0);
```

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
bahusoid: Do you know what null means here? It looks like some legacy code considering that `HashSet` which is used for storing elements can't have null values.

fredericDelaporte: I see nothing that prevents `HashSet<T>` to have a `null`. They accept `null`.

src/NHibernate/Async/Collection/Generic/PersistentGenericSet.cs
fredericDelaporte: It wrecks the logic. The snapshot must be taken from the persistence context, not from the `WrappedSet`. Otherwise the collection will never be considered dirty. It seems to me you are mixing up `GetSnapshot()` with `GetSnapshot(ICollectionPersister persister)`. Although they share the same name, their purpose is not the same. The former get the snapshot representing the collection state as it was in the database upon loading, while the later extract a snapshot from the current collection state.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
bahusoid: Ok.. I might mistake it with dictionary key or something else. But our code explicitly ignores null values from adding:
https://github.com/nhibernate/nhibernate-core/blob/20a61d1ca05dd792cd72ad3a6cb76307c7cb5642/src/NHibernate/Collection/Generic/PersistentGenericSet.cs#L168

src/NHibernate/Async/Collection/Generic/PersistentGenericSet.cs
bahusoid: Yeah.. Sorry  my mistake...

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
fredericDelaporte: This is a thing which was done in af769586 as an "unrelated" by the way change, which according to comment of that time was required due to a limitation of the Iesi `set` NHibernate was using. (The Iesi set was required by lack of set in the .Net framework prior to .Net 4.)

I think keeping this limitation is now undue, it prevents supporting a set of `int?` elements with a null inside it, by example, with no reason. But removing is out of the scope of this PR.

bahusoid: > I think keeping this limitation is now undue, it prevents supporting a set of int? elements with a null inside it, by example, with no reason. But removing this limitation is out of the scope of this PR.

Null removing is consistent with Java version. And for one thing it would break many-to-many filtering (I removed null elements for list also not so long ago - #2181). So I don't think this behavior should change

bahusoid: So to conclude: maybe this change can be removed but skipping nulls should stay for entity types. And now you suggested change seems won't be used and can't be tested..

hazzik:
Maybe we stick to the dictionary to avoid using conditional compilation?

hazzik:
Also, hibernate just uses `Map` without any additional wrappers.

maca88:
Hibernate is using [HashMap](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/collection/internal/PersistentSet.java#L103), which [supports null keys](https://www.mycompiler.io/view/AF2TZuQ), so I think we should too.

> Maybe we stick to the dictionary to avoid using conditional compilation?

To me using a dictionary is only a temporary solution until we upgrade to .NET Standard 2.1 or higher as `HashSet` is more memory efficent and a simpler solution for the problem (less code).

src/NHibernate/Async/Collection/Generic/PersistentGenericSet.cs
bahusoid: Async regen is missed for `GetSnapshotTyped`. Or just kill commit adding it as it's not very useful in the end..

maca88: Reverted commit.

hazzik:
Is it possible to add unit tests for SetSnapshot (but no reflection please)?

maca88:
Tests added.

PR #2396, opened 2020-05-24
Rewrite string.Contains/string.StartsWith/string.EndsWith to SqlMethods.Like so DB can reuse query plan

### LINKED ISSUE #725: NH-2325 - problems with StartsWith /  EndsWith / Contains / Substring / IndexOf
**Hung Tran** created an [issue](https://nhibernate.jira.com/browse/NH-2325) — <time datetime="2010-09-07T03:18:55.179Z" title="2010-09-07T03:18:55.179Z">7th September 2010, 3:18:55</time>:
> My mapping is
> 
> 
> ```xml 
> <hibernate-mapping xmlns="urn:nhibernate-mapping-2.2" auto-import="true" default-access="field.camelcase-underscore" default-lazy="true" default-cascade="none" >
>     <class name="Order" dynamic-update="true" dynamic-insert="true">
>         <id name="Id" unsaved-value="-1" type="Int64" length="18">
>             <generator class="native"/>
>         </id>
>         <property name="Code" type="String" length="255" index="IX*CODE*D1436656544BAE8D"/>
>         <property name="Comment" type="String" length="255"/>
>         <bag name="Lines" fetch="subselect" lazy="false" cascade="all-delete-orphan">
>             <key column="Order"/>
>             <one-to-many class="OrderLine"/>
>         </bag>
>     </class>
>     <class name="OrderLine" dynamic-update="true" dynamic-insert="true">
>         <id name="Id" unsaved-value="-1" type="Int64" length="18">
>             <generator class="native"/>
>         </id>
>         <many-to-one name="Product" class="Product" column="Product" index="IX*FK*95888E80952904EB" lazy="proxy"/>
>         <property name="Quantity" type="Int32"/>
>         <property name="RetailPrice" type="Decimal"/>
>     </class>
>     <class name="Product" dynamic-update="true" dynamic-insert="true">
>         <id name="Id" unsaved-value="-1" type="Int64" length="18">
>             <generator class="native"/>
>         </id>
>         <property name="Code" type="String" length="255" index="IX*CODE*952904EB544BAE8D"/>
>         <property name="CostPrice" type="Decimal"/>
>         <property name="Description" type="String" length="255"/>
>         <property name="Name" type="String" length="255"/>
>     </class>
> </hibernate-mapping>
> ```
> 
> LINQ: 
> ```c#
> (from o in db.AsQueryable<Order>() where o.Code.Contains("001") select o).ToList();
> ```
> 
>  SQL: 
> ```sql
> select order0*.[ID] as ID1_0_, order0_.[CODE] as CODE2_0_, order0_.[COMMENT] as COMMENT3_0_ from [ORDER] order0_ where order0*.[CODE] like ('%'+@p0+'%');@p0 = '001' [Type: String (4000)]
> ```
> 
> LINQ: 
> ```c#
> (from o in db.AsQueryable<Order>() where o.Code.StartsWith("001") select o).ToList();
> ```
> 
>  SQL: 
> ```sql
> select order0*.[ID] as ID1_0_, order0_.[CODE] as CODE2_0_, order0_.[COMMENT] as COMMENT3_0_ from [ORDER] order0_ where order0*.[CODE] like (@p0+%');@p0 = '001' [Type: String (4000)]
> ```
> 
> LINQ: 
> ```c#
> (from o in db.AsQueryable<Order>() where o.Code.EndsWith("001") select o).ToList();
> ```
>  SQL:
> ```sql
> select order0*.[ID] as ID1_0_, order0_.[CODE] as CODE2_0_, order0_.[COMMENT] as COMMENT3_0_ from [ORDER] order0_ where order0*.[CODE] like ('%'+@p0);@p0 = '001' [Type: String (4000)]
> ```
> 
> => The where clause in all cases above should be `where order0_.[CODE] like @p0', and @p0 = '%001%' / @p0 = '001%' / @p0 = '%001'`, the one will gain more performance when the query above run many times. 
> => `StartsWith`/`EndsWith`/`Contains` methods can reuse the same SQL Execution Plan by database engine
> 
> 
> LINQ: 
> ```c#
> (from o in db.AsQueryable<Order>() where o.Code.Substring(2, 3) == "000" select o).ToList()
> ```
> 
> SQL: 
> ```sql
> select order0*.[ID] as ID1_0_, order0_.[CODE] as CODE2_0_, order0_.[COMMENT] as COMMENT3_0_ from [ORDER] order0_ where (substring(order0_.[CODE], @p0, @p1) is null) and (@p2 is null) or substring(order0*.[CODE], @p0, @p1)=@p2;@p0 = 2 [Type: Int32 (0)], @p1 = 3 [Type: Int32 (0)], @p2 = '000' [Type: String (4000)]
> ```
> 
> => The where clause should be `where length(order0*.[CODE]) >= (@p0 + @p1) and substring(order0*.[CODE], :p0, :p1) == :p2`

---
**Fabio Maulo** added a comment — <time datetime="2010-09-07T09:19:26.288Z" title="2010-09-07T09:19:26.288Z">7th September 2010, 9:19:26</time>:

> 1. The Linq-provider is available only in NH3.
> 2. Ok in case of constant-parameter

---
**Oskar Berggren** added a comment — <time datetime="2012-04-05T16:02:33.555Z" title="2012-04-05T16:02:33.555Z">5th April 2012, 16:02:33</time>:

> @tdhung80 Could you please check if these problems persist in 3.3.0CR1?

---
**Alexander Zaytsev** added a comment — <time datetime="2013-01-23T03:06:20.591Z" title="2013-01-23T03:06:20.591Z">23rd January 2013, 3:06:20</time>:

> I think it is possible to make new `ExpressionType` to handle this.

hazzik:
Fixes #725

### DIFF
--- a/src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
+++ b/src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
@@ -1,0 +1,60 @@
+using System.Linq.Expressions;
+using System.Reflection;
+using NHibernate.Util;
+using Remotion.Linq.Parsing.ExpressionVisitors.Transformation;
+
+namespace NHibernate.Linq.ExpressionTransformers
+{
+	/// <summary>
+	/// Replace <see cref="string.StartsWith(string)"/>, <see cref="string.EndsWith(string)"/> and <see cref="string.Contains(string)"/>
+	/// with <see cref="SqlMethods.Like(string, string)"/>
+	/// </summary>
+	internal class LikeTransformer : IExpressionTransformer<MethodCallExpression>
+	{
+		private static readonly MethodInfo Like = ReflectHelper.FastGetMethod(SqlMethods.Like, default(string), default(string));
+
+		public ExpressionType[] SupportedExpressionTypes { get; } = {ExpressionType.Call};
+
+		public Expression Transform(MethodCallExpression expression)
+		{
+			if (expression.Method == ReflectionCache.StringMethods.StartsWith)
+			{
+				return Expression.Call(
+					Like,
+					expression.Object,
+					Concat(expression.Arguments[0], Expression.Constant("%"))
+				);
+			}
+
+			if (expression.Method == ReflectionCache.StringMethods.EndsWith)
+			{
+				return Expression.Call(
+					Like,
+					expression.Object,
+					Concat(Expression.Constant("%"), expression.Arguments[0])
+				);
+			}
+
+			if (expression.Method == ReflectionCache.StringMethods.Contains)
+			{
+				return Expression.Call(
+					Like,
+					expression.Object,
+					Concat(Concat(Expression.Constant("%"), expression.Arguments[0]), Expression.Constant("%"))
+				);
+			}
+
+			return expression;
+		}
+
+		private static Expression Concat(Expression arg1, Expression arg2)
+		{
+			if (arg1 is ConstantExpression const1 && arg2 is ConstantExpression const2)
+			{
+				return Expression.Constant(string.Concat(const1.Value, const2.Value));
+			}
+
+			return Expression.Add(arg1, arg2, ReflectionCache.StringMethods.Concat);
+		}
+	}
+}
--- a/src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
+++ b/src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
@@ -17,7 +17,7 @@ namespace NHibernate.Linq.ExpressionTransformers
 
 		public Expression Transform(MethodCallExpression expression)
 		{
-			if (expression.Method.Name == nameof(string.StartsWith) && expression.Method == ReflectionCache.StringMethods.StartsWith)
+			if (expression.Method == ReflectionCache.StringMethods.StartsWith)
 			{
 				return Expression.Call(
 					Like,
@@ -26,7 +26,7 @@ namespace NHibernate.Linq.ExpressionTransformers
 				);
 			}
 
-			if (expression.Method.Name == nameof(string.EndsWith) && expression.Method == ReflectionCache.StringMethods.EndsWith)
+			if (expression.Method == ReflectionCache.StringMethods.EndsWith)
 			{
 				return Expression.Call(
 					Like,
@@ -35,7 +35,7 @@ namespace NHibernate.Linq.ExpressionTransformers
 				);
 			}
 
-			if (expression.Method.Name == nameof(string.Contains) && expression.Method == ReflectionCache.StringMethods.Contains)
+			if (expression.Method == ReflectionCache.StringMethods.Contains)
 			{
 				return Expression.Call(
 					Like,
--- a/src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
+++ b/src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
@@ -17,7 +17,7 @@ namespace NHibernate.Linq.ExpressionTransformers
 
 		public Expression Transform(MethodCallExpression expression)
 		{
-			if (expression.Method == ReflectionCache.StringMethods.StartsWith)
+			if (expression.Method.Name == nameof(string.StartsWith) && expression.Method == ReflectionCache.StringMethods.StartsWith)
 			{
 				return Expression.Call(
 					Like,
@@ -26,7 +26,7 @@ namespace NHibernate.Linq.ExpressionTransformers
 				);
 			}
 
-			if (expression.Method == ReflectionCache.StringMethods.EndsWith)
+			if (expression.Method.Name == nameof(string.EndsWith) && expression.Method == ReflectionCache.StringMethods.EndsWith)
 			{
 				return Expression.Call(
 					Like,
@@ -35,7 +35,7 @@ namespace NHibernate.Linq.ExpressionTransformers
 				);
 			}
 
-			if (expression.Method == ReflectionCache.StringMethods.Contains)
+			if (expression.Method.Name == nameof(string.Contains) && expression.Method == ReflectionCache.StringMethods.Contains)
 			{
 				return Expression.Call(
 					Like,
--- a/src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
+++ b/src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
@@ -17,44 +17,47 @@ namespace NHibernate.Linq.ExpressionTransformers
 
 		public Expression Transform(MethodCallExpression expression)
 		{
-			if (expression.Method == ReflectionCache.StringMethods.StartsWith)
+			if (IsLike(expression, out var value))
 			{
 				return Expression.Call(
 					Like,
 					expression.Object,
-					Concat(expression.Arguments[0], Expression.Constant("%"))
-				);
-			}
-
-			if (expression.Method == ReflectionCache.StringMethods.EndsWith)
-			{
-				return Expression.Call(
-					Like,
-					expression.Object,
-					Concat(Expression.Constant("%"), expression.Arguments[0])
-				);
-			}
-
-			if (expression.Method == ReflectionCache.StringMethods.Contains)
-			{
-				return Expression.Call(
-					Like,
-					expression.Object,
-					Concat(Concat(Expression.Constant("%"), expression.Arguments[0]), Expression.Constant("%"))
+					Expression.Constant(value)
 				);
 			}
 
 			return expression;
 		}
 
-		private static Expression Concat(Expression arg1, Expression arg2)
+		private static bool IsLike(MethodCallExpression expression, out string value)
 		{
-			if (arg1 is ConstantExpression const1 && arg2 is ConstantExpression const2)
+			if (expression.Method == ReflectionCache.StringMethods.StartsWith)
 			{
-				return Expression.Constant(string.Concat(const1.Value, const2.Value));
+				if (expression.Arguments[0] is ConstantExpression constantExpression)
+				{
+					value = string.Concat(constantExpression.Value, "%");
+					return true;
+				}
+			}
+			else if (expression.Method == ReflectionCache.StringMethods.EndsWith)
+			{
+				if (expression.Arguments[0] is ConstantExpression constantExpression)
+				{
+					value = string.Concat("%", constantExpression.Value);
+					return true;
+				}
+			}
+			else if (expression.Method == ReflectionCache.StringMethods.Contains)
+			{
+				if (expression.Arguments[0] is ConstantExpression constantExpression)
+				{
+					value = string.Concat("%", constantExpression.Value, "%");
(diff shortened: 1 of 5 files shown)


src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
```diff
@@ -0,0 +1,60 @@
...
+	/// </summary>
+	internal class LikeTransformer : IExpressionTransformer<MethodCallExpression>
+	{
+		private static readonly MethodInfo Like = ReflectHelper.FastGetMethod(SqlMethods.Like, default(string), default(string));
+
+		public ExpressionType[] SupportedExpressionTypes { get; } = {ExpressionType.Call};
+
+		public Expression Transform(MethodCallExpression expression)
+		{
+			if (expression.Method.Name == nameof(string.StartsWith) && expression.Method == ReflectionCache.StringMethods.StartsWith)
```
bahusoid, line 20: Why do you think it optimizes anything?

hazzik: What exactly do you mean?

bahusoid: I just don't understand this optimization with method name comparison. Are you aware of some  heavy internal method equals logic?

bahusoid: Quick fiddle test shows no performance gain. It's only lost:
https://dotnetfiddle.net/Di9PUY 

Did I miss something?

hazzik: Ok, reverted.

src/NHibernate/Linq/ExpressionTransformers/LikeTransformer.cs
```diff
@@ -0,0 +1,60 @@
...
+		public ExpressionType[] SupportedExpressionTypes { get; } = {ExpressionType.Call};
+
+		public Expression Transform(MethodCallExpression expression)
+		{
+			if (expression.Method == ReflectionCache.StringMethods.StartsWith)
+			{
+				return Expression.Call(
+					Like,
+					expression.Object,
+					Concat(expression.Arguments[0], Expression.Constant("%"))
```
bahusoid, line 25: How escaping works for such queries? I mean do we really properly support queries like `Where(x => x.Name.StartsWith("%"))`?

hazzik: It is not supported at the moment.

fredericDelaporte: Related: #484, #488, [NH-3726](https://nhibernate.jira.com/browse/NH-3726), 9be7d064bddcf64d37e73e6716162ef43f2a72a2.
The `StartsWith`, `EndsWith`, `Contains` cases were a comment "side note" in NH-3726 and #484. The main point of [NH-3726](https://nhibernate.jira.com/browse/NH-3726) was only about `SqlMethod.Like`. #484 and #488 were PR doing this, and it seems the corresponding fix has ended up committed directly as 9be7d06.
Anyway we lack an opened issue for supporting escaping with these methods, and as said in #484, it will be a possible breaking change.

hazzik: No, we have @fredericDelaporte. #877

src/NHibernate/Util/ReflectionCache.cs
```diff
@@ -222,5 +222,13 @@ internal static class TypeMethods
 			internal static readonly MethodInfo GetTypeFromHandle =
 				ReflectHelper.FastGetMethod(System.Type.GetTypeFromHandle, default(RuntimeTypeHandle));
 		}
+
+		internal static class StringMethods
+		{
+			public static readonly MethodInfo EndsWith = ReflectHelper.GetMethodDefinition<string>(x => x.EndsWith(null));
+			public static readonly MethodInfo StartsWith = ReflectHelper.GetMethodDefinition<string>(x => x.StartsWith(null));
+			public static readonly MethodInfo Contains = ReflectHelper.GetMethodDefinition<string>(x => x.Contains(null));
+			public static readonly MethodInfo Concat = ReflectHelper.FastGetMethod(string.Concat, default(string), default(string));
```
fredericDelaporte, line 231: This `Concat` addition is used nowhere, I do not think it should be added just in case it is used later.

hazzik: It was used. I refacrored it out. Will remove.

fredericDelaporte:
I have a concern about this change [here](https://github.com/nhibernate/nhibernate-core/issues/725#issuecomment-633203702):

> The proposed change may in some cases play badly with SQL-Server parameter sniffing, especially when putting in common cache plans for `StartsWith` with others. SQL-Server can use an index with a `StartsWith` expression where parameter sniffing would see that the parameter value is `startOfString%`, so the like will be `like N'startOfString%'`, and then the cached query plan would rely on that index. But that index use will then be useless with a subsequent parameter value of `%searchedString`, causing the query to perform badly due to re-using a inadequate query plan for that parameter value.
>
> We may have better to keep the `StartsWith` case specially handled, with a SQL in the form of `like @p1 + N'%'`. For `Contains` and `EndsWith`, having the same query plan should not cause issues with parameter sniffing.

What are your thoughts about it?

maca88:
> SQL-Server can use an index with a StartsWith expression where parameter sniffing would see that the parameter value is startOfString%

SQL-Server seems to create the same query plan, test case:
```SQL
CREATE TABLE Location(
Id NVARCHAR(5) NOT NULL,
Value NVARCHAR(5) NOT NULL,
CONSTRAINT [PK_Location] PRIMARY KEY CLUSTERED (Id ASC)
)
CREATE NONCLUSTERED INDEX [IX_Location] ON Location (Value ASC)

DECLARE @cnt INT = 0;

WHILE @cnt < 10000
BEGIN
   INSERT INTO Location(Id, Value) VALUES('A' + CAST(@cnt AS VARCHAR), 'A' + CAST(@cnt AS VARCHAR))
   SET @cnt = @cnt + 1;
END;

DECLARE @id NVARCHAR(5)
DECLARE @id2 NVARCHAR(5)
SET @id = 'A1'
SET @id2 = 'A1%'

SELECT Value FROM Location WHERE Value LIKE 'A1' + '%'
SELECT Value FROM Location WHERE Value LIKE @id + '%'
SELECT Value FROM Location WHERE Value LIKE @id2
```
The last two queries have the same query plan:
image

On PostgreSQL the same query plan is created for all three queries:
```SQL
create table location (
id varchar(5) not null constraint location_pkey primary key,
value varchar(5)
);
CREATE INDEX location_value_index ON location(value text_pattern_ops);

INSERT INTO location(id, value)
SELECT 'A' || g, 'A' || g
FROM generate_series(1, 10000) g;

EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON) SELECT Value FROM location WHERE Value LIKE 'A1' || '%'
EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON) SELECT Value FROM location WHERE Value LIKE :p1 || '%'
EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON) SELECT Value FROM location WHERE Value LIKE :p2
```

If the same applies for other databases then we don't need to handle `StartsWith` differently.

fredericDelaporte:
To check about parameter sniffing, you must try with different parameters value, with an index on the "liked" column, with some parameters values starting with `%` and some other without, and also, you must clear the query cache plan between each test. Read [here](http://www.sommarskog.se/query-plan-mysteries.html) for an in-depth explanation.

maca88:
> Read here for an in-depth explanation.

Thanks for the link. I played with different parameters by clearing query cache each time and realized that also `%p1%` (Contains) and `%p1` (EndsWith) may end out having a diffenent query plan, example (Sql Server):
```SQL
DBCC FREEPROCCACHE;
exec sp_executesql N'SELECT l.Value FROM Location l left join Location l2 on l.Id = l2.Id WHERE l.Value LIKE @id2 AND l2.Id <> @id2', N'@id2 NVARCHAR(5)', @id2='%1%'
DBCC FREEPROCCACHE;
exec sp_executesql N'SELECT l.Value FROM Location l left join Location l2 on l.Id = l2.Id WHERE l.Value LIKE @id2 AND l2.Id <> @id2', N'@id2 NVARCHAR(5)', @id2='1%'
DBCC FREEPROCCACHE;
exec sp_executesql N'SELECT l.Value FROM Location l left join Location l2 on l.Id = l2.Id WHERE l.Value LIKE @id2 AND l2.Id <> @id2', N'@id2 NVARCHAR(5)', @id2='%1'
```

in the above example all three statements have a different query plan. Now I don't quite understand the following statement from #725:

> The where clause in all cases above should be where order0_.[CODE] like @p0', and @p0 = '%001%' / @p0 = '001%' / @p0 = '%001', the one will gain more performance when the query above run many times.

From my current understanding, having one plan for all three statements won't increase the performance, infact it may only reduce it when the database engine would produce different query plans. Currently, we don't have any reported issues regarding producing too many query plans that would cause a performance drop, so maybe we shouldn't reuse the same query plan even for `Contains` and `EndsWith`.

fredericDelaporte:
Well, recomputing a query plan is slower than using an existing one, if the existing one is suitable. But this gain is in my view marginal against the gain of using a suitable plan.
Also for applications generating a lot of various queries, using more different query plans than needed may create excessive pressure on the query plan cache, if the SQL server is "small". (SQL Express, limited memory available, think about cloud deployment where the ops tries to have the smaller server possible for minimizing the exploitation cost, ... I am not on that side personally currently, the company I currently works for has a "huge over-sizing" policy about servers, but that is not the most popular way of handling things these days.)

maca88:
> Well, recomputing a query plan is slower than using an existing one, if the existing one is suitable

Correct, I forgot about that. So these are the current differences:

Current master:
[+] Better performance when queries have different query plans
[-] Additional query plans

This PR:
[+] Less query plans, which means less memory usage
[+] Faster first time execution when reusing an existing plan (e.g. query using `EndsWith` would reuse a plan created by `Contains` method)
[-] Decrease in performance when queries would have different plans

For me the benefit of having a different query plan for each method outweighs the benefits from this PR, as a bad query plan may cause the query to run multiple times slower. In order to support both scenarios, I made an alternative PR #2411, that adds the ability to configure the pre-transformation by the user.

hazzik:
Closing in favor of #2411

CetinBasoz:
This discussion seems to be the problem I am having :( We are using an old version of nHibernate where it generates StartsWith as:
@p + '%'

While on SQL server it may seem that both @p+'%' and the one that include '%' in parameter generating the same execution plan, SQL server generates a plan per call !

fredericDelaporte:
Have you read [this](https://github.com/nhibernate/nhibernate-core/issues/725#issuecomment-646777527) and tried using the new feature introduced by #2411 to solve the issue you encounter?

PR #2399, opened 2020-05-25
Optimize PersistentGenericBag.EqualsSnapshot

bahusoid:
Added tests directly calling `EqualsSnapshot` to verify it works correctly.

Notes: This check is called when NHibernate detects cases when underlying collection can be modified directly in user code and not via `IPersistenCollection` wrapper, for instance array mappings and `PersistentArryaHolder`.
For bags it's executed if new instance is created with bag then flushed (so `List<Entity>` become `PersistentGenericBag`).  Check is executed for all additional flushes in this session as theoretically user can still have instance of original bag `List<Entity>` not wrapped in `PersistentGenericBag` that can be modified directly

### DIFF
--- a/src/NHibernate/Collection/Generic/PersistentGenericBag.cs
+++ b/src/NHibernate/Collection/Generic/PersistentGenericBag.cs
@@ -295,12 +295,15 @@ namespace NHibernate.Collection.Generic
 				return false;
 			}
 
-			foreach (var elt in _gbag)
+			for (var i = 0; i < _gbag.Count; i++)
 			{
-				if (CountOccurrences(elt, _gbag, elementType) != CountOccurrences(elt, sn, elementType))
-				{
+				if(elementType.IsSame(_gbag[i], sn[i]))
+					continue;
+
+				var elt = _gbag[i];
+				var countInSnapshot = CountOccurrences(elt, sn, elementType);
+				if (countInSnapshot == 0 || CountOccurrences(elt, _gbag, elementType) != countInSnapshot)
 					return false;
-				}
 			}
 
 			return true;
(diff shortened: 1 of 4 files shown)


src/NHibernate/Collection/Generic/PersistentGenericBag.cs
```diff
@@ -295,12 +295,15 @@ public override bool EqualsSnapshot(ICollectionPersister persister)
 				return false;
 			}
 
-			foreach (var elt in _gbag)
+			for (var i = 0; i < _gbag.Count; i++)
 			{
-				if (CountOccurrences(elt, _gbag, elementType) != CountOccurrences(elt, sn, elementType))
-				{
+				if(elementType.IsSame(_gbag[i], sn[i]))
```
bahusoid, line 300: No need to count elements placed on the same positions.

src/NHibernate/Collection/Generic/PersistentGenericBag.cs
```diff
@@ -295,12 +295,15 @@ public override bool EqualsSnapshot(ICollectionPersister persister)
...
+					continue;
+
+				var elt = _gbag[i];
+				var countInSnapshot = CountOccurrences(elt, sn, elementType);
```
bahusoid, line 304: Added shortcut if element is no longer present in snapshot

src/NHibernate/Collection/Generic/PersistentGenericBag.cs
fredericDelaporte: New white space glitch after if.

bahusoid:
> But I was thinking about something much heavier : building dictionaries of bag elements. Not sure it is worth it.

I don't think it's worth for bags. As I tried to explain in PR description this check is rarely executed and when executed in most cases it's unmodified collection. So added `IsSame` shortcut covers most use cases.

PR #2401, opened 2020-05-26
Optimize JoinWalker.WhereString method

hazzik:
It was generating a lot of memory pressure. I've just inlined all objects creation and removed conditions which are always false.

### DIFF
--- a/src/NHibernate/Loader/JoinWalker.cs
+++ b/src/NHibernate/Loader/JoinWalker.cs
@@ -980,42 +980,96 @@ namespace NHibernate.Loader
 			{
 				// if not a composite key, use "foo in (?, ?, ?)" for batching
 				// if no batch, and not a composite key, use "foo = ?"
-				string tableAlias = GenerateAliasForColumn(alias, columnNames[0]);
-				InFragment inf = new InFragment().SetColumn(tableAlias, columnNames[0]);
+				var columnName = columnNames[0];
 
-				for (int i = 0; i < batchSize; i++)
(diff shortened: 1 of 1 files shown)


src/NHibernate/Loader/JoinWalker.cs
```diff
@@ -980,42 +980,96 @@ protected SqlStringBuilder WhereString(string alias, string[] columnNames, int b
...
+					whereString.Add(" in (");
+					for (var i = 0; i < batchSize; i++)
+					{
+						var value = Parameter.Placeholder;
+						if (added)
+						{
+							whereString.Add(StringHelper.CommaSpace);
+						}
 
-				return new SqlStringBuilder(inf.ToFragmentString());
```
bahusoid, line 989: Maybe we should instead extend `InFragment`  with provided `SqStringBuilder` in ctor or add property to return underlying builder? Maybe the same can be done with `DisjunctionFragment`.

Also `InFragment` contains some logic regarding `null`/`not null` values handling. It's now missed - though not sure if it's applicable  here.

hazzik: > though not sure if it's applicable here.

Not applicable. The values are always parameter placeholder, which is never null.

hazzik: > Maybe we should instead extend InFragment with provided SqStringBuilder in ctor or add property to return underlying builder? Maybe the same can be done with DisjunctionFragment.

Current logic creates a lot of unnecessary objects and arrays, not just SqlStringBuilder.

PR #2402, opened 2020-05-26
Fix Oracle binary floating point setting

maca88:
Fix for the setting introduced by #2349. I was testing the mentioned PR with `CanGetAverageOfIntegersAsDouble` method, but I didn't noticed that it works even without having the setting enabled.

### DIFF
--- a/src/NHibernate/Dialect/Oracle10gDialect.cs
+++ b/src/NHibernate/Dialect/Oracle10gDialect.cs
@@ -26,12 +26,12 @@ namespace NHibernate.Dialect
 
 		public override void Configure(IDictionary<string, string> settings)
 		{
-			base.Configure(settings);
-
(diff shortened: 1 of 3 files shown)


src/NHibernate/Dialect/Oracle10gDialect.cs
```diff
@@ -26,12 +26,12 @@ public override JoinFragment CreateOuterJoinFragment()
...
 		public override void Configure(IDictionary<string, string> settings)
 		{
-			base.Configure(settings);
-
 			_useBinaryFloatingPointTypes = PropertiesHelper.GetBoolean(
 				Environment.OracleUseBinaryFloatingPointTypes,
 				settings,
 				false);
+
+			base.Configure(settings);
```
maca88, line 34: Method `RegisterFloatingPointTypeMappings` is called inside `base.Configure`, so the field must be set before calling it.

PR #2406, opened 2020-05-27
Make Render method virtual for Cast function

### LINKED ISSUE #2404: Cannot override default CastFunction
NHibernate Version: 4.0.4.4000
Versions Affected: **4.0.x,** **5.0.x**, **5.1.x**, **5.2.x**, **master**

Calling RegisterFunction("cast", ...) with anything but an instance of CastFunction results in an incorrect return type. The following lines in SessionFactoryHelperExtension explicitly check if the sql function is of type CastFunction.

[SessionFactoryHelperExtension:85](https://github.com/nhibernate/nhibernate-core/blob/14fef3f829141a398eb37a01ffabd21266d382fe/src/NHibernate/Hql/Ast/ANTLR/SessionFactoryHelperExtensions.cs#L85)
[SessionFactoryHelperExtension:115](https://github.com/nhibernate/nhibernate-core/blob/14fef3f829141a398eb37a01ffabd21266d382fe/src/NHibernate/Hql/Ast/ANTLR/SessionFactoryHelperExtensions.cs#L115)

The problem with this type check is that the implementation of CastFunction cannot be changed by implementing a derived class as none of the implemented interface methods are virtual.

Some possible solutions are to:
1. Check if the `functionName` parameter is equal to `"cast"`.
2. Make interface methods of CastFunction virtual so a derived class can override them.
3. Add an abstract base class that CastFunction is derived from and change the type checks accordingly.

See the implementation of [JetCastFunction](https://github.com/deAtog/nhcontrib/blob/master/src/NHibernate.JetDriver/src/NHibernate.JetDriver/JetCastFunction.cs) for why a Dialect would want to implement their own CastFunction.

maca88:
In order to override it in case the dialect does not support cast function.

Fixes #2404

### DIFF
--- a/src/NHibernate/Dialect/Function/CastFunction.cs
+++ b/src/NHibernate/Dialect/Function/CastFunction.cs
@@ -54,7 +54,7 @@ namespace NHibernate.Dialect.Function
 			get { return true; }
 		}
 
-		public SqlString Render(IList args, ISessionFactoryImplementor factory)
+		public virtual SqlString Render(IList args, ISessionFactoryImplementor factory)
 		{
 			if (args.Count != 2)
 			{
@@ -87,13 +87,13 @@ namespace NHibernate.Dialect.Function
 				throw new QueryException(string.Format("invalid Hibernate type for cast(): type {0} not found", typeName));
 			}
 
-			//TODO 6.0: Remove pragma block with its content
+			// TODO 6.0: Remove pragma block with its content
 #pragma warning disable 618
 			if (!CastingIsRequired(sqlType))
 				return new SqlString("(", args[0], ")");
 #pragma warning restore 618
 
-			return Render(args, sqlType, factory);
+			return Render(args[0], sqlType, factory);
 		}
 
 		#endregion
@@ -108,13 +108,13 @@ namespace NHibernate.Dialect.Function
 		/// <summary>
 		/// Renders the SQL fragment representing the SQL cast.
 		/// </summary>
-		/// <param name="args">The cast arguments.</param>
+		/// <param name="expression">The cast argument.</param>
 		/// <param name="sqlType">The SQL type to cast to.</param>
 		/// <param name="factory">The session factory.</param>
 		/// <returns>A SQL fragment.</returns>
-		protected virtual SqlString Render(IList args, string sqlType, ISessionFactoryImplementor factory)
+		protected virtual SqlString Render(object expression, string sqlType, ISessionFactoryImplementor factory)
 		{
-			return new SqlString("cast(", args[0], " as ", sqlType, ")");
+			return new SqlString("cast(", expression, " as ", sqlType, ")");
 		}
 
 		#region IFunctionGrammar Members
--- a/src/NHibernate/Dialect/Function/CastFunction.cs
+++ b/src/NHibernate/Dialect/Function/CastFunction.cs
@@ -87,11 +87,13 @@ namespace NHibernate.Dialect.Function
 				throw new QueryException(string.Format("invalid Hibernate type for cast(): type {0} not found", typeName));
 			}
 
+			//TODO 6.0: Remove pragma block with its content
 #pragma warning disable 618
-			return CastingIsRequired(sqlType)
+			if (!CastingIsRequired(sqlType))
+				return new SqlString("(", args[0], ")");
 #pragma warning restore 618
-				? Render(args, sqlType, factory)
-				: new SqlString("(", args[0], ")");
+
+			return Render(args, sqlType, factory);
 		}
 
 		#endregion
--- a/src/NHibernate/Dialect/Function/CastFunction.cs
+++ b/src/NHibernate/Dialect/Function/CastFunction.cs
@@ -87,11 +87,17 @@ namespace NHibernate.Dialect.Function
 				throw new QueryException(string.Format("invalid Hibernate type for cast(): type {0} not found", typeName));
 			}
 
-			return CastingIsRequired(sqlType) ? Render(args, sqlType, factory) : new SqlString("(", args[0], ")");
+#pragma warning disable 618
+			return CastingIsRequired(sqlType)
+#pragma warning restore 618
+				? Render(args, sqlType, factory)
+				: new SqlString("(", args[0], ")");
 		}
 
 		#endregion
 
+		// Since v5.3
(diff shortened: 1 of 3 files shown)


src/NHibernate/Dialect/Function/CastFunction.cs
```diff
@@ -87,14 +87,7 @@ public SqlString Render(IList args, ISessionFactoryImplementor factory)
...
 
-			if (CastingIsRequired(sqlType))
-			{
-				return new SqlString("cast(", args[0], " as ", sqlType, ")");
-			}
-			else
-			{
-				return new SqlString("(", args[0], ")");
-			}
+			return CastingIsRequired(sqlType) ? Render(args, sqlType, factory) : new SqlString("(", args[0], ")");
```
deAtog, line 90: With Render being virtual, you should remove CastingIsRequired and override Render in the TransparentCastFunction class to avoid this check.

deAtog: Nevermind, I didn't see that SQLiteCastFunction overrides CastingIsRequired as well and returns true/false based on the sqlType. I think these changes are probably fine.

deAtog: The SQLiteCastFunction class is slated to be removed after v5.3 so maybe the refactoring I suggest above should be done at that time as well.

maca88: Made `CastingIsRequired` obsolete and overridden the `Render` method in `TransparentCastFunction`.

src/NHibernate/Dialect/Function/CastFunction.cs
```diff
@@ -87,23 +87,27 @@ public SqlString Render(IList args, ISessionFactoryImplementor factory)
...
-			}
-			else
-			{
-				return new SqlString("(", args[0], ")");
-			}
+#pragma warning disable 618
+			return CastingIsRequired(sqlType)
+#pragma warning restore 618
+				? Render(args, sqlType, factory)
+				: new SqlString("(", args[0], ")");
```
bahusoid, line 94: To simplify 6.0 code clean up I think we should try to put full obsolete branch of code to `#pragma` block.
Something like:
```C#
//TODO 6.0: Remove pragma block with it's content
#pragma warning disable 618
if (!CastingIsRequired(sqlType))
    return new SqlString("(", args[0], ")");
#pragma warning restore 618

return Render(args, sqlType, factory);
```

maca88: Done as suggested.

src/NHibernate/Dialect/Function/CastFunction.cs
```diff
@@ -87,23 +87,29 @@ public SqlString Render(IList args, ISessionFactoryImplementor factory)
...
 		#endregion
 
+		// Since v5.3
+		[Obsolete("This method has no usages and will be removed in a future version")]
 		protected virtual bool CastingIsRequired(string sqlType)
 		{
 			return true;
 		}
 
+		protected virtual SqlString Render(IList args, string sqlType, ISessionFactoryImplementor factory)
```
fredericDelaporte, line 108: We should xml document members meant to be used or overridden by user code.

```suggestion
		/// <summary>
		/// Renders the SQL fragment representing the SQL cast.
		/// </summary>
		/// <param name="args">The cast arguments.</param>
		/// <param name="sqlType">The SQL type to cast to.</param>
		/// <param name="factory">The session factory.</param>
		/// <returns>A SQL fragment.</returns>
		protected virtual SqlString Render(IList args, string sqlType, ISessionFactoryImplementor factory)
```

I wonder if we should replace `args` with just an `expression` parameter with `args[0]` into it. It is less versatile, in case someone would want to recompute differently the `sqlType`. But for this need I say we should instead make the main `Render` virtual too. While here we have a kind of redundancy between `args` and `sqlType`, causing `args[1]` to be useless in all current cases.

src/NHibernate/Dialect/Function/TransparentCastFunction.cs
```diff
@@ -8,9 +11,16 @@ namespace NHibernate.Dialect.Function
...
 	public class TransparentCastFunction : CastFunction
 	{
+		// Since v5.3
+		[Obsolete("This method has no usages and will be removed in a future version")]
 		protected override bool CastingIsRequired(string sqlType)
 		{
 			return false;
 		}
+
+		protected override SqlString Render(IList args, string sqlType, ISessionFactoryImplementor factory)
```
fredericDelaporte, line 21:
```suggestion
		/// <summary>
		/// Renders the SQL fragment representing the casted expression without actually casting it.
		/// </summary>
		/// <param name="args">The cast arguments.</param>
		/// <param name="sqlType">The SQL type to cast to, ignored for rendering.</param>
		/// <param name="factory">The session factory.</param>
		/// <returns>A SQL fragment.</returns>
		protected override SqlString Render(IList args, string sqlType, ISessionFactoryImplementor factory)
```

src/NHibernate/Dialect/Function/CastFunction.cs
maca88: > I wonder if we should replace args with just an expression parameter with args[0] into it.

I think we should. Replaced `args` with `expression` and made the main `Render` method virtual.

hazzik:
I'm not really sure, but maybe we should remove this special treatment of the CAST function from HQL?

maca88:
The only way we could remove it is by modifiyng `GetReturnType` first parameter type from `IEnumerable<IType>` to `IEnumerable<IASTNode>` so that each `ISQLFunction` would need to cast `IASTNode` nodes to `SqlNode` except for `CastFunction`, which has a special logic. Casting `IASTNode` to `IType` could be moved to a helper method in order to avoid duplicate code. As `GetReturnType` is not yet released, we could change it prior releasing `5.3`. What are your thoughts about it?

deAtog:
Personally, I don't think the CastFunction ought to have been treated differently from any of the others. I like the changes introduced here as it allows code reuse, E.g. The JetCastFunction can be updated to extend the CastFunction class and implement only the new virtual Render method. However, it leaves the original issue that if a dialect registers a class that does not derive from CastFunction then the code will not behave in an expected manner. Regardless, I don't see a way to fix the stated issue in a backwards compatible manner, i.e. without making changes to the original JetCastFunction.

PR #2416, opened 2020-06-09
Fix querying with delayed entity insert

### LINKED ISSUE #2414: Error when query with delayed insert entity in where expression
Hi,
Throw System.InvalidCastException: Unable to cast object of type 'NHibernate.Action.DelayedPostInsertIdentifier' to type 'System.IConvertible'

```
    var entity = new Entity();
    session.Persist(entity);
    var entities= session.Query<Entity>().Where(x => x == entity).ToList();
```

galamai:
Fix #2414


src/NHibernate/Type/NullableType.cs
```diff
@@ -152,6 +153,15 @@ public sealed override void NullSafeSet(DbCommand st, object value, int index, I
...
 				st.Parameters[index].Value = DBNull.Value;
 			}
+			else if (value is DelayedPostInsertIdentifier)
+			{
+				if (IsDebugEnabled)
+				{
+					Log.Debug("binding 0 to parameter: {0}", index);
+				}
+
+				st.Parameters[index].Value = 0;
```
fredericDelaporte, line 163: This assumes the identifier is compatible with an `int`, and this assumes `0` is not a valid value for it. This will be wrong in many cases.

galamai: DelayedPostInsertIdentifier exists only for auto-increment fields, so I believe that the value 0 in this case is correct

galamai: I compared this behavior with entity framework core 3 comparison by reference for an entity that is not yet stored in the table returns exactly parameter 0 in all cases

bahusoid: What if `entity.Id` is not 0 for transient entity and for instance it's -1?
Your fix is incorrect for the following SQL Server table:
```SQL
CREATE TABLE Entity
(  
Id int IDENTITY(0,1), -- start from 0
Name varchar (20) 
 );  
```

hazzik:
With changes from #3043 now tests fail.

PR #2418, opened 2020-06-16
Use BestGuessEntityName instead of GuessEntityName

othman-fayed:
GuessEntityName fails to retrieve the EntityName of *ProxyFieldInterceptor.
Switch to use BestGuessEntityName instead since it already handles this case.


fredericDelaporte:
Thanks,

May you a a test demonstrating in which cases it makes a difference?

PR #2431, opened 2020-07-15
Document the configuration provider

fredericDelaporte:
Add some documentation about the new configuration provider.

We should think about documenting new features a bit more in my opinion.


fredericDelaporte:
Requalified as a fix since the documented feature has been added in the same milestone.

PR #2435, opened 2020-07-19
Fix iconUrl warning

fredericDelaporte:
Fix the `iconUrl` warning occurring on NuGet push.
See https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageiconurl


hazzik:
Oops. Should have gone to master.

fredericDelaporte:
It was intended to target 5.3.x.
I expect we will have some fixes to do, and then it will be merged to master once 5.3.1 will be released.

PR #2448, opened 2020-07-23
Avoid lambda compilation for constant and member access expressions in Criteria

bahusoid:
```C#
//Example of constant
query.Where(x => x.Id == 1);

//Example of member access
var paramId = 1;
query.Where(x => x.Id == paramId)
```

### DIFF
--- a/src/NHibernate/Impl/ExpressionProcessor.cs
+++ b/src/NHibernate/Impl/ExpressionProcessor.cs
@@ -252,6 +252,26 @@ namespace NHibernate.Impl
 		/// </summary>
 		public static object FindValue(Expression expression)
 		{
+			if (expression.NodeType == ExpressionType.Constant)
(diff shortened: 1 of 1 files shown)


hazzik:
@bahusoid is performance gain significant?

bahusoid:
@hazzik Benchmark for slower member access case (`Where(x => x.Id == paramId)`):
```
| Method |       Mean |     Error |     StdDev | Ratio |
|------- |-----------:|----------:|-----------:|------:|
|    New |   1.562 us | 0.0283 us |  0.0236 us | 0.007 |
|    Old | 233.980 us | 4.6614 us | 12.1980 us | 1.000 |

  1 us   : 1 Microsecond (0.000001 sec)
```

Source: https://gist.github.com/bahusoid/c68cb6b81f8dde9252d0cc67229136b4

PR #2450, opened 2020-07-24
Release 5.3.1

fredericDelaporte:
In case it is wished to release these 5.3.1 fixes soon, here is the the release PR.


hazzik:
Should we release today?

fredericDelaporte:
We could wait in case there are more regressions to fix, but it could also hinder detecting those other regressions due to failing before with the fixed ones, unless using the MyGet dev build.

So I am agree with releasing it today.
I will do this in 2 hours I think, unless someone ask for waiting for some reasons in the mean time.

fredericDelaporte:
By the way, the release announcements on the nhuser group are not yet approved. It seems I am administrator only of the nhdevelopement group, and so someone else has to approve my announcements on nhuser.

PR #2455, opened 2020-07-27
#2454 fix ConditionalProjection GetType from outer query

delta-emil:
This should fix #2454.

The AsyncGenerator was failing for me (I'll open another issue about that) from the build menu, so I wrote the async version of the test by hand.

PR #2460, opened 2020-07-29
Simplify single alias retrieval for SimpleProjections

bahusoid:
There is no need to parse projection just to retrieve single alias for projections based on `SimpleProjection`.
See https://github.com/nhibernate/nhibernate-core/pull/2455#issuecomment-664922877

### DIFF
--- a/src/NHibernate.Test/Criteria/AddNumberProjection.cs
+++ b/src/NHibernate.Test/Criteria/AddNumberProjection.cs
@@ -35,7 +35,7 @@ namespace NHibernate.Test.Criteria
 				.Add(" + ")
 				.Add(criteriaQuery.NewQueryParameter(typedValue).Single())
 				.Add(") as ")
-				.Add(GetColumnAlias(0, 0 ))
+				.Add(GetColumnAlias(0))
 				.ToSqlString();
 		}
 
--- a/src/NHibernate.Test/Criteria/AddNumberProjection.cs
+++ b/src/NHibernate.Test/Criteria/AddNumberProjection.cs
@@ -35,7 +35,7 @@ namespace NHibernate.Test.Criteria
 				.Add(" + ")
 				.Add(criteriaQuery.NewQueryParameter(typedValue).Single())
 				.Add(") as ")
-				.Add(GetColumnAliases(0, criteria, criteriaQuery)[0])
+				.Add(GetColumnAlias(0, 0 ))
 				.ToSqlString();
 		}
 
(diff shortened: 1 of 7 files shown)


src/NHibernate/Criterion/SimpleProjection.cs
```diff
@@ -26,11 +26,18 @@ public virtual IType[] GetTypes(string alias, ICriteria criteria, ICriteriaQuery
 			return null;
 		}
 
+		// Since v5.3
+		[Obsolete("This method has no more usage in NHibernate and will be removed in a future version.")]
```
bahusoid, line 30: Not sure why it's been left. But it doesn't work properly for multi-column projections. Maybe to allow to retrieve default alias name outside of criteria context? So let's expose instead `GetColumnAlias(int position, int columnIndex)`

src/NHibernate.Test/Criteria/AddNumberProjection.cs
```diff
@@ -35,7 +35,7 @@ public override SqlString ToSqlString(ICriteria criteria, int position, ICriteri
 				.Add(" + ")
 				.Add(criteriaQuery.NewQueryParameter(typedValue).Single())
 				.Add(") as ")
-				.Add(GetColumnAliases(0, criteria, criteriaQuery)[0])
+				.Add(GetColumnAlias(0, 0 ))
```
hazzik, line 38: Why do you need second 0 here? It is almost always `0`.

src/NHibernate/Criterion/SimpleProjection.cs
```diff
@@ -26,11 +26,18 @@ public virtual IType[] GetTypes(string alias, ICriteria criteria, ICriteriaQuery
...
 		public virtual string[] GetColumnAliases(int loc)
 		{
-			return new string[] {"y" + loc + "_"};
+			return new[] {GetColumnAlias(loc, 0)};
 		}
-		
+
+		public virtual string GetColumnAlias(int position, int columnIndex)
```
hazzik, line 36: Make this non virtual protected 
Call `GetColumnAliases(int loc)`

src/NHibernate.Test/Criteria/AddNumberProjection.cs
bahusoid: To allow `GetColumnAlias` method to change multicolumn projection aliases in user overrides. With you suggestion to get rid of virtual - it's not useful so I removed it.

src/NHibernate/Criterion/SimpleProjection.cs
bahusoid: Done

Issue #2457, opened 2020-07-28
Regression: Session.Refresh on entity with lazy collection set throws LazyInitializationException


deAtog:
After upgrading to 5.3.1 from 4.0.4 we now get a LazyInitializationException when trying to refresh an entity with a lazy collection. The reason for the exception given is: _no session or session was closed_ Minimal pseudo code to reproduce the issue is as follows:

```C#
public function CanRefreshEntity(int id) {
    using (ISession s = session_factory.OpenSession()) {
        EntityWithLazyCollection entity;

        using (ITransaction t = s.BeginTransaction()) {
            entity = s.Get<EntityWithLazyCollection>(id);
        }

        using (ITransaction t = s.BeginTransaction()) {
            // LazyInitializationException thrown here..
           s.Refresh(entity);
        }
    }
}
```

Pseudo xml mapping we are using is as follows:
```xml
<class name="EntityWithLazyCollection">
   <id name="Id" column="ParentId" generator="identity"/>
   <set name="Children" lazy="true" inverse="true" cascade="save-update">
      <key unique="true">
         <column name="ParentId" not-null="true"/>
      </key>
      <one-to-many class="Child"/>
   </set>
</class>
```

bahusoid:
I can't reproduce it. Exception stack trace?
And full reproducible example (using for instance https://github.com/nhibernate/nhibernate-core-testcase) would also be good

deAtog:
Stack trace is as follows:

```C#
at NHibernate.Collection.AbstractPersistentCollection.ThrowLazyInitializationException(String message)
at NHibernate.Collection.AbstractPersistentCollection.ThrowLazyInitializationExceptionIfNotConnected()
at NHibernate.Collection.AbstractPersistentCollection.Initialize(Boolean writing)
at NHibernate.Collection.AbstractPersistentCollection.Read()
at NHibernate.Collection.Generic.PersistentGenericSet`1.Equals(Object other)
at System.Object.Equals(Object objA, Object objB)
at Domain.EntityWithLazyCollection.set_Children(ISet`1 value)
at (Object , Object[] , SetterCallback )
at NHibernate.Bytecode.Lightweight.AccessOptimizer.SetPropertyValues(Object target, Object[] values)
at NHibernate.Tuple.Entity.PocoEntityTuplizer.SetPropertyValuesWithOptimizer(Object entity, Object[] values)
[...]

PR #2461, opened 2020-07-29
Remove persistent collections Equals/GetHashCode overrides

bahusoid:
To make behavior consistent with standard collections.
Related to #2457

### DIFF
--- a/src/NHibernate/Collection/Generic/PersistentGenericList.cs
+++ b/src/NHibernate/Collection/Generic/PersistentGenericList.cs
@@ -238,23 +238,6 @@ namespace NHibernate.Collection.Generic
 			return entry != null;
 		}
 
-		public override bool Equals(object obj)
-		{
-			var that = obj as ICollection<T>;
-			if (that == null)
-			{
-				return false;
-			}
-			Read();
-			return CollectionHelper.SequenceEquals(WrappedList, that);
-		}
-
-		public override int GetHashCode()
-		{
-			Read();
-			return WrappedList.GetHashCode();
-		}
-
 		#region IList Members
 
 		int IList.Add(object value)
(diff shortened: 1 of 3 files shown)


fredericDelaporte (commented):
I am not convinced aligning with standard collections on this matter is worth the possible breaking change for application code currently relying on NHibernate `Equals` overrides for these collections.

As showcased by #2457, these overrides may cause in some cases failures, but it does not seem to be a frequently annoying pitfall, judging by the number of issues I have seen related to this. (#2457 is the first and only I have seen currently.)

(Due to this I am not decided on which type label it should have. Improvement? Task?)

bahusoid:
> I am not convinced aligning with standard collections on this matter is worth the possible breaking change for application code currently relying on NHibernate Equals overrides for these collections.

I'm pretty sure that the fact that `Equals` call make collection be eager loaded is not noticed by most users  but it's not desired behavior for most of them. As it's stated in the issue - they have such problem for long time unnoticed and only fail shows them the issue.

Also this behavior is not consistent with our other persistent collections - for instance there is no override for bags.

Also the fact that behavior is different for persistent and transient entity is also disturbing.

Maybe Equals/GetHashcode methods really need to be overridden but with the following logic:
If it's initialized call E Equals/GetHashcode on wrapped collection instead. If it's not - call reference equality.

But as first step - get rid of object equality I consider as good change anyway.

fredericDelaporte:
> I'm pretty sure that the fact that `Equals` call makes collection be eager loaded is not noticed by most users but it's not desired behavior for most of them.

Well ok.

fredericDelaporte (approved):

hazzik (approved):

PR #2464, opened 2020-07-29
Fix entity join for entity with discriminator

### LINKED ISSUE #2463: Path expected for join
After I updated to the latest version 5.3.1, all Joins made with linq where the table is mapped with xml and contains discriminator, simply stop working, this just for Join, because when I do just one ToList only in the table it works normally.

bahusoid:
Fixes #2463

I wonder why it's still not fixed in hibernate. They seem behave the same - [discriminator replaces text node](
https://github.com/hibernate/hibernate-orm/blob/20273b81ee623d74d4c3d8efed2e7f2ab2f79c4e/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/util/LiteralProcessor.java#L81-L85) and then used for [persister retrieval](https://github.com/hibernate/hibernate-orm/blob/20273b81ee623d74d4c3d8efed2e7f2ab2f79c4e/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/HqlSqlWalker.java#L453-L457)


fredericDelaporte:
Is this a 5.3 regression? Or an older one? #2463 does not tell in which version it was working, and no one has confirmed this was a regression.

bahusoid:
Yes it's regression. Entity Join is implemented in 5.3

PR #2474, opened 2020-08-10
Fetch all lazy properties when entity is already loaded fails

danko-d:
Fix for the new test `TestFetchAllPropertiesAfterEntityIsInitialized` in `FetchLazyPropertiesFixture`.

Prior to this fix, the second query below would fail:
```
Person person;
using(var s = OpenSession())
{
    person = s.CreateQuery("from Person where Id = 1").UniqueResult<Person>();
    var image = person.Image;
    person = s.CreateQuery("from Person fetch all properties where Id = 1").UniqueResult<Person>();
}
```
The problem was with the `indexes` and `lazyIndexes` arrays in  `InitializeLazyProperties()` of `AbstractEntityPersister` when `allLazyProperties` was `true`. This could, in turn, cause `IndexOutOfRangeException` or `InvalidCastException`.

### DIFF
--- a/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
+++ b/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
@@ -1408,7 +1408,6 @@ namespace NHibernate.Persister.Entity
 			{
 				throw new HibernateException($"Entity is not associated with the session: {id}");
 			}
-
 		
 			var lazyProperties = allLazyProperties ? LazyProperties.ToArray() : uninitializedLazyProperties;
 			var metadata = InstrumentationMetadata.LazyPropertiesMetadata;
@@ -1420,7 +1419,6 @@ namespace NHibernate.Persister.Entity
 				indexes[i] = descriptor.PropertyIndex;
 				lazyIndexes[i] = descriptor.LazyIndex;
 			}
-		
 
 			var values = Hydrate(rs, id, entity, suffixedPropertyColumns, null, true, indexes, session);
 			for (var i = 0; i < lazyIndexes.Length; i++)
--- a/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
+++ b/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
@@ -1409,24 +1409,18 @@ namespace NHibernate.Persister.Entity
 				throw new HibernateException($"Entity is not associated with the session: {id}");
 			}
 
-			int[] indexes;
-			int[] lazyIndexes;
-			if (allLazyProperties)
+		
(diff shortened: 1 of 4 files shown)


src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -1408,24 +1408,16 @@ public virtual object InitializeLazyProperty(string fieldName, object entity, IS
 			{
 				throw new HibernateException($"Entity is not associated with the session: {id}");
 			}
-
-			int[] indexes;
-			int[] lazyIndexes;
-			if (allLazyProperties)
-			{
-				lazyIndexes = indexes = lazyPropertyNumbers;
```
bahusoid, line 1416: > The problem was with the indexes and lazyIndexes arrays 

What problem do you see with the `indexes`?  I see it only for `lazyIndexes`. Can't fix be  simplified to something like this:
```C#
if (allLazyProperties)
{
	indexes = lazyPropertyNumbers;
	lazyIndexes = new int[lazyPropertyNumbers.Length];
	for (int i = 0; i < lazyIndexes.Length; i++)
	{
		lazyIndexes[i] = i;
	}
}
```

danko-d: You're right; there's no problem with `indexes`. I've simplified the fix.

maca88:
@danko-d thanks for the fix, which fixes a regression made by #1922.

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -1427,7 +1432,7 @@ public virtual object InitializeLazyProperty(string fieldName, object entity, IS
 					lazyIndexes[i] = descriptor.LazyIndex;
 				}
 			}
-
+
```
maca88, line 1435: This line should be reverted.
```suggestion

```

bahusoid:
@danko-d Since it's regression can you please rebase it on 5.3.x branch and force push

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
maca88: By revert I meant to remove the added tabs and leave the new line as it was prior this PR. `InitializeLazyProperties` method must have the same structure as its async counterpart  `InitializeLazyPropertiesAsync`, so that when the async generator is re-run there will be no changes. Now when running the async generator it will make a change in the `InitializeLazyPropertiesAsync`, which we want to avoid.

danko-d: I only committed your suggestion which removed the line. 
Should be fixed now. Thanks!

PR #2480, opened 2020-08-13
Modification to allow Linq subqueries to have paging (OrderBy/Take)

### LINKED ISSUE #2479: When using a paged sub-query in Linq, generates incorrect SQL
I have an top level query that I control paging on, then use it along with `ToFuture()` to pull various related lists, then combine at the client.

The top level query has a `.Top()` along with `.OrderBy()`. When I use it in a `.Any()` sub-query, the `ORDER BY` gets duplicated into the `EXISTS`/`IN` query and SQL fails.

Something like this:

```cs
	[TestFixture]
	public class WhereSubqueryTests : LinqTestCase
	{
		[Test]
		public void OrdersWithSubquery9()
		{
			var query = db.Orders
			              .Where(x => x.Employee.EmployeeId > 5)
			              .OrderBy(x => x.OrderId)
			              .Take(2);

			var listFuture = db.OrderLines
			                   .Where(x => query.Any(o => o == x.Order))
			                   .OrderBy(x => x.Id)
			                   .ToFuture();

			var results = query.ToFuture().ToList();
			var listResults = listFuture.ToList();

			Assert.That(results.Count, Is.EqualTo(2), nameof(results));
			Assert.That(listResults.Count, Is.EqualTo(6), nameof(listResults));
		}
	}
```

This results in the following error:

>System.Data.SqlClient.SqlException : The ORDER BY clause is invalid in views, inline functions, derived tables, subqueries, and common table expressions, unless TOP, OFFSET or FOR XML is also specified.

The generated SQL looked like this:

```sql
select orderline0_.OrderLineId as orderlineid1_4_, orderline0_.OrderId as orderid2_4_, orderline0_.ProductId as productid3_4_, orderline0_.UnitPrice as unitprice4_4_, orderline0_.Quantity as quantity5_4_, orderline0_.Discount as discount6_4_
from OrderLines orderline0_
where exists (
    select order1_.OrderId 
    from Orders order1_ 
    where order1_.OrderId=orderline0_.OrderId and (order1_.OrderId in (
        select TOP (?) order2_.OrderId 
        from Orders order2_ where order2_.EmployeeId=? 
        order by order2_.OrderId asc)) 
    order by order1_.OrderId asc)
order by orderline0_.OrderLineId asc;
```

I think the issue is in `PagingRewriter`, because I can remove this `if` section and then all tests pass:

https://github.com/nhibernate/nhibernate-core/blob/ac39173567b31bcfad475ca32687c9faf0d37f87/src/NHibernate/Linq/GroupBy/PagingRewriter.cs#L56-L61

ngbrown:
Fixes #2479

Adds a test with the general pattern I was trying to accomplish.

The fix involves removing the copying of `OrderBy` in `PagingRewriter`.  All existing tests still passed.


bahusoid:
To me simply removing `if` block looks as very suspicious fix. It's true it's not really required inside your `exists` subquery.

But I'm pretty sure if subquery returns some results - order by is required to be added. I meant something like this:

```C#
var query = db.Orders
			.Where(x => x.Employee.EmployeeId > 5)
			.OrderBy(x => x.OrderId)
			.Take(2);

var listFuture = db.OrderLines
					.Where(
						x =>
							//Check only first with not empty shipper from subquery
							query.Where(x2 => x2.Shipper != null).First() == x.Order)
					.ToFuture();
```

bahusoid:
Maybe fix should skip `order by` processing only if `queryModel.ResultOperators` consists of only bool operators (like Any/All/Contains...)

ngbrown:
@bahusoid So based on the test results, that portion of the code was needed for SQL engines other than MS SQL Server.

I haven't ever work on the LINQ or AST areas of the code before, so I'm pretty out of my depth.  I did lots of watching with the debugger stepping through, and saw that this place was what was duplicating the `OrderBy` but I don't really understand why the PagingRewriter` visitor is needed.

I'm trying to avoid multiple round trips with the database with the nested query.

bahusoid:
> needed for SQL engines other than MS SQL Server.

No it's just coincidence that tests work on SQL Server (apparently if IN is used for id column - SQL Server returns them in that order).  Just change `OrderBy` column to something else and it will fail for SQL Server too. For instance add the following test to `PagingTests`:
```C#
```

bahusoid:
> This includes fixing some tests in CriteriaAssertFixture so they wouldn't report a failure when really passing in ReSharper.

Can you move this change to separate PR?

ngbrown:
I've moved the `CriteriaAssertFixture` fixture change into #2489.

@bahusoid, your example test, `PagedProductsWithOuterWhereClause2` didn't fail as expected without the `order by` cloning on SQL Server.

ngbrown:
> Maybe fix should skip `order by` processing only if `queryModel.ResultOperators` consists of only bool operators (like Any/All/Contains...)

Are you thinking something like this?

```cs
if (!queryModel.BodyClauses.OfType<OrderByClause>().Any() &&
    queryModel.BodyClauses.Any(x => (x as WhereClause)?.Predicate.Type != typeof(bool)))
```

It seems to work, but not sure if it's too broad.

bahusoid:
> didn't fail as expected 

Ok.. Sorry I didn't test it. And I just realized it would still be a query with IN on identifier column - so changing OrderBy doesn't make any difference for SQL Server. But the problem is the following. Now generated query looks like:
```sql
    select
   *
    from
        Products product0_ 
    where
        product0_.UnitsInStock>@p0 
        and (
            product0_.ProductId in (
                SELECT
                    TOP (@p1) ProductId 
                FROM
                    (select
                        product1_.ProductId,
                        ROW_NUMBER() OVER(
                    ORDER BY
                        product1_.UnitPrice,
                        product1_.ProductId) as __hibernate_sort_row 
                    from
                        Products product1_) as query 
                WHERE
                    query.__hibernate_sort_row > @p2 
                ORDER BY
                    query.__hibernate_sort_row)
            );
```

With your fix as you can see `ORDER BY` is only placed inside IN subquery. There is no `ORDER BY` for main query. And most DBs apply default sorting rules for results - so test is failing. And the fact that SQL Server returns those records in order supplied by IN subquery is not documented behavior.

bahusoid:
> Are you thinking something like this?

No I meant something like this:
```C#
if (!queryModel.BodyClauses.OfType<OrderByClause>().Any()
	//There is no need to apply 'order by` if subquery is wrapped in some bool expression (like exists)
	&& !(queryModel.ResultOperators.Count == 1
		&& (queryModel.ResultOperators[0] is AnyResultOperator || queryModel.ResultOperators[0] is AllResultOperator)))
{
	var orderByClauses = subQueryModel.BodyClauses.OfType<OrderByClause>();
	foreach (var orderByClause in orderByClauses)
		queryModel.BodyClauses.Add(orderByClause);
}
```

Though I also don't know this parts of code. So not sure if it's a good fix.

ngbrown:
I've added additional tests for `Contains` and `All`. Each test was red without the fix and are green with it.

I tested locally with both MSSQL and SQLite. That seems to cover the differences between engines.

src/NHibernate.Test/Linq/PagingTests.cs
```diff
@@ -386,6 +386,28 @@ public void PagedProductsWithOuterWhereClause()
 			Assert.That(ids, Is.EqualTo(inMemoryIds));
 		}
 
+		[Test]
+		public void PagedProductsWithOuterWhereClause2()
```
bahusoid, line 390: It adds no value compared to `PagedProductsWithOuterWhereClause`. Better to remove it.

ngbrown:
I've made all the suggested changes and this is ready for further and wider review.

hazzik:
Ideally we should get this query:

```sql
select
    orderline0_.OrderLineId as orderlineid1_4_,
    orderline0_.OrderId as orderid2_4_,
    orderline0_.ProductId as productid3_4_,
    orderline0_.UnitPrice as unitprice4_4_,
    orderline0_.Quantity as quantity5_4_,
    orderline0_.Discount as discount6_4_ 
from
    OrderLines orderline0_ 
where
	orderline0_.OrderId in (
        select
            TOP (2) order2_.OrderId 
        from
            Orders order2_ 
        where
            order2_.EmployeeId>5
        order by
            order2_.OrderId asc
	)
order by
    orderline0_.OrderLineId asc;
```

ngbrown:
@hazzik I'm not sure which query you are referring to.

The closest one is `OrdersWithSubquery9()`. With the fix, it now generates the following:

```sql
DECLARE @p0 AS INT = 5;
DECLARE @p1 AS INT = 2;

select
    orderline0_.OrderLineId as orderlineid1_4_,
    orderline0_.OrderId as orderid2_4_,
    orderline0_.ProductId as productid3_4_,
    orderline0_.UnitPrice as unitprice4_4_,
    orderline0_.Quantity as quantity5_4_,
    orderline0_.Discount as discount6_4_ 
from
    OrderLines orderline0_ 
where
    exists (
        select
            order1_.OrderId 
        from
            Orders order1_ 
        where
            order1_.OrderId=orderline0_.OrderId 
            and (
                order1_.OrderId in (
                    select
                        TOP (@p0) order2_.OrderId 
                    from
                        Orders order2_ 
                    where
                        order2_.EmployeeId>@p1 
                    order by
                        order2_.OrderId asc
                )
            )
        ) 
    order by
        orderline0_.OrderLineId asc;
```

Your query is slightly more efficient in the query planner.

Screenshot - 2020-08-28_15-34-34

I assumed the output is a side effect of the translation from Linq to SQL.

ngbrown:
To improve the generated SQL I think need to do something entirely different in the `PagingRewriter`. I'm not exactly sure what though.

```
PagingRewriter input:
from Order o in {from Order x in value(NHibernate.Linq.NhQueryable`1[NHibernate.DomainModel.Northwind.Entities.Order]) where ([x].Employee.EmployeeId > 5) orderby [x].OrderId asc select [x] => Take(2)} where ([o] == [x].Order) select [o] => Any()

PagingRewriter output:
from Order x in value(NHibernate.Linq.NhQueryable`1[NHibernate.DomainModel.Northwind.Entities.Order]) where ([x] == [x].Order) where {from Order x in value(NHibernate.Linq.NhQueryable`1[NHibernate.DomainModel.Northwind.Entities.Order]) where ([x].Employee.EmployeeId > 5) orderby [x].OrderId asc select [x] => Take(2) => Contains([x])} select [x] => Any()
```

I got it as close as this:

```
from Order x in value(NHibernate.Linq.NhQueryable`1[NHibernate.DomainModel.Northwind.Entities.Order]) where ([o] == [x].Order) where ([x].Employee.EmployeeId > 5) orderby [x].OrderId asc select [o] => Any()
```

But it's still generating `exists` instead of `orderline0_.OrderId in ...`

ngbrown:
In a very narrow case, I got something that can generate the more efficient SQL:

```cs
else if (queryModel.ResultOperators.Count == 1 && queryModel.ResultOperators.All(x => x is AnyResultOperator))
{
	var joinOnBodyClause = ((System.Linq.Expressions.BinaryExpression)((WhereClause) queryModel.BodyClauses[0]).Predicate).Right;
	var cro = new ContainsResultOperator(joinOnBodyClause);

	queryModel.ResultOperators[0] = cro;
	queryModel.ResultTypeOverride = typeof (bool);

	queryModel.BodyClauses.Clear();
	foreach (var orderByClause in subQueryModel.BodyClauses)
	{
		queryModel.BodyClauses.Add(orderByClause);
	}

	queryModel.ResultOperators.Clear();
	foreach (var resultOperator in subQueryModel.ResultOperators)
	{
		queryModel.ResultOperators.Add(resultOperator);
	}

	queryModel.ResultOperators.Add(cro);
	queryModel.ResultTypeOverride = typeof (bool);
}
```

```
from Order x in value(NHibernate.Linq.NhQueryable`1[NHibernate.DomainModel.Northwind.Entities.Order]) where ([x].Employee.EmployeeId > 5) orderby [x].OrderId asc select [x] => Take(2) => Contains([x].Order)
```

Obviously it somehow need to be more generic in getting the correct `Left`/`Right` expression. Maybe there's already visitors that do most of this?

hazzik:
Ok, it is likely that we're trying to fix query in a wrong place. It should be done somewhere else. If you remove `.Take(...)` from `OrdersWithSubquery9` test then it will fail with the same error.

ngbrown:
@hazzik If you manually modify the ideal query text to remove the TOP of an IN sub-query you will get the same error.  It's expected for MS SQL Server to fail queries that have an ordered SQL sub-query without a TOP.

Can you help me with the correctness of the [optimization code](https://github.com/ngbrown/nhibernate-core/commit/1386cf4b4bf7d2f29424c934cf44582f0586afd9) if that's what's needed? Or if the double sub-query, with the slight inefficiency, is fine, are there any other things to check for before merging?

hazzik:
> @hazzik If you manually modify the ideal query text to remove the TOP of an IN sub-query you will get the same error. 

Yes, I know.

> It's expected for MS SQL Server to fail queries that have an ordered SQL sub-query without a TOP.

There is already some code which removes order-by for the `count(*)` result operators. I think it should do the same for other.

hazzik:
> Can you help me with the correctness of the optimization code if that's what's needed?

It's incorrect. What required (and I think it is outside of scope of this PR) is to rewrite `something.Any(y => y == other)` to `something.Contains(other)`.

hazzik (approved):

bahusoid:
> The correct way is to detect if one side of BinaryExpression is QuerySourceReferenceExpression and it is referencing the QueryModel.MainFromClause then other side is an argument to ContainsResultOperator.

Code should also check that's it's simple  single ==/!= call and not something like  `Any(o => o.SomeValue > x.SomeValue)`. I'm not sure it's worth the trouble considering that user can use `Contains` himself.

Issue #2478, opened 2020-08-13
How to correctly make join by IQueryable with NHibernate


RockNHawk:
Hi,

I have searched may document about how to correctly make join by IQueryable with NHibernate, but sill cant make it work.

I have tried join like this:

	var queryableLinq = (from user in _session3.Query<AppUserPartIndex>()
		join cii in _session3.Query<ContentItemIndex>() on user.DocumentId equals cii.DocumentId 
		where userIds.Contains(user.UserId) && cii.Published == true
		select user);
					
	var items = queryableLinq.ToList();
	Console.WriteLine(items);

but throws `Path expected for join` SemanticException.

	NHibernate.Hql.Ast.ANTLR.SemanticException: Path expected for join! [
		.......
	   at NHibernate.Hql.Ast.ANTLR.HqlSqlWalker.CreateFromJoinElement(IASTNode path, IASTNode alias, Int32 joinType, IASTNode fetchNode, IASTNode propertyFetch, IASTNode with)
	   at NHibernate.Hql.Ast.ANTLR.HqlSqlWalker.joinElement()
	   at NHibernate.Hql.Ast.ANTLR.HqlSqlWalker.fromElement()
		.....
	   at NHibernate.Hql.Ast.ANTLR.HqlSqlTranslator.Translate()
	   at NHibernate.Hql.Ast.ANTLR.QueryTranslatorImpl.Analyze(String collectionRole)
		.....

Or add an statement `into` like this:

	var queryableLinq = (from user in _session3.Query<AppUserPartIndex>()
		join cii in _session3.Query<ContentItemIndex>() on user.DocumentId equals cii.DocumentId into alias
		from a in alias
		where userIds.Contains(user.UserId) && a.Published == true
		select user);

but throws `The method or operation is not implemented` NotImplementedException

	System.NotImplementedException: The method or operation is not implemented.
	   at NHibernate.Linq.Visitors.QueryModelVisitor.VisitGroupJoinClause(GroupJoinClause groupJoinClause, QueryModel queryModel, Int32 index)
	   at Remotion.Linq.Clauses.GroupJoinClause.Accept(IQueryModelVisitor visitor, QueryModel queryModel, Int32 index)
	   at Remotion.Linq.QueryModelVisitorBase.VisitBodyClauses(ObservableCollection`1 bodyClauses, QueryModel queryModel)
	   at Remotion.Linq.QueryModelVisitorBase.VisitQueryModel(QueryModel queryModel)
[...]

PR #2485, opened 2020-08-14
Throw entity not mapped exception for entity join in hql if possible

bahusoid:
Consider it as regression as correct exception was thrown from LINQ in 5.2 (with implicit join) and to avoid reporting issues like https://github.com/nhibernate/nhibernate-core/issues/2478

### DIFF
--- a/src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
@@ -794,7 +794,8 @@ namespace NHibernate.Hql.Ast.ANTLR
 			if (path.Type == IDENT)
 			{
 				var pathIdentNode = (IdentNode) path;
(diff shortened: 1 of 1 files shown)


src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
```diff
@@ -794,7 +794,8 @@ private IQueryable ResolveEntityJoinReferencedPersister(IASTNode path)
 			if (path.Type == IDENT)
 			{
 				var pathIdentNode = (IdentNode) path;
-				return SessionFactoryHelper.FindQueryableUsingImports(pathIdentNode.Path);
+				//Until IDENT node is not expected for implicit join path we can throw on not found persister
```
fredericDelaporte, line 797:
```suggestion
				// Since IDENT node is not expected for implicit join path, we can throw on not found persister
```

Issue #2483, opened 2020-08-13
v5.3.2 Invalid cast on nullable custom type with Linq Equals Method


gokhanabatay:
Hi,
We are trying to upgrade NH 5.3.2.
We come across an issue, I implemented below test cases, we are using Equals method instead of == operator as a workaround to remove unnecessary IS NULL conditions from queries.

https://github.com/gokhanabatay/NHibernateBugTest/blob/master/NHibernateBugTest/CustomTypeSelectFailsTest.cs

Get_DateCustomType_NullableDateValueEqualsV2_IncorrectCast
```stack
 Message: 
    NHibernate.Exceptions.GenericADOException : could not execute query
    [ select usersessio0_.GUID as guid1_5_, usersessio0_.MBR_ID as mbr2_5_, usersessio0_.USER_CODE as user3_5_, usersessio0_.OPEN_DATE as open4_5_, usersessio0_.EXPIRE_DATE_TIME as expire5_5_, usersessio0_.IS_OPEN as is6_5_, usersessio0_.REMOTE_IP_ADDRESS as remote7_5_, usersessio0_.REMOTE_PORT as remote8_5_, usersessio0_.LOCAL_IP_ADDRESS as local9_5_, usersessio0_.LOCAL_PORT as local10_5_, usersessio0_.DEVICE_ID as device11_5_, usersessio0_.CLAIMS as claims12_5_ from USER_SESSION usersessio0_ where usersessio0_.MBR_ID = :p0 and usersessio0_.OPEN_DATE=:p1 ]
      Name:p1 - Value:8/14/2020 1:51:13 AM
      ----> Npgsql.PostgresException : 42883: operator does not exist: integer = timestamp without time zone
    Data:
      Severity: ERROR
      InvariantSeverity: ERROR
      SqlState: 42883
      MessageText: operator does not exist: integer = timestamp without time zone
      Hint: No operator matches the given name and argument types. You might need to add explicit type casts.
      Position: 551
      File: parse_oper.c
      Line: 731
      Routine: op_error
[...]

PR #2488, opened 2020-08-14
Fix parameter detection for Equals and CompareTo methods for Linq provider

maca88:
Fix another regression from #2365, where `Equals` and `CompareTo` were forgot to be added in the logic of detecting parameter types.

Related #2483


fredericDelaporte:
This should be retargeted on 5.3.3

PR #2491, opened 2020-08-17
Avoid unnecessary casting for Linq provider

### LINKED ISSUE #2490: Unnecessary cast in sql with Linq are causing performance issues
Hi,
With NH 5.3.2 update Equals method is not working with linq throws "System.NotSupportedException : Boolean Equals(System.Object)" exception.
So we changed to == operator but if field is nullable short its now adding extra cast in sql,  earlier it was not with Equals method as expected.

https://github.com/gokhanabatay/NHibernateBugTest/blob/master/NHibernateBugTest/CustomTypeSelectFailsTest.cs
Test Case:
NullableShortFieldIsNullConditionTest

```sql
select customer0_.GUID as guid1_1_, customer0_.NAME as name2_1_ from CUSTOMER customer0_ where exists (select addresses1_.GUID from ADDRESS addresses1_ where customer0_.GUID=addresses1_.CUSTOMER_GUID and cast(addresses1_.ADDRESS_TYPE as int4)=?)
```

v5.2 working as expected test case: NullableShortFieldIsNullConditionTestV5_2

maca88:
This PR contains an improvement for parameter detection that are related to nullable expressions, which was needed to avoid unnecessary casting reported in #2490.

Fixes #2490


src/NHibernate.Test/Async/Linq/ParameterTests.cs
```diff
@@ -125,6 +125,15 @@ public async Task UsingValueTypeParameterTwiceAsync()
 				1));
 		}
 
+		[Test]
+		public async Task UsingValueTypeParameterTwiceOnNullablePropertyAsync()
+		{
+			short value = 1;
+			await (AssertTotalParametersAsync(
+				db.Users.Where(o => o.NullableShort == value && o.NullableShort != value && o.Short == value),
+				1));
```
maca88, line 134: Prior this PR there would be 3 parameters.

src/NHibernate.Test/Async/Linq/NullComparisonTests.cs
```diff
@@ -560,6 +577,15 @@ public async Task NullInequalityAsync()
 
 			await (ExpectAsync(session.Query<User>().Where(o => o.CreatedBy.ModifiedBy.Id != 5), Does.Contain("is null").IgnoreCase));
 			await (ExpectAsync(session.Query<User>().Where(o => 5 != o.CreatedBy.ModifiedBy.Id), Does.Contain("is null").IgnoreCase));
+
+			short value = 3;
```
gokhanabatay, line 581: unnecessary cast is exists short/long with == / <= / >= operators could you add additional test cases.

fredericDelaporte:
Does this case still qualifies as a regression? It is not really clear when reading #2490.

src/NHibernate.Test/Async/Linq/NullComparisonTests.cs
maca88: When comparing two different types `short == long`, a cast will always be added when the mapped sql type of `short` and `long` are different. I am thinking of restoring the previous behavior of `Equals` in case a user wants to compare two different types without an explicit cast as this is not possible anymore in `5.3`.

maca88:
The problem is that with #1996, `Equals` method was also modified to work the same as `==` operator, where previously it was directly converted without any additional logic. The new logic does not behave the same as it may add a cast where previously by using `Equals` was not added. This PR improves the new logic to not add a cast when both sides are casted by the compiler due to lack of an operator (`short == short`), but still does not match the old `Equals` behavior, for example:
```C#
o.Short.Equals(3)
```
In 5.2.x a cast is not added, where in 5.3 it is added due to a type mismatch (`short` and `int`). I think it would be better to restore the old `Equals` behavior in case the user does not want the new logic to add the is null check and possible a cast.

> Does this case still qualifies as a regression?

Technically it is a regression but this PR does not solves it entirely, that is why I am thinking to restore the old `Equals` behavior, so then this PR may be added in the next version as an improvement or to the next minor as a bug fix.

src/NHibernate.Test/Async/Linq/NullComparisonTests.cs
gokhanabatay: I think the operators and methods should work exactly the same, otherwise the software developer might make an error about which one to use and under what condition. If not careful, sql clauses are created which can cause performance problems.

The strange thing is that it casts unnecessarily in short to short or long to long comparisons.

Equals method with == operator 
CompareTo method result == 0 with <=  operator
CompareTo method result < 0 with <= operator
CompareTo method result > 0 with >= operator

fredericDelaporte:
I think we have a flaw in the casting logic: it should cast the parameter whenever possible, instead of the column. Casting a parameter has far less potential for causing performances issues than casting a column. And a parameter cast could be even pre-evaluated.
Another possibility may be to handle implicit cast detection, where it is known the SQL engine does not need the cast because it does it implicitly between the involved types. But that is very likely to be database dependent and as such quite more heavy to implement.

gokhanabatay:
@fredericDelaporte @maca88 
Take a look below test case with sql statements, as you can see there is no short parameter two short column compare and casting is already exists.

https://github.com/gokhanabatay/NHibernateBugTest/blob/master/NHibernateBugTest/CustomTypeSelectFailsTest.cs
AnyExistsWithNullableShort

```c#
var query = session.Query<Customer>().Where(x => x.Addresses.Any(y => y.AddressType.Value == y.AddressType2.Value));
```

```sql
select customer0_.GUID as guid1_1_, customer0_.NAME as name2_1_ from CUSTOMER customer0_ 
where exists (select addresses1_.GUID from ADDRESS addresses1_ where customer0_.GUID=addresses1_.CUSTOMER_GUID 
and (cast(addresses1_.ADDRESS_TYPE as int4)=cast(addresses1_.ADDRESS_TYPE2 as int4) 
or (cast(addresses1_.ADDRESS_TYPE as int4) is null) and (cast(addresses1_.ADDRESS_TYPE2 as int4) is null)))
```

maca88:
> it should cast the parameter whenever possible, instead of the column

I do agree, I will try to implement the following rules:
- When comparing integral numeric types, pre-evaluate the parameter `o.Integer == 5L || o.Short < 4 || o.Long >= 5UL`
- When comparing integral numeric parameters with floating-point numeric columns, pre-evaluate the parameter `o.Decimal == 4 || o.Double < 5L`
- When comparing floating-point numeric types, pre-evaluate the parameter `o.Double == 2.3m || o.Float < 2.1d || o.Decimal > 3.3f`
- When a floating-point numeric type parameter is used for integral and floating-point numeric columns, cast columns `o.Double < decimalParam || o.Integer >= decimalParam`
- When a floating-point numeric type parameter is used for different floating-point numeric columns, cast the parameter `o.Double < decimalParam || o.Float > decimalParam`
- When an integral numeric type parameter is used for different numeric columns, cast the parameter `o.Double < intParam || o.Float > intParam || o.Short > intParam`

> Another possibility may be to handle implicit cast detection

I think that the above rules should be good enough.

> Two short column compared with casting

This is already fixed with the current state of this PR, I will add additional tests for that.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -81,7 +81,15 @@ public override Expression Visit(Expression expression)
 				#region NH additions
 				// Variables should be evaluated only when they are part of an evaluatable expression (e.g. o => string.Format("...", variable))
 				expression is UnaryExpression unaryExpression &&
-				ExpressionsHelper.IsVariable(unaryExpression.Operand, out _, out _))
+				(
```
hazzik, line 84: Can the expression inside parenthesis please be moved to a method/methods with a meaningful name?

Remember, if something requires a comment it means it should go to a method.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -496,11 +526,14 @@ protected HqlTreeNode VisitUnaryExpression(UnaryExpression expression)
 				case ExpressionType.Convert:
 				case ExpressionType.ConvertChecked:
 				case ExpressionType.TypeAs:
-					return IsCastRequired(expression.Operand, expression.Type, out var existType)
+					var notCastable = _notCastableExpressions.TryGetValue(expression, out var castType);
+					castType = castType ?? expression.Type;
```
hazzik, line 530: `castType` is null when not `notCastable`. Can this please made explicit?

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
maca88: Done.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
maca88: Done.

src/NHibernate.Test/Linq/NullComparisonTests.cs
```diff
@@ -460,6 +461,33 @@ public void NullEquality()
...
+			Expect(db.NumericEntities.Where(o => o.Short == o.NullableShort), Does.Not.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+
+			short value = 3;
+			Expect(db.NumericEntities.Where(o => o.NullableShort == value), Does.Not.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+			Expect(db.NumericEntities.Where(o => value == o.NullableShort), Does.Not.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+
+			Expect(db.NumericEntities.Where(o => o.NullableShort.Value == value), Does.Not.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+			Expect(db.NumericEntities.Where(o => value == o.NullableShort.Value), Does.Not.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+			Expect(db.NumericEntities.Where(o => o.Short == value), Does.Not.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+			Expect(db.NumericEntities.Where(o => value == o.Short), Does.Not.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
```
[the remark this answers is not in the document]
maca88: Done.

src/NHibernate.Test/Linq/NullComparisonTests.cs
```diff
@@ -548,6 +576,43 @@ public void NullInequality()
...
+			Expect(db.NumericEntities.Where(o => o.Short != o.NullableShort), Does.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+
+			short value = 3;
+			Expect(db.NumericEntities.Where(o => o.NullableShort != value), Does.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+			Expect(db.NumericEntities.Where(o => value != o.NullableShort), Does.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+
+			Expect(db.NumericEntities.Where(o => o.NullableShort.Value != value), Does.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+			Expect(db.NumericEntities.Where(o => value != o.NullableShort.Value), Does.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+			Expect(db.NumericEntities.Where(o => o.Short != value), Does.Not.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
+			Expect(db.NumericEntities.Where(o => value != o.Short), Does.Not.Contain("is null").IgnoreCase.And.Not.Contain("cast"));
```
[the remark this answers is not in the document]
maca88: Done.

src/NHibernate/Linq/Visitors/ParameterTypeLocator.cs
```diff
@@ -80,55 +101,68 @@ public static class ParameterTypeLocator
 			var visitor = new ConstantTypeLocatorVisitor(removeMappedAsCalls, targetType, parameters, sessionFactory);
 			queryModel.TransformExpressions(visitor.Visit);
 
-			foreach (var pair in visitor.ConstantExpressions)
+			var processedConstants = new HashSet<ConstantExpression>();
```
hazzik, line 104: `processedConstants` is not used

src/NHibernate/Linq/Visitors/ParameterTypeLocator.cs
```diff
@@ -80,55 +101,68 @@ public static class ParameterTypeLocator
...
+				// do not change the parameter type, but instead cast the parameter when comparing with different column types.
 				if (type == null)
 				{
 					type = constantExpression.Value != null
 						? ParameterHelper.TryGuessType(constantExpression.Value, sessionFactory, namedParameter.IsCollection)
-						: ParameterHelper.TryGuessType(constantExpression.Type, sessionFactory, namedParameter.IsCollection);
+						: ParameterHelper.TryGuessType(constantType, sessionFactory, namedParameter.IsCollection);
 				}
 
 				namedParameter.Type = type;
```
hazzik, line 168: Please move `type` calculation into a method

src/NHibernate/Linq/Visitors/ParameterTypeLocator.cs
```diff
@@ -80,55 +101,68 @@ public static class ParameterTypeLocator
...
-					// MappedAs was used
-					namedParameter.Type = type;
-					continue;
+					if (visitor.RelatedExpressions.TryGetValue(expression, out var relatedExpressions))
+					{
+						parameterRelatedExpressions.AddRange(relatedExpressions);
+					}
 				}
 
+				var candidateTypes = new HashSet<IType>();
```
hazzik, line 125: Please extract `candidateTypes` calculation into a method.

src/NHibernate/Linq/Visitors/ParameterTypeLocator.cs
```diff
@@ -80,55 +101,68 @@ public static class ParameterTypeLocator
...
+				var constantType = constantExpression.Type.UnwrapIfNullable();
+				IType type = null;
+				if (
+					candidateTypes.Count == 1 &&
+					// When comparing an integral column with a floating-point parameter, the parameter type must remain floating-point type
+					// and the column needs to be casted in order to prevent invalid results (e.g. Where(o => o.Integer >= 2.2d)).
+					!(candidateTypes.Any(t => IntegralNumericTypes.Contains(t.ReturnedClass)) && FloatingPointNumericTypes.Contains(constantType))
+				)
+				{
+					type = candidateTypes.FirstOrDefault();
```
hazzik, line 155: `candidateTypes` is of size 1 here. You can use `First` method.

src/NHibernate/Linq/Visitors/ParameterTypeLocator.cs
bahusoid: As far as I can see method can return either single `candidateType` or null in case multiple results/not found case. Also the whole routine can be skipped if `!FloatingPointNumericTypes.Contains(constantType)`

hazzik:
I did some refactoring for readability.

maca88:
Thanks!

bahusoid:
I find it a bit confusing when 2 methods with same name have some subtle difference in functionality - so I merged them. I hope readability is not affected )

hazzik (approved):

bahusoid (approved):

src/NHibernate/Type/SingleType.cs
```diff
@@ -62,7 +62,7 @@ public override System.Type ReturnedClass
 
 		public override void Set(DbCommand rs, object value, int index, ISessionImplementor session)
 		{
-			rs.Parameters[index].Value = value;
+			rs.Parameters[index].Value = Convert.ToSingle(value);
```
bahusoid, line 65: Maybe it should be  `Convert.ToSingle(value, CultureInfo.InvariantCulture)`?

maca88: Normally the argument is a numeric type, so I am not sure whether `CultureInfo` is even used in that case. If we add the invariant culture for `Single` then we should do it also for other types to be consistent, which I think should be done in a separate PR.

bahusoid: Ok

hazzik:
@maca88 does it solve #2490?

maca88:
Yes it does, updated the description of this PR.

PR #2492, opened 2020-08-20
Replace casting with NodeType checks in Criteria ExpressionProcessor

bahusoid:
Single check (`is/as` cast vs `NodeType` with direct cast)  is around 11% faster when casting is failed and 3% percent slower for successful casting case.

### DIFF
--- a/src/NHibernate/Impl/ExpressionProcessor.cs
+++ b/src/NHibernate/Impl/ExpressionProcessor.cs
@@ -318,45 +318,27 @@ namespace NHibernate.Impl
 
 		private static Expression UnwrapConvertExpression(Expression expression)
 		{
-			if (expression is UnaryExpression unaryExpression)
-			{
-				if (!IsConversion(unaryExpression.NodeType))
-				{
-					if (IsSupportedUnaryExpression(unaryExpression))
-						return null;
-
-					throw new ArgumentException("Cannot interpret member from " + expression, nameof(expression));
-				}
-				return unaryExpression.Operand;
-			}
-
-			return null;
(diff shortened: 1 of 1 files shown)


hazzik:
Why?

bahusoid:
> Why?

@hazzik  Because it's a little bit faster and don't make much difference in readability. So why not.. I actually expected a little more gain (actually maybe in Full Framework there is more gain). So why not...

src/NHibernate/Impl/ExpressionProcessor.cs
```diff
@@ -494,26 +472,18 @@ private static System.Type FindMemberType(Expression expression)
...
-			if (methodCallExpression != null)
-			{
-				return methodCallExpression.Method.ReturnType;
-			}
-
-			if (expression is BinaryExpression || expression is UnaryExpression)
-				return expression.Type;
-
-			throw new ArgumentException("Could not determine member type from " + expression, nameof(expression));
+			return expression.Type;
```
fredericDelaporte, line 475: The previous code would have throw if the expression is not a binary, unary or member expression. Why do we not need this check? Is it because it would be already somehow be checked by the `FindMemberProjection` call that is currently done prior to calling `FindMemberType`? It does not look to me as being the case.

bahusoid: I just thought it's not that important to have strict check inside this private method as long `FindMemberProjection` is able to parse expression.. But maybe not - I restored exception

fredericDelaporte (approved):

PR #2494, opened 2020-08-21
Fix missing joins in subquery in case joins are present in main query

bahusoid:
Fixes https://github.com/nhibernate/nhibernate-core/issues/1326

Fix is ported from hibernate: https://github.com/hibernate/hibernate-orm/blob/5ecf45020ff5f7d269425abf7d25823ca4776f62/hibernate-core/src/main/java/org/hibernate/hql/internal/ast/tree/DotNode.java#L555-L564


hazzik (approved):

fredericDelaporte:
Shouldn't it be `next_minor`?

hazzik:
Yes, true

PR #2499, opened 2020-08-26
Cast operation fails when an enum is mapped as an AnsiString

gliljas:
Regression from #2036 

Worked fine for enums mapped as String, but not if an AnsiString or FixedLength type was used.

### DIFF
--- a/src/NHibernate/Util/DbTypeExtensions.cs
+++ b/src/NHibernate/Util/DbTypeExtensions.cs
@@ -1,0 +1,14 @@
+﻿using System.Data;
+
+namespace NHibernate.Util
+{
+	public static class DbTypeExtensions
+	{
+		/// <summary>
+		/// Checks whether the type is a <see cref="DbType.String"/>, <see cref="DbType.AnsiString"/>, <see cref="DbType.StringFixedLength"/> or <see cref="DbType.AnsiStringFixedLength"/>
+		/// </summary>
+		/// <param name="dbType"></param>
+		/// <returns></returns>
+		public static bool IsStringType(this DbType dbType) => dbType == DbType.String || dbType == DbType.AnsiString || dbType == DbType.StringFixedLength || dbType == DbType.AnsiStringFixedLength;
+	}
+}
--- a/src/NHibernate/Util/DbTypeExtensions.cs
+++ b/src/NHibernate/Util/DbTypeExtensions.cs
@@ -2,7 +2,7 @@
 
 namespace NHibernate.Util
 {
-	public static class DbTypeExtensions
+	internal static class DbTypeExtensions
 	{
 		/// <summary>
 		/// Checks whether the type is a <see cref="DbType.String"/>, <see cref="DbType.AnsiString"/>, <see cref="DbType.StringFixedLength"/> or <see cref="DbType.AnsiStringFixedLength"/>
--- a/src/NHibernate.Test/Async/Linq/EnumTests.cs
+++ b/src/NHibernate.Test/Async/Linq/EnumTests.cs
@@ -21,13 +21,18 @@ namespace NHibernate.Test.Linq
 {
 	using System.Threading.Tasks;
 	using System.Threading;
+	using System.Linq.Expressions;
+
 	[TestFixture(typeof(EnumType<TestEnum>))]
 	[TestFixture(typeof(EnumStringType<TestEnum>))]
 	[TestFixture(typeof(EnumAnsiStringType<TestEnum>))]
 	public class EnumTestsAsync : TestCaseMappingByCode
 	{
 		private IType _enumType;
+		private ISession _session;
 
+		private IQueryable<EnumEntity> TestEntities { get; set; }
+		private IQueryable<EnumEntity> TestEntitiesInDb { get; set; }
 
 		public EnumTestsAsync(System.Type enumType)
 		{
@@ -43,6 +48,7 @@ namespace NHibernate.Test.Linq
 					rc.Table("EnumEntity");
 					rc.Id(x => x.Id, m => m.Generator(Generators.Identity));
 					rc.Property(x => x.Name);
+					rc.Property(x => x.BatchId);
 					rc.Property(x => x.Enum, m => m.Type(_enumType));
 					rc.Property(x => x.NullableEnum, m => m.Type(_enumType));
 					rc.ManyToOne(x => x.Other, m => m.Cascade(Mapping.ByCode.Cascade.All));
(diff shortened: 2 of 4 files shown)


bahusoid:
Should be targeted to 5.3.x branch and async regen is missing

src/NHibernate/Util/DbTypeExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Data;
+
+namespace NHibernate.Util
+{
+	public static class DbTypeExtensions
```
maca88, line 5: As this is a regression fix, I think that we should not expose new type/methods.

gliljas: A fair point. Would you agree with just making internal?

hazzik: I made it internal.

hazzik (approved):

src/NHibernate/Util/DbTypeExtensions.cs
gliljas: Thanks!

hazzik:
@bahusoid addressed.

bahusoid (changes requested):
Failed tests on some dialects...

gliljas:
Checking.

gliljas:
I will make the "complex" test a bit more legible and predictable.

gliljas:
I was unable to run the async generator. Never happened before. So I did the asyncification manually...

`  [exec] Generating async code started
     [exec] Opening project 'c:\Projects\github\nhibernate-core\src\NHibernate\NHibernate.csproj' started
     [exec] System.InvalidOperationException: One or more errors occurred while opening the project:
     [exec] Msbuild failed when processing the file 'c:\Projects\github\nhibernate-core\src\NHibernate\NHibernate.csproj' with message: C:\Program Files\dotnet\sdk\3.1.401\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets: (241, 5): The "ResolvePackageAssets" task failed unexpectedly.
     [exec] System.IO.FileNotFoundException: Could not load file or assembly 'NuGet.ProjectModel, Version=5.7.0.7, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
     [exec] File name: 'NuGet.ProjectModel, Version=5.7.0.7, Culture=neutral, PublicKeyToken=31bf3856ad364e35' ---> System.IO.FileNotFoundException: Could not load the specified file.`

gliljas:
I believe that the failing query may actually have failed before too, but it surfaced now. Will try to investigate with Firebird or SQLite.

fredericDelaporte:
For the async generator, there is an issue with it depending on which Core SDK you have on your setup. There is a workaround [here](https://github.com/maca88/AsyncGenerator/issues/150#issuecomment-663883863).

bahusoid:
> I believe that the failing query may actually have failed before too, but it surfaced now.

I believe you just changed logic a bit when did conversion to use `EnumEntity`. Few things I noticed at quick glance:

1) In original tests special type is used `EnumStoredAsStringType` with some additional logic (such as `Unspecified` enum value is set as NULL  parameter)
2) User.NullableEnum property is mapped as formula

P.S. I don't think changes you did in last commit are required and I would prefer to keep your initial version - with minimal modifications to original tests.

gliljas:
> > I believe that the failing query may actually have failed before too, but it surfaced now.
> 
> I believe you just changed logic a bit when did conversion to use `EnumEntity`. Few things I noticed at quick glance:
> 
> 1. In original tests special type is used `EnumStoredAsStringType` with some additional logic (such as `Unspecified` enum value is set as NULL  parameter)
> 2. `User.NullableEnum1` property is mapped as formula
> 
Good catch!  I didn't see that hidden complexity.

> P.S. I don't think changes you did in last commit are required and I would prefer to keep your initial version - with minimal modifications to original tests.

I mostly agree, although I think a test that checks for count=0 easily hides problems. I'll revert a bit, and probably replicate the EnumStoredAsStringType "features".

bahusoid:
I just checked generated SQL for SQLite and I think this test indeed reveals new issue (at least for SQLite dialect). I see unnecessary `cast` in generated SQL. Though I'm pretty sure this issue is unrelated to fix made in this PR:

```sql
 select
        enumtests_0_.Id as col_0_0_,
        enumtests_0_.Enum as col_1_0_,
        case 
            when enumtests_0_.Enum=@p0 then 1 
            else 0 
        end as col_2_0_,
        enumtests_0_.Enum as col_3_0_,
        (case 
            when enumtests_0_.Enum = 'Unspecified' then null 
            else enumtests_0_.Enum 
        end) as col_4_0_,
        enumtests_0_.Id as id1_0_,
        enumtests_0_.Name as name2_0_,
        enumtests_0_.Enum as enum3_0_,
        enumtests_0_.Other as other4_0_,
        (case 
            when enumtests_0_.Enum = 'Unspecified' then null 
            else enumtests_0_.Enum 
        end) as formula2_ 
    from
        EnumEntity enumtests_0_ 
    where
        cast(case 
            when (case 
                when enumtests_0_.Enum = 'Unspecified' then null 
                else enumtests_0_.Enum 
            end)=@p1 then @p2 
            else coalesce((case 
                when enumtests_0_.Enum = 'Unspecified' then null 
                else enumtests_0_.Enum 
            end), enumtests_0_.Enum) 
        end as INTEGER)=@p3; -- THIS CAST IS INCORRECT
```

@maca88  Can you take a look?

bahusoid:
I'm puzzled by firebird errors:
```
 Error : NHibernate.Test.Linq.FunctionTests.Trim
NHibernate.Exceptions.GenericADOException : could not get next sequence value[SQL: select gen_id(hibernate_sequence, 1 ) from RDB$DATABASE]
  ----> FirebirdSql.Data.FirebirdClient.FbException : invalid request BLR at offset 61
generator HIBERNATE_SEQUENCE is not defined
Data:
  actual-sql-query: select gen_id(hibernate_sequence, 1 ) from RDB$DATABASE
```

Any idea what's going on?

bahusoid (approved):

maca88:
> Can you take a look?

I did check it and it is indeed not related to this PR but instead to #707, where `transparentcast` is added for the `ConditionalExpression`, which is converted to a cast in SQLite. Prior #707 it would fail for all databases, so it is now minimized to SQLite. Should we fix the issue in this PR or in a separate one?

bahusoid:
> Should we fix the issue in this PR or in a separate one?

I think better to do it in separate PR on top of changes made here.

hazzik (approved):

hazzik:
Quick question. What about `text`, `clob`, etc?

maca88:
I did a quick test by setting the column sql type (`m.Column(cm => cm.SqlType("text"))`) and it does not work for neither of them.

With `clob` it fails when inserting rows in the database with error:
`System.Data.OracleClient.OracleException : ORA-01722: invalid number`

With `text` the query fails with error:
`System.Data.SqlClient.SqlException : The data types text and varchar are incompatible in the equal to operator.`

None of the above errors are related to this PR as `DbType` is still the same, so the casting logic is not affected.

hazzik:
Actually these errors might be because `.IsStringType()` does not account of these `DbType`s.

hazzik:
Ok, checked myself. For `StringClobSqlType` the `DbType` is `String`. So this works as expected.

PR #2501, opened 2020-08-29
Upgrade MySql client and remove allowed failures on CI builds

bahusoid:
They claim issue is no longer reproduced on [latest client](https://www.nuget.org/packages/MySql.Data/): https://bugs.mysql.com/bug.php?id=90086


hazzik:
It is fixed in .22 which is not released yet. I’m waiting for the release.

bahusoid:
Yeah.. Too soon

hazzik:
@bahusoid still does not work


This is what I was trying to fix:

```
Failed : NHibernate.Test.TypesTest.StringTypeFixtureAsync.InsertUnicodeValueAsync

  String lengths are both 53. Strings differ at index 0.

  Expected: "길동 최고 新闻 地图 ます プル éèêëîïôöõàâäåãçùûü бджзй αβ ខគឃ ضذخ"

  But was:  "?? ?? ?? ?? ?? ?? éèêëîïôöõàâäåãçùûü ????? aß ??? ???"

  -----------^

   at NHibernate.Test.TypesTest.StringTypeFixtureAsync.InsertUnicodeValueAsync() in /home/travis/build/nhibernate/nhibernate-core/src/NHibernate.Test/Async/TypesTest/StringTypeFixture.cs:line 68
```

bahusoid:
> This is what I was trying to fix

I got it. But it failed the same with you changes, so I assumed that such characters are 4-bytes long so changed to UTF8mb4. This issue is gone with this change . But other tests failed with new error like
```
NHibernate.Test.SubselectFetchTest.SubselectFetchFixture
NHibernate.HibernateException : Specified key was too long; max key length is 767 bytes
  ----> MySql.Data.MySqlClient.MySqlException : Specified key was too long; max key length is 767 bytes
```

hazzik:
```
+----------------------+-------------------+
| Variable_name        | Value             |
+----------------------+-------------------+
| collation_connection | utf8_general_ci   |
| collation_database   | latin1_swedish_ci |
| collation_server     | latin1_swedish_ci |
+----------------------+-------------------+
```

🤦‍♂️

hazzik (approved):

hazzik:
Ok. Combination of `utf8_general_ci` and mysql 5.7 has fixed the MySQL issues. The only problem that MySQL 5.7 is not supported on trusty. Upgrade to bionic/xenial, unfortunately, breaks Firebird.

bahusoid:
> Combination of utf8_general_ci and mysql 5.7

I was  close :) Did you get why no explicit changes are required for appveyour?

> Upgrade to bionic/xenial, unfortunately, breaks Firebird.

Hm.. I made a quick search and found seems to [working ](https://travis-ci.com/github/asfernandes/node-firebird-drivers/jobs/320011060) Firebird installation on xenial travis:
https://github.com/asfernandes/node-firebird-drivers/blob/7fddf29970066b185a580e3f2dd26cbc0b6e6848/.travis.yml

Seems the relevant parts:
https://github.com/asfernandes/node-firebird-drivers/blob/7fddf29970066b185a580e3f2dd26cbc0b6e6848/.travis.yml#L18-L20
https://github.com/asfernandes/node-firebird-drivers/blob/7fddf29970066b185a580e3f2dd26cbc0b6e6848/.travis.yml#L35-L38

hazzik:
> Did you get why no explicit changes are required for appveyor?

Because AppVeyor already runs 5.7 with `utf8_general_ci` collation.

hazzik:
> Seems the relevant parts

First link is what we do but with curl instead of nuget.

The second one is what we also have.

I think usermod command is the key.

hazzik:
> I think usermod command is the key.

It should be. I've used clean docker image of `ubuntu:xenial` and Firebird works there using our steps. The difference is that in docker everything is run under `root` user. Here as `travis`.

.travis.yml
```diff
@@ -41,7 +37,7 @@ before_install:
       echo -e "nhibernate = /tmp/firebird/nhibernate.fdb" | sudo tee -a /opt/firebird/databases.conf
       echo -e "AuthServer = Srp\nAuthClient = Srp\nUserManager = Srp\nWireCrypt = Enabled" | sudo tee -a /opt/firebird/firebird.conf
       sudo /opt/firebird/bin/gsec -modify SYSDBA -pw masterkey -admin yes
-      sudo service firebird restart
+      sudo systemctl restart firebird-superserver
```
hazzik, line 40: The reason it was failing is because the service name on 3.0.x has changed. Probably 14.04 has some residue service definitions for `firebird` service and so it was working as intended.

.travis.yml
```diff
@@ -13,25 +12,22 @@ env:
...
     fi
   - |-
     if [[ "$DB" == "Firebird" ]]
     then
-      sudo apt-get install -y libicu-dev libtommath-dev curl
-      curl -L -O https://github.com/FirebirdSQL/firebird/releases/download/R3_0_5/Firebird-3.0.5.33220-0.amd64.tar.gz
+      sudo apt-get install -y libtommath-dev
+      # This would be required on bionic and above
+      # sudo ln -s /usr/lib/x86_64-linux-gnu/libtommath.so.1 /usr/lib/x86_64-linux-gnu/libtommath.so.0
+      wget -q https://github.com/FirebirdSQL/firebird/releases/download/R3_0_5/Firebird-3.0.5.33220-0.amd64.tar.gz
```
hazzik, line 30: `curl -L -O URL` and `wget URL` do exactly same. I decided to leave `wget` as it is a standard GNU utility.

.travis.yml
```diff
@@ -13,25 +12,22 @@ env:
...
     if [[ "$DB" == "MySQL" ]]
     then
-      echo -e '[server]\nlower_case_table_names=1' | sudo tee -a /etc/mysql/my.cnf
+      echo -e '[server]\nlower_case_table_names=1\n[mysqld]\ncharacter-set-server=utf8\ncollation-server=utf8_general_ci' | sudo tee -a /etc/mysql/my.cnf
       sudo service mysql restart
     fi
   - |-
     if [[ "$DB" == "Firebird" ]]
     then
-      sudo apt-get install -y libicu-dev libtommath-dev curl
```
hazzik, line 33: `libicu-dev` and `curl` are already installed so no point to install them again.

.travis.yml
```diff
@@ -13,25 +12,22 @@ env:
   - DB=Firebird 
   - DB=MySQL CONNECTION_STRING="Server=127.0.0.1;Uid=root;Database=nhibernate;Old Guids=True;"   
   - DB=SQLite 
-matrix:
-  allow_failures:
-    - env: DB=MySQL CONNECTION_STRING="Server=127.0.0.1;Uid=root;Database=nhibernate;Old Guids=True;"
 before_install:
-  - curl https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
-  - curl https://packages.microsoft.com/config/ubuntu/14.04/prod.list | sudo tee /etc/apt/sources.list.d/microsoft.list
```
hazzik, line 21: Not needed.

hazzik (approved):

.travis.yml
```diff
@@ -1,8 +1,7 @@
-dist: trusty
+dist: xenial
 language: csharp
 mono: none
 dotnet: 2.1.300
-sudo: required
```
hazzik, line 5: This option is obsolete

.travis.yml
hazzik, line 21: Not needed

.travis.yml
hazzik, line 33: `libicu-dev` and `curl` are already installed so no point to install them again.

.travis.yml
hazzik, line 30: `curl -L -O URL` and `wget URL` do exactly same. I decided to leave `wget` as it is a standard GNU utility.

.travis.yml
hazzik, line 40: The reason it was failing is because the service name on 3.0.x has changed. Probably 14.04 has some residue service definitions for `firebird` service and so it was working as intended.

hazzik (approved):

hazzik (approved):

hazzik:
Я джва года ждал этого релиза.

bahusoid:
> Я два года ждал этого релиза.

:) непреодолимое желание подружить nhibernate с mysql на линухе?

.travis.yml
bahusoid: Why did you change it back to `libtommath-dev`? I suspect with `libtommath0`  hacky stuff below with `ln` wouldn't be required for bionic and above. Or am I wrong?

hazzik: The `libtommath-dev` references either `libtommath0` or `libtommath1` depending on the distributive. 16.04 is the last version which will include `libtommath0`. The 18.04 and above uses `libtommath1`.

Firebird 3.x is built against `libtommath0` and they refuse to rebuild it to `libtommath1`, so the `ln` will be required in 18.04 or above when using the official distribution.

However, it seems that 18.04 and above have Firebird 3.0 in the `apt`.

PR #2504, opened 2020-08-30
Throw for LINQ queries on unmapped entities

### LINKED ISSUE #1106: NH-2978 - LINQ: Queries for unmapped entity types return empty result set
**James Nail** created an [issue](https://nhibernate.jira.com/browse/NH-2978) — <time datetime="2011-12-16T22:52:56.812Z" title="2011-12-16T22:52:56.812Z">16th December 2011, 22:52:56</time>:
> in NH 3.2.0, if you do a LINQ query for an unmapped entity type it returns an empty result set instead of throwing an exception.
```C#
mySession.Query<MyUnmappedEntityType>().ToList();
mySession.Query<MyUnmappedEntityType>().Delete();
```

Since the HQL query implementation throws an exception in this scenario (ex, NHibernate.Hql.Ast.ANTLR.QuerySyntaxException --> "MyUnmappedEntityType is not mapped"), I would expect the LINQ provider to have the same behavior (it should throw an exception)
> 
---
**Anton Chelnokov** added a comment — <time datetime="2015-09-02T17:43:34.663Z" title="2015-09-02T17:43:34.663Z">2nd September 2015, 17:43:34</time>:
> 4 years...
>

bahusoid:
Fixes #1106


src/NHibernate/Hql/Ast/ANTLR/AstPolymorphicProcessor.cs
```diff
@@ -29,8 +29,11 @@ private IASTNode[] Process()
             // Find all the polymorphic query sources
             _nodeMapping = new PolymorphicQuerySourceDetector(_factory).Process(_ast);
 
-            if (_nodeMapping.Count() > 0)
+            if (_nodeMapping.Count > 0)
             {
+                foreach (var kv in _nodeMapping.Where(x => x.Value.Length == 0))
```
hazzik, line 34: Shall we throw one exception for all?

bahusoid: I still undecided it's a good place for throwing. Also check seems incorrect for polymorphic case as some failed tests show. Seems should be `_nodeMapping.All(x => x.Value.Length == 0)`

bahusoid: > Also check seems incorrect for polymorphic case as some failed tests show.

`Can_LazyPropertyNotCastingException` failure reveals that empty database check doesn't work for entities mapped with `polymorphism="explicit"`

bahusoid:
I simply skipped invalid tests in `FunctionTests` as they didn't really test anything (as those test are simply based on the fact that query can be executed  and dump results to output).

P.S. We need to do something with tests inside Linq folder.
As now all of them are executed under `LinqReadonlyTestsContext` . So it's problematic to add some non readonly test or test not based on Northwind database. And mixing non-readonly and readonly tests can lead to some vague issues (like[ missing sequence](https://github.com/nhibernate/nhibernate-core/pull/2499#issuecomment-683278785) on firebird for seems unrelated tests).

hazzik:
As I understand it the idea of LinqReadOnlyTextContext was exactly about that - running read only tests. So, IMO, non read only tests should go somewhere else.

bahusoid:
> So, IMO, non read only tests should go somewhere else.

Scope of this LinqReadOnlyTextContext  is the whole LINQ folder with subfolders (namespace based). Seems excessive to me.

bahusoid:
Maybe rename it to LinqReadonly.. Or put in Readonly subfolder. It's just a bit contr-intuitive that you have to put LINQ related tests in some other place and not inside Linq folder/subfolder

hazzik (dismissed):

hazzik (approved):

PR #2506, opened 2020-08-31
Exception on unmapped class DML query

### LINKED ISSUE #2167: Query<UnmappedType>
### Linq on unmapped type is NOP.

Consider code below it will executes successfully and do nothing.
It is clear that it is mistake on  programmer`s part.
Such operation should throw exception same as HQL.
```
    class Program
    {
        public class UnmappedClass
        {
        }
        static void Main(string[] args)
        {
            var configuration = new NHibernate.Cfg.Configuration();
            configuration.AddProperties(new Dictionary<string, string> {
                {NHibernate.Cfg.Environment.Dialect, typeof(MySQL5Dialect).FullName},
                {NHibernate.Cfg.Environment.ConnectionDriver, typeof(MySqlDataDriver).FullName},
                {NHibernate.Cfg.Environment.ConnectionString, "host=localhost; user=root"},
            });
            ISession session = configuration.BuildSessionFactory().OpenSession();
            session.Query<UnmappedClass>().Delete();
        }
    }
```

radiy:
Trying to fix #2167.
Do not know if it is right approach, but it seems to work.

### DIFF
--- a/src/NHibernate.Test/LinqBulkManipulation/Fixture.cs
+++ b/src/NHibernate.Test/LinqBulkManipulation/Fixture.cs
@@ -181,6 +181,18 @@ namespace NHibernate.Test.LinqBulkManipulation
 
 		#region INSERTS
 
+		public class NotMappedEntity
+		{
+		}
+
+		[Test]
+		public void NotMapped()
+		{
+			using(var s = OpenSession()) {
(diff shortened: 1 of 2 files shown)


radiy:
Whoops, it seems I lost this race. Closing.

PR #2510, opened 2020-09-02
Remove OrderByClause from query models with Contains, All and Any result operators

hazzik:
Improvement for an unwanted behavior found on #2480

### DIFF
--- a/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
+++ b/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
@@ -621,9 +621,6 @@ where c.Order.Customer.CustomerId = 'VINET'
 		[Test]
 		public void OrdersWithSubquery11A()
 		{
-			//if (Dialect is MsSqlCeDialect)
-			//	Assert.Ignore("MS SQL CE does not support sorting on a subquery.");
-
 			var ordersQuery = db.Orders
 			                    .OrderByDescending(x => x.OrderLines.Count).ThenBy(x => x.OrderId);
 
--- a/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
+++ b/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
@@ -534,6 +534,25 @@ where c.Order.Customer.CustomerId = 'VINET'
 			Assert.That(orderLines.Count, Is.EqualTo(6), nameof(orderLines));
 		}
 
+		[Test]
+		public void OrdersWithSubquery9A()
+		{
+			var ordersQuery = db.Orders
+			                    .Where(x => x.Employee.EmployeeId > 5)
+			                    .OrderBy(x => x.OrderId);
+
+			var orderLinesFuture = db.OrderLines
+			                         .Where(x => ordersQuery.Any(o => o == x.Order))
+			                         .OrderBy(x => x.Id)
+			                         .ToFuture();
+
+			var orders = ordersQuery.ToFuture().ToList();
+			var orderLines = orderLinesFuture.ToList();
+
+			Assert.That(orders.Count, Is.EqualTo(286), nameof(orders));
+			Assert.That(orderLines.Count, Is.EqualTo(711), nameof(orderLines));
+		}
+
 		[Test(Description = "GH2479")]
 		public void OrdersWithSubquery10()
 		{
@@ -558,6 +577,26 @@ where c.Order.Customer.CustomerId = 'VINET'
 			Assert.That(products.Count, Is.EqualTo(6), nameof(products));
 		}
 
+		[Test]
+		public void OrdersWithSubquery10A()
+		{
+			var ordersQuery = db.Orders
+			                    .Where(x => x.Employee.EmployeeId > 5)
+			                    .OrderBy(x => x.OrderId);
+
+			var productsQuery = ordersQuery.SelectMany(x => x.OrderLines).Select(x => x.Product);
+			var productsFuture = db.Products
+			                       .Where(x => productsQuery.Contains(x))
+			                       .OrderBy(x => x.ProductId)
+			                       .ToFuture();
+
+			var orders = ordersQuery.ToFuture().ToList();
+			var products = productsFuture.ToList();
+
+			Assert.That(orders.Count, Is.EqualTo(286), nameof(orders));
+			Assert.That(orders.Count, Is.EqualTo(286), nameof(orders));
+		}
+
 		[Test(Description = "GH2479")]
 		public void OrdersWithSubquery11()
 		{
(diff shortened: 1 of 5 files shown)


bahusoid:
You can't simply remove `OrderBy` as it does matter for sub-queries with TOP:

```C#
var ordersQuery = db.Orders
                    .Where(x => x.Employee.EmployeeId > 5)
                    .OrderBy(x => x.OrderId)
		    .Take(2);

var orderLinesFuture = db.OrderLines
                         .Where(x => ordersQuery.Any(o => o == x.Order))
                         .OrderBy(x => x.Id)
                         .ToFuture();
```

In case paging rewriter this OrdereBy and TOP goes to IN subquery so it's OK to remove it in outer query for `exists` and `in`:

``sql
select
*
from
    OrderLines orderline0_ 
where
    exists (
        select
            order1_.OrderId 
        from
            Orders order1_ 
        where
            order1_.OrderId=orderline0_.OrderId 
            and (
                order1_.OrderId in (
                    select
                        TOP (@p0) order2_.OrderId 
                    from
                        Orders order2_ 
                    where
                        order2_.EmployeeId>@p1 
                    order by
                        order2_.OrderId asc -- this order by CAN'T be removed
                )
            --order by order2_.OrderId asc -- this can be removed
            )
        ) 
    order by
        orderline0_.OrderLineId asc;
```

**Comment by hazzik** [comment]
> You can't simply remove OrderBy

I can.

> as it does matter for sub-queries with TOP

In case of paging the order-by would be in another subquery.

**Comment by bahusoid** [comment]
> In case of paging the order-by would be in another subquery.

No it doesn't.:
```C#
var ordersQuery = db.Orders
                    .Where(x => x.Employee.EmployeeId > 5)
                    .OrderBy(x => x.OrderId)
		    .Take(2);

var orderLinesFuture = db.OrderLines
                         .Where(x => ordersQuery.Contains(o => o == x.Order))
                         .OrderBy(x => x.Id)
                         .ToFuture();
```

I see no ORDER BY in subquery nor other subquery with your changes:
```SQL
    select *
    from
        OrderLines orderline0_ 
    where
        exists (
            select
                order1_.OrderId 
            from
                Orders order1_ 
            where
                order1_.OrderId=orderline0_.OrderId 
                and (
                    order1_.OrderId in (
                        select
                            TOP (@p0) order2_.OrderId 
                        from
                            Orders order2_ 
                        where
                            order2_.EmployeeId>@p1
                    )
                )
            ) 
        order by
            orderline0_.OrderLineId asc;
```

hazzik:
Yeah, it seems it runs a second time again. And removes after PagingRewriter....

hazzik:
@bahusoid ok. Seems to be fixed.

bahusoid:
Now it seems we are doing twice job that should be done only once by your code. How safe would it be to move `RemoveUnnecessaryBodyOperators` right after `PagingRewriter`? Then it seems we could drop check introduced in https://github.com/nhibernate/nhibernate-core/commit/848b915129c5dd573302f1b59ed04c65a5f6ea15:
https://github.com/nhibernate/nhibernate-core/blob/848b915129c5dd573302f1b59ed04c65a5f6ea15/src/NHibernate/Linq/GroupBy/PagingRewriter.cs#L57

bahusoid:
> How safe would it be

Not quite...  `CanExecuteCountWithOrderByArguments` fails...

hazzik:
This thing would not pick up the QueryModel which is processed by PagingRewriter. Or at least it should not.

bahusoid:
It' just got me thinking that https://github.com/nhibernate/nhibernate-core/pull/2480 doesn't fix subqueries with `COUNT` operator:

```C#
[Test(Description = "GH2479")]
public void OrdersWithSubquery9Count()
{
	if (Dialect is MySQLDialect)
		Assert.Ignore("MySQL does not support LIMIT in subqueries.");

	var ordersQuery = db.Orders
	                    .Where(x => x.Employee.EmployeeId > 5)
	                    .OrderByDescending(x => x.OrderId)
	                    .Take(2);

	var orderLinesFuture = db.OrderLines
							.Where(x => ordersQuery.Count(o => o == x.Order) == 2)
							.OrderBy(x => x.Id)
							.ToFuture();

	var orderLines = orderLinesFuture.ToList();
}
```

Though for some reasons I don't see `Count` result operator in generated subquery.

bahusoid:
> This thing would not pick up the QueryModel which is processed by PagingRewriter. Or at least it should not.

Not sure what you meant but calling `RemoveUnnecessaryBodyOperators` right after `PagingRewriter` does remove `ORDER BY` without changes applied in https://github.com/nhibernate/nhibernate-core/commit/848b915129c5dd573302f1b59ed04c65a5f6ea15

So in theory we can just call it right from `PagingRewriter` when we detect that it might be useful instead of copy/paste checks. But for some reasons COUNT for subquery is not present in ResultOperators so it won't fix Count queries

hazzik:
> Though for some reasons I don't see Count result operator in generated subquery.

It is rewrited to `NhCountExpression` earlier.

hazzik:
> calling RemoveUnnecessaryBodyOperators right after PagingRewriter 

But RemoveUnnecessaryBodyOperators  needs to be called earlier because _some_ other rewrites could be dependent on the removed queries.

bahusoid:
> But RemoveUnnecessaryBodyOperators needs to be called earlier

I meant calling it second time only when query is converted to Contains subquery. But maybe not as it's anyway doesn't handle Count

hazzik:
@bahusoid how about that implementation?

Let's deal with `count` in a separate request.

bahusoid:
> how about that implementation?

I like it

src/NHibernate.Test/Linq/WhereSubqueryTests.cs
```diff
@@ -579,6 +618,24 @@ public void OrdersWithSubquery11()
 			Assert.That(productsNotInLargestOrders.Count, Is.EqualTo(49), nameof(productsNotInLargestOrders));
 		}
 
+		[Test]
+		public void OrdersWithSubquery11A()
+		{
+			//if (Dialect is MsSqlCeDialect)
+			//	Assert.Ignore("MS SQL CE does not support sorting on a subquery.");
+
```
bahusoid, line 626:
```suggestion
```

bahusoid: And async regen is missing

hazzik: I'll remove the commented code after the MsSqlCeDialect finished.

hazzik: Addressed.

bahusoid (approved):

PR #2516, opened 2020-09-03
Suppress Codefactor single class per file rule for test project

bahusoid:


hazzik:
I tried it did not work

bahusoid:
I tried it with #2513 and was under impression it works.. Seems got it confused with WIP :) 

Codefactor uses some strange logic when he decides that check is passed. I wouldn't be surprised if  it ignores rules from PR but will use them after commit...

bahusoid:
"не получилось, не фартануло" )  Seems we should ask codefactor  to shed light on this matter...

PR #2517, opened 2020-09-03
Fix Linq Fetch over component after fetching an association

### LINKED ISSUE #2511: Linq Fetch over component after fetching a many-to-one throws exception
I have a Linq query that looks like this:

    session.Query<MyEntity>()
        .Fetch(x => x.UnrelatedManyToOneProperty)
        .Fetch(x => x.ComponentProperty).ThenFetch(x => x.ManyToOneInComponent)
        .Where(x => x.Id == someId)
        .ToArray();

This query worked in NHibernate 5.2.7 (although it didn't actually fetch the many-to-one in the component, see #1232). But with NHibernate 5.3.2 this query now throws an exception:

    InvalidOperationException: Property x.ComponentProperty does not exist on entity TypeOfUnrelatedManyToOneProperty
       at NHibernate.Persister.Entity.AbstractEntityPersister.PropertySelectFragment(String name, String suffix, ICollection`1 fetchProperties, Boolean allProperties)

If I remove the `.Fetch(x => x.UnrelatedManyToOneProperty)` part of the query it works (and even correctly fetches the many-to-one in the component, which means that #1232 was fixed, except that we now have this regression).

maca88:
Fix a regression made by #1922.

Fixes #2511

### DIFF
--- a/src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
+++ b/src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
@@ -20,10 +20,10 @@ namespace NHibernate.Linq.Visitors.ResultOperatorProcessors
 
 		public void Process(FetchRequestBase resultOperator, QueryModelVisitor queryModelVisitor, IntermediateHqlTree tree, string sourceAlias)
 		{
-			Process(resultOperator, queryModelVisitor, tree, tree.GetFromRangeClause(), sourceAlias);
+			Process(resultOperator, queryModelVisitor, tree, null, sourceAlias);
 		}
 
-		public void Process(
+		private void Process(
 			FetchRequestBase resultOperator,
 			QueryModelVisitor queryModelVisitor,
 			IntermediateHqlTree tree,
@@ -45,11 +45,6 @@ namespace NHibernate.Linq.Visitors.ResultOperatorProcessors
 			HqlTreeNode currentNode,
 			IType propType)
 		{
-			if (currentNode == null)
-			{
-				throw new InvalidOperationException($"Property {resultOperator.RelationMember.Name} cannot be fetched for this type of query.");
-			}
-
 			if (resultOperator is FetchOneRequest)
 			{
 				if (propType == null)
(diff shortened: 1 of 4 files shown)


src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
```diff
@@ -61,20 +82,21 @@ private void Process(FetchRequestBase resultOperator, QueryModelVisitor queryMod
...
+						Process(innerFetch, queryModelVisitor, tree, memberPath, currentNode, componentType.Subtypes[subTypeIndex]);
 					}
 
 					return;
 				}
 			}
 
 			var alias = queryModelVisitor.Model.GetNewName("_");
-			tree.AddFromClause(tree.TreeBuilder.LeftFetchJoin(memberPath, tree.TreeBuilder.Alias(alias)));
+			currentNode = tree.TreeBuilder.LeftFetchJoin(memberPath, tree.TreeBuilder.Alias(alias));
```
hazzik, line 93: For the `FetchManyRequest` the `currentNode` is always set, so the null check above does not matter.

src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
```diff
@@ -19,16 +19,37 @@ public void Process(FetchRequestBase resultOperator, QueryModelVisitor queryMode
 		}
 
 		public void Process(FetchRequestBase resultOperator, QueryModelVisitor queryModelVisitor, IntermediateHqlTree tree, string sourceAlias)
+		{
+			Process(resultOperator, queryModelVisitor, tree, tree.GetFromRangeClause(), sourceAlias);
+		}
+
+		public void Process(
```
hazzik, line 26: Can this be private please?

src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
bahusoid: It seems not only `FetchManyRequest` but all fetch requests except components.

That got me thinking that `tree.GetFromRangeClause()` call should be moved inside component block as it's the only case it's needed.

src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
bahusoid: Done

src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
maca88: It is required only for property fetching so that the node is added to the correct from/join node. The applied change is valid, so I am marking this as resolved.

bahusoid (approved):

PR #2518, opened 2020-09-04
Support Aggregate subqueries with paging on MS SQL Server

bahusoid:
Continuation of  https://github.com/nhibernate/nhibernate-core/pull/2480  with addition of aggregate statements (sum/max...)

### DIFF
--- a/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
+++ b/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
@@ -559,6 +559,9 @@ where c.Order.Customer.CustomerId = 'VINET'
 			if (Dialect is MySQLDialect)
 				Assert.Ignore("MySQL does not support LIMIT in subqueries.");
 
+			if (!Dialect.SupportsScalarSubSelects)
+				Assert.Ignore(Dialect.GetType().Name + " does not support scalar sub-queries");
+
 			var ordersQuery = db.Orders
 								.Where(x => x.Employee.EmployeeId > 5)
 								.OrderByDescending(x => x.OrderId)
--- a/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
+++ b/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
@@ -553,6 +553,25 @@ where c.Order.Customer.CustomerId = 'VINET'
 			Assert.That(orderLines.Count, Is.EqualTo(711), nameof(orderLines));
 		}
 
+		[Test]
+		public void OrdersWithSubquery9Count()
+		{
+			if (Dialect is MySQLDialect)
+				Assert.Ignore("MySQL does not support LIMIT in subqueries.");
+
+			var ordersQuery = db.Orders
+								.Where(x => x.Employee.EmployeeId > 5)
+								.OrderByDescending(x => x.OrderId)
+								.Take(2);
+
+			var orderLines = db.OrderLines
+									.Where(x => ordersQuery.Count(o => o == x.Order) > 0)
+									.OrderBy(x => x.Id)
+									.ToList();
+
+			Assert.That(orderLines.Count, Is.EqualTo(4), nameof(orderLines));
+		}
+
 		[Test(Description = "GH2479")]
 		public void OrdersWithSubquery10()
 		{
--- a/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
+++ b/src/NHibernate.Test/Linq/WhereSubqueryTests.cs
@@ -575,6 +575,28 @@ where c.Order.Customer.CustomerId = 'VINET'
 			Assert.That(orderLines.Count, Is.EqualTo(4), nameof(orderLines));
 		}
 
+		[Test]
+		public void OrdersWithSubquery9Sum()
+		{
+			if (Dialect is MySQLDialect)
+				Assert.Ignore("MySQL does not support LIMIT in subqueries.");
+
+			if (!Dialect.SupportsScalarSubSelects)
+				Assert.Ignore(Dialect.GetType().Name + " does not support scalar sub-queries");
+
+			var ordersQuery = db.Orders
+								.Where(x => x.Employee.EmployeeId > 5)
+								.OrderByDescending(x => x.OrderId)
+								.Take(2);
+
+			var orderLines = db.OrderLines
+								.Where(x => ordersQuery.Where(o => o == x.Order).Sum(o => o.Freight.Value) > 0)
+								.OrderBy(x => x.Id)
+								.ToList();
+
+			Assert.That(orderLines.Count, Is.EqualTo(4), nameof(orderLines));
+		}
+
 		[Test(Description = "GH2479")]
 		public void OrdersWithSubquery10()
 		{
--- a/src/NHibernate/Linq/GroupBy/PagingRewriter.cs
+++ b/src/NHibernate/Linq/GroupBy/PagingRewriter.cs
@@ -1,4 +1,5 @@
 using System.Linq;
+using NHibernate.Linq.ReWriters;
 using NHibernate.Linq.Visitors;
 using Remotion.Linq;
 using Remotion.Linq.Clauses;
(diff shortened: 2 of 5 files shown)


src/NHibernate.Test/Linq/WhereSubqueryTests.cs
```diff
@@ -553,6 +553,28 @@ public void OrdersWithSubquery9A()
...
+			if (!Dialect.SupportsScalarSubSelects)
+				Assert.Ignore(Dialect.GetType().Name + " does not support scalar sub-queries");
+
+			var ordersQuery = db.Orders
+								.Where(x => x.Employee.EmployeeId > 5)
+								.OrderByDescending(x => x.OrderId)
+								.Take(2);
+
+			var orderLines = db.OrderLines
+									.Where(x => ordersQuery.Count(o => o == x.Order) > 0)
```
hazzik, line 571: In fact it could be anything. Count, sum, avg, etc.

src/NHibernate/Linq/ReWriters/RemoveUnnecessaryBodyOperators.cs
```diff
@@ -21,20 +22,32 @@ public static void ReWrite(QueryModel queryModel)
...
 
+		internal static bool IsOrderByNeeded(QueryModel queryModel)
+		{
+			switch (queryModel.ResultOperators.Count)
+			{
+				case 1:
+					var r = queryModel.ResultOperators[0];
+					return !(r is AnyResultOperator || r is AllResultOperator || r is ContainsResultOperator);  
+				case 0:
+					return !(queryModel.SelectClause.Selector is NhCountExpression);
```
hazzik, line 45:
```suggestion
					return !(queryModel.SelectClause.Selector is NhAggregatedExpression);
```

src/NHibernate.Test/Linq/WhereSubqueryTests.cs
hazzik: Please add a test `.Where(x => ordersQuery.Where(o => o == x.Order).Sum(o => o.OrderId) > 0)`

src/NHibernate/Linq/ReWriters/RemoveUnnecessaryBodyOperators.cs
```diff
@@ -21,20 +22,32 @@ public static void ReWrite(QueryModel queryModel)
...
 		}
 
+		internal static bool IsOrderByNeeded(QueryModel queryModel)
+		{
+			switch (queryModel.ResultOperators.Count)
+			{
+				case 1:
+					var r = queryModel.ResultOperators[0];
+					return !(r is AnyResultOperator || r is AllResultOperator || r is ContainsResultOperator);  
+				case 0:
```
hazzik, line 44: This should probably go into the aggregate paging rewiter instead.

hazzik (changes requested):

src/NHibernate.Test/Linq/WhereSubqueryTests.cs
bahusoid: Yeah.  I forgot to mention it. I tried to add aggregate functions but they don't work properly.

bahusoid: For instance sum is generated along with TOP statement:
```C#
var orderLines = db.OrderLines
					.Where(x => ordersQuery.Sum(o => o.Freight.Value) > 0)
					.OrderBy(x => x.Id)
					.ToList();

```

Will give:
```SQL
where (select TOP (@p0) (sum(order1_.Freight)) from Orders order1_ where order1_.EmployeeId>@p1 order by order1_.OrderId desc)>@p2 order by orderline0_.OrderLineId asc]
```

src/NHibernate/Linq/ReWriters/RemoveUnnecessaryBodyOperators.cs
hazzik: But... distinct would probably require order-by.

src/NHibernate.Test/Linq/WhereSubqueryTests.cs
bahusoid: Hm.. Though with additional `Where(o => o == x.Order)` as in your suggested test it works properly.. So OK I will add it

bahusoid: Done

src/NHibernate/Linq/ReWriters/RemoveUnnecessaryBodyOperators.cs
bahusoid: It's not required but we shouldn't loose it

src/NHibernate/Linq/ReWriters/RemoveUnnecessaryBodyOperators.cs
bahusoid: meaning?

hazzik: I tried to say that this logic probably belongs to some other rewriter.

src/NHibernate/Linq/ReWriters/RemoveUnnecessaryBodyOperators.cs
```diff
@@ -21,20 +22,35 @@ public static void ReWrite(QueryModel queryModel)
...
+			switch (queryModel.ResultOperators.Count)
+			{
+				case 1:
+					var r = queryModel.ResultOperators[0];
+					return !(r is AnyResultOperator || r is AllResultOperator || r is ContainsResultOperator);
+				case 0:
+					var s = queryModel.SelectClause.Selector;
+					return s is NhAggregatedExpression
+						? s is NhDistinctExpression
+						: true;
```
hazzik, line 48:
```suggestion
					return !(s is NhAggregatedExpression) || s is NhDistinctExpression;
```

bahusoid: OK... But I find it more readable the way it was...

src/NHibernate/Linq/ReWriters/RemoveUnnecessaryBodyOperators.cs
bahusoid: I don't get  how do you see it expressed in code? 
Something like:
```C#
case 0:
return AggregatingGroupJoinRewriter.IsOrderByNeeded(queryModel):
```

hazzik: No.

src/NHibernate/Linq/GroupBy/PagingRewriter.cs
```diff
@@ -53,7 +54,7 @@ private static void FlattenSubQuery(SubQueryExpression subQueryExpression, Query
 				var where = new WhereClause(new SubQueryExpression(newSubQueryModel));
 				queryModel.BodyClauses.Add(where);
 
-				if (!queryModel.BodyClauses.OfType<OrderByClause>().Any())
+				if (RemoveUnnecessaryBodyOperators.IsOrderByNeeded(queryModel) && !queryModel.BodyClauses.OfType<OrderByClause>().Any())
```
hazzik, line 57: Why have you re-introduced this check again?

src/NHibernate/Linq/ReWriters/RemoveUnnecessaryBodyOperators.cs
hazzik: I was thinking of moving the entire `0` case into `AggregatingGroupJoinRewriter` where we create `NhAggregatedExpression`'s. But it did not work, because this check is also used when we copy order-by into the `QueryModel` with `NhAggregatedExpression`

src/NHibernate/Linq/GroupBy/PagingRewriter.cs
bahusoid: Micro optimization to avoid unnecessary collection modification. What else :)

hazzik: семь пятниц на неделе.

hazzik (approved):

src/NHibernate/Linq/GroupBy/PagingRewriter.cs
bahusoid: I'm not pushing it. We can drop it you don't like it. Last time I was concerned about code duplication.

hazzik: I'm thinking. Do we actually need/want the conversion from `CountResultOperator` to `NhCountExpression` (and other similar conversions?).

bahusoid: :) That's also something I asked myself - to check why this rewriting is needed at all. As it's my understanding it's also the source of issues like https://github.com/nhibernate/nhibernate-core/issues/1124

PR #2528, opened 2020-09-11
Test case that NHibernate starts to compute the boolean value incorrect after a certain scenario

WGroenestein:


WGroenestein:
Seems already to be fixed in 5.3.3

PR #2534, opened 2020-09-14
Fix asymmetrical SqlType.Equals

bahusoid:
Fixed equality issue described in https://github.com/nhibernate/nhibernate-core/issues/1180#issuecomment-664946300

Now it's strict check that returns true only in case of full object equality. Maybe should go as possible breaking change?

### DIFF
--- a/src/NHibernate/SqlTypes/SqlType.cs
+++ b/src/NHibernate/SqlTypes/SqlType.cs
@@ -143,13 +143,13 @@ namespace NHibernate.SqlTypes
 			if (ScaleDefined != rhsSqlType.ScaleDefined)
 				return false;
 
-			if (LengthDefined && Length != rhsSqlType.Length)
+			if (Length != rhsSqlType.Length)
 				return false;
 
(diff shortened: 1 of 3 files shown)


hazzik (approved):

src/NHibernate/SqlTypes/SqlType.cs
```diff
@@ -125,24 +125,34 @@ public override bool Equals(object obj)
...
+			if (LengthDefined != rhsSqlType.LengthDefined)
+				return false;
+
+			if (PrecisionDefined != rhsSqlType.PrecisionDefined)
+				return false;
+
+			if (ScaleDefined != rhsSqlType.ScaleDefined)
+				return false;
+
+			if (LengthDefined && Length != rhsSqlType.Length)
```
hazzik, line 146: `LengthDefined &&` is not needed I believe.

hazzik (dismissed):

hazzik (approved):

PR #2538, opened 2020-09-15
Remove no longer needed alias substitution for filtered many-to-many collection in hql

bahusoid:
Partial undo of https://github.com/nhibernate/nhibernate-core/commit/c4b806cae771fa2229a66a6e2d0ad239635d0f08
as no longer needed with table group joins https://github.com/nhibernate/nhibernate-core/commit/bc1484c76e962f51eaa65e0a178a9cc95488d3cf

Test case: TestFilteredLinqQuery

### DIFF
--- a/src/NHibernate/Loader/Hql/QueryLoader.cs
+++ b/src/NHibernate/Loader/Hql/QueryLoader.cs
@@ -203,11 +203,6 @@ namespace NHibernate.Loader.Hql
 			get { return _collectionPersisters; }
 		}
 
-		protected override IDictionary<string, string[]> GetCollectionUserProvidedAlias(int index)
-		{
-			return null;
-		}
-
 		private void Initialize(SelectClause selectClause)
 		{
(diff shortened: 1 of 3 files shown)


src/NHibernate/Loader/Hql/QueryLoader.cs
```diff
@@ -206,7 +205,7 @@ protected internal override ICollectionPersister[] CollectionPersisters
 
 		protected override IDictionary<string, string[]> GetCollectionUserProvidedAlias(int index)
```
hazzik, line 206: Shall this be made obsolete?

hazzik (dismissed):

src/NHibernate/Loader/Hql/QueryLoader.cs
bahusoid: Let's keep it as feature. It's still might be useful.

hazzik (approved):

Issue #2540, opened 2020-09-15
Unable to use external predicate in subquery


krzkowal:
While this query works as expected:
```
[Test]
public void CanSelectWithWhereSubQuery()
{
    var query = from timesheet in db.Timesheets
                select new
                {
                    timesheet.Id,
                    Entries = timesheet.Entries.Where(e => e.NumberOfHours >= 0).ToList()
                };

    var list = query.ToList();

    Assert.AreEqual(3, list.Count);
}
```
(The test is a modified version of the already exising [Linq/SelectionTests.CanSelectWithAnySubQuery](https://github.com/nhibernate/nhibernate-core/blob/7fd1a9a62b7ea7e7192255baff16a92ba291e5fd/src/NHibernate.Test/Linq/SelectionTests.cs#L270))

The next test generates incorrect SQL query which results in an exception when executed

```
[Test]
public void CanSelectWithAsQueryableAndWhereSubQueryWithExternalPredicate()
{
    Expression<Func<TimesheetEntry, bool>> predicate = e => e.NumberOfHours >= 0;

    var query = from timesheet in db.Timesheets
                select new
                {
                    timesheet.Id,
                    Entries = timesheet.Entries.AsQueryable().Where(predicate).ToList()
                };

    var list = query.ToList();

    Assert.AreEqual(3, list.Count);
}
```
<details open>
  <summary>Error details</summary>

```
NHibernate.Exceptions.GenericADOException : could not execute query
[ select timesheet0_.TimesheetId as col_0_0_, (select entries1_.TimesheetEntryId from TimesheetEntries entries1_ where timesheet0_.TimesheetId=entries1_.TimesheetID and entries1_.NumberOfHours>=@p0) as col_1_0_ from Timesheets timesheet0_ ]
  Name:p1 - Value:0
[SQL: select timesheet0_.TimesheetId as col_0_0_, (select entries1_.TimesheetEntryId from TimesheetEntries entries1_ where timesheet0_.TimesheetId=entries1_.TimesheetID and entries1_.NumberOfHours>=@p0) as col_1_0_ from Timesheets timesheet0_]
  ----> System.Data.SqlClient.SqlException : Subquery returned more than 1 value. This is not permitted when the subquery follows =, !=, <, <= , >, >= or when the subquery is used as an expression.
Data:
  HelpLink.ProdName: Microsoft SQL Server
  HelpLink.ProdVer: 14.00.3192
[...]

PR #2541, opened 2020-09-15
#2540 LINQ SELECT tests with WHERE subquery

krzkowal:
Test case for #2540


bahusoid:
Moved tests to `NestedSelectsTests` and made async regen.

bahusoid (approved):

PR #2547, opened 2020-09-17
Fix paging in DB2Dialect 

hazzik:


src/NHibernate.Test/ReadOnly/TextHolder.cs
```diff
@@ -10,7 +10,7 @@ public class TextHolder
 		/// </summary>
 		public static bool SupportedForDialect(Dialect.Dialect dialect)
 		{
-			return !(dialect is FirebirdDialect || dialect is Oracle8iDialect || dialect is MsSqlCeDialect || dialect is HanaRowStoreDialect);
+			return !(dialect is FirebirdDialect || dialect is Oracle8iDialect || dialect is MsSqlCeDialect || dialect is HanaRowStoreDialect || dialect is DB2Dialect);
```
hazzik, line 13: Strictly speaking this is not related, but let it be here.

src/NHibernate/Dialect/DB2Dialect.cs
```diff
@@ -232,28 +232,29 @@ public override bool SupportsVariableLimit
...
 			if (offset == null)
 			{
-				return new SqlString(querySqlString,
-									 " fetch first ",
-									 limit,
-									 " rows only");
+				return new SqlString(sql, " fetch first ", limit, " rows only");
 			}
 
+			ExtractColumnOrAliasNames(sql, out var selectColumns, out _, out _);
```
fredericDelaporte, line 242:
```suggestion
			var selectColumns = ExtractColumnOrAliasNames(sql);
```

src/NHibernate/Dialect/DB2Dialect.cs
```diff
@@ -232,28 +232,29 @@ public override bool SupportsVariableLimit
...
 			 * " ) as tempresult where rownum between ? and ?"
 			 */
-			string rownumClause = GetRowNumber(querySqlString);
+			string rownumClause = GetRowNumber(sql);
 
 			SqlStringBuilder pagingBuilder = new SqlStringBuilder();
 			pagingBuilder
-				.Add("select * from (select ")
+				.Add("select " )
+				.Add(string.Join(",", selectColumns))
```
fredericDelaporte, line 254:
```suggestion
				.Add(selectColumns)
```

src/NHibernate/Dialect/DB2Dialect.cs
hazzik: This would require lifting `ExtractColumnOrAliasNames` from `OracleXXXDialect` to `DialectBase`

src/NHibernate/Dialect/DB2Dialect.cs
hazzik: There is no such overload.

src/NHibernate/Dialect/DB2Dialect.cs
fredericDelaporte: Sorry, I had not realized it was not in the base class.

src/NHibernate/Dialect/DB2Dialect.cs
fredericDelaporte: It would have been a string with the previous suggested change, if that previous suggestion had been valid.

fredericDelaporte (approved):

PR #2551, opened 2020-09-17
Add support for joining a subquery in hql

### LINKED ISSUE #2092: Projection and join fetch in hql leads to duplicated column aliases
```C#
var results =  session
.CreateQuery("select o.Employee.FirstName, o from Order o join fetch o.Customer")
.SetMaxResults(1)
.UniqueResult<object[]>();
```
The query above leads to duplicated column aliases for fetched Customer ID:
```sql
SELECT employee2_.FirstName AS col_0_0_
	,order0_.OrderId AS col_1_0_
	,customer1_.CustomerId AS customerid1_0_1_ -- <-HERE
	,order0_.OrderId AS orderid1_3_0_
	,customer1_.CustomerId AS customerid1_0_1_ -- <-AND HERE
...
```
And that doesn't work on Oracle (for queries with paging) and  SQL Server CE, SQL Anywhere.

Related issue:
https://nhibernate.jira.com/browse/NH-1773

maca88:
Basic hql syntax:
```
from Order o 
inner join (
  from Order where OrderId = 10248
) o2 on (o.OrderId - 1) = o2.OrderId
```

In case the subquery returns multiple values, each value can be referenced by using an alias:
```
select o, o2.ord, o2.cu
from Order o 
inner join (
  select o1 as ord, o1.Customer as cu
  from Order o1 where o1.OrderId = 10248
) o2 on (o.OrderId - 1) = o2.ord.OrderId
```

Aliases can be nested in case of nested subqueries:
```
select 
  o4.ord3, o4.ord3.ShippedTo, o4.cu3, o4.ord3.ShippingAddress.City,
  o4.sub2.ord2, o4.sub2.ord2.ShippedTo, o4.sub2.cu2, o4.sub2.cu2.Address.City,
  o4.sub2.sub1.ord1, o4.sub2.sub1.ord1.ShippedTo, o4.sub2.sub1.cu1, o4.sub2.sub1.cu1.Address.City
from Order o5
inner join (
  select o as ord3, o.Customer as cu3, o3 as sub2
  from Order o
  left join (
    select o1 as ord2, o1.Customer as cu2, o2 as sub1
    from Order o1
    left join (
      select o0 as ord1, o0.Customer as cu1
      from Order o0
    ) o2 on o1.OrderId = o2.ord1.OrderId-1
  ) o3 on o.OrderId = o3.ord2.OrderId-1
  where o.OrderId = 10248
) o4 on o5.OrderId = o4.ord3.OrderId
```

This feature is a requirement in order to support the `Include` method on all databases when paging results. Sql `IN` operator does not work with composite keys and in some databases even for normal keys (e.g. MySql), so with this feature we can write the following query:
```
select o
from CompositeOrder o 
inner join (
  select id
  from CompositeOrder
  order by OrderId skip 2 take 2
) o2 on o.id = o2.id
left join fetch o.OrderLines
```

Known limitation:
- Selecting a hql constructor inside a subquery is not supported 

Also fixes #2092


src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
```diff
@@ -509,6 +521,10 @@ private void DereferenceEntityJoin(string classAlias, EntityType propertyType, b
 
 			if ( ! useFoundFromElement )
 			{
+				// Create a new FROM node for the referenced class.
+				var associatedEntityName = propertyType.GetAssociatedEntityName();
+				var tableAlias = AliasGenerator.CreateName(associatedEntityName);
```
maca88, line 526: This was moved only to avoid generating aliases names that won't be used.

src/NHibernate/Hql/Ast/ANTLR/Tree/AggregateNode.cs
```diff
@@ -14,7 +14,7 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 	/// Ported by: Steve Strong
 	/// </summary>
 	[CLSCompliant(false)]
-	public class AggregateNode : AbstractSelectExpression, ISelectExpression
```
maca88, line 17: Removed `ISelectExpression` as `AbstractSelectExpression` implements it.

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -2350,21 +2391,19 @@ private void InternalInitSubclassPropertyAliasesMap(string path, IEnumerable<Pro
...
+				var cols = new string[prop.ColumnSpan];
+				var l = 0;
+				foreach (var thing in prop.ColumnIterator)
+				{
+					aliases[l] = thing.GetAlias(Factory.Dialect, prop.Value.Table);
+					cols[l] = thing.GetText(Factory.Dialect);
+					l++;
 				}
+
+				subclassPropertyAliases[propName] = aliases;
```
maca88, line 2405: Now `subclassPropertyAliases` contains also component properties which is needed in order to select component properties from a subquery.

bahusoid:
Does it work with parameters in subquery? I see no such tests...
```sql
from Order o 
inner join (
  from Order where OrderId = :id
) o2 on (o.OrderId - 1) = o2.OrderId
```

maca88:
Forgot to test it, now it works and added tests.

src/NHibernate/Hql/Ast/ANTLR/Tree/JoinSubqueryFromElement.cs
```diff
@@ -57,7 +57,9 @@ private IType CreateDataType(SelectClause selectClause, out IEntityPersister ent
 
 		public SqlString RenderText(SqlString subQuery, ISessionFactoryImplementor sessionFactory)
 		{
-			return RenderText(sessionFactory).Replace("{query}", subQuery.ToString());
+			var array = RenderText(sessionFactory).Split("{query}");
+
+			return new SqlString(new object[] {array[0], subQuery, array[1]});
```
bahusoid, line 62: I think more optimal would be doing `IndexOf` and `Substring`. Something like:
```C#
var index = renderText.IndexOfOrdinal("{query}");
return new SqlString(
	renderText.Substring(0, index),
	subQuery,
	renderText.Substring(index + 7));
```

maca88: Suggestion applied.

bahusoid:
There is I believe another restriction (which is also present for entity join). It's not possible to add association join dependent on subquery join. Something like:
```sql
from Order o 
inner join (from Order where OrderId = :id) o2 on (o.OrderId - 1) = o2.OrderId 
inner join o.Employee e on e = o2.Employee 
```
Employee join will be added right after Order join (and it seems only SQLite is able to handle such queries with invalid order of joins). AFAIK it's not a big issue for LINQ as we don't support association joins there. That's just FYI maybe you see an easy fix for this.

bahusoid:
And your alias related changes seem to fix issue reported here:
https://github.com/nhibernate/nhibernate-core/issues/2092

I see no duplicates in generated SQL

maca88:
> It's not possible to add association join dependent on subquery join

With the last commit the issue is fixed.

src/NHibernate/Hql/Ast/ANTLR/Tree/FromElementType.cs
```diff
@@ -65,32 +72,26 @@ public virtual IType DataType
...
 			{
-				if (_entityType == null)
+				if (!(_type is EntityType entityType))
 				{
 					return null;
 				}
 
 				bool shallow = _fromElement.FromClause.Walker.IsShallowQuery;
-				return TypeFactory.ManyToOne(_entityType.GetAssociatedEntityName(), shallow);
+				return TypeFactory.ManyToOne(entityType.GetAssociatedEntityName(), shallow);
```
bahusoid, line 94: Can we cache it? Maybe calculate it once right inside ctor? It seems it's called quite extensively when query is processed

src/NHibernate/Persister/Entity/SubqueryPropertyMapping.cs
```diff
@@ -0,0 +1,346 @@
...
+				)
+				{
+					return true;
+				}
+			}
+
+			if (propertyName == EntityPersister.EntityID)
+			{
+				columns = GetIdentifiersColumns(alias).ToArray();
+				return columns.Length > 0;
```
bahusoid, line 287: Should we cache results in `_propertyColumns`?

src/NHibernate/Hql/Ast/ANTLR/Tree/FromElementType.cs
bahusoid: > It seems SelectType is called quite extensively when query is processed

Actually no. I'm wrong about "quite extensively". Still it's executed twice from `InitializeDerivedSelectClause`. Can you refactor it to avoid double object creation? Maybe something like:
```C#
IType type;
if (fromElement.IsFetch || fromElement.IsCollectionOfValuesOrComponents || ((type = fromElement.SelectType) == null))
...
node.DataType = type;
```

src/NHibernate/Hql/Ast/ANTLR/Tree/JoinSubqueryFromElement.cs
```diff
@@ -0,0 +1,80 @@
...
+		{
+			var tableAlias = fromClause.AliasGenerator.CreateName("subquery");
+			var querySelectClause = queryNode.GetSelectClause();
+			var dataType = CreateDataType(querySelectClause, out var entityPersister);
+			PropertyMapping = new SubqueryPropertyMapping(dataType, querySelectClause);
+			QueryNode = queryNode;
+			DataType = dataType;
+			foreach (var fromElement in querySelectClause.SelectExpressions.Select(o => o.FromElement)
+				.Union(querySelectClause.NonScalarExpressions?.Select(o => o.FromElement) ?? Enumerable.Empty<FromElement>())
+				.Union(querySelectClause.CollectionFromElements ?? Enumerable.Empty<FromElement>())
```
bahusoid, line 23: If I'm not mistaken each `Union` call leads to `Set` creation with all existing items So it's better  to do `Concat` with single `Distinct` call in the end.

src/NHibernate/Persister/Collection/IQueryableCollection.cs
```diff
@@ -91,4 +95,43 @@ public interface IQueryableCollection : IPropertyMapping, IJoinable, ICollection
...
+				return collectionPersister.GetSelectFragment(alias, columnSuffix);
+			}
+
+#pragma warning disable 618
+			var renderedText = queryable.SelectFragment(alias, columnSuffix);
+#pragma warning restore 618
+			var identifierAlias = queryable.GetIdentifierColumnAlias(null);
+			var indexAliases = queryable.GetIndexColumnAliases(null);
+			var columnAliases = queryable.GetKeyColumnAliases(null)
+			                             .Union(queryable.GetElementColumnAliases(null));
```
bahusoid, line 122: Same.. Better to `Concat` and then once `Distinct`

src/NHibernate/Hql/Ast/ANTLR/Tree/JoinSubqueryFromElement.cs
```diff
@@ -0,0 +1,80 @@
...
+				renderText.Substring(0, index),
+				subQuery,
+				renderText.Substring(index + 7));
+		}
+
+		public override string GetIdentityColumn()
+		{
+			// Return null for a scalar subquery instead of throwing an exception.
+			// The node will be removed in the SelectClause
+			if (DataType is SubqueryComponentType && PropertyMapping.GetIdentifiersColumns(TableAlias).Count == 0)
```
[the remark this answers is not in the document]
maca88: Refactored to use `GetIdentifiersColumns` method result.

src/NHibernate/Hql/Ast/ANTLR/Tree/FromElementType.cs
maca88: Suggestion applied.

src/NHibernate/Hql/Ast/ANTLR/Tree/JoinSubqueryFromElement.cs
maca88: Done as suggested.

src/NHibernate/Persister/Collection/IQueryableCollection.cs
maca88: Done as suggested.

src/NHibernate/Persister/Entity/SubqueryPropertyMapping.cs
maca88: `_propertyColumns` contains only columns names, where `GetIdentifiersColumns` returns qualified columns. As alias may be different (nested subqueries) I rather cached non scalar `FromElement`s to avoid using a `HashSet` in `GetPropertiesColumns` and `GetIdentifiersColumns` methods.

fredericDelaporte (commented):
I have not reviewed the SelectClause changes. I lack knowledge of how it works there.

src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
```diff
@@ -682,6 +743,32 @@ void PrepareFilterParameter()
...
+		void CreateJoinSubquery(
+			IASTNode query,
+			IASTNode alias,
+			int joinType,
+			IASTNode with)
+		{
+			if (log.IsDebugEnabled())
+			{
+				log.Debug($"Creating subquery-join FromElement [{alias?.Text}]");
+			}
```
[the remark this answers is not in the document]
maca88: I copied it from method `CreateEntityJoin` without much thinking, but I do agree with you.

bahusoid:
@maca88 Some tests are failed after merge... Can you take a look?

maca88:
The tests were failing because a new formula was added to `Order` entity:
`<property name="RequiredOrderDate" formula="OrderDate" insert="false" update="false" />`
I just increased the expected number of columns by two as we are selecting two orders in one query.

bahusoid (approved):

bahusoid (dismissed):

bahusoid (approved):

src/NHibernate/Hql/Ast/ANTLR/Tree/SelectClause.cs
```diff
@@ -390,139 +437,286 @@ private void InitAliases(ISelectExpression[] selectExpressions)
...
-						{
-							RenderNonScalarProperties(appender, fromElement, nonscalarSize, k);
-							k++;
-						}
-					}
+					continue;
+				}
+
+				var node = (IASTNode) e;
+				if (processedElements.Add(fromElement))
```
fredericDelaporte, line 497: @maca88, this causes #3489 regression.
The query is
```
insert into Enrolment (Course, Student)
  select e.Course, e.Student from Enrolment e
```

At the second column, the from is already added, causing the execution to go into the else and not rendering the second column.

In debug, forcing the execution to jump directly to `RenderNonScalarIdentifiers` (so, not adding again the from to combinedFromElements) allows the test to succeed.

But surely the fix is not as simple as throwing away the else in all cases with something like:
```c#
if (processedElements.Add(fromElement))
{
	combinedFromElements.Add(fromElement);
}
RenderNonScalarIdentifiers(fromElement, inheritedExpressions.ContainsKey(e) ? null : e, appender);
```

fredericDelaporte: Doing that change seems to break only `HqlDuplicateEntitySelectionSubQuery` which looks like testing invalid cases, containing duplicate selection of entities. Why should we support theses cases?

Issue #2530, opened 2020-09-11
Unable to perform AfterTransactionCompletion callback: System.NullReferenceException after update to 5.3.2 from 5.2.7


xumix:
This started after updating to 5.3.2 from 5.2.7. I can't create a repro since some tests work ok.

```
Stack Trace: 
    AfterTransactionCompletionProcessQueue.AfterTransactionCompletion(Boolean success)
    SessionImpl.AfterTransactionCompletion(Boolean success, ITransaction tx)
    AdoTransaction.AfterTransactionCompletion(Boolean successful)
    AdoTransaction.Commit()
// ......... Skipped
    TestHelper.Execute(Action action, Boolean useInterceptors) line 69
    StatisticsJobTests.Execute_ActiveCount_NoExceptions() line 414
    ----- Inner Stack Trace -----
    ReadOnlyCache.Clear()
    SessionFactoryImpl.EvictEntity(IEnumerable`1 entityNames)
    SessionFactoryExtension.EvictEntity(ISessionFactory factory, IEnumerable`1 entityNames)
    BulkOperationCleanupAction.EvictEntityRegions()
    BulkOperationCleanupAction.ExecuteAfterTransactionCompletion(Boolean success)
    AfterTransactionCompletionProcessQueue.AfterTransactionCompletion(Boolean success)
```
Config:
```
<hibernate-configuration xmlns="urn:nhibernate-configuration-2.2">
    <session-factory>
      <property name="connection.provider">NHibernate.Connection.DriverConnectionProvider</property>
      <!--<property name="dialect">NHibernate.Dialect.MsSql2012Dialect</property>
      <property name="connection.driver_class">NHibernate.Driver.SqlClientDriver</property>-->
      <!--<property name="connection.driver_class">Granit.Core.Data.NHibernate.Oracle.OracleManagedDataClientDriver, Granit.Core</property>
      <property name="dialect">NHibernate.Dialect.Oracle10gDialect</property>-->
      <property name="connection.connection_string_name">Current</property>
      <property name="command_timeout">60</property>
      <property name="show_sql">false</property>
      <property name="format_sql">true</property>
      <property name="adonet.batch_size">100</property>
      <property name="cache.provider_class">NHibernate.Caches.SysCache.SysCacheProvider, NHibernate.Caches.SysCache</property>
      <property name="cache.use_second_level_cache">true</property>
      <property name="cache.use_query_cache" >true</property>
    </session-factory>
  </hibernate-configuration>
[...]

PR #2554, opened 2020-09-20
Avoid NRE in cache strategies

fredericDelaporte:
This fixes a regression from #1480, where destroying the externally supplied and shared cache has been replaced by setting it to null.

Related: #2530.

But this change causes further usages of the cache to cause null reference exceptions, instead of meaningful errors.

This PR is in my opinion the most correct minimal fix.

Another option could be to just remove the line setting the cache to null, effectively causing `Destroy` to be no-op. But this would mean letting the user use "destroyed" caches, which seems a bit wrong. And in the case of read-write caches, other failures would occur due to some internal object being disposed.

### DIFF
--- a/src/NHibernate/Cache/CacheFactory.cs
+++ b/src/NHibernate/Cache/CacheFactory.cs
@@ -60,7 +60,7 @@ namespace NHibernate.Cache
 		public static ICacheConcurrencyStrategy CreateCache(string usage, CacheBase cache)
 		{
 			if (log.IsDebugEnabled())
-				log.Debug("cache for: {0} usage strategy: {1}", cache.RegionName, usage);
+				log.Debug("cache for: {0} usage strategy: {1}", cache?.RegionName, usage);
 
 			ICacheConcurrencyStrategy ccs;
 			switch (usage)
--- a/src/NHibernate/Cache/NonstrictReadWriteCache.cs
+++ b/src/NHibernate/Cache/NonstrictReadWriteCache.cs
@@ -47,7 +47,6 @@ namespace NHibernate.Cache
 					throw new InvalidOperationException(_isDestroyed ? "The cache has already been destroyed" : "The concrete cache is not defined");
 				return _cache;
 			}
-			set => _cache = value;
 		}
 
 		// 6.0 TODO: remove
--- a/src/NHibernate/Cache/NonstrictReadWriteCache.cs
+++ b/src/NHibernate/Cache/NonstrictReadWriteCache.cs
@@ -38,7 +38,19 @@ namespace NHibernate.Cache
 			set { _cache = value?.AsCacheBase(); }
 		}
 
-		// 6.0 TODO: make implicit and switch to auto-property
+		// 6.0 TODO: Rename to Cache and make public (possible breaking change for reader when null).
+		private CacheBase InternalCache
+		{
+			get
+			{
+				if (_cache == null || _isDestroyed)
+					throw new InvalidOperationException(_isDestroyed ? "The cache has already been destroyed" : "The concrete cache is not defined");
+				return _cache;
+			}
+			set => _cache = value;
+		}
+
+		// 6.0 TODO: remove
 		CacheBase IBatchableCacheConcurrencyStrategy.Cache
 		{
 			get => _cache;
@@ -50,13 +62,12 @@ namespace NHibernate.Cache
 		/// </summary>
 		public object Get(CacheKey key, long txTimestamp)
 		{
-			CheckCache();
 			if (log.IsDebugEnabled())
 			{
 				log.Debug("Cache lookup: {0}", key);
 			}
 
-			var result = Cache.Get(key);
+			var result = InternalCache.Get(key);
 			if (log.IsDebugEnabled())
 			{
 				log.Debug(result != null ? "Cache hit: {0}" : "Cache miss: {0}", key);
@@ -67,13 +78,12 @@ namespace NHibernate.Cache
 
 		public object[] GetMany(CacheKey[] keys, long timestamp)
 		{
-			CheckCache();
 			if (log.IsDebugEnabled())
 			{
 				log.Debug("Cache lookup: {0}", string.Join(",", keys.AsEnumerable()));
 			}
 
-			var results = _cache.GetMany(keys);
+			var results = InternalCache.GetMany(keys);
 			if (log.IsDebugEnabled())
 			{
 				log.Debug("Cache hit: {0}", string.Join(",", keys.Where((k, i) => results != null)));
@@ -97,8 +107,6 @@ namespace NHibernate.Cache
 				return result;
 			}
 
-			CheckCache();
-
 			var checkKeys = new List<object>();
 			var checkKeyIndexes = new List<int>();
 			for (var i = 0; i < minimalPuts.Length; i++)
@@ -109,10 +117,12 @@ namespace NHibernate.Cache
 					checkKeyIndexes.Add(i);
 				}
 			}
+
+			var cache = InternalCache;
 			var skipKeyIndexes = new HashSet<int>();
 			if (checkKeys.Any())
 			{
-				var objects = _cache.GetMany(checkKeys.ToArray());
+				var objects = cache.GetMany(checkKeys.ToArray());
 				for (var i = 0; i < objects.Length; i++)
 				{
 					if (objects[i] != null)
@@ -144,7 +154,7 @@ namespace NHibernate.Cache
 				putValues[j++] = values[i];
 				result[i] = true;
 			}
-			_cache.PutMany(putKeys, putValues);
+			cache.PutMany(putKeys, putValues);
 			return result;
 		}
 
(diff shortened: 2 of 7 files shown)


src/NHibernate/Cache/CacheFactory.cs
```diff
@@ -60,7 +60,7 @@ public static class CacheFactory
 		public static ICacheConcurrencyStrategy CreateCache(string usage, CacheBase cache)
 		{
 			if (log.IsDebugEnabled())
-				log.Debug("cache for: {0} usage strategy: {1}", cache.RegionName, usage);
+				log.Debug("cache for: {0} usage strategy: {1}", cache?.RegionName, usage);
```
fredericDelaporte, line 63: Creating a strategy with a `null` cache is very likely not expected, but still, it does "work", unless debug logs are enabled.
It would be better to unconditionally throw in case of null, but that would be a possible breaking change, so it should not be done in a 5.3.x

src/NHibernate/Cache/NonstrictReadWriteCache.cs
```diff
@@ -18,13 +19,14 @@ public partial class NonstrictReadWriteCache : IBatchableCacheConcurrencyStrateg
...
 		private CacheBase _cache;
+		private bool _isDestroyed;
 
 		/// <summary>
 		/// Gets the cache region name.
 		/// </summary>
 		public string RegionName
 		{
-			get { return Cache.RegionName; }
+			get { return Cache?.RegionName; }
```
fredericDelaporte, line 29: We could `CheckCache` here instead, but this seems to me good enough.

src/NHibernate/Cache/NonstrictReadWriteCache.cs
```diff
@@ -261,5 +274,11 @@ public bool AfterInsert(CacheKey key, object value, object version)
 		{
 			return false;
 		}
+
+		private void CheckCache()
+		{
+			if (_cache == null || _isDestroyed)
+				throw new InvalidOperationException(_isDestroyed ? "The cache has already been destroyed" : "The concrete cache is not defined");
```
fredericDelaporte, line 281: These strategies classes are instantiated with a `null` underlying cache, and a later set on their `Cache` property is supplying their underlying cache. Nothing prevents this underlying cache of staying `null` or being set to `null`, without the cache having been destroyed.

So we cannot just "guess" that the strategy has been destroyed if its underlying cache is `null`, it may be wrong.

I do not see any valid reason for the build pattern actually used by theses classes. It should likely be reworked for taking the underlying cache as a constructor parameter, mandatory, and setters should be obsoleted and removed in next major.
But that would be out of the scope of this PR.

bahusoid:
TBH I don't like this change. All these checks just to throw not NRE  but different kind of exception feels like not useful change that unnecessarily clutters the code.

So I would either go with:
> just remove the line setting the cache to null, effectively causing Destroy to be no-op

Or maybe introduce some kind of "Null Object" cache and assign it instead. Which either simply makes no op on calls (it seems we already have one - `FakeCache`) or throws  `InvalidOperationException`.

fredericDelaporte:
And what about the read-write case, which will fail on its disposed lock? It is no better than a NRE and we should also handle this case more gracefully, don't you think?

bahusoid:
> And what about the read-write case

Sorry I don't know this part of code -  so have no idea what would fail on its disposed lock and in which case (Empty destroy method or with "Null Value Object" cache or in both cases). 
So implementing `FakeCache` like class that throws user friendly exception and use it instead of null won't work? Ok.. Then I have no better suggestions..

> It is no better than a NRE and we should also handle this case more gracefully, don't you think?

I am for graceful behavior if it's possible to be implemented in one place. I don't like scattering such checks across all implementations. If it's the only choice - I think it's not worth it for the rare cases when object is used for disposed session factory.

So that's just my thoughts I'm not requesting you to make any changes (sorry forgot to nit :) ). Let it be reviewed by someone else.

src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -96,6 +98,7 @@ private int NextLockId()
 		/// </remarks>
 		public object Get(CacheKey key, long txTimestamp)
 		{
+			CheckCache();
 			using (_asyncReaderWriterLock.ReadLock())
```
fredericDelaporte, line 102: In the (strict) read-write strategy case, most cache operations acquire this lock. This class `Destroy` does dispose it, so most operations, in case of continued usage, would likely throw a disposed object exception from within the lock.

Somehow it is actually better than a NRE, but still not good: the strategies classes are "destroyed", so throwing a disposed object exception can be considered suitable. Excepted that it should be the strategy class throwing it directly, instead of some internal object it holds. Letting this been thrown by some internal object can puzzle end users as whether it is a bug inside NHibernate or not, like that strategy class having a bug causing usage of a disposed object.

hazzik:
Would it be better if `CheckCache` is encapsulated into `Cache` property and the usage of `_cache` field is updated with the usage of property?

hazzik:
Another thought -- can CheckCache be incorporated into `CacheBase`?

fredericDelaporte:
Yes I dislike too this bloat code for checking the object state everywhere.

But there is also the `_asyncReaderWriterLock` which can be disposed and so needs to be protected by those checks. Maybe we can use your first suggestion and dodge the lock trouble by nullifying it and using `?.` every where on it to dodge the issue and let the later `Cache` access raise the exception, but that would be a code smell I think, those `?.` on the lock.

I will use your suggestion for the two other caches at least.

About incorporating `CheckCache` in `CacheBase`, I do not get it right now, I will have another look.

hazzik:
> incorporating CheckCache in CacheBase, I do not get it right now, I will have another look.

Now most of our calls look like this:

```csharp
// prop definition simplified
CacheBase Cache => _cache;

void SomeMethod() {
    CheckCache();
    _cache.DoSomething();
...
}
```

```csharp
// prop definition simplified
CacheBase Cache { get { CheckCache(); return _cache; } }

void SomeMethod() {
    Cache.DoSomething();
...
}
```

fredericDelaporte:
I thought you wanted me to put something in the `CacheBase` class, but instead you refer to the `CacheBase` typed property versus the `ICache` one.
One trouble is, currently, the `CacheBase` one is an explicit interface implementation and so is unpractical to directly call.
The `ICache` one is on the path to removal.

Maybe we should dodge that by adding an `InternalCache` property, with a `todo 6.0 move code into CacheBase Cache and remove property`.

It would also avoid adding a possible breaking change in a minor by starting to throw on read of those interface properties for `null` caches.

src/NHibernate/Cache/ReadWriteCache.cs
fredericDelaporte: Now it is protected by acquiring the cache prior to accessing the lock, with the cache acquisition throwing if the strategy is already destroyed.
This is a bit brittle pattern since we could acquire the cache later...

hazzik:
> I thought you wanted me to put something in the CacheBase class, but instead you refer to the CacheBase typed property versus the ICache one.

There were 2 suggestions actually. One is about CacheBase class and another one about properties (both properties in fact).

hazzik:
Can you explain why you need additional property `InternalCache`?

fredericDelaporte:
As commented before doing it:
- To avoid a possible breaking change in a SemVer patch release. Changing one of the existing property for throwing when the cache is null, instead of just yielding it, is a possible breaking change since theses properties are public.
- Doing it in the `CacheBase` typed property would require some cumbersome, bloat code everywhere to ensure we call it, since it is an explicit implementation. Something like `IBatchableCacheConcurrencyStrategy strategy = this; var cache = strategy.Cache;`.

fredericDelaporte:
I have added a commit to remove `InternalCache` property. But as explained, it adds some other kind of bloat code, in order to be able to call the explicitly implemented `CacheBase` property. The `ICache` property is not suitable because `ICache` does not have all the required methods. By example we cannot call `GetMany` on an `ICache` typed property.

hazzik:
> I have added a commit to remove InternalCache property.

I have not had anything against this property. I was just wondering why it is needed. The explanation and reasoning was fine.

fredericDelaporte:
Reverted.

fredericDelaporte:
This PR is holding the release of 5.3.4 since ten days (latest merge done on the 5.3.x branch). It should be validated or rejected, and then I would prepare the release of the 5.3.4 fixes.

bahusoid:
> It should be validated or rejected

I still think the issue is not worth changes made in this PR. And it shouldn't hold the release - this PR can wait till next round as NRE or other exception it's about user invalid  usage case scenario.

hazzik:
I lean on agreeing with @bahusoid on the issue. We probably should release the 5.3.4 without this change.

mazharqayyum:
> > It should be validated or rejected
> 
> I still think the issue is not worth changes made in this PR. And it shouldn't hold the release - this PR can wait till next round as NRE or other exception it's about user invalid usage case scenario.

Can you please expand on invalid usage scenario? ISQL DDL query causes this exception, I have the L2 SysCache configured. Its a known behavior of Nhibernate that any direct update to database will wipe the L2 cache. It seems like a good deal for really rare scenario which in my case is. So what is the recommended way in 5.3.x?

bahusoid:
This PR just changes one exception with another. It doesn't fix anything. So it won't help you

> Can you please expand on invalid usage scenario

https://github.com/nhibernate/nhibernate-core/issues/2530#issuecomment-691234620

So if you can prove that it's reproducible with still open session factory (check ISessionFactory.IsClosed) we should investigate it more closely.

PR #2555, opened 2020-09-21
Add spaces around concat operator

hazzik:
Some dialects (DB2 for example) are confused by the concat operator without spaces. So let's just tweak it.

I did consider overriding only DB2Dialect but I think it is more beneficial if applied to all dialects.

### DIFF
--- a/src/NHibernate/Dialect/Dialect.cs
+++ b/src/NHibernate/Dialect/Dialect.cs
@@ -101,7 +101,7 @@ namespace NHibernate.Dialect
 			RegisterFunction("cast", new CastFunction());
 			RegisterFunction("transparentcast", new TransparentCastFunction());
 			RegisterFunction("extract", new AnsiExtractFunction());
-			RegisterFunction("concat", new VarArgsSQLFunction(NHibernateUtil.String, "(", "||", ")"));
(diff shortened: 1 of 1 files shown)


bahusoid (approved):

fredericDelaporte (commented):
Isn't that more a bug fix than an improvement? I tend to think that

> Some dialects (DB2 for example) are confused by the concat operator without spaces

means these dialects yield wrong results in such case or fail to execute the query.

This PR could also have removed the now redundant identical fix in HanaDialectBase.

hazzik:
> This PR could also have removed the now redundant identical fix in HanaDialectBase.

It is commented out there

> Isn't that more a bug fix than an improvement?

Maybe... I’ll reclassify

bahusoid:
> It is commented out there

Nope:
https://github.com/nhibernate/nhibernate-core/blob/14e2d87abb66fd8ffa2a986158205f7db786e50c/src/NHibernate/Dialect/HanaDialectBase.cs#L457

hazzik:
> Nope

Ok, I was thinking about some other dialect.

fredericDelaporte:
Yes another dialect has it commented out, with a bunch of other registrations. (Code commented out should be removed, there is not point in keeping it in most case. But that was not the subject of this PR of course.)

PR #2568, opened 2020-09-25
Support internal entity classes by proxy factory

bahusoid:
We do support proxy for internal interfaces since https://github.com/nhibernate/nhibernate-core/pull/1814
So not supporting internal classes looks like omission to me


fredericDelaporte:
Do you mean it was supported prior to #1814? It would be a regression then.

bahusoid:
>Do you mean it  it was supported prior to #1814

No. Why? #1814 added support only for internal interfaces. This one adds support for internal base class.

hazzik (approved):

PR #2570, opened 2020-09-26
Update Relinq and LinFu links

wallymathieu:
Updated links on the readme since the projects pages have moved.


fredericDelaporte (approved):
Thanks

PR #2571, opened 2020-09-28
Default value for CancellationToken in IQueryBatch.GetResultAsync 

lofcz:
I'd like to propose default value of second parameter in function `IQueryBatch.GetResultAsync` to be `default(CancellationToken)`. This would be consistent with the rest of async functions `SingleOrDefaultAsync, ListAsync` and `RowCountAsync`, where in each and every case this is the default value for cancellation. 

To my best knowledge this should be nonbreaking change. Due to triviality I don't think any tests are required here.


hazzik:
These files are auto-generated. This would need to be done with AsyncGenerator.yml

lofcz:
Thanks @hazzik

fredericDelaporte (approved):

PR #2573, opened 2020-09-30
Criteria API - Added support for Full-Text Search in SQL Server & SQLite

mihaicodrean:
Verified in production as well, against SQL Server.
The SQLite bit was only tested as far as to see the generated SQL, to verify the abstraction.

### DIFF
--- a/src/NHibernate/Criterion/ContainsExpression.cs
+++ b/src/NHibernate/Criterion/ContainsExpression.cs
@@ -1,0 +1,79 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using NHibernate.Dialect;
+using NHibernate.Engine;
+using NHibernate.SqlCommand;
+using NHibernate.Util;
+
+namespace NHibernate.Criterion
+{
+	public class ContainsExpression : AbstractCriterion
+	{
+		private readonly object value;
+		private readonly IProjection projection;
+		private readonly TypedValue typedValue;
+		private readonly bool freetext;
+
+		public ContainsExpression(IProjection projection, object value, bool freetext)
+		{
+			this.projection = projection;
+			this.value = value;
+			typedValue = new TypedValue(NHibernateUtil.String, this.value, false);
+			this.freetext = freetext;
+		}
+
+		public ContainsExpression(string propertyName, object value, bool freetext)
+			: this(Projections.Property(propertyName), value, freetext)
+		{
+		}
+
+		public override SqlString ToSqlString(ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
+			var columns = CriterionUtil.GetColumnNamesAsSqlStringParts(projection, criteriaQuery, criteria);
+			var value = criteriaQuery.NewQueryParameter(typedValue).Single();
+			var arguments = new[] { columns[0], value };
+			var functionName = GetFunctionName(criteriaQuery);
+
+			var dialectFunction = criteriaQuery.Factory.SQLFunctionRegistry.FindSQLFunction(functionName);
+			if (dialectFunction == null)
+			{
+				throw new HibernateException(string.Format("The current dialect '{0}' doesn't support the function: '{1}'",
+					criteriaQuery.Factory.Dialect, functionName));
+			}
+			return dialectFunction.Render(arguments, criteriaQuery.Factory);
+		}
+
+		private string GetFunctionName(ICriteriaQuery criteriaQuery)
+		{
+			var dialect = criteriaQuery.Factory.Dialect;
+
+			if (dialect is MsSql2000Dialect)
+				return freetext ? "freetext" : "contains";
+
+			if (dialect is SQLiteDialect)
+				return "match";
+
+			return "contains";
+		}
+
+		public override TypedValue[] GetTypedValues(ICriteria criteria, ICriteriaQuery criteriaQuery)
+		{
+			return new TypedValue[] { typedValue };
+		}
+
+		public override IProjection[] GetProjections()
+		{
+			if (projection != null)
+			{
+				return new IProjection[] { projection };
+			}
+			return null;
+		}
+
+		/// <summary></summary>
+		public override string ToString()
+		{
+			return projection + " contains/freetext " + value;
+		}
+	}
+}
--- a/src/NHibernate/Criterion/ContainsExpression.cs
+++ b/src/NHibernate/Criterion/ContainsExpression.cs
@@ -1,12 +1,12 @@
-﻿using System.Collections.Generic;
+﻿using System;
 using System.Linq;
 using NHibernate.Dialect;
 using NHibernate.Engine;
 using NHibernate.SqlCommand;
-using NHibernate.Util;
 
 namespace NHibernate.Criterion
 {
+	[Serializable]
 	public class ContainsExpression : AbstractCriterion
 	{
 		private readonly object value;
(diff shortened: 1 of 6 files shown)


src/NHibernate/Criterion/ContainsExpression.cs
```diff
@@ -0,0 +1,79 @@
...
+					criteriaQuery.Factory.Dialect, functionName));
+			}
+			return dialectFunction.Render(arguments, criteriaQuery.Factory);
+		}
+
+		private string GetFunctionName(ICriteriaQuery criteriaQuery)
+		{
+			var dialect = criteriaQuery.Factory.Dialect;
+
+			if (dialect is MsSql2000Dialect)
```
gliljas, line 50: Asking for type is a code smell. Not that it doesn't happen elsewhere in NHibernate, but I'd rather not introduce more of it.

src/NHibernate/Criterion/ContainsExpression.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System;
+using System.Linq;
+using NHibernate.Dialect;
+using NHibernate.Engine;
+using NHibernate.SqlCommand;
+
+namespace NHibernate.Criterion
+{
+	[Serializable]
+	public class ContainsExpression : AbstractCriterion
```
gliljas, line 10: This is a very DBMS specific thing. SQLite "match" is not the same thing as SQL Server "contains" (which required enabling full text indexing). Calling it ContainsExpression may lead to unintended use. Perhaps it should be in companion libraries instead.

src/NHibernate/Criterion/ContainsExpression.cs
mihaicodrean: What do you suggest to do instead?

src/NHibernate/Criterion/ContainsExpression.cs
mihaicodrean: MATCH is used in SQLite to search against the FTS index - see [here](https://www.sqlitetutorial.net/sqlite-full-text-search/) -, so what am I missing?

BTW, the choice of "ContainsExpression" for FTS was not lightly taken: "contains" is used by more major players (e.g. SQL Server, Oracle, IBM Db2) than "match" is (e.g. PostgreSQL, MySQL, SQLite).

bahusoid: Can you clarify why we need this file in a first place. We don't need `AbstractCriterion` based classes for other functions. So how this `Contains` is different?

src/NHibernate/Criterion/ContainsExpression.cs
gliljas: Yes, it's certainly smelly there too :) But I'd suggest just try checking for whether a commonly named function is registered in the dialect or not.

src/NHibernate/Criterion/ContainsExpression.cs
gliljas: I'm just thinking it sounds quite a lot like a standard "string contains", when it's actually a "fulltext contains", and what that means differs between DBMS implementations.

gliljas: @bahusoid For exposing the function in the criteria API.

src/NHibernate/Criterion/ContainsExpression.cs
mihaicodrean: Thanks for the constructive discussion.

Checking for a function by name would mean searching for at least two possible flavors, "contains" & "match". For my purposes, I would also need to look for "freetext" for SQL Server. To me that sounds more fragile than type checking.

There's also a potential future clash in some RDBMs, e.g. [Oracle Text](https://docs.oracle.com/cd/B28359_01/text.111/b28303/query.htm#i1007233) has both "contains" & "matches" keywords.

src/NHibernate/Criterion/ContainsExpression.cs
mihaicodrean: @gliljas Good point, especially to programmers. How about "FullTextExpression" and Restrictions.FullText?

bahusoid: > For exposing the function in the criteria API.

And what's wrong with `Projections.SqlFunction`?

Anyway I think I got it. SQL Server `Contains` function can't be used in comparison expressions (`WHERE Contains(prop, 'xx') = 1`). And there is no way in Criteria to generate bare projection/function call in WHERE statement (`WHERE Contains(prop, 'xx')`)

Am I right?

I think it's better to implement class for bare bool expressions (`BoolExpression`?) which would be rendered simply as wrapped projection. So usage in code can look something like:
```C#
Restrictions.Where(
  Projections.Bool(
    Projections.SqlFunction(
      "contains",
      NHibernateUtil.Bool,
      Projections.Property("Name"),
      Projections.Constant("xx")))
```
And we can leave nice wrapping for user.

gliljas: Yes, something along those lines, but based on an adjective, to align with other restrictions, e.g FullTextContains

bahusoid: >  e.g FullTextContains

I see no such restriction. `LikeExpression` and `InsensitiveLikeExpression` is the closest I see. And I'm not mistaken they can also be simplified to `SQLFunctionTemplate` + `Projections.SqlFunction` call.

src/NHibernate/Criterion/ContainsExpression.cs
gliljas: The name doesn't have to be "contains" or "match" (see below discussion about the ambiguity). Sure, it has an element of fragility, but type checking is completely non extensible. Consider if you'd like to add this support in a different dialect. Having to modify ContainsExpression to do that would stink. I would hope that at some point, "NHibernate.Core" doesn't even know what MsSql2000Dialect is. It just knows about dialects or whatever similar construct may take its place in the future. 

As I mentioned, this kind of smell is not uncommon, but I'd rather add big TODO:s to fixing them, rather than adding more.

src/NHibernate/Criterion/ContainsExpression.cs
bahusoid: Ohh :))) I see.  It wasn't for me about `FullTextContains` :) So just ignore my last message ( I deleted it)

mihaicodrean: @bahusoid Please keep in mind that the FTS functions (e.g. contains, match) have no return value.

src/NHibernate/Criterion/ContainsExpression.cs
mihaicodrean: Of course it doesn't *have* to be "contains" or "match", but the name still has to be familiar to someone coming from the DB world. Also, I'm sure the major players put some serious though into it as well, and the more I think about it, the more I find "contains" to best fit the job, in that it hints about tokenization. However, an ORM connects two worlds and it's addressing / facing programmers, where indeed Contains from String becomes debatable.

With all that in mind, I think your final proposal below with "FullTextContains" fits best. Good idea. I'll rename after agreeing on the rest below.

For the name checking vs. my question above ("Add a virtual property in the Dialect base class for the function name?"), how about meeting in the middle, by defining an abstract Dialect.FullTextFunction string property, then have the actual dialects implement it? Or, to not have all the custom dialects out there complain during compilation after upgrade, the property could be non-abstract and equal to an empty string by default or null, overridden by the dialects that care about it. And then, the FullTextContainsExpression will check it for !IsEmptyOrNull.

mihaicodrean: Done both & committed, please re-review.

hazzik:
I would prefer more generic solution which allows for a sql-function restriction. Something like `Restrictions.SqlFunction("xxx", "a", "b", "c")`.

hazzik:
Closing in favor of  #2846

PR #2576, opened 2020-10-01
Fix caching of OneToOneType from second level cache.

### LINKED ISSUE #2552: One-to-one second level cache issue
When assembling an object with one-to-one relationships, a second level cache miss occurs while trying to assemble the related object. The OneToOneType has the following code:
```c#
public override object Assemble(object cached, ISessionImplementor session, object owner)
{
	//this should be a call to resolve(), not resolveIdentifier(), 
	//'cos it might be a property-ref, and we did not cache the
	//referenced value
	return ResolveIdentifier(session.GetContextEntityIdentifier(owner), session, owner);
}
```
I'm not sure why the implementation of this function is ignoring the value of `cached`, but feel this should be changed to:
```c#
public override object Assemble(object cached, ISessionImplementor session, object owner)
{
	return ResolveIdentifier(cached, session, owner);
}
```
or:
```c#
public override object Assemble(object cached, ISessionImplementor session, object owner)
{
	if (cached != null) {
		return ResolveIdentifier(session.GetContextEntityIdentifier(owner), session, owner);
	}

	return null;
}
```

Currently when `cached` is null, the code presumes that the Id of the one-to-one object is the same as the ID of the owning object. This causes a cache miss and a subsequent query to verify that no related object exists. I'm not sure why the original code does not simply use the `cached` Id here..

Hibernate ORM has this same issue, so coordinating a fix in both would benefit all.

deAtog:
The changes here are based on the current implementation in the ManyToOneType. These changes fix issue #2552 by storing the ID of the related one-to-one object when disassembling the parent/owner. When assembling the parent/owner the ID of the one-to-one is no longer assumed to be that of the owner and when it is null, it is reliably so. This eliminates the cache-miss and subsequent database query when the related one-to-one object was null.

Fixes #2552


hazzik:
Need to have tests.

bahusoid:
Maybe we should keep old logic when it's safe to use owner id (when `IsReferenceToPrimaryKey && !IsNullable`)...

deAtog:
> Maybe we should keep old logic when it's safe to use owner id (when `IsReferenceToPrimaryKey && !IsNullable`)...

I think adding the check for that particular case would only add additional unnecessary complexity.

bahusoid (approved):
So basically `ManyToOneType` handling is copied to `OneToOneType`. It seems we need to introduce base type `ToOneType` with common functionality (no need to do it in this PR)

hazzik:
> It seems we need to introduce base type ToOneType with common functionality (no need to do it in this PR)

`EntitiyType` is that `ToOneType`

deAtog:
This seems to cause a regression when committing a transaction that adds a related one-to-one object to an existing entity. Nhibernate does not currently have a test case for this, but Hibernate ORM does. I'll try to resolve this issue next week and add the related test.

bahusoid (commented):
Hibernate Issue: https://hibernate.atlassian.net/browse/HB-992
> Trouble is that no SQL UPDATE is needed, so Hibernate does not consider the owning object to be dirty (at least as far as the db is concerned). However, it /is/ dirty as far as the cache is concerned.

deAtog:
I added the basic one-to-one persistence test that was failing in Hibernate ORM. This test fails after changing the disassemble and assemble methods, but passes once the isDirty method is implemented. The isDirty check causes an update to the owner to occur, but since the one-to-one property is not updatable, it only hits the cache and not the database. In the case that was failing, this causes the null id of the one-to-one object to be updated with the correct id in the cache.

bahusoid:
> but since the one-to-one property is not updatable, it only hits the cache and not the database

Are you sure? I don't think it works this way. If object considered is dirty NHibernate will trigger update for all properties unless dynamic-update is specified. So I believe now unnecessary update statement is generated. And I'm not even sure that dynamic-update case is handled properly big chance that invalid SQL is generated for this case.

deAtog:
@bahusoid Yes, I am 100% certain of this. An Update action is added to the ActionQueue with the only change being the ID of the related one-to-one property. Since that value is not undateable on the owner, it is not considered as part of the database update. Since nothing else is modified in this case, the database update is skipped but the change is still persisted in the cache.

deAtog:
To be more specific, the flow is as follows:
1. An [EntityUpdateAction ](https://github.com/nhibernate/nhibernate-core/blob/6cb478fd7644507941140dd05b33e2fa9f66702e/src/NHibernate/Action/EntityUpdateAction.cs )is added to the [ActionQueue ](https://github.com/nhibernate/nhibernate-core/blob/2ad61c6f3b935dadfb6b5c7b885744607ae2f7f8/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs) for the owner.
2. All actions in the queue are executed via a call to [ActionQueue.ExecuteActions()](https://github.com/nhibernate/nhibernate-core/blob/6cb478fd7644507941140dd05b33e2fa9f66702e/src/NHibernate/Engine/ActionQueue.cs#L244) which calls [EntityUpdateAction.Execute()](https://github.com/nhibernate/nhibernate-core/blob/6cb478fd7644507941140dd05b33e2fa9f66702e/src/NHibernate/Action/EntityUpdateAction.cs#L44).
3. EntityUpdateAction.Execute() obtains the entity persister which in this case is a [SingleTableEntityPersister ](https://github.com/nhibernate/nhibernate-core/blob/6cb478fd7644507941140dd05b33e2fa9f66702e/src/NHibernate/Persister/Entity/SingleTableEntityPersister.cs) which is derived from the [AbstractEntityPersister](https://github.com/nhibernate/nhibernate-core/blob/6cb478fd7644507941140dd05b33e2fa9f66702e/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs).
4. Once the persister is obtained Execute() calls [persister.Update()](https://github.com/nhibernate/nhibernate-core/blob/6cb478fd7644507941140dd05b33e2fa9f66702e/src/NHibernate/Action/EntityUpdateAction.cs#L79) with the list of dirty fields. (In this case only the OneToOneType field is dirty.)
5. [AbstractEntityPersister.Update()](https://github.com/nhibernate/nhibernate-core/blob/2ad61c6f3b935dadfb6b5c7b885744607ae2f7f8/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs#L3409) calls [GetTableUpdateNeeded](https://github.com/nhibernate/nhibernate-core/blob/2ad61c6f3b935dadfb6b5c7b885744607ae2f7f8/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs#L1218) which iterates over the list of dirty fields and marks the tables that need updated based on which fields are dirty **and** updatable.
6. GetTableUpdateNeeded returns false for all tables as the OneToOneType field is dirty but **not** updatable.
7. [AbstractEntityPersister.Update()](https://github.com/nhibernate/nhibernate-core/blob/2ad61c6f3b935dadfb6b5c7b885744607ae2f7f8/src/NHibernate/Persister/Entity/AbstractEntityPersister.cs#L3465) iterates over the list of tables requiring updates and performs the update if needed. (No updates are needed and the function returns.)
8. Finally, EntityUpdateAction.Execute() [updates the cache](https://github.com/nhibernate/nhibernate-core/blob/6cb478fd7644507941140dd05b33e2fa9f66702e/src/NHibernate/Action/EntityUpdateAction.cs#L120) with the current state of the entity.

deAtog:
@hazzik In my opinion, this is ready to be integrated after review. These changes mirror the ones I've proposed for Hibernate

src/NHibernate/Type/OneToOneType.cs
```diff
@@ -153,7 +189,7 @@ public override object Assemble(object cached, ISessionImplementor session, obje
 		/// </summary>
 		public override bool IsAlwaysDirtyChecked
 		{
-			get { return false; } //TODO: this is kinda inconsistent with CollectionType
+			get { return true; } //TODO: this is kinda inconsistent with CollectionType
```
hazzik, line 192: This comment is probably outdated now.

src/NHibernate.Test/OneToOneType/Owner.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace NHibernate.Test.OneToOneType
+{
+	public class Owner
+	{
+		private Details _details;
+
+		public virtual int Id { get; protected set; }
+		public string Name { get; set; }
+		public virtual Details Details {
+			get => _details;
```
hazzik, line 16: `get { return _details;}` please

src/NHibernate.Test/OneToOneType/Owner.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace NHibernate.Test.OneToOneType
+{
+	public class Owner
+	{
+		private Details _details;
+
+		public virtual int Id { get; protected set; }
+		public string Name { get; set; }
+		public virtual Details Details {
```
hazzik, line 15: Code format.

src/NHibernate.Test/NHSpecificTest/GH2552/Fixture.cs
```diff
@@ -0,0 +1,178 @@
...
+					ids.Add(s.Save(person));
+				}
+
+				tx.Commit();
+			}
+
+			return ids;
+		}
+
+		public IList<TPerson> getPersons<TPerson>(List<object> ids) where TPerson : Person
```
hazzik, line 148: Please use C# naming convention.

src/NHibernate/Type/OneToOneType.cs
```diff
@@ -59,12 +59,31 @@ public override bool IsOneToOne
 
 		public override bool IsDirty(object old, object current, ISessionImplementor session)
```
hazzik, line 60: Because this code is not identical for `OneToOneType` and `ManyToOneType` it would be better to move it to `EntityType` (the base of both types). 

For discussion: 
- do we want to implement `IsDirty(object old, object current, ISessionImplementor session)` in `EntityType` or;
- create a new method `IsDirtyCore...` and call it from overrides? /cc @bahusoid @fredericDelaporte @maca88

src/NHibernate/Type/OneToOneType.cs
```diff
@@ -135,15 +154,32 @@ public override bool UseLHSPrimaryKey
 
 		public override object Disassemble(object value, ISessionImplementor session, object owner)
```
hazzik, line 155: Same as with `IsDirty`

src/NHibernate/Type/OneToOneType.cs
```diff
@@ -135,15 +154,32 @@ public override bool UseLHSPrimaryKey
...
+
+			if (id == null)
+			{
+				throw new AssertionFailure("cannot cache a reference to an object with a null id: " + GetAssociatedEntityName());
+			}
+
+			return GetIdentifierType(session).Disassemble(id, session, owner);
 		}
 
 		public override object Assemble(object cached, ISessionImplementor session, object owner)
```
hazzik, line 172: Same as with `IsDirty`

hazzik (commented):
There are some comments for consideration and suggestions to code.

src/NHibernate/Type/OneToOneType.cs
deAtog: The implementation of IsDirty in this case is identical to the implementation of the ManyToOneType. Maybe the common functionality can be combined in common ToOneType in a future enhancement. The purpose of this change set is only to correct the above mentioned issue.

bahusoid: @hazzik  Either way works for me. Simply moving IsDirty to base looks as less code...

It seems `IsModified` should also be modified accordingly for `OneToOneType`. But have no idea when it's really needed...

hazzik:
@deAtog it seems you've forgot to push the fix with your latest force-push

deAtog:
@hazzik Sorry about that I got busy working on other things for work. I didn't push everything at once because I wanted the build to complete before pushing everything else.

src/NHibernate/Type/OneToOneType.cs
deAtog: @bahusoid `IsModified` is inherited from `AbstractType` which returns true/false based on `IsDirty`.

src/NHibernate/Type/OneToOneType.cs
deAtog: Removed comment in latest update.

src/NHibernate.Test/OneToOneType/Owner.cs
deAtog: Changed.

src/NHibernate.Test/OneToOneType/Owner.cs
deAtog: Fixed.

src/NHibernate.Test/NHSpecificTest/GH2552/Fixture.cs
deAtog: Fixed.

src/NHibernate/Type/OneToOneType.cs
hazzik: I made a typo. It should be `code is now identical` not `code is no*t* identical`. Fixed.

bahusoid: > IsModified is inherited from AbstractType which returns true/false based on IsDirty

And still for some reasons `ManyToOne` overrides it:
https://github.com/nhibernate/nhibernate-core/blob/18667cfbdbbb0ee7743be4e1725583052d38d007/src/NHibernate/Type/ManyToOneType.cs#L137-L151

So if it's required here it seems it's now required for OneToOne too

deAtog: I think the difference here might be due to the fact the owner of a many-to-one can be modified whereas the one-to-one cannot.

deAtog: I agree there _might_ be an issue here, but it is unrelated to the one that this pull request attempts to fix. I.E. It looks like `IsModified` _may_ not handle the case when `old` is an instance of a OneToOneType and not an identifier. If you can verify this is the case, then I'd suggest creating a separate issue to correct this. After all, this problem would exist despite these changes.

bahusoid: > IsModified is inherited from AbstractType which returns true/false based on IsDirty.

Actually this statement is wrong. `IsModified` is overridden in `OneToOneType` and always returns false:
https://github.com/nhibernate/nhibernate-core/blob/eb137cc8668982ae0c0f68bd264e834c138324f8/src/NHibernate/Type/OneToOneType.cs#L70-L73

bahusoid: > IsModified may not handle the case when old is an instance of a OneToOneType and not an identifier

I don't get why you think it's unrelated... It's related.
Old logic: handles both identifier and entity type in old parameter (because it simply returns null)
If you make IsModified call IsDirty - your new logic will fail when `IsModified` is called with identifier. Because now `GetIdentifier` might be called on `old` instance.

deAtog: You are correct, `IsModified` always returns `false`. This is consistent with the original behavior of the type that you cannot modify owner of the owner of the OneToOneType once assigned. I'm not against changing this behavior but I think that doing so would be the topic of another issue. `IsDirty` determines whether or not the parent should be considered dirty or not, while `IsModified` determines whether or not an update is required to the identifier of the OneToOneType. Since updates are not currently permitted, `IsModified` always returns `false`.

bahusoid: Maybe I just don't understand something. But from what I see [DefaultFlushEntityEventListener.DirtyCheck](https://github.com/nhibernate/nhibernate-core/blob/6ab71d08b6c7f0a9c95895e94700b466b0743d6e/src/NHibernate/Event/Default/DefaultFlushEntityEventListener.cs#L426-L496) can call both methods for obtaining dirty properties depending on current session state:
https://github.com/nhibernate/nhibernate-core/blob/6ab71d08b6c7f0a9c95895e94700b466b0743d6e/src/NHibernate/Event/Default/DefaultFlushEntityEventListener.cs#L452

https://github.com/nhibernate/nhibernate-core/blob/6ab71d08b6c7f0a9c95895e94700b466b0743d6e/src/NHibernate/Event/Default/DefaultFlushEntityEventListener.cs#L480

With only difference that `IsModified` is called on databaseSnapshot (and identifier related code was added to many-to-one to handle such db snapshots). So to me it looks like those method should behave the same as both of them can be called for dirty check.

So regarding
>IsDirty determines whether or not the parent should be considered dirty or not, while IsModified determines whether or not an update is required to the identifier of the OneToOneType

Can you clarify where did you get this difference and how it plays with `DefaultFlushEntityEventListener.DirtyCheck`? And I don't really understand what `determines whether or not an update is required to the identifier of the OneToOneType` means considering identifier of the entity is not updateable. And if you mean in parent object - it's still can be changed from null to something and vice-versa.

src/NHibernate.Test/OneToOneType/Fixture.cs
```diff
@@ -0,0 +1,61 @@
...
+			{
+				s.CreateQuery("delete from Details").ExecuteUpdate();
+				s.CreateQuery("delete from Owner").ExecuteUpdate();
+
+				tx.Commit();
+			}
+		}
+
+		[Test]
+		public void OneToOnePersistedOnOwnerUpdate()
```
bahusoid, line 22: Can you also add test that checks that cache is properly reset when `owner.Details`  is changed from not-null to null.

src/NHibernate/Type/OneToOneType.cs
deAtog: `IsDirty` has the following comment:
```C#
/// <summary>
/// Should the parent be considered dirty, given both the old and current 
/// field or element value?
/// </summary>
```

`IsModified` has the following comment
```C#
/// <summary>
/// Says whether the value has been modified
/// </summary>
```
Pseudo code for `DefaultFlushEntityEventListener.DirtyCheck` the following:

L428-434 Get entity, current property values, loaded property values, id, and persiseter.
L436 Check if the Interceptor handled the Dirty check.
L443 If the Interceptor did not handle the dirty check, then attempt to dirty check the entity.
L448 If the loaded state of the entity is null, then the entity is transient and cannot be dirty checked.
L449-453 If the entity was loaded then compare current property values against loaded property values to determine if the entity is dirty.
L454 If the entity was deleted, compare the current property values against the deleted state.
L474-484 Get the current values from the database if select before update is required or check the id against the unsaved value. If a database entry exists or the id is not equal to the unsaved value, check for modified property values.

Based on the above and the fact that the OneToOneType is not updatable, I see no reason why `IsModified` should ever return anything but `false` for the OneToOneType.

bahusoid: > L448 If the loaded state of the entity is null, then the entity is transient and cannot be dirty checked.

`loadedState` can also be null if entity is added to session via via `session.Update`. It actually has explicit comment about it... So I expect with current PR state we still have issue with cache when owner is added to session via  `session.Update` and then updated. Can you prove me wrong with test case?

src/NHibernate/Type/OneToOneType.cs
```diff
@@ -135,25 +154,41 @@ public override bool UseLHSPrimaryKey
...
-		/// We don't need to dirty check one-to-one because of how 
-		/// assemble/disassemble is implemented and because a one-to-one 
-		/// association is never dirty
+		/// We always need to dirty check as our identifier is dependent on
+		/// whether or not a persistent entry exists.
 		/// </summary>
 		public override bool IsAlwaysDirtyChecked
 		{
-			get { return false; } //TODO: this is kinda inconsistent with CollectionType
+			get { return true; }
```
bahusoid, line 191: For many-to-one it's true only for `not-found="ignore"`. Why it's always true here? Shouldn't it be here similarly true only for `constrained="false"` associations (so basically `return IsNullable;`)

deAtog: There are two use cases for a one-to-one with constrained="false" on the parent side and constrained="true" on the child side. From my understanding of `IsAlwaysDirtyChecked` only determines whether or not the relations should be checked to see if it is dirty. When constrained="false" we definitely want to always perform the dirty check. When constrained="true", the id of the child should match the id of the parent and so presumably it can never be dirty. I think returning `IsNullable` here would be acceptable.

bahusoid (changes requested):
So to finalize this PR the following needs to be done:
1) `IsModified` needs the same handling as in many-to-one (see `OneToOnePersistedOnOwnerUpdateForSessionUpdate`);
2) `IsAlwaysDirtyChecked` should also be [adjusted](https://github.com/nhibernate/nhibernate-core/pull/2576/files#r523748008);
3) Move common logic to `EntityType`. I'm actually fine to make this refactoring as separate task later.. So I see it as optional point

bahusoid (dismissed):

hazzik (approved):

src/NHibernate.Test/NHSpecificTest/GH2552/Fixture.cs
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		private int DeleteDetailsFromFirstPerson<TPerson>() where TPerson:Person
+		{
+			using (ISession s = Sfi.OpenSession())
+			using (ITransaction tx = s.BeginTransaction())
+			{
+				// Get the first person with details.
+				Person person = s.QueryOver<TPerson>()
+					.Where(p => p.Details != null)
```
bahusoid, line 125: It will generate something like `select x.* from Obj x where x.Id != null`. Not what you expect to get.
Unfortunately such queries in QueryOver are not working properly. Left Join on `Details` is required to make it work properly but implicit joins are not supported in `QueryOver` . So you have either to use Linq Query here or add join explicitly.

src/NHibernate.Test/NHSpecificTest/GH2552/Fixture.cs
```diff
@@ -0,0 +1,214 @@
...
+					.Take(1)
+					.SingleOrDefault();
+
+				Assert.NotNull(person);
+				Assert.NotNull(person.Details);
+
+				s.SaveOrUpdate(person);
```
bahusoid, line 132: `person` is already loaded to session. There is no need to call `SaveOrUpdate` here (this method should only be used when you want to update object loaded in different session)

bahusoid (dismissed):

src/NHibernate.Test/NHSpecificTest/GH2552/Fixture.cs
deAtog: Ah okay. I wasn't really aware of the implicit nature of sessions and transactions until quite recently. I guess I should have realized by now that committing the transaction was all that was needed. Thanks for correcting the tests.

src/NHibernate.Test/NHSpecificTest/GH2552/Fixture.cs
deAtog: Yeah, I thought this might have been the case. Thanks again for correcting the test.

bahusoid (approved):

fredericDelaporte:
Possible breaking change: now one-to-one changes trigger a version increment. #3216 is required for reverting to the old behavior.

PR #2361, opened 2020-04-23
Table group joins support in hql

bahusoid:
Loose port of https://github.com/hibernate/hibernate-orm/commit/753858ce735b2a4eb1a88e3dcd676a0f4e444c2b

Example of SQL generated with "group joins":
```sql
SELECT *
FROM Person person0_
INNER JOIN (
	IndividualCustomer individual1_ 
	INNER JOIN Customer individual1_1_ ON individual1_.IndividualCustomerID = individual1_1_.Id
	) ON person0_.Id = individual1_.PersonID AND individual1_1_.Deleted = @p0
```

It's useful when withClause needs to be applied to multiple tables:
* subclasses with withClause referencing base tables
* Complex withClause with implied joins (like `from a in A join b in B b on a.Id equals b.C.D.Id` also see added test case `WithImpliedJoinOnAssociation`) 
* Many-to-many with filters/ withClause (makes fix implemented in https://github.com/nhibernate/nhibernate-core/pull/1995 obsolete as with group joins no null data is retrieved)

Fixes hql/Linq for #1209 and #1215

PR #2581, opened 2020-10-11
Fix joining with an inherited property from another table for Linq provider

### LINKED ISSUE #2580: InvalidWithClauseException when join polymorphic entity
### Situation
Updating NHibernate from 5.2.7 to 5.3.3 we ran into issues with queries operating on inheritance hierarchies. We join inherited entities and run operators such as distincts or simple selects. Analyzing the exception `with clause can only reference columns in the driving table` we soon came to find the cause being the `HqlSqlWalker` and its `WithClauseVisitor`. There the names of aliases are compared and if different, an exception is thrown.

#### Comparison with Hibernate
After some research we found this to be an issue in Hibernate as well: https://hibernate.atlassian.net/browse/HHH-11383. It was fixed here: https://github.com/hibernate/hibernate-orm/commit/5ac46eb70359a45a34f7a64e859df106928baa94. 

We use NHibernate as our ORM for large-scale projects and ensure compability with a test coverage of 20K+ tests (hence we discovered this issue quickly). For testing purposes we disabled the exception (as in Hibernate) and ran our queries, which then work smoothly. 

### Request
Is it possible to adapt the change from Hibernate to NHibernate and disable the check of alias names in this case and consequently not throwing the exception?

On a side note: perhaps this became a problem in the first place after NHibernate release 5.3.0 and this change https://github.com/nhibernate/nhibernate-core/pull/2078

maca88:
Fix for a regression made by #2327. When merging into master only the test cases should be merged as #2361 makes this fix obsolete.

Fixes #2580

### DIFF
--- a/src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
+++ b/src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
@@ -535,7 +535,7 @@ namespace NHibernate.Linq.Visitors
 			// When associations or members from another table are located inside the inner key selector we have to use a cross join
 			// instead of an inner join and add the condition in the where statement.
 			if (queryModel.BodyClauses.OfType<NhJoinClause>().Any(o => o.ParentJoinClause == joinClause) ||
-			    queryModel.BodyClauses.OfType<JoinClause>().Any(o => baseMemberCheker.ContainsBaseMember(o.InnerKeySelector)))
+			    queryModel.BodyClauses.OfType<JoinClause>().Any(baseMemberCheker.ContainsBaseMember))
 			{
 				if (!innerJoin)
 				{
@@ -566,10 +566,18 @@ namespace NHibernate.Linq.Visitors
 				_sessionFactory = sessionFactory;
 			}
 
-			public bool ContainsBaseMember(Expression node)
+			public bool ContainsBaseMember(JoinClause joinClause)
 			{
+				// Visit the join inner key only for entities that have subclasses
+				if (joinClause.InnerSequence is ConstantExpression constantNode &&
+				    constantNode.Value is IEntityNameProvider entityNameProvider &&
+				    !_sessionFactory.GetEntityPersister(entityNameProvider.EntityName).EntityMetamodel.HasSubclasses)
+				{
+					return false;
+				}
+
 				_result = false;
-				Visit(node);
+				Visit(joinClause.InnerKeySelector);
 
 				return _result;
 			}
--- a/src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
+++ b/src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using NHibernate.Engine;
 using NHibernate.Hql.Ast;
 using NHibernate.Linq.Clauses;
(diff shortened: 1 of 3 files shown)


src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -527,10 +529,13 @@ private void AddJoin(JoinClause joinClause, QueryModel queryModel, bool innerJoi
 			var withClause = equalityVisitor.Visit(joinClause.InnerKeySelector, joinClause.OuterKeySelector);
 			var alias = _hqlTree.TreeBuilder.Alias(VisitorParameters.QuerySourceNamer.GetName(joinClause));
 			var joinExpression = HqlGeneratorExpressionVisitor.Visit(joinClause.InnerSequence, VisitorParameters);
+			var baseMemberCheker = new BaseMemberChecker(VisitorParameters.SessionFactory);
+
 			HqlTreeNode join;
-			// When associations are located inside the inner key selector we have to use a cross join instead of an inner
```
bahusoid, line 531: A bit offtopic: Is this check still required in master? Or it also should be handled by #2361 and can be removed?

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -551,6 +556,46 @@ private void AddJoin(JoinClause joinClause, QueryModel queryModel, bool innerJoi
...
+			protected override Expression VisitMember(MemberExpression node)
+			{
+				if (ExpressionsHelper.TryGetMappedType(
+						_sessionFactory,
+						node,
+						out _,
+						out var persister,
+						out _,
+						out var propertyPath) &&
+					persister is IOuterJoinLoadable joinLoadable &&
```
bahusoid, line 586: Maybe skip property checks if no subclasses? Something like
```C#
persister.EntityMetamodel.HasSubclasses &&
persister is IOuterJoinLoadable joinLoadable ...
```

Or skip this visitor completely if `joinClause` entity has no subclasses?

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
maca88: No, it was obsolete by #2539. We should remove the `ParentJoinClause` property on the master branch as now is never set.

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
maca88: > Or skip this visitor completely if joinClause entity has no subclasses?

Added a check before visiting the node.

src/NHibernate/Linq/Visitors/QueryModelVisitor.cs
```diff
@@ -551,6 +556,54 @@ private void AddJoin(JoinClause joinClause, QueryModel queryModel, bool innerJoi
...
+
+			public BaseMemberChecker(ISessionFactoryImplementor sessionFactory)
+			{
+				_sessionFactory = sessionFactory;
+			}
+
+			public bool ContainsBaseMember(JoinClause joinClause)
+			{
+				// Visit the join inner key only for entities that have subclasses
+				if (joinClause.InnerSequence is ConstantExpression constantNode &&
```
maca88, line 572: Constant check seems to be enough as `join o2 in db.Mammals.Where(o => o.Pregnant)` is not supported.

bahusoid (approved):

Issue #2591, opened 2020-10-31
Add method AsNoTracking to prevent query results to be stored in first level cache


rjperes:
The idea is to have a method like EF Core's AsNoTracking (https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.entityframeworkqueryableextensions.asnotracking?view=efcore-3.1). When applied to an IQueryable<T>, it will prevent the results of the query to be stored in the first level cache of the originating session, by replacing it with an IStatelessSession.
Merge request included here: https://github.com/nhibernate/nhibernate-core/pull/2592.

hazzik:
The consensus seems that this functionality should be implemented in some "contrib" or "extensions" project.

PR #2592, opened 2020-10-31
#2591 Add method AsNoTracking to prevent query results to be stored in first level cache

rjperes:
https://github.com/nhibernate/nhibernate-core/issues/2591

### DIFF
--- a/src/NHibernate/Linq/LinqExtensionMethods.cs
+++ b/src/NHibernate/Linq/LinqExtensionMethods.cs
@@ -2614,6 +2614,28 @@ namespace NHibernate.Linq
 			throw new InvalidOperationException("The method should be used inside Linq to indicate a type of a parameter");
 		}
 
+		public static IQueryable<T> AsNoTracking<T>(this IQueryable<T> query)
+		{
+			if (query is NhQueryable<T> nhQuery)
+			{
+				var oldProvider = nhQuery.Provider as DefaultQueryProvider;
+
+				if (oldProvider.Session is IStatelessSession)
+				{
+					return query;
+				}
+
+				var statelessSession = (oldProvider.Session as SessionImpl)
+					.StatelessSessionWithOptions()
+					.Connection(oldProvider.Session.Connection)
+					.OpenStatelessSession() as ISessionImplementor;
+				var newProvider = new DefaultQueryProvider(statelessSession, null);
+				return new NhQueryable<T>(newProvider, query.Expression);
+			}
+
+			return query;
+		}
+
 		internal static INhQueryProvider GetNhProvider<TSource>(this IQueryable<TSource> source)
 		{
 			if (source == null)
--- a/src/NHibernate.Test/Linq/QueryNoTrackingTests.cs
+++ b/src/NHibernate.Test/Linq/QueryNoTrackingTests.cs
@@ -20,5 +20,20 @@ namespace NHibernate.Test.Linq
 			Assert.True(customers.All(c => !session.Contains(c)));
 			Assert.AreEqual(0, this.session.GetSessionImplementation().PersistenceContext.EntityEntries.Count);
 		}
+
+		[Test]
+		public void FutureQueryIsNoTracking()
+		{
+			var customers = db.Customers
+				.Where(c => c.ContactTitle == "Sales Representative")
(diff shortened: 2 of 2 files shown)


src/NHibernate/Linq/LinqExtensionMethods.cs
```diff
@@ -2614,6 +2614,28 @@ public static T MappedAs<T>(this T parameter, IType type)
...
+
+				if (oldProvider.Session is IStatelessSession)
+				{
+					return query;
+				}
+
+				var statelessSession = (oldProvider.Session as SessionImpl)
+					.StatelessSessionWithOptions()
+					.Connection(oldProvider.Session.Connection)
+					.OpenStatelessSession() as ISessionImplementor;
```
bahusoid, line 2631: I believe all castings can be dropped in this code using something like this:
```suggestion
				var statelessSession = oldProvider.Session
					.StatelessSessionWithOptions()
					.Connection()
					.OpenStatelessSession()
					.GetSessionImplementation()
```

And it looks like it won't work properly for `Future` queries  (when both `AsNoTracking` and common queries are present in batch).

rjperes: It does work with ToFuture, included a unit test. As for your code, it does not compile.

bahusoid: > It does work with ToFuture,

Your test does not cover scenario I described - when both AsNoTracking and common queries are present in batch. Multiple future queries should be executed as one db roundtrip. I believe It won't happen with you changes (simply because `Future` queries are batched in session batcher - and your implementation just creates new session for each no tracking query)

> it does not compile. 

Missing semicolon or what? Yes I didn't compile it and give it as example -  I thought you would able to figure out what's missing.

bahusoid: Ok I see. `oldProvider.Session` returns implementor so one cast is still required: `(ISession) oldProvider.Session`

rjperes: Your remark about mixing AsNoTracking and common queries might be true, haven't tested it, but I don't see that as a problem. I see people using AsNoTracking when all they have is an ISession and they want, for example, to populate a list on a web page, that is, when they don't need the entities to be tracked. Anyway, that could always be mentioned in the documentation.

rjperes: Well, every extension method is a shortcut, so, in that sense, yes. But if you actually look at the code you can see that it's a bit more complex than that.

bahusoid (commented):
So to conclude our [discussion ](https://github.com/nhibernate/nhibernate-core/pull/2592/files#r515768435)- I don't like this implementation as it breaks batching for `Future` queries (`AsNoTracking` queries are simply not batched as executed in separate stateless sessions with own batchers).

User can achieve the same results with working `Future` queries and without hard castings made in this implementation and not limited to `IQueryable`:
```C#
var noTrackingSession = session.StatelessSessionWithOptions().Connection().OpenStatelessSession();
var resutls1 = noTrackingSession.Query<Entity>().ToFuture();
var results2 = noTrackingSession.QueryOver<Entity>().Future();
```

hazzik:
I tend to agree with @bahusoid.

I don't see a point why we need to "port" this feature from a EF. NHibernate already has a mechanism to run stateles (or "untracked") queries. 

If the underlying reason is another attempt to be "persistent ignorant" than I think it is wrong, because "persistence ignorance" is a big fat lie.

fredericDelaporte (commented):
> Anyway, it's purpose is not to be used with future queries.

The purpose you have in mind, maybe, but I do not see anything that would prevent thinking about using it with Future too. And then we would have a bug to fix.

Anyway I think NHibernate should not provide this feature as it already has a robust pattern for it and users can easily implement themselves, with very little code, this `AsNoTracking`. 

Maybe contributing your extension snippet as a "show and tell" [discussion](https://github.com/nhibernate/nhibernate-core/discussions) would be more suitable.

hazzik:
The consensus seems that this functionality should be implemented in some "contrib" or "extensions" project.

PR #2605, opened 2020-11-13
Upgrade AsyncGenerator to 0.19.1

maca88:
With this upgrade we can now generate async code also with .NET Core 3.1 and 5.0. Based on the installed .NET Core version, the appropriate async generator binaries will be used, so the `global.json` won't be needed anymore.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "csharpasyncgenerator.tool": {
+      "version": "0.19.1",
+      "commands": [
+        "async-generator"
+      ]
+    }
+  }
+}
--- a/ShowBuildMenu.sh
+++ b/ShowBuildMenu.sh
@@ -178,34 +178,13 @@ testRun(){
 }
 
 generateAsync(){
-	dotnet restore ./src/NHibernate.sln
-
-	getAsyncGeneratorPath
 	cd src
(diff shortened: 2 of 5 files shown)


hazzik (dismissed):

hazzik:
I'll redo it to use a tool instead.

hazzik:
@maca88 or you want to do it?

maca88:
> I'll redo it to use a tool instead.

Sure, why not. Ideally would be nice to port all tools but most likely there are some that are not published as a .NET tool.

>@maca88 or you want to do it?

I am fine, you can proceed with the conversion.

hazzik:
> Ideally would be nice to port all tools but most likely there are some that are not published as a .NET tool.

Yes, it would be great. However, only AsyncGenerator and GitReleaseManager has tool packages.

hazzik (approved):

fredericDelaporte:
Has the Wip removal been overlooked, or is there still something more to do?

maca88:
I've added WIP for upgrading from 0.19.0 to 0.19.1, which has been done. Removed it, as I think it is ready to be merged.

hazzik:
With this change, I think, the SDK 3+ would be required.

fredericDelaporte (approved):
SDK 3.0 is [indeed required](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-tool-restore), and keeping a global.json asking for a lower sdk causes the async generation to fail.

PR #2609, opened 2020-11-22
Fix #2608 - persisting a one-to-one with delayed insert fails

### LINKED ISSUE #2608: Delay entity insert may fail with Merge
We are having problems upgrading to new version 5.3.0 due to a breaking change on delay entity insert on persist: #1754

I read the discusson and agreed that Persist should delay entity insert, but Save does it immediatly.

Is it correct that Merge methos doesn't create identity immediatly?
I think that Merge method should work like Save.

For this reason, if I create a OneToOne relation between two objects with the same primary key:
- Merge gives "Specific cast is not valid"
- Save works correctly

Example:

```
public TableAMap()
        {
            Schema("[dbo]");
            Table("TEST_TABLE_A");
            Not.LazyLoad();

            Id(x => x.Id).Column("ID");

            Map(x => x.Name).Column("NAME").Length(100).Not.Nullable();
            Map(x => x.Description).Column("DESCRIPTION").Length(1000).Nullable();
            Map(x => x.CreationDate).Column("CREATION_DATE").Not.Nullable();

            HasOne(x => x.TableB).Not.LazyLoad().Cascade.All();
        }
```

```
public TableBMap()
        {
            Schema("[dbo]");
            Table("TEST_TABLE_B");
            Not.LazyLoad();

            Id(x => x.Id).Column("ID").GeneratedBy.Foreign("TableA");
            Map(x => x.Color).Column("COLOR").Length(100).Nullable();

            HasOne(x => x.TableA).Constrained();
        }
```

```
using (var session = sessionFactory.OpenSession())
            {
                {
                    var tableA = new TableA {Name = "test", Description = "it works?", CreationDate = DateTime.UtcNow};
                    tableA.TableB = new TableB { TableA = tableA, Color = "red" };

                    session.Save(tableA);
                    //session.Merge(tableA);

                    transaction.Commit();

                    var test = session.Query<TableA>().FirstOrDefault(x => x.Id == tableA.Id);
                }
            }
```

fredericDelaporte:
Fix #2608.


src/NHibernate.Test/Operations/OneToOne.hbm.xml
```diff
@@ -32,10 +32,12 @@
 
 	<class name="PersonalDetails" table="OPS_PERS_DETAIL">
 		<id name="Id" column="ID" type="long">
-			<generator class="increment"/>
+			<generator class="foreign">
+				<param name="property">Person</param>
+			</generator>
```
fredericDelaporte, line 37: We already had a test really near what is required for reproducing #2608, unfortunately ignored and anyway, not using an id strategy on its parent which would be adequate for triggering the bug. As switching the parent id strategy would impact other tests, I have just fix that other test.
Strictly speaking it should go in a dedicated PR, but let it be here.

src/NHibernate.Test/NHSpecificTest/GH2608/Fixture.cs
```diff
@@ -0,0 +1,33 @@
...
+
+		[Test]
+		public void MergeBidiPrimaryKeyOneToOne()
+		{
+			using (var s = OpenSession())
+			using (var tx = s.BeginTransaction())
+			{
+				var p = new Person { Name = "steve" };
+				p.Details = new PersonalDetails { SomePersonalDetail = "I have big feet", Person = p };
+				s.Merge(p);
```
bahusoid, line 28: It looks like the same exception should be thrown on `Persist` too. Am I right?

fredericDelaporte: No, persist does not fail, it least with my tests on a modified MergeFixture.

bahusoid: Strange... I thought it's directly related to #1754 in which case Persist should fail the same (at least for this test case). I added test for Persist - let's see...

bahusoid: >  it least with my tests on a modified MergeFixture.

But in current state it also doesn't fail on Merge.... The only failed tests are newly added GH2608

fredericDelaporte: The `MergeFixture` one fails if we change its id generator from `increment` to `identity`. But as this would mean changing from a app-side generated id to a db-side one, I have not done that change, it would change too much the test.

I have re-checked these GH2608 tests on 5.2.x just in case, both succeed on 5.2.x.

fredericDelaporte:
About working on a fix, it will be maybe on next week-end, if not done by someone else before that.

bahusoid:
So it's regression after #1754.  Reason: Delayed foreign id is not supported (exception is thrown when `DelayedPostInsertIdentifier` is tried to be set for entity with foreign id generator). Doesn't look easy fixable (if we want to keep delayed insertion)

fredericDelaporte:
The fix is not hard after all, while keeping things delayed.
I have rebased this on 5.3.x, it was on master.

hazzik (approved):

PR #2612, opened 2020-11-23
WIP Fix nullable entity comparison with null

### LINKED ISSUE #2611: One-to-zero-or-one relation not returning data when checking for null
# Issue/Scenario

We are having troubles with some of our queries that contain null checks and don't return any data when the null check is in place. Our setup is as follows:

- database first (legacy database)
- one-to-zero-or-one relationship between two tables
- some queries contain null checks on the zero-or-one side and do not return data because of the null check

We are not sure if this is 

- expected behavior/by design (zero-to-one-side mapped with Id generated by foreign)
- wrong mapping on our side
- a bug

## Steps to reproduce

- Query data:
  - `.Query<Sheep>().Where(s => s.Name == "Dolly")` 
  - **Result: *** Sheep "Dolly" returned
- Add a null check to the query (zero-or-one-side):
  -  `.Query<Sheep>().Where(s => s.Name == "Dolly" && s.SlaughterInfo == null)` 
  - **Result: *** no data returned due to generated SQL statement containing ` (slaughterInfo.SheepId is null)` 

I have created a [sample project](https://github.com/marcelstuder/nhibernate-one-to-zero-or-one-issue) that demonstrates the issue.

Even if this is the expected behavior: from a developer's perspective who writes the Linq query it is not obvious that this null check is causing the query to fail without looking at the underlying mapping (or writing a test :-)).

## Notes

We are using the following Versions:

- NHibernate 5.3.5
- FluentNhibernate 3.1.0

## Workarounds

- check all queries in code and remove null checks if possible

bahusoid:
Apparently fix made in #2081 wasn't enough as implicit joins were generated for comparisons like `where entity.OneToOne is null`. It should be left join instead...

Fixes #2611

### DIFF
--- a/src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
@@ -409,9 +409,15 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 			}
 			else
 			{
-				comparisonWithNullableEntity = (Walker.IsComparativeExpressionClause && entityType.IsNullable);
-				joinIsNeeded = generateJoin || (Walker.IsInSelect && !Walker.IsInCase) || (Walker.IsInFrom && !Walker.IsComparativeExpressionClause)
-				               || comparisonWithNullableEntity;
+				if (Walker.IsComparativeExpressionClause && entityType.IsNullable)
+				{
+					joinIsNeeded = comparisonWithNullableEntity = true;
+					_joinType = JoinType.LeftOuterJoin;
+				}
+				else
+				{
+					joinIsNeeded = generateJoin || (Walker.IsInSelect && !Walker.IsInCase) || (Walker.IsInFrom && !Walker.IsComparativeExpressionClause);
+				}
 			}
 
 			if ( joinIsNeeded ) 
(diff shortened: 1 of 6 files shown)


src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
```diff
@@ -409,9 +410,19 @@ private void DereferenceEntity(EntityType entityType, bool implicitJoin, string
...
 			{
-				comparisonWithNullableEntity = (Walker.IsComparativeExpressionClause && entityType.IsNullable);
-				joinIsNeeded = generateJoin || (Walker.IsInSelect && !Walker.IsInCase) || (Walker.IsInFrom && !Walker.IsComparativeExpressionClause)
-				               || comparisonWithNullableEntity;
+				if (Walker.IsComparativeExpressionClause && entityType.IsNullable)
+				{
+					joinIsNeeded = comparisonWithNullableEntity = true;
+
+					//TODO: Fix this hack. Ideally we should always left join here. Skip left join for nullable entity if query contains implicit joins. We currently don't support left joins for such queries (see OneToOneCompositeQueryCompareWithJoinOrIsNull)
+					var fromJoins = ASTUtil.CollectChildren<FromElement>(Walker.CurrentFromClause, x => x is FromElement fe && !fe.IsImplied && fe.Type == HqlSqlWalker.FROM_FRAGMENT);
```
bahusoid, line 418: So hacky... :( Not even sure it handles all the cases..

fredericDelaporte:
Is this fixing a regression, or fixing an additional case of an old issue which was only partially fixed? If that is the later, it would be better to target master.

bahusoid:
It's fixing additional case... And I don't like this fix at all.. So will close it now. Need some thinking when I have time..

PR #2618, opened 2020-11-24
Fix limits in polymorphic queries 

### LINKED ISSUE #2614: Obvious bug in two HQLQueryPlan classes with distinction Set
We have a system still running with an old version of NHibernate and I've found a bug when experimenting with current release of NHibernate:

When the needsLimit variable is true in PerformList or PerformListAsync, the Limit is done in memory. To detect duplicates, there's an ISet<object> distinction. In current code, the items are skipped if they can be added to the distinction set but that means they have not yet been added to the combinedResults list. It must be the other way round.

I'll try to prepare a PR.

csharper2010:
Results from the query should be skipped if they are already in the ISet distinction thus if distinction.Add is false. Currently, they are only added if they have been there before.

Fixes #2614


hazzik:
Need tests to prevent future regressions.

hazzik (approved):

PR #2622, opened 2020-11-25
Fix enums used as parameter for string columns

### LINKED ISSUE #2621: Regression bug with enums used as parameter for string column
It used to be possible to provide an enum as a parameter for HQL queries where a criterion using that parameter matches on a string column.

I know that there are different options to have an enum directly in the property and map it to a string column on the database but sometimes there just might be the convention to use an enum for defining the possible values of a string column.

I would generally argue that enums should be treated like strings when mapped on string columns.

csharper2010:
A special treatment of enums mapped on string columns is required, because all reasons for checking string lengths or streamlining the length of parameters of queries on string columns to the length of the column for SQL Server
are also valid for enums, since they are sometimes by convention used to model the possible values of a string column.

Fixes #2621.


hazzik (approved):

csharper2010:
Sorry for not seeing this... Adding the logic directly to the AbstractStringType is surely better.

One question: should we make it more explicit for Enums because the Convert.ToType is very generic. Might help in some cases but might also cover real bugs in an application.

So handle Enums explicitly like I did in the Dialect classes?
`if (value is Enum)
    value = Enum.GetName(value.GetType(), value);`

Or leave it as is?

hazzik:
> Or leave it as is?

I don't know.

src/NHibernate.Test/NHSpecificTest/GH2621Enum/Fixture.cs
```diff
@@ -0,0 +1,20 @@
...
+	{
+		[Test]
+		public void TestOk()
+		{
+			using (var s = OpenSession())
+			using (s.BeginTransaction())
+			{
+				var query = s.CreateQuery(
+					@"SELECT Name FROM NHibernate.Test.NHSpecificTest.GH2621Enum.ClassWithString ROOT WHERE ROOT.Kind = :kind");
+				query.SetParameter("kind", Kind.SomeKind);
```
hazzik, line 15: Why not using one of the following methods?

```
/* 1 */ query.SetParameter("kind", Kind.SomeKind, NHibernateUtil.Enum(typeof(Kind)));
/* 2 */ query.SetEnum("kind", Kind.SomeKind);
/* 3 */ query.SetParameter("kind", Kind.SomeKind.ToString());
```

csharper2010: Because it's an existing code base where I cannot check all calls to SetParameter.

I would say that the change is a very specific one that does not hurt.

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -58,13 +58,16 @@ public override void Set(DbCommand cmd, object value, int index, ISessionImpleme
 		{
 			var parameter = cmd.Parameters[index];
 
+			if (value is Enum)
+				value = Enum.GetName(value.GetType(), value);
```
hazzik, line 62: Why not just `value = value.ToString()`?

gliljas:
If the objective is to solve a problem in an existing code base, why not just change the string type used, to one where this modification has been applied? That would require fixing the mapping, not all calls to SetParameter.

Using Enums as string constants is bad practice.

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -58,13 +58,16 @@ public override void Set(DbCommand cmd, object value, int index, ISessionImpleme
...
+
 			//Allow the driver to adjust the parameter for the value
 			session.Factory.ConnectionProvider.Driver.AdjustParameterForValue(parameter, SqlType, value);
 
 			// set the parameter value before the size check, since ODBC changes the size automatically
 			parameter.Value = value;
 
-			if (parameter.Size > 0 && ((string)value).Length > parameter.Size)
+			if (parameter.Size > 0 && value != null && ((string)value).Length > parameter.Size)
```
fredericDelaporte, line 70: `Set` is called by `NullSafeSet` which checks for `null` and does not call `Set` in case of a `null` value. So this null check here is moot, there is no need for it.

fredericDelaporte (commented):
I am not in favor of this change. It is too specific, and easy to manage on user side by deriving a custom type as suggested by Gunnar.

hazzik:
@fredericDelaporte what about changes as in e8c8a3b07e3a06c545c7bdf43472877220a04927 (minus null check)?

fredericDelaporte:
That sounds better to me, why not accepting this e8c8a3b (minus null check).

hazzik:
@fredericDelaporte @gliljas please check. This change is consistent with how some other types handle parameters, for example:

https://github.com/nhibernate/nhibernate-core/blob/7f60bc3a875a721d6ad4b76b530f2447da79454f/src/NHibernate/Type/Int32Type.cs#L59-L62

fredericDelaporte (dismissed):

csharper2010:
> 
> 
> @fredericDelaporte @gliljas please check. This change is consistent with how some other types handle parameters, for example:
> 
> https://github.com/nhibernate/nhibernate-core/blob/7f60bc3a875a721d6ad4b76b530f2447da79454f/src/NHibernate/Type/Int32Type.cs#L59-L62

Sorry for picking up that discussion that late but to be honest, I fear that the general approach using Convert.ToString might cover programming bugs of app developers. Convert.ToString(new object()) leads to "System.Object", the result of Convert.ToString(1.5m) is locale-dependent.

On the one hand I am interested in a solution but that might be a bit risky. That's why I originally had the enum handled specifically because it has a natural way of being expressed as string (admittedly there are the edge cases with integers not being represented as enum string so my solution with Enum.GetName might not be perfect, e.g. it leads to null for (FileAccess)0 whereas  Convert.ToString((FileAccess)0) gives the string "0").

Maybe we should guard the Convert.ToString with a list of defined cases where conversion to string is safe in some way.

`var stringValue = value as string;`
`if (stringValue == null && value != null)`
`{`
`    if (value is Enum || value is xxx || value is yyy)`
`        stringValue = Convert.ToString(value);`
`    else`
`        throw new ArgumentException();`
`}`

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -57,14 +57,15 @@ public AbstractStringType(SqlType sqlType)
 		public override void Set(DbCommand cmd, object value, int index, ISessionImplementor session)
 		{
 			var parameter = cmd.Parameters[index];
+			var stringValue  = Convert.ToString(value);
```
bahusoid, line 60: > I fear that the general approach using Convert.ToString might cover programming bugs of app developers. Convert.ToString(new object()) leads to "System.Object", the result of Convert.ToString(1.5m) is locale-dependent.

I totally agree with @csharper2010 on this. We should not silently convert parameters with incorrect types to string. I see it as a source of hard to find bugs

csharper2010:
Hi again,

I would like to pick that up again...
Should I revert the change back to the original one with just handling Enums? Do I have a chance to get that landed? There will not be a risk of regression bugs if we make it small and there is also no risk of covering application bugs as with Convert.ToString().

Thanks

hazzik:
Rebased and dropped all other commits.

hazzik:
Reverted the 'fix' as I'm trying to understand what are we fixing here.

hazzik:
Ok, this was dragging for a long time. We need to decide what to do. Just for context enum might not have a name (!) and [`Enum.GetName(...)`](https://learn.microsoft.com/en-us/dotnet/api/system.enum.getname) will return `null`. It is the case for `[Flags]` enums and enums with out of bound values

> **Returns**
> String
> A string containing the name of the enumerated constant in `enumType` whose value is `value`; or `null` if no such constant is found.

These unnamed enum values however, have a valid `ToString()` representation that can be parsed back.

```csharp
using System;

Print(E.A | E.B);             // -> Name: ''; ToString: 'A, B'
Print(Enum.Parse<E>("A, B")); // -> Name: ''; ToString: 'A, B'

Print((E) 100);               // -> Name: ''; ToString: '100'
Print(Enum.Parse<E>("100"));  // -> Name: ''; ToString: '100'

void Print(E e) => Console.WriteLine($"Name: '{Enum.GetName(e.GetType(), e)}'; ToString: '{e.ToString()}'");

[Flags]
enum E
{
	A = 1 << 0,
	B = 1 << 1,
	C = 1 << 2
}
```

So, with the proposed solution, it is possible to have the value lost in transit, and this is **NOT** a desirable outcome.

I personally do not want to add a workaround for clearly invalid mapping. However, I'm open to accept the special treatment for the enums in a form of `ToString(...)` call if @nhibernate/core agrees:

```csharp
if (value is Enum e)
      value = e.ToString(CultureInfo.InvariantCulture);
```

hazzik:
PS: should have re-read the whole thread before posting my reply :-(

src/NHibernate/Type/AbstractStringType.cs
hazzik, line 62:
```suggestion
			if (value is Enum e)
				value = e.ToString(CultureInfo.InvariantCulture);
```

hazzik (approved):

bahusoid (dismissed):

hazzik (dismissed):

src/NHibernate/Type/AbstractStringType.cs
```diff
@@ -58,13 +58,16 @@ public override void Set(DbCommand cmd, object value, int index, ISessionImpleme
...
+				value = value.ToString();
+
 			//Allow the driver to adjust the parameter for the value
 			session.Factory.ConnectionProvider.Driver.AdjustParameterForValue(parameter, SqlType, value);
 
 			// set the parameter value before the size check, since ODBC changes the size automatically
 			parameter.Value = value;
 
-			if (parameter.Size > 0 && ((string)value).Length > parameter.Size)
+			if (parameter.Size > 0 && value != null && ((string) value).Length > parameter.Size)
```
fredericDelaporte, line 70: Undue change. `Set` is not supposed to be called with a `null` value. It is called by `NullSafeSet` only if the value is not null.

hazzik:
```suggestion
			if (parameter.Size > 0 && ((string)value).Length > parameter.Size)
```

hazzik: @fredericDelaporte reverted.

bahusoid (approved):

fredericDelaporte (approved):

hazzik (approved):

Issue #2642, opened 2020-12-29
Linq expression parser removes required Convert nodes since 5.3


nkreipke:
Issue is reproducible with this test case:

```csharp
private readonly struct GuidWrapper
{
	public readonly Guid Id;

	public GuidWrapper(Guid id)
	{
		Id = id;
	}

	public static implicit operator Guid(GuidWrapper idWrapper)
	{
		return idWrapper.Id;
	}
}

[Test]
public void CanUseImplicitCastOperatorToGuid()
{
	var id = new GuidWrapper(new Guid("{01234567-abcd-abcd-abcd-0123456789ab}"));

	// Implicit cast operator is used here, expression for "id" is something like
	// Convert(value(...).id)
	var query = session.Query<Shipper>().Where(a => a.Reference == id);

	Assert.DoesNotThrow(() =>
	{
		var _ = query.ToList();
	});
}
```

This used to work in 5.2. In 5.3, we get the following errors depending on the DB driver:

**SQL Server**:
```
System.InvalidCastException: Failed to convert parameter value from a GuidWrapper to a Guid.
---> System.InvalidCastException: Object must implement IConvertible.
```

**PostgreSQL**:
```
System.InvalidCastException: Can't write CLR type NHibernate.Test.Linq.WhereTests+GuidWrapper with handler type UuidHandler
```

When creating the `NamedParameter`, the "Convert" expression is ignored and instead the parameter is assigned the value of `GuidWrapper` directly. Will try to provide a PR fixing this shortly.

hazzik:
@maca88 this is a result of changes in `NhPartialEvaluatingExpressionVisitor` could you please take a look?

maca88:
Correct, this is a regression from #2335. I think that @nkreipke fix is the best way to solve the issue as it still allows the parameter detection logic to optimize the parameter count in case the variable is used multiple times in the query.

fredericDelaporte:
Fixed by #2642

PR #2643, opened 2020-12-29
Add special case for constants contained within Convert nodes (#2642) 

nkreipke:
Test case and fix for issue #2642 

- Extracted `AddConstantExpressionParameter` method, added an extra parameter to "override" the value instead of always using `expression.Value`
- Added a special case for a constant contained within a "Convert" node. Instead of using the constant value, evaluate the node so the conversion is performed correctly.

Tested with SQL Server and PostgreSQL.

### DIFF
--- a/src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
+++ b/src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
@@ -146,7 +146,9 @@ namespace NHibernate.Linq.Visitors
 			// If we have an expression like "Convert(<constant>)" we no not want to lose the conversion operation
 			// because it might be necessary if the types are incompatible with each other, which might happen if
 			// the expression uses an implicitly or explicitly defined cast operator.
-			if (node.NodeType == ExpressionType.Convert && node.Operand is ConstantExpression constantExpression)
+			if (node.NodeType == ExpressionType.Convert &&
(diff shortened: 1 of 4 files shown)


nkreipke:
Build is failing because the language feature "readonly struct" is unavailable. Should I remove this or can we upgrade to C# 7.2?

maca88:
@nkreipke thanks for the PR, I added a small optimization and some additional tests.

src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -135,49 +135,72 @@ protected override Expression VisitConstant(ConstantExpression expression)
...
+		}
 
-				_parameters.Add(expression, parameter);
+		protected override Expression VisitUnary(UnaryExpression node)
+		{
+			// If we have an expression like "Convert(<constant>)" we no not want to lose the conversion operation
+			// because it might be necessary if the types are incompatible with each other, which might happen if
+			// the expression uses an implicitly or explicitly defined cast operator.
+			if (node.NodeType == ExpressionType.Convert &&
+			    node.Method != null && // The implicit/explicit operator method
```
maca88, line 150: Avoid compiling and evaluating the value for .NET built-in conversions.

fredericDelaporte (approved):

PR #2645, opened 2021-01-07
Allow specifying the size of the query plan cache

gliljas:
Implemented "query.plan_cache_max_size" and "query.plan_parameter_metadata_max_size" from Hibernate.

### DIFF
--- a/src/NHibernate/Cfg/Environment.cs
+++ b/src/NHibernate/Cfg/Environment.cs
@@ -5,6 +5,7 @@ using System.Reflection;
 using NHibernate.Bytecode;
 using NHibernate.Cfg.ConfigurationSchema;
 using NHibernate.Engine;
+using NHibernate.Engine.Query;
 using NHibernate.Linq;
 using NHibernate.Linq.Visitors;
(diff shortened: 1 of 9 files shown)


src/NHibernate/Cfg/Environment.cs
```diff
@@ -396,6 +397,30 @@ public static string Version
...
+		/// 
+		/// maintained by <see cref="QueryPlanCache"/>. Default is 128.
+		/// </summary>
+		public const string QueryPlanCacheMaxSize = "query.plan_cache_max_size";
+
+		/// <summary>
+		/// The maximum number of <see cref="ParameterMetadata"/> maintained
+		/// by <see cref="QueryPlanCache"/>. Default is 128.
+		/// </summary>
+		public const string QueryPlanCacheParameterMetadataMaxSize = "query.plan_parameter_metadata_max_size";
```
hazzik, line 422: does it need to be added to the configuration schema?

gliljas: Good catch! Done.

hazzik (approved):

PR #2656, opened 2021-01-29
Make sure dbcommand is disposed

RogerKratz:
DbCommands created in GenericBatchingBatcher aren't disposed.

### DIFF
--- a/src/NHibernate/AdoNet/GenericBatchingBatcher.cs
+++ b/src/NHibernate/AdoNet/GenericBatchingBatcher.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
 using System.Text;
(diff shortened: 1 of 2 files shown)


fredericDelaporte (approved):
I have qualified it as just a minor improvement instead of a major bug because in most cases with db commands, this disposal lack will not cause an unmanaged leak and will not prevent the command from being garbage collected.

If I am wrong, this fix should then be re-tagged, but also should then target 5.3.x and likely be back-ported down to the introduction of the generic batching batcher.

RogerKratz:
Agreed, I think it's minor as well, no unmanaged leak when `SqlCommand` is used I think. Not even sure it matters GC perf wise (maybe `GC.SuppressFinalize` is always called internally in `SqlCommand` anyhow, haven't looked closely)?

Still good to have it fixed (as "minor", yes).

RogerKratz:
....on the other hand it might cause an unmanged leak if an other `DbCommand` than `SqlCommand` is used. However, I don't know much about these, only thinking out loud now.

hazzik:
@fredericDelaporte maybe for 5.3.6?

fredericDelaporte:
Why not.

fredericDelaporte:
Changing the base did not went well, so I have made it manually with a forced push.

PR #2664, opened 2021-02-04
Fix subquery "Item with Same Key has already been added” exception for Linq provider

### LINKED ISSUE #2659: IQueryable<T> filter by subquery gives "Item with Same Key has already been added”
Execute this query, gives the following exception:
```
statelessSession.Query<Sale>()
    .Select(x => new SaleDto
    {
        Id = x.Id,
        SaleDate = x.ListChildren.Max(y => y.Date)
    })
    .Where(x => x.SaleDate <= DateTime.Today && x.SaleDate >= DateTime.Today)
    .ToList();
```
Exception:
`System.ArgumentException: Item with Same Key has already been added.`

maca88:
Fixes #2659


src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -317,8 +317,8 @@ protected HqlTreeNode VisitBinaryExpression(BinaryExpression expression)
 			var rightType = GetExpressionType(expression.Right);
 			if (leftType != null && leftType == rightType)
 			{
-				_notCastableExpressions.Add(expression.Left, leftType);
-				_notCastableExpressions.Add(expression.Right, rightType);
+				_notCastableExpressions.TryAdd(expression.Left, leftType);
```
hazzik, line 320: TryAdd is not available in .net standard 2.0 and .net 4.6.1

maca88: Added an extension method for them.

fredericDelaporte (approved):

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
hazzik: @maca88 can we replace Add with Set? 

_notCastableExpressions[expression.Right] = rightType;

hazzik:
@maca88 one last question from me, see in the comments section.

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
maca88: Somehow I overlooked the comment. Yes we can, replaced with Set. 
Also now I realized that I created the branch to this repository instead on my fork. We should remove the branch after merging.

hazzik (approved):

src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
hazzik: >Somehow I overlooked the comment.

This is what I thought.

PR #2668, opened 2021-02-06
Upgrade log4net

fredericDelaporte:
Fix a security warning raised for our test project.


hazzik (approved):

fredericDelaporte:
Re-tagged as a fix as I do not think this is worth being in release notes.

PR #2705, opened 2021-03-17
Propagate ExpectedType to CaseNode in hql

### LINKED ISSUE #2704: IEnhancedUserType from string to bool fails in some circumstances
I have a custom mapping type that transform a string (N or S) from database to a bool on the entity side.
This custom mapping type works fine in NH 5.2.x. After the migration to NH 5.3.x the custom mapping type works well except in one scenario: when the query check the inverse value of one variable, as show under:

```
var no = false;
Assert.IsTrue(_database.GetSession().Query<Entity1>().Where(x => x.IsChiusa == !no).Any());
```

I attach here a unit test to show the issue in action.

Thanks,
Pasquale.

[issue_in_action.zip](https://github.com/nhibernate/nhibernate-core/files/6150580/issue_in_action.zip)

bahusoid:
Fixes #2704

The following expression `x.IsChiusa == !no`

In 5.2 is generated simply as `alias.IsChiusa=@p0`  (`@p0` is set to `!no`)
In 5.3 is generated as `alias.IsChiusa=case when  not (@p0='S') then 1 else 0 end`  (`@p0` is set to `no`). Where `1` and `0` are invalid values inside `case` statement, as `IsChiusa` is string column.

Not sure what causes different query (maybe #2335) but provided fix makes sure that constants inside `case` statement are properly translated for custom type


maca88:
> Not sure what causes different query (maybe #2335)

That is correct, this is a regression of #2335 which prevented variables to be evaluated. Thanks for fixing the issue.

maca88 (approved):

Issue #2696, opened 2021-03-08
Transaction dispose should reset internal state


uciprian:
If transaction object is disposed then the status flags
- IsActive should be set to false
- WasRolledBack should be set to true

A client otherwise may attempt to rollback it and will throw Cannot access a disposed object

Or provide a way to tell internal state has been disposed like having public IsDisposed flag

fredericDelaporte:
In normal circumstances, `IsActive` will be `false` after a disposal. See the code.
`IsActive`:
https://github.com/nhibernate/nhibernate-core/blob/47992e73ed6114093f31ed99829dd8c4c7663829/src/NHibernate/Transaction/AdoTransaction.cs#L310-L313
`Dispose(bool disposing)`, when `disposing` is `true`, so when a `Dispose()` call is made:
https://github.com/nhibernate/nhibernate-core/blob/47992e73ed6114093f31ed99829dd8c4c7663829/src/NHibernate/Transaction/AdoTransaction.cs#L386-L390
`session` can be null only after a call to `AfterTransactionCompletion`, so if the transaction is still active, it will not be null and `AfterTransactionCompletion` will be called.
`AfterTransactionCompletion`:
https://github.com/nhibernate/nhibernate-core/blob/47992e73ed6114093f31ed99829dd8c4c7663829/src/NHibernate/Transaction/AdoTransaction.cs#L173-L183

So how can you get a transaction that would still have `IsActive` yielding `true` after a disposal? Hasn't your application thrown any exception?

Still we can set the `rolledBack` flag in the `// Assume we are rolled back`, but I do not think it will change much for your case.

PR #2706, opened 2021-03-21
Set the rolledBack flag when disposing active transactions

fredericDelaporte:
This is a slight improvement related to #2696. When disposing an active transaction, the underlying `DbTransaction` is disposed. Most implementations if not all trigger a rollback in such case, so we should flag the transaction has having been rollbacked.

(As it seems the opener obtains disposed transactions with `IsActive` being still `true`, I do not think it will fix his case, because the current code base as I read it does not allow such a situation in normal circumstances (without an exception occurring during disposal, while having explicitly disposed the transaction).)

### DIFF
--- a/src/NHibernate/Transaction/AdoTransaction.cs
+++ b/src/NHibernate/Transaction/AdoTransaction.cs
@@ -383,10 +383,12 @@ namespace NHibernate.Transaction
 						log.Debug("DbTransaction disposed.");
 					}
 
-					if (IsActive && session != null)
+					if (IsActive)
 					{
 						// Assume we are rolled back
-						AfterTransactionCompletion(false);
+						rolledBack = true;
+						if (session != null)
+							AfterTransactionCompletion(false);
 					}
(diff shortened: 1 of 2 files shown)


hazzik (approved):

src/NHibernate/Transaction/AdoTransaction.cs
```diff
@@ -383,10 +383,12 @@ protected virtual void Dispose(bool isDisposing)
 						log.Debug("DbTransaction disposed.");
 					}
 
-					if (IsActive && session != null)
+					if (IsActive)
 					{
 						// Assume we are rolled back
-						AfterTransactionCompletion(false);
+						rolledBack = true;
```
hazzik, line 389: But it can be that transaction is already committed, no?

hazzik: Ah, I see. `IsActive` checks for `!rolledBack`

Issue #2714, opened 2021-03-29
QueryExecutionCount statistic does not count all queries


deAtog:
When loading data from a session there are several cases where queries are executed without being counted in the QueryExecutionCount. E.g. The following code will cause the QueryExecutionCount to be incremented:

```c#
Entity e = session.QueryOver<Entity>()
    .Where(e => e.Id == id)
    .SingleOrDefault();
```

However, the following does NOT increment the QueryExecutionCount:
```c#
Entity e = session.Get<Entity>(id);
```

The implications of this issue is that tests which rely on this statistic are broken as the statistic does not accurately represent the number database queries executed.

bahusoid:
>  the following does NOT increment the QueryExecutionCount:
> session.Get<Entity>(id)


It looks intentional to me. Query here is not about any DB queries - but about loading via Query API.  `session.Get` is counted separately in `EntityLoadCount`

deAtog:
@bahusoid The current usage hides unexpected behavior as tests cannot verify the total number of queries actually executed against the database. `session.Get` is merely one such example, there are many others. Just because something functions correctly does not mean it's implemented correctly. This issue is related to another issue that I have yet to post. Without a fix for this issue, tests cannot be written to correct the unexpected behavior as the EntityLoadCount is correct, but the queries executed and the query count is wrong.

fredericDelaporte:
Query Statistics are not about SQL queries as explained by bahusoid, but about NHibernate queries. A get or load is not a NHibernate query. Trying to use them for counting SQL queries cannot be reliable, they are just not meant for this. And changing them for counting SQL queries would corrupt its semantic, and would be a breaking change.

deAtog:
@fredericDelaporte Yes, I understand. Maybe another metric should be added to measure/count all database queries executed? For the time being, I'll see if the other metrics capture the inconsistent behavior I'm seeing as suggested by @bahusoid.

fredericDelaporte:
I would need to dig the code a bit to asses this, but maybe an [interceptor or some events](https://nhibernate.info/doc/nhibernate-reference/events.html) could allow you to get this kind of statistics. So no built-in way, but it may be an injection point allowing you to build the feature.

PR #2715, opened 2021-03-29
Ensure all queries executed are counted.

deAtog:
This is a WIP to address issue #2714.


fredericDelaporte:
This PR goes a wrong way, trying to change the intended semantic of the Statistics class from counting NHibernate queries to counting SQL queries. This class is not meant for this as said in #2714.

PR #2723, opened 2021-04-07
Avoid double param type guessing and better NULL parameter handling in LINQ

bahusoid:
And added test cases for #1166


hazzik:
> Figure out why firebird generates ...

Because Firebird requires parameters to be casted.

hazzik:
@bahusoid please resolve conflicts.

Also as I understand some changes needs to be reverted from 5.3.9

bahusoid:
I believe it would also need changes similair to https://github.com/nhibernate/nhibernate-core/pull/2836

bahusoid:
> please resolve conflicts.
Also as I understand some changes needs to be reverted from 5.3.9

Done

fredericDelaporte:
There is a conflict with #2964 to resolve.

hazzik:
Hopefully I've resolved the conflict correctly. @maca88 and @bahusoid could you please check.

maca88:
From my point of view the conflict was correctly solved. Even if the modification in `GetCharsGenerator` class fixes the tests, I would still leave the change from #2964 as there could be other cases (e.g. with user defined hql functions) for which we don't have tests.

bahusoid:
It seems some of the changes no longer needed. I need to look at it more closely when I have time...

bahusoid:
#2964 implements most of the logic of original PR. Now this PR is transformed in parameter handling improvement so changed the name.

fredericDelaporte (approved):

hazzik (approved):

Issue #2735, opened 2021-04-21
UpdateTimestampsCache non cached entities concurrency issue


gokhanabatay:
Hi @hazzik @maca88,
As I mentioned earlier current UpdateTimestampsCache has still concurrency issue. #2129

In real life scenario some of entities uses second level cache but most of entities does not uses second level cache but still try to update UpdateTimestampsCache. At least it should be configurable in entity level or session factory level, default may be true for backward compability issue.

Non cached entities could be related(foreign key) with cached entity may be better solution StandardQueryCache property spaces and cached entity property spaces only triggers UpdateTimestampsCache?

Current UpdateTimestampsCache uses AsyncReaderWriterLock if one writer grants lock other writers of different entities or readers wait for it. If this is true you could realize the concurrency issue that we are facing under high concurrency (400-500tps).

If still not acceptable please consider a way to implement Custom UpdateTimestampsCache via UpdateTimestampsFactory like cache.provider_class attribute in configuration. Could be named cache.update_timestamps_class.

Currently we are trying to replace current UpdateTimestampsCache via reflection but its not cleanest solution.

```c#
var cacheablePersisters = ISessionFactory.GetAllClassMetadata()
                .Where(x => x.Value.GetType() == typeof(SingleTableEntityPersister))
                .Select(x => x.Value)
                .Cast<SingleTableEntityPersister>()
                .Where(x => x.HasCache)
                .ToList();

var sessionFactoryImpl = (SessionFactoryImpl)ISessionFactory;
var standardQueryCache = (StandardQueryCache)sessionFactoryImpl.QueryCache;


var updateTimestamps = (CacheBase)typeof(UpdateTimestampsCache).GetField("_updateTimestamps", BindingFlags.Instance | BindingFlags.NonPublic).GetValue(sessionFactoryImpl.UpdateTimestampsCache);

var updateTimestampsCache = new DistributedUpdateTimestampsCache(updateTimestamps, cacheablePersisters.SelectMany(x => x.PropertySpaces).ToHashSet());

typeof(SessionFactoryImpl).GetField("updateTimestampsCache", BindingFlags.Instance | BindingFlags.NonPublic).SetValue(sessionFactoryImpl, updateTimestampsCache);

if (standardQueryCache != null)
{
    typeof(StandardQueryCache).GetField("_updateTimestampsCache", BindingFlags.Instance | BindingFlags.NonPublic).SetValue(standardQueryCache, updateTimestampsCache);
}

```

[...]

PR #2742, opened 2021-04-28
Remove locks from UpdateTimestampsCache

hazzik:
See #2735

> Hibernate has done [it] a [long time ago](https://github.com/hibernate/hibernate-orm/commit/f85e9247e918c1ddb2b76a6e870f6ad730020788) to improve performance. 

Possible breaking change: 
> In theory the read/write locking prevents that `IsUpToDate` would return stale data, but that holds true only when one instance of the application is running. In case multiple application instances are running then `IsUpToDate` may return stale data due to another application instance writing data at the same time. So we could remove the locking entirely as Hiberanate did or add a configuration setting for disabling it.

_Originally posted by @maca88 in https://github.com/nhibernate/nhibernate-core/issues/2735#issuecomment-826136231_

### DIFF
--- a/src/NHibernate/Async/Cache/UpdateTimestampsCache.cs
+++ b/src/NHibernate/Async/Cache/UpdateTimestampsCache.cs
@@ -11,8 +11,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.CompilerServices;
-
 using NHibernate.Cfg;
 using NHibernate.Util;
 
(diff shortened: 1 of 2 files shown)


maca88 (approved):

gokhanabatay:
@maca88 When we use [redis pub/sub distributed cache in your implementation](https://github.com/nhibernate/NHibernate-Caches/blob/master/StackExchangeRedis/NHibernate.Caches.StackExchangeRedis/DistributedLocalCacheRegionStrategy.cs) non cached entities causes much more syncronization events. Think about a configuration in mapping `Cache.Never()` that tells this entity never cached with query cache or entity cache ever.
For performance needs this is absolutly necessary.

gokhanabatay:
5.3.9 released but does not contains this pr, it was tagged as next minor am I missing something. @fredericDelaporte @hazzik

fredericDelaporte:
Current minor is 5.3.x. Next minor will be 5.4.x, or 6.0.x if we decide to go for a next major instead.

hazzik:
@gokhanabatay yes, this was for 5.4.x, because we cannot change the behavior or add new features in 5.3.x

fredericDelaporte:
@gokhanabatay, in case you were not knowing, you can still use this feature by using the latest [MyGet package](https://www.myget.org/gallery/nhibernate).

gokhanabatay:
@fredericDelaporte thanks we are using myget latest nhibernate package for performance needs.

Issue #2632, opened 2020-12-08
ISession.Merge() not returning the entity id for the new entity on version 5.3.x


Afshin1980:
I updated the NHibernate from 5.2.7 to 5.3.0 and noticed that there is a difference between the returning object from the merge function.  In version 5.3.x the merge function doesn't return the entity id if the entity is not in the database but in version 5.2.7 the merge function returns the entity id. I don't know this is a bug or not 

Here is the code 
```
using NHibernate.Cfg;
using NHibernate.Dialect;
using NHibernate.Driver;
using NHibernate.Mapping.ByCode;
using NHibernate.Mapping.ByCode.Conformist;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Reflection;
using System.Text;
using System.Threading.Tasks;

namespace TestNHibernate
{
    class Program
    {
        static void Main(string[] args)
        {
            var cfg = new Configuration();

            cfg.DataBaseIntegration(x =>
            {
                x.ConnectionString = "Data Source=.;integrated security=true;Initial Catalog=TestDB;MultipleActiveResultSets=True";
                x.Driver<SqlClientDriver>();
                x.Dialect<MsSql2008Dialect>();
            });

            cfg.AddAssembly(Assembly.GetExecutingAssembly());

            var sefact = cfg.BuildSessionFactory();

            using (var session = sefact.OpenSession())
            {
                var person = new Person
                {
                    Name= "Jamie"
                };

                using (var tx = session.BeginTransaction())
                {
                    var merged = session.Merge(person);

                    session.SaveOrUpdate(merged);
                    tx.Commit();
                }
                                                
                Console.ReadLine();
            }
        }
    }

    public class Person
    {
        public virtual int ID { get; set; }
        public virtual string Name { get; set; }
        public virtual string Tel { get; set; }
    }
}
```

![5 2 7](https://user-images.githubusercontent.com/22458044/101498908-a2f84c80-393a-11eb-8974-662a33f6fdb9.png)
![5 3 0](https://user-images.githubusercontent.com/22458044/101498909-a2f84c80-393a-11eb-96c8-8d23af50d9be.png)

fredericDelaporte:
This is a change coming from #1754. But it is unclear for me whether `Merge` should follow that change or not.

We should at least document it in the possible breaking changes of 5.3.

bahusoid:
[...]

PR #2752, opened 2021-05-10
Change cascade style for DefaultDirtyCheckEventListener to persist to avoid flushing the session

### LINKED ISSUE #1413: Make ISession.IsDirty() method safe, effective and useful

hazzik:
The default cascade for it was SaveUpdate (via base AbstractFlushingEventListener) which triggered
insertion of the entities with post-insert style ID generators (eg. identity). Persist on the other hand will
make sure that insertion of the entity is delyed until flush is called.

Related to #1413

### DIFF
--- a/src/NHibernate.Test/Async/NHSpecificTest/GH1413/FixtureByCode.cs
+++ b/src/NHibernate.Test/Async/NHSpecificTest/GH1413/FixtureByCode.cs
@@ -74,13 +74,12 @@ namespace NHibernate.Test.NHSpecificTest.GH1413
 			}
 		}
 
-		[Test]
-		[KnownBug("#1413")]
-		public async Task SessionIsDirtyShouldNotTriggerCascadeSavingAsync()
+		[Theory]
+		public async Task SessionIsDirtyShouldNotTriggerCascadeSavingAsync(bool beginTransaction)
 		{
 			Sfi.Statistics.IsStatisticsEnabled = true;
 			using (var session = OpenSession())
-			using (session.BeginTransaction())
+			using (beginTransaction ? session.BeginTransaction() : null)
 			{
 				var parent = await (GetParentAsync(session));
 				var entityChild = new EntityChild
--- a/src/NHibernate.Test/Async/NHSpecificTest/GH1413/FixtureByCode.cs
+++ b/src/NHibernate.Test/Async/NHSpecificTest/GH1413/FixtureByCode.cs
@@ -9,6 +9,7 @@
 
 
 using NHibernate.Cfg.MappingSchema;
+using NHibernate.Engine;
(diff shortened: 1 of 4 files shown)


hazzik:
@bahusoid @fredericDelaporte could you please review?

fredericDelaporte (approved):

hazzik:
@fredericDelaporte can I please get some thoughts why this change might be a bad idea? To be honest I wasn't looking for a silent tick ;-)

bahusoid:
Well it's not exactly immutable operation (session state is changed) but it's definitely an improvement over old behavior. So why not..

But I still would like to have IsDirty check that doesn't modify session in any way  - so maybe let's not close #1413 yet.

> why this change might be a bad idea?

Not exactly sure about real use case but something similar to https://github.com/nhibernate/nhibernate-core/issues/2632 might come up

fredericDelaporte:
I did think a lot about what could possibly wrong by switching the cascade style for the dirty check from SaveUpdate to Persist, but I could not find anything. So I have approved.

hazzik:
> But I still would like to have IsDirty check that doesn't modify session in any way - so maybe let's not close #1413 yet.

Sure.

Roman-Evseev:
This change still allows IsDirty to modify the database, causing a NullReferenceException in certain scenarios. Therefore, the commit description does not accurately reflect the intended goals.

PR #2755, opened 2021-05-13
Fix stateless batcher

### LINKED ISSUE #2750: Using System.Transaction with IStatelessSession doesn't always flush batches to database
I'm having two issues with using `System.Transaction` and `IStatelessSession` together.

First, it doesn't work at all with the default `UseConnectionOnSystemTransactionPrepare` value of `true`. Committing a transaction like:
```
using (var transaction = new TransactionScope())
{
	using (var session = sessionFactory.OpenStatelessSession())
	{
		// save a change here
	}
	transaction.Complete();
}
```
always results in this exception:
```
System.Transactions.TransactionAbortedException: The transaction has aborted. ---> System.InvalidOperationException: Batcher still has opened ressources at time of processing from system transaction.
   at NHibernate.AdoNet.ConnectionManager.BeginProcessingFromSystemTransaction(Boolean allowConnectionUsage)
   at NHibernate.Transaction.AdoNetWithSystemTransactionFactory.SystemTransactionContext.Prepare(PreparingEnlistment preparingEnlistment)
   at System.Transactions.TransactionStateAborted.EndCommit(InternalTransaction tx)
   at System.Transactions.CommittableTransaction.Commit()
   at System.Transactions.TransactionScope.InternalDispose()
   at System.Transactions.TransactionScope.Dispose()
   at StatelessBatchFlush.Program.ValidateBatching(ISessionFactory sessionFactory) in C:\Projects\NHibernate-Stateless\Program.cs:line 68
```

Setting `UseConnectionOnSystemTransactionPrepare=false` does cause this exception to go away, but with that set the last batch of items does not get flushed to the database. `IStatelessSession` does not expose a `Flush()` method, but we've found that calling `session.GetSessionImplementation().Flush()` before completing the transaction is a workaround for this.

I've created a small console app to demo the problems I'm having here: https://github.com/jklawrence/NHibernate-Stateless

These same scenarios work fine when using an NHibernate transaction instead of System.Transaction.

NHibernate version: 5.3.8

fredericDelaporte:
Fix #2750.


src/NHibernate/Impl/StatelessSessionImpl.cs
```diff
@@ -351,7 +363,7 @@ public override bool IsOpen
 
 		public override FlushMode FlushMode
 		{
-			get { return FlushMode.Commit; }
+			get { return FlushMode.Always; }
```
fredericDelaporte, line 366: This is more consistent with the stateless semantic of a stateless session, and also with:
https://github.com/nhibernate/nhibernate-core/blob/07ecb9f25534d3894dfba6a3a8e8f22932bd06cf/src/NHibernate/Impl/SessionFactoryImpl.cs#L1605
(That is in the stateless session builder.)

src/NHibernate.Test/NHSpecificTest/GH2750/FixtureByCode.cs
```diff
@@ -0,0 +1,173 @@
...
+
+			using (var session = Sfi.OpenStatelessSession())
+			{
+				using (var transaction = new TransactionScope())
+				{
+					foreach (var entity in entities)
+					{
+						session.Insert(entity);
+					}
+					session.FlushBatcher();
```
hazzik, line 89: Is there any reason why we cannot just call `_session.Flush()` in `SystemTransactionContext.Prepare(PreparingEnlistment)`?

fredericDelaporte: Yes we cannot. That is too late, we are already in the 2PC, exposed to threading concurrency issues if trying to do anything that will use the session current connection.

At this point, the "use connection on system transaction" setting is already required for allowing attempt to use a connection in the 2PC. So anyway if this setting is not enabled (as recommended in the [documentation](https://nhibernate.info/doc/nhibernate-reference/transactions.html#transactions-scopes)), we cannot flush here.
But in the case of stateless session we cannot flush even if the setting is enabled, because if there is anything to flush in the batcher, it has already acquired the session current connection. And so the connection switch done in `ConnectionManager.BeginProcessingFromSystemTransaction` will have no effect on the batcher, causing it to flush on the session connection which may be executing its own 2PC on another thread at the same time.

There no way to be safe here excepted the [recommend documented one](https://nhibernate.info/doc/nhibernate-reference/transactions.html#transactions-scopes):
> For new applications, it is recommended to set transaction.use_connection_on_system_prepare to false, and to flush explicitly your sessions before scope completion.

hazzik:
Need to restrict tests to the providers that support distributed transactions.

fredericDelaporte:
I have amended the "more test" commit and rebased.

src/NHibernate.Test/NHSpecificTest/GH2750/FixtureByCode.cs
```diff
@@ -0,0 +1,179 @@
...
```
hazzik, line 95: `StatelessSession` has `ManagedFlush` method which is not exposed through interface. Maybe we should expose it instead?

fredericDelaporte: I choosed `FlushBatcher` rather than just `Flush` because in principle the stateless session should never need a `Flush`, since it is stateless. So I want to stress what this flush is for: flushing the batcher that can be enabled with the stateless session, and which may retain some operations not yet sent to the database.

hazzik: Just wanted to clarify that I'm not talking about the general `Flush` method, but about `ManagedFlush`.

fredericDelaporte: In the case of the stateless session, that is the same.

But anyway, what does it mean, `ManagedFlush`? I do not make any sense of it. `FlushBatcher` is explicit about its purpose, while that is not the case of `ManagedFlush` in my opinion.

https://github.com/nhibernate/nhibernate-core/blob/2bd3c1680e105dc496a4c97f6d8630ca362feee8/src/NHibernate/Impl/StatelessSessionImpl.cs#L385-L396

hazzik (approved):

Issue #2792, opened 2021-05-26
Arithmetic operations adding casts to SQLite that cause incorrect results


PleasantD:
Affects v5.3.0+

*How to Reproduce*
`var result = session.Query<Entity>().Where(e => e.DecimalValue != null && (offset + e.DecimalValue) == test).ToList();`
`var result = session.Query<Entity>().Where(e => e.IntValue != null && (offset + (decimal?)e.IntValue) == test).ToList();`

Where `offset` and `test` are `decimal` or `decimal?` types.

*Expected Result*
`result` contains a list of entities where the equality condition matches.
This works fine for PostgreSQL and MS SQL Server.

*Actual Result*
SQLite gives no errors and always produces 0 results.

PleasantD:
Prior to v5.3.0 the SQL generated for all 3 DB types looked like
```
SELECT /* omitted for berevity*/
FROM "Entity" entity0_
WHERE (entity0_."IntValue" IS NOT NULL)
	AND 5 /* @p1 */ + entity0_."IntValue" = 15 /* @p2 */
```

After v5.3.0 the PostgreSQL and MSSQL queries look like
```
SELECT /* omitted for berevity*/
FROM "Entity" entity0_
WHERE (entity0_."IntValue" IS NOT NULL)
	AND 5 /* @p1 */ +  cast(entity0_."IntValue" AS DECIMAL(29, 10)) = 15 /* @p2 */
```

While the SQLite query looks like 
```
SELECT /* omitted for berevity*/
FROM "Entity" entity0_
WHERE (entity0_."IntValue" IS NOT NULL)
	AND 5 /* @p1 */ +  cast(entity0_."IntValue" AS REAL) = 15 /* @p2 */
```

PR #2793, opened 2021-05-30
Fix parameter detection for custom hql method generators

### LINKED ISSUE #2791: Custom Equality Fails
As of v5.3+ custom equality (and comparison) method handling is no longer possible for non-primitive types.
Let me unpack that.

We have a custom HQL generator implemented for `object.Equals`, since that isn't supported by NH directly.
The reason we need it is due to custom code that generates LINQ query expressions which need to be both C# and NH compatible, so in some cases we must use `object.Equals(object)` to correctly test for equality.

An example generated expression might look equivalent to `e => (MyEnum.A).Equals((object)e.MyEnumProp)`

Prior to v5.3.0 this worked fine for us. 
After v5.3.0 this  causes us errors, specifically when the mapped property is a String Enum type.

----
Some investigation led me to  the following code
https://github.com/nhibernate/nhibernate-core/blob/5.3.8/src/NHibernate/Linq/Visitors/ParameterTypeLocator.cs#L249

This method of the visitor is hit prior to evaluating the types of the parameters.
It appears to be building a dictionary of "related" expression subtrees so that the parameter evaluator can figure out that  `Enum.A` (a constant, and therefore an SQL parameter) is related to `e.MyEnumProp` (a mapped property) and can use the NH mapping to convert the parameter to the correct `IType`.

Only this branch never runs because `object.Equals(object)` is not on the hard-coded list of methods.
As a consequence, no relationship is added between the calling object and the mapped property argument.
Lacking any relationships, the parameter type evaluation falls back to guessing and, in this case, guesses that the parameter should be `EnumType` instead of `EnumStringType `.  This generates SQL that attempts to compare a string column with an int parameter, with errors or bad results depending on the DB.

maca88:
Fix for a regression made by #2365.

Fixes #2791


bahusoid:
What do you think of https://github.com/nhibernate/nhibernate-core/pull/2723?

I believe this change is not required on master if https://github.com/nhibernate/nhibernate-core/pull/2723 is merged. Do you see any drawbacks? All tests are green except the one I added (and most likely  it fails on dialects that don't distinguish/support AnsiString type)

maca88:
> What do you think of #2723?

I think that your approach is better as we preserve the guessed type in case we would need it in the future. So probably the best would be to rebase your PR to `5.3.x` and close this one.

> All tests are green except the one I added (and most likely it fails on dialects that don't distinguish/support AnsiString type)

The `StartsWith` and `Contains` tests can be fixed by adding their method definition to this line:
https://github.com/nhibernate/nhibernate-core/blob/5.3.8/src/NHibernate/Linq/Visitors/ParameterTypeLocator.cs#L249

maybe we need to make this line configurable by extending `IHqlGeneratorForMethod`.

bahusoid:
> So probably the best would be to rebase your PR to 5.3.x and close this one.

I'm not sure. IMO it's too general change for patch release - as you can see I had to adjust few places to make it work (like in `StringGenerator` and `CaseNode` classes) .  So other unnoticed regressions are also possible.

maca88:
I understand, so a safer option would be to merge this one and remove its changes in your PR.

PleasantD:
> The `StartsWith` and `Contains` tests can be fixed by adding their method definition to this line:
> https://github.com/nhibernate/nhibernate-core/blob/5.3.8/src/NHibernate/Linq/Visitors/ParameterTypeLocator.cs#L249
> 
> maybe we need to make this line configurable by extending `IHqlGeneratorForMethod`.

I agree that exposing the list of methods that are eligible for this relationship finding is desirable.
Since generator logic is customizable, it makes sense to me that this could be needed by arbitrary user defined methods.

Otherwise, this seems a minimally adequate solution for issue #2792

fredericDelaporte (approved):

bahusoid:
Errh... Firebird again... Though exceptions are different:
```C#
  Expected: not null
  But was:  null
 at NHibernate.Test.NHSpecificTest.NH3634.ByCodeFixtureAsync.<QueryOverComponentWithANullPropertyAsync>d__3.MoveNext() in C:\projects\nhibernate-core\src\NHibernate.Test\Async\NHSpecificTest\NH3634\FixtureByCode.cs:line 133
```

PR #2807, opened 2021-06-02
Fix decimal equality comparison for Sqlite

### LINKED ISSUE #2792: Arithmetic operations adding casts to SQLite that cause incorrect results

maca88:
It is a fix for a regression made by #2335.

Fixes #2792


src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs
```diff
@@ -579,8 +579,12 @@ protected HqlTreeNode VisitConstantExpression(ConstantExpression expression)
 			if (_parameters.ConstantToParameterMap.TryGetValue(expression, out namedParameter))
 			{
 				_parameters.RequiredHqlParameters.Add(new NamedParameterDescriptor(namedParameter.Name, null, false));
+				var parameter = _hqlTreeBuilder.Parameter(namedParameter.Name).AsExpression();
 
-				return _hqlTreeBuilder.Parameter(namedParameter.Name).AsExpression();
+				// SQLite driver binds decimal parameters to text, which can cause unexpected results in arithmetic operations.
+				return expression.Type.UnwrapIfNullable() == typeof(decimal) && _parameters.SessionFactory.Dialect.IsDecimalStoredAsFloatingPointNumber
```
bahusoid, line 585: Better swap those checks... Also maybe make `IsDecimalStoredAsFloatingPointNumber` public and use in tests instead of Dialect checks.

maca88: Done.

bahusoid (approved):
Why WIP? Fix looks complete to me

maca88:
I forgot to remove it.

PR #2808, opened 2021-06-02
Update Travis CI to use Ubuntu Xenial on 5.3.x

hazzik:
Seems trusty is broken on Travis. So borrow changes from master to 5.3.x builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
-dist: trusty
+os: linux
+dist: xenial
 language: csharp
 mono: none
 dotnet: 2.1.300
-sudo: required
 services:
   - mysql
   - postgresql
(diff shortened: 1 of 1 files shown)


hazzik:
https://www.powershellgallery.com certificate has expired :(

bahusoid:
> https://www.powershellgallery.com certificate has expired

It's already resolved. Also if you also can't no longer restart Travis jobs you need to select Free plan in account settings.

bahusoid (approved):

hazzik:
>  Also if you also can't no longer restart Travis jobs you need to select Free plan in account settings.

It seems it was needed once per org.

bahusoid:
Ouch.. It seems setting Free plan wasn't the best idea. It looks like Travis CI is no longer free for open source projects. Not sure why builds were still running for us with no plan set (glitch in their system?) - but now with free plan we have only 1000 build minutes and then need to request OSS credits https://docs.travis-ci.com/user/billing-faq/#what-if-i-am-building-open-source

fredericDelaporte:
Change to fix because it is obsoleted by #2812, so it will not be actually released. I will not list it in release notes.

PR #2812, opened 2021-06-06
Add GitHub Actions build

bahusoid:
As replacement for Travis.

Depends on #2811 

Pretty much the same experience except you currently can restart only all jobs (not possible to restart only failed jobs). And each job is displayed as separate check on github UI.

See results:
https://github.com/bahusoid/nhibernate-core/actions/runs/911127398


.github/workflows/NetCoreTests.yml
```diff
@@ -0,0 +1,57 @@
...
+        include:
+          - DB: SqlServer2008
+            CONNECTION_STRING: "Server=localhost;initial catalog=nhibernate;User Id=sa;Password=P@ssw0rd;packet size=4096;"
+          - DB: PostgreSQL
+            CONNECTION_STRING: "Host=localhost;Port=5432;Username=nhibernate;Password=nhibernate;Database=nhibernate;Enlist=true;"
+          - DB: Firebird
+            CONNECTION_STRING: "DataSource=localhost;Database=nhibernate;User=SYSDBA;Password=nhibernate;charset=utf8;"
+          - DB: MySQL
+            CONNECTION_STRING: "Server=localhost;Uid=root;Password=nhibernate;Database=nhibernate;Old Guids=True;"
+            ALLOW_FAILURE: true
```
bahusoid, line 19: This line should be removed when merged to master.

fredericDelaporte:
This PR comes with a commit from another not yet merged PR.

fredericDelaporte:
> Depends on #2811

That is the reason for this commit? What cause it to depend on #2811?

bahusoid:
Because #2811 for some reasons is almost 100% reproducible on github runners. So I wasn't able to test it properly.

bahusoid:
> This PR comes with a commit from another not yet merged PR. I It should be removed.

Ok. Done. It's no longer needed here as I finished testing. But #2811 should be merged first to avoid failed builds.

bahusoid:
@hazzik  Regarding adding windows in matrix. See https://github.com/bahusoid/nhibernate-core/blob/ghawithWin/.github/workflows/NetCoreTests.yml#L10

But containers don't work on windows: https://github.com/bahusoid/nhibernate-core/actions/runs/913556186

hazzik:
Maybe we can use [services](https://docs.github.com/en/actions/guides/about-service-containers) instead of docker

hazzik:
Ok, it seems services also require linux.

bahusoid:
> Maybe we can use services instead of docker

If I got it right "service" is just another way to configure docker container in Github Action. Only it doesn't support conditional configurations for jobs (like install only Firebird for this job)

.github/workflows/NetCoreTests.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Set up PostgreSQL
+      if: matrix.DB == 'PostgreSQL'
+      run: |
+       docker run -d -e POSTGRES_USER=nhibernate  -e POSTGRES_PASSWORD=nhibernate -e POSTGRES_DB=nhibernate -p 5432:5432 postgres
+
+    - name: Set up Firebird
+      if: matrix.DB == 'Firebird'
+      run: |
+        docker run --name firebird -e EnableWireCrypt=true -e FIREBIRD_USER=nhibernate -e FIREBIRD_PASSWORD=nhibernate -e ISC_PASSWORD=nhibernate -e FIREBIRD_DATABASE=nhibernate -p 3050:3050 -d jacobalberty/firebird
```
hazzik, line 47: Are we sure we want to test against some random container?

bahusoid: Not random. Latest stable... But you are right maybe better to  specify 3.0 explicitly - I suspect it would fail on 4.0 for some tests...

bahusoid: Or did you mean something else? Like should we trust/depend on some not official container? It's open source container  - I see no issue with it. It works for us - so why not...

bahusoid: I also saw Oracle container: https://hub.docker.com/r/gvenzl/oracle-xe. I think we should add it too sometime later.

hazzik (approved):

.github/workflows/NetCoreTests.yml
hazzik: >  Like should we trust/depend on some not official container?

Yes, I meant that.

> It's open source container with 1M+ downloads

It does not mean anything, really.

PR #2813, opened 2021-06-07
Fix intermittent Firebird test errors

bahusoid:
I hope it fixes it everywhere. Tests shown on 14 Github Action runs no failed FB test:
https://github.com/bahusoid/nhibernate-core/actions/runs/914629949
https://github.com/bahusoid/nhibernate-core/actions/runs/914686728
https://github.com/bahusoid/nhibernate-core/runs/2765215509

Without this change CI sometime fails with errors like:

```sql
FirebirdSql.Data.FirebirdClient.FbException : unsuccessful metadata update
CREATE TABLE TESTENTITY failed
Table TESTENTITY already exists
```
See https://github.com/bahusoid/nhibernate-core/runs/2753973440?check_suite_focus=true#logs
https://ci.appveyor.com/project/nhibernate/nhibernate-core/builds/39434450/job/ici409t79f6kie1c/tests

Let's add it to 5.3 as it's still active branch to avoid annoying build restarts.

### DIFF
--- a/src/NHibernate.Test/TestCase.cs
+++ b/src/NHibernate.Test/TestCase.cs
@@ -303,6 +303,9 @@ namespace NHibernate.Test
 
 		public static void DropSchema(bool useStdOut, SchemaExport export, ISessionFactoryImplementor sfi, Func<DbConnection> getConnection = null)
 		{
+			using(var optionalConnection = getConnection?.Invoke())
+				export.Drop(useStdOut, true, optionalConnection);
+
 			if (sfi?.ConnectionProvider.Driver is FirebirdClientDriver fbDriver)
 			{
 				// Firebird will pool each connection created during the test and will marked as used any table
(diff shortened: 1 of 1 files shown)


src/NHibernate.Test/TestCase.cs
```diff
@@ -303,6 +303,9 @@ protected virtual void DropSchema()
 
 		public static void DropSchema(bool useStdOut, SchemaExport export, ISessionFactoryImplementor sfi, Func<DbConnection> getConnection = null)
 		{
+			using(var optionalConnection = getConnection?.Invoke())
+				export.Drop(useStdOut, true, optionalConnection);
```
bahusoid, line 307: So my theory Drop commands may also be pooled. So ClearPool should be called after Drop.

bahusoid:
I accidentally opened another PR #2814 so will keep testing there. Also I easily reproduced error on this branch too with GH Action:
https://github.com/bahusoid/nhibernate-core/runs/2766079024

Turns out that my initial valid tests were with upgraded FB library. So it seems we need both - this code change +latest FB lib

PR #2816, opened 2021-06-09
Add Projections.Conditional for CASE expressions with multiple conditions

### LINKED ISSUE #2804: Projections.Conditional for CASE expressions with multiple conditions
Hello,

the NHibernate.Criterion.ConditionalProjection currently supports only 1 "when" and the "else" of the known SQL switch-case-expression.
About switch-case, see https://www.w3schools.com/sql/sql_case.asp 

Is there a specific reason why thats not available (compatibility?) and could it be please implemented in a future (soon) release?
Below i've stated a working example solution (i think).

### Our use-case:
We want to "order by" our Enums with translation (from culture resources), so we use switch-case to map the Enum-value to its culture specific sorting index (determined by the enum value's translation).

### Current state:
`
...
ORDER BY
case 
(when _this.BigEnum = @p10 then @p11 
else (when _this.BigEnum = @p12 then @p13 
else (when _this.BigEnum = @p14 then @p15 
else  (... else (.... else(... ) ) ) ) 
`
"p10" is the enum value ("Value10" or "Wert10" (german)) and "p11" its sorting index (10).

### The problem:
This way, the maximum case-intendation of 10 for MS-SQL server and 18 for SQLite is reached really fast and aborted with an exception ("parser stack overflow").
Multiple "when" parts without intendation would solve that.

### The possible solution:
I've already took some investigation in how to produce SQL with multiple "when-then" using Criteria with our own "ConditionalsProjection". See atttachement.
[NHibernate Criterion ConditionalsProjection multi-when-case.zip](https://github.com/nhibernate/nhibernate-core/files/6584711/NHibernate.Criterion.ConditionalsProjection.multi-when-case.zip)

`
...
ORDER BY
case 
(when _this.BigEnum = @p10 then @p11 
when _this.BigEnum = @p12 then @p13 
when _this.BigEnum = @p14 then @p15 
when ... then
when... then...
else ... )
`

### My further questions:
- Is the provided solution working for other aspects/requirements (compatibility) i don't see?
- Are there further changes in other files required? (registrations, tests?)

I hope that's enough information to specify my point ;) 

**Thank you very much and have a nice day.**

**Edit: fixed missing file in attachement! Sorry /fail**

perahoky:
Fixes #2804 

Not sure whether the testing is alright. I simply put MSSQLLocalDb or SQLite-InMemory for local testing.
Hope its okay.

Creates a "switch" expression with multiple "when then" parts.
For example:

`
select this_.Name, this_.Enum
from Examples 
order by 
case (
when this_Enum = "EnumValueEnglish" then 1,
when this_Enum = "EnumValueGerman" then 2,
when this_Enum = "EnumValueFrance" then 3
else 99)
`

have a nice day.
Pera


hazzik:
> The following classes was expected to be marked as Serializable:

>    NHibernate.Criterion.ConditionalsProjection

perahoky:
Thank you hazzik :)

hazzik:
We also need feature parity from QueryOver

hazzik (approved):

src/NHibernate/Criterion/Projections.cs
```diff
@@ -356,6 +356,18 @@ public static IProjection Conditional(ICriterion criterion, IProjection whenTrue
...
 		}
 
+		/// <summary>
+		/// Conditionally returns one of the <see cref="ConditionalProjectionCase.Projection"/>s depending on the <see cref="ConditionalProjectionCase.Criterion"/>s of <paramref name="cases"/> or the <paramref name="elseProjection"/>.
+		/// This produces an switch-case expression with multiple when-then parts.
+		/// </summary>
+		/// <param name="cases">The <see cref="ConditionalProjectionCase"/>s which contain your <see cref="ICriterion"/>s and <see cref="IProjection"/>s.</param>
+		/// <param name="elseProjection">The else <see cref="IProjection"/>.</param>
+		/// <returns>A <see cref="IProjection"/> for a switch-expression with multiple <see cref="ICriterion">Criterions</see> ("when") <see cref="IProjection">Projections</see> ("then").</returns>
+		public static IProjection Conditional(ConditionalProjectionCase[] cases, IProjection elseProjection)
```
bahusoid, line 366: Maybe better make it `IList<ConditionalProjectionCase> cases`?

bahusoid (approved):

src/NHibernate/Criterion/Projections.cs
perahoky: Hi, yes i would agree but it should be an array due to performance. You could pass an array to IList (the interface doesn't really change the performance), but that would move the control of the performance to the caller: they need to "know to pass an array" if they care about performance (and we do, we pass 100+ cases multiple times a second).
Yeah it's not much, but if we have the choice?

bahusoid: I don't get your point about `know to pass an array`. Why  `List<ConditionalProjectionCase>` or any other collection that implements `IList<>` would perform worse than array?

hazzik: I'll just leave this as is.

perahoky: > 
> 
> I don't get your point about `know to pass an array`. Why `List<ConditionalProjectionCase>` or any other collection that implements `IList<>` would perform worse than array?

Just to conclude my point: :) 
Using IList<> enables you to deliver every better or worser IList<>-type like "slower" List<> instead of an array.
The current implemention forces you to deliver an array - which is much faster to iterate than a - probably - used List<>.
A developer may not know that the implementation's performance relies on the use of thhe delivered real type (List<>, Array[], ...) for the IList<>.
And as we don't change the array, we don't need IList<> which only enables you to add something (what we don't do).
That's my reason to force everyone to use array.
We may discuss whether to use IReadOnlyCollection https://docs.microsoft.com/en-us/dotnet/api/system.array.asreadonly?view=net-5.0 
but that enables you to use an List<> again - which is a bit slower. It's not much difference, but we take every bit performance we can get.
Or we create a second cctor with IList<> as "a backdoor" for using other types - if its really that neccessary - and write a note to documentation "performance depends on real used type" or something like that.

Have a nice day and thank you all for accepting :) 
Paul

PR #2841, opened 2021-06-30
Fix possible test failure for SqlServer 2019

bahusoid:
Apparently latest SQL Server 2019 has enabled "verbose truncation warning" by default. Which gives more detailed exception and changes error code: 
"2628 – String or binary data would be truncated in table ‘[TableName]’, column ‘[ColumnName]’. Truncated value: ‘[Value]’."
vs old
"8152 -String or binary data would be truncated"


fredericDelaporte (approved):
Since long, I had this failure on my own setup, but never took the time to check what was going on.

PR #2844, opened 2021-07-04
Merge 5.3.9

fredericDelaporte:
I intend to push the merge directly to master once this PR is validated. (It would require that nothing else is merged before.)


fredericDelaporte:
Maybe I should keep MySql failures allowed in this merge, and let its troubles be fixed in a dedicated PR.

bahusoid:
> Maybe I should keep MySql failures allowed in this merge

I already squashed this fix in merge commit.. So should I force push or will fix it separately?

fredericDelaporte:
Ok, force push.

hazzik (approved):

Issue #2872, opened 2021-07-21
LINQ Query with .Where(f => names.Where(d => d != null).Contains(f.Name)) throws InvalidCastException starting version 5.3.2


haefele:
Hello, I noticed another regression in the newer NHibernate versions.

We have quite a lot of places where some query-parameter list is filtered inside the .Where Expression like this:
```csharp
var dave = session.Query<Person>()
                  .Where(f => names.Where(d => d != null).Contains(f.Name))
                  .FirstOrDefault();
```

Focus on the `names.Where(d => d != null)` part.

This used to work without any issues up until NHibernate version `5.2.7`, but started returning `NULL` with version `5.3.0`, and now throws a `InvalidCastException` starting from version `5.3.2` up to the current version `5.3.9`

<details>
  <summary>Full Exception with StackTrace</summary>

```
NHibernate.Exceptions.GenericADOException: could not execute query
[ select person0_.Id as id1_0_, person0_.Name as name2_0_ from dbo.[Person] person0_ where person0_.Name in (?) ORDER BY CURRENT_TIMESTAMP OFFSET 0 ROWS FETCH FIRST 1 ROWS ONLY ]
  Name:p1 - Value:System.Collections.Generic.List`1[System.String]
[SQL: select person0_.Id as id1_0_, person0_.Name as name2_0_ from dbo.[Person] person0_ where person0_.Name in (?) ORDER BY CURRENT_TIMESTAMP OFFSET 0 ROWS FETCH FIRST 1 ROWS ONLY] ---> System.InvalidCastException: Das Objekt des Typs "System.Collections.Generic.List`1[System.String]" kann nicht in Typ "System.String" umgewandelt werden.
   bei NHibernate.Type.AbstractStringType.Set(DbCommand cmd, Object value, Int32 index, ISessionImplementor session)
   bei NHibernate.Type.NullableType.NullSafeSet(DbCommand st, Object value, Int32 index, ISessionImplementor session)
   bei NHibernate.Param.NamedParameterSpecification.Bind(DbCommand command, IList`1 multiSqlQueryParametersList, Int32 singleSqlParametersOffset, IList`1 sqlQueryParametersList, QueryParameters queryParameters, ISessionImplementor session)
   bei NHibernate.Param.NamedParameterSpecification.Bind(DbCommand command, IList`1 sqlQueryParametersList, QueryParameters queryParameters, ISessionImplementor session)
[...]

PR #2873, opened 2021-07-22
WIP Evaluate parameter expression for Enumerable.Contains calls

bahusoid:


bahusoid:
It seems does the job but I wonder how to make it more optimal. Now expressions evaluation  are duplicated  inside `NhPartialEvaluatingExpressionVisitor.TryGetCollectionParameter` and as part of common parsing inside `NhPartialEvaluatingExpressionVisitor.Visit(Expression expression)`.

bahusoid:
> Now expressions evaluation are duplicated

Seems to be fixed by https://github.com/nhibernate/nhibernate-core/pull/2873/commits/15381b0a47c42f23de76b9296c303f5b7974d433

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -155,6 +158,42 @@ private Expression EvaluateSubtree(Expression subtree)
...
+		{
+			if (expression.Method.DeclaringType != typeof(Enumerable))
+				return IsCollectionParameter(expression);
+
+			var arg = expression.Arguments[0];
+
+			if (IsCollectionParameter(arg))
+				return true;
+
+			return arg.NodeType == ExpressionType.Call && TryGetCollectionParameter((MethodCallExpression) arg);
```
hazzik, line 189: Why do we need to do that recursively?

bahusoid: To support chain of LINQ commands. And to make sure it doesn't start with mapped association (somethin like
`db.Orderes.Where( x => x.OrderLines.Select(...).Where(...).Contains(...))` should not be evaluated)

hazzik:
Imho, there is a bug in `EvaluatableTreeFindingExpressionVisitor.Analyze` which marks any expression containing lambda as not evaluatable. This happens because any lambda has parameters and parameters are not evaluatable. I believe, the `EvaluatableTreeFindingExpressionVisitor` should visit lambdas to evaluate the subtrees, but the lambdas itself should not affect the expressions containing those lambdas.

I made the tests working by copying EvaluatableTreeFindingExpressionVisitor to our codebase and modifying `VisitLambda` as following:

		protected override Expression VisitLambda<T>(Expression<T> expression)
		{
			if (expression == null) throw new ArgumentNullException("expression");

			var parent = _isCurrentSubtreeEvaluatable;
			_isCurrentSubtreeEvaluatable = true;

			var vistedExpression = base.VisitLambda(expression);

			_isCurrentSubtreeEvaluatable = parent;

			// Testing the parent expression is only required if all children are evaluatable
			if (_isCurrentSubtreeEvaluatable)
				_isCurrentSubtreeEvaluatable = _evaluatableExpressionFilter.IsEvaluatableLambda(expression);

			return vistedExpression;
		}

There are some side effects (like lambdas are marked as evaluatable which it should not do, but it can be fixed further down the track)

hazzik:
In other words the @v-kabanov's solution is more correct.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
```diff
@@ -155,6 +158,42 @@ private Expression EvaluateSubtree(Expression subtree)
...
+		private void DetectEvaluatableExpressionOnCollectionContains(MethodCallExpression expression)
+		{
+			if (!expression.Method.IsGenericMethod || ContainsMethodInfo != expression.Method.GetGenericMethodDefinition())
+				return;
+			var argument = expression.Arguments[0];
+			if (argument.NodeType != ExpressionType.Call)
+				return;
+
+			if(TryGetCollectionParameter((MethodCallExpression)argument))
+				_partialEvaluationInfo.AddEvaluatableExpression(argument);
```
hazzik, line 176: Here you're modifying external dependency, which you should not.

bahusoid: I don't get it. Why I shouldn't touch it? It's just a wrapper around set of expressions that needs evaluation.  I can store it in separate field (as I did initially) - but to minimize change noise I used already present class for storing evaluatable expressions.

hazzik:
With the proposed implementation following expression would be considered as evaluatable:

```
db.Orders.Where(o => ids.Where(i => i == o.OrderId).Contains(o.OrderId))
```

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
hazzik: > I don't get it. Why I shouldn't touch it?

Because this is the information provided by another concerning party. We already let this logic spread across solution that we trying to catch bugs doing 10th bugfix release.

bahusoid:
> With the proposed implementation following expression would be considered as evaluatable:

Yeah. I know.. But it's not supported anyway. It seems we don't support much with collection parameters except `Contains` call

bahusoid:
> In other words the @v-kabanov's solution is more correct.

I agree. Though I don't like including another bunch of slightly modified files from Re-Linq. IMO better local small targeted fix. We can drop it if his Relinq PR is accepted https://github.com/re-motion/Relinq/pull/14

> There are some side effects (like lambdas are marked as evaluatable which it should not do, but it can be fixed further down the track)

Sounds like it can cause more regressions (at least more unnecessary attempts to compile) than my fix

bahusoid:
>With the proposed implementation following expression would be considered as evaluatable:
```C#
db.Orders.Where(o => ids.Where(i => i == o.OrderId).Contains(o.OrderId))
```

Old behavior: 
v5.2: Where part is silently ignored;
v5.3: Where part is ignored and throws InvalidCastException when setting parameter (see #2872) ;

New behavior: Exception is thrown: `Evaluation failure on value(NHibernate.DomainModel.Northwind.Entities.Order[]).Where(i => (i == o.OrderId))` So I think it's an improvement anyway as it shows exact place of failure; So it's clear which part of query is not supported.

src/NHibernate/Linq/Visitors/NhPartialEvaluatingExpressionVisitor.cs
bahusoid: Well.. Ok... It can be stored separately. But I still think it's more proper to store it in `_partialEvaluationInfo` - external or not, it stores list of expressions that needs evaluation. So it's better to fill it also with our expressions - so all other external parties can handle it appropriately as evaluatable expression.

But before making further changes -  do you agree with this fix on principle?

bahusoid:
> Imho, there is a bug in EvaluatableTreeFindingExpressionVisitor.Analyze which marks any expression containing lambda as not evaluatable. 

https://re-motion.atlassian.net/browse/RMLNQ-96  It looks like it's known limitation since 2016.

And here is author's  thoughts about  @v-kabanov's fix:  https://re-motion.atlassian.net/browse/RMLNQ-127?focusedCommentId=21851

hazzik:
> But before making any further changes - do you agree with this fix on principle?

After sleeping on the issue overnight, I think we should not fix it. Not at 5.3.x at least. 

I would prefer a fix of RMLNQ-96 in re-linq itself.

bahusoid:
Ok. I rebased it on top of master. 
I would prefer to fix it in 5.4 one way or another. So let it hang till we decide what to do with Re-Linq

PR #2877, opened 2021-07-27
Fix property type retrieval for Composite User Type in LINQ

### LINKED ISSUE #2856: Distinct on Composite User Type property fails
This is a regression introduced in v5.3.0

If I define an `ICompositeUserType` that directly maps properties to columns, the LINQ query should act on the mapped type's columns.

Say I define `PhoneNumber` with `Number` and `Ext` properties and map them to separate columns.

Prior to v5.3.0 the following worked
```
using (var s = OpenSession())
{
    s.Save(new Entity { 
        Name = "Company",
        Phone = new PhoneNumber("745-555-1234"),
    });
    s.Save(new Entity
    {
        Name = "Bob",
        Phone = new PhoneNumber("745-555-1234", "x123"),
    });
    s.Save(new Entity
    {
        Name = "Jane",
        Phone = new PhoneNumber("745-555-1235"),
    });
    s.Flush();
}

using (var s = OpenSession())
{
    var numbers = s.Query<Entity>().Select(x => x.Phone.Number).Distinct().ToList();
    Assert.That(numbers.Count, Is.EqualTo(2));
}
```

After v5.3.0 the Count returned is 3.

Inspecting the SQL generated reveals that both columns of the composite type are being included in the generated SQL rather than the single selected column and thus the `DISTINCT` clause is returning unexpected results.

bahusoid:
Fixes #2856 
Regression from https://github.com/nhibernate/nhibernate-core/pull/2036 (most likely)


src/NHibernate/Util/ExpressionsHelper.cs
```diff
@@ -360,25 +360,7 @@ internal static IType GetType(VisitorParameters parameters, Expression expressio
...
 					case IAbstractComponentType componentType:
 						currentComponentType = componentType;
-						if (currentEntityPersister == null)
-						{
-							// When persister is not available (q.OneToManyCompositeElement[0].Prop), try to get the type from the component
-							currentType = TryGetComponentPropertyType(componentType, member.Path);
-						}
-						else
-						{
-							// Concatenate the component property path in order to be able to use EntityMetamodel.GetPropertyType to retrieve the type.
```
bahusoid, line 370: EntityMetamodel also can't handle ICustomUserType components. So why bother to use it at all if we can retrieve all info from `componentType`

bahusoid: > Concatenate the component property path in order to be able to use EntityMetamodel.GetPropertyType

From original comments it looks like it's some internal calculations specifically for EntityMetamodel call. But this calculated path is returned in out param `memberPath`.

@maca88 Are you sure this concatenation is not required for the branch "When persister is not available".
You have it covered in the following test:
https://github.com/nhibernate/nhibernate-core/blob/a97627d8f1b06388aa68d4b0c5ddf386a0a1781b/src/NHibernate.Test/Linq/TryGetMappedTests.cs#L461-L466

Where you expect `memberPath` for `Select(o => o.Components[0].Subcomponent.Name)` to be `Name`. But why? Shouldn't it be `Subcomonent.Name` instead?

Also should such query `session.Query<Baz>().Select(o => o.Components[0].Subcomponent.Name).ToList()` work at all? Currently it fails to parse generated hql...

bahusoid: > But why? Shouldn't it be Subcomonent.Name instead?

I made a quick look and it seems not important. Both cases are handled.

hazzik (approved):

PR #2889, opened 2021-08-09
Fix nullable entity comparison with null

### LINKED ISSUE #2611: One-to-zero-or-one relation not returning data when checking for null

bahusoid:
Fixes #2611

### DIFF
--- a/src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
@@ -417,8 +417,7 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 
 			if ( joinIsNeeded )
 			{
-				//TODO: Find correct place for this if.
-				if (Walker.IsNullComparison)
+				if (comparisonWithNullableEntity && Walker.IsNullComparison)
 				{
 					implicitJoin = false;
 					_joinType = JoinType.LeftOuterJoin;
--- a/src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
@@ -389,9 +389,12 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 			bool joinIsNeeded;
 
 			//For nullable entity comparisons we always need to add join (like not constrained one-to-one or not-found ignore associations)
-			//NOTE: This fix is not fully correct. It doesn't work for comparisons with null (where e.OneToOneProp is null)
-			// as by default implicit join is generated and to work propelry left join is required (see GH-2611)
-			bool comparisonWithNullableEntity = entityType.IsNullable && Walker.IsComparativeExpressionClause;
+			bool comparisonWithNullableEntity = entityType.IsNullable && Walker.IsComparativeExpressionClause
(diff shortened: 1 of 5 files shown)


fredericDelaporte:
So this would somehow fix #2611 by reverting to pre #2081 behavior for a special and narrow case?

bahusoid:
>  this would somehow fix #2611

No as one-to-one case remains completely broken but with different results as Id of the owner is used in this case. By example:
```C#
s.Query<Parent>().Where(p => p.OneToOne == null);
```

5.2 (and 5.3.10 with this PR applied):
```sql
select * from Parent p where p.Id is null --Returns all parent records
```

5.3.9 :
```sql
select * from Parent p, OneToOne  o where p.Id = o.Id and o.Id is null -- Returns zero records
```

hazzik:
Do I understand correctly that the correct SQL would be like this?

```sql
select * from Parent p left join OneToOne  o on p.Id = o.Id where o.Id is null -- Returns zero records
```

bahusoid:
Yes. Left join is needed instead of implicit

hazzik:
image

I made it working. Need to resolve TODO. Feel free to drop the commit if you don't like it.

src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
```diff
@@ -417,8 +415,15 @@ private void DereferenceEntity(EntityType entityType, bool implicitJoin, string
 				               || comparisonWithNullableEntity;
 			}
 
-			if ( joinIsNeeded ) 
+			if ( joinIsNeeded )
 			{
+				//TODO: Find correct place for this if.
+				if (Walker.IsNullComparison)
```
hazzik, line 421: Need to find correct place for this check. It probably needs to be done somewhere earlier. I don't have time right now, could look later, but not promising it.

bahusoid:
> I made it working.

I made similar changes in https://github.com/nhibernate/nhibernate-core/pull/2612. Only then I tried to make it more general fix not bound to is null comparison. Yeah I also realized that merging both PRs might be a good idea..

hazzik (approved):

hazzik (approved):

PR #2916, opened 2021-09-26
Merge 5.3.10

fredericDelaporte:
I intend to push directly, but since there were some conflicts, I still ask for a review.


teamcity.build
```diff
@@ -207,7 +207,7 @@
 		<property name="db-service" value="MySQL57" />
 		<property name="nhibernate.connection.driver_class" value="NHibernate.Driver.MySqlDataDriver" />
 		<property name="nhibernate.dialect"	value="NHibernate.Dialect.MySQL5Dialect" />
-		<property name="nhibernate.connection.connection_string" value="Data Source=localhost;Database=nhibernate;User ID=nhibernate;Password=nhibernate;Old Guids=True;" />
+		<property name="nhibernate.connection.connection_string" value="Data Source=localhost;Database=nhibernate;User ID=nhibernate;Password=nhibernate;Old Guids=True;SslMode=none;" />
```
fredericDelaporte, line 210: Conflict here between removing `memory` communication on master due to a bug supposed to be fixed in current MySql provider (which we are not yet using) and adding ssl deactivation on 5.3.x.

bahusoid: We don't need SSL deactivation on master (See https://github.com/nhibernate/nhibernate-core/pull/2870). So I think it's better to simply skip this change.

hazzik (approved):

PR #2921, opened 2021-10-04
Fix test for SAP SQL Anywhere

bahusoid:


src/NHibernate.Test/Linq/LinqQuerySamples.cs
```diff
@@ -2001,9 +2001,12 @@ public void ReplaceFunctionWithNullArgument()
 		[Test(Description = "GH-2860")]
 		public void StringFormatWithTrim()
 		{
+			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
```
hazzik, line 2004: It is interesting how we fix it for Firebird by wrapping parameters into a cast expression, and don’t do same for other DBs.

hazzik (approved):

PR #2944, opened 2021-11-17
Add monitor based sync only locker for ReadWrite cache

### LINKED ISSUE #2862: NHibernate AsyncReaderWriterLock stalls under load
We've been investigating our process stalling under load, multiple crash dumps show a similar set of stack traces to this:
image

After some testing I put together a crude harness over NHibernate.Util.AsyncReadWriteLock which reproduces the problem - run this on .NET Framework 4.8 and the unit test doesn't complete after an hour, switch to .NET Core 3.0 and it runs < 1 minute.

```
  [TestMethod]
        public Task LockHammer53()
        {
            var asyncReaderWriterLock = new AsyncReaderWriterLock();
          
            return Task.WhenAll(Enumerable.Range(1, 100).Select(x => Task.Run(() =>
              {
                  for (var i = 0; i < 1000000; i++)
                  {
                      using (asyncReaderWriterLock.WriteLock())
                      {
                      }
                  }
              })));
        }
```

It looks like it might be related to this bug in .NET 
https://github.com/dotnet/runtime/issues/28717
It was fixed in .NET Core 3.0 but remains unfixed in .NET Framework and it looks like they have no plans to fix it.

Interestingly StackExchange.Redis was majorly impacted by this and worked around this issue by switching from SemaphoreSlim to MutexSlim
https://github.com/StackExchange/StackExchange.Redis/commit/687394109f9b7faa6a25851b3744417bfae5cd51

So I would propose either:
a) Reverting the locking to the design in 5.2
b) Rewrite the AsyncReaderWriterLock to avoid the bug in SemaphoreSlim

bahusoid:
Fixes #2862 

Adds ability to use old `lock` based locking for sync only scenarios. Reason: As described [here](https://github.com/nhibernate/nhibernate-core/issues/2862#issuecomment-898876785) in  full .NET Framework current locking code is 300 times slower.

Throws `InvalidOperationException` when used in async code.

Usage:
```xml
<session-factory>
  <property name="cache.read_write_lock_factory">sync</property>
</session-factory>
```

Or by code:
```C#
Configuration cfg = ...;
cfg.SetProperty(Environment.CacheReadWriteLockFactory, "sync");
```


src/NHibernate/Cache/ICacheLock.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.Threading.Tasks;
+using NHibernate.Util;
+
+namespace NHibernate.Cache
+{
+	public interface ICacheLock : IDisposable
```
fredericDelaporte, line 7:
```suggestion
	/// <summary>
	/// Implementors provide a locking mechanism for the cache.
	/// </summary>
	public interface ICacheLock : IDisposable
```

src/NHibernate/Cache/ICacheLock.cs
```diff
@@ -0,0 +1,35 @@
...
+	{
+		IDisposable ReadLock();
+		IDisposable WriteLock();
+		Task<IDisposable> ReadLockAsync();
+		Task<IDisposable> WriteLockAsync();
```
fredericDelaporte, line 12:
```suggestion
		/// <summary>
		/// Acquire synchronously a read lock.
		/// </summary>
		/// <returns>A read lock.</returns>
		IDisposable ReadLock();

		/// <summary>
		/// Acquire synchronously a write lock.
		/// </summary>
		/// <returns>A write lock.</returns>
		IDisposable WriteLock();
		

		/// <summary>
		/// Acquire asynchronously a read lock.
		/// </summary>
		/// <returns>A read lock.</returns>
		Task<IDisposable> ReadLockAsync();

		/// <summary>
		/// Acquire asynchronously a write lock.
		/// </summary>
		/// <returns>A write lock.</returns>
		Task<IDisposable> WriteLockAsync();
```

src/NHibernate/Cache/ICacheLock.cs
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	public interface ICacheReadWriteLockFactory
```
fredericDelaporte, line 15:
```suggestion
	/// <summary>
	/// Define a factory for cache locks.
	/// </summary>
	public interface ICacheReadWriteLockFactory
```

src/NHibernate/Cache/ICacheLock.cs
```diff
@@ -0,0 +1,35 @@
...
+	{
+		ICacheLock Create();
```
fredericDelaporte, line 17:
```suggestion
		/// <summary>
		/// Create a cache lock provider.
		/// </summary>
		ICacheLock Create();
```

src/NHibernate/Cache/ICacheLock.cs
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	class AsyncCacheReadWriteLockFactory : ICacheReadWriteLockFactory
```
fredericDelaporte, line 20:
```suggestion
	internal class AsyncCacheReadWriteLockFactory : ICacheReadWriteLockFactory
```

src/NHibernate/Cache/ICacheLock.cs
```diff
@@ -0,0 +1,35 @@
...
+	{
+		public ICacheLock Create()
+		{
+			return new AsyncReaderWriterLock();
+		}
+	}
+
+	class SyncCacheReadWriteLockFactory : ICacheReadWriteLockFactory
```
fredericDelaporte, line 28:
```suggestion
	internal class SyncCacheReadWriteLockFactory : ICacheReadWriteLockFactory
```

src/NHibernate/nhibernate-configuration.xsd
```diff
@@ -103,6 +103,7 @@
 										<xs:enumeration value="cache.provider_class" />
 										<xs:enumeration value="cache.use_query_cache" />
 										<xs:enumeration value="cache.query_cache_factory" />
+										<xs:enumeration value="cache.read_write_lock_factory" />
```
fredericDelaporte, line 106:
```suggestion
										<xs:enumeration value="cache.read_write_lock_factory">
											<xs:annotation>
												<xs:documentation>
													Specify the cache lock factory to use for read-write cache regions.
													Defaults to the built-in async cache lock factory.
													Use async, or sync, or classname.of.CacheLockFactory, assembly.
												</xs:documentation>
											</xs:annotation>
										</xs:enumeration>
```

fredericDelaporte (commented):
May you also insert a documentation entry in the below file after the query cache factory setting?
https://github.com/nhibernate/nhibernate-core/blob/92075ee3e0202b797c5d51d9fe23a7f6d0f3e390/doc/reference/modules/configuration.xml#L600-L612
```xml
                    <row>
                        <entry>
                            <literal>cache.read_write_lock_factory</literal>
                        </entry>
                        <entry>
                            Specify the cache lock factory to use for read-write cache regions.
                            Defaults to the built-in <literal>async</literal> cache lock factory.
                            <para>
                                <emphasis role="strong">eg.</emphasis>
                                <literal>async</literal>, or <literal>sync</literal>, or <literal>classname.of.CacheLockFactory, assembly</literal>
                            </para>
                            <para>
                                <literal>async</literal> uses a single writer multiple readers locking mechanism supporting asynchronous operations.
                            </para>
                            <para>
                                <literal>sync</literal> uses a single access locking mechanism which will throw on asynchronous
                                operations but may have better performances than the <literal>async</literal> provider for applications using
                                the .Net Framework (4.8 and below).
                            </para>
                        </entry>
                    </row>
```

bahusoid:
> May you also insert a documentation entry

Done

src/NHibernate.Test/CacheTest/SyncOnlyCacheFixture.cs
```diff
@@ -0,0 +1,180 @@
...
+using NHibernate.Cache;
+using NHibernate.Cache.Access;
+using NHibernate.Cfg;
+using NUnit.Framework;
+using Environment = NHibernate.Cfg.Environment;
+
+namespace NHibernate.Test.CacheTest
+{
+	[TestFixture]
+	public class SyncOnlyCacheFixture : TestCase
```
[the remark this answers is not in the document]
bahusoid: Done

fredericDelaporte (approved):

src/NHibernate/Cfg/Settings.cs
```diff
@@ -101,6 +101,7 @@ public Settings()
 		public ConnectionReleaseMode ConnectionReleaseMode { get; internal set; }
 
 		public ICacheProvider CacheProvider { get; internal set; }
+		public ICacheReadWriteLockFactory CacheReadWriteLockFactory { get; internal set; }
```
hazzik, line 104: Please add new line before the property.

src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -36,7 +36,17 @@ public interface ILockable
 
 		private CacheBase _cache;
 		private int _nextLockId;
-		private readonly AsyncReaderWriterLock _asyncReaderWriterLock = new AsyncReaderWriterLock();
+		private readonly ICacheLock _asyncReaderWriterLock;
+
+		public ReadWriteCache()
+		{
+			_asyncReaderWriterLock = new AsyncReaderWriterLock();
+		}
```
hazzik, line 44:
```suggestion
		public ReadWriteCache(ICacheLock locker) : this(new AsyncReaderWriterLock()) 
		{
		}
```

PR #2945, opened 2021-11-19
Fix distinct paging queries for MSSql2012Dialect

### LINKED ISSUE #1169: NH-3594 - MSSql2012Dialect generates invalid order by clause for paging distinct query
**Gerke Geurts** created an [issue](https://nhibernate.jira.com/browse/NH-3594) — <time datetime="2014-01-24T16:13:56.855Z" title="2014-01-24T16:13:56.855Z">24th January 2014, 16:13:56</time>:
> The SQL Server 2012 dialect generates invalid SQL when paging or limits are used on SELECT DISTINCT queries. The NHibernate.Test.NHSpecificTest.NH2214.Fixture.ShouldWorkUsingDistinctAndLimits() results in the following exception:
> 
> NHibernate.Exceptions.GenericADOException : could not execute query
> < select distinct person0*.Id as Id5061_, person0_.FirstName as FirstName5061_ from Person person0_ ORDER BY CURRENT*TIMESTAMP OFFSET 0 ROWS FETCH FIRST @p0 ROWS ONLY >
> [SQL: select distinct person0*.Id as Id5061_, person0_.FirstName as FirstName5061_ from Person person0_ ORDER BY CURRENT*TIMESTAMP OFFSET 0 ROWS FETCH FIRST @p0 ROWS ONLY]
>   ----> System.Data.SqlClient.SqlException : ORDER BY items must appear in the select list if SELECT DISTINCT is specified.
> 
---
**Gerke Geurts** added a comment — <time datetime="2014-01-24T16:27:19.560Z" title="2014-01-24T16:27:19.560Z">24th January 2014, 16:27:19</time>:
> Fix proposed in https://github.com/nhibernate/nhibernate-core/pull/251
> 
---
**Oskar Berggren** added a comment — <time datetime="2014-07-26T10:35:00.740Z" title="2014-07-26T10:35:00.740Z">26th July 2014, 10:35:00</time>:
> Merged in 8531dafd9e42fc4de9b00dc27934038520e28112.
> 
---
**Oskar Berggren** added a comment — <time datetime="2014-07-26T13:08:14.914Z" title="2014-07-26T13:08:14.914Z">26th July 2014, 13:08:14</time>:
> Merge reverted due to many new test failures for e.g. sql server 2008.
>

bahusoid:
Fixes #1169

Simply fallback to old logic when new one doesn't work.


hazzik:
Why not just `ORDER BY 1` and fix tests?

bahusoid:
> Why not just ORDER BY 1

I don't like it. How it's better? It looks hackish to me...

src/NHibernate/Dialect/MsSql2012Dialect.cs
```diff
@@ -58,25 +58,25 @@ protected override void RegisterFunctions()
...
+
+				var hasOrderBy = tokenEnum.TryParseUntil("order");
+				if(!hasOrderBy && isDistinct)
+						return base.GetLimitString(querySqlString, offset, limit);
 
 				var result = new SqlStringBuilder(querySqlString);
-				if (!tokenEnum.TryParseUntil("order"))
+				if (!hasOrderBy)
 				{
 					result.Add(" ORDER BY CURRENT_TIMESTAMP");
```
fredericDelaporte, line 75: Ordering by the first column of the result set seems less hackish to me than ordering by the current date and time. This thing attempts to emulate the behavior of unordered query on SQL-Server. I do not think NHibernate should try to do that, since NHibernate is more about abstracting away the specificity of database vendors rather than surfacing them.

I am tempted to say we should not even try to support unordered paged queries, because conceptually, writing such queries is a mistake: there is no predictability about which data will be sent back. There is not even any guarantee the paging will be consistent: it may include the same row in different pages, and not yield some rows whatever the asked page. (With SQL Server, it depends on the indexes used in the computed execution plan. If for some reasons it uses another plan, the ordering yielded by an unordered query will change.)
But removing all support for it will undoubtedly break a lot of existing application, so we have to keep that legacy.

So, replacing `CURRENT_TIMESTAMP` by `1` for this case looks better to me.

Using a fallback on the row_number paging seems to me as trying to somewhat preserve the default behavior of SQL-Server with unordered queries, which I do not think is valuable for NHibernate.

bahusoid: > Ordering by the first column of the result set seems less hackish

IMO it's better to use existing behavior (hackish or not) than introduce another dialect specific hack (even less hackish) with different behavior.

hazzik: > I am tempted to say we should not even try to support unordered paged queries, because conceptually, writing such queries is a mistake

I fully agree. 

I think the user should see an error and then try to fix their query.

PR #2951, opened 2021-12-01
Run tests on .NET 6

hazzik:


.github/workflows/NetCoreTests.yml
```diff
@@ -54,9 +54,9 @@ jobs:
 
     - uses: actions/checkout@v2
     - name: Setup .NET
-      uses: actions/setup-dotnet@v1.8.0
+      uses: actions/setup-dotnet@v1
```
bahusoid, line 57: Downgrade? Why? Latest now 1.9

ReleaseProcedure.txt
```diff
@@ -58,6 +58,6 @@ Tools\gitreleasemanager\0.11.0\tools\GitReleaseManager.exe create -o nhibernate
   * If this was the master branch, create the stable branch for the new release, named
     according to its version with "x" instead of its patch number. By example, "5.3.x".
     
-  * If next version is decided set it in NHibernate.props under build-common folder
+  * If next version is decided set it in Directory.Build.props under build-common folder
```
bahusoid, line 61: under build-common -> under src

src/Directory.Build.props
```diff
@@ -13,18 +13,25 @@
...
     <FileVersion Condition="'$(FileVersion)' == ''">$(VersionPrefix).0</FileVersion>
 
-    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;netcoreapp2.0</NhAppTargetFrameworks>
+    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;net6.0</NhAppTargetFrameworks>
     <NhLibTargetFrameworks Condition ="$(NhLibTargetFrameworks) == ''">net461;netcoreapp2.0;netstandard2.0</NhLibTargetFrameworks>
     <NhNetFx>false</NhNetFx>
     <NhNetFx Condition="$(TargetFramework.StartsWith('net4'))">true</NhNetFx>
-    <DefineConstants Condition="$(NhNetFx) AND $(NhVbNet) == ''" >NETFX;$(DefineConstants)</DefineConstants>
+    <DefineConstants Condition="$(NhNetFx) AND '$(MSBuildProjectExtension)' != '.vbproj'">NETFX;$(DefineConstants)</DefineConstants>
+    <DefineConstants Condition="$(NhNetFx) AND '$(MSBuildProjectExtension)' == '.vbproj'">NETFX,$(DefineConstants)</DefineConstants>
```
bahusoid, line 21: Can you please add comment like `Use coma separation for VB projects`. As it took me sometime to grasp what's going on...

src/NHibernate.Test.VisualBasic/Program.vb
```diff
@@ -1,4 +1,4 @@
-﻿#If NETCOREAPP2_0
+﻿#If NETCOREAPP2_0_OR_GREATER
```
bahusoid, line 1: nit: I don't like it. Why not simply NETCORE?

src/NHibernate.Test/Async/StaticProxyTest/InterfaceHandling/Fixture.cs
```diff
@@ -179,23 +179,23 @@ public async Task ProxyInterfaceIdAccessFromDifferentInterfacesAsync()
 
 		private void ThrowOnIEntityNameAccess(IEntity entity)
 		{
-			Assert.That(() => entity.Name, Throws.TypeOf<ObjectNotFoundException>(), "IEntity.Name access should lead to proxy initialization");
+			Assert.That<string>(() => entity.Name, Throws.TypeOf<ObjectNotFoundException>(), "IEntity.Name access should lead to proxy initialization");
```
bahusoid, line 182: Why such changes are necessary?

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -1,13 +1,12 @@
...
   <PropertyGroup>
     <Description>The Unit Tests for NHibernate.</Description>
     <TargetFrameworks>$(NhAppTargetFrameworks)</TargetFrameworks>
     <IsTestProject>true</IsTestProject>
-    <NoWarn>$(NoWarn);3001;3002;3003;3005</NoWarn>
+    <NoWarn>$(NoWarn);3001;3002;3003;3005;SYSLIB0003;SYSLIB0011;SYSLIB0012</NoWarn>
     <ContentSQLiteInteropFiles>true</ContentSQLiteInteropFiles>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
```
bahusoid, line 9: Why not introduce global `NhNetCore` property and use it for all such places instead to simplify further retargeting?

src/Directory.Build.props
```diff
@@ -13,18 +13,25 @@
     <FileVersion Condition="'$(VersionSuffix)' != '' AND '$(BuildNumber)' != ''">$(VersionPrefix).$(BuildNumber)</FileVersion>
```
bahusoid, line 13: https://github.com/nhibernate/nhibernate-core/blob/master/build-common/NHibernate.props#L2

What about NHibernate.dev.props? Should it also be renamed to Directory.Build.dev.props?

.github/workflows/NetCoreTests.yml
hazzik: > Downgrade? 

No. It takes latest 1.x this way.

ReleaseProcedure.txt
hazzik: Thanks

src/Directory.Build.props
hazzik: Sure.

src/NHibernate.Test.VisualBasic/Program.vb
hazzik: Because `NETCOREAPP` could mean `NETCOREAPP1_0`. These are standard symbols. I've also decided not to change our custom defined `NETFX` to standard `NETFRAMEWORK` or `NET461_OR_GREATER`

src/NHibernate.Test/Async/StaticProxyTest/InterfaceHandling/Fixture.cs
hazzik: Because .NET 6 broke [NUnit](https://github.com/nunit/nunit/issues/3993).

src/Directory.Build.props
hazzik: I don't know.

src/NHibernate.Test/NHibernate.Test.csproj
hazzik: Because this is not so easy... I tried it when targeting both .NET Core 3.1 and .NET 6... The dependencies could/were a little different. For example, `Microsoft.AspNetCore.OData` does not have problems on 3.1 but has problems in 6.0.

src/Directory.Build.props
hazzik: Ok, renamed

src/NHibernate.Test/Async/StaticProxyTest/InterfaceHandling/Fixture.cs
hazzik: Probably need to wait next 6.0.x SDK version before merging.

src/NHibernate.Test.VisualBasic/Program.vb
bahusoid: Oh I see. Though I still think using simple NETCORE would minimize the frictions in the future..

src/NHibernate.Test/Async/StaticProxyTest/InterfaceHandling/Fixture.cs
bahusoid: Yeah.. Quite annoying bug...

src/NHibernate.Test/NHibernate.Test.csproj
bahusoid: Ehm...  I want to convert checks like `<PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">` to something like `<PropertyGroup Condition="NhNetCore">`.  Looks easy enough for me.. Not sure I understand your problem - are you talking about targeting both 3.1 and 6.0? IMO that's a bit different issue and should be handled separately

hazzik: > Looks easy enough

Yes, it is easy, but I doubt it will solve any real issue.

> are you talking about targeting both 3.1 and 6.0?

I was going to make it target .NET 3.1 and .NET 6 rc initially. I started working on the switch around last betas and before .NET 6 rc.1 came out. But I did not have enough time and motivation to finish that at that time. So because of differences in the supported features (like SYSLIB0011 - BinaryFormatter depreciation which is quite aggressive) and different dependencies between .NET 3.1 and .NET 6.0 it required having 2 different monikers like NhNetCore31 and NhNet60. Which defied the purpose of the simplification. 

Also, the same could happen when we transition from .NET 6 to next version. Depends on the path we're willing to take.  

So, frankly, I do not believe that a simple bump of a version in NhNetCore would be enough to migrate to the later versions. Seems like we're trying to solve the problem which we might not have. 

For NhNet it does have sense to use it because Microsoft has promised that 4.x API would stay backwards compatible.

src/NHibernate.Test.VisualBasic/Program.vb
hazzik: > I still think using simple NETCORE would minimize the frictions in the future..

It's quite opposite. Because we would need to maintain that constant somehow. 

I actually tried doing so. I wanted to be ready before .NET 6 release and that the effort to enable .NET 6 would be as simple as adding `;net60` to NhAppTargetFrameworks constant. So that was needed to be compatible and I had to maintain the constant manually because I did not want to push people into upgrading of the SDK until the .NET 6 release.

The ugliness and burden of this approach is a big part of why I decided to wait until .NET 6 release to propose the changes.

src/Directory.Build.props
bahusoid: Maybe we should keep NHibernate.props in place. We can just import it here instead?

hazzik: Why? I don't like it. IMO, better to stick to the standard way of doing things.

appveyor.yml
```diff
@@ -42,7 +42,7 @@ before_test:
         Start-Service 'MySQL57'
         # Create nhibernate database (not handled by NHibernate.TestDatabaseSetup.dll)
         $env:MYSQL_PWD = 'Password12!'
-        & 'C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql' -e 'CREATE DATABASE nhibernate;' --user=root
+        & 'C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql' -e 'CREATE DATABASE nhibernate COLLATE utf8_general_ci;' --user=root
```
hazzik, line 45: @fredericDelaporte do you remember how we've fixed this (or similar) problem in the past?

src/Directory.Build.props
bahusoid: Because I don't like looking at different places in different branches. In particular that I need to modify different dev.props files.

We still do it standard way  by importing our Nhibernate.props inside Directory.Build.props. And keep things the same across all branches.

hazzik: Maybe I just revert `.dev.props` change then?

hazzik: What do you propose to have in Nhibernate.props?

bahusoid: > What do you propose to have in Nhibernate.props?


I propose don't change Nhibernate.props. And content of `Directory.Build.props` should look something like:
```xml
<Project>
  <Import Project="../build-common/NHibernate.props"/>
</Project>
```

appveyor.yml
bahusoid: try with `CREATE DATABASE mydatabase CHARACTER SET utf8 COLLATE utf8_general_ci;`

src/Directory.Build.props
hazzik: image

appveyor.yml
hazzik: I'll try when this test run finishes.

src/NHibernate.Test/Async/StaticProxyTest/InterfaceHandling/Fixture.cs
hazzik: Fixed in 6.0.101

src/Directory.Build.props
hazzik: `NHibernate.props` restored.

src/NHibernate.Test/Async/Insertordering/FamilyModel/Fixture.cs
```diff
@@ -104,8 +105,8 @@ public async System.Threading.Tasks.Task CircularReferencesAsync()
 			using (var session = OpenSession())
 			using (var tran = session.BeginTransaction())
 			{
-				Assert.AreEqual(9, await (session.Query<Woman>().CountAsync()));
-				Assert.AreEqual(6, await (session.Query<Man>().CountAsync()));
+				Assert.AreEqual(9, session.Query<Woman>().Count());
+				Assert.AreEqual(6, session.Query<Man>().Count());
```
hazzik, line 109: @maca88 do you know why it did not pick up the `CountAsync` here?

src/NHibernate.sln
```diff
@@ -12,6 +12,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		..\build-common\common.xml = ..\build-common\common.xml
 		..\appveyor.yml = ..\appveyor.yml
 		..\ReleaseProcedure.txt = ..\ReleaseProcedure.txt
+		..\src\Directory.Build.props = ..\src\Directory.Build.props
```
bahusoid, line 15: Do we need it?

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -35,11 +34,12 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
-  <ItemGroup Condition="'$(TargetFramework)'=='netcoreapp2.0'">
+  <ItemGroup Condition="'$(TargetFramework)'=='net6.0'">
     <Compile Remove="**\CfgTest\DefaultNsAssmFixture.cs" />
     <Compile Remove="**\CfgTest\HbmOrderingFixture.cs" />
     <Compile Remove="**\DynamicProxyTests\PeVerifyFixture.cs" />
     <Compile Remove="**\Insertordering\**" />
+    <Compile Remove="**\Linq\ODataTests.cs" />
```
bahusoid, line 42: I remember you mentioned some problem with .NET 6 and OData. Are you sure it's still broken with latest Microsoft.AspNetCore.OData  8.x version?

src/NHibernate.Test/Async/Insertordering/InsertOrderingFixture.cs
```diff
@@ -675,16 +675,25 @@ public IBatcher CreateBatcher(ConnectionManager connectionManager, IInterceptor
...
+
+
+
+
+
+
+
+
+		
+
```
bahusoid, line 692: Hm.. Some whitespace glitch in AsyncGenerator?

global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "6.0.101"
```
bahusoid, line 3: Did you add it to warn users about minimal supported version?
Then we need `rollForward` too with some less restrictive rule. Default `latestPatch` is too restrictive (6.0.200 would be discarded), it at least should be `latestFeature`

src/NHibernate.Test/NHibernate.Test.csproj
hazzik: Yes, sure. It requires `Web` sdk,

src/NHibernate.Test/Async/Insertordering/FamilyModel/Fixture.cs
maca88: The problem lies in this part of the `NHibernate.Test.csproj`:
```xml
<ItemGroup Condition="'$(TargetFramework)'=='net6.0'">
    <Compile Remove="**\CfgTest\DefaultNsAssmFixture.cs" />
    <Compile Remove="**\CfgTest\HbmOrderingFixture.cs" />
    <Compile Remove="**\DynamicProxyTests\PeVerifyFixture.cs" />
    <Compile Remove="**\Insertordering\**" />
    <Compile Remove="**\Linq\ODataTests.cs" />
    <Compile Remove="**\NHSpecificTest\NH1850\**" />
    <Compile Remove="**\NHSpecificTest\NH2484\**" />
    <Compile Remove="**\NHSpecificTest\NH2188\**" />
    <Compile Remove="**\NHSpecificTest\NH3121\**" />
  </ItemGroup>
```

When the generator loads the project by using `net6.0` as the target framework, files matching the above globs won't be included and therefore the code won't be generated for them. By deleting the whole `Async` folder and re-generating the async code again, this fixture along with some others matching these globs won't be generated. They are not removed by the generator, because it removes only generated files that are present in the loaded project prior code generation.

src/NHibernate.Test/Async/Insertordering/InsertOrderingFixture.cs
maca88: This file is not generated anymore by the generator due to the [added globs](https://github.com/nhibernate/nhibernate-core/pull/2951#discussion_r770933890). When adding globs for ignoring files for the target framework used to generate async code (`net6.0` in this case), it is best to delete the `Async` folder prior generating to code to get rid of such files.

global.json
hazzik: Done

src/NHibernate.Test/Async/Insertordering/FamilyModel/Fixture.cs
hazzik: Thanks

src/NHibernate.sln
hazzik: Why not?

bahusoid: nit: Looks like unnecessary noise for me.

hazzik: It helps me not to switch to other editor from VS.

bahusoid: In current state it's just used as convenient way to import NHibernate.props to all projects. And that's all. So really there is need to edit it in solution - NHibernate.props should be edited instead.

bahusoid:
Should we add step to install .net 6.0.101 to AppVeyor?

bahusoid (dismissed):

bahusoid (approved):

bahusoid:
Just noticed that Rider no longer recognizes  symbols from ANTLR generated  code. Is it only me or I wonder if it's somehow related  to .NET 6 SDK.

For instance HqlSqlWalker no longer sees generated constants:
image

hazzik:
Same here. Probably issue with Rider. Probably worth to create a ticket in their issue tracker.

build-common/NHibernate.props
```diff
@@ -13,18 +13,26 @@
...
     <Copyright>Licensed under LGPL.</Copyright>
     <Authors>NHibernate community, Hibernate community</Authors>
     <NeutralLanguage>en-US</NeutralLanguage>
 
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+    <TreatSpecificWarningsAsErrors />
+    
+    <DisableImplicitPackageTargetFallback>True</DisableImplicitPackageTargetFallback>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(IsPackable)' == 'True'">
```
bahusoid, line 35: What  this `$(IsPackable)` about? I don't think it works properly. At least it seems all this information is now missing in generated nuget package

hazzik: :-(

hazzik: @bahusoid. `'$(IsPackable)'` evaluates to `'true'` instead of `'True'`.

hazzik: Should be fixed here: https://github.com/nhibernate/nhibernate-core/pull/3125

PR #2953, opened 2021-12-01
WIP Refactor BestGuessEntityName

bahusoid:
Original version has some strange logic for handling proxies..


src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -957,37 +957,14 @@ public void PersistOnFlush(object obj)
...
-					INHibernateProxy proxy = entity as INHibernateProxy;
-					ILazyInitializer initializer = proxy.HibernateLazyInitializer;
-
-					// it is possible for this method to be called during flush processing,
-					// so make certain that we do not accidentally initialize an uninitialized proxy
-					if (initializer.IsUninitialized)
-					{
-						return initializer.PersistentClass.FullName;
-					}
-					entity = initializer.GetImplementation();
```
fredericDelaporte, line 973: The entity name may be different from the proxy entity name. Getting the entity name can be more precise than getting the entity name known by the proxy.
This happens in case of mapped inheritance, when we have a proxy of the base class. When the proxy is initialized, the entity name will be the actual concrete implementation entity name, while the entity name known by the proxy will only be the one of the base class.
So the change here is a possible breaking change: with the change, entity-name guessed from proxies on base classes will always be the base class entity name. With current code it would be the concrete class entity name for initialized proxies.

bahusoid: And now old code makes sense :) Fixed and commented.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -337,11 +337,16 @@ public virtual IQuery GetNamedSQLQuery(string name)
 		[Obsolete("Replaced by FutureBatch")]
 		public abstract FutureQueryBatch FutureQueryBatch { get; protected internal set; }
 
-		protected private static string TryGetProxyEntityName(object entity)
+		protected private static string TryGetProxyEntityName(ref object entity)
```
hazzik, line 340: Technically this is a breaking change.

bahusoid: It's new method added in this PR

hazzik: True...

hazzik (dismissed):

hazzik:
@bahusoid can you update PR title and description to reflect what changes have been done? "Refactor" is just too ambiguous. And strictly speaking this is not a "refactoring", because behaviour of StatelessSession may have changed. Also, can we have tests for this new behavior of StatelessSession?

Issue #2952, opened 2021-12-01
Linq performance issue with non cacheable linq query plan 


gokhanabatay:
Hi,
We have a performance issue with below linq query is not added to query plan cache. Because x.F2.Mask(6, 4, '*', MaskOption.Default) of this function call makes query uncacheable. This function is masks card number middle digits other than first 6 and last 4 digits. This api called concurrent 100 request at the same time.

First of all why this function call leads to query plan uncacheable, is this desired behaviour?
When we look at Dynatrace pure path below image added there is a suspension at QueryTranslatorImpl.Compile that I cannot understand and takes to much time to compile the query?

To report all uncacheable queries could you change QueryPlanCache.GetHqlQueryPlan log.Debug("Query plan not cacheable") to log.Warn("Query plan not cacheable {0}", queryExpression.Key) Adding elapsed time could be better for performance issues, we cannot find this issue without Dynatrace.

Environment
NHibernate 5.3.9
.net 5.0
linux container

```
            var transactions = Query<TxnAcquirerOnl>().WithOptions(opt => opt.SetReadOnly(true))
                                                   .Where(predicateTxnAcquirerOnl)
                                                   .OrderByDescending(order => order.Guid)
                                                   .Select(x => new GetTransactionsResponse
                                                   {
                                                       TxnGuid = x.Guid,
                                                       CardNo = x.F2.Mask(6, 4, '*', MaskOption.Default),
                                                       CardDci = x.CardDci,
                                                       CardSource = x.CardSource,
                                                       TerminalType = x.TerminalType,
                                                       TransactionName = x.TxnDef.Description.GetText(CurrentLanguage),
                                                       TxnEffect = x.TxnDef.TxnEffect,
                                                       TxnEntry = x.TxnEntry,
                                                       TransactionStatus = x.TxnStt,
                                                       TransactionDate = x.F13,
                                                       TransactionTime = x.F12,
                                                       TxnRegion = x.TxnRegion,
                                                       BinProductType = x.BinProductType,
                                                       MerchantCode = x.F42,
                                                       MerchantName = x.F43Name,
                                                       MerchantLocation = x.F43,
                                                       MerchantCity = x.F43City,
                                                       MerchantCountry = x.F43Country,
                                                       TerminalCode = x.F41,
                                                       TerminalStan = x.F11Trm,
                                                       BatchNo = x.BatchNo,
                                                       InstallCount = x.TxnAcquirerInstall.InstallCnt,
[...]

PR #2955, opened 2021-12-02
Linq performance issue with non cacheable linq query plan #2952

mikailcetinkaya:
Log statement updating talked in #2952


mikailcetinkaya:
My friend Gokhan Abatay made a full request

PR #2959, opened 2021-12-16
Support Dynamic BatchFetchStyle

### LINKED ISSUE #1316: NH-3530 - memory when using default_batch_fetch_size
**Kris Desmadryl** created an [issue](https://nhibernate.jira.com/browse/NH-3530) — <time datetime="2013-09-12T20:45:09.062Z" title="2013-09-12T20:45:09.062Z">12th September 2013, 20:45:09</time>:
> As since NH3+ setting the default_batch_fetch_size to something has very bad impact on memory as soon the SessionFactory is created. I'm using 3 session factories in an web app (I'm accessing 3 db's). It uses 250MB without the setting but 750MB (default_batch_fetch_size=16) with the setting. When I switch back to 2.1.2 the problem is gone. My webserver recycles because the private memory is reached!
> 
> grtz,
> Kris
> 
---
**Filip Duyck** added a comment — <time datetime="2013-09-13T07:29:56.090Z" title="2013-09-13T07:29:56.090Z">13th September 2013, 7:29:56</time>:
> In addition to what Kris said, I'd like to share some results from our dotTrace profiling session.
> 
> Performance trace can be seen at  -- notice how in NHibernate 3, EntityLoader..ctor is called about 10x as much as in 2.1.2 (even though the calls to BatchingEntityLoader.CreateBatchingEntityLoader are more or less of the same order of magnitude).
> 
> We also did memory profiling which showed similar results, which you can see at . In this trace, you can see that BatchingEntityLoader.CreateBatchingEntityLoader creates many, many more objects in 3.3.3 than it did in 2.1.2.
> 
> I have looked at the code of both methods but I can't really identify a change. Our setup for NH2.1.2 and NH3.3.3 use virtually the same configuration. The problem is, as Kris mentioned, amplified by the fact that we have default*batch_fetch*size set to 16 which causes more looping inside the CreateBatchingEntityLoader method, but this setting is the same during 2.1.2 and 3.3.3 testing.
> 
> Can anyone from the NH team shed a light on this? Especially the memory problem is killing us in production.
> 
---
**Filip Duyck** added a comment — <time datetime="2014-07-16T13:13:52.838Z" title="2014-07-16T13:13:52.838Z">16th July 2014, 13:13:52</time>:
> It's been almost a year. Is anyone **ever** going to look at this?
>

bahusoid:
Fixes #1316
Port of https://github.com/hibernate/hibernate-orm/commit/06b0faaf5770da47681d9d32e06ea204519a3846 (without Padded fetch style)

To start using Dynamic batch style add the following property to session-factory settings:
```xml
<property name="batch_fetch_style">Dynamic</property>
```
Or configuration by code:
```C#
Configuration configuration = ...
configuration.SetProperty(Environment.BatchFetchStyle, BatchFetchStyle.Dynamic.ToString());
```


src/NHibernate.Test/Async/Legacy/FooBarTest.cs
```diff
@@ -1187,6 +1187,7 @@ public async Task BatchLoadAsync()
 				await (s.FlushAsync());
 			}
 
+			using(new SqlLogSpy())
```
hazzik, line 1190:
```suggestion
 			using (new SqlLogSpy())
```

A space is missing.

src/NHibernate.Test/Async/NHSpecificTest/GH1920/Fixture.cs
```diff
@@ -36,18 +39,43 @@ protected override void OnSetUp()
 		[TestCase(false)]
 		public async Task CanLoadEntityAsync(bool loadProxyOfOtherEntity)
 		{
+			using(new SqlLogSpy())
```
hazzik, line 42:
```suggestion
			using (new SqlLogSpy())
```

A space is missing.

src/NHibernate.Test/Legacy/FooBarTest.cs
```diff
@@ -1175,6 +1175,7 @@ public void BatchLoad()
 				s.Flush();
 			}
 
+			using(new SqlLogSpy())
```
hazzik, line 1178:
```suggestion
			using (new SqlLogSpy())
```

A space is missing.

src/NHibernate.Test/NHSpecificTest/GH1920/Fixture.cs
```diff
@@ -25,18 +28,43 @@ protected override void OnSetUp()
 		[TestCase(false)]
 		public void CanLoadEntity(bool loadProxyOfOtherEntity)
 		{
+			using(new SqlLogSpy())
```
hazzik, line 31:
```suggestion
			using (new SqlLogSpy())
```

A space is missing.

src/NHibernate.Test/NHSpecificTest/NH3142/ChildrenTest.cs
```diff
@@ -60,6 +60,7 @@ protected override void OnTearDown()
 		[Test]
 		public void ChildrenCollectionOfAllParentsShouldContainsThreeElements()
 		{
+			using(new SqlLogSpy())
```
hazzik, line 63:
```suggestion
			using (new SqlLogSpy())
```

A space is missing.

src/NHibernate/Engine/QueryParameters.cs
```diff
@@ -228,5 +234,7 @@ public bool CanPutToCache(ISessionImplementor session)
 		{
 			return (CacheMode ?? session.CacheMode).HasFlag(NHibernate.CacheMode.Put);
 		}
+
```
hazzik, line 237: Extra space, please cleanup.

hazzik (commented):
Please fix formatting issues identified by codefactor and by me.

bahusoid:
Please don't review till it's WIP. It's in working state AFAIK but not polished and no real tests

src/NHibernate/Util/StringHelper.cs
```diff
@@ -3,6 +3,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using NHibernate.SqlCommand;
```
hazzik, line 6: Please remove

src/NHibernate/Loader/DynamicBatchingHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+			parameters = new Parameter[paramsCount];
+			for (var i = 0; i < parameters.Length; i++)
+			{
+				parameters[i] = Parameter.Placeholder;
+			}
+
+			wherePart.SubstituteBogusParameters(parameters, bogusParam);
+			result = sqlString.ReplaceLast(DynamicBatchingHelper.BatchIdPlaceholder, wherePart);
+		}
+
```
hazzik, line 31: Please split into 2 methods. One to build where statement and another one to replace placeholder with that where statement.

src/NHibernate.Test/NHSpecificTest/NH3530/Fixture.cs
```diff
@@ -0,0 +1,183 @@
...
+
+			using (var session = OpenSession())
+			{
+				var proxy = session.Load<Entity>(_ids[0]);
+				var result = session.Get<Entity>(_ids[1]);
+
+				Assert.That(result.Name, Is.Not.Null);
+
+				var childrenCount = result.Children.Count;
+				//Assert.That(NHibernateUtil.IsInitialized(proxy), Is.True);
```
bahusoid, line 37: This entity is already loaded by batch loading. But proxy doesn't know about it - NHibernateUtil.IsInitialized(proxy) returns false. Looks like an issue to me.

src/NHibernate/Loader/DynamicBatchingHelper.cs
bahusoid: Ok. But common it's 10 lines method - IMO splitting doesn't improve anything.

hazzik: This method just makes zero sense.

bahusoid: Meaning? It's reused in multiple places

hazzik: > Meaning?

1. The method does not have a clear purpose it does a lot of half things.
2. It has 2 unrelated `out` parameters.

> It's reused in multiple places

Here are the 2 places that user that method:

```csharp
DynamicBatchingHelper.ExpandBatchIdPlaceholder(sqlString, queryParameters, columns, Factory.Dialect, out var parameters, out var result);
parameterSpecifications.UnionWith(CreateParameterSpecificationsAndAssignBackTrack(parameters));
```

```csharp
DynamicBatchingHelper.ExpandBatchIdPlaceholder(sqlString, queryParameters, columns, Factory.Dialect, out var parameters, out var result);
parameterSpecifications.UnionWith(CreateParameterSpecificationsAndAssignBackTrack(parameters));
```

Spot the difference... 

`CreateParameterSpecificationsAndAssignBackTrack` are also identical with only difference is the key type.

bahusoid: Purpose - minimize code duplication.

It returns SQL command and parameters used in this command - looks pretty related to me.

`CreateParameterSpecificationsAndAssignBackTrack` - same name, different methods  . I don't want to bother unifying it.

hazzik: > `CreateParameterSpecificationsAndAssignBackTrack` - same name, different methods. 

It is exactly same method, just differs in one thing that could be an argument. 

But after a close inspection turns out that the whole construct is just a cart in front of a horse.

src/NHibernate/Loader/DynamicBatchingHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+		private static SqlString GenerateWherePart(QueryParameters queryParameters, string[] columns, Dialect.Dialect dialect, out Parameter[] parameters)
+		{
+			var bogusParam = Parameter.Placeholder;
+			var wherePart = InExpression.GetSqlString(columns, queryParameters.PositionalParameterValues.Length, dialect, bogusParam);
+			var paramsCount = wherePart.GetParameterCount();
+			parameters = new Parameter[paramsCount];
+			for (var i = 0; i < parameters.Length; i++)
+			{
+				parameters[i] = Parameter.Placeholder;
+			}
```
hazzik, line 32:
```suggestion
			parameters = Parameter.GenerateParameters(paramsCount)
```

src/NHibernate/Loader/DynamicBatchingHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+		private static string BatchIdPlaceholder = "$$BATCH_ID_PLACEHOLDER$$";
+
+		public static SqlStringBuilder BuildBatchFetchRestrictionFragment()
+		{
+			return new SqlStringBuilder(1).Add(BatchIdPlaceholder);
+		}
+
+		public static void ExpandBatchIdPlaceholder(SqlString sqlString, QueryParameters queryParameters, string[] columns, Dialect.Dialect dialect, out SqlString result, out Parameter[] parameters)
+		{
+			var wherePart = GenerateWherePart(queryParameters, columns, dialect, out parameters);
```
hazzik, line 19: At this point you already know the number of parameters required. `parameters.Length == columns.Length * queryParameters.PositionalParameterValues.Length`. So you can pass the parameter's array without need of an out parameter.

hazzik: Then... `InExpression.GetSqlString` is always used in this combination:

```csharp
var bogusParam = Parameter.Placeholder;
var wherePart = InExpression.GetSqlString(columns, paramsCount, dialect, bogusParam);
wherePart.SubstituteBogusParameters(parameters, bogusParam);
```
This could and should be one method.

bahusoid: This split is not very useful in the first place and now you are suggesting to complicate it even further by some pre-calculations based on the internal implementation details. I don't like it. Sorry I'm going to undo your last commit.

src/NHibernate/Loader/Entity/AbstractEntityLoader.cs
```diff
@@ -74,7 +74,7 @@ protected virtual object Load(ISessionImplementor session, object id, object opt
 
 		protected IType UniqueKeyType { get; private set; }
 
-		private IEnumerable<IParameterSpecification> CreateParameterSpecificationsAndAssignBackTrack(IEnumerable<Parameter> sqlPatameters)
+		internal protected List<IParameterSpecification> CreateParameterSpecificationsAndAssignBackTrack(IEnumerable<Parameter> sqlPatameters)
```
hazzik, line 77: This change is not required.

src/NHibernate/Loader/DynamicBatchingHelper.cs
bahusoid: It's 5 lines of code now - just let it be..

bahusoid: I'm just re-using the pattern that's used in `Loader`. Just check how `CreateSqlCommand` works and see `GetParameterSpecifications`

hazzik: The difference between `GetParameterSpecifications` mentioned and this method that here we know the structure of the string fragment and number of parameters and can write more optimal and readable code. And in `GetParameterSpecifications` the `SqlString` could be anything. It could be a user-provided select statement or a generated string, or something else.

bahusoid: You've just introduced yet another version of `CreateParameterSpecificationsAndAssignBackTrack`. I really don't see how it improves anything...

fredericDelaporte (approved):

gokhanabatay:
@bahusoid  @hazzik 
For dynamic batch fetch style works correctly does `default_batch_fetch_size` needs to be bigger than 1, if it is than is it cause to dublicate queries in memory for those with 2 and 1 parameters.

property in (p0:,p1) 
property in (p0)

```c#
public virtual IUniqueEntityLoader BuildLoader(IOuterJoinLoadable persister, int batchSize, LockMode lockMode, ISessionFactoryImplementor factory, IDictionary<string, IFilter> enabledFilters)
{
	if (batchSize <= 1)
	{
		// no batching
		return new EntityLoader(persister, lockMode, factory, enabledFilters);
	}

	return BuildBatchingLoader(persister, batchSize, lockMode, factory, enabledFilters);
}
```

PR #2964, opened 2021-12-26
Fix parameter detection when using custom hql functions

### LINKED ISSUE #2963: Time is incompatible with bigint for TimeAsTimeSpanType
### Context
Updating v5.1.7 to 5.3.10 together with [FluentNHibernate 3.1.0](https://github.com/nhibernate/fluent-nhibernate/releases/tag/v3.1.0)

NHibernate LINQ providers don't support `DateTime.TimeOfDay` to automatically convert to SQL `time`

### Approach
`RegisterFunction()` to the rescue to adjust the dialect, where in 5.1.7 the following code works for _MS SQL 2012 dialect_
```cs
RegisterFunction(nameof(Ext.GetTime), 
                 new SQLFunctionTemplate(NHibernateUtil.TimeAsTimeSpan, "cast(?1 as time)"));
```
where `Ext.GetTime` only accesses `TimeOfDay`
```cs
public static class Ext
{
    public static TimeSpan GetTime(this DateTime dateTime)
    {
        return dateTime.TimeOfDay;
    }
}
```

### Usage
```cs
var start = filter.Start.GetTime();
var end = filter.End.GetTime();

return Session.Query<Entity>().Where(x => x.Start.GetTime() < end && x.End.GetTime() > start).ToList();

class Entity
{
    public virtual DateTime Start { get; set; }
    public virtual DateTime End { get; set; }
}
```

### Exception in 5.3.10
While v5.3.10 throws an exception
> ---- System.Data.SqlClient.SqlException : Operand type clash: time is incompatible with bigint

I understand _TimeSpan_ can have more than 24h, whereas this method is used only for less than 24h intervals.

**Related:** [TimeAsTimeSpanType](https://github.com/nhibernate/nhibernate-core/blob/47992e73ed6114093f31ed99829dd8c4c7663829/src/NHibernate/Type/TimeAsTimeSpanType.cs) has changes from v5.1.3 related to casting types to `TimeSpan`.

### Questions
Trying to figure out 
1. Where `InternalTicks` from `TimeOfDay` tries to map to SQL type `time` in NHibernate
1. Documentation related to this breaking change
1. Why v5.1.7 works and what is the new approach in v5.3.10 version

Open to ideas how to contribute to documentation 😃

maca88:
Fixes #2963


src/NHibernate.Test/TypesTest/CharClassFixture.cs
```diff
@@ -31,5 +32,33 @@ public void ReadWrite()
...
+				var result = (from e in session.Query<CharClass>()
+							where e.AnsiChar == 'B'
+							select e).ToList();
+				
+				Assert.That(logSpy.GetWholeLog(), Does.Contain("Type: AnsiString"));
+			}
+		}
+
+		[Test]
+		public void ParameterTypeForCharInAnsiStringInLinq()
```
bahusoid, line 51: Added test case that I described in https://github.com/nhibernate/nhibernate-core/issues/2963#issuecomment-1001112687

bahusoid: Hm... It also fails in 5.2 for some other reasons (I didn't look into it). So not really  a regression - so let's not touch it in 5.3...

bahusoid (approved):

src/NHibernate.Test/TypesTest/CharClassFixture.cs
maca88: It doesn't work in hql because `AnsiString[0]` is translated to function `substring` which is registered by `AnsiSubstringFunction` that has `NHibernateUtil.String` as the return type. With `5.3` it is detected as `StringFixedLengthSqlType`, which is better than before but still not correct, as it should be `AnsiStringFixedLengthSqlType`.

src/NHibernate.Test/TestDialect.cs
```diff
@@ -42,6 +42,8 @@ public bool NativeGeneratorSupportsBulkInsertion
 				(IIdentifierGenerator) Cfg.Environment.ObjectsFactory.CreateInstance(
 					_dialect.NativeIdentifierGeneratorClass));
 
+		public virtual bool SupportsTime => true;
```
bahusoid, line 45: Maybe no need to override it with:
```C#
public virtual bool SupportsTime => _dialect.GetTypeName(new SqlType(DbType.Time)) != _dialect.GetTypeName(new SqlType(DbType.DateTime));
```

maca88: I agree, changed your suggestion.

src/NHibernate.Test/Linq/CustomExtensionsExample.cs
```diff
@@ -81,6 +100,19 @@ public void CanUseObjectEquals()
...
+		[Test(Description = "GH-2963")]
+		public void CanUseComparisonWithExtensionOnMappedProperty()
+		{
+			if (!TestDialect.SupportsTime)
+			{
+				Assert.Ignore("Time type is not supported");
+			}
+
+			var time = DateTime.UtcNow.GetTime();
+			//using(new SqlLogSpy())
```
hazzik, line 112: Please delete

src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
```diff
@@ -98,6 +99,21 @@ public override void ReportError(RecognitionException e)
...
 
+		internal IStatement Transform()
+		{
+			var tree = (IStatement) statement().Tree;
+			// Use the guessed type in case we weren't been able to detect the type
+			foreach (var parameter in _parameters)
+			{
+				if (parameter.ExpectedType == null && _guessedParameterTypes.TryGetValue(parameter, out var guessedType))
+				{
+					parameter.ExpectedType = guessedType;
```
hazzik, line 110: Shouldn't just `break`?

bahusoid: One case: when parameter value is NULL it can't be guessed by hql.
It's already covered by some tests.

hazzik: But would not the `_guessedParameterTypes.TryGetValue(..., out guessedType)` expression return false?

bahusoid: No. It's guessed  during LINQ  processing as type is known there.

hazzik: So the question really is... can the `guessedType` variable be null? If not, then the cycle should break after assignment.

bahusoid: Ehh. Cycle break? Why? Multiple parameters might be NULL. All of them should use guessed type if ExpectedType is null

bahusoid: So I misunderstood the question. But still I don't get why you want to break the cycle if one of the parameter uses `guessedType`. Next parameter might also be NULL and might also use `guessedType`. This cycle is about different parameters. Not about detecting parameter type for single parameter

hazzik: Yeah, you're right. My bad. Sorry.

src/NHibernate.Test/Async/Linq/CustomExtensionsExample.cs
```diff
@@ -42,6 +42,19 @@ public async Task CanUseObjectEqualsAsync()
...
+		[Test(Description = "GH-2963")]
+		public async Task CanUseComparisonWithExtensionOnMappedPropertyAsync()
+		{
+			if (!TestDialect.SupportsTime)
+			{
+				Assert.Ignore("Time type is not supported");
+			}
+
+			var time = DateTime.UtcNow.GetTime();
+			//using(new SqlLogSpy())
```
hazzik, line 54:
```suggestion
```

hazzik (approved):

src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
hazzik, line 110: Why do we need `_guessedParameterTypes` dictionary? Why we cannot just do this: 

```csharp
if (parameter.ExpectedType == null && parameter.GuessedType != null)
{
    parameter.ExpectedType = parameter.GuessedType;
}
```

hazzik: Answering my own question: because `parameter` here does not have `GuessedType` property.

src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
```diff
@@ -1072,7 +1088,10 @@ IASTNode GenerateNamedParameter(IASTNode delimiterNode, IASTNode nameNode)
 				// Add the parameter type information so that we are able to calculate functions return types
 				// when the parameter is used as an argument.
 				if (namedParameter.IsGuessedType)
+				{
+					_guessedParameterTypes[paramSpec] = namedParameter.Type;
 					parameter.GuessedType = namedParameter.Type;
```
hazzik, line 1093: can we do here this?

```csharp
paramSpec.ExpectedType = namedParameter.Type;
```

hazzik: Or is it too early?

bahusoid: Yes. Too early. HQL parameter detection logic is not yet executed

bahusoid: I have an idea how to get rid of `_guessedParameterTypes`. But it requires new method in `IQuery`, so not a good fit for a patch release.

PR #2966, opened 2021-12-28
WIP Concat should return supplied string type

bahusoid:
Useful for parameter detection logic (i.e. to properly detect AnsiString type for parameter in hql)

Fixes part of the problem described in #1166

### DIFF
--- a/src/NHibernate.Test/Criteria/Lambda/FunctionsIntegrationFixture.cs
+++ b/src/NHibernate.Test/Criteria/Lambda/FunctionsIntegrationFixture.cs
@@ -134,9 +134,6 @@ namespace NHibernate.Test.Criteria.Lambda
 		[Test]
 		public void Concat()
 		{
-			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
-				Assert.Ignore("Current dialect does not support this test");
-
 			using (var s = OpenSession())
 			using (s.BeginTransaction())
 			{
--- a/src/NHibernate.Test/Criteria/ProjectionsTest.cs
+++ b/src/NHibernate.Test/Criteria/ProjectionsTest.cs
@@ -58,9 +58,6 @@ namespace NHibernate.Test.Criteria
 		[Test]
 		public void UsingSqlFunctions_Concat()
 		{
-			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
-				Assert.Ignore("Current dialect does not support this test");
-
 			using (ISession session = Sfi.OpenSession())
 			{
 				string result = session.CreateCriteria(typeof(Student))
@@ -78,12 +75,8 @@ namespace NHibernate.Test.Criteria
 		[Test]
 		public void UsingSqlFunctions_Concat_WithCast()
 		{
-			if(Dialect is Oracle8iDialect)
-			{
+			if (Dialect is Oracle8iDialect)
 				Assert.Ignore("Not supported by the active dialect:{0}.", Dialect);
-			}
-			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
-				Assert.Ignore("Current dialect does not support this test");
 
 			using (ISession session = Sfi.OpenSession())
 			{
--- a/src/NHibernate.Test/Linq/LinqQuerySamples.cs
+++ b/src/NHibernate.Test/Linq/LinqQuerySamples.cs
@@ -2001,9 +2001,6 @@ namespace NHibernate.Test.Linq
 		[Test(Description = "GH-2860")]
 		public void StringFormatWithTrim()
 		{
-			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
-				Assert.Ignore("Current dialect does not support this test");
-
 			var q =
 				from e in db.Employees
 				select new { Name = $"{e.FirstName} {e.LastName}".Trim(), Phone = e.Address.PhoneNumber };
--- a/src/NHibernate.Test/Linq/LinqQuerySamples.cs
+++ b/src/NHibernate.Test/Linq/LinqQuerySamples.cs
@@ -2001,6 +2001,9 @@ namespace NHibernate.Test.Linq
 		[Test(Description = "GH-2860")]
 		public void StringFormatWithTrim()
 		{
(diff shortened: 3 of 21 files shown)


hazzik (dismissed):

src/NHibernate.Test/NHSpecificTest/NH1280/NH1280Fixture.cs
```diff
@@ -41,9 +41,6 @@ protected override void OnTearDown()
 		[Test]
 		public void HavingUsingSqlFunctions_Concat()
 		{
-			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
```
hazzik, line 44: This test does not have anything to do with selected parameters. The only dialect that currently has `HasBrokenTypeInferenceOnSelectedParameters == true` is `SapSQLAnywhere17Dialect`. This test tests concat, so I thought this change should re-enable this test.

src/NHibernate.Test/Linq/LinqQuerySamples.cs
```diff
@@ -2001,9 +2001,6 @@ public void ReplaceFunctionWithNullArgument()
 		[Test(Description = "GH-2860")]
 		public void StringFormatWithTrim()
 		{
-			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
```
hazzik, line 2004: This test has nothing to do with "selected parameters".

src/NHibernate.Test/Criteria/ProjectionsTest.cs
```diff
@@ -78,12 +75,8 @@ public void UsingSqlFunctions_Concat()
 		[Test]
 		public void UsingSqlFunctions_Concat_WithCast()
 		{
-			if(Dialect is Oracle8iDialect)
-			{
+			if (Dialect is Oracle8iDialect)
 				Assert.Ignore("Not supported by the active dialect:{0}.", Dialect);
-			}
-			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
```
hazzik, line 85: Has nothing to do with "selected parameters"

src/NHibernate.Test/Criteria/ProjectionsTest.cs
```diff
@@ -58,9 +58,6 @@ protected override void OnTearDown()
 		[Test]
 		public void UsingSqlFunctions_Concat()
 		{
-			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
```
hazzik, line 61: Has nothing to do with "selected parameters". It tests concat.

src/NHibernate.Test/Criteria/Lambda/FunctionsIntegrationFixture.cs
```diff
@@ -134,9 +134,6 @@ public void FunctionsToLowerToUpper()
 		[Test]
 		public void Concat()
 		{
-			if (TestDialect.HasBrokenTypeInferenceOnSelectedParameters)
```
hazzik, line 137: Has nothing to do with "selected parameters". It tests concat.

src/NHibernate/Dialect/SybaseSQLAnywhere10Dialect.cs
```diff
@@ -257,7 +257,6 @@ protected virtual void RegisterStringFunctions()
 			RegisterFunction("char_length", new StandardSQLFunction("char_length", NHibernateUtil.Int32));
 			RegisterFunction("compare", new VarArgsSQLFunction(NHibernateUtil.Int32, "compare(", ",", ")"));
 			RegisterFunction("compress", new VarArgsSQLFunction(NHibernateUtil.BinaryBlob, "compress(", ",", ")"));
-			RegisterFunction("concat", new VarArgsSQLFunction(NHibernateUtil.String, "(", "+", ")"));
```
hazzik, line 260: `+` behaves differently than `||` and requiring conversion to string first.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -101,7 +101,7 @@ protected Dialect()
 			RegisterFunction("cast", new CastFunction());
 			RegisterFunction("transparentcast", new TransparentCastFunction());
 			RegisterFunction("extract", new AnsiExtractFunction());
-			RegisterFunction("concat", new VarArgsSQLFunction(NHibernateUtil.String, "(", " || ", ")"));
+			RegisterFunction("concat", new VarArgsSQLFunction("(", " || ", ")"));
```
fredericDelaporte, line 104: I think this is an invalid change. It is not the first parameter which matters, but all parameters. As pointed in my first comment, all parameters must be ANSI strings for the return type to be an ANSI string. By not specifying the return type, we let NHibernate infer the returned type from the first parameter, which is then incorrect if the first is an ANSI string but not all others.

It is also worst to incorrectly type something as an ANSI string than incorrectly type something as an Unicode string: the former may cause garbled text and performance issues, while the later may only cause performance issues. (Well, since with SQL Server 2019, varchar can be [UTF-8 encoded](https://docs.microsoft.com/en-us/sql/relational-databases/collations/collation-and-unicode-support#utf8), the former in such case could not have the garbled text issue. But that remains a corner case for now I think.)

So I think this change is a regression.
We would need instead a specialized concat function which check the type of all its arguments, and would yield an ANSI string only if all of them are ANSI strings, otherwise it would yield an Unicode string.

bahusoid: In case of simple comparison `p.AnsiColumn = @param` if `@param` type is not explicitly specified it will be treated as ANSI. Even though `@param` might contain some UTF-8 symbols...  

To me this case is not much of a difference. It's just an assumption that's used for type detection for not specified explicitly parameter types. So I don't think any garbled text is possible here (only wrong comparisons results, the same as with simple comparison)

But OK let's be on a safe side...

hazzik:
Ok, I'll extract my changes for SqlAnywhere into a separate PR since this is WIP.

hazzik:
@fredericDelaporte @bahusoid what is the conclusion on this PR? This is the last one marked for 5.4.

bahusoid:
> what is the conclusion on this PR? 

It's not ready. I removed it from 5.4 milestone

PR #2967, opened 2021-12-30
WIP Add testing DB2 to GitHub Actions

bahusoid:
Too many tests error..


hazzik:
DB2 requires similar to Firebird workaround for parameters. Parameters that are used in the SELECT statements should be wrapped into `cast`.


```
2021-12-31T14:59:15.4008216Z 14) Error : NHibernate.Test.Criteria.ProjectionsTestAsync.UsingConditionalsAsync
2021-12-31T14:59:15.4010114Z NHibernate.Exceptions.GenericADOException : could not execute query
2021-12-31T14:59:15.4011188Z [ SELECT (case when this_.studentId = @p0 then @p1 else @p2 end) as y0_ FROM Student this_ ]
2021-12-31T14:59:15.4012866Z   Name:cp0 - Value:27 [Type: NHibernate.Type.Int64Type (SqlType: Int64)]  Name:cp1 - Value:yes [Type: NHibernate.Type.StringType (SqlType: String)]  Name:cp2 - Value:no [Type: NHibernate.Type.StringType (SqlType: String)]
2021-12-31T14:59:15.4014355Z [SQL: SELECT (case when this_.studentId = @p0 then @p1 else @p2 end) as y0_ FROM Student this_]
2021-12-31T14:59:15.4016006Z   ----> IBM.Data.DB2.Core.DB2Exception : ERROR [42610] [IBM][DB2/LINUXX8664] SQL0418N  The statement was not processed because the statement contains an invalid use of one of the following: an untyped parameter marker, the DEFAULT keyword, or a null value.
2021-12-31T14:59:15.4017111Z Data:
2021-12-31T14:59:15.4017905Z   actual-sql-query: SELECT (case when this_.studentId = @p0 then @p1 else @p2 end) as y0_ FROM Student this_
```

See this: https://www.ibm.com/docs/es/db2/11.1?topic=design-parameters-markers#d97366e93

hazzik:
I think we need to have a systematic approach rather than trying to fix this problem for different DBs. (I think Hana has the same problem)

bahusoid:
> DB2 requires similar to Firebird

I don't really like how it's done in Firebird. We need to think about more elegant solution rather than "brute force"  parameters wrapping in cast. The request to disable this functionality proves it can cause more harm than good

hazzik:
Yes, agree.

bahusoid:
```C#
171) Error : NHibernate.Test.TypesTest.StringTypeWithLengthFixture.HqlEqualityParameterCanExceedColumnSize
NHibernate.Exceptions.GenericADOException : could not execute query
[ select stringclas0_.Id as id1_0_, stringclas0_.StringValue as stringvalue2_0_, stringclas0_.LongStringValue as longstringvalue3_0_ from StringClass stringclas0_ where stringclas0_.StringValue=@p0 ]
  Name:likeValue - Value:AAAAAAAAABx [Type: NHibernate.Type.StringType (SqlType: String(Length=10))]
  ----> IBM.Data.DB2.Core.DB2Exception : ERROR [22001] [IBM] CLI0109E  String data right truncation. SQLSTATE=22001
```
This test is skipped if `SupportsNonDataBoundCondition` is false. But I see no non data bound conditions. It seems we need to adjust parameter sizes similar how it's done for MSSQL in `SqlClientDriver.AdjustParameterForValue`

hazzik:
> This test is skipped if SupportsNonDataBoundCondition is false.

Skip condition is incorrect then.

src/NHibernate.Test/TestDialect.cs
```diff
@@ -120,10 +120,15 @@ public bool SupportsSqlType(SqlType sqlType)
 		/// </summary>
 		public virtual bool SupportsModuloOnDecimal => true;
 
+		/// <summary>
+		/// Supports sub-selects in order by clause
+		/// </summary>
+		public virtual bool SupportsSubSelectsInOrderBy => _dialect.SupportsScalarSubSelects;
```
hazzik, line 126: DB2 supports scalar sub selects, but it does not support them in order-by clause.

bahusoid: And in SELECT statement seems too

hazzik: @bahusoid this test works just fine:

https://github.com/nhibernate/nhibernate-core/blob/6576a9675cbfc48b560c58f85794013b4512131d/src/NHibernate.Test/Linq/SelectionTests.cs#L408-L464

hazzik:
>  Use NET5.IBM.Data.DB2 lib instead of IBM.Data.DB2.Core 

I was thinking about this. But why?

hazzik:
Also... The assembly name and namespace has reverted back to .NET version. Need to change driver.

bahusoid:
> I was thinking about this. But why?

Just to check how it works... It's the latest version. So why not?

bahusoid:
> Just to check how it works.

And it doesn't work... So let's just drop it for now..

hazzik:
Yeah, it is `IBM.Data.Db2.DB2Connection` (in assembly) vs `IBM.Data.DB2.DB2Connection` (expected)

bahusoid:
Changing casing in namespace. Smart move...

bahusoid:
Interesting async regen failure.
<details> 
  <summary>Logs </summary>
crit: AsyncGenerator.CommandLine[0]
      One or more errors occurred while opening the project:
      Msbuild failed when processing the file '/home/runner/work/nhibernate-core/nhibernate-core/src/NHibernate.Test/NHibernate.Test.csproj' with message: /usr/share/dotnet/sdk/6.0.405/Microsoft.Common.CurrentVersion.targets: (2302, 5): There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "/home/runner/.nuget/packages/ibm.data.db2.core-lnx/3.1.0.500/lib/netstandard2.1/IBM.Data.DB2.Core.dll", "AMD64". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.
      Hint: For suppressing irrelevant errors use SuppressDiagnosticFailures option.
      System.InvalidOperationException: One or more errors occurred while opening the project:
      Msbuild failed when processing the file '/home/runner/work/nhibernate-core/nhibernate-core/src/NHibernate.Test/NHibernate.Test.csproj' with message: /usr/share/dotnet/sdk/6.0.405/Microsoft.Common.CurrentVersion.targets: (2302, 5): There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "/home/runner/.nuget/packages/ibm.data.db2.core-lnx/3.1.0.500/lib/netstandard2.1/IBM.Data.DB2.Core.dll", "AMD64". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.
      Hint: For suppressing irrelevant errors use SuppressDiagnosticFailures option.
         at AsyncGenerator.AsyncCodeGenerator.CheckForErrors(MSBuildWorkspace workspace, String itemType, ImmutableArray`1 supressFailuresPredicates, ILogger logger) in C:\Workspace\Git\AsyncGenerator\Source\AsyncGenerator\AsyncCodeGenerator.cs:line 437
         at AsyncGenerator.AsyncCodeGenerator.OpenProject(MSBuildWorkspace workspace, String filePath, ImmutableArray`1 supressFailuresPredicates, ILogger logger, CancellationToken cancellationToken) in C:\Workspace\Git\AsyncGenerator\Source\AsyncGenerator\AsyncCodeGenerator.cs:line 375
         at AsyncGenerator.AsyncCodeGenerator.GenerateAsync(AsyncCodeConfiguration configuration, CancellationToken cancellationToken) in C:\Workspace\Git\AsyncGenerator\Source\AsyncGenerator\AsyncCodeGenerator.cs:line 76
         at AsyncGenerator.CommandLine.Program.Main(String[] args) in C:\Workspace\Git\AsyncGenerator\Source\AsyncGenerator.CommandLine\Program.cs:line 51
</details>

robsosno:
> Changing casing in namespace. Smart move...

I've just moved my app from .NET 3.1 to .NET 6.0. Next step it would be to have new Db2 driver supporting .NET 6.0 (I expect better performance). However I was surprised with driver name change - [db2-net-packages](https://community.ibm.com/community/user/datamanagement/blogs/vishwa-hs1/2020/07/12/db2-net-packages-download-and-configure) (breaking changes paragraph).
I see some logic in the naming: IBM.Data.DB2 is dedicated to classic .NET framework, IBM.Data.DB2.Core to .NET Core and IBM.Data.Db2 to .NET 5+ . They removed Core similarly as Framework did and to avoid possibility of incorrect linking 3.1 with 5+ they've changed DB2 to Db2. So it seems that all future versions will be in accordance with this new schema.
I saw some changes on [db2Net5](https://github.com/bahusoid/nhibernate-core/tree/db2Net5/src/NHibernate/Driver) branch. When do you plan to merge it to the core?
I would vote to implement driver differently - as an additional class, let it be named Db2Driver or Db2Driver5. I trust more to manual configuration than smart switches.

hazzik:
Note to myself:

On Windows need to set `DB2LANG=en_US` environment variable.

hazzik:
Current state: 91 tests are failing.

PR #2968, opened 2022-01-02
Fix QueryStatistics.ExecutionAvgTime calculation

maikschott:
QueryStatistics.ExecutionAvgTime was incorrectly calculated.

Assuming a query first took 10ms and then 30ms, the average between both should be 20ms.
However, the old implementation was _newAvgTime = (oldAvgTime * n + curTime) / n_ which in the provided example would result in (10*2+30)/2 = 50/2 = 25.

As the old average time was calculated on (_n_-1) samples the first instance of _n_ should be _n_-1 instead, which correctly results in (10*(2-1)+30)/2 = 40/2 = 20.


hazzik (dismissed):

bahusoid:
Seems async regen is needed for tests

hazzik:
@bahusoid thanks. Done.

hazzik (approved):

PR #2980, opened 2022-01-11
Allow entity not found delegate to handle when an entity is not found

cedar-cranson:
NHibernate 5.3 changed the behavior of entity not found.  In the code there is this comment that _//6.0 TODO Add to IEntityNotFoundDelegate interface_ that made me assume it will be correctly fixed in 6.0.  My guess was that in the point version it was desired to not add additional methods to public interfaces so `void HandleEntityNotFound(string entityName, string propertyName, object key);` was not added to IEntityNotFoundDelegate but because DefaultEntityNotFoundDelegate is internal it appeared to make it so there was no way to changed the handling of EntityNotFound in the case of when there was a property associated with it.  So I added a new interface that an IEntityNotFoundDelegate could also implement so the behavior could be modified.  I think it would be better if IEntityNotFoundDelegate just had the additional method but by making a new interface it does not break current compile but make it easy for projects that need the override to change the behavior.

### DIFF
--- a/src/NHibernate/Proxy/IEntityNotFoundDelegate.cs
+++ b/src/NHibernate/Proxy/IEntityNotFoundDelegate.cs
@@ -16,7 +16,7 @@ namespace NHibernate.Proxy
 			string propertyName,
 			object key)
 		{
-			if (interceptor is Impl.SessionFactoryImpl.DefaultEntityNotFoundDelegate x)
+			if (interceptor is IEntityNotFoundPropertyDelegate x)
 			{
 				x.HandleEntityNotFound(entityName, propertyName, key);
 				return;
@@ -26,6 +26,14 @@ namespace NHibernate.Proxy
 		}
 	}
 
+	/// <summary> 
+	/// Delegate to handle the scenario of an entity not found by a specified id and property. 
+	/// </summary>
+	public interface IEntityNotFoundPropertyDelegate
+	{
+		void HandleEntityNotFound(string entityName, string propertyName, object key);
+	}
+	
 	/// <summary> 
 	/// Delegate to handle the scenario of an entity not found by a specified id. 
 	/// </summary>
--- a/src/NHibernate/Proxy/IEntityNotFoundDelegate.cs
+++ b/src/NHibernate/Proxy/IEntityNotFoundDelegate.cs
@@ -16,7 +16,7 @@ namespace NHibernate.Proxy
 			string propertyName,
 			object key)
 		{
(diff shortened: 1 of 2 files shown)


src/NHibernate/Proxy/IEntityNotFoundDelegate.cs
```diff
@@ -26,6 +26,14 @@ internal static class EntityNotFoundDelegateExtension
 		}
 	}
 
+	/// <summary> 
+	/// Delegate to handle the scenario of an entity not found by a specified id and property. 
+	/// </summary>
+	public interface IEntityNotFoundPropertyDelegate
```
hazzik, line 32: Can we have another name please? This is slightly confusing.

bahusoid: So it's about https://github.com/nhibernate/nhibernate-core/pull/2049

It's a temp interface. So please add comment to remove it in 6.0. Something like `//6.0 TODO: Remove`

Also I think it can be considered as regression so can be retargeted to 5.3.x branch.

>Can we have another name please?

`IEntityNotFoundByUniqueKeyDelegate` or `IEntityNotFoundByPropertyRefDelegate`?

hazzik: I like `IEntityNotFoundByPropertyRefDelegate` better.

hazzik: > IEntityNotFoundByUniqueKeyDelegate or IEntityNotFoundByPropertyRefDelegate?

Any of them.

cedar-cranson: Thank you for reviewing the commit I have made the requested changes please let me know if there is anything else I can do.

bahusoid:
Just to clarify things. 

> NHibernate 5.3 changed the behavior of entity not found. 

So now it respects `not-found` mapping setting (which is `exception` by default). You can specify `not-found="ignore"` in property mapping to ignore not found records. So do we really need this extension? Do you use it for something else?

hazzik (approved):

hazzik (approved):

bahusoid (changes requested):
I don't think we  need this change.  The proper way to avoid exception here is to use `not-found="ignore"`  in mapping (as I already said in https://github.com/nhibernate/nhibernate-core/pull/2980#issuecomment-1011374350)

hazzik:
Ok. Closing as per @bahusoid

PR #2989, opened 2022-01-14
Handle null and not null checks in queries on not-found='ignore' properties

### LINKED ISSUE #2988: More query issues when using not-found='ignore' in entity mapping
There was a [previous](https://github.com/nhibernate/nhibernate-core/issues/2611) fix to queries for entites that use the "not-found='ignore'" mapping that worked prior to 5.3. That fixed an issue when there was a single null check, but there are still some other edge cases that aren't being handled correctly.

The first is when querying for null or another entity. For example, giving two entities:
```
public class Human
{
    public virtual int Id { get; protected set; }
}

public class Dog
{
    public virtual int Id { get; protected set; }
}
```

Giving the mapping
```
public class DogMap : ClassMap<Dog>
{
    public DogMap()
    {
        Id(x => x.Id);
        Map(x => x.Name)
            .Length(16)
            .Not.Nullable();
        References(x => x.Human)
            .NotFound.Ignore();
    }
}
```
And data:
```
Fido -> Joe (deleted)
Buddy -> Jane
```
and we query
```
session.Query<Dog>().Where(d => d.Human == null || d.Human.Id == janeId).Select(d => d.Name)
```
Only one dog with the human that still exists will show up due to an implicit comma join:
```
select
	dog0_.Name as col_0_0_
from
	"Dog" dog0_
left outer join
	 "Human" human1_ on dog0_.Human_id=human1_.Id,
	 "Human" human2_
where
	dog0_.Human_id=human2_.Id and (human1_.Id is null or human2_.Id=@p0);
@p0 = 2 [Type: Int32 (0:0:0)]
```

The second issue is when you try and query for the deleted id directly:
```
session.Query<Dog>().Where(d => d.Human.Id == joeId).Select(d => d.Name)
```
This used to bring back dogs whose owner is the now deleted Joe. This seems to also be due to the implicit comma join
```
select
	dog0_.Name as col_0_0_
from
	"Dog" dog0_,
	"Human" human1_
where
	dog0_.Human_id=human1_.Id and human1_.Id=@p0;
@p0 = 1 [Type: Int32 (0:0:0)]
```

Interestingly enough, after Joe has been removed, and you remove the NotFound.Ignore property on the dog mapping, then the query using Joe's old id works. This seems like it may be a workaround if you don't need to delete new entities.

I've attached a minimal working example using NHibernate 5.3.10 and FluentNhibernate that demonstrates these issues
[NHibernateTest.zip](https://github.com/nhibernate/nhibernate-core/files/7866051/NHibernateTest.zip)

bahusoid:
Fixes #2988

### DIFF
--- a/src/NHibernate.Test/Async/Hql/EntityJoinHqlTest.cs
+++ b/src/NHibernate.Test/Async/Hql/EntityJoinHqlTest.cs
@@ -349,11 +349,23 @@ namespace NHibernate.Test.Hql
 					from x2 in session.Query<NullableOwner>()
 					where x == x2 && x.ManyToOne == null && x.OneToOne.Name == null
 					select x2).ToListAsync());
+
+				var validManyToOne = new OneToOneEntity{Name = "valid"};
+				await (session.SaveAsync(validManyToOne));
+				nullableOwner2.ManyToOne = validManyToOne;
+				await (session.FlushAsync());
+
+				//GH-2988
+				var withNullOrValidList = await (session.Query<NullableOwner>().Where(x => x.ManyToOne.Id == validManyToOne.Id || x.ManyToOne == null).ToListAsync());
+				var withNullOrValidList2 = await (session.Query<NullableOwner>().Where(x =>  x.ManyToOne == null || x.ManyToOne.Id == validManyToOne.Id).ToListAsync());
+
 				Assert.That(fullList.Count, Is.EqualTo(2));
 				Assert.That(withValidManyToOneList.Count, Is.EqualTo(0));
 				Assert.That(withValidManyToOneList2.Count, Is.EqualTo(0));
 				Assert.That(withNullManyToOneList.Count, Is.EqualTo(2));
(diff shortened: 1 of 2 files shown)


src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
```diff
@@ -417,13 +417,14 @@ private void DereferenceEntity(EntityType entityType, bool implicitJoin, string
 
 			if ( joinIsNeeded )
 			{
+				bool forceJoinType = false;
 				if (comparisonWithNullableEntity && Walker.IsNullComparison)
 				{
-					implicitJoin = false;
 					_joinType = JoinType.LeftOuterJoin;
+					forceJoinType = true;
 				}
```
hazzik, line 425:
```suggestion
				var forceJoinType = comparisonWithNullableEntity && Walker.IsNullComparison;
				if (forceJoinType) 
				{
					_joinType = JoinType.LeftOuterJoin
				}
```

fredericDelaporte (approved):

PR #2990, opened 2022-01-14
Use inner join instead of implicit join for implied entity joins

bahusoid:
Should be safe with table group join support


src/NHibernate/NHibernate.csproj
```diff
@@ -12,7 +12,6 @@
 
     <PackageDescription>NHibernate is a mature, open source object-relational mapper for the .NET framework. It is actively developed, fully featured and used in thousands of successful projects.</PackageDescription>
     <PackageTags>ORM; O/RM; DataBase; DAL; ObjectRelationalMapping; NHibernate; ADO.Net; Core</PackageTags>
-    <LangVersion>7.3</LangVersion>
```
hazzik, line 15: Need to explicitly specify language version. I propose use 10.

bahusoid: Maybe should be added to NHibernate.props instead

bahusoid: latestMajor (default) | The compiler accepts syntax from the latest released major version of the compiler.

I don't understand why it doesn't work by default. Shouldn't latestMajor be 10.0 too?

hazzik: > I don't understand why it doesn't work by default. Shouldn't latestMajor be 10.0 too?

It is tied to the target framework somehow.

bahusoid: That explains it, though it's not what description says... Some strange compilations issues with 10.0  Let's start with 9.0 for now..

src/NHibernate.Test/Legacy/ParentChildTest.cs
```diff
@@ -372,10 +372,15 @@ public void CollectionQuery()
...
 			c.ManyToMany = l;
+			c.One = new One { Value = "one" };
+			s.Save(c.One);
 			s.Save(c);
 
 			Assert.AreEqual(1,
 			                s.CreateQuery("select c from c in class ContainerX, s in class Simple where c.OneToMany[2] = s").List
+			                	().Count);			
+			Assert.AreEqual(1,
+			                s.CreateQuery("select c from c in class ContainerX, s in class Simple where c.OneToMany[2] = s and c.One != null").List
```
bahusoid, line 383: I'm a bit lost.. This check works on my machine with both Sql Server and SQLite.. How come it fails on CI???

hazzik: Do you run .net 6 tests? Or .net 4.6.1?

bahusoid: net6.0

bahusoid: I'm still puzzled why original tests fails on linux (seems it's the only difference). But fail is not related to PR changes (as this check is about proper SQL generation with valid order of joins)

src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
```diff
@@ -519,7 +519,7 @@ private void DereferenceEntityJoin(string classAlias, EntityType propertyType, b
 				// If this is an implied join in a from element, then use the impled join type which is part of the
 				// tree parser's state (set by the gramamar actions).
 				JoinSequence joinSequence = SessionFactoryHelper
-					.CreateJoinSequence( impliedJoin, propertyType, tableAlias, _joinType, joinColumns );
+					.CreateJoinSequence(false, propertyType, tableAlias, _joinType, joinColumns);
```
fredericDelaporte, line 522: The comment is no more relevant and should be removed.

fredericDelaporte (approved):

hazzik (approved):

PR #2992, opened 2022-01-16
Fix schema validator for Npgsql 5

### LINKED ISSUE #2876: Schema validation not working with NpgSql v5
Database schema auto validation is not working after the update of NpgSql to version 5.0.0.
(Enabled e.g by: `properties.SchemaAction = SchemaAutoAction.Validate;`)
The validation always fails with errors like:
```
Wrong column type in test.systems for column ID. Found: **integer**, Expected **int4**; 
Wrong column type in test.systems for column Name. Found: **character varying**, Expected **varchar(30)**;
...
```

Exception stack:
```
Received unexpected error: Schema validation failed: see list of validation errors.
   at NHibernate.Cfg.Configuration.ValidateSchema(Dialect dialect, IDatabaseMetadata databaseMetadata)
   at NHibernate.Tool.hbm2ddl.SchemaValidator.Validate()
   at NHibernate.Impl.SessionFactoryImpl..ctor(Configuration cfg, IMapping mapping, Settings settings, EventListeners listeners)
   at NHibernate.Cfg.Configuration.BuildSessionFactory()
   at NHibernate.NetCore.ServiceCollectionExtensions.<>c.<AddHibernate>b__5_1(IServiceProvider provider)
```  
Problem description:
The code in Table.cs `Table.ValidateColumns` 
https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Mapping/Table.cs
is comparing column types:
```c#
bool typesMatch = column.GetSqlType(dialect, mapping).StartsWith(columnInfo.TypeName, StringComparison.OrdinalIgnoreCase);
```

The `columnInfo.TypeName` is getting column type from NpgSql `NpgsqlSchema.GetColumns`
https://github.com/npgsql/npgsql/blob/main/src/Npgsql/NpgsqlSchema.cs
where the value is taken from `data_type` column from SQL query:
```sql
SELECT
    table_catalog, table_schema, table_name, column_name, ordinal_position, column_default, is_nullable,
    udt_name::regtype::text AS data_type, character_maximum_length, character_octet_length, numeric_precision,
    numeric_precision_radix, numeric_scale, datetime_precision, character_set_catalog, character_set_schema,
    character_set_name, collation_catalog
  FROM information_schema.columns");
```

Unfortunately in version 5.0.0, there were https://github.com/npgsql/npgsql/pull/3012, which changed the returned value from `udt_name AS data_type` to `udt_name::regtype::text AS data_type`.
https://github.com/npgsql/npgsql/commit/90366685b88456446988ca087ec9c1a53e0613bd#diff-9e5577d0cfb25bc8279488c960f8dc2427d580f90156386cda9d14953caafd1d
This change caused, that "user-facing type names" are now returned from the database.

Here are some example results returned from SQL query: 
```sql
SELECT udt_name AS old_data_type, udt_name::regtype::text AS new_data_type
  FROM information_schema.columns
```

```
old_data_type	new_data_type
int8          	bigint
int4          	integer
char          	character
varchar       	character varying
bool          	boolean
```

Please, fix this issue.

hazzik:
Npgsql 5 returns different type names for some of the types. Normalize them back to what dialect expects.  

Fixes #2876

### DIFF
--- a/src/NHibernate/Dialect/Schema/PostgreSQLMetadata.cs
+++ b/src/NHibernate/Dialect/Schema/PostgreSQLMetadata.cs
@@ -138,7 +138,43 @@ namespace NHibernate.Dialect.Schema
 			this.SetNumericalPrecision(rs["NUMERIC_PRECISION"]);
 
 			Nullable = Convert.ToString(rs["IS_NULLABLE"]);
-			TypeName = Convert.ToString(rs["DATA_TYPE"]);
+			TypeName = NormalizeTypeNames(Convert.ToString(rs["DATA_TYPE"]));
+		}
+
+		private static string NormalizeTypeNames(string typeName)
+		{
+			switch (typeName)
+			{
+				case "double precision":
+					return "float8";
+				case "real":
+					return "float4";
+				case "smallint":
+					return "int2";
+				case "integer":
+					return "int4";
+				case "bigint":
+					return "int8";
+			}
+
+			if (typeName.StartsWith("character", StringComparison.Ordinal))
+			{
+				return typeName.Replace("character varying", "varchar").Replace("character", "char");
+			}
+
+			if (typeName.EndsWith(" with time zone", StringComparison.Ordinal))
+			{
+				return typeName.StartsWith("timestamp", StringComparison.Ordinal)
+					? typeName.Replace(" with time zone", string.Empty).Replace("timestamp", "timestamptz")
+					: typeName.Replace(" with time zone", string.Empty).Replace("time", "timetz");
+			}
+
+			if (typeName.EndsWith(" without time zone", StringComparison.Ordinal))
+			{
+				return typeName.Replace(" without time zone", string.Empty);
+			}
+
+			return typeName;
 		}
 	}
 
--- a/src/NHibernate.Test/NHibernate.Test.csproj
+++ b/src/NHibernate.Test/NHibernate.Test.csproj
@@ -64,7 +64,7 @@
     <PackageReference Include="NUnit3TestAdapter" Version="3.13.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.1.1" />
     <PackageReference Include="FirebirdSql.Data.FirebirdClient" Version="6.6.0" />
-    <PackageReference Include="Npgsql" Version="4.0.3" />
+    <PackageReference Include="Npgsql" Version="5.0.11" />
   </ItemGroup>
   <ItemGroup Condition="$(NhNetFx)">
     <Reference Include="System.Configuration" />


hazzik:
Imo this is the least intrusive approach that works for both Npgsql 5 and Npgsql 4. I was considering doing other way around (eg normalize to Npgsql 5 type system. But that require more changes and could be a breaking changes due to changes of the dialect for some users (if, eg, they do textual schema compare).

src/NHibernate/Dialect/Schema/PostgreSQLMetadata.cs
```diff
@@ -138,7 +138,43 @@ public PostgreSQLColumnMetadata(DataRow rs)
...
+					return "float4";
+				case "smallint":
+					return "int2";
+				case "integer":
+					return "int4";
+				case "bigint":
+					return "int8";
+			}
+
+			if (typeName.StartsWith("character"))
```
bahusoid, line 160: Please use `StringComparison.Ordinal` with `StartsWith` and `EndsWith`

src/NHibernate/Dialect/Schema/PostgreSQLMetadata.cs
```diff
@@ -138,7 +138,43 @@ public PostgreSQLColumnMetadata(DataRow rs)
...
+			{
+				return typeName.Replace("character varying", "varchar").Replace("character", "char");
+			}
+
+			if (typeName.EndsWith("with time zone"))
+			{
+				return typeName.StartsWith("timestamp")
+					? typeName.Replace("with time zone", "").Replace("timestamp", "timestamptz").Trim()
```
bahusoid, line 168: Can't we replace `" with time zone"` (starting with space) so Trim would be no op

hazzik: Done

src/NHibernate/Dialect/Schema/PostgreSQLMetadata.cs
hazzik: Done

src/NHibernate/Dialect/Schema/PostgreSQLMetadata.cs
```diff
@@ -138,7 +138,43 @@ public PostgreSQLColumnMetadata(DataRow rs)
...
+			}
+
+			if (typeName.StartsWith("character", StringComparison.Ordinal))
+			{
+				return typeName.Replace("character varying", "varchar").Replace("character", "char");
+			}
+
+			if (typeName.EndsWith(" with time zone", StringComparison.Ordinal))
+			{
+				return typeName.StartsWith("timestamp")
```
bahusoid, line 167:
```suggestion
				return typeName.StartsWith("timestamp", StringComparison.Ordinal)
```

bahusoid (approved):

hazzik:
@bahusoid, @fredericDelaporte are you ok with including this to 5.3.x?

bahusoid:
> including this to 5.3.x

Yeah. Why not...

hazzik:
> Yeah. Why not...

Rebased to 5.3.x

src/NHibernate/Dialect/Schema/PostgreSQLMetadata.cs
```diff
@@ -138,7 +138,43 @@ public PostgreSQLColumnMetadata(DataRow rs)
...
+					return "float8";
+				case "real":
+					return "float4";
+				case "smallint":
+					return "int2";
+				case "integer":
+					return "int4";
+				case "bigint":
+					return "int8";
+			}
```
fredericDelaporte, line 158: We can see the naming changes in [Npgsql change commit](https://github.com/npgsql/npgsql/pull/1945/files?diff=split&w=1), or infer it from [PostgreSQL documentation](https://www.postgresql.org/docs/current/datatype.html#DATATYPE-TABLE).

It looks like we were already having an issue with older Npgsql drivers, because we were using the complete name, `boolean`, while older drivers were previously supposed to yield the short name, `bool`. That trouble would be fixed since we do not translate it to its short name here.

But then we keep an inconsistency: the dialect and now the metadata are using PostgreSQL type short aliases for most types excepted for the boolean type.

For the same reasons than in this [comment](https://github.com/nhibernate/nhibernate-core/pull/2992#issuecomment-1013960183), normalizing this would be breaking. So, let it stay that way.

src/NHibernate/Dialect/Schema/PostgreSQLMetadata.cs
```diff
@@ -138,7 +138,43 @@ public PostgreSQLColumnMetadata(DataRow rs)
...
+
+			if (typeName.StartsWith("character", StringComparison.Ordinal))
+			{
+				return typeName.Replace("character varying", "varchar").Replace("character", "char");
```
fredericDelaporte, line 162: It is likely we are fixing another old trouble by the way here: for fixed length character types, the older Npgsql drivers could yield `bpchar`, while our dialects are using `char`.

fredericDelaporte (approved):

fredericDelaporte:
I am used to the "auto-merge" feature on Azure DevOps, but not here on GitHub. It was quite a surprise for me. It lacks visibility I think.

PR #2997, opened 2022-01-18
WIP Fix validating similar but not equal column types

### LINKED ISSUE #1895: Column data type validation not quite exact
With a PostgreSql database, when I run a **schema validation** the following happens:

- Property mapped as `timestamp` against a `timestamptz` column => Schema Validation Failed (as expected)
- Property mapped as `timestamptz` against a `timestamp` column => Schame Validation OK (**not** as expected)

After cheching the repository code, I found where the error might be (src/NHibernate/Mapping/Table.cs, method `ValidateColumns`:

`bool typesMatch = column.GetSqlType(dialect, mapping).StartsWith(columnInfo.TypeName, StringComparison.OrdinalIgnoreCase);`

Is it correct to check if both types match (mapping and column) with `.StartsWith(...)`?
Wouldn't be better to use `.Equals(...)` here?

hazzik:
Just fixes similar column types checks for precision and scale is not implemented, because:

_From @fredericDelaporte:_
> Validating length/scale/precision when available should be done as an improvement, with a dedicated PR.

Fixes #1895


src/NHibernate/Mapping/Table.cs
```diff
@@ -1035,7 +1036,10 @@ public IEnumerable<string> ValidateColumns(Dialect.Dialect dialect, IMapping map
 				}
 
 				//TODO: Add new method to ColumnMetadata :getTypeCode
-				bool typesMatch = column.GetSqlType(dialect, mapping).StartsWith(columnInfo.TypeName, StringComparison.OrdinalIgnoreCase);
```
bahusoid, line 1038: Why not simply check that next char is not a letter (`char.IsLetter`)?

hazzik: I was/am thinking about that. Even have code about to push. But it is not that simple.

bahusoid: Still see no reasons to use regex - check additionally for not a digit (`char.IsDigit`). Alphanumeric character check should suffice

hazzik: Yes, I hate the regex myself. But that is not that easy really. For example, Db2's `CHAR(16) FOR BIT DATA` would match to `CHAR(16)` even in my implementation with regex.

PR #3027, opened 2022-02-23
Avoid allocations on lock in SyncCacheLock

bahusoid:


src/NHibernate/Cache/SyncCacheLock.cs
```diff
@@ -6,34 +6,46 @@ namespace NHibernate.Cache
...
 		public IDisposable ReadLock()
 		{
-			return new MonitorLock(this);
+			return _monitorLock.Lock();
 		}
 
 		public IDisposable WriteLock()
 		{
-			return new MonitorLock(this);
+			return _monitorLock.Lock();
```
hazzik, line 48: But then it would be same lock for read and write. Is this expected?

bahusoid: Yes. It's the same lock for read and write without this change. This locker doesn't distinguish read and write.

hazzik: What about this?


	class SyncCacheLock : ICacheLock
	{
		public void Dispose()
		{
			Monitor.Exit(this);
		}

		public IDisposable ReadLock()
		{
			return Lock();
		}

		public IDisposable WriteLock()
		{
			return Lock();
		}

                private IDisposable Lock()
		{
                        Monitor.Enter(this);
			return this;
		}

                //....

hazzik (approved):

src/NHibernate/Cache/SyncCacheLock.cs
bahusoid: No. ICacheLock is disposed separately from locks.

Issue #3029, opened 2022-03-02
StaticProxyFactory.GetProxy throws InvalidOperationException when the object has explicit implementation of a generic method


yyordanov:
Consider the following domain model (mapping is provided in the attached demo project):

    public class Parent
    {
        public virtual int Id { get; set; }
        public virtual Generic Generic { get; set; }
    }

    public class Generic : IGeneric
    {
        public virtual int Id { get; set; }

        T[] IGeneric.GetTypes<T>()
        {
            return null;
        }
    }

    public interface IGeneric
    {
        T[] GetTypes<T>();
    }

When trying to load an instance of `Parent` via `Session.Get`, the exception below occurs. The problem occurs when the class being proxied implements a generic interface method explicitly. The problem doesn't occur with implicit implementation. 
The issue is introduced with NHibernate 5.3.0, so it should be treated as a regression. It is valid for the latest version as well (5.3.11). 5.2.7 and lower work OK.

It seems the problem is within StaticProxyFactoryFactory or its internal implementation. The problem doesn't exist when using DefaultStaticProxyFactoryFactory under .NET 4.8. Unfortunately, this is not a valid solution for .NET 5 / 6 because DefaultProxyFactoryFactory throws "Bad IL format" -> it is not compatible with .NET 5 / 6

A demo project is attached.
[ProxyProblem.zip](https://github.com/nhibernate/nhibernate-core/files/8167797/ProxyProblem.zip)

```
NHibernate.HibernateException: Creating a proxy instance failed
 ---> System.InvalidOperationException: The signature of the MethodBuilder can no longer be modified because an operation on the MethodBuilder caused the methodDef token to be created. For example, a call to SetCustomAttribute requires the methodDef token to emit the CustomAttribute token.
   at System.Reflection.Emit.MethodBuilder.DefineGenericParameters(String[] names)
   at NHibernate.Proxy.ProxyBuilderHelper.GenerateMethodSignature(String name, MethodInfo method, TypeBuilder typeBuilder)
   at NHibernate.Proxy.NHibernateProxyBuilder.ImplementCallMethodOnImplementation(TypeBuilder typeBuilder, MethodInfo method, FieldInfo lazyInitializerField, Type parentType)
   at NHibernate.Proxy.NHibernateProxyBuilder.CreateProxiedMethod(TypeBuilder typeBuilder, MethodInfo method, FieldInfo lazyInitializerField, Type parentType)
[...]

PR #3034, opened 2022-03-13
Proxy creation fail for class with explicit generic interface

fredericDelaporte:
Add test for #3029


fredericDelaporte:
This test runs well on 5.2.x branch, so I confirm this is a 5.3.x regression.

fredericDelaporte:
It was a regression caused by #2264.

hazzik (approved):

PR #3035, opened 2022-03-13
Supports tests in VS 2022

fredericDelaporte:
Tests fail to run in VS 2022, which is unpractical.

### DIFF
--- a/build-common/NHibernate.props
+++ b/build-common/NHibernate.props
@@ -13,7 +13,7 @@
     <FileVersion Condition="'$(VersionSuffix)' != '' AND '$(BuildNumber)' != ''">$(VersionPrefix).$(BuildNumber)</FileVersion>
     <FileVersion Condition="'$(FileVersion)' == ''">$(VersionPrefix).0</FileVersion>
 
-    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;netcoreapp2.0</NhAppTargetFrameworks>
+    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;netcoreapp2.1</NhAppTargetFrameworks>
     <NhLibTargetFrameworks Condition ="$(NhLibTargetFrameworks) == ''">net461;netcoreapp2.0;netstandard2.0</NhLibTargetFrameworks>
     <NhNetFx>false</NhNetFx>
     <NhNetFx Condition="$(TargetFramework.StartsWith('net4'))">true</NhNetFx>
--- a/build-common/NHibernate.props
+++ b/build-common/NHibernate.props
@@ -13,11 +13,13 @@
     <FileVersion Condition="'$(VersionSuffix)' != '' AND '$(BuildNumber)' != ''">$(VersionPrefix).$(BuildNumber)</FileVersion>
     <FileVersion Condition="'$(FileVersion)' == ''">$(VersionPrefix).0</FileVersion>
 
-    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;netcoreapp2.1</NhAppTargetFrameworks>
+    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;netcoreapp2.0;netcoreapp2.1</NhAppTargetFrameworks>
     <NhLibTargetFrameworks Condition ="$(NhLibTargetFrameworks) == ''">net461;netcoreapp2.0;netstandard2.0</NhLibTargetFrameworks>
     <NhNetFx>false</NhNetFx>
     <NhNetFx Condition="$(TargetFramework.StartsWith('net4'))">true</NhNetFx>
     <DefineConstants Condition="$(NhNetFx) AND $(NhVbNet) == ''" >NETFX;$(DefineConstants)</DefineConstants>
+    <NhNetCoreApp>false</NhNetCoreApp>
+    <NhNetCoreApp Condition="$(TargetFramework.StartsWith('netcoreapp'))">true</NhNetCoreApp>
 
     <Product>NHibernate</Product>
     <Company>NHibernate.info</Company>
--- a/build-common/NHibernate.props
+++ b/build-common/NHibernate.props
@@ -20,6 +20,7 @@
     <DefineConstants Condition="$(NhNetFx) AND $(NhVbNet) == ''" >NETFX;$(DefineConstants)</DefineConstants>
     <NhNetCoreApp>false</NhNetCoreApp>
     <NhNetCoreApp Condition="$(TargetFramework.StartsWith('netcoreapp'))">true</NhNetCoreApp>
+    <DefineConstants Condition="$(NhNetCoreApp) AND $(NhVbNet) == ''" >NETCOREAPP2_0_OR_GREATER;$(DefineConstants)</DefineConstants>
 
     <Product>NHibernate</Product>
     <Company>NHibernate.info</Company>
--- a/src/AsyncGenerator.yml
+++ b/src/AsyncGenerator.yml
@@ -183,7 +183,7 @@
     scanForMissingAsyncMembers:
     - all: true
 - filePath: NHibernate.Test/NHibernate.Test.csproj
-  targetFramework: netcoreapp2.0
+  targetFramework: netcoreapp2.1
   concurrentRun: true
   applyChanges: true
   suppressDiagnosticFailures:
--- a/src/AsyncGenerator.yml
+++ b/src/AsyncGenerator.yml
@@ -183,7 +183,7 @@
     scanForMissingAsyncMembers:
     - all: true
 - filePath: NHibernate.Test/NHibernate.Test.csproj
-  targetFramework: netcoreapp2.1
+  targetFramework: netcoreapp2.0
   concurrentRun: true
   applyChanges: true
   suppressDiagnosticFailures:
--- a/src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
+++ b/src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
@@ -1372,6 +1372,20 @@ namespace NHibernate.Test.Linq
 			}
 		}
 
+		[Test]
+		public void ReplaceFunctionWithNullArgumentAsync()
+		{
+			var query = from e in db.Employees
+			            select e.FirstName.Replace(e.LastName, null);
+			List<string> results = null;
+			Assert.That(
+				async () =>
+				{
+					results = await (query.ToListAsync());
+				}, Throws.Nothing, "Expected REPLACE(FirstName, LastName, NULL) to be supported");
+			Assert.That(results, Is.Not.Null);
+		}
+
 		[Test(Description = "GH-2860")]
 		public async Task StringFormatWithTrimAsync()
 		{
--- a/src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
+++ b/src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
@@ -1372,20 +1372,6 @@ namespace NHibernate.Test.Linq
 			}
 		}
 
-		[Test]
-		public void ReplaceFunctionWithNullArgumentAsync()
-		{
-			var query = from e in db.Employees
-			            select e.FirstName.Replace(e.LastName, null);
-			List<string> results = null;
-			Assert.That(
-				async () =>
-				{
-					results = await (query.ToListAsync());
-				}, Throws.Nothing, "Expected REPLACE(FirstName, LastName, NULL) to be supported");
-			Assert.That(results, Is.Not.Null);
-		}
-
 		[Test(Description = "GH-2860")]
 		public async Task StringFormatWithTrimAsync()
 		{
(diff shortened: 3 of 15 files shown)


build-common/NHibernate.props
```diff
@@ -13,7 +13,7 @@
     <FileVersion Condition="'$(VersionSuffix)' != '' AND '$(BuildNumber)' != ''">$(VersionPrefix).$(BuildNumber)</FileVersion>
     <FileVersion Condition="'$(FileVersion)' == ''">$(VersionPrefix).0</FileVersion>
 
-    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;netcoreapp2.0</NhAppTargetFrameworks>
+    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;netcoreapp2.1</NhAppTargetFrameworks>
```
fredericDelaporte, line 16: In order to run test in VS 2022, the NUnit test adapter has to be updated to its fourth version at least. But this version does no more support target netcoreapp2.0, and requires at least netcoreapp2.1.

May we have better to directly target net6.0 for our tests?

src/AsyncGenerator.yml
```diff
@@ -183,7 +183,7 @@
     scanForMissingAsyncMembers:
     - all: true
 - filePath: NHibernate.Test/NHibernate.Test.csproj
-  targetFramework: netcoreapp2.0
+  targetFramework: netcoreapp2.1
```
fredericDelaporte, line 186: I have upgrade only the test project. Unfortunately, it seems the async generator does not support generating a project for core 2.1 which depends on core 2.0 libs : it tries to compile the 2.0 libs as 2.1 libs.
@maca88 

A workaround would be to add a core 2.1 target to the NHibernate libs too, but just for being able able to runs tests in VS2022, it sounds overkill to me.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -1,7 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-#if NETCOREAPP2_0
+#if NETCOREAPP2_0 || NETCOREAPP2_1
```
fredericDelaporte, line 4: This file is directly included as source in the test project too. So here we need to handle core 2.0 and 2.1 compilation.

build-common/NHibernate.props
hazzik: I think 2.1 should be enough.

build-common/NHibernate.props
```diff
@@ -13,11 +13,13 @@
     <FileVersion Condition="'$(VersionSuffix)' != '' AND '$(BuildNumber)' != ''">$(VersionPrefix).$(BuildNumber)</FileVersion>
     <FileVersion Condition="'$(FileVersion)' == ''">$(VersionPrefix).0</FileVersion>
 
-    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;netcoreapp2.0</NhAppTargetFrameworks>
+    <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net461;netcoreapp2.0;netcoreapp2.1</NhAppTargetFrameworks>
```
fredericDelaporte, line 16: To work around the async generation issue, I target both net core 20 and 2.1.
The test explorer in Visual Studio 2022 then show tests for three targets, net461, netcoreapp2.0 and 2.1. But it can run only tests for net461 and netcoreapp2.1. This is good enough I think.

src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
```diff
@@ -1372,20 +1372,6 @@ public async Task DLinqJoin10aLeftJoinAsync()
...
-			            select e.FirstName.Replace(e.LastName, null);
-			List<string> results = null;
-			Assert.That(
-				async () =>
-				{
-					results = await (query.ToListAsync());
-				}, Throws.Nothing, "Expected REPLACE(FirstName, LastName, NULL) to be supported");
-			Assert.That(results, Is.Not.Null);
-		}
-
```
fredericDelaporte, line 1388: I have no idea why this async generation change occurs...

src/NHibernate.Test/Linq/WhereTests.cs
```diff
@@ -432,7 +432,7 @@ public void StringComparisonParamEmitsWarning()
 					AssertStringComparisonWarning(x => x.CustomerId.EndsWith("ANATR", StringComparison.Ordinal), 1);
 					AssertStringComparisonWarning(x => x.CustomerId.IndexOf("ANATR", StringComparison.Ordinal) == 0, 1);
 					AssertStringComparisonWarning(x => x.CustomerId.IndexOf("ANATR", 0, StringComparison.Ordinal) == 0, 1);
-#if NETCOREAPP2_0
+#if NETCOREAPP2_0 || NETCOREAPP2_1
```
fredericDelaporte, line 435: I have tried with just `#if  NETCOREAPP` instead, but it looks like the async generator does not handle it.

src/AsyncGenerator.yml
maca88: > Unfortunately, it seems the async generator does not support generating a project for core 2.1 which depends on core 2.0 libs : it tries to compile the 2.0 libs as 2.1 libs.

I do apologize as I didn't saw this earlier. Unfortunately, I didn't found a workaround yet that would not require to add `netcoreapp2.1` to other projects.

> A workaround would be to add a core 2.1 target to the NHibernate libs too

Another workaround would be to use `netcoreapp3.1` for the test project which works fine as opposed to `netcoreapp2.1`.

src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
maca88: After some debugging I found out that the problem seems to be with Roslyn which loads incorrectly the NUnit `Assert.Than` symbols marking async counterparts parameters as nullable but not sync ones (`object?[]` vs `object[]`) . This seems to be an issue only for Roslyn version `3.0.0` as with the newer versions (e.g. `3.7.0`) it marks both as nullable. As Roslyn version `3.7.0` is not supported for `netcoreapp2.1`, I will add a loosely matching so that it will still match despite the nullable differences.

src/NHibernate.Test/Linq/WhereTests.cs
maca88: For me `NETCOREAPP` is working fine, which .NET Core sdk version are you using? I think that `NETCOREAPP` [was added](https://github.com/dotnet/sdk/pull/2073) in `2.1.3XX`.

src/AsyncGenerator.yml
fredericDelaporte: No apologies are required! We are all contributors on our free time, no one is supposed to have any "SLA" with short response time, ...

Thanks for your investigation. So you say it work well to generate async code for a project targeting 3.1 and referencing project targeting 2.0?

That would be better than my current workaround, targeting both 2.0 and 2.1 in the test project and generating async for 2.0. @hazzik, are you agree?

src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
fredericDelaporte: So this is triggered by trageting 2.1 in the test project.Will it also happen with 3.1? If no, then it will be one reason more to target 3.1 in the test project instead of 2.1.

src/NHibernate.Test/Linq/WhereTests.cs
fredericDelaporte: Sdk 2.0.0, 3.1.417, 5.0.406, 6.0.201. So, I removed that old sdk 2: no more async generation. I have installed a 2.2 SDK: it works fine now, thanks. (But that was not an issue for compiling with VS: strange.)

src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
maca88: > Will it also happen with 3.1?

Yes, the same issue is present when the test project is targeting `netcoreapp3.1`, not sure why, but only happens for Roslyn version `3.0.0.`. I released a new version `0.18.4` which fixes this issue by using a loosely matching when searching for async counterparts.

src/AsyncGenerator.yml
maca88: > So you say it work well to generate async code for a project targeting 3.1 and referencing project targeting 2.0?

That is correct, [here](https://github.com/maca88/nhibernate-core/commit/a86287916d2e35045428f8bde1ded3e8a1f63feb) you can find the changes that I used to test `3.1`.

src/NHibernate.Test/Async/Linq/LinqQuerySamples.cs
fredericDelaporte: Thanks again.

build-common/NHibernate.props
hazzik: What is the issue with generating 2.0 and 2.1?

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
hazzik, line 4: Let's do `NETCOREAPP2_0_OR_GREATER` to minimize changes to master.

src/NHibernate.TestDatabaseSetup/Program.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if NETCOREAPP2_0
+﻿#if NETCOREAPP
```
hazzik, line 1: Let's do `NETCOREAPP2_0_OR_GREATER` to minimize changes to master.

src/NHibernate.TestDatabaseSetup/NHibernate.TestDatabaseSetup.csproj
```diff
@@ -7,19 +7,24 @@
...
     <ProjectReference Include="..\NHibernate.Test\NHibernate.Test.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="NUnit" Version="3.12.0" />
-    <PackageReference Include="NUnit3TestAdapter" Version="3.13.0" />
+    <PackageReference Include="NUnit" Version="3.13.2" />
   </ItemGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
-    <PackageReference Include="NUnitLite" Version="3.12.0" />
+  <ItemGroup Condition="$(NhNetFx)">
```
hazzik, line 20: This and 2 following groups are confusing, I'm not sure I understand the intention.

src/NHibernate.TestDatabaseSetup/NHibernate.TestDatabaseSetup.csproj
```diff
@@ -7,19 +7,24 @@
...
+    <PackageReference Include="NUnit3TestAdapter" Version="4.2.1" />
```
hazzik, line 21: Let's use same version as in master.

build-common/NHibernate.props
fredericDelaporte: It causes the test explorer to offer to run tests for 2.0, but that does not work, by lack of an adequate test adapter.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
fredericDelaporte: This requires a higher SDK. According to the [documentation](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-a-target-framework), a .net 5 SDK is required for these `_OR_GREATER` symbols to be supported.
It may be an issue for the async generator.

src/NHibernate.TestDatabaseSetup/NHibernate.TestDatabaseSetup.csproj
fredericDelaporte: 4.1.0? Why not, although I do not see the point of not upgrading this dependency to 4.2.1 here then in master when merging 5.3.x.

src/NHibernate.TestDatabaseSetup/NHibernate.TestDatabaseSetup.csproj
fredericDelaporte: netcoreapp2.0 cannot compile with a reference to NUnit3TestAdapter. So it has to be referenced only for netFx and 2.1.

Maybe I should not mix $(NhNetCoreApp) references (for both 2.0 and 2.1) and explicit 2.1 references, for easing understanding what is done.

Or even maybe I could just drop the NUniteLite reference for 2.0 too, I have not checked.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
hazzik: I know that it requires 5.4 sdk, but I think we can register them manually. This would be better than resolve tens of conflicts.

fredericDelaporte: ok

src/NHibernate.TestDatabaseSetup/NHibernate.TestDatabaseSetup.csproj
fredericDelaporte: I cannot remove NUniteLite reference for netcoreap2.0, or else I would have to use a `NETCOREAPP2_1` only condition in test program files, instead of using the `NETCOREAPP2_0_OR_GREATER` condition already in master.

So I have changed for using explicite target framework references for netcore, which duplicates quite a few references for the main test project.

hazzik (approved):

Issue #2549, opened 2020-09-17
ApplyFilter does not work on join statements in LINQ


esabah:
Hi 
We upgraded NH from 5.2.7 to 5.3.3.
After upgrade we had an issue that ApplyFilter  does not work with joins.
Same join statement working fine with version 5.2.7.

You can find testcase implementation below link.
https://github.com/esabah/NHibernateBugTest/blob/master/NHibernateBugTest/TxnDefTest.cs

TxnDefMember entity has a custom filter which applies filter on mbrId field.

`         ApplyFilter<MemberConditionFilter>();`

But mbrId  not included in sql query in 5.3.3 version.
`(from txndef in session.Query<TxnDef>()
join txnDefMember in session.Query<TxnDefMember>() on txndef.Guid equals txnDefMember.Id.TxnDefGuid
select txnDefMember.BucketName)`

same query returns 8 records in 5.3.3
and returns 4 records in 5.2.7

bahusoid:
Can you check how it works with latest dev build?
https://www.myget.org/feed/nhibernate/package/nuget/NHibernate

esabah:
I works correctly with NHibernate 5.4.0-dev.3147.

bahusoid:
Good. Thanks for report

gokhanabatay:
Would it be better to add Active Filters in HasMany Relations by default? Why do we force adding filters with ApplyFilter <T>?

PR #3048, opened 2022-04-24
Fix many-to-one disabled filters for entity joins

### LINKED ISSUE #3046: Regression for filters on entity joins with many-to-one disabled
I'm trying to migrate my project from v5.2.7 to the latest 5.3.x version but apparently I've hit a potential regression / breaking change on how filters are applied.

My filter is defined like this: 

```
            configuration.AddFilterDefinition(new FilterDefinition(
                LabelFilterName,
                "Label IN ('A', 'B', 'C')",
                new Dictionary<string, IType>(),
                false));
```

On "simple" queries it is applied as expected, but on a more complex one it is not: 

```
            var query = from b in session.Query<OtherTable>() 
                        join jta in session.Query<JoinTable>() on b.Id equals jta.OtherTableId
                        join ft in session.Query<FilteredTable>() on jta.FilteredTableId equals ft.Id
                        select new
                        {
                            jta.FilteredTableId,
                            ft.Label
                        };
```

I've attached a small repro to demonstrate the issue: on 5.2.7 both unit tests are green, but on 5.3.11 one of them fails.

[repro.zip](https://github.com/nhibernate/nhibernate-core/files/8548584/repro.zip)

fredericDelaporte:
fix #3046


src/NHibernate.Test/NHSpecificTest/GH3046/Mappings.hbm.xml
```diff
@@ -0,0 +1,26 @@
...
+    </class>
+
+    <class name="Person">
+        <id name="Id">
+            <generator class="assigned" />
+        </id>
+        <property name="Name" />
+    </class>
+
+    <filter-def name="DeletedCustomer" use-many-to-one="false">
```
fredericDelaporte, line 22: `use-many-to-one="false"` is the only change compared to #2549 tests.

src/NHibernate/Engine/JoinSequence.cs
```diff
@@ -175,7 +175,7 @@ public JoinFragment ToJoinFragment(IDictionary<string, IFilter> enabledFilters,
 			{
 				Join join = joins[i];
 				string on = join.AssociationType.GetOnCondition(join.Alias, factory, enabledFilters);
-				SqlString condition = new SqlString();
+				SqlString condition;
```
fredericDelaporte, line 178: Little cleanup by the way.

src/NHibernate/Engine/JoinSequence.cs
```diff
@@ -175,7 +175,7 @@ public JoinFragment ToJoinFragment(IDictionary<string, IFilter> enabledFilters,
 			{
 				Join join = joins[i];
 				string on = join.AssociationType.GetOnCondition(join.Alias, factory, enabledFilters);
```
fredericDelaporte, line 177: I initially though the `GetOnCondition` implementation may be troublesome in some cases, because it does filter `enabledFilters` according to their `use-many-to-one` setting:
https://github.com/nhibernate/nhibernate-core/blob/e99cf4378d7588c77fd13e487e8fa825675d23dd/src/NHibernate/Type/EntityType.cs#L501-L512

And later here, we take the `on` clause if it is not empty, instead of computing it with all filters if `ForceFilter`. It appears this `on` clause will always be empty in our case, arbitrary joins, because `GetOnCondition` current implementation always yields that excepted for property-ref associations.
https://github.com/nhibernate/nhibernate-core/blob/e99cf4378d7588c77fd13e487e8fa825675d23dd/src/NHibernate/Type/EntityType.cs#L496-L499

So, the current change is enough to fix the trouble. I cannot think of a related test case demonstrating a need to change `GetOnCondition`. This implementation looks a bit brittle to me, but a patch change is not the place to get things better from that viewpoint, unless we can demonstrate a failing case.

src/NHibernate/Engine/JoinSequence.cs
```diff
@@ -195,10 +195,17 @@ public JoinFragment ToJoinFragment(IDictionary<string, IFilter> enabledFilters,
...
 					// NH Different behavior : NH1179 and NH1293
 					// Apply filters for entity joins and Many-To-One association
-					var enabledForManyToOne = FilterHelper.GetEnabledForManyToOne(enabledFilters);
-					condition = new SqlString(string.IsNullOrEmpty(on) && (ForceFilter || enabledForManyToOne.Count > 0)
-					            	? join.Joinable.FilterFragment(join.Alias, enabledForManyToOne)
-					            	: on);
+					if (string.IsNullOrEmpty(on))
+					{
+						var enabledFiltersForJoin = ForceFilter ? enabledFilters : FilterHelper.GetEnabledForManyToOne(enabledFilters);
+						condition = new SqlString(ForceFilter || enabledFiltersForJoin.Count > 0
```
fredericDelaporte, line 201: I have considered `ForceFilter` was meant to apply all filters, not just many-to-one enabled filters. But maybe I am wrong and it is here just for discriminator filters and other non session filters related filtering. In such case this change may not be suitable.

`ForceFilter` is enabled for entity joins. I think entity joins are arbitrary joins and should not respect the many-to-one filter setting, which seems to me about entity association joins. In such case, this fix would be legit.

But if instead we consider the many-to-one filter setting should be taken into account for arbitrary joins, meaning the filters should be disabled on arbitrary joins if that setting is disabled, then we should instead just add a breaking change in the releases note for v5.3.0.

bahusoid (approved):

src/NHibernate.Test/NHSpecificTest/GH3046/Fixture.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System.Linq;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.GH3046
+{
+	// Duplicated from GH2549 with minor changes to trigger the trouble.
```
hazzik, line 6: Maybe better to collapse test into GH2549?

hazzik (approved):

src/NHibernate.Test/NHSpecificTest/GH3046/Fixture.cs
fredericDelaporte: Like 9f99166?

(Or maybe by defining the filter by code in `OnSetup` and changing its definition for #3046, but it was just easier to do the way I have done it.)

hazzik: I was thinking just adding a new test to the #2549 tests.

fredericDelaporte: It seemed to me simpler to parameterize them (0420362).

hazzik: Perfect

PR #3000, opened 2022-01-19
Add .net 6 and net48 targets

bahusoid:

PR #3050, opened 2022-04-27
Add .NET Standard 2.1 target

hazzik:


bahusoid:
Similarly to  #3000 will cause intermittent failure builds like https://github.com/nhibernate/nhibernate-core/pull/3000#issuecomment-1016601861

hazzik:
> Similarly to https://github.com/nhibernate/nhibernate-core/pull/3000 will cause intermittent (quite often) failure builds like https://github.com/nhibernate/nhibernate-core/pull/3000#issuecomment-1016601861

I've set maxcpucount to 1, so it should help. However, builds might become slower.

hazzik:
> However, builds might become slower.

Seems the same time.

src/NHibernate/NHibernate.csproj
```diff
@@ -48,18 +48,23 @@
...
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.1'">
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.7.0" />
-    <PackageReference Include="System.Reflection.Emit" Version="4.7.0" />
-    <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
   </ItemGroup>
-
+
```
fredericDelaporte, line 67: Undue white space adding an undue change.

```suggestion

```

src/NHibernate/Driver/DriverBase.cs
```diff
@@ -338,7 +338,7 @@ public DbParameter GenerateOutputParameter(DbCommand command)
 
 		public virtual bool RequiresTimeSpanForTime => false;
 
-#if NETCOREAPP2_0_OR_GREATER
+#if NETCOREAPP2_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
 		public virtual bool SupportsSystemTransactions => false;
 
 		public virtual bool SupportsNullEnlistment => false;
```
fredericDelaporte, line 344: Shouldn't we invert the condition?
```
#if NETFX
		public virtual bool SupportsSystemTransactions => true;

		public virtual bool SupportsNullEnlistment => true;
#else
		public virtual bool SupportsSystemTransactions => false;

		public virtual bool SupportsNullEnlistment => false;
#endif
```

src/NHibernate/NHibernate.csproj
hazzik: Thanks

src/NHibernate/Driver/DriverBase.cs
hazzik: Previously, for some reason, for .NET Standard 2.0 it was the same as for .NET Fx.

hazzik: @fredericDelaporte done

fredericDelaporte: That was coming from [here](https://github.com/nhibernate/nhibernate-core/pull/1523/files#r160946816). Theoretically, System.Transactions are supported in all our targets. But at the time, only NET Fx data providers were actually supporting them. (Even non-distributed ambient transaction were actually supported only by NET Fx data providers.)

As these properties are only used for tests, and do not change the behavior of the NHibernate library, setting them to false for the appcore target was the way to avoid testing these unsupported ambient transactions under that target. I do not think there was much thoughts about whether it should be set to true or false for netstandard.

fredericDelaporte (approved):

PR #3052, opened 2022-04-27
Support for BeginTransactionAsync

ramonsmits:
WIP, clone of sync methods. Cancellation tokens missing. Missing some #if to as BeginTransactionAsync only exists since NET Standard 2.1

### DIFF
--- a/src/NHibernate/Async/AdoNet/ConnectionManager.cs
+++ b/src/NHibernate/Async/AdoNet/ConnectionManager.cs
@@ -113,5 +113,19 @@ namespace NHibernate.AdoNet
 			EnlistInTransaction(result);
 			return result;
 		}
+
+		public async Task<ITransaction> BeginTransactionAsync(IsolationLevel isolationLevel)
+		{
+			EnsureTransactionIsCreated();
+			await _transaction.BeginAsync(isolationLevel).ConfigureAwait(false);
+			return _transaction;
+		}
+
+		public async Task<ITransaction> BeginTransactionAsync()
+		{
+			EnsureTransactionIsCreated();
+			await _transaction.BeginAsync().ConfigureAwait(false);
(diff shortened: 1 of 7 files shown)


src/NHibernate/Async/AdoNet/ConnectionManager.cs
```diff
@@ -113,5 +113,21 @@ public async Task<DbCommand> CreateCommandAsync(CancellationToken cancellationTo
...
+			return _transaction;
+		}
+
+		public async Task<ITransaction> BeginTransactionAsync()
+		{
+			EnsureTransactionIsCreated();
+			await _transaction.BeginAsync().ConfigureAwait(false);
+			return _transaction;
+		}
+#endif
```
fredericDelaporte, line 131: No manual editing of generated code files (see their header) can be allowed.

Almost all the "coding" for async happens in the AsyncGenerator yaml file. To diagnose why it does not generate a BeginTransactionAsync by itself, run the aync generator (see contributing.md) and analyze its output.

Either we have disabled it for this member on purpose (then look in git history from where it comes to find the rational) or it may not actually trigger a connection opening as you seem to assume. (I have not checked anything.) Maybe some other cause is in play, like an async candidate access triggered by a property access, case which cannot be converted to async. Such cases will de diagnosed in the async generator output.

hazzik: > why it does not generate a BeginTransactionAsync by itself

Two reasons:

1. The target generation is set to netstandard2.0
2. The return type is `ValueTask<>` that is yet not supported by AsyncGenerator.

ramonsmits: I didn't notice these files were autogenerated :-(

PR #3062, opened 2022-05-05
feat(restrictions): add IsEqual when using Restrictions.On(() => prop)

spindel87:


hazzik:
What SQL would be generated? Also, we need to add tests.

spindel87:
So our problem was that using combination of Restrictions.Or and Restrictions.And.
We want to create criteria with Restrictions.On(() => car) and this support .IsNotNull, IsLike and some other but not Equal 

This is made up but somewhat close to what was needed.                                                                      

```charp
.Where(Restrictions.Or(
                    Restrictions.And(
                        Restrictions.On(() => car.User).IsEqual(userIdentityId),
                        Restrictions.On(() => car.Status).IsEqual("Approved"),
                    ),
                    Restrictions.And(
                        Restrictions.On(() => car.Status).IsEqual("Pending"),
                        Restrictions.On(() => owner.User).IsEqual(userIdentityId)
                    )
                ))
```

               
Tests of course that was expected, was kind of lost in all the test couldn't find :)
But *RestrictionsFixture.cs* looks like one place for sure, somewhere else?

hazzik:
The API is confusing, but this is what you're looking for, I believe: 
```csharp
.Where(Restrictions.Or(
    Restrictions.And (
        Restrictions.Where(() => car.User == userIdentityId),
        Restrictions.Where(() => car.Status == "Approved"),
    ),
    Restrictions.And (
        Restrictions.Where(() => car.Status == "Pending"),
        Restrictions.Where(() => owner.User == userIdentityId)
    )
))
```

spindel87:
Alright didn't know that, nice thank you my good sir

PR #3074, opened 2022-06-16
Added support for the CancelQuery() method in IStatelessSession

### LINKED ISSUE #809: NH-2799 - Provide the CancelQuery() method in IStatelessSession
**cremor** created an issue — <time datetime="2011-07-20T06:05:18.165Z" title="2011-07-20T06:05:18.165Z">20th July 2011, 6:05:18</time>:
> Just as the title says, it would be good to have that method to cancel long running queries in stateless sessions.
>

SoftStoneDevelop:
Resolve #809


src/NHibernate/IStatelessSession.cs
```diff
@@ -143,6 +143,15 @@ public partial interface IStatelessSession : IDisposable
...
 		void Close();
 
+		/// <summary>
+		/// Cancel execution of the current query.
+		/// </summary>
+		/// <remarks>
+		/// May be called from one thread to stop execution of a query in another thread.
+		/// Use with care!
+		/// </remarks>
+		void CancelQuery();
```
fredericDelaporte, line 153: Adding a new member to an interface is a binary breaking change. These kind of changes have to be postponed until the next major version.

As a workaround, we use extension methods to avoid the breaking change, with a comment telling to do the change on the next major. See the extension class at the start of this file for examples.

fredericDelaporte (commented):
Thanks for your contribution,

We do not accept PR without corresponding tests. See [contributing](https://github.com/nhibernate/nhibernate-core/blob/master/CONTRIBUTING.md). May you add a test?

src/NHibernate/IStatelessSession.cs
SoftStoneDevelop: Okay, I moved it to an extension class.

SoftStoneDevelop: ISession don't have test  for ISession.CancelQuery(). Why IStatelessSession.CancelQuery() must have it?

Test hould be like this(in PostgreSQL), I suppose:
```C#

[Test]
public async Task CancelQuery()
{
    IStatelessSession ss = Sfi.OpenStatelessSession();
    var queryTask = Task.Factory.StartNew(() =>
    {
        var result = ss.CreateSQLQuery(@"
SELECT 
  clock_timestamp(),
  pg_sleep(10);
").UniqueResult();
    });

    Thread.Sleep(2000);
    ss.CancelQuery();
    try
    {
        await queryTask;
    }
    catch (Exception ex)
    {
        Assert.IsNotNull(ex.InnerException);
	Assert.AreEqual(ex.InnerException.GetType(), typeof(OperationCanceledException));
	var canceledException = (OperationCanceledException) ex.InnerException;
        Assert.IsNotNull(canceledException.InnerException);
	var postgresException = canceledException.InnerException as PostgresException;
	Assert.IsNotNull(postgresException);
	Assert.AreEqual(postgresException.SqlState, PostgresErrorCodes.QueryCanceled);
    }
}
```
We need to be sure of the request execution time in order to have time to cancel it(' pg_sleep(10);').
The main problem is that this query is related to PostgreSQL, just like the error code.

Do we need a test that checks for the DbCommand.Cancel() call? This is not the responsibility of Hibernate?
In AbstractBatcher.cs:
```C#
public void CancelLastQuery()
		{
			try
			{
				if (_lastQuery != null)
				{
					_lastQuery.Cancel();
				}
			}
			catch (HibernateException)
			{
				// Do not call Convert on HibernateExceptions
				throw;
			}
			catch (Exception sqle)
			{
				throw Convert(sqle, "Could not cancel query");
			}
		}
```

fredericDelaporte: Development practices change over time. Not everything has been developed with proper tests in the past, but that is not a valid reason for adding features or fixing bugs without corresponding tests.

About your proposal, we may accept a test specific to some databases, when devising one working for all is not practical, provided the test runs for at least one of the databases we have automated tests for. So a test specific to PostgreSql can be accepted. Use Assert.Ignore to skip other databases.

But maybe you can test it in a more database agnostic way, by using another session to exclusively lock some data, and then perform queries on the same data with exclusive locks too with another session on which `CancelQuery` will be called.

The main pitfall with your change is, does the session always use the batcher to run queries? If not, calling that function would do nothing, and that would be a bug. Ideally all ways of querying with a session should have a corresponding `CancelQuery` test.

SoftStoneDevelop: If it's a select query (not an insert/update/delete command), then yes, it always uses Batcher when calling "List" or "List<T>" (as far as I can see).
Added tests.

src/NHibernate.Test/Stateless/StatelessSessionCancelQueryFixture.cs
```diff
@@ -0,0 +1,110 @@
...
+		public void CancelLinqQuery()
+		{
+			CancelQueryTest(s => s.Query<Document>().WithLock(LockMode.Upgrade).ToList());
+		}
+
+		[Test]
+		public void CancelQueryOverQuery()
+		{
+			CancelQueryTest(s => s.QueryOver<Document>().Lock().Upgrade.List());
+		}
```
fredericDelaporte, line 108: Added test for QueryOver (covering the Criteria case by the way), removed SQL query test which cannot be "database agnostic".

src/NHibernate.Test/Stateless/StatelessSessionCancelQueryFixture.cs
```diff
@@ -0,0 +1,110 @@
...
+				using (var t2 = s2.BeginTransaction())
+				{
+					var queryTask = Task.Factory.StartNew(() => queryAction(s2));
+
+					Thread.Sleep(200);
+					s2.CancelQuery();
+					Assert.That(() => queryTask,
+						Throws.InnerException.TypeOf(typeof(OperationCanceledException))
+							.Or.InnerException.Message.Contains("cancelled")
+							.Or.InnerException.Message.Contains("canceled"));
```
fredericDelaporte, line 87: Newer tests should be written with `Assert.That` assertions.

src/NHibernate/IStatelessSession.cs
```diff
@@ -74,6 +74,22 @@ public static void FlushBatcher(this IStatelessSession session)
...
+		/// May be called from one thread to stop execution of a query in another thread.
+		/// Use with care!
+		/// </remarks>
+		public static void CancelQuery(this IStatelessSession session)
+		{
+			var implementation = session.GetSessionImplementation();
+			using (implementation.BeginProcess())
+			{
+				implementation.Batcher.CancelLastQuery();
+			}
```
fredericDelaporte, line 91: Granted, `GetSessionImplementation()` is just `=> this;`, but still, calling twice a function while expecting to always get the same thing should not be done.

fredericDelaporte (commented):
I have not checked all databases on my local setup, so waiting for the automated tests to do this for me.

src/NHibernate.Test/Stateless/StatelessSessionCancelQueryFixture.cs
```diff
@@ -0,0 +1,110 @@
...
+			using (var t1 = s1.BeginTransaction())
+			{
+				s1.Get<Document>(_documentName, LockMode.Upgrade);
+
+				using (var s2 = Sfi.OpenStatelessSession())
+				using (var t2 = s2.BeginTransaction())
+				{
+					var queryTask = Task.Factory.StartNew(() => queryAction(s2));
+
+					Thread.Sleep(200);
```
fredericDelaporte, line 82: Reduced from 2 seconds down to 200ms. We have more than 10 000 tests, even 200ms is already a big wait for us.

fredericDelaporte:
On GitHub action, AFAIK we are running SQL Server under Linux, and we seem to hit dotnet/SqlClient#109. It has been fixed in Microsoft.Data.SqlClient 2.0, but we are using System.Data.SqlClient, which is unlikely to be fixed, since Microsoft has announced it will go on developing only Microsoft.Data.SqlClient.

Should we switch Sql Server tests to Microsoft.Data.SqlClient? Well, ideally I think we should go on testing System.Data.SqlClient under .Net Framework, but test Microsoft.Data.SqlClient under net6.0.

src/NHibernate.Test/Stateless/StatelessSessionCancelQueryFixture.cs
SoftStoneDevelop: Thanks, I didn't know NHibernate had lock methods (good to know).

fredericDelaporte:
@hazzik, it seems this cancel test freezes Oracle test suit. (Maybe Oracle does not support Cancel...) The NHibernate agent on TeamCity seems gone. Can you check what is going on please?

SoftStoneDevelop:
> @hazzik, it seems this cancel test freezes Oracle test suit. (Maybe Oracle does not support Cancel...) The NHibernate agent on TeamCity seems gone. Can you check what is going on please?

Today I'll find out what's wrong.
So far, I think the problem is that oracle does not automatically release the lock when the transaction closes.

hazzik:
@fredericDelaporte 

> The NHibernate agent on TeamCity seems gone. Can you check what is going on please?

My Credit Card in Azure is expired and MS has shut down the environment. That is why.

hazzik:
> Should we switch Sql Server tests to Microsoft.Data.SqlClient? 

We should at least add it to the test suit. Maybe we test it in both platforms?

hazzik:
> My Credit Card in Azure is expired and MS has shut down the environment. That is why.

Ok, I've re-enabled my Azure subscription. The agent is back online, but now it is not authorized :(

bahusoid:
> The agent is back online, but now it is not authorized :(

It seems all you have to do is to click Authorize on this page  https://teamcity.jetbrains.com/agents.html?tab=unauthorizedAgents
( I see no such button but I might not have all required rights)

hazzik:
I don’t have the button either. Need to contact Jetbrains people.

hazzik (approved):

hazzik:
I've raised a ticket to solve it. I don't have permission to change pool nor authorize agents. Hopefully it will be resolved soon. Sorry for the delays - I don't have good access to a workstation.

hazzik:
Ok, agent is back. I've also increased tier for the VM, so builds are slightly faster now.

hazzik:
For some reason I cannot get any logs for Oracle

hazzik:
Oracle on TC works fine :-/

SoftStoneDevelop:
@hazzik 
I checked locally on Oracle 19c: test passed. Perhaps the problem is that on TC Oracle 18c. In release 18c, "Manual completion of unmanaged queries" was added (https://docs.oracle.com/en/database/oracle/oracle-database/18/newft/new-features.html#GUID-EFC832BE-DC2C-4997-8C38-8A9E27A4FFC4). But from what version 18.x?

I'll try installing 18.3 on my machine and run the test: if it passes then I have no idea why TC failed.

SoftStoneDevelop:
On 18.3(on my local server) work fine, all test passed. The only difference is I have Oracle on Windows.

SoftStoneDevelop:
> On 18.3 (on my local server) it works fine, all tests passed. The only difference from TC is that I have Oracle on Windows.

"Canceled in 360m - Oracle" misled me, I checked locally for nothing. At the TC: [all test(NHibernate Oracle Managed 32) pass](https://teamcity.jetbrains.com/project.html?projectId=NHibernate&tab=projectOverview&branch_NHibernate=pull%2F3074).

bahusoid:
> "Canceled in 360m - Oracle" misled me

Why misled? There is clearly some issue with Git Hub Action build (Oracle on Linux) - https://github.com/nhibernate/nhibernate-core/runs/7643016174?check_suite_focus=true

SoftStoneDevelop:
On "Git Hub Action" yes but not on teamcity(I initially thought that the assembly hung on the TC due to deadlock). So "misled me" means that the problem is not where I thought it was.

hazzik:
But I wrote that:

> Oracle on TC works fine :-/

SoftStoneDevelop:
Yes, yes I am blind, sometimes.

bahusoid:
I think we should simply skip tests for Oracle and Sql Server on linux.

fredericDelaporte (approved):

bahusoid (approved):
`SupportsCancelQuery` introduction looks not necessary to me (I expected skip right inside test case). But why not.

fredericDelaporte:
It makes for a central point for checking database limitations, which may help when we wish to check if some of these limitations do no more apply. Tests skipped directly inside them for some databases are harder to spot.

hazzik:
I think the problem is that the implementation relies on the batcher implementation. And SqlServer and Oracle use custom batcher implementation that might behave differently.

hazzik:
Created issue to track SqlServer and Oracle #3167

hazzik (approved):

PR #3077, opened 2022-07-07
SqlStringHelper.RemoveAsAliasesFromSql remove *all* aliases

Louis9902:
This removes multiple occurrences of aliases from the sql string not just the last one.

I had the problem while using the following projection, which results in invalid sql (sqlserver)
```csharp
Projections.Distinct(Projections.ProjectionList()
.Add(Projections.Max(() => ability.Role))
.Add(Projections.GroupProperty(Projections.RootEntity())))
```

### DIFF
--- a/src/NHibernate/Criterion/GroupedProjection.cs
+++ b/src/NHibernate/Criterion/GroupedProjection.cs
@@ -25,6 +25,9 @@ namespace NHibernate.Criterion
 		{
 			if (projection is IPropertyProjection propertyProjection)
 				return new SqlString(string.Join(", ", criteriaQuery.GetColumns(criteria, propertyProjection.PropertyName)));
+			// This is because if grouped by an entity, multiple columns are in an projection which all have as aliases, which are not allowed.
+			if (projection is EntityProjection)
+				return SqlStringHelper.RemoveAllAsAliasesFromSql(renderedProjection);
 
 			//This is kind of a hack. The hack is based on the fact that ToGroupSqlString always called after ToSqlString.
 			return SqlStringHelper.RemoveAsAliasesFromSql(renderedProjection);
--- a/src/NHibernate/SqlCommand/SqlString.cs
+++ b/src/NHibernate/SqlCommand/SqlString.cs
@@ -408,6 +408,7 @@ namespace NHibernate.SqlCommand
 		/// </summary>
 		/// <param name="text">Text to look for in the <see cref="SqlString" />. Must be in lower
 		/// case.</param>
+		/// <param name="startIndex">The zero-based index of the search starting position.</param>
 		/// <remarks>
 		/// The text must be located entirely in a string part of the <see cref="SqlString" />.
 		/// Searching for <c>"a ? b"</c> in an <see cref="SqlString" /> consisting of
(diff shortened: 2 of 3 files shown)


Louis9902:
@hazzik can you or somebody else maybe approve the ci run, so I can check if everything works?

Louis9902:
Okay, the MySQL failure seems to be unrelated, because it somehow breaks in the authentication step.

src/NHibernate/Criterion/GroupedProjection.cs
```diff
@@ -25,6 +25,9 @@ public virtual SqlString ToGroupSqlString(ICriteria criteria, ICriteriaQuery cri
 		{
 			if (projection is IPropertyProjection propertyProjection)
 				return new SqlString(string.Join(", ", criteriaQuery.GetColumns(criteria, propertyProjection.PropertyName)));
+			// This is because if grouped by an entity, multiple columns are in an projection which all have as aliases, which are not allowed.
+			if (projection is EntityProjection)
+				return SqlStringHelper.RemoveAllAsAliasesFromSql(renderedProjection);
```
bahusoid, line 30: With such specific `EntityProjection` check I would implement `EntityProjection.GetColumns` and call it instead (similar to how it's done for `IPropertyProjection` above)

Louis9902: > With such specific `EntityProjection` check I would implement `EntityProjection.GetColumns` and call it instead (similar to how it's done for `IPropertyProjection` above)

Do you mean with 'I' me or would you implement it? Aside from that, could you lead me in the direction where I can find similar code.

hazzik: > Do you mean with 'I' me or would you implement it? 

He meant 'If I were you' or 'it would be better to'. So it implied that it is better to make the change.

src/NHibernate/SqlCommand/SqlString.cs
```diff
@@ -408,16 +408,17 @@ public int GetParameterCount()
...
+		/// <param name="startIndex">The zero-based index of the search starting position.</param>
 		/// <remarks>
 		/// The text must be located entirely in a string part of the <see cref="SqlString" />.
 		/// Searching for <c>"a ? b"</c> in an <see cref="SqlString" /> consisting of
 		/// <c>"a ", Parameter, " b"</c> will result in no matches.
 		/// </remarks>
 		/// <returns>The index of the first occurrence of <paramref name="text" />, or -1
 		/// if not found.</returns>
-		public int IndexOfCaseInsensitive(string text)
+		public int IndexOfCaseInsensitive(string text, int startIndex = 0)
```
hazzik, line 419: Need to add a new method instead

hazzik:
We need some tests, otherwise we would not be able to accept the contribution.

src/NHibernate/Criterion/GroupedProjection.cs
```diff
@@ -25,6 +25,9 @@ public virtual SqlString ToGroupSqlString(ICriteria criteria, ICriteriaQuery cri
 		{
 			if (projection is IPropertyProjection propertyProjection)
 				return new SqlString(string.Join(", ", criteriaQuery.GetColumns(criteria, propertyProjection.PropertyName)));
+			// This is because if grouped by an entity, multiple columns are in an projection which all have as aliases, which are not allowed.
+			if (projection is EntityProjection)
```
hazzik, line 29: You should just override method in `EntityProjection`.

PR #3120, opened 2022-08-21
Guards against use of a disposed session factory

fredericDelaporte:
Help diagnose #3026 reports.

I suspect those reports are due to user code errors, using a disposed session factory. It appears a disposed session factory does not throw when used. It lets things fail quite later with harder to understand exceptions. This leads to bug reports for what is likely user bugs.

This PR currently targets 5.3.x. But this may be considered as too much of a possible breaking change for this branch, even if using a disposed session factory is a bug on user side.
In my opinion it is ok to have it in 5.3.x.

### DIFF
--- a/src/NHibernate/Impl/SessionFactoryImpl.cs
+++ b/src/NHibernate/Impl/SessionFactoryImpl.cs
@@ -469,7 +469,11 @@ namespace NHibernate.Impl
 
 		public EventListeners EventListeners
 		{
-			get { return eventListeners; }
+			get
+			{
+				CheckNotClosed();
+				return eventListeners;
+			}
 		}
 
 		#region IObjectReference Members
@@ -513,6 +517,7 @@ namespace NHibernate.Impl
 
 		public ISessionBuilder WithOptions()
 		{
+			CheckNotClosed();
 			return new SessionBuilderImpl(this);
 		}
 
@@ -563,6 +568,7 @@ namespace NHibernate.Impl
 
 		public IStatelessSessionBuilder WithStatelessOptions()
 		{
+			CheckNotClosed();
 			return new StatelessSessionBuilderImpl(this);
 		}
 
@@ -580,6 +586,7 @@ namespace NHibernate.Impl
 
 		public IEntityPersister GetEntityPersister(string entityName)
 		{
+			CheckNotClosed();
 			IEntityPersister value;
 			if (entityPersisters.TryGetValue(entityName, out value) == false)
 				throw new MappingException("No persister for: " + entityName);
@@ -588,6 +595,7 @@ namespace NHibernate.Impl
 
 		public IEntityPersister TryGetEntityPersister(string entityName)
 		{
+			CheckNotClosed();
 			IEntityPersister result;
 			entityPersisters.TryGetValue(entityName, out result);
 			return result;
@@ -595,6 +603,7 @@ namespace NHibernate.Impl
 
 		public ICollectionPersister GetCollectionPersister(string role)
 		{
+			CheckNotClosed();
 			ICollectionPersister value;
 			if (collectionPersisters.TryGetValue(role, out value) == false)
 				throw new MappingException("Unknown collection role: " + role);
@@ -863,6 +872,14 @@ namespace NHibernate.Impl
 			Close();
 		}
 
+		private void CheckNotClosed()
+		{
+			if (isClosed)
+			{
+				throw new ObjectDisposedException($"Session factory {Name} with id {Uuid}");
+			}
+		}
+
 		/// <summary>
 		/// Closes the session factory, releasing all held resources.
 		/// <list>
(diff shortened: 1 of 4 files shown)


src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -469,7 +469,11 @@ private IQueryCache BuildQueryCache(string queryCacheName)
 
 		public EventListeners EventListeners
 		{
-			get { return eventListeners; }
+			get
+			{
+				CheckNotDisposed();
+				return eventListeners;
```
fredericDelaporte, line 475: The session factory closing `Destroy` the event listeners, so better guard against any later use of them.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -513,6 +517,7 @@ public object GetRealObject(StreamingContext context)
 
 		public ISessionBuilder WithOptions()
 		{
+			CheckNotDisposed();
 			return new SessionBuilderImpl(this);
```
fredericDelaporte, line 521: I have mostly protected accesses to resources destroyed or disposed by the session factory closing, except here and in stateless equivalent, where there is no direct accesses to these resources. But that is the main usage of session factories, so, better to directly throw here when using it on a disposed factory.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -580,6 +586,7 @@ public IStatelessSession OpenStatelessSession(DbConnection connection)
 
 		public IEntityPersister GetEntityPersister(string entityName)
 		{
+			CheckNotDisposed();
 			IEntityPersister value;
 			if (entityPersisters.TryGetValue(entityName, out value) == false)
```
fredericDelaporte, line 591: `entityPersisters` have their cache property destroyed by the factory closing, so guarding against their use. This also causes most session usages to throw if their session factory is disposed, which is also the aim of this PR.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -588,13 +595,15 @@ public IEntityPersister GetEntityPersister(string entityName)
...
 			IEntityPersister result;
 			entityPersisters.TryGetValue(entityName, out result);
 			return result;
 		}
 
 		public ICollectionPersister GetCollectionPersister(string role)
 		{
+			CheckNotDisposed();
 			ICollectionPersister value;
 			if (collectionPersisters.TryGetValue(role, out value) == false)
```
fredericDelaporte, line 608: Same here, their cache is destroyed by the factory closing.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1105,6 +1122,7 @@ public UpdateTimestampsCache UpdateTimestampsCache
 		public IDictionary<string, ICache> GetAllSecondLevelCacheRegions()
 #pragma warning restore 618
 		{
+			CheckNotDisposed();
 			return
 				_allCacheRegions
```
fredericDelaporte, line 1127: Caches are destroyed by the closing.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1119,6 +1137,7 @@ public UpdateTimestampsCache UpdateTimestampsCache
 		public ICache GetSecondLevelCacheRegion(string regionName)
 #pragma warning restore 618
 		{
+			CheckNotDisposed();
 			_allCacheRegions.TryGetValue(regionName, out var result);
 			return result;
```
fredericDelaporte, line 1142: `_allCacheRegions` is also accessed by `private GetCache`, which I do not guard because it is currently called only by already directly guarded methods or upon initialization of the session factory.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -1145,7 +1164,12 @@ public IStatisticsImplementor StatisticsImplementor
 
 		public IQueryCache QueryCache
 		{
-			get { return queryCache; }
+			get
+			{
+				if (queryCache != null)
+					CheckNotDisposed();
```
fredericDelaporte, line 1170: Not sure I should have that if. Done for consistency with `EvictQueries`.

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -852,6 +861,14 @@ public string GetImportedClassName(string className)
 			return collectionMetadata;
 		}
 
+		private void CheckNotDisposed()
+		{
+			if (disposed)
+			{
+				throw new ObjectDisposedException($"Session factory {Name} with id {Uuid}");
+			}
+		}
```
fredericDelaporte, line 870: I am going to changed this to `CheckNotClosed`, because we can close a factory without disposing it, and closing it is enough to cause the reported troubles.

hazzik (approved):

fredericDelaporte:
I had to force the MyGet publishing on TeamCity. Maybe I should have just waited, but I was not seeing any reason for it to have not run after the TeamCity "Release" build was finished.

PR #3126, opened 2022-08-22
Fix MappingByCode discriminator string type throws

### LINKED ISSUE #2717: MappingByCode discriminator column with string type throws exception
I am using `ConventionModelMapper` to map my entities, I am using a table per hierarchy strategy to map two entities with a shared parent. When I try to map the discriminator column to type string, it fails:

```C#
mapper.Class<NotificationChannel>(
    m =>
    {
        m.Table("notificationchannels");
        m.Discriminator(
            x =>
            {
                x.Column("discriminator");
                x.Type(NHibernateUtil.String);
            });
    });
```

Error: No parameterless constructor defined for type 'NHibernate.Type.StringType'.

I tried a lot of ways to make this work but nothing worked. The only way out for me is to map it as a numeric value as it seems the default is a numeric type.

bahusoid:
Fixes #2717


src/NHibernate/Mapping/ByCode/Impl/DiscriminatorMapper.cs
```diff
@@ -57,15 +57,15 @@ public void Column(Action<IColumnMapper> columnMapper)
...
+			Type(discriminatorType);
 		}
 
 		public void Type(IDiscriminatorType persistentType)
 		{
-			Type(persistentType.GetType());
+			discriminatorMapping.type = persistentType.Name;
 		}
 
 		public void Type<TPersistentType>() where TPersistentType : IDiscriminatorType
```
fredericDelaporte, line 71: This method and the next one will still breaks, I think.

Trying to map using `rc.Discriminator(d => d.Type<StringType>())` or `rc.Discriminator(d => d.Type(typeof(StringType))` should still break in the same way.

The root cause of the trouble seems to me:
 - Either having the `StringType` parameter-less [constructor](https://github.com/nhibernate/nhibernate-core/blob/9f6a965beb1c04b63e1fb725631ab016dce8ec06/src/NHibernate/Type/StringType.cs#L13) `internal` instead of `public`.
 - Or having the [objectActivator](https://github.com/nhibernate/nhibernate-core/blob/9f6a965beb1c04b63e1fb725631ab016dce8ec06/src/NHibernate/Bytecode/ActivatorObjectsFactory.cs#L9) not using the overload accepting parameter-less non-public constructor.
 - Or not using a specific instantiator for this case, which would support internal constructors.

Looking at IDiscriminatorType, it seems to me only string and character types have that internal restriction on their constructor, along with the `TrueFalseType` (although that is basically the same thing than the `YesNoType`, just using different chars). (I am not counting types which cannot be directly used but must be derived from, like non generic enum types and the CustomType.)

The internal restriction on the StringType parameter-less constructor is here since long: 2004-02-10. That change seems to have been done on variable length types. Likely the idea is to promote the use of the TypeFactory and reuse of instance types instead of directly instantiating them.
I am unsure we should lift that restriction on all variable length types.

"Fixing" the objectActivator would just fix the default one, not changing the trouble for people using their own activator, if they do not allow non-public constructors. And that would relax the constraint about public parameter-less constructor to almost all NHibernate dependencies, which may be a trouble for custom implementations, if they want to support what our default object factory does support.

Ceasing to use the ObjectsFactory for types could be an issue for some users if they use a custom factory to instantiate their custom types.

fredericDelaporte:
So the missing part is done with #3127. Do we need two separated PR?

bahusoid:
> So the missing part is done with https://github.com/nhibernate/nhibernate-core/pull/3127. Do we need two separated PR?

Yeap. Both changes makes sense to me. This one unifies old behavior and local optimization and #3127 is a global fix.

fredericDelaporte (dismissed):

hazzik (approved):

PR #3127, opened 2022-08-22
Register IType CLR types as aliases

bahusoid:
To allow directly use it in mapping by code `m.Type<StringType>()` instead of awkward `m.Type(NHibernateUtil.String)`


src/NHibernate/Type/TypeFactory.cs
```diff
@@ -140,6 +140,7 @@ public static void RegisterType(System.Type systemType, IType nhibernateType, IE
 		{
 			var typeAliases = new List<string>(aliases);
 			typeAliases.AddRange(GetClrTypeAliases(systemType));
+			typeAliases.AddRange(GetClrTypeAliases(nhibernateType.GetType()));
```
fredericDelaporte, line 143: That is another way to fully fix #2717.

fredericDelaporte (approved):

hazzik:
> To allow directly use it in mapping by code `m.Type<StringType>()` instead of awkward `m.Type(NHibernateUtil.String)`

I think this was made intentionally because `StringType` normally requires a length specified.

bahusoid:
> I think this was made intentionally because StringType normally requires a length specified.

I see no such intentions. Length is not required and `NHibernateUtil.String` returns instance of `new StringType()`. So how it makes things any different? The only advantage I see - usage of the same `StringType` instance across app.

But with this change there is no calls to  default `StringType` ctor on each `m.Type<StringType>()` . Cached type is returned:
https://github.com/nhibernate/nhibernate-core/blob/47992e73ed6114093f31ed99829dd8c4c7663829/src/NHibernate/Type/TypeFactory.cs#L547

src/NHibernate/Type/TypeFactory.cs
hazzik, line 143: Not all types are registered through this path. `Currency` type fails.

hazzik: Also, `TrueFalseType` is using different path.

src/NHibernate.Test/TypesTest/TimestampTypeFixture.cs
```diff
@@ -24,7 +24,7 @@ public void ObsoleteMessage()
 					var log = spy.GetWholeLog();
 					Assert.That(
 						log,
-						Does.Contain($"NHibernate.Type.TimestampType ({NHibernateUtil.Timestamp.Name}) is obsolete. Please use DateTimeType instead.").IgnoreCase);
+						Does.Contain($"NHibernate.Type.TimestampType is obsolete. Please use DateTimeType instead.").IgnoreCase);
```
bahusoid, line 27: Source and message of log is changed. Before:
https://github.com/nhibernate/nhibernate-core/blob/f027dad690566f7fbf7d4938a897b708f57f3dfd/src/NHibernate/Type/TypeFactory.cs#L690
Now:
https://github.com/nhibernate/nhibernate-core/blob/f027dad690566f7fbf7d4938a897b708f57f3dfd/src/NHibernate/Type/TypeFactory.cs#L549-L550

I don't think we should really care about keeping old message - so I just adjusted test

hazzik (approved):
nit: not completely happy with over-DRYed code.

Issue #3113, opened 2022-08-15
Join fails on Oracle9Dialect


khoshrou:
Hi 
I have just upgraded to 5.3.12 (it works in 5.2) and have a simple query running on an oracle database using Oracle9Dialect . Seems its creating wrong SQL.
Also for left outer joins I see old fashiond oracle SQL like:
```SQL
A = B (+) 
```
which is not ANSI standard like: 
```SQL
Left Outer Join A on A.x = B.y
```

Here is my query: 

``` CSharp
var sampleQuery = (from globalUser in Session.Query<GlobalUser>()

                                           join userRole in Session.Query<UserRole>() on globalUser.ID
                                               equals
                                               userRole.GlobalUser.ID

                                           select new
                                           {
                                               UserName = globalUser.UserName,
                                               userRole.Role.Name
                                           }
                            );

var list = sampleQuery.Select(x=>x.UserName).ToList();
```

Exception I get:

NHibernate.Exceptions.GenericADOException: could not execute query
[ select globaluser0_.USERNAME as col_0_0_ from ACUSER globaluser0_ ACUSERROLE userrole1_ where (userrole1_.ACUSERID=globaluser0_.ID) ]

``` SQL 
select globaluser0_.USERNAME as col_0_0_ from ACUSER globaluser0_ ACUSERROLE userrole1_ 
where (userrole1_.ACUSERID=globaluser0_.ID)
```
 ---> System.Data.OracleClient.OracleException: ORA-00933: SQL command not properly ended

clearly its a wrong sql generation. 
Do you have any idea what's happening here?

bahusoid:
>I have just upgraded to 5.3.12 
Which version did you use before upgrade?

> Also for left outer joins I see old fashioned oracle SQL
What dialect are you using? Try with `Oracle10gDialect`

khoshrou:
I was upgraded from 5.2 and using Oracle9Dialect

bahusoid:
Indeed simple join in LINQ fails in 5.3 for legacy dialects that don't support ANSI joins. Test case: https://github.com/bahusoid/nhibernate-core/tree/gh3113Master

It should work OK with `Oracle10gDialect`

khoshrou:
Thanks it works with Oracle10gDialect

bahusoid:
No need to close it. We might fix it for old dialects

khoshrou:
Sure and Thanks again

fredericDelaporte:
[...]

PR #2025, opened 2019-02-27
Port HQL Entity Join

bahusoid:
It's a port for the following hibernate [commit](https://github.com/hibernate/hibernate-orm/commit/f4b5f2c649b069b49affd194d1ded640e0561c88) adding support for explicit joins on unrelated classes

Fixes #981

PR #3137, opened 2022-08-28
Fix gh3113

fredericDelaporte:
Minimal fix proposal for #3113, which is a 5.3.0 regression introduced by #2327 along with #2025.

This trouble does by the way demonstrate that we lack an automated tests run for dialects using only theta style joins. Hundreds of tests fail for such a dialect, and maybe there is a bunch of them which are regressions.


hazzik (approved):

fredericDelaporte:
Before merging, I have to compare a test run on Oracle8i dialect without the change to one with the change, in order to make sure it does not break anything new.

bahusoid (approved):

fredericDelaporte:
Here are the test results. Down from 563 failures to 511, no new failure seen. (But checked manually.)
[Test-Results.zip](https://github.com/nhibernate/nhibernate-core/files/9448403/Test-Results.zip)

A test run on 5.2.x have "only" an hundred failures or so, with maybe 2 000 less tests. I think there is quite a bunch of other regressions for Oracle8i in 5.3.x, but it is a bit tedious to detect manually which tests are newly failing.

hazzik:
I think "we do not care" and should deprecate all older Oracle dialects.

fredericDelaporte:
I should have rewritten the commit message a bit better, sorry for having overlooked this.

PR #3139, opened 2022-08-29
WIP Add support for BeginTransactionAsync

hazzik:


src/NHibernate/AdoNet/AsyncExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Data;
+using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NHibernate
+{
+	internal static class AsyncExtensions
+	{
+		public static Task<DbTransaction> BeginTransactionAsync(this DbConnection connection, CancellationToken cancellationToken = default)
```
hazzik, line 12: @maca88 do you know why async generator does not pick up these extensions for `DbConnection`?

maca88: Async extension methods need to be registered by using `asyncExtensionMethods` option in `AsyncGenrator.yaml`. [Here](https://github.com/maca88/nhibernate-core/commit/61ed5ddab25c3f9fdbdb1b2a4c7455f3a93da612#diff-60e3e40b5223a1c12eeb3d32a9be6ceb498de4f79d2d62008da4a6330e30bfdfR247) you can see what changes are needed in the configuration to generate `BeginTransactionAsync` calls.

hazzik: Ok, thanks, I forgot about that.

src/AsyncGenerator.yml
```diff
@@ -139,6 +142,10 @@
     - conversion: Ignore
       name: DependentContext
       containingTypeName: AdoNetWithSystemTransactionFactory
+    asyncExtensionMethods:
+      projectFiles:
+      - fileName: AsyncExtensions.cs
```
hazzik, line 147: Am I doing something wrong, @maca88?

maca88: Upgrade to AsyncGenerator version `0.21.0` as it include some fixes regarding extension methods. As you will find out, the generated code does not compile for `.netstandard2.1` as `BeginTransactionAsync` return `ValueTask` and not `Task`. Therefore I suggest to first merge #3000 and then update the `targetFramework` of project `NHibernate` from `netcoreapp2.0` to `net6.0`.

src/AsyncGenerator.yml
```diff
@@ -212,6 +219,8 @@
       projectFiles:
       - fileName: LinqExtensionMethods.cs
         projectName: NHibernate
+      - fileName: AsyncExtensions.cs
```
maca88, line 222: This will not work even with version `0.21.0` as `AsyncExtensions` class does not exist for `net6.0` which the `NHibernate.Test` project targets. You need to add extension methods for both `ISession` and `IStatelessSession` in case it is not desired to have a breaking change and then registered them like this:
```
    asyncExtensionMethods:
      projectFiles:
      - fileName: LinqExtensionMethods.cs
        projectName: NHibernate
      assemblyTypes:
      - fullTypeName: NHibernate.SessionExtensions
        assemblyName: NHibernate
      - fullTypeName: NHibernate.StatelessSessionExtensions
        assemblyName: NHibernate
```
With version AsyncGenerator 0.21.0 onwards, it is recommended that `assemblyTypes` option is used instead of `projectFiles` when referencing types from a different project, otherwise it is possible that the extension methods won't be used.

hazzik: > This will not work even with version 0.21.0 as AsyncExtensions class does not exist for net6.0

Yeah, this was my mistake, I've put it into incorrect place at first, I just did not care to clear it.

src/NHibernate/Async/Driver/DriverBase.cs
```diff
@@ -0,0 +1,61 @@
...
+		{
+			if (cancellationToken.IsCancellationRequested)
+			{
+				return Task.FromCanceled<DbTransaction>(cancellationToken);
+			}
+			try
+			{
+				if (isolationLevel == IsolationLevel.Unspecified)
+				{
+					return connection.BeginTransactionAsync(cancellationToken);
```
hazzik, line 46: @maca88 I tried to fix this by using 'alwaysAwait' option, but I could not quite figure out the syntax.

src/AsyncGenerator.yml
```diff
@@ -1,13 +1,13 @@
-﻿projects:
+projects:
 - filePath: NHibernate/NHibernate.csproj
-  targetFramework: netcoreapp2.0
+  targetFramework: netcoreapp2.1
```
hazzik, line 3: This failed with following error:

```
crit: AsyncGenerator.CommandLine[0]
      Unable to find System.Threading.Tasks.Task type
      System.InvalidOperationException: Unable to find System.Threading.Tasks.Task type
         at AsyncGenerator.Internal.ThreadSleepAsyncCounterpartFinder.Initialize(Project project, IProjectConfiguration configuration, Compilation compilation) in C:\Workspace\Git\AsyncGenerator\Source\AsyncGenerator\Internal\ThreadSleepAsyncCounterpartFinder.cs:line 30
         at AsyncGenerator.AsyncCodeGenerator.GenerateProject(ProjectData projectData, ILoggerFactory loggerFactory, ILogger logger, CancellationToken cancellationToken)
         at AsyncGenerator.AsyncCodeGenerator.GenerateAsync(AsyncCodeConfiguration configuration, CancellationToken cancellationToken) in C:\Workspace\Git\AsyncGenerator\Source\AsyncGenerator\AsyncCodeGenerator.cs:line 84
         at AsyncGenerator.CommandLine.Program.Main(String[] args) in C:\Workspace\Git\AsyncGenerator\Source\AsyncGenerator.CommandLine\Program.cs:line 51
```

src/AsyncGenerator.yml
hazzik: I'll try to make it working on .net core 2.0 (bugging you by the way, so sorry, but please bear with me)

src/NHibernate/Async/Driver/DriverBase.cs
maca88: To fix this change the `alwaysAwait` section to:
```
alwaysAwait:
    - name: BeginTransaction
``` 
The filter `BeginTransactionAsync` could also work but the `containingTypeName` filter would need to be changed from `AsyncExtensions` to `DriverBase`.

src/AsyncGenerator.yml
maca88: The `targetFramework` can be set only to targets that the project supports, for `NHibernate` project it can be set to one of the following: `net461, netcoreapp2.0, netstandard2.0, netstandard2.1`

hazzik: Yeah, I'm stupid...

src/NHibernate/Async/Driver/DriverBase.cs
hazzik: How does it work? I thought it identifies the method that should be always awaited.

maca88: With `alwaysAwait` you define filters that are applied on every genereated async method to check whether to force await invocations inside the method body. So if we take the above filter as an example, every generated async method that has a sync counterpart named "BeginTransaction" will be generated by using `async/await` even if some of them could be genereated without it.

PR #3144, opened 2022-08-30
Fix named parameter leaking to query cache

### LINKED ISSUE #3030: Memory leak named parameter holds entity references
Investigating a memory leak in an app, it was evident that the NHibernate.Engine.Query.QueryPlanCache was holding on to references to entities via instances of NHibernate.Param.NamedParameter.  This only happens if the linq query expression embeds the entity itself as a parameter.

Below is a complete example which demonstrates this behaviour, note that only the first time a unique query is used is the entity pinned, see ` var first_reference_is_not_null = firstReference.Target;`  :

```csharp
using FluentNHibernate.Cfg;
using FluentNHibernate.Cfg.Db;
using FluentNHibernate.Mapping;
using NHibernate;
using NHibernate.Dialect;
using NHibernate.Tool.hbm2ddl;

public class programm
{
    public static void Main()
    {
        WeakReference sessionReference = null;
        WeakReference firstReference = null;
        WeakReference secondReference = null;

        new Action(() =>
        {
            var session = ConfigureSessionFactory();

            var first = new Test() { Id = 1 };
            var second = new Test() { Id = 2 };

            var f = session.Query<Test>().FirstOrDefault(f => f == first);
            var s = session.Query<Test>().FirstOrDefault(f => f == second);


            sessionReference = new WeakReference(session, true);
            firstReference = new WeakReference(first, true);
            secondReference = new WeakReference(second, true);

            session.Dispose();
        })();

        GC.Collect();
        GC.WaitForPendingFinalizers();

        var session_reference_is_null = sessionReference.Target;
        var first_reference_is_not_null = firstReference.Target;
        var second_reference_is_null = secondReference.Target;
    }

    private static ISession ConfigureSessionFactory()
    {
        var configuration = Fluently.Configure().Database(SQLiteConfiguration.Standard.InMemory).Mappings(m =>
        {
            m.FluentMappings.AddFromAssemblyOf<Test>();
        });

        configuration.ExposeConfiguration(c => SchemaMetadataUpdater.QuoteTableAndColumns(c, new MsSql2012Dialect()));

        var config = configuration.BuildConfiguration();

        var factory = config.BuildSessionFactory();
        var session = factory.OpenSession();
        var wr = new StringWriter();

        new SchemaExport(config).Execute(true, true, false, session.Connection, wr);
        var t = wr.ToString();
        return session;
    }

    public class TestMap : ClassMap<Test>
    {
        public TestMap()
        {
            Id(x => x.Id).GeneratedBy.Assigned();
        }
    }

    public class Test
    {
    }
}

````

### PR DESCRIPTION
**Pull request by hazzik**
Fixes #3030

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/GH3030/ByCodeFixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH3030/ByCodeFixture.cs
@@ -1,0 +1,72 @@
+﻿using System;
+using System.Linq;
+using NHibernate.Cfg.MappingSchema;
+using NHibernate.Mapping.ByCode;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.GH3030
+{
+	[TestFixture]
+	public class ByCodeFixture : TestCaseMappingByCode
+	{
+		protected override HbmMapping GetMappings()
+		{
+			var mapper = new ModelMapper();
+			mapper.Class<Entity>(
+				rc =>
+				{
+					rc.Table("Entity");
+					rc.Id(x => x.Id, m => m.Generator(Generators.Assigned));
+				});
+
+			return mapper.CompileMappingForAllExplicitlyAddedEntities();
+		}
+
+		protected override void OnTearDown()
+		{
+			using (var session = OpenSession())
+			using (var transaction = session.BeginTransaction())
+			{
+				session.CreateQuery("delete from System.Object").ExecuteUpdate();
+				transaction.Commit();
+			}
+		}
+
+		[Test]
+		public void LinqShouldNotLeakEntityParameters()
+		{
+			WeakReference sessionReference = null;
+			WeakReference firstReference = null;
+			WeakReference secondReference = null;
+
+			new System.Action(
+				() =>
+				{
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by fredericDelaporte — APPROVED** [review]

File: src/NHibernate.Test/NHSpecificTest/GH3030/ByCodeFixture.cs [hunk 1]
```diff
@@ -0,0 +1,72 @@
...
+			new System.Action(
+				() =>
+				{
+					using (var session = ((DebugSessionFactory) Sfi).ActualFactory.OpenSession())
+					{
+						var first = new Entity { Id = 1 };
+						var second = new Entity { Id = 2 };
+
+						_ = session.Query<Entity>().FirstOrDefault(f => f == first);
+						_ = session.Query<Entity>().FirstOrDefault(f => f == second);
```
bahusoid, line 51: MIssing async regen?

hazzik: No, it is not async-able because of action. First I tried with local function, but test was failing, as I thought because local function closure works differently, so I decided to make test exactly as it was submitted.

PR #3154, opened 2022-09-08
Fix query dynamic component on unmapped interface

### LINKED ISSUE #3150: LINQ query dynamic component by interface hangs the application
I have the following query which executes fine:

```
var posts = await session.Query<BlogPost>()
    .Where(p => p.Attributes["Field1"] != null)
   .ToListAsync();
```

However if I change it to use an interface instead then the query hangs.

```
var posts = await session.Query<IBlogPost>()
    .Where(p => p.Attributes["Field1"] != null)
    .ToListAsync();
```

Please see the attached application which helps demonstrate this. This problem only happens when filtering against an attribute.

[ConsoleApp1.zip](https://github.com/nhibernate/nhibernate-core/files/9483340/ConsoleApp1.zip)

bahusoid:
Fixes #3150


src/NHibernate/Linq/Visitors/ExpressionParameterVisitor.cs
```diff
@@ -110,7 +110,7 @@ protected override Expression VisitMethodCall(MethodCallExpression expression)
 				_collectionParameters.Add(collectionParameter);
 			}
 
-			if (VisitorUtil.IsDynamicComponentDictionaryGetter(expression, _sessionFactory))
+			if (VisitorUtil.TryGetPotentialDynamicComponentDictionaryMember(expression, out _))
```
bahusoid, line 113: It's too early to fully resolve expression so just guess here...

hazzik (approved):

Issue #3156, opened 2022-09-12
Evaluation failure when using `Nullable` without a value in LINQ


krzkowal:
Using `Nullable<T>.HasValue` in LINQ  throws `NHibernate.HibernateException: Evaluation failure on null.HasValue` if no value is set.

<details open>
  <summary>Test case</summary>

```
[Test]
public void CanUseNullableWithoutAValue()
{
    bool? isActive = null;

    Assert.DoesNotThrow(
        () => _ = session.Query<Role>()
                     .Where(r => !isActive.HasValue || r.IsActive == isActive.Value)
                     .ToList());
}
```
</details>

<details open>
  <summary>Error details</summary>

```
NHibernate.HibernateException: Evaluation failure on null.HasValue ---> System.Reflection.TargetException: Non-static method requires a target.
   at System.Reflection.RuntimeMethodInfo.CheckConsistency(Object target)
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
   at NHibernate.Impl.ExpressionProcessor.FindValue(Expression expression) in C:\src\nhibernate-core\src\NHibernate\Impl\ExpressionProcessor.cs:line 269
   at NHibernate.Linq.Visitors.NhPartialEvaluatingExpressionVisitor.EvaluateSubtree(Expression subtree) in C:\src\nhibernate-core\src\NHibernate\Linq\Visitors\NhPartialEvaluatingExpressionVisitor.cs:line 151
   at NHibernate.Linq.Visitors.NhPartialEvaluatingExpressionVisitor.Visit(Expression expression) in C:\src\nhibernate-core\src\NHibernate\Linq\Visitors\NhPartialEvaluatingExpressionVisitor.cs:line 92
   --- End of inner exception stack trace ---
   at NHibernate.Linq.Visitors.NhPartialEvaluatingExpressionVisitor.Visit(Expression expression) in C:\src\nhibernate-core\src\NHibernate\Linq\Visitors\NhPartialEvaluatingExpressionVisitor.cs:line 99
[...]

PR #3158, opened 2022-09-12
LINQ Nullable in WHERE condition tests

krzkowal:
Test cases for https://github.com/nhibernate/nhibernate-core/issues/3156

### DIFF
--- a/src/NHibernate.Test/Linq/WhereTests.cs
+++ b/src/NHibernate.Test/Linq/WhereTests.cs
@@ -810,6 +810,28 @@ namespace NHibernate.Test.Linq
 			Assert.That(query.Count, Is.EqualTo(3));
 		}
 
+		[Test(Description = "GH-3156")]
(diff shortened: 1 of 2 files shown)


src/NHibernate.Test/Linq/WhereTests.cs
```diff
@@ -738,6 +738,28 @@ public void BitwiseQuery3()
...
 		}
 
+		[Test(Description = "GH-3156")]
+		public void CanUseNullableWithoutAValue()
+		{
+			bool? isActive = null;
+
+			Assert.DoesNotThrow(
+				() => _ = session.Query<Role>()
+				             .Where(r => !isActive.HasValue || r.IsActive == isActive.Value)
```
fredericDelaporte, line 748: Accessing `.Value` on a null parameter remains unsupported, but that is not a regression. And it is a minimal priority issue in my opinion, since this `.Value` access is not needed and has not any advantage as far as I know.

PR #3160, opened 2022-09-13
Allow internal entity classess/interfaces in .NET Standard 2.0

bahusoid:


hazzik (approved):

hazzik:
It was done under an assumption that there could be other runtimes than .NET 4.6 and .NET Core that might not support that.

PR #3171, opened 2022-09-29
Fix fetching lazy component with formula

### LINKED ISSUE #3164: Fetching a lazy loaded component regression
I have identified a regression introduced in version 5.3.0. It happens when trying to fetch a lazy loaded component which has a read-only property and is mapped as an interface.

It's a little querky to explain but I've tried to produce a minimal sample to show you the issue.

Please see the attached application (making sure to set the connection string). The database must have the following structure:

```
CREATE TABLE [dbo].[ContentItems](
	[Id] [int] IDENTITY(1,1) NOT NULL,
	[Name] [nvarchar](100) NOT NULL,
	[PageTitle] [nvarchar](1024) NULL
)
```

[Test NHibernate.zip](https://github.com/nhibernate/nhibernate-core/files/9644422/Test.NHibernate.zip)

bahusoid:
Fixes #3164 

Regression made by #1922
Note: fetching lazy component in 5.2 is not supported but it's processed without errors and component is lazy loaded later. In 5.3 it throws.


fredericDelaporte (approved):

hazzik (approved):

src/NHibernate/Persister/Entity/AbstractEntityPersister.cs
```diff
@@ -1673,14 +1673,10 @@ private string PropertySelectFragment(string name, string suffix, ICollection<st
...
-					// Formulas will have all null values
-					if (columnNames.All(o => o == null))
-					{
-						columnNames = SubclassPropertyFormulaTemplateClosure[index];
-					}
 
-					foreach (var columnName in columnNames)
+					for (var i = 0; i < columnNames.Length; i++)
 					{
+						var columnName = columnNames[i] ?? SubclassPropertyFormulaTemplateClosure[index][i];
```
hazzik, line 1679: Please extract `SubclassPropertyFormulaTemplateClosure[index]` into a field.

bahusoid: Ok. But I wonder if it's really improvement. Array index access is super fast (memory offset) and your change involves stack allocation for temp variable.

hazzik: Compiler should be smart enough to optimize this stack allocation. It is more of a readability issue.

PR #3178, opened 2022-10-11
Don't ignore minimalPut parameter in ReadWriteCache

### LINKED ISSUE #3176: Second-level query cache - entity always fetches lazy loaded data
I have an issue when I wish to retrieve an entity (which has already been cached) but this time I fetch a lazy loaded property it will trigger a hit to database. For example say I have the following models:

```cs
public class Entity {
    public virtual Component? Component { get; set; }
}

public class Component {
    public virtual string Field { get; set; }
}
```

Which are mapped like so:

```cs
public class EntityMapping : ClassMapping<Entity> {
    public EntityMapping() {
        Id(x => x.Id, m => m.Generator(Generators.Identity));
        Property(x => x.Name);
        Component(x => x.Component, m => {
                m.Property(x => x.Field);
                m.Lazy(true);
        });
        Cache(x => x.Usage(CacheUsage.ReadWrite));
    }
}
```

Here's my sample application, indicating each time the database is hit:

```cs
public static async Task Main() {
    using (var sessionFactory = CreateSessionFactory()) {
        using (var session = CreateSession(sessionFactory))
        using (var transaction = session.BeginTransaction()) {
            // Load the entities into the second-level cache.
            session.Query<Entity>().WithOptions(o => o.SetCacheable(true)).ToList(); // Hits the database.

            var result = session.Query<Entity>().WithOptions(o => o.SetCacheable(true)).Fetch(e => e.Head).FirstOrDefault(); // Hits the database.
            var title = result?.Head?.Title;
        }

        using (var session = CreateSession(sessionFactory))
        using (var transaction = session.BeginTransaction()) {
            var result = session.Query<Entity>().WithOptions(o => o.SetCacheable(true)).Fetch(e => e.Head).FirstOrDefault();
            var title = result?.Head?.Title; // Hits the database.
        }
    }
}
```

Notice when I try to retrieve an entity via (which has already been cached), the first time it does a fresh hit to the database (to fetch the lazy loaded data as requested) but the cache is not updated. Now any further calls to the lazy loaded data will also trigger a hit to the database since I assume was because it was not included in the original query which cached the entity.

See https://github.com/nhibernate/nhibernate-core/discussions/3175 for the initial discussion which suggests this is not the intended behaviour.

bahusoid:
Fixes #3176


src/NHibernate/Cache/CacheLock.cs
```diff
@@ -85,10 +85,20 @@ public bool IsPuttable(long txTimestamp, object newVersion, IComparer comparator
 			{
 				return false;
 			}
-			return Version == null ?
```
bahusoid, line 88: Seems `CachedItem` converted to lock `CacheLock` behaves differently for non-versioned entity.

`CachedItem` behaved as `minimalPut` is always true and `CacheLock` as `minimalPut` is always `false` (so `CacheLock` allows to update existing non-versioned entity, and CachedItem - always skips put for existing item)

Not sure if it's a bug - I kept this behavior for now in CacheLock.

P.S. Also I'm a bit confused how UnlockTimestamp and Timeout properties are used..

src/NHibernate.Test/NHSpecificTest/GH3176/Entity.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace NHibernate.Test.NHSpecificTest.GH3176 {
```
hazzik, line 1:
```suggestion
﻿namespace NHibernate.Test.NHSpecificTest.GH3176
﻿{
```

src/NHibernate/Cache/CachedItem.cs
```diff
@@ -104,13 +104,19 @@ public bool IsGettable(long txTimestamp)
 		/// <param name="txTimestamp"></param>
 		/// <param name="newVersion"></param>
 		/// <param name="comparator"></param>
+		/// <param name="minimalPut"></param>
```
hazzik, line 107: These are empty, better to remove all of them.

src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -24,11 +41,20 @@ namespace NHibernate.Cache
 	/// </remarks>
 	public partial class ReadWriteCache : IBatchableCacheConcurrencyStrategy
 	{
+		//TODO 6.0: Merge with ILockable
+		internal interface ILockableNextVer
```
hazzik, line 45: Maybe `IMinimalPutAwareLockable`? I don't like V2, nextVer, etc names. They are too broad. Also, maybe make this an abstract class?

bahusoid: We've already discussed it. It's bad for open API. It's fine for temp internal interface and requires less frictions if additional changes are required for old interface. But OK it's really nit for me in this context let it be `IMinimalPutAwareLockable`.

src/NHibernate/Cache/CacheLock.cs
bahusoid: > CacheLock behaves differently for non-versioned entity

Actually let's try to properly support it

hazzik (approved):

src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -6,6 +7,22 @@
 
 namespace NHibernate.Cache
 {
+	public static class LocableExtension
+	{
+		//TODO 6.0: Remove after ILockableNextVer merge
```
fredericDelaporte, line 12:
```suggestion
		//TODO 6.0: Remove after IMinimalPutAwareLockable merge
```

src/NHibernate/Cache/CacheLock.cs
fredericDelaporte: `UnlockTimestamp` initial value (-1) is irrelevant until Multiplicity has dropped to 0, at which point `UnlockTimestamp` is updated with the unlocking timestamp and starts to be relevant: it avoids a concurrent "stale" transaction to put in cache stale data.

The `Timeout` here is not actually a timeout, it is a `TimeoutTimestamp`: it has been computed from the locking timestamp to which the timeout has been added.

src/NHibernate/Cache/ReadWriteCache.cs
```diff
@@ -24,11 +41,20 @@ namespace NHibernate.Cache
 	/// </remarks>
 	public partial class ReadWriteCache : IBatchableCacheConcurrencyStrategy
```
fredericDelaporte, line 42: The summary comment states:

> Repeatable read isolation is compromised in the case of concurrent writes.

Which means implementing correctly minimal put may cause more violations of repeatable read semantic.
But since it was not guaranteeing it anyway, and since not putting when minimal put is false causes bugs, bette implement it anyway I think.

src/NHibernate/Cache/CacheLock.cs
bahusoid: > it avoids a concurrent "stale" transaction to put stale data in the cache.

So it's OK that `UnlockTimestamp`  is ignored for versioned entity?

fredericDelaporte: If the cached version is older than the one to be put yes. But with your change, it may be as old, in which case this is debattable.

src/NHibernate/Cache/CacheLock.cs
```diff
@@ -85,10 +85,20 @@ public bool IsPuttable(long txTimestamp, object newVersion, IComparer comparator
...
-			       UnlockTimestamp < txTimestamp :
-			       comparator.Compare(Version, newVersion) < 0;
-				//by requiring <, we rely on lock timeout in the case of an unsuccessful update!
+
+			return Version == null 
+				? !minimalPut && UnlockTimestamp < txTimestamp 
+				: comparator.Compare(Version, newVersion) < (minimalPut ? 0 : 1);
+			//by requiring <, we rely on lock timeout in the case of an unsuccessful update!
```
fredericDelaporte, line 92: With your change this comment is no more making sense when `!minimalPut`.

```suggestion
			//by requiring < 0, we rely on lock timeout in the case of an unsuccessful update!
```

src/NHibernate/Cache/CacheLock.cs
bahusoid: > But with your change, it may be as old, in which case this is debattable.

Why old? My change allows only to put the same version (`Compare` returns 0).

fredericDelaporte: I am just meaning the same age: cached version as old as the one to put, (or as new as,) that is just a way to tell "same age". This case is normally useful only for adding data for lazy loaded properties which was not already loaded and cached.

If the version is strictly newer, we know for sure that the cached data is stale and so, there is no harm in updating it with newer data, even if the session putting data in cache is not timestamped after the unlock timestamp. (Better to have less stale data.)

If the version is the same while the unlock timestamp tell us the session putting in cache may be stale, is it better to let other sessions see the additional date the put may add (for lazy properties), or let them hit the db?

fredericDelaporte (approved):

PR #3172, opened 2022-09-29
Skip caching for auto discovery type queries with result transformer

bahusoid:
Fixes #3169  

Exception was triggered by changes in 5.3.10 (https://github.com/nhibernate/nhibernate-core/pull/2943)  (see details https://github.com/nhibernate/nhibernate-core/issues/3169#issuecomment-1262101953)

PR #3184, opened 2022-10-16
Support caching queries with autodiscovered types

### LINKED ISSUE #3169: InvalidOperationException: This transformer is not initialized by Cached Query
Between version 5.3.10 and 5.3.11 the behaviour of cached SQL Query results changed when using AliasToBean transform. 

```csharp
ISession session;
session.CreateSQLQuery("select 'REGIONCODE' as regionCode  ")
		.AddScalar("regionCode", NHibernateUtil.String)
		.SetResultTransformer(Transformers.AliasToBean<ResultDto>())
		.SetCacheable(true)
		.List();
```

This works the first time through when caching is enabled, but when run again it fails with a "System.InvalidOperationException : This transformer is not initialized" exception.

I've created the attached Test Case which gives the exception in the second call to .List in the GetCacheableSqlQueryResults method.

In my actual code the SELECT statement involves tables and join, etc - its been simplified for this test case to not include any entities.  

[GH3169.zip](https://github.com/nhibernate/nhibernate-core/files/9657395/GH3169.zip)

fredericDelaporte:
re-fix #3169


src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -86,8 +85,15 @@ public void Clear()
 		{
 			// 6.0 TODO: inline the call.
 #pragma warning disable 612
-			return Put(key, returnTypes, result, queryParameters.NaturalKeyLookup, session);
+			var cached = Put(key, returnTypes, result, queryParameters.NaturalKeyLookup, session);
```
bahusoid, line 88: Can't we instead of caching additional value wrap result in DTO class with required data.
Something like `Put(..., new AutDiscoveryCacheValue(result, aliases))`

fredericDelaporte: This would be a breaking change for some cache serializers (like the json one), which would need to know about that DTO.

The cached data is an list of objects, mainly arrays of atomic types. So, we may instead handle that in `GetCacheableResult` which already adds the timestamps in that list.

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -510,23 +553,24 @@ public void Destroy()
...
 					break;
 				}
 			}
 			else
 			{
 				var foundTypes = 0;
-				foreach (object[] row in cacheable)
+				foreach (var cached in GetResultsEnumerable(cacheable))
 				{
+					var row = (object[]) cached;
```
fredericDelaporte, line 573: This code was forgetting the timestamp put at index 0, and so, was plain wrong.

src/NHibernate/Loader/Loader.cs
```diff
@@ -1842,8 +1842,7 @@ internal virtual bool IsCacheable(QueryParameters queryParameters)
 
 		internal bool IsCacheable(QueryParameters queryParameters, bool supportsQueryCache, IEnumerable<IPersister> persisters)
 		{
-			bool isCacheable = Factory.Settings.IsQueryCacheEnabled && queryParameters.Cacheable
-                && !(queryParameters.HasAutoDiscoverScalarTypes && queryParameters.ResultTransformer != null);
+			bool isCacheable = Factory.Settings.IsQueryCacheEnabled && queryParameters.Cacheable;
             if (isCacheable && !supportsQueryCache)
```
hazzik, line 1846:
```suggestion
		if (isCacheable && !supportsQueryCache)
```

hazzik:
I think adding prefixes in the middle can break already cached objects because internal format of the cache will change. Can we add indexes to the end? Or what do you think?

src/NHibernate/Cache/StandardQueryCache.cs
bahusoid: > This would be a breaking change for some cache serializers (like the json one), which would need to know about that DTO

But doesn't it mean the same issue applies with your initial implementation? Shouldn't such serializers also know about `QueryAliasesKey`?

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -335,9 +372,9 @@ public void Destroy()
 			ICacheAssembler[] returnTypes,
 			ISessionImplementor session,
 			IList result,
-			long ts)
+			long ts, string[] aliases)
 		{
-			var cacheable = new List<object>(result.Count + 1) { ts };
+			var cacheable = new List<object>(result.Count + 2) { ts, aliases };
```
bahusoid, line 377: Maybe put it as single array item instead?

`var cacheable = new List<object>(result.Count + 1) {new[]{ ts, aliases} };`

hazzik: I think this would work.

fredericDelaporte: Overall it is better, but it is not enough for not breaking if mixed versions of Nhibernate put data in shared cache. the old ones will be broken, the newer one could support the old one data through some specific code, but would it be worth it?

I think we should consider this a possible breaking change, mandating to purge the cache and upgrade all application instances using the cache if it is a distributed cache.

fredericDelaporte: Putting the aliases at the end could be non-breaking for distributed cache scenario mixing NHibernate versions (or not purging the cache on upgrade), but are we willing to really supports this?

Putting them at the end in a non-breaking way would be a bit cumbersome: we would have to add them only for auto-discovered types queries, which are not cacheable with older NHibernate version, so which would not break them. But then the logic to iterate the result list would have to be different according to whether the query is with auto-discovery or not.

src/NHibernate/Cache/StandardQueryCache.cs
fredericDelaporte: Yes I was overlooking this trouble for that new key type.

src/NHibernate/Cache/StandardQueryCache.cs
bahusoid: >  only for auto-discovered types queries

Only for auto-discovered types queries with not null result transformer. 

Anyway we just need to add note about possible breaking change.
As I don't think it's possible to do completely in non-breaking way. It's not cacheable only since 5.3.14 (#3172). It throws in 5.3.11 - 5.3.13 on attempt to retrieve from cache (caused by #2943 which fixed cached auto discovery queries with null transformer)
But in 5.3.0 - 5.3.10 - it's cached successfully but not retrieved from cache. So mixing with  5.3.0 - 5.3.10 would fail.

I'm suggesting single reserved item just because I think it's more clean solution and it would be easier to extend it further in future.

We can also keep backward compatibility for not affected queries using something like:
`var cacheable = new List<object>(result.Count + 1) {aliases == null ? (object)ts : new[]{ ts, aliases} };`
Though I don't think we should really do it (but not against it too)..

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -184,16 +209,22 @@ public IList Get(QueryKey key, ICacheAssembler[] returnTypes, bool isNaturalKeyL
...
 			}
 
 			_cache.PutMany(cachedKeys.ToArray(), cachedResults.ToArray());
 
 			return cached;
 		}
 
+		private static string[] GetAutoDiscoveredAliases(QueryKey key)
+		{
+			return key.ResultTransformer?.AutoDiscoverTypes == true ? key.ResultTransformer.AutoDiscoveredAliases : null;
```
bahusoid, line 225: Why not simply?
```suggestion
			return key.ResultTransformer?.AutoDiscoveredAliases;
```

src/NHibernate/Cache/StandardQueryCache.cs
bahusoid: > Maybe we should add some test using NHibernate Caches instead

Good idea

src/NHibernate/Cache/StandardQueryCache.cs
fredericDelaporte: Just code changes leading to sub-optimal code without me realizing.

src/NHibernate/Cache/StandardQueryCache.cs
hazzik: I think we should do what @bahusoid proposes. For the caches we’re trying to fix nothing will change (the we’re not working anyway), but for other types we would need to flush cache. 

With the changes proposed by Roman we can skip flushing.

src/NHibernate/Cache/StandardQueryCache.cs
fredericDelaporte: Actually no, the key is not serialized, so, no trouble with that `QueryAliasesKey` class, excepted that solution use more code.

src/NHibernate.Test/SqlTest/Query/NativeSQLQueriesFixture.cs
```diff
@@ -294,23 +276,64 @@ void AssertQuery(bool fromCache)
 							.List();
 					t.Commit();
 
-					Assert.AreEqual(1, l.Count);
-					//TODO: Uncomment if we properly fix caching auto discovery type queries with transformers
-					// var msg = "results are expected from " + (fromCache ? "cache" : "DB");
-					// Assert.That(Sfi.Statistics.QueryCacheMissCount, Is.EqualTo(fromCache ? 0 : 1), msg);
-					// Assert.That(Sfi.Statistics.QueryCacheHitCount, Is.EqualTo(fromCache ? 1 : 0), msg);
+					Assert.That(l.Count, Is.EqualTo(1));
```
bahusoid, line 279: Can you also add similar test checking caching for empty list? (basically copy/paste removing  test data creation part and adjusting this check)

src/NHibernate/Cache/StandardQueryCache.cs
bahusoid: > Actually no, the key is not serialized,?

How? Or is it just implementation detail of our cache? Something like `ToString()` call? But then it's just no trouble with this particular implementation...

fredericDelaporte: Depending on the cache, the key is used "as is", or its string representation is used . (It can be appended with its hashcode.)

Prior to latest version, the CoreDistributedCahe was adding the key into the cached data, which was causing the [JsonSerializer to fail](https://github.com/nhibernate/NHibernate-Caches/issues/106). This has been fixed.

src/NHibernate.Test/SqlTest/Query/NativeSQLQueriesFixture.cs
fredericDelaporte: Done, with multiple fixes.

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -219,14 +249,19 @@ public IList Get(QueryKey key, ICacheAssembler[] returnTypes, bool isNaturalKeyL
...
+
 				var querySpaces = spaces[i];
 				if (queryParameters[i].NaturalKeyLookup || querySpaces.Count == 0)
 					continue;
 
 				spacesToCheck.Add(querySpaces);
 				checkedSpacesIndexes.Add(i);
-				// The timestamp is the first element of the cache result.
-				checkedSpacesTimestamp.Add((long) cacheable[0]);
+				checkedSpacesTimestamp.Add(timestamp);
```
fredericDelaporte, line 264: For aliases, we need to get them before short-circuiting if no timestamp checking is needed. So timestamp retrieval is also moved before the short-circuit.

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -271,7 +307,7 @@ public IList Get(QueryKey key, ICacheAssembler[] returnTypes, bool isNaturalKeyL
 
 				for (var i = 0; i < keys.Length; i++)
 				{
-					if (finalReturnTypes[i] == null)
+					if (cacheables[i] == null)
```
fredericDelaporte, line 310: Checking for null return types as a marker of cache miss is wrong for empty result lists with type auto-discovery: their return types are null, but that is no cache miss. So instead, we need to nullify the cacheable when it is a cache miss due to its timestamp.

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -335,9 +371,14 @@ public void Destroy()
...
 			ISessionImplementor session,
 			IList result,
-			long ts)
+			long ts, string[] aliases)
 		{
-			var cacheable = new List<object>(result.Count + 1) { ts };
+			var cacheable =
+				new List<object>(result.Count + 1)
+				{
+					aliases == null ? ts : new object[] { ts, aliases }
```
fredericDelaporte, line 379: I have still to add tests for the JsonSerializer, for checking if a `string[]` will be supported or will have to turned into an `object[]`.

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -492,6 +572,10 @@ public void Destroy()
 			IList cacheable)
 		{
 			Log.Debug("returning cached query results for: {0}", key);
+
+			if (!HasResults(cacheable))
+				return new List<object>();
```
fredericDelaporte, line 577: That short-circuit could be removed, as the methods called here also handle it. (These other methods needs to handle it themselves for the GetMany case.)

src/NHibernate/Transform/CacheableResultTransformer.cs
```diff
@@ -225,12 +232,14 @@ public object TransformTuple(object[] tuple, string[] aliases)
...
 		/// <param name="includeInTuple"></param>
-		/// <returns>transformedResults, with each element re-transformed (if necessary).</returns>
+		/// <returns><paramref name="transformedResults"/>, with each element re-transformed (if necessary).</returns>
 		public IList RetransformResults(IList transformedResults,
 		                                string[] aliases,
 		                                IResultTransformer transformer,
 		                                bool[] includeInTuple)
 		{
+			if (transformedResults.Count == 0)
+				return transformedResults;
```
fredericDelaporte, line 242: This short-circuit, and the one in `UntransformToTuples`, could be removed for a single one in the `Loader`, which is the only caller of these methods. But it seems more appropriate to me to put them here, especially since their code documentation states:

> If not necessary, the original List is returned unchanged.

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -376,7 +376,7 @@ public void Destroy()
 			var cacheable =
 				new List<object>(result.Count + 1)
 				{
-					aliases == null ? ts : new object[] { ts, aliases }
+					aliases == null ? ts : new object[] { ts, aliases.ToArray<object>() }
```
bahusoid, line 379: `ToArray` call seems not needed. We still can read it as `object[]` after de-serialization without this transformation..

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -346,23 +387,56 @@ public void Destroy()
...
 		}
 
+		private static long GetResultsMetadata(IList cacheable, out string[] aliases)
+		{
+			aliases = null;
+
+			var metadata = cacheable[0];
+			var timestamp = metadata as long?;
+			if (timestamp.HasValue)
+				return timestamp.Value;
```
bahusoid, line 404:
```suggestion
			if (metadata is long timestamp)
				return timestamp;
```

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -510,23 +594,24 @@ public void Destroy()
...
-			var colCount = (cacheable[0] as object[])?.Length ?? 1;
+			var colCount = (GetResultsEnumerable(cacheable).First() as object[])?.Length ?? 1;
 			var returnTypes = new ICacheAssembler[colCount];
 			if (colCount == 1)
 			{
-				foreach (var obj in cacheable)
+				foreach (var cached in GetResultsEnumerable(cacheable))
 				{
-					if (obj == null)
+					if (cached == null)
```
bahusoid, line 603: Huh... We have different treatment for single column vs multi column result(object vs object[]).. So it's better to add test for colCount > 1 too...

src/NHibernate.Test/CacheTest/JsonSerializerCacheFixture.cs
```diff
@@ -0,0 +1,525 @@
...
+using NHibernate.Linq;
+using NHibernate.Multi;
+using NHibernate.Transform;
+using NUnit.Framework;
+using Environment = NHibernate.Cfg.Environment;
+
+namespace NHibernate.Test.CacheTest
+{
+	[TestFixture]
+	public class JsonSerializerCacheFixture : TestCase
```
bahusoid, line 17: Why it's implemented as separate new class? Can't it be based on existing one (NativeSQLQueriesFixture?) with configuration for using JsonSerializer cache

fredericDelaporte: It is a mix of NativeSQLQueriesFixture and BatchableCacheFixture, for checking non-sql-native queries too.

src/NHibernate/Cache/StandardQueryCache.cs
fredericDelaporte: [This](https://github.com/nhibernate/nhibernate-core/pull/3184/files#diff-ca8980fb225a4228f73f304c834c6d4793a8d2370627ab0cd99c4bc63e663b30R97) fails without this call, because a `string[]` ends up in the cached value, which is not supported by the JsonCacheSerializer (unless adding a registration for it, but then, that means this fix here is a breaking change).

src/NHibernate/Cache/StandardQueryCache.cs
fredericDelaporte: Added.

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -456,6 +531,9 @@ public void Destroy()
 			IList assembleResult,
 			IList cacheResult)
 		{
+			if (!HasResults(cacheResult))
```
hazzik, line 534: Maybe rename it to `IsEmpty`?

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -346,23 +387,55 @@ public void Destroy()
...
+		private static long GetResultsMetadata(IList cacheable, out string[] aliases)
+		{
+			aliases = null;
+
+			var metadata = cacheable[0];
+			if (metadata is long timestamp)
+				return timestamp;
+
+			var metadataArray = (object[]) metadata;
+			aliases = ((object[]) metadataArray[1]).Cast<string>().ToArray();
```
hazzik, line 406: Would this work?
```suggestion
			aliases = ((IEnumerable) metadataArray[1]).ToArray<string>();
```

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -335,9 +371,14 @@ public void Destroy()
 			ICacheAssembler[] returnTypes,
 			ISessionImplementor session,
 			IList result,
-			long ts)
+			long ts, string[] aliases)
```
hazzik, line 374:
```suggestion
			long ts,
			string[] aliases)
```

src/NHibernate/Cache/StandardQueryCache.cs
```diff
@@ -219,14 +249,19 @@ public IList Get(QueryKey key, ICacheAssembler[] returnTypes, bool isNaturalKeyL
 					continue;
 				}
 
+				var timestamp = GetResultsMetadata(cacheable, out var aliases); var key = keys[i];
```
hazzik, line 252:
```suggestion
				var timestamp = GetResultsMetadata(cacheable, out var aliases);
				var key = keys[i];
```

src/NHibernate/Cache/StandardQueryCache.cs
bahusoid: Ok. But it looks like an implementation flaw to me. Who registers `object[]` as supported type? This code should treat any arrays as `object[]` which .NET supports natively.

src/NHibernate/Cache/StandardQueryCache.cs
bahusoid: There is no LINQ extensions on `IEnumerable` except casting one to generic one Cast/OfType

hazzik: I think it should work for `IEnumerable<object>`

fredericDelaporte: No, `ToArray` definition is `TSource[] ToArray<TSource>(this IEnumerable<TSource>)`, so only a `ToArray<object>` can compile without a cast.

src/NHibernate/Cache/StandardQueryCache.cs
fredericDelaporte: This security feature quite bother me, but since I have not studied what attacks may be possible by relaxing it, I just do with it.

We could also choose to do a new release of JsonCacheSerializer with a few more innocuous types registered like `string[]` or even all basic atomic type arrays. Then tell as a possible breaking change that users of that serializer needs to upgrade to its latest version and check their own registration for potential conflicts with new built-in registrations. But so, that would mean a possible breaking change here in NHibernate and another one in the caches.

src/NHibernate/Cache/StandardQueryCache.cs
fredericDelaporte: Oops (my turn).

bahusoid (approved):

hazzik (approved):

bahusoid (approved):

PR #3203, opened 2022-11-22
Fix a wrong example in configuration documentation

janv8000:


fredericDelaporte:
Side note: that reminds me the 5.4 doc still has to be published.

fredericDelaporte (approved):
Indeed a bad copy-paste, thanks.

fredericDelaporte:
I have re-targeted it to 5.4.x. I would publish the 5.4 doc after that fix, likely without waiting for 5.4.1 release.

PR #3209, opened 2022-12-09
Allow custom query loader

micmerchant:
Hello,

I'm currently working on a NHibernate-Linq plugin to support [MSSQL Table-Valued-Functions](https://learn.microsoft.com/en-us/sql/relational-databases/user-defined-functions/user-defined-functions?view=sql-server-ver16#table-valued-functions), which I plan to release as open source software.

My entry points are the following ones:
`<property name="query.linq_provider_class"/>`
`<property name="query.factory_class"/>`

The plugin propagates the Table-Value parameters down to a custom query loader via the following steps
1) inherited `DefaultQueryProvider` implementation to override the `PrepareQuery` method
2) decorated `NhLinqExpression` to keep track of 
  - `ParameterDescriptors`
  - `ParameterValuesByName`
  - `NamedParameters`
and to pass the query/parameter verification checks.
3) custom query translator factory to create a special query translator
4) expand the `ParameterMetaData` by Table-Valued parameters via decorated `BuildParameterMetadata` method of query translator
5) inherited `QueryLoader` implementation to expand parameters by Table-Valued parameters via overridden `CreateSqlCommand` method. Expanding the parameters works quite similar like the session filter parameter implementation.

But to get this working and not to use reflection, I need some changes to support Dependency Inversion in the following areas:
1) A custom query loader must be injectable into the query translator. This requires the following changes:
  - change some access modifiers from internal to public
  - extracted `ILoader` and `IQueryLoader` interfaces
  - `QueryTranslatorImpl` constructor extended by a factory method to create an `IQueryLoader` on demand
2) To support query plan caching and proper copying of query plan/expression also for Table-Valued Function queries, the following changes are necessary:
  - `NhLinqExpression` implements `ILinqQueryExpression`, which extends `IQueryExpression, ICacheableQueryExpression`
  - checks changed from `NhLinqExpression` to `ILinqQueryExpression` in `QueryPlanCache::PreparePlanToCache, PreparePlanToCache::CopyIfRequired`
  - `CopyExpressionTranslation` public accessible via `ILinqQueryExpression` now
  - `NhLinqExpressionCache` depends on `ILinqQueryExpression` now instead of `NhLinqExpression`
  - due to interface extraction the access modifier of some other classes/interfaces had to be changed to public

Basically, the functionality of NHibernate has not changed at all. Some classes depend on abstractions now instead on concrete implementations. 

What do you think about these changes?

BR,
Michael


src/NHibernate/Hql/Ast/ANTLR/ASTQueryTranslatorFactory.cs
```diff
@@ -33,8 +34,17 @@ public IQueryTranslator[] CreateQueryTranslators(IQueryExpression queryExpressio
 
 			var translators = polymorphicParsers
 				.ToArray(hql => queryExpression is NhLinqExpression linqExpression
```
fredericDelaporte, line 36: Don't you need to change this cast to `ILinqQueryExpression` too?

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -48,12 +49,14 @@ public partial class QueryTranslatorImpl : IFilterTranslator
...
 		/// <param name="factory">The session factory constructing this translator instance.</param>
+		/// <param name="queryLoaderFactory">A query loader factory method.</param>
 		public QueryTranslatorImpl(
 			string queryIdentifier,
 			IASTNode parsedQuery,
 			IDictionary<string, IFilter> enabledFilters,
-			ISessionFactoryImplementor factory)
-			: this(queryIdentifier, parsedQuery, enabledFilters, factory, null)
+			ISessionFactoryImplementor factory,
+			Func<QueryTranslatorImpl, ISessionFactoryImplementor, SelectClause, IQueryLoader> queryLoaderFactory)
```
fredericDelaporte, line 58: It is not a pattern in-use in NHibernate. It is hard to supply by configuration a factory which is just a method pointer, while we can easily do this for a factory interface. So, I think you should instead use something like:

```suggestion
			IQueryLoaderFactory queryLoaderFactory)
```

with:
```
/// <summary>
/// Creates query loaders.
/// <summary>
public interface IQueryLoaderFactory
{
    /// <summary>
    /// Creates a query loader.
    /// <summary>
    /// ...
    IQueryLoader Create(QueryTranslatorImpl queryTranslator, ISessionFactoryImplementor factory, SelectClause selectClause); 
}

File: src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs [hunk 3]
```diff
@@ -48,12 +49,14 @@ public partial class QueryTranslatorImpl : IFilterTranslator
 		/// <param name="parsedQuery">The hql query to translate</param>
 		/// <param name="enabledFilters">Currently enabled filters</param>
 		/// <param name="factory">The session factory constructing this translator instance.</param>
+		/// <param name="queryLoaderFactory">A query loader factory method.</param>
 		public QueryTranslatorImpl(
```
fredericDelaporte, line 53: As this constructor is public, an overload without the new parameter and adequate implementation for it to work properly must be kept, unless this change targets a new major version of NHibernate.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -192,7 +198,7 @@ private void ErrorIfSelect()
 			}
 		}
 
-		public NHibernate.Loader.Loader Loader
+		public ILoader Loader
```
fredericDelaporte, line 201: Another breaking change. If willing to merge this for 5.5, the property `Loader` cannot change its type. It seems this change will need a new major version.

src/NHibernate/IQueryExpression.cs
```diff
@@ -6,7 +6,7 @@
 namespace NHibernate
 {
 	//TODO 6.0: Merge into IQueryExpression
-	internal interface ICacheableQueryExpression
+	public interface ICacheableQueryExpression
```
fredericDelaporte, line 9: As explained elsewhere, this interface is just transitional, meant to avoid a binary breaking change in a public interface until v6.0, and should disappear then. So, if the changes int this PR anyway require a v6.0, it is better to merge it in IQueryExpression and get rid of it.

src/NHibernate/Loader/Loader.cs
```diff
@@ -646,9 +646,11 @@ private IEnumerable<SubselectFetch> CreateSubselects(List<EntityKey[]> keys, Que
...
 		}
 
-		internal void InitializeEntitiesAndCollections(
-			IList hydratedObjects, DbDataReader reader, ISessionImplementor session, bool readOnly,
-			CacheBatcher cacheBatcher)
+		public void InitializeEntitiesAndCollections(IList hydratedObjects, 
+		                                             DbDataReader reader, 
+		                                             ISessionImplementor session, 
+		                                             bool readOnly,
+		                                             CacheBatcher cacheBatcher)
```
fredericDelaporte, line 653: The previous formatting has a reason: with tab indentation, it is a lot more stable than this new formatting, which tends to "break" when some reformatting tool change spaces back to tabs. Please keep previous formatting, or change it with first parameter on its own line:

```suggestion
		public void InitializeEntitiesAndCollections(
			IList hydratedObjects, 
			DbDataReader reader, 
			ISessionImplementor session, 
			bool readOnly,
			CacheBatcher cacheBatcher)
```

src/NHibernate/Loader/Loader.cs
```diff
@@ -379,12 +379,12 @@ internal static EntityKey GetOptionalObjectKey(QueryParameters queryParameters,
...
-											EntityKey optionalObjectKey, IList hydratedObjects, EntityKey[] keys,
-											bool returnProxies, IResultTransformer forcedResultTransformer,
-											QueryCacheResultBuilder queryCacheResultBuilder,
-		                                    Action<IEntityPersister, CachePutData> cacheBatchingHandler)
+		public object GetRowFromResultSet(DbDataReader resultSet, ISessionImplementor session,
+		                                  QueryParameters queryParameters, LockMode[] lockModeArray,
+		                                  EntityKey optionalObjectKey, IList hydratedObjects, EntityKey[] keys,
+		                                  bool returnProxies, IResultTransformer forcedResultTransformer,
+		                                  QueryCacheResultBuilder queryCacheResultBuilder,
+		                                  Action<IEntityPersister, CachePutData> cacheBatchingHandler)
```
fredericDelaporte, line 387: That formatting may not be kept depending on the developer IDE (VS's formatting vs Rider, ...). For avoiding new undue noise at next change, it is better to format this with tab only, like so:

```suggestion
		public object GetRowFromResultSet(
			DbDataReader resultSet, ISessionImplementor session,
			QueryParameters queryParameters, LockMode[] lockModeArray,
			EntityKey optionalObjectKey, IList hydratedObjects, EntityKey[] keys,
			bool returnProxies, IResultTransformer forcedResultTransformer,
			QueryCacheResultBuilder queryCacheResultBuilder,
			Action<IEntityPersister, CachePutData> cacheBatchingHandler)
```

src/NHibernate/Loader/Loader.cs
```diff
@@ -1563,8 +1565,9 @@ protected internal virtual void AutoDiscoverTypes(DbDataReader rs)
 			AutoDiscoverTypes(rs, new QueryParameters(), null);
 		}
 
-		protected internal virtual void AutoDiscoverTypes(
-			DbDataReader rs, QueryParameters queryParameters, IResultTransformer forcedResultTransformer)
+		public virtual void AutoDiscoverTypes(DbDataReader rs, 
+		                                      QueryParameters queryParameters,
+		                                      IResultTransformer forcedResultTransformer)
```
fredericDelaporte, line 1570: Revert the formatting.

src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -91,9 +92,8 @@ protected class QueryInfo : ICachingInformation, ICachingInformationWithFetches
 			/// <param name="loader">The loader.</param>
 			/// <param name="querySpaces">The query spaces.</param>
 			/// <param name="session">The session of the query.</param>
-			public QueryInfo(
-				QueryParameters parameters, Loader.Loader loader, ISet<string> querySpaces,
-				ISessionImplementor session)
+			public QueryInfo(QueryParameters parameters, ILoader loader, ISet<string> querySpaces,
+			                 ISessionImplementor session)
```
fredericDelaporte, line 96: Revert the formatting.

(I have ceased flagging binary breaking changes, it seems this PR will not avoid them and will have to be merged only for 6.0.)

src/NHibernate/NHibernate.csproj
```diff
@@ -13,6 +13,7 @@
     <PackageDescription>NHibernate is a mature, open source object-relational mapper for the .NET framework. It is actively developed, fully featured and used in thousands of successful projects.</PackageDescription>
     <PackageTags>ORM; O/RM; DataBase; DAL; ObjectRelationalMapping; NHibernate; ADO.Net; Core</PackageTags>
     <PackageReadmeFile>NHibernate.readme.md</PackageReadmeFile>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
fredericDelaporte, line 16: Why this change?

fredericDelaporte (commented):
Tests for the change should be added: check that providing a custom query loader factory works as expected, for all querying interface NHibernate has. I suspect it may break where I have asked about a possible missing type change.

Also, the possibility of supplying a custom query loader factory should likely be surfaced to the NHibernate configuration, in a similar way to what we do for most other factories (see `linqtohql.generatorsregistry`, `collectiontype.factory_class`, `query.factory_class`, ...).

build-common/NHibernate.props
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <NhVersion     Condition="'$(NhVersion)' == ''"    >5.4</NhVersion>
-    <VersionPatch  Condition="'$(VersionPatch)'  == ''">0</VersionPatch>
+    <VersionPatch  Condition="'$(VersionPatch)'  == ''">1</VersionPatch>
```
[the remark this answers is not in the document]
micmerchant: Sorry, of course my fault. Was for building a custom NuGet package for internal testing.

src/NHibernate/NHibernate.csproj
micmerchant: Sorry, of course my fault. Was for building a custom NuGet package for internal testing.

src/NHibernate/Loader/Loader.cs
micmerchant: OK. Do you prefer one parameter each line like 
```
		public void InitializeEntitiesAndCollections(
			IList hydratedObjects, 
			DbDataReader reader, 
			ISessionImplementor session, 
			bool readOnly,
			CacheBatcher cacheBatcher)
```
, or multiple ones?
```
public object GetRowFromResultSet(
			DbDataReader resultSet, ISessionImplementor session,
			QueryParameters queryParameters, LockMode[] lockModeArray,
			EntityKey optionalObjectKey, IList hydratedObjects, EntityKey[] keys,
			bool returnProxies, IResultTransformer forcedResultTransformer,
			QueryCacheResultBuilder queryCacheResultBuilder,
			Action<IEntityPersister, CachePutData> cacheBatchingHandler)
```

src/NHibernate/Engine/Query/QueryPlanCache.cs
```diff
@@ -69,10 +69,14 @@ public IQueryExpressionPlan GetHQLQueryPlan(IQueryExpression queryExpression, bo
 
 				plan = new QueryExpressionPlan(queryExpression, shallow, enabledFilters, factory);
 				// 6.0 TODO: add "CanCachePlan { get; }" to IQueryExpression interface
-				if (!(queryExpression is ICacheableQueryExpression linqExpression) || linqExpression.CanCachePlan)
+				if(queryExpression is ICacheableQueryExpression linqExpression && linqExpression.CanCachePlan)
```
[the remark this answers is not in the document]
micmerchant: Yeah, I've thought about this several times and also talked to a colleague about it. I really wasn't sure if this piece of code of you was intended or not. 

But honestly, my change wasn't meant to be committed. I simple forgot to revert it and want to apologize for wasting your time for this and other e.g. format changes.

Beside of that, this would be a clear change of the functionality of NHibernate.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
micmerchant: What about this solution?

```suggestion
		//Interface:
		
		// 6.0 TODO : remove
		Loader.Loader Loader { get; }
			
		// 6.0 TODO : rename to Loader when overloaded property gets removed
		ILoader QueryLoader { get; }
		
		// Implementation:
		
		// 6.0 TODO : remove
		public Loader.Loader Loader => QueryLoader as Loader.Loader ?? throw new NotSupportedException("Loader is not of supported type.");

		// 6.0 TODO : rename to Loader when overloaded property gets removed
		public ILoader QueryLoader => _queryLoader;
```

I don't really like the cast, but don't see a way around it without breaking the interface.

src/NHibernate/Multi/QueryBatchItemBase.cs
micmerchant: Possible solution without breaking it.

```suggestion
		
			// 6.0 TODO : remove
			
			/// <summary>
			/// Create a new <c>QueryInfo</c>.
			/// </summary>
			/// <param name="parameters">The query parameters.</param>
			/// <param name="loader">The loader.</param>
			/// <param name="querySpaces">The query spaces.</param>
			/// <param name="session">The session of the query.</param>
			public QueryInfo(
				QueryParameters parameters, Loader.Loader loader, ISet<string> querySpaces,
				ISessionImplementor session)
				:this(parameters, (ILoader)loader, querySpaces, session)
			{
				
			}

			/// <summary>
			/// Create a new <c>QueryInfo</c>.
			/// </summary>
			/// <param name="parameters">The query parameters.</param>
			/// <param name="loader">The loader.</param>
			/// <param name="querySpaces">The query spaces.</param>
			/// <param name="session">The session of the query.</param>
			public QueryInfo(
				QueryParameters parameters, ILoader loader, ISet<string> querySpaces,
				ISessionImplementor session)
			{
				Parameters = parameters;
				Loader = loader;
				QuerySpaces = querySpaces;

				IsCacheable = loader.IsCacheable(parameters);
				if (!IsCacheable)
					return;

				CacheKey = Loader.GenerateQueryKey(session, Parameters);
				CanGetFromCache = Parameters.CanGetFromCache(session);
				CanPutToCache = Parameters.CanPutToCache(session);
			}
```

micmerchant:
> Tests for the change should be added: check that providing a custom query loader factory works as expected, for all querying interface NHibernate has. I suspect it may break where I have asked about a possible missing type change.
> 
> Also, the possibility of supplying a custom query loader factory should likely be surfaced to the NHibernate configuration, in a similar way to what we do for most other factories (see `linqtohql.generatorsregistry`, `collectiontype.factory_class`, `query.factory_class`, ...).

1) I'll add some tests.
2) Yes, you are right, the query loader could be made configurable. But this would also require a "major" documentation update.

src/NHibernate/Cache/QueryCacheResultBuilder.cs
```diff
@@ -14,12 +15,12 @@ public sealed class QueryCacheResultBuilder
 		private readonly IType[] _resultTypes;
 		private readonly Loader.Loader.QueryCacheInfo _cacheInfo;
 
-		public static bool IsCacheWithFetches(Loader.Loader loader)
+		public static bool IsCacheWithFetches(ILoader loader)
 		{
 			return loader.CacheTypes.Length > loader.ResultTypes.Length;
 		}
```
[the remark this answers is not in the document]
micmerchant: My IDE tells me that we need a cast, to not be a recursive call.

```suggestion
		// 6.0 TODO : remove
		public static bool IsCacheWithFetches(Loader.Loader loader)
			=> IsCacheWithFetches((ILoader)loader);
		
		public static bool IsCacheWithFetches(ILoader loader)
		{
			return loader.CacheTypes.Length > loader.ResultTypes.Length;
		}
```

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
micmerchant: Nice solution.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
micmerchant:
```suggestion
				/// <summary>
		/// Creates a new AST-based query translator.
		/// </summary>
		/// <param name="queryIdentifier">The query-identifier (used in stats collection)</param>
		/// <param name="parsedQuery">The hql query to translate</param>
		/// <param name="enabledFilters">Currently enabled filters</param>
		/// <param name="factory">The session factory constructing this translator instance.</param>
		public QueryTranslatorImpl(
			string queryIdentifier,
			IASTNode parsedQuery,
			IDictionary<string, IFilter> enabledFilters,
			ISessionFactoryImplementor factory)
			: this(queryIdentifier, parsedQuery, enabledFilters, factory, new QueryLoaderFactory(), null)
		{
		}

		/// <summary>
		/// Creates a new AST-based query translator.
		/// </summary>
		/// <param name="queryIdentifier">The query-identifier (used in stats collection)</param>
		/// <param name="parsedQuery">The hql query to translate</param>
		/// <param name="enabledFilters">Currently enabled filters</param>
		/// <param name="factory">The session factory constructing this translator instance.</param>
		/// <param name="queryLoaderFactory">The query loader factory.</param>
		public QueryTranslatorImpl(
			string queryIdentifier,
			IASTNode parsedQuery,
			IDictionary<string, IFilter> enabledFilters,
			ISessionFactoryImplementor factory,
			IQueryLoaderFactory queryLoaderFactory)
			: this(queryIdentifier, parsedQuery, enabledFilters, factory, queryLoaderFactory, null)
		{
		}

		/// <summary>
		/// Creates a new AST-based query translator.
		/// </summary>
		/// <param name="queryIdentifier">The query-identifier (used in stats collection)</param>
		/// <param name="parsedQuery">The hql query to translate</param>
		/// <param name="enabledFilters">Currently enabled filters</param>
		/// <param name="factory">The session factory constructing this translator instance.</param>
		/// <param name="queryLoaderFactory">The query loader factory.</param>
		/// <param name="namedParameters">The named parameters information.</param>
		public QueryTranslatorImpl(
			string queryIdentifier,
			IASTNode parsedQuery,
			IDictionary<string, IFilter> enabledFilters,
			ISessionFactoryImplementor factory,
			IQueryLoaderFactory queryLoaderFactory,
			IDictionary<string, Tuple<IType, bool>> namedParameters)
		{
			_queryIdentifier = queryIdentifier;
			_stageOneAst = parsedQuery;
			_compiled = false;
			_shallowQuery = false;
			_enabledFilters = enabledFilters;
			_factory = factory;
			_queryLoaderFactory = queryLoaderFactory;
			_namedParameters = namedParameters;
		}
```

QueryLoaderFactory implements your suggested interface. This should ensure backwards compatibility.

src/NHibernate/IQueryExpression.cs
micmerchant: OK, i could decouple the ICacheableQueryExpression definition from the ILinqQueryExpression again and revert it back to internal accessibility. But caching of custom ILinqQueryExpression implementations would only work cause all expressions not implementing ICacheableQueryExpression are cached per default. This seems a bit misleading to me. But you mentioned that already earlier.

The changes to copy the plan/expression should also work with the suggested change.

```suggestion
	public interface ILinqQueryExpression: IQueryExpression
	public class NhLinqExpression : ILinqQueryExpression, ICacheableQueryExpression	
```

src/NHibernate/Cache/QueryCacheResultBuilder.cs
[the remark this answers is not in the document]
fredericDelaporte: Yes the cast is needed indeed.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
fredericDelaporte: Ok.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
fredericDelaporte: Yes, adding a member to an interface is a binary breaking change, due to the lack of a default implementation for external implementations of the interface.

But we are speaking of which interface? `ILoader`? Since it is introduced here, there is no trouble changing it until a release with it is done. And for 6.0, binary breaking changes will be allowed, so the todo would not be an issue.

So the new member `QueryLoader` would have to be added to a new "transitional interface".

About the cast, provided the code make sure to use the new `QueryLoader` property when handling an `ILoader`, instead of calling the old `Loader` property, it should be fine. For helping not missing this, the old property should be flagged with the `Obsolete` attribute, instructing to use `QueryLoader`. For avoiding a source breaking change with 6.0 event for those not calling any obsolete member, the TODO should be change to just "Remove Loader property" without the renaming todo on `QueryLoader`.

src/NHibernate/Loader/Loader.cs
fredericDelaporte: The later, at least when there are that many parameters.

src/NHibernate/Multi/QueryBatchItemBase.cs
fredericDelaporte: Ok.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
micmerchant: No, the problem is the `public partial interface IQueryTranslator`. `IQueryTranslator` currently (Master) provides the property `Loader.Loader Loader { get; }`, which must be still supported if no breaking changes should be introduced.
So `QueryTranslatorImpl` has to implement this property, but the backing field is of type `IQueryLoader` now, due to the `IQueryLoaderFactory` introduction. This makes a cast from `IQueryLoader to Loader.Loader` necessary now.

**Honestly, I don't want to wait for a major release :)** So I would suggest the following change for `IQueryTranslator/QueryTranslatorImpl`:
```suggestion
		//IQueryTranslator:

		// 6.0 TODO : change type to ILoader
		Loader.Loader Loader { get; }
		
		// QueryTranslatorImpl:
		
		// 6.0 TODO : change type to ILoader
		public Loader.Loader Loader => QueryLoader as Loader.Loader ?? throw new NotSupportedException("Loader is not of supported type.");
```

**But the cast looks kinda like a code smell.** 

Alternativly, as u suggest we could add a new transitional interface which provides the above QueryLoader property of type ILoader, but some classes/interfaces are depending on `Loader.Loader` e.g.
image
So a new transitional interface wouldn't help much.

src/NHibernate/Hql/Ast/ANTLR/ASTQueryTranslatorFactory.cs
micmerchant: I guess you are right. I'll change that, but have to publish the `GetNamedParameterTypes()` via the `ILinqQueryExpression` of course too.

micmerchant:
I've added tests that should use a custom `QueryTranslatorFactory` in combination with a custom `QueryLoader` for `Criteria, HQL, Linq and QueryOver` queries. I've not added a NHibernate config option for custom `QueryLoaders`.

But, it seems that `Criteria` and `QueryOver` queries don't depend on a `QueryTranslatorFactory`, so these two TestCases are kinda useless. Only `Linq` and `HQL` queries instantiate a `QueryTranslatorFactory`.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
```diff
@@ -191,11 +213,9 @@ private void ErrorIfSelect()
...
 			}
 		}
-
-		public NHibernate.Loader.Loader Loader
-		{
-			get { return _queryLoader; }
-		}
+		
+		// 6.0 TODO : change type to ILoader
+		public Loader.Loader Loader => _queryLoader as Loader.Loader ?? throw new NotSupportedException("Loader is not of supported type.");
```
fredericDelaporte, line 218: This property is used by the multi-query batcher (so, by futures too). This implementation means custom query loaders not deriving from the Loader class will not be supported when using the query batching feature. (Either explicitly through the use of a query batch or implicitly through the use of futures.)

Can you add some test about this?

fredericDelaporte: For supporting them, it will likely be needed to add an `ILoader QueryLoader` property to this class, obsolete the `Loader` one and use the new property "everywhere" the older one was used.

Of course adding that QueryLoader property to the IQueryTranslator interface would be a breaking change, so if in need of adding it there, a new transitional interface would be needed.

src/NHibernate/Multi/QueryBatchItemBase.cs
```diff
@@ -31,7 +32,7 @@ protected class QueryInfo : ICachingInformation, ICachingInformationWithFetches
 			/// <summary>
 			/// The query loader.
 			/// </summary>
-			public Loader.Loader Loader { get; set; }
+			public ILoader Loader { get; set; }
```
fredericDelaporte, line 35: This change here needs also to be avoided.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
micmerchant: You are right, this implementation causes some troubles with Future-Queries. 

Here is a possible solution, where the underlying QueryLoader of a decorated QueryLoader can be exposed and used:
```suggestion

		// 6.0 TODO : change type to ILoader
		public Loader.Loader Loader => _queryLoader.GetUnderlyingQueryLoader() as Loader.Loader ?? throw new NotSupportedException("Loader is not of supported type.");

	/// <summary>
	/// Defines a query loader.
	/// </summary>
	public partial interface IQueryLoader: ILoader
	{	
		// 6.0 TODO : Remove when Loader.Loader properties are replaced by ILoader properties.
		/// <summary>
		/// Returns the underlying query loader.
		///
		/// E.g. When a query loader gets decorated, it is possible to return the decoratee here. 
		/// </summary>
		/// <returns></returns>
		IQueryLoader GetUnderlyingQueryLoader();
	}
	
	public partial class QueryLoader : BasicLoader, IQueryLoader
	{
		// 6.0 TODO : Remove when Loader.Loader properties are replaced by ILoader properties.
		/// <summary>
		/// Returns the underlying query loader which is the current implementation.
		/// </summary>
		/// <returns></returns>
		public virtual IQueryLoader GetUnderlyingQueryLoader()
		{
			return this;
		}
	}
	
	internal sealed class CustomQueryLoader: IQueryLoader
	{
		private readonly IQueryLoader _queryLoader;

		public CustomQueryLoader(IQueryLoader queryLoader)
		{
			_queryLoader = queryLoader;
		}

		public IQueryLoader GetUnderlyingQueryLoader()
		{
			return _queryLoader;
		}
	}
	
	Test:
		[Test]
		public void FutureQueryTest()
		{
			var customers = _session.CreateQuery("select c from Customer c")
			                        .Future<Customer>()
			                        .ToList();
			
			Assert.AreEqual(1, customers.Count);
		}
```

src/NHibernate/Multi/QueryBatchItemBase.cs
micmerchant: Type of property reverted and additional constructor removed again.

src/NHibernate/Hql/Ast/ANTLR/QueryTranslatorImpl.cs
fredericDelaporte: That does not sound to me as complete fix, and it seem more complex than what I propose for implementors. (But allow for less changes in NHibernate code.) I'll try to spare some time to push a commit with what I think could be a better solution.

fredericDelaporte:
I have pushed the commit I had written about.

I got by the way a better view of what was actually needed. So, about some of my previous comments:

> Also, the possibility of supplying a custom query loader factory should likely be surfaced to the NHibernate configuration, in a similar way to what we do for most other factories (see `linqtohql.generatorsregistry`, `collectiontype.factory_class`, `query.factory_class`, ...).

It is indeed not needed, the current translator factory is enough.

src/NHibernate/Persister/IPersister.cs
```diff
@@ -5,7 +5,7 @@
 namespace NHibernate.Persister
 {
 	// TODO 6.0: Make this public and make IEntityPersister and ICollectionPersister derive it.
```
fredericDelaporte, line 7: Comment now slightly obsolete:

```suggestion
	// TODO 6.0: Make IEntityPersister and ICollectionPersister derive it.
```

micmerchant:
> I have pushed the commit I had written about.
> 
> I got by the way a better view of what was actually needed. So, about some of my previous comments:
> 
> > Also, the possibility of supplying a custom query loader factory should likely be surfaced to the NHibernate configuration, in a similar way to what we do for most other factories (see `linqtohql.generatorsregistry`, `collectiontype.factory_class`, `query.factory_class`, ...).
> 
> It is indeed not needed, the current translator factory is enough.
> 
> > For all possible querying interface
> 
> Indeed a bit broad since that change does apply neither to Criteria nor QueryOver (nor native queries).

I'm glad that you got a better view on it and I thankful for your commit. I'll still add the FutureQuery-Test to ensure that there it is no breaking change.

micmerchant:
I think `HqlTranslatorWrapper` needs to implement `ITranslatorWithCustomizableLoader` to use the `QueryLoader` property. Now the new future query test runs also successfully.
```
	internal class HqlTranslatorWrapper : ITranslatorWithCustomizableLoader
	{
		/// <inheritdoc />
		public ILoader QueryLoader => innerTranslator.GetQueryLoader();
```

src/NHibernate.Test/QueryTranslator/CustomQueryLoaderFixture.cs
```diff
@@ -0,0 +1,115 @@
...
+			_session.Dispose();
+		}
+
+		[Test(Description = "Tests criteria queries.")]
+		public void CriteriaQueryTest()
+		{
+			var customers = _session.CreateCriteria(typeof(Customer))
+			                        .List<Customer>();
+			
+			Assert.AreEqual(1, customers.Count);
```
fredericDelaporte, line 75: I forgot about that: we now use only `Assert.That` in our new tests. It tends to be more descriptive than the short-hands.

src/NHibernate/Impl/MultiQueryImpl.cs
```diff
@@ -802,13 +803,61 @@ private int AddQueryForLaterExecutionAndReturnIndexOfQuery(System.Type resultGen
...
+				Log.Warn("{0} is obsolete, it should implement {1} to support customizable loaders", translator, nameof(ITranslatorWithCustomizableLoader));
+			}
+
+#pragma warning disable CS0618 // Type or member is obsolete
+			return translator.Loader;
+#pragma warning restore CS0618 // Type or member is obsolete
+		}
+	}
+
+	internal class HqlTranslatorWrapper : ITranslatorWithCustomizableLoader
```
fredericDelaporte, line 860: Yes I forgot to use the new interface... For consistency, I am adding it on the other translator implementation, even if it does not need it. It will reduce the work when doing the todo.

src/NHibernate.Test/QueryTranslator/CustomQueryLoaderFixture.cs
```diff
@@ -0,0 +1,115 @@
...
+		}
+		
+		[Test(Description = "Tests future queries.")]
+		public void FutureQueryTest()
+		{
+			var customers = _session.CreateQuery("select c from Customer c")
+			                        .Future<Customer>()
+			                        .ToList();
```
fredericDelaporte, line 83: Better batch together more than one query to actually test the batching feature.

fredericDelaporte (commented):
I am going to adjust these additional points.

micmerchant:
> I am going to adjust these additional points.
:thumbsup:

I can recommend this NuGet package for more verbose assertions [FluentAssertions](https://github.com/fluentassertions/fluentassertions).

fredericDelaporte (approved):

micmerchant:
Just for information, you can find the plugin to support MSSQL Table-Valued Functions with Linq here:
[MSSQLTableValuedFunctions.NHibernate.Linq](https://github.com/micmerchant/nhibernate-mssql-table-valued-functions)

PR #2459, opened 2020-07-28
Fix detecting parameter type for Contains method for Linq provider

maca88:
This is a regression from #2365, where `Contains` method which is transformed into `ContainsResultOperator` on pre-transform was not handled correctly by `QueryModel.TransformExpressions`. `ContainsResultOperator` is a special result operator that unfortunately has to be manually handled.

Fixes: #2453

PR #3212, opened 2022-12-19
Fix some Contains subqueries failures with Linq

gliljas:
This is most likely an old regression, where the smartness around LINQ Contains queries ignores the fact that source to the Contains can contain body clauses. In my case it was the use of `MappedAs` that was left hanging in the querymodel, even though `removeMappedAsCalls` was set to true, but I guess there may be other issues.


fredericDelaporte:
It was not clear for me if this was an improvement or a bug fix. Checking the test case, the query throws without the change. So, labeling as bug.

gliljas:
Yes, it's a bug.

fredericDelaporte:
Though not a recent regression, should we target 5.4.1 instead of master?

fredericDelaporte (approved):

src/NHibernate/Linq/Visitors/ParameterTypeLocator.cs
```diff
@@ -302,6 +302,7 @@ private bool TryLinkContainsMethod(QueryModel queryModel)
 				// ContainsResultOperator where the constant expression is dislocated from the related expression,
 				// we have to manually link the related expressions.
 				if (queryModel.ResultOperators.Count != 1 ||
+					queryModel.BodyClauses.Count > 0 ||
```
bahusoid, line 305: We also have different issue regarding `Contains` on parameter constants. See #2872. In 5.3 we didn't handle it properly but exception is thrown.  
I wonder if this change makes `Where` part be silently ignored.

gliljas: I must admit I don't quite understand why TryLinkContainsMethod is needed.

bahusoid: It's required for proper parameter type recognition(See #2459). So it's required for queries like `Where(order => collectionParameter.Contains(order.Id))` or `Where(order => order.Lines.Contains(parameter))`
So in theory your changes might break second type of queries (Contains on subquery) with Body clauses. But removing completely this check I see fails only for first type of queries (Contains on collection parameter). So it's either not properly covered in tests or this change not needed any more...

bahusoid: > So in theory your changes might break second type of queries

So maybe to be on a safe side we should keep logic adding related expressions and instead replace final "return true" that skips further processing with something like:
```C#
//Subqueries not based on parameters need further processing
return right.NodeType != ExpressionType.Constant;
```

bahusoid: My last suggestion is actually wrong. And it should be actually pretty safe as is... So marking it as resolved.

bahusoid:
Replaced by #3274

PR #2502, opened 2020-08-30
Support count distinct on function in hql

bahusoid:
Fixes #1285

PR #3217, opened 2023-01-02
Fix Count Distinct on path with hql reserved words

### LINKED ISSUE #3215: Count(Distinct ...)  does not work
below HQL does not work with NH 5.4 but works with NH 5.3.14 :
select l.Name, Count(Distinct l.Member.ID) as MemCount from Loan l group by l.Name

bahusoid:
Regression from #2502
Fixes #3215


src/NHibernate/Hql/Ast/ANTLR/Hql.g
```diff
@@ -600,7 +600,7 @@ vectorExpr
 // NOTE: handleDotIdent() is called immediately after the first IDENT is recognized because
 // the method looks a head to find keywords after DOT and turns them into identifiers.
 identPrimary
-	: identifier { HandleDotIdent(); }
+	: identifier {{ HandleDotIdent(); }}
```
bahusoid, line 603: `HandleDotIdent` should be executed even while backtracking

fredericDelaporte (approved):

PR #3222, opened 2023-01-16
Automatically generate async code on pull requests for 5.3

bahusoid:
My test run: https://github.com/bahusoid/nhibernate-core/actions/runs/3932985711

### DIFF
--- a/Tools/.gitignore
+++ b/Tools/.gitignore
@@ -4,4 +4,4 @@ csharpasyncgenerator.commandline/
 gitreleasemanager/
 obj/
 microsoft.*
-netstandard.library/
+netstandard.*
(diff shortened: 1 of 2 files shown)


Tools/.gitignore
```diff
@@ -4,3 +4,4 @@ csharpasyncgenerator.commandline/
 gitreleasemanager/
 obj/
 microsoft.*
+netstandard.library/
```
fredericDelaporte, line 7: In master we have instead:

```suggestion
netstandard.*
```

Wouldn't it better to align on it, or do I miss something?

bahusoid: Yeah Ok. I didn't check master because it wasn't part of this change. And AsyncGenerator is used as global tool there it's most likely no longer needed

fredericDelaporte (approved):

PR #3233, opened 2023-02-08
Make object destructors not proxiable

### LINKED ISSUE #3205: `Object.Finalize` should not be proxiable
Currently in DefaultDynamicProxyMethodCheckerExtensions methods `IsProxiable` and `ShouldBeProxiable` are treating overridden  `Finalize` methods as proxiable:
```
...
&& (method.DeclaringType != typeof(object) || !"finalize".Equals(method.Name, StringComparison.OrdinalIgnoreCase))
...
```
Both of these methods are used for lazy loading.
If during finalization object is not yet loaded, lazy proxy attempts to pull it from DB.
Since Finalize is called by GC at random times and random threads, it is unlikely that session will still be open during Finalize.
In most cases Finalize for lazy not yet loaded proxies throws an exception "Could not initialize proxy - no Session".

This bug occurs only if proxyable class has a base class, that is overriding `Finalize`, ie:
```
class Base { ~Base() { } }
class LazyChild : Base { } // this class will not be finalized
```
I've created a small demonstration project: https://github.com/ssharunas/nhibernate-finalization-bug-demonstration

IMHO Finalize should **not** be proxyable even in delivered classes.

hazzik:
Fixes #3205


bahusoid:
Can you clarify current behavior. It seems desctructor for proxy should be generated but be empty. Is it what happens in this PR?

bahusoid:
>  It seems desctructor for proxy should be generated but be empty

Nope. I'm wrong. It should be generated but base class be called only when proxy is initialized.

src/NHibernate/Proxy/DefaultDynamicProxyMethodCheckerExtensions.cs
```diff
@@ -47,5 +45,11 @@ private static bool IsDisposeMethod(MethodInfo method)
 			return method.Name.Equals("Dispose") && method.MemberType == MemberTypes.Method && method.GetParameters().Length == 0;
 			// return method.Name.Equals("Dispose") && method.IsMethodOf(typeof(IDisposable));
 		}
+
+		private static bool IsFinalizeMethod(MethodInfo method)
+		{
+			return "finalize".Equals(method.Name, StringComparison.OrdinalIgnoreCase) &&
```
bahusoid, line 51: Why this method name check? Seems not needed.

hazzik: It was in the original code, so I left it here.

hazzik:
> it should be generated but base class be called only when proxy is initialized.

The problem is that finalizers are called in some random thread. So even if session is still open and proxy is initialized we could get some problems because NHibernate is not thread safe in general.

src/NHibernate/Proxy/DefaultDynamicProxyMethodCheckerExtensions.cs
bahusoid: But it's not original check anymore so let's remove it as obsolete.

bahusoid:
> it should be generated but base class be called only when proxy is initialized.

When suggesting it I've been thinking of standard destructor implementation. It looks something like:

```C#
if(!disposed)
   Dispose();
```

So you made destructor not proxiable but `IDisposable.Dispose` is proxiable. So it fixes nothing in this case. But then I'm not that sure that it's safe to skip base destructor completely for  un-initialized proxy.

hazzik:
Dispose is already not proxiable, or am I missing something?

hazzik:
@bahusoid do you think we should implement empty finalizers on the proxy if base class has a finalizer? Or what shall we do?

bahusoid:
> do you think we should implement empty finalizers 

No. I think your fix is correct. I just don't understand why you want to keep 2 checks:
```C#
return "finalize".Equals(method.Name, StringComparison.OrdinalIgnoreCase) &&
			       method.GetBaseDefinition() == ReflectionCache.ObjectMethods.Finalize;
```

when it's enough to have single check:
```C#
return method.GetBaseDefinition() == ReflectionCache.ObjectMethods.Finalize;
```

hazzik:
> I just don't understand why you want to keep 2 checks

I just had no time to update the PR, that all.

bahusoid (approved):

PR #3236, opened 2023-02-10
Add MySQL8Dialect and MySQL8InnoDBDialect

dennis-gr:
MySQL 8 is deprecating integer display widths, see [here](https://dev.mysql.com/worklog/task/?id=13127) and [here](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html) for more information.

The MySQL8Dialect uses BOOLEAN instead of TINYINT(1) for DbType.Boolean.


src/NHibernate/Dialect/MySQL8Dialect.cs
```diff
@@ -0,0 +1,9 @@
+using System.Data;
+
+namespace NHibernate.Dialect
+{
+	public class MySQL8Dialect : MySQL57Dialect
+	{
+		public MySQL8Dialect() => RegisterColumnType(DbType.Boolean, "BOOLEAN");
```
fredericDelaporte, line 7: This `BOOLEAN` type does not actually exist for MySql.

See [11.9 Using Data Types from Other Database Engines](https://dev.mysql.com/doc/refman/8.0/en/other-vendor-data-types.html), MySql goes on using `TINYINT` for this.

See also [Numeric Data Type Syntax](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html)
> `BOOL`, `BOOLEAN`
>
>These types are synonyms for [`TINYINT(1)`](https://dev.mysql.com/doc/refman/8.0/en/integer-types.html).

So, that is mere synonyms. And if that documentation is accurate, switching to them does not cease using the deprecated feature, display width.

fredericDelaporte: So, it should be:
```suggestion
		public MySQL8Dialect() => RegisterColumnType(DbType.Boolean, "TINYINT");
```

fredericDelaporte (commented):
Thanks for your contribution. It seems to me it misses the point you were willing to address, display width being deprecated for integers types.

dennis-gr:
No, it actually needs to be BOOLEAN since TINYINT(1) is treated differently by [connectors](https://mysqlconnector.net/tutorials/best-practices/). Because of this, TINYINT(1) is probably the only case where specifiying the display width is/was actually useful.

You can verify this yourself:
```
mysql> create table a (b TINYINT(1));
Query OK, 0 rows affected, 1 warning (0.00 sec)

mysql> show warnings;
+---------+------+------------------------------------------------------------------------------+
| Level   | Code | Message                                                                      |
+---------+------+------------------------------------------------------------------------------+
| Warning | 1681 | Integer display width is deprecated and will be removed in a future release. |
+---------+------+------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> create table b (c BOOLEAN);
Query OK, 0 rows affected (0.00 sec)

mysql> show warnings;
Empty set (0.00 sec)
```

fredericDelaporte:
According to your link, the connector documentation tell to use `BOOL`. It do not seem to tell anything about `BOOLEAN`. So, should the mapping use `BOOL` instead, just in case?

dennis-gr:
Since the official [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html) specifies that both BOOL and BOOLEAN are valid, I think it's fine to leave it as is.

fredericDelaporte (approved):

PR #3242, opened 2023-02-22
Linq: add enum Equals and object Equals support

tykovec:


tykovec:
Seems not so easy with enum equals as it is boxed in expression to object.Equals(object) (what leads to problem in my first iteration with BaseHqlGeneratorForMethod approach)

bahusoid:
We have `EqualsGenerator` class with registrations for various Equals methods. Can't it be simply extended with additional registration for Enum.Equals?

tykovec:
> We have `EqualsGenerator` class with registrations for various Equals methods. Can't it be simply extended with additional registration for Enum.Equals?

I tried it but it is throwing exception:
```
System.InvalidOperationException : The binary operator Equal is not defined for the types 'NHibernate.DomainModel.Northwind.Entities.Gender' and 'System.Object'.

Stack Trace: 
Expression.GetEqualityComparisonOperator(ExpressionType binaryType, String opName, Expression left, Expression right, Boolean liftToNull)
Expression.Equal(Expression left, Expression right, Boolean liftToNull, MethodInfo method)
Expression.Equal(Expression left, Expression right)
EqualsGenerator.BuildHql(MethodInfo method, Expression targetObject, ReadOnlyCollection arguments, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor) line 79
HqlGeneratorExpressionVisitor.VisitMethodCallExpression(MethodCallExpression expression) line 605
```

GetMethodBasedBinaryOperator is expecting that types on both sides of equal operation are same ...

bahusoid:
Stacktrace shows the same source of exception:
> EqualsGenerator.BuildHql

 I can't to look into it right now. But I'm pretty sure `EqualsGenerator.BuildHql` can be adjusted (like casting to object one of argument) to make it work

tykovec:
I tried it but Enum.Equals is  `[MethodImplAttribute(MethodImplOptions.InternalCall)]` and when i try to get method `ReflectHelper.GetMethodDefinition<Enum>(x => x.Equals(default(Enum)))` this is evaluated to `object.Equals` and then is `EqualsGenerator` class used for object.Equals comparison which is not wanted.

bahusoid:
> then is EqualsGenerator class used for object.Equals comparison which is not wanted.

Why? Does it cause any particular issues (except mentioned problem from `EqualsGenerator.BuildHql`)? Maybe we just need to register `object.Equals`.

hazzik:
hi @tykovec could you please remove unrelated formatting changes?

tykovec:
> hi @tykovec could you please remove unrelated formatting changes?

done, btw were you not thinking about some code cleanup rules or something like that, white spaces and spacing is really mess ...

tykovec:
@hazzik how can i retrigger teamcity build, has some issue with cleanup ....

fredericDelaporte:
I have relaunched them.

hazzik:
@fredericDelaporte I actually had to restart the build agent, sorry, did not have time to re-queue tests

tykovec:
@bahusoid @hazzik -> i have repushed changes, now there is no problem with team city builds, needs successful merge something more?

src/NHibernate.Test/NHSpecificTest/GH1879/ExpansionRegressionTests.cs
```diff
@@ -30,35 +21,6 @@ protected override void OnSetUp()
 			}
 		}
 
-		protected override void Configure(Configuration configuration)
```
hazzik, line 33: This, unfortunately, makes the test useless.

bahusoid: It actually looks wrong that user can't use own generator for already registered method.
I think we should simply change code to allow user overrides (basically change `dictionary.Add(method, handler)` with `dictionary[method] = handler` in affected places)

tykovec: i agree, some method handlers could be already defined by some developers (like in our case :-)) what could leads to exceptions, should i rewrite the code and revert test

bahusoid: > should i rewrite the code and revert test

I would say yes. Please do.

hazzik (approved):

PR #3259, opened 2023-03-16
Treat transparentcast as guessed type in hql

### LINKED ISSUE #3256: Invalid SQL is generated for string Enum used in conditional LINQ
It's an old issue we discussed in https://github.com/nhibernate/nhibernate-core/pull/2499#issuecomment-682420425 and related to #707 

Test cases for `WhereTests.cs`:
```C#
//Fails for all dialects 
[Test]
public void CanUseStringEnumInConditional()
{
	using var log = new SqlLogSpy();
	var query = db.Users
			            .Where(user => (user.Enum1 == EnumStoredAsString.Small
				                   ? EnumStoredAsString.Small
				                   : EnumStoredAsString.Large) == user.Enum1 )
			.Select(x => x.Enum1);

	Assert.That(query.Count(), Is.GreaterThan(0));
}

//Fails for SQLite 
[Test]
public void CanUseStringEnumInConditional2()
{
	using var log = new SqlLogSpy();
	var query = db.Users
			            .Where(user => (user.Enum1 == EnumStoredAsString.Small
				                   ? user.Enum1
				                   : EnumStoredAsString.Large) == user.Enum1 )
			.Select(x => x.Enum1);

	Assert.That(query.Count(), Is.GreaterThan(0));
}
```

Example of invalid SQL for `CanUseStringEnumInConditional2` test on SQLite:
```sql
    select
        cast(count(user0_.Enum1) as INTEGER) as col_0_0_ 
    from
        Users user0_ 
    where
        cast(case 
            when user0_.Enum1=@p0 then user0_.Enum1 
            else @p1 
        end as INTEGER)=user0_.Enum1; -- cast to INTEGER is invalid as user0_.Enum1 is a string column
```

The problem goes away by removing the following transparent cast and returning case directly :
https://github.com/nhibernate/nhibernate-core/blob/8e22b702102970eef1d4ae00b2585278feae1db0/src/NHibernate/Linq/Visitors/HqlGeneratorExpressionVisitor.cs#L633

It's not a suggested fix - just a place to look at.

bahusoid:
Fixes #3256

Related to #707
The whole `transparentcast` concept is a bit hacky and breaks hql parameter detection logic in case some custom type is used (enum string is one example). So let's extend this hack a bit more to let hql know that `transparentcast` is just a guessed type hint.

### DIFF
--- a/src/NHibernate/Hql/Ast/ANTLR/Tree/TransparentCastNode.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/Tree/TransparentCastNode.cs
@@ -28,8 +28,8 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 			{
 				_expectedType = value;
 				var node = GetChild(0).NextSibling.GetChild(0);
-				//transparentcast on parameters is a special use case - skip it.
-				if (node.Type != HqlSqlWalker.NAMED_PARAM && node is IExpectedTypeAwareNode  typeNode && typeNode.ExpectedType == null)
+				// A transparent cast on parameters is a special use case - skip it.
+				if (node.Type != HqlSqlWalker.NAMED_PARAM && node is IExpectedTypeAwareNode typeNode && typeNode.ExpectedType == null)
 					typeNode.ExpectedType = value;
 			}
 		}
@@ -37,7 +37,7 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 		public override SqlString Render(IList args)
 		{
 			return ExpectedType != null
-				// Provide expected type in case transparent cast is transformed to real
+				// Provide the expected type in case the transparent cast is transformed to an actual cast.
 				? ((CastFunction) SQLFunction).Render(args, ExpectedType, SessionFactoryHelper.Factory)
 				: base.Render(args);
 		}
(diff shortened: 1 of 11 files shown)


src/NHibernate/Hql/Ast/ANTLR/Tree/TransparentCastNode.cs
```diff
@@ -0,0 +1,45 @@
...
+
+		public IType ExpectedType
+		{
+			get => _expectedType;
+			set
+			{
+				_expectedType = value;
+				var node = GetChild(0).NextSibling.GetChild(0);
+				//transparentcast on parameters is a special use case - skip it.
+				if (node.Type != HqlSqlWalker.NAMED_PARAM && node is IExpectedTypeAwareNode  typeNode && typeNode.ExpectedType == null)
```
fredericDelaporte, line 32:
```suggestion
				// A transparent cast on parameters is a special use case - skip it.
				if (node.Type != HqlSqlWalker.NAMED_PARAM && node is IExpectedTypeAwareNode typeNode && typeNode.ExpectedType == null)
```

src/NHibernate/Hql/Ast/ANTLR/Tree/TransparentCastNode.cs
```diff
@@ -0,0 +1,45 @@
...
+					typeNode.ExpectedType = value;
+			}
+		}
+
+		public override SqlString Render(IList args)
+		{
+			return ExpectedType != null
+				// Provide expected type in case transparent cast is transformed to real
```
fredericDelaporte, line 40:
```suggestion
				// Provide the expected type in case the transparent cast is transformed to an actual cast.
```

fredericDelaporte (approved):

PR #3304, opened 2023-05-13
Fix SetSnapShot CopyTo variance failure

fredericDelaporte:
This failure causes tests to fail when upgrading the NUnit engine.

### DIFF
--- a/src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
+++ b/src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
@@ -120,16 +120,10 @@ namespace NHibernate.Collection.Generic.SetHelpers
 				return;
 			}
 
-			if (array.GetType().GetElementType().IsAssignableFrom(typeof(T)))
-			{
-				if (_hasNull)
-					array.SetValue(default(T), index);
-				ICollection keysCollection = _values.Keys;
-				keysCollection.CopyTo(array, index + (_hasNull ? 1 : 0));
-				return;
-			}
-
-			throw new ArgumentException($"Array must be of type {typeof(T[])}.", nameof(array));
+			if (_hasNull)
+				array.SetValue(default(T), index);
+			ICollection keysCollection = _values.Keys;
+			keysCollection.CopyTo(array, index + (_hasNull ? 1 : 0));
 		}
 
 		public int Count => _values.Count + (_hasNull ? 1 : 0);
--- a/src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
+++ b/src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
@@ -114,12 +114,22 @@ namespace NHibernate.Collection.Generic.SetHelpers
 
 		void ICollection.CopyTo(Array array, int index)
 		{
-			if (!(array is T[] typedArray))
+			if (array is T[] typedArray)
 			{
-				throw new ArgumentException($"Array must be of type {typeof(T[])}.", nameof(array));
+				CopyTo(typedArray, index);
+				return;
 			}
 
-			CopyTo(typedArray, index);
+			if (array.GetType().GetElementType().IsAssignableFrom(typeof(T)))
+			{
+				if (_hasNull)
+					array.SetValue(default(T), index);
+				ICollection keysCollection = _values.Keys;
+				keysCollection.CopyTo(array, index + (_hasNull ? 1 : 0));
+				return;
+			}
+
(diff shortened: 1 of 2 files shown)


src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -114,12 +114,22 @@ IEnumerator IEnumerable.GetEnumerator()
...
-			if (!(array is T[] typedArray))
+			if (array is T[] typedArray)
 			{
-				throw new ArgumentException($"Array must be of type {typeof(T[])}.", nameof(array));
+				CopyTo(typedArray, index);
+				return;
 			}
 
-			CopyTo(typedArray, index);
+			if (array.GetType().GetElementType().IsAssignableFrom(typeof(T)))
```
hazzik, line 123: Do we need this check? Wouldn't it fail regardless? Do we need to test this behaviour?

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -114,12 +114,22 @@ IEnumerator IEnumerable.GetEnumerator()
 
 		void ICollection.CopyTo(Array array, int index)
```
hazzik, line 115: You've forgot about overload for .NET Core https://github.com/nhibernate/nhibernate-core/blob/2ee35131208be016cc038fafc62b6c55d69d59dc/src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs#L164-L172

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
bahusoid: I don't think we actually need even first check. Implementation can simply be equal to generic CopyTo

bahusoid: Ok. It can't be fully equal. What I meant:

```C#
void ICollection.CopyTo(Array array, int index)
{
	if (_hasNull)
		array.SetValue(default(T), index);
	((ICollection) _values.Keys).CopyTo(array, index + (_hasNull ? 1 : 0));
}
```

bahusoid:
`ICollection.CopyTo` implementation is broken by https://github.com/nhibernate/nhibernate-core/pull/2394
Shouldn't it be treated as 5.3 regression?

fredericDelaporte:
While technically a regression from 5.3.0, SetSnapShot is an internal technical class which NHibernate current internal uses do not suffer from the introduced bug. So, it appears there is no regression from an user standpoint. Only the NHibernate.Test project has an issue with this "regression", and only if upgrading our testing framework, which works on that internal class.

That is why targeting master seems enough to me. But why not targeting 5.3.x.

About:
> I've added test for the incompatible array type. .NET 4.x and .NET Core behave differently. The .NET 4.x throws ArgumentException while .NET Core throws InvalidCastException. Not sure if we need/want to make this consistent.

I do not see a need for that, due to this class being internal and having its current use cases without a possibility of triggering these exceptions.

hazzik:
> I do not see a need for that

Sure. Feel free to drop my test then.

bahusoid:
>  is an internal technical class which NHibernate current internal uses do not suffer from the introduced bug

Yeah but it's still exposed to user via `IPersistentCollection.GetSnapshot` call.  I think it's worth fixing at least in 5.4.

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -153,12 +157,26 @@ protected SetSnapShot(SerializationInfo info, StreamingContext context) : base(i
...
+			if (index < 0)
+				throw new ArgumentOutOfRangeException(nameof(index), index, "Array index cannot be negative");
+
+			if (index > array.Length || Count > array.Length - index)
+				throw new ArgumentException("Provided array is too small", nameof(array));
+
+			foreach (var value in this)
+			{
+				array.SetValue(value, index);
+				index++;
```
bahusoid, line 178:
```suggestion
				array.SetValue(value, index++);
```

hazzik:
Closed and opened PR to fix wrong target on `Generate Async code`

src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
fredericDelaporte: I avoid combining in one line of code two unrelated operations. That post-increment feature which allows to use a value then increment it altogether looks as an anti-pattern to me.

hazzik (approved):

PR #3309, opened 2023-05-27
Test Case for Invalid SQL with property-ref on property using a formula

PapisKang:
The test case for #1313  is not complete yet cause i ve some troubles with the issue description. 
The description of the issue defines a situation where a many to one relationship with property-ref is targetting a property having a formula. 
But after spending a lot of time on the documentation, i feel like it s something that is not allowed anymore. 
A formula is not a  column in the database (it is just calculated), so i m asking how can i target a formula with a many to one relationship ? Also  specifying both column and formula attributes is invalid 
`<property name="SerialNumber" unique="true" type="string" column="SerialNumber" formula="(select Id from Customer where Customer.Id = Id)" />`

AnyWay, i tried to write a test case and i hope someone would help to understand what we re trying to do here.


src/NHibernate.Test/NHSpecificTest/GH3513/Fixture.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using System.Linq;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.GH3513
```
fredericDelaporte, line 5: It should be moved to `NHibernate.Test.NHSpecificTest.GH1313`. #1313 was a bug reported on our old Jira as NH-3513, which was using a NH prefix. We keep NH prefix for them, to avoid conflict with GitHub issues number, which we prefix with GH. So it could be `NHibernate.Test.NHSpecificTest.NH3513` too.

PapisKang: Oh okay. I am going to fix that

src/NHibernate.Test/NHSpecificTest/GH1313/Fixture.cs
```diff
@@ -0,0 +1,49 @@
...
+		{
+			using var session = OpenSession();
+			using var transaction = session.BeginTransaction();
+
+			session.CreateQuery("delete from System.Object").ExecuteUpdate();
+			transaction.Commit();
+		}
+
+		[Test]
+		[Explicit("Not fixed yet")]
```
fredericDelaporte, line 37: Since this is not fixed yet, marking as explicit allows the test case to be ignore by the CI. This will have to be removed once the trouble is fixed.

fredericDelaporte (approved):

PR #3310, opened 2023-06-03
Move persist tracking context out of the loop in AbstractFlushingEventListener

### LINKED ISSUE #3307: IsDirty performance hit from 5.3.16 to 5.4.2
I faced a longer IsDirty run after upgrading.

NHibernate 3.3.1:

image

NHibernate 5.4.2:

image

The first question is, was there a bug in old implementation that now lead to this cascade?
Since the result of IsDirty is true in both cases, I can imagine that there is optimization potential.

The test case of IsDirty is there that we have a table which stores for every element a parent element (not lazy load). The children are as HasMany relation. We remove just near all elements, and this lead to this bad result.

hazzik:
Fixes #3307

### DIFF
--- a/src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
+++ b/src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
@@ -188,6 +188,7 @@ namespace NHibernate.Event.Default
 		{
 			log.Debug("processing flush-time cascades");
 
+			var anything = Anything;
 			ICollection list = IdentityMap.ConcurrentEntries(session.PersistenceContext.EntityEntries);
 			//safe from concurrent modification because of how entryList() is implemented on IdentityMap
 			foreach (DictionaryEntry me in list)
@@ -196,7 +197,7 @@ namespace NHibernate.Event.Default
 				Status status = entry.Status;
 				if (status == Status.Loaded || status == Status.Saving || status == Status.ReadOnly)
 				{
-					CascadeOnFlush(session, entry.Persister, me.Key, Anything);
+					CascadeOnFlush(session, entry.Persister, me.Key, anything);
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
```diff
@@ -196,7 +197,7 @@ protected virtual void PrepareEntityFlushes(IEventSource session)
 				Status status = entry.Status;
 				if (status == Status.Loaded || status == Status.Saving || status == Status.ReadOnly)
 				{
-					CascadeOnFlush(session, entry.Persister, me.Key, Anything);
+					CascadeOnFlush(session, entry.Persister, me.Key, anything);
```
fredericDelaporte, line 200: I do not get it. The property implementation just always directly yield null, and is not overridden anywhere to do anything else. What does this change?

bahusoid: > not overridden anywhere

Not until #2752. Looks like you check on 5.3 codebase.

hazzik: Yes, as Roman said: it was overridden in #2752:

https://github.com/nhibernate/nhibernate-core/blob/170210aa211a522f145c61437d4429af794f1f7f/src/NHibernate/Event/Default/DefaultDirtyCheckEventListener.cs#L16

See the discussion starting in https://github.com/nhibernate/nhibernate-core/issues/3307#issuecomment-1562296466 for more context.

fredericDelaporte: Indeed, I was not on the adequate code base.

fredericDelaporte (approved):

PR #3312, opened 2023-06-10
Fix referencing nullable entity in correlated subquery

### LINKED ISSUE #3306: NHibernate can produce invalid SQL (missing comma) for a query: Suspected regression 5.3.10 -> 5.3.11
We have recently tried upgrading NHibernate for our private project and found that the upgrade caused crashes against some particular queries that had previously returned the expected results.

I have narrowed this problem down to having been introduced in NHibernate **5.3.11** (last working version was 5.3.10).  I have also created [a standalone reproduction project, which I have placed online].  I have tried to keep the example as minimal as possible, to match the query & techniques that we are actually using.  Some of the query/mappings/entities might prove to be irrelevant though.

Full repro steps and the analysis I have already performed are available in the **Readme** to that repo.  You'll also find a sample schema creation script in there.  It also includes copies of the SQL produced by 5.3.10 and 5.3.11 and an indication of where they differ.  In short, in a multi-table subquery which NHibernate creates, it omits a comma between two tables in the FROM part of the query.  This then causes a crash error because the DB rejects it.

```sql
FROM tableOne tableOneAlias tableTwo tableTwoAlias
-- ----------------------- ^ This is where the comma is missing
```

There is another difference in the SQL though, which might prove to be another problem entirely _(I'm happy to open a new ticket if someone determines that it is indeed unrelated)_.  The criterion which uses the formula for a `many-to-one` association (which is mapped using a formula) is omitted from the WHERE from 5.3.11 onward.  I think that even if the invalid SQL (missing comma) issue were fixed, the query would still return incorrect results because of the omitted criterion.

I have reproduced this with T-SQL (MS SQL Server) but _I have not tried it with other database drivers/dialects_ such as SQLite.  It may or may not affect those too.  I have also reproduced it with the current latest NH version (at time of writing) 5.4.2.

[a standalone reproduction project, which I have placed online]: https://github.com/craigfowler/NHibernateInvalidSql

bahusoid:
Fixes #3306 

regression from https://github.com/nhibernate/nhibernate-core/pull/2989.


fredericDelaporte (approved):

fredericDelaporte:
My adaptation of the opener test case succeeds with your branch.

PR #3320, opened 2023-06-16
Fix fetching lazy component on not mapped interface

### LINKED ISSUE #3289: Fetching lazy loaded component causes n + 1 query when querying a subclass abstraction
I've found that when doing a LINQ query and fetching a component it seems to work fine when querying a class, but if I try to query the interface the class implements then the fetch doesn't appear to work.

For example say I have the following:

```cs
public class Entity : IEntity
{
    public virtual int Id { get; set; 
    public virtual Component Component { get; set; }
}

public class SubEntity : Entity, ISubEntity
{
    public virtual bool SomeProperty { get; set; }
}

public class Component
{
    public virtual string Field { get; set; }
}

public interface IEntity
{
    int Id { get; set; }
    string Name { get; set; }
    Component Component { get; set; }
}

public interface ISubEntity : IEntity
{
    public bool SomeProperty { get; set; }
}
```

Which is mapped like so:

```cs
protected override HbmMapping GetMappings()
{
    var mapper = new ModelMapper();
    mapper.Class<Entity>(rc =>
    {
        rc.Id(x => x.Id, m => m.Generator(Generators.Identity));
        rc.Property(x => x.Name);
        rc.Component(x => x.Component);
    });
    mapper.JoinedSubclass<SubEntity>(rc =>
    {
        rc.EntityName(typeof(ISubEntity).FullName);
        rc.Key(k => k.Column("Id"));
        rc.Property(x => x.SomeProperty);
    });
    mapper.Component<Component>(rc =>
    {
        rc.Property(x => x.Field);
        rc.Lazy(true);
    });

    return mapper.CompileMappingForAllExplicitlyAddedEntities();
}
```

Now if I was to say the following, the database would be hit once:

```cs
var result = session.Query<SubEntity>()
    .Fetch(e => e.Component)
    .ToList();

// Make sure the component field is executed.
foreach (var entity in result)
{
    var foo = entity.Component?.Field;
}
```

However if I say the following, the database would be hit for each entity in the result:

```cs
var result = session.Query<ISubEntity>()
    .Fetch(e => e.Component)
    .ToList();

// Make sure the component field is executed.
foreach (var entity in result)
{
    var foo = entity.Component?.Field;
}
```

**Edit:** See below for a test case.

bahusoid:
Fixes #3289

not a regression but let's fix in 5.4


src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
```diff
@@ -52,6 +52,16 @@ public void Process(FetchRequestBase resultOperator, QueryModelVisitor queryMode
 				{
 					var metadata = queryModelVisitor.VisitorParameters.SessionFactory
 													.GetClassMetadata(resultOperator.RelationMember.ReflectedType);
+					if (metadata == null)
+					{
+						var entityName = queryModelVisitor.VisitorParameters.SessionFactory.GetImplementors(
+							resultOperator.RelationMember.ReflectedType.FullName).SingleOrDefault();
```
fredericDelaporte, line 58: Is there a risk to have many implementors, so a crash here, in case of inheritance mapping?

bahusoid: For some reasons I didn't think it's a valid scenario. Fixed.

hazzik:
We usually traverse all possible implementations in other places, not just
select first.

bahusoid:
> We usually traverse all possible implementations in other places, not just
select first.

Nope.  And why we need to do it here? Other place with explanation:
https://github.com/nhibernate/nhibernate-core/blob/90d50c03475628d10336cbad368f071375056fef/src/NHibernate/Util/ExpressionsHelper.cs#L498-L503

Ideally we should use `ExpressionsHelper.TryGet` methods but it's too early for it...

hazzik:
Ok, thanks for the clarification.

Another question, why do we fix it in 5.4.x? Was it a regression?

bahusoid:
> Was it a regression?

No (it's mentioned in pr description). But fix is simple so why not?

fredericDelaporte (approved):
There is always a risk of introducing some new regression with each change. Ideally 5.4.x releases are meant to stabilize (fix regressions) the 5.4 minor version. Additional fixes added into it may delay its stabilization. And also, adding new fixes in it instead of vNext (5.5) is not an incentive to finish 5.5.

That said, sure, the fix seems to come with very little risks of unexpected consequences, and we have already accepted other little additional fixes in the past.

hazzik (approved):

PR #3326, opened 2023-06-19
Fixes cascading orphan delete on versioned entity

### LINKED ISSUE #3325: Cascading orphan delete may not work on versioned entity
In a certain scenario, the orphan delete of a child entity is not performed:

Parent entity must be versioned.

Way to reproduce
* new Parent created
* Child entity is added to Parent
* Parent is saved (cascaded to child)
* Child is removed
* Flush()

Child is then added to the database with NULL as Parent pointer.

Pull request is created.

I would appreciate addition to next 5.4.x release.

csharper2010:
VisitCollectionsBeforeSave must be called no matter if substitute is already true. On versioned entities this is always the case.

Fixes #3325


bahusoid:
Looks like similar hibernate fix https://github.com/hibernate/hibernate-orm/pull/3513

src/NHibernate.Test/NHSpecificTest/GH3325/Mappings.hbm.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<hibernate-mapping xmlns="urn:nhibernate-mapping-2.2" assembly="NHibernate.Test"
+                   namespace="NHibernate.Test.NHSpecificTest.GH3325">
+
+  <class name="Entity">
+    <id name="Id" generator="guid.comb"/>
+    <version name="Version" column="Version" type="Int32" unsaved-value="-1" />
+    <property name="Name"/>
+    <set name="Children" cascade="persist,delete,save-update,evict,lock,replicate,delete-orphan" collection-type="PersistentSetType`1[ChildEntity]">
```
fredericDelaporte, line 9: Why the custom collection type? Is-it required to trigger the trouble?

src/NHibernate/Event/Default/AbstractSaveEventListener.cs
```diff
@@ -235,7 +235,7 @@ protected virtual object PerformSaveOrReplicate(object entity, EntityKey key, IE
 
 			if (persister.HasCollections)
 			{
-				substitute = substitute || VisitCollectionsBeforeSave(entity, id, values, types, source);
+				substitute = substitute | VisitCollectionsBeforeSave(entity, id, values, types, source);
```
fredericDelaporte, line 238: Better explain, to avoid a later bad optimization, even if the test case is also here to prevent the mistake to be done.

```suggestion
				// We have to visit the collections whatever the current value of substitute.
				substitute = VisitCollectionsBeforeSave(entity, id, values, types, source) || substitute;
```

src/NHibernate.Test/NHSpecificTest/GH3325/Mappings.hbm.xml
fredericDelaporte: Bette remove it then, cherry pick the commit on a branch without your fix, and check it triggers the trouble.

csharper2010: I've removed the custom collection from the mapping
Problem is independent of the usage of such a collection

src/NHibernate.Test/Immutable/EntityWithMutableCollection/AbstractEntityWithManyToManyTest.cs
```diff
@@ -485,7 +485,7 @@ public void CreateWithNonEmptyManyToManyCollectionMergeWithNewElement()
 			s.Close();
 	
 			AssertInsertCount(1);
-			AssertUpdateCount(isContractVersioned && isPlanVersioned ? 1 : 0); // NH-specific: Hibernate issues a separate UPDATE for the version number
+			AssertUpdateCount(isContractVersioned && isPlanVersioned ? 2 : 0);
```
hazzik, line 488: https://github.com/hibernate/hibernate-orm/blob/59fdc462542ebe968c8086e5dcfd676f41d16c68/hibernate-core/src/test/java/org/hibernate/orm/test/immutable/entitywithmutablecollection/AbstractEntityWithManyToManyTest.java#L278

src/NHibernate.Test/ReadOnly/ReadOnlyVersionedNodes.cs
```diff
@@ -624,7 +624,7 @@ public void MergeDetachedChildWithNewParentCommitWithReadOnlyChild()
 				t.Commit();
 			}
 
-			AssertUpdateCount(0); // NH-specific: Hibernate issues a separate UPDATE for the version number
+			AssertUpdateCount(1);
```
hazzik, line 627: https://github.com/hibernate/hibernate-orm/blob/59fdc462542ebe968c8086e5dcfd676f41d16c68/hibernate-core/src/test/java/org/hibernate/orm/test/readonly/ReadOnlyVersionedNodesTest.java#L575

src/NHibernate.Test/ReadOnly/ReadOnlyVersionedNodes.cs
```diff
@@ -675,7 +675,7 @@ public void GetChildMakeReadOnlyThenMergeDetachedChildWithNewParent()
...
```
hazzik, line 678: https://github.com/hibernate/hibernate-orm/blob/59fdc462542ebe968c8086e5dcfd676f41d16c68/hibernate-core/src/test/java/org/hibernate/orm/test/readonly/ReadOnlyVersionedNodesTest.java#L619

hazzik (approved):
I've aligned tests to Hibernate and also applied @fredericDelaporte's suggestion

bahusoid (dismissed):
Hibernate fix also uses optimization with additional `isNewlyInstantiated` check
```
default boolean isNewlyInstantiated() {
return getKey() == null && !isDirty();
}
```

It's wrong and breaks case when unwrapped collection is used for modifications. Adjusted test   case to catch this case if someone decides to port this "optimization"

bahusoid (dismissed):

hazzik:
> It's wrong and breaks case when unwrapped collection is used for modifications.

Shall we notify them?

src/NHibernate.Test/NHSpecificTest/GH3325/Fixture.cs
```diff
@@ -0,0 +1,51 @@
...
+		{
+			Guid parentId;
+			Guid childId;
+			using (var session = OpenSession())
+			using (var t = session.BeginTransaction())
+			{
+				var parent = new Entity { Name = "Parent" };
+				var child = new ChildEntity { Name = "Child" };
+				var parentChildren = parent.Children;
+				parentChildren.Add(child);
```
hazzik, line 32: Not sure why there could be differences in behaviour? Could you explain please @bahusoid?

hazzik: Ok, now I see..

However the test is incorrect in itself. I'll adjust the tests.

hazzik (approved):

PR #3328, opened 2023-06-20
Fix AND/OR negation logic in hql

### LINKED ISSUE #3327: HqlParser does not correctly negate EXISTS-nodes below an AND/OR
For the not explicitly handled cases in HqlParser.NegateNode, a new node is created but AND/OR do not correctly replace the children with the newly created ones.

This leads to wrong sql query.

**Example base query (correctly handled):**
```
SELECT COUNT(ROOT.Id)
FROM Entity AS ROOT
WHERE (
    EXISTS (FROM ChildEntity AS CHILD WHERE CHILD.Parent = ROOT)
    AND ROOT.Name = 'Test'
)
```

**Example base query (wrong):**
```
SELECT COUNT(ROOT.Id)
FROM Entity AS ROOT
WHERE NOT (
    EXISTS (FROM ChildEntity AS CHILD WHERE CHILD.Parent = ROOT)
    AND ROOT.Name = 'Test'
)
```

Pull request is created. Problem is obvious and the fix straightforward.
As someone has stated, a similar bugfix was made in Hibernate.

I would appreciate addition to next 5.4.x release.

csharper2010:
AND/OR below NOT must use result of subsequent NegateNode calls to create a new result node instead of partially mutating the existing node to consider functional signature of the NegateNode method

Fixes #3327


bahusoid:
Hibernate similar fix for reference: https://github.com/hibernate/hibernate-orm/commit/aac7e3ee7e91004402f8d9da1edf1b58d67b3314

fredericDelaporte:
For inclusion in 5.4.x as asked in its corresponding issue, #3327, it would need to target the 5.4.x branch and being based on it.

The test case needs to be fixed for some databases too. (I have not checked what is the trouble. It could be that they cannot support what the test does. In such case the test needs likely to do some feature support checks as done in some other tests. It needs investigation.)

Doing this will not guarantee we will merge it in 5.4.x: there is always a risk of introducing some new regression with each change. Ideally 5.4.x releases are meant to stabilize (fix regressions) the 5.4 minor version. Additional fixes added into it may delay its stabilization. And also, adding new fixes in it instead of vNext (5.5) is not an incentive to finish 5.5.

Still we do it sometimes.

csharper2010:
> The test case needs to be fixed for some databases too. (I have not checked what is the trouble. It could be that they cannot support what the test does. In such case the test needs likely to do some feature support checks as done in some other tests. It needs investigation.)

There seems to pe a foreign key problem on setup of the entities or the db structure is not generated as expected. I will have a look next week.

src/NHibernate/Hql/Ast/ANTLR/HqlParser.cs
```diff
@@ -212,17 +212,17 @@ public IASTNode NegateNode(IASTNode node)
 			switch (node.Type)
 			{
 				case OR:
-					node.Type = AND;
-					node.Text = "{and}";
-					NegateNode(node.GetChild(0));
-					NegateNode(node.GetChild(1));
-					return node;
+					IASTNode newNode = (IASTNode) TreeAdaptor.Create(AND, "{and}");
```
bahusoid, line 215: I don't think new node is actually required. We can keep reusing old node. Just need to set negated nodes as children:

```C#
case OR:
	node.Type = AND;
	node.Text = "{and}";
	node.SetChild(0, NegateNode(node.GetChild(0)));
	node.SetChild(1, NegateNode(node.GetChild(1)));
	return node;
```

csharper2010: I didn't check the SetChild-Methode... If it replaces, that variant would also be fine.

bahusoid (dismissed):

fredericDelaporte (approved):

hazzik (approved):

PR #3331, opened 2023-06-21
Lazy property with nosetter.pascalcase-underscore accessor remains uninitialized

sashakboo:
In my project after upgrading from version 5.2 to 5.3 there is a problem with lazy properties that have nosetter.pascal-underscore accessor in their mapping.
- After accessing the property it remains uninitialized
- Each call to a property getter returns a new object

I have added tests which demonstrate the problem.

There is no such problem in version 5.2 #3332


fredericDelaporte:
Thanks for the test case. It has been cherry-picked on a 5.3.x based branch for being included with the fix. See #3345.

PR #3338, opened 2023-06-26
Gh3334 2

csharper2010:
Now I had some time to work on a cleaner solution for #3334 which would make the pull request #3335 obsolete.

The change makes things explicit instead of working around trying to guess the kind of situation we are in.


src/NHibernate/Hql/Ast/ANTLR/HqlSqlWalker.cs
```diff
@@ -835,10 +835,10 @@ void PrepareFilterParameter()
...
-					// Most likely means association join is used in invalid context
-					// I.e. in subquery: from EntityA a where exists (from EntityB join a.Assocation)  
-					// Maybe we should throw exception instead
-					fromElement.FromClause.AddChild(fromElement);
+					// happens e.g. on ToMany association join with "EXISTS(FROM ELEMENTS(ROOT.Children))"
+					// or on an association join in subquery from a parent entity in parent from clause
+					// like in "from EntityA a where exists (from EntityB join a.Assocation)"
+					fromElement.FromClause.AppendFromElement(fromElement);
 					if (fromElement.IsImplied)
 						fromElement.JoinSequence.SetUseThetaStyle(true);
```
bahusoid, line 843:
```suggestion
```
This hack should no longer be needed if it's a proper fix.

bahusoid: Unexpected results. Some tests are failed but only added in this PR.  So this hack is still required for some reasons I don't understand. But anyway it's definitely a better fix for #3334.

csharper2010: Thanks for your comment (and effort)... I also think, it makes some things more explicit but does not fully eliminate the need for hacks.
I was experimenting also with removing the need for the ```if (elem.Parent == null)```-Hack but not yet successful. Quite complex this thing.

bahusoid: I probably OK with keeping it for now. But this PR causes some troubles with MySql on old tests. It needs to be investigated.

fredericDelaporte:
Replaced by #3369.

PR #3345, opened 2023-06-30
Fix lazy property handling with field accessors

### LINKED ISSUE #3333: Lazy property with nosetter.pascalcase-underscore accessor remains uninitialized
In my project after upgrading from version 5.2 to 5.3 there is a problem with lazy properties that have nosetter.pascal-underscore accessor in their mapping.
- After accessing the property it remains uninitialized
- Each call to a property getter returns a new object

I have added tests which demonstrate the problem #3331

There is no such problem in version 5.2 #3332

bahusoid:
Fixes #3333 
Replaces #3331
Regression from https://github.com/nhibernate/nhibernate-core/pull/1943 and https://github.com/nhibernate/nhibernate-core/pull/1947

### DIFF
--- a/src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
+++ b/src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
@@ -12,6 +12,7 @@ using NHibernate.Type;
 using NHibernate.Util;
 using System.Runtime.Serialization;
 using NHibernate.Bytecode.Lightweight;
+using NHibernate.Intercept;
 
 namespace NHibernate.Tuple.Entity
 {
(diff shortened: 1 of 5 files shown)


src/NHibernate/Tuple/Entity/PocoEntityTuplizer.cs
```diff
@@ -306,6 +307,16 @@ public override bool IsLifecycleImplementor
 
 		public override void SetPropertyValue(object entity, int i, object value)
 		{
+			//If there is no property setter we need to manually intercept value for proper lazy property handling
```
fredericDelaporte, line 310: Nit for readability: needs space between comment chars and comment.
```suggestion
			// If there is no property setter we need to manually intercept value for proper lazy property handling.
```

fredericDelaporte (approved):

PR #3350, opened 2023-07-03
BuildCache method returns now a CacheBase type as per TODO

Ivandavidmircea:
Hi, 

I believe that the returned type of the BuildCache method in ICacheProvider should be CacheBase.
It was written as summary long time ago as a TODO but perhaps it was forgotten.

Hope this helps.


bahusoid:
> // 6.0 TODO: return a CacheBase instead

TODO also suggests it's for next major release 6.0. It's a breaking change and can't be added to current 5.x development. Will do it when we decide it's time. Such PR is not really useful. Closing it.

PR #3365, opened 2023-07-17
Call BeforeAssemble on persistent collections' InitializeFromCache to allow batched cache fetching

### LINKED ISSUE #3359: AbstractPersistentCollection InitializeCollectionFromCache performance
I believe there is an issue in second level cache when NHibernate initializing a collection which call GetMany to get all values for array of keys but then when it goes into for loop to Assemble then it again calls GetMany for each item

I have Entity which has a bag collection of EntityFields and both have `batch-size=2000` attribute. When I try to load the collection EntityFields then NHibernate calls `DefaultInitializeCollectionEventListener.InitializeCollectionFromCache` then it gets a 2000 CacheKey which calls `GetMany()` till that point everything is fine but then it goes inside `CacheEntry.Assemble` which calls `InternalLoad` and `DoLoad` which again tries to load from the cache `LoadFromSecondLevelCache` and calls again GetMany ... attached the call stack image.

My understanding is that once it found it all in Second Level Cache then it should not keep calling `GetMany` again for each record one by one, right?

Stack

hazzik:
Fixes #3359

### DIFF
--- a/src/NHibernate.Test/Async/CacheTest/BatchableCacheFixture.cs
+++ b/src/NHibernate.Test/Async/CacheTest/BatchableCacheFixture.cs
@@ -1565,8 +1565,31 @@ namespace NHibernate.Test.CacheTest
 			Assert.That(Sfi.Statistics.QueryCacheHitCount, Is.EqualTo(future ? 2 : 1), "Unexpected cache hit count");
 		}
 
+		[Test]
+		public async Task CollectionLazyInitializationFromCacheIsBatchedAsync()
+		{
+			using (var s = OpenSession())
+			{
+				var readOnly = await (s.GetAsync<ReadOnly>(await (s.Query<ReadOnly>().Select(x => x.Id).FirstAsync())));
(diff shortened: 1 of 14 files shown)


bahusoid:
I probably OK with this fix though I don't think it's fully correct.
Currently it calls `ScheduleBatchLoadIfNeeded(object id, ISessionImplementor session, bool addToQueryCacheBatch)`
with `addToQueryCacheBatch` set to true. But it this context query cache seems irrelevant so it should be `false`

hazzik:
I think it does not really matter. In this case the `QueryCacheQueue` will likely be `null` and so the parameter would be, pretty much, ignored.

fredericDelaporte (approved):

fredericDelaporte:
Though, why 5.4? It does not look like a regression.

hazzik:
> why 5.4? It does not look like a regression.

Yes, it is not regression, but a really nasty bug that renders collection cache unusable in some cases.

fredericDelaporte:
Ok.

PR #3390, opened 2023-08-01
Enable Not node handling in HqlParser.NegateNode

bahusoid:


hazzik (approved):

src/NHibernate.Test/NHSpecificTest/GH3327/Fixture.cs
```diff
@@ -48,5 +48,21 @@ public void NotIsCorrectlyHandled()
...
+			using var log = new SqlLogSpy();
+			using var session = OpenSession();
+			var results = session.CreateQuery(
+				@"SELECT COUNT(ROOT.Id)
+					FROM Entity AS ROOT
+					WHERE NOT (
+						NOT EXISTS (FROM ChildEntity AS CHILD WHERE CHILD.Parent = ROOT)
+						AND NOT ROOT.Name = 'Parent'
+					)").List();
+			Assert.That(log.GetWholeLog(), Does.Not.Contains(" NOT ").IgnoreCase);
```
fredericDelaporte, line 64: This test is about a more complex case than the comment in the parser: `not (not a and not b)` instead of `not (not x)`. That is not a simple double not. Eliminating the not requires to change it to `(a or b)`. Is it what is actually done?

bahusoid: > Eliminating the not requires to change it to (a or b).  Is it what is actually done?

Yes. #3328 was actually about proper handling of and/or negation.

PR #3392, opened 2023-08-02
Partial fix fetching lazy property after Select in Linq

### LINKED ISSUE #3356: Fetch after Select applies to wrong entity
See `FetchAfterSelect` test case added with GH3352 test case.

Query source is wrongly detected and fetch is applied to main entity instead of selected.

bahusoid:
Partial fix of #3356

Added proper from node detection required for lazy property fetching.

To fully fix #3356 we need to find a way to reliable detect query source (see https://github.com/nhibernate/nhibernate-core/pull/3392#discussion_r1281514034)

### DIFF
--- a/src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
+++ b/src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using NHibernate.Hql.Ast;
+using NHibernate.Hql.Ast.ANTLR;
 using NHibernate.Persister.Entity;
 using NHibernate.Type;
 using Remotion.Linq.EagerFetching;
@@ -21,7 +22,7 @@ namespace NHibernate.Linq.Visitors.ResultOperatorProcessors
 
 		public void Process(FetchRequestBase resultOperator, QueryModelVisitor queryModelVisitor, IntermediateHqlTree tree, string sourceAlias)
 		{
-			Process(resultOperator, queryModelVisitor, tree, null, sourceAlias);
+			Process(resultOperator, queryModelVisitor, tree, tree.GetFromNodeByAlias(sourceAlias), sourceAlias);
 		}
 
 		private void Process(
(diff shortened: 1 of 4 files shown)


src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
```diff
@@ -11,12 +12,15 @@ public class ProcessFetch
 	{
 		public void Process(FetchRequestBase resultOperator, QueryModelVisitor queryModelVisitor, IntermediateHqlTree tree)
 		{
-			var querySource = QuerySourceLocator.FindQuerySource(
```
bahusoid, line 14: `QuerySourceLocator.FindQuerySource` can't reliably detect query source. Instead of actually query source it finds first source with "compatible" type

src/NHibernate/Linq/Visitors/ResultOperatorProcessors/ProcessFetch.cs
```diff
@@ -11,12 +12,15 @@ public class ProcessFetch
...
-				queryModelVisitor.Model,
-				resultOperator.RelationMember.DeclaringType);
-			var name = queryModelVisitor.VisitorParameters.QuerySourceNamer.GetName(querySource);
+			var lastFromNode = tree.GetLastFromChildNode();
```
hazzik, line 15: Why have you decided to find the from with HQL? `QuerySourceSelector` (which is currently an inner class of `Joiner`) can reliably find the correct query source.

bahusoid:
> My understanding of the problem currently is that fetch requests are not properly associated to the query source.

No. It's still about proper query source detection. Now it doesn't detected properly for `ReuseJoinScalarSelect` test case.

bahusoid:
Need to find a way to reliable detect query source. I don't quite understand why query source is not part of `FetchBaseRequest`.

bahusoid:
#3356 test casee exposes 2 issues:
1) Invalid query source detection (possible when joins on the same or assignable types are involved)
2) Fetch is added to wrong from node.

This PR fixes second issue (currently it's unclear how to fix first issue)

hazzik (approved):

PR #3396, opened 2023-08-05
Get rid of select queries for each ManyToMany not found ignored element in Criteria and lazy loading

bahusoid:
Use table group join with inner joined entity table for many-to-many not-found="ignore" mapping:
```sql
FROM Device device0_
LEFT OUTER JOIN (
	DriveOfDevice drives1_ 
	INNER JOIN Drive drive2_ ON drives1_.DriveId = drive2_.id
	) ON device0_.id = drives1_.DeviceId
```


src/NHibernate.Test/NHSpecificTest/NH750/ManyToManyThrowsForNotFoundFixture.cs
```diff
@@ -44,6 +45,7 @@ protected override void OnTearDown()
 		[Test]
 		public void LazyLoad()
 		{
+			using var log = new SqlLogSpy();
```
fredericDelaporte, line 48: It is not used by the test. Forgotten debug temporary change?

bahusoid: Removed

fredericDelaporte (approved):

PR #3398, opened 2023-08-06
Do not store mapping field in Configuration

### LINKED ISSUE #3397: GenerateSchemaCreationScript creates many identical dialect instances
Hello
During generation script for DB creation Nhib 5.4.4 creates dialect instances many times.

image

It looses passed dialect instance in GetDialectTypeName calling GetSqlTypeCode

GetSqlTypeCode do not take dialect as param....

It creates slowdown and memory traffic.

hazzik:
Fixes #3397.

(Actually patched with minimal changes in 5.3x with #3399, then fixed more robustly here.)


src/NHibernate/Cfg/Configuration.cs
```diff
@@ -1299,33 +1291,22 @@ protected virtual void ConfigureProxyFactoryFactory()
...
-			try
-			{
-				ConfigureProxyFactoryFactory();
-				SecondPassCompile();
-				Validate();
-				Environment.VerifyProperties(properties);
-				Settings settings = BuildSettings();
+			var m = BuildMapping();
+			var settings = BuildSettings();
+			var mapping = new StaticDialectMappingWrapper(m, settings.Dialect);
```
hazzik, line 1300: `SessionFactoryImpl` will use dialect from session, so use it here as well.

src/NHibernate/Cfg/Configuration.cs
```diff
@@ -124,7 +123,7 @@ public void GetObjectData(SerializationInfo info, StreamingContext context)
 		{
 			ConfigureProxyFactoryFactory();
 			SecondPassCompile();
-			Validate();
+			Validate(BuildMapping());
```
hazzik, line 126: Not sure if we want/need to validate configuration on serialization?

src/NHibernate.Test/DialectTest/MsSql2008DialectFixture.cs
```diff
@@ -148,16 +148,7 @@ public void ScaleTypes()
...
-			typeof(Configuration).GetField("mapping", BindingFlags.Instance | BindingFlags.NonPublic);
-
-		private static IMapping GetMapping(Configuration cfg)
-		{
-			Assert.That(_mappingField, Is.Not.Null, "Unable to find field mapping");
-			var mapping = _mappingField.GetValue(cfg) as IMapping;
-			Assert.That(mapping, Is.Not.Null, "Unable to find mapping object");
-			return mapping;
-		}
+		private static IMapping GetMapping(Configuration cfg) => (IMapping) cfg.BuildSessionFactory();
```
hazzik, line 151: The true equivalent for the original code in the new way would be `GetMapping(Configuration cfg) => cfg.BuildMapping()`, but `Dialect` would throw an `InvalidOperationException`. However, in this case it does not matter and we can use fully resolved session factory instead.

src/NHibernate/Cfg/Configuration.cs
```diff
@@ -238,18 +236,18 @@ public bool HasNonIdentifierPropertyNamedId(string className)
 			}
 
 			public Dialect.Dialect Dialect =>
-				NHibernate.Dialect.Dialect.GetDialect(configuration.Properties);
+				throw new InvalidOperationException("The dialect is not ready at this stage of the configuration.");
```
hazzik, line 239: To prevent further misuses of this property - throw IOE.

src/NHibernate/Cfg/Configuration.cs
bahusoid: Considering that this static wrapping is still required I don't think field removal is actually necessary.

bahusoid:
Looks hacky. I think proper fix should obsolete existing `BuildMapping` method along with existing Mapping class. Introduce new `BuildMapping` method taking dialect as parameter . So wrapping is no longer required.

src/NHibernate/Cfg/Configuration.cs
bahusoid: I would say no

src/NHibernate/Cfg/Configuration.cs
hazzik: Wrap in static wrapper is not required. It is possible to remove it entirely. 

The mapping field is not need as well as it itself is just a wrapper around configuration field.

hazzik: I've decided to remove mapping field to prevent further misuses of the same nature. I've initially started with exact code you have in your PR.

src/NHibernate/Cfg/Configuration.cs
hazzik: I would leave it as is though.

src/NHibernate/Cfg/Configuration.cs
```diff
@@ -275,23 +273,20 @@ public bool HasNonIdentifierPropertyNamedId(string className)
...
 
-		private void InitBlock()
+		public virtual IMapping BuildMapping()
 		{
-			mapping = BuildMapping();
+			return new Mapping(this);
 		}
 
-		public virtual IMapping BuildMapping()
+		private IMapping BuildMapping(Dialect.Dialect dialect)
```
hazzik, line 287: I don't think I want to make this method public.

src/NHibernate/Cfg/Configuration.cs
bahusoid: Why? Wouldn't it throw IOE? At least you shouldn't use obsolete method here.

hazzik: > Wouldn't it throw IOE? 

Highly likely not. 

Ok, I've removed the validation.

src/NHibernate/Cfg/Configuration.cs
bahusoid: Yeah I got it. I'm just saying that with required wrapping it's not any better. You can still create mapping and forgot to wrap it. Runtime exception is not very nice way to report about it.

src/NHibernate/Cfg/Configuration.cs
bahusoid: To me, it looks like it can be overridden to adjust session factory creation logic. So making it private and suggesting the use of `BuildSessionFactory` instead is not exactly an equivalent. But really, I have no knowledge of how exactly overriding it can help someone and if it's really useful.

hazzik: > To me, it looks like it can be overridden to adjust session factory creation logic.

The `IMapping` interface is passed to the session factory, but really not used there (only some piece of code - I did not look carefully - about temporary tables) and, probably, can be removed. So it does not really help to session factory logic. 

> suggesting the use of `BuildSessionFactory` instead is not exactly an equivalent.

Yes, it is not an equivalent. `IMapping` produced by `BuildMapping` is a half-baked mapping at some intermediate step, while `SessionFactory` is fully resolved working instance of `IMapping`. So, yes, it is not equivalent, but more correct way to get `IMapping` unless you're doing some black-magic type of configurations and introspections early on.

src/NHibernate/Cfg/Configuration.cs
bahusoid: I think we don't need this exception anymore. Now misuse in our code is prevented by `GetMapping` obsoletion.

src/NHibernate/Cfg/Configuration.cs
bahusoid: nit: I would keep it public

src/NHibernate/Cfg/Configuration.cs
hazzik: Reverted

src/NHibernate/Cfg/Configuration.cs
hazzik: Made public virtual

src/NHibernate.Test/DialectTest/MsSql2008DialectFixture.cs
bahusoid: Now new public "BuildMapping" can be used.

src/NHibernate/Cfg/Configuration.cs
```diff
@@ -279,23 +272,25 @@ public bool HasNonIdentifierPropertyNamedId(string className)
...
 		protected Configuration(SettingsFactory settingsFactory)
 		{
-			InitBlock();
 			this.settingsFactory = settingsFactory;
 			Reset();
 		}
 
-		private void InitBlock()
+		// Since v5.5
+		[Obsolete("Use BuildSessionFactory to obtain a new ISessionFactory instance instead.")]
```
bahusoid, line 282: Use overload with dialect parameter

src/NHibernate.Test/DialectTest/MsSql2008DialectFixture.cs
hazzik: I was thinking about it and don't like it. `cfg.BuildMapping` just should not exist as it provides an incomplete version of mapping, that can create a lot of problems.

bahusoid: I don't understand your concern. In my understanding `IMapping` was specifically designed to be safely used in "uncompiled" state. See interface description:
https://github.com/nhibernate/nhibernate-core/blob/a0bc61f3a4de1fed25c4c2cbd6011227b5aa0468/src/NHibernate/Engine/IMapping.cs#L6-L7

So as long you use interface methods and don't do some funny castings you should be safe.

You agreed to keep method public - so it would be good to have it in tests too.

bahusoid (approved):

src/NHibernate.Test/DialectTest/MsSql2008DialectFixture.cs
hazzik: I don't think the `IMapping` was designed. It was probably appeared incidentally from some other bad design decision. Problem with the interface that it is not clear when it is safe to use compiled or raw mappings. For example, `SingleTableEntityPersister` accepts both `IMapping` and `ISessionFactoryImplementor`. Can it use only `ISessionFactoryImplementor`? 

https://github.com/nhibernate/nhibernate-core/blob/3e353c1db3c8f106aa43f5d2f5b3b0c232785676/src/NHibernate/Persister/Entity/SingleTableEntityPersister.cs#L79-L81

This is really not clear.

Also, Hibernate has removed this interface and redesigned the mapping compilation.

fredericDelaporte (approved):

PR #3399, opened 2023-08-06
Wrap mapping usages in StaticDialectMappingWrapper in Configuration class

### LINKED ISSUE #3397: GenerateSchemaCreationScript creates many identical dialect instances

bahusoid:
Fixes #3397

This is version of fix with minimal changes and impact specific for released branches. (as hazzik works on more general fix for master) I think it's worth fixing at least in 5.4 (as seems reported issue is [5.0 regression](https://github.com/nhibernate/nhibernate-core/issues/3397#issuecomment-1666660703))


fredericDelaporte:
It could target 5.2.x in my opinion. According to [NuGet stats](https://www.nuget.org/stats/packages/NHibernate?groupby=Version), it still has some significant usage. 5.1.x is not so far below, but targeting it would require to back-port #2011 to v5.4.x.

It should at least target 5.3.x.

Aggregated by minor versions, the [NuGet download stats](https://www.nuget.org/stats/packages/NHibernate?groupby=Version) for last six weeks are:
- 5.4.x: 209 724
- 5.3.x: 588 152
- 5.2.x: 187 245
- 5.1.x: 88 131
- 5.0.x: 7 772
- 4.1.x: 39 710
- 4.0.x: 111 330
- 3.4.x: 11 002
- 3.3.x: 81 307
- 3.2.x: 15 791
- 3.1.x: 7 655
- 3.0.x: 2 985
- 2.1.2: 5 118

src/NHibernate/Cfg/Configuration.cs
```diff
@@ -941,11 +945,12 @@ public string[] GenerateSchemaCreationScript(Dialect.Dialect dialect)
 
 			var script = new List<string>();
 
+			var staticDialectMapping = new StaticDialectMappingWrapper(mapping, dialect);
```
hazzik, line 948: You would potentially need to do same as in `BuildSessionFactory` if you want to leave `mapping` field in place: replace `mapping` field with wrapped instance and back when operation finished, otherwise something can access the unwrapped `mapping` field and then have different dialect than was configured. 

Even if there are no places that would be affected now - this could be in the future. 

This is the reason I've decided to remove `mapping` field altogether.

bahusoid:
> It could target 5.2.x in my opinion. 

Ok. Done

hazzik:
IMO, 5.2 is too old. The issue is too minor and affects only schema generation which usually happens only on a startup of the application.

bahusoid:
5.2 also requires fixing the following errors:
> error NU5125: The 'licenseUrl' element will be deprecated. Consider using the 'license' element instead. 
error NU5048: The 'PackageIconUrl'/'iconUrl' element is deprecated. Consider using the 'PackageIcon'/'icon' element instead. Learn more at https://aka.ms/deprecateIconUrl [D:\BuildAgent\work\30546188361a242\src\NHibernate\NHibernate.csproj]

So 5.3. then?

fredericDelaporte:
Ok then, 5.3.x

src/NHibernate/Cfg/Configuration.cs
fredericDelaporte: As a fix of previous versions, it is good enough for me.

fredericDelaporte (approved):

Issue #3403, opened 2023-08-10
Cascade all-delete-orphan of constrained one-to-one association throws InvalidCastException on explicit flush


nkreipke:
When trying to delete an orphaned child entity in a `one-to-one` association, an `InvalidCastException` is thrown if `session.Flush()` is called explicitly.

Example (full solution at https://github.com/nkreipke/NHOneToOneBug):

```xml
<class name="Entity1">
    <id name="Id">
        <generator class="guid"/>
    </id>
    <one-to-one name="Child" cascade="all-delete-orphan" constrained="true" />
</class>
<class name="Entity2">
    <id name="Id">
        <generator class="foreign">
            <param name="property">Parent</param>
        </generator>
    </id>
    <one-to-one name="Parent" />
</class>
```

```csharp
var entity = session.Query<Entity1>().First(x => x.Id == id);

// Child entity should be deleted due to all-delete-orphan cascading
entity.Child = null;

session.Update(entity);

// !!! This is required for the bug to trigger
session.Flush();

// !!! InvalidCastException
transaction.Commit();
```

This will throw the following exception at `transaction.Commit()`:

```
System.InvalidCastException: Unable to cast object of type 'NHOneToOneBug.Entity2' to type 'NHOneToOneBug.Entity1'.
   at (Object )
   at NHibernate.Bytecode.Lightweight.AccessOptimizer.GetSpecializedPropertyValue(Object target)
   at NHibernate.Bytecode.AccessOptimizerExtensions.GetSpecializedPropertyValue(IAccessOptimizer optimizer, Object target)
   at NHibernate.Tuple.Entity.PocoEntityTuplizer.GetIdentifierPropertyValue(Object entity)
   at NHibernate.Tuple.Entity.AbstractEntityTuplizer.GetIdentifier(Object entity)
   at NHibernate.Persister.Entity.AbstractEntityPersister.GetIdentifier(Object obj)
   at NHibernate.Persister.Entity.AbstractEntityPersister.IsTransient(Object entity, ISessionImplementor session)
   at NHibernate.Engine.ForeignKeys.IsTransientFast(String entityName, Object entity, ISessionImplementor session)
   at NHibernate.Engine.ForeignKeys.IsTransientSlow(String entityName, Object entity, ISessionImplementor session)
   at NHibernate.Event.Default.DefaultDeleteEventListener.OnDelete(DeleteEvent event, ISet`1 transientEntities)
[...]

PR #3406, opened 2023-08-13
Fix orphan removal for detached one-to-one

bahusoid:
Related to #3403 (Reported issue still not fixed throwing some other error. But I believe constrained + delete-orphan is invalid mapping combination)


src/NHibernate/Engine/Cascade.cs
```diff
@@ -203,7 +203,7 @@ private void CascadeProperty(object parent, object child, IType type, CascadeSty
 
 							if (loadedValue != null)
 							{
-								eventSource.Delete(entry.Persister.EntityName, loadedValue, false, null);
```
bahusoid, line 206: This `entry` is for parent. So parent entity name is supplied for removing child entity. This entity name is only used when one-to-one entity is detached - so it's quite rare case.

src/NHibernate.Test/NHSpecificTest/GH3403OneToOne/Fixture.cs
```diff
@@ -0,0 +1,51 @@
...
+			using var session = OpenSession();
+			using var transaction = session.BeginTransaction();
+
+			var entity = session.Get<Entity1>(_id);
+			session.Evict(entity.Child);
+			entity.Child = null;
+
+			session.Flush();
+
+			transaction.Commit();
```
fredericDelaporte, line 48: It would be better to ascertain the result in database too.

```suggestion
			Guid childId;
			using (var session = OpenSession())
			using (var transaction = session.BeginTransaction())
			{
				var entity = session.Get<Entity1>(_id);
				childId = entity.Child.Id
				session.Evict(entity.Child);
				entity.Child = null;

				session.Flush();
				transaction.Commit();
			}

			using (var session = OpenSession())
			using (var transaction = session.BeginTransaction())
			{
				var entity = session.Get<Entity1>(_id);
				Assert.That(entity, Is.Not.Null);
				Assert.That(entity.Child, Is.Null, "Unexpected child on parent");

				var child = session.Get<Entity1>(_id);
				Assert.That(child , Is.Null, "Child is still in database");
			}
```

src/NHibernate.Test/NHSpecificTest/GH3403OneToOne/Fixture.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		[Test]
+		public void OrphanDeleteForDetachedOneToOne()
+		{
+			Guid childId;
+			using (var session = OpenSession())
+			using (var transaction = session.BeginTransaction())
+			{
+				var entity = session.Get<Entity1>(_id);
+				childId = entity.Child.Id
```
fredericDelaporte, line 44: Typo, sorry.
```suggestion
				childId = entity.Child.Id;
```

src/NHibernate.Test/NHSpecificTest/GH3403OneToOne/Fixture.cs
```diff
@@ -0,0 +1,63 @@
...
+				transaction.Commit();
+			}
+
+			using (var session = OpenSession())
+			{
+				var entity = session.Get<Entity1>(_id);
+				Assert.That(entity, Is.Not.Null);
+				Assert.That(entity.Child, Is.Null, "Unexpected child on parent");
+
+				var child = session.Get<Entity2>(_id);
```
fredericDelaporte, line 58: And of course:

```suggestion
				var child = session.Get<Entity2>(childId);
```

fredericDelaporte (approved):

PR #3411, opened 2023-08-15
Remove ISessionFactoryImplementor parameter from TableGroupJoinHelper

hazzik:


bahusoid (approved):
So it's a cleanup for #2680 that made parameter not used

Issue #3438, opened 2023-10-25
[DB2400] ArgumentException Column 'SQL_TYPE_NAME' does not belong to table DataTypes


shessane:
Using NHibernate to access DB2 AS400 database. When upgrading to NHibernate 5.3.13 from 5.2.7, we face a regression in the FluentConfiguration.BuildSessionFactory() :

> FluentNHibernate.Cfg.FluentConfigurationException : An invalid or incomplete configuration was used while creating a SessionFactory. Check PotentialReasons collection, and InnerException for more detail.
> 
> 
> ---- System.ArgumentException : Column 'SQL_TYPE_NAME' does not belong to table DataTypes.
> 
>   Stack Trace: 
> FluentConfiguration.BuildSessionFactory()
>
>  ----- Inner Stack Trace -----
> DB2MetaData.GetReservedWords()
> SchemaMetadataUpdater.GetReservedWords(Dialect dialect, IConnectionHelper connectionHelper)
> SchemaMetadataUpdater.UpdateDialectKeywords(Dialect dialect, IConnectionHelper connectionHelper)
> SchemaMetadataUpdater.Update(ISessionFactoryImplementor sessionFactory)
> SessionFactoryImpl.ctor(Configuration cfg, IMapping mapping, Settings settings, EventListeners listeners)
> Configuration.BuildSessionFactory()
> FluentConfiguration.BuildSessionFactory()

The problem seems the same as [#fluent-nhibernate/504](https://github.com/nhibernate/fluent-nhibernate/issues/504)

Here's the line that create the problem : https://github.com/nhibernate/nhibernate-core/blob/4dac4ca42841ca600e2e70803585c1f46bbe900d/src/NHibernate/Dialect/Schema/DB2MetaData.cs#L39

In a debug session I can see that there's no colomn named "SQL_TYPE_NAME". I have instead a column named "TypeName".

The workaround consists of disabling IncludeDataTypesInReservedWords :

``` csharp
// ...
fluentConfiguration.Database(DB2400Configuration.Standard.ShowSql().ConnectionString(connectionString)
    .Dialect<CustomDB2As400Dialect>())
//...

public class CustomDB2As400Dialect : DB2400Dialect
{
    public override IDataBaseSchema GetDataBaseSchema(DbConnection connection)
    {
        return new CustomAs400MetaData(connection);
    }

    private class CustomAs400MetaData : DB2MetaData
    {
[...]

PR #3439, opened 2023-10-25
Fix ArgumentException : Column 'SQL_TYPE_NAME' does not belong to table DataTypes (#3438)

shessane:
Fix for #3438

Data types will be included in the DB2 reserved words, only if a a column existes with the name "SQL_TYPE_NAME". If not il will not throw an ArgumentException : Column 'SQL_TYPE_NAME' does not belong to table DataTypes.

This PR includes only the uncaught exception. But if you are ok I can add a fallback to the TypeName column if the first does not exist.

### DIFF
--- a/src/NHibernate/Dialect/Schema/DB2MetaData.cs
+++ b/src/NHibernate/Dialect/Schema/DB2MetaData.cs
@@ -37,15 +37,19 @@ namespace NHibernate.Dialect.Schema
 
 			var dtTypes = Connection.GetSchema(DbMetaDataCollectionNames.DataTypes);
 
-			var typeNameColumn = dtTypes.Columns.Cast<DataColumn>()
-							.FirstOrDefault(column => column.ColumnName == "SQL_TYPE_NAME");
+			var typeNameColumnIndex = dtTypes.Columns.IndexOf("SQL_TYPE_NAME");
 
-			if (typeNameColumn == null) //todo We can try to fallback to "TypeName" columnName
-				return result;
+			if (typeNameColumnIndex == -1)
+			{
+				typeNameColumnIndex = dtTypes.Columns.IndexOf("TypeName");
+
+				if (typeNameColumnIndex == -1)
+					return result;
+			}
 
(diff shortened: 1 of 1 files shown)


src/NHibernate/Dialect/Schema/DB2MetaData.cs
```diff
@@ -31,13 +32,20 @@ public override ISet<string> GetReservedWords()
 				result.Add(row["ReservedWord"].ToString());
 			}
 
-			if (IncludeDataTypesInReservedWords)
+			if (!IncludeDataTypesInReservedWords)
+				return result;
+
+			var dtTypes = Connection.GetSchema(DbMetaDataCollectionNames.DataTypes);
+
+			var typeNameColumn = dtTypes.Columns.Cast<DataColumn>()
```
bahusoid, line 40: How many columns are returned?

I believe this check can be simplified with `dtTypes.Columns.Contains("SQL_TYPE_NAME")`. Or  `dtTypes.Columns.IndexOf` and refactor  to use index instead.

shessane: 23 columns in my debug :

image

Just pushed a new version that uses IndexOf. It will also fallback to "TypeName" if "SQL_TYPE_NAME" is not found.

hazzik (dismissed):

fredericDelaporte:
This is fixing a regression introduced in 5.3 by #2058.

Shouldn't we re-target this on the 5.3.x branch?

bahusoid:
> Shouldn't we re-target this on the 5.3.x branch?

Up to you. Should we also re-target mysql appveyour fix?

hazzik:
This fix is not entirely correct. `DB2400Dialect` should use own metadata and not trying to add some fallbacks to a sister-dialect's metadata.

shessane:
Agree, but since the DB2AS400 inherit the implementation, I thought this was the simplest fix without breaking the logic.

I can cancel the PR, up to you

PR #3447, opened 2023-11-19
Migrate appveyor build to MySql 8

fredericDelaporte:
This is an attempt to fix the appveyor build. MySql 5.7 seems to have disappeared from the appveyor images without notice. We would have to use MySql 8 instead.

See appveyor/ci#3895.


hazzik (dismissed):

fredericDelaporte:
According to this [SO answer](https://stackoverflow.com/a/45500633/1178314), it seems we must use at least MySql.Data 8.0.30 for recent MySql servers. We are using the 8.0.27 one.

fredericDelaporte:
All MySql builds fail now, downgrading from 8.2.0 to 8.0.33. There is some issues with lock tests and 8.2.0.

hazzik:
@fredericDelaporte .33 will fail as well.

In https://github.com/nhibernate/nhibernate-core/pull/3095#issuecomment-1616533416:
> @bahusoid: Looks like [v8.0.30](https://github.com/nhibernate/nhibernate-core/actions/runs/2975635759) is the latest working version.


So, the only option is to go to .30: https://stackoverflow.com/a/45500633/259946

hazzik:
https://stackoverflow.com/q/59735301/259946

hazzik:
@fredericDelaporte, this is the issue: https://github.com/nhibernate/nhibernate-core/issues/1302

hazzik:
I would suggest to quote `groups` table in this test case for the purpose of this ticket, and drop the keywords from here.

bahusoid:
> @fredericDelaporte, this is the issue: https://github.com/nhibernate/nhibernate-core/issues/1302

Doesn't look like it to me. #1302 is about using `Configuration.GenerateSchemaCreationScript` directly (not via `SchemaExport`). But quoting is not executed there, instead `SchemaExport.Create` should be used.   Am I missing something?

Also if I understand correctly some new words are now reserved in MySql 8 (like groups)  and  we don't do auto-quoting by default in tests. So failed tests are expected to me - we need to quote new reserved words or register new words + enable auto-qouting in affected test cases.

hazzik:
@bahusoid I’ve debugged the test before I commented here. That is our issue, I’m 100% sure: `SchemaExport` uses scripts generated by `Configuration.GenerateSchemaCreationScript`

bahusoid:
>  SchemaExport uses scripts generated by Configuration.GenerateSchemaCreationScript

Ok. But it also quotes tables and columns:

https://github.com/nhibernate/nhibernate-core/blob/7a86566437b8bcad019213f83f9b63c02af559a3/src/NHibernate/Tool/hbm2ddl/SchemaExport.cs#L63-L71

And direct call to `Configuration.GenerateSchemaCreationScript` in user code won't quote anything (which is your issue about)

hazzik:
@bahusoid, yes, you are right. Sorry, my bad.

fredericDelaporte:
So, I could either enable auto-quoting on taht test, or drop the change and directly quote that new reserved word.

hazzik (approved):

fredericDelaporte:
Replaced by #3453

PR #3461, opened 2023-12-14
Support for DbBatch

gliljas:
### Observations

**A DbBatch looks like a DbCommand, but just on the surface.**
Unlike batching which uses an underlying `DbCommand`, `DbBatch `will require special handling for things such as `Prepare`, `Enlist `etc. While the `DbBatch `could theoretically be wrapped in a `DbCommand `"adapter" or even a NH specific "`DbCommandOrBatch`", since `DbBatch `is now a part of the core .NET 6+ API, I think the route forward is to be explicit about it, e.g. having both `Enlist(DbCommand)` and `Enlist(DbBatch)`.

This means that a transparent move is harder, since custom derivatives of e.g MicrosoftDataSqlClientDriver may have vital implementations of e.g `AdjustCommand `that will have to be duplicated for `DbBatch`.

**No really good way to create a DbBatchCommand from a DbCommand**
If I'm not missing something, I think this is a bit of an oversight on the part of the `DbBatch` creators, especially since they already use "`CanDoXxxx`" properties.
All the parameters in a `DbCommand` will have to be recreated for the `DbBatchCommand`, since they can't be attached to two `DbParameterCollections` at the same time. The good ol' `SqlCommandSet` uses a bit of [trickery](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/System.Data/System/Data/SqlClient/SqlCommandSet.cs#L137) for that, but I just opted to make sure that the `DbParameter` is `ICloneable`, and used that. Perhaps the value should be explicitly cloned too, if it's `ICloneable`. Making it possible to override this logic is probably a good idea.

**ReflectionBasedDriver doesn't use DbProviderFactoryDriveConnectionCommandProvider on .NET 5 and above**
Any reason for this, or just an effect of NETSTANDARD2_1_OR_GREATER not including NET5 etc.?


src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -62,7 +62,7 @@
   <ItemGroup>
     <PackageReference Include="log4net" Version="2.0.15" />
     <PackageReference Include="Microsoft.AspNetCore.OData" Version="7.7.0" />
-    <PackageReference Include="Microsoft.Data.SqlClient" Version="3.1.3" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="5.2.0-preview4.23342.2" />
```
hazzik, line 65: Why this version?

gliljas: SqlClient support was just recently merged.

https://github.com/dotnet/SqlClient/issues/19#issuecomment-1848577977

hazzik: I see

src/NHibernate/Driver/MicrosoftDataSqlClientDriver.cs
```diff
@@ -98,6 +103,8 @@ public override void Configure(IDictionary<string, string> settings)
 			base.Configure(settings);
 
 			_dialect = Dialect.Dialect.GetDialect(settings);
+
+			_useDbBatch = settings.TryGetValue("usedbbatch", out var value) && bool.Parse(value);
```
hazzik, line 107: Do we need this as an option? Users already can configure the batching factory.

src/NHibernate/AdoNet/AbstractBatcher.cs
```diff
@@ -135,6 +136,45 @@ protected void Prepare(DbCommand cmd)
...
+#if NET6_0_OR_GREATER
+		/// <summary>
+		/// Prepares the <see cref="DbBatch"/> for execution in the database.
+		/// </summary>
+		/// <remarks>
+		/// This takes care of hooking the <see cref="DbBatch"/> up to an <see cref="DbConnection"/>
+		/// and <see cref="DbTransaction"/> if one exists.  It will call <c>Prepare</c> if the Driver
+		/// supports preparing batches.
+		/// </remarks>
+		protected void Prepare(DbBatch batch)
```
hazzik, line 148: It feels that this method does not belong to `AbstractBatcher` and should instead be on the `DbBatchBatcher`. This would allow to:
1. Contain logic related to `DbBatch` in `DbBatchBatcher`
2. Have more flexibility in terms of methods

src/NHibernate/Driver/MicrosoftDataSqlClientDriver.cs
gliljas: Removed

hazzik:
Just a few thoughts for the discussion (do not need to implement straight away). The .NET Core since 3 supports default interface members. With it we can add methods to interfaces without introducing breaking changes (it is similar to adding a virtual method for the class in prior versions). Since the DbBatch is .NET 6 and above we could use default methods to extend interfaces instead of doing our usual shimming. This should simplify code a little bit.

src/NHibernate/AdoNet/DbBatchBatcher.cs
```diff
@@ -0,0 +1,297 @@
...
+		}
+
+		protected override async Task DoExecuteBatchAsync(DbCommand ps, CancellationToken cancellationToken)
+		{
+			cancellationToken.ThrowIfCancellationRequested();
+			try
+			{
+				Log.Debug("Executing batch");
+				await (CheckReadersAsync(cancellationToken)).ConfigureAwait(false);
+				//await (PrepareAsync(_currentBatch, cancellationToken)).ConfigureAwait(false);
```
hazzik, line 171: Why not prepare async?

gliljas:
> Just a thought for the discussion (do not need to implement straight away). The .NET Core since 3 supports default interface members. With it we can add methods to interfaces without introducing breaking changes (it is similar to adding a virtual method for the class in prior versions). Since the DbBatch is .NET 6 and above we could use default methods to extend interfaces instead of doing our usual shimming. This should simplify code a little bit.

Yes! I wanted to use that but didn't realize the compiler directives actually meant that I could. Fixed.

src/NHibernate/AdoNet/DbBatchBatcher.cs
gliljas: Good catch. I think I was experimenting with getting the AsyncGenerator to do the job, but got stuck and forgot this one. I'm guessing the compiler directive can get in the way of AsyncGenerator?

Fixed, manually.

gliljas:
We could make it select the DbBatchBatcher automatically when possible, but perhaps that is something we should/could wait with.

gliljas:
@hazzik Thanks for helping out!

A couple of ideas.

1. Maybe we should add the commands to a List<DbCommand> instead of directly to the batch. It would make reuse of e.g existing AdjustCommand implementations easier (we'd call both AdjustCommand and AdjustBatch), and if the batch only contains one command, there no need to add the overhead of creating a batch and cloning the command. Perhaps the IDriver method should be CreateDbBatchFromDbCommands, and it will be called just before executing the batch.
2. The CanCreateBatch naming is maybe a bit off, since it answers the question "Can you create batches **and** can you create batch commands from dbcommands"

fredericDelaporte (commented):
This looks good to me excepted a few commented details.

src/NHibernate/Driver/DriverBase.cs
```diff
@@ -329,6 +329,70 @@ public virtual void AdjustCommand(DbCommand command)
...
+		}
+
+		/// <summary>
+		/// Override to make any adjustments to each DbBatch object before it added to the batcher.
+		/// </summary>
+		/// <param name="batch">The batch.</param>
+		/// <remarks>
+		/// This method is similar to the <see cref="OnBeforePrepare"/> but, instead be called just before execute the command (that can be a batch)
+		/// is executed before add each single command to the batcher and before <see cref="OnBeforePrepare"/> .
+		/// If you have to adjust parameters values/type (when the command is full filled) this is a good place where do it.
```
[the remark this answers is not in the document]
gliljas: It was just copied from AdjustCommand, which is equally weird (not an excuse, I should have seen it). However, it is a bit weird anyway, since in DriverBase, AdjustCommand and OnBeforePrepare are run back to back. There's nothing you can do in AdjustCommand that you can't do in OnBeforePrepare and vice versa.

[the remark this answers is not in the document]
gliljas: Maybe just drop AdjustBatch altogether. "Before added to the batch" doesn't make sense when it's already a batch.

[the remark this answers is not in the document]
fredericDelaporte: Indeed the comment does not make much sense with this method. And if we do not see a need for this batch adjustment point, better not put it in place for now. It is easier to add than to remove.

src/NHibernate/Driver/DriverBase.cs
```diff
@@ -329,6 +329,55 @@ public virtual void AdjustCommand(DbCommand command)
...
+				batch.Prepare();
+			}
+		}
+
+		/// <summary>
+		/// Override to make any adjustments to the DbBatch object.  (e.g., Oracle custom OUT parameter)
+		/// Parameters have been bound by this point, so their order can be adjusted too.
+		/// This is analogous to the RegisterResultSetOutParameter() function in Hibernate.
+		/// </summary>
+		protected virtual void OnBeforePrepare(DbBatch command)
```
fredericDelaporte, line 348: Nit:

```suggestion
		protected virtual void OnBeforePrepare(DbBatch batch)
```

hazzik (commented):
Can you please change not implemented exceptions to not supported?

fredericDelaporte (approved):

fredericDelaporte:
I intend merging this in the coming days (up to next Sunday).

src/NHibernate/Driver/IDriveConnectionCommandProvider.cs
```diff
@@ -6,5 +7,9 @@ public interface IDriveConnectionCommandProvider
 	{
 		DbConnection CreateConnection();
 		DbCommand CreateCommand();
+#if NET6_0_OR_GREATER
+		DbBatch CreateBatch() => throw new NotImplementedException();
```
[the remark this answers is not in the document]
fredericDelaporte: You have marked this as resolved while it is not changed, without another comment. I guess it was a missed click.

src/NHibernate/Driver/DriverBase.cs
```diff
@@ -329,6 +329,55 @@ public virtual void AdjustCommand(DbCommand command)
...
+		/// Parameters have been bound by this point, so their order can be adjusted too.
+		/// This is analogous to the RegisterResultSetOutParameter() function in Hibernate.
+		/// </summary>
+		protected virtual void OnBeforePrepare(DbBatch batch)
+		{
+		}
+
+		public virtual DbBatch CreateBatch()
+		{
+			throw new NotSupportedException();
```
fredericDelaporte, line 354: Switch to not supported (was not implemented) for consistency with the interface default implementation, en because ["not implemented" means "still in dev"](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-notimplementedexception#throw-the-exception) while ["not supported"](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-notsupportedexception#throw-a-notsupportedexception-exception) is meant for implementations that might not support some features of a base class / interface. When `CanCreateBatch` yields `false`, it is adequate to throw `NotSupportedException` from `CreateBatch`.

fredericDelaporte (approved):

PR #3369, opened 2023-07-18
Support association joins from main query to be used in subqueries

bahusoid:
Fixes #3334
Closes #3338
Closes #3335

PR #3476, opened 2024-02-01
Do not append from element when already in from

### LINKED ISSUE #3465: Invalid SQL created for some joins in a subquery
I have found what appears to be a regression in the 5.5.0 release (this worked fine previously with 5.4.6.

The mappings are rather complex, so I'll do my best to explain here:

The hql fragment causing the issue is as follows:

```
and exists(
from om.ActivityShiftAssignments ass, ScheduledActivityShift s 
inner join s.Activity a
where ass.ActivityShift.Id = s.Id
and a.Active = 1
and ( s.ShiftInterval.Start between :fromDate and :toDate or :fromDate between s.ShiftInterval.Start and s.ShiftInterval.End )
)
```

This generates the following sql which is invalid:

```
and (exists (
select activitysh2_.ActivityShiftAssignmentId, scheduleda3_.ActivityShiftId, activity4_.ActivityId 
from ActivityShiftAssignment activitysh2_ 
where scheduleda3_.ActivityTypeId=3 
and organizati0_.OrganizationMemberId=activitysh2_.OrganizationMemberId 
and organizati0_.OrganizationMemberId=activitysh2_.OrganizationMemberId 
and activitysh2_.ActivityShiftId=scheduleda3_.ActivityShiftId 
and activity4_.Active=1 
and (scheduleda3_.StartDate between ? and ? or ? between scheduleda3_.StartDate and scheduleda3_.EndDate))) 
```

As you can see "scheduleda3_" is never included in the from statement, and a join for "activity4_" is never generated either.

I'm not sure what details may be relevant here, other than to say that ScheduledActivityShift is a subclass of ActivityShift, which is directly referenced in the ActivityShiftAssignment entity. I used the implicit join in order to access some of the subclass-specific properties previously.

I can attempt to create a minimal test case, but wanted to put this out there first in case others (with simpler mappings) have encountered the same issue.

fredericDelaporte:
Fix #3465.


fredericDelaporte:
@bahusoid, this fixes a regression from #3369. May you have a look?

It does succeed the test cases from #3369, but I have not check all the other tests, I am waiting for the CI builds.

hazzik (approved):

PR #3481, opened 2024-02-02
Support an arbitrary length for CultureInfoType

### LINKED ISSUE #3473: The sql type of the CultureInfoType is too short
The current generated SQL type for the `CultureInfo` object is string(5):
https://github.com/nhibernate/nhibernate-core/blob/07c367fb859086970737c35f24fe66c089a034e2/src/NHibernate/Type/CultureInfoType.cs#L21

But some culture names are much longer. The longest culture names are 11 characters:
en-US-POSIX
mni-Beng-IN
sat-Olck-IN
shi-Latn-MA
shi-Tfng-MA
tzm-Arab-MA
tzm-Tfng-MA
vai-Latn-LR
vai-Vaii-LR

They are very specific, but in our usecase  we use cultures with length of 10 characters. So this is a real problem for us.
Now we get the following exception:
NHibernate.Exceptions.GenericADOException: 'could not execute batch command.[SQL: SQL not available]'
SqlException: String or binary data would be truncated in table ...

fredericDelaporte:
fix #3473


src/NHibernate.Test/Async/Legacy/FooBarTest.cs
```diff
@@ -5162,6 +5162,7 @@ public async Task EmbeddedCompositeIDAsync()
 			l.CountryCode = "AU";
 			l.Description = "foo bar";
 			l.Locale = CultureInfo.CreateSpecificCulture("en-AU");
+			l.ExtendedLocale = CultureInfo.CreateSpecificCulture("en-US-POSIX");
```
fredericDelaporte, line 5165: There are no dedicated test for CultureInfoType. This one was testing it, along other things. So, I have added the test for the variable length case here. Otherwise a dedicated test in TypesTest could be added.

src/NHibernate/Type/TypeFactory.cs
```diff
@@ -344,11 +346,10 @@ private static void RegisterDefaultNetTypes()
 		}
 
 		/// <summary>
-		/// Register other NO Default .NET type
+		/// Register types which are not a default for a .NET type.
 		/// </summary>
 		/// <remarks>
-		/// These type will be used only when the "type" attribute was is specified in the mapping.
-		/// These are in here because needed to NO override default CLR types and be available in mappings
+		/// These types will be used only when the "type" attribute is specified in the mapping.
```
fredericDelaporte, line 352: I had a hard time making sense of this. I think it is more clear like this.

src/NHibernate.Test/Async/Legacy/FooBarTest.cs
hazzik: I would probably prefer a dedicated test for the type.

fredericDelaporte: Done.

fredericDelaporte:
The DeepSource "failure" is plain wrong: attempting to apply their suggestion can only result in not compiling code. (And classifying as a major issue the use of the old using pattern instead of the newer one is quite an overstatement.)

hazzik (approved):

PR #3483, opened 2024-02-05
Fix concurrency issues on TransactionScope timeout

### LINKED ISSUE #3355: InvalidOperationException on SequencedHashMap.OrderedEnumerator.MoveNext
Every now and then, we see the following exception in our logs:

```
System.InvalidOperationException: Enumerator was modified
   at NHibernate.Util.SequencedHashMap.OrderedEnumerator.MoveNext()
   at NHibernate.Engine.StatefulPersistenceContext.AfterTransactionCompletion()
   at NHibernate.Impl.SessionImpl.AfterTransactionCompletion(Boolean success, ITransaction tx)
   at NHibernate.Transaction.AdoNetWithSystemTransactionFactory.SystemTransactionContext.CompleteTransaction(Boolean isCommitted)
```

To be honest, we have no idea where it exactly comes from or when it occurs. It's quite rare and we haven't been able to reproduce it.

However, it seems to be very similar to #1216. Is it possible this issue wasn't completely fixed?

fredericDelaporte:
fix #3355


src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -458,7 +456,28 @@ protected virtual void CompleteTransaction(bool isCommitted)
...
+					// Thread safety of a concurrent session BeginProcess is ensured by the Wait performed by BeginProcess.
+					var context = _session.BeginProcess();
+					if (context == null)
+					{
+						var timeOutGuard = new Stopwatch();
+						timeOutGuard.Start();
+						while (context == null && timeOutGuard.ElapsedMilliseconds < _systemTransactionCompletionLockTimeout)
+						{
+							// Na�ve yield.
+							Thread.Sleep(10);
```
hazzik, line 471: Can we do this somehow without spin lock?

fredericDelaporte: I guess we could, if we were putting thread locking mechanisms directly in the session, probably whatever the transaction mechanism, instead of keeping them in the transaction factory and applied only on transaction scope finalization.

The `Wait` done by `BeginProcess` actually belongs to the transaction factory and does nothing by default: it locks only during transaction scope finalization. This was done that way to minimize the impact of locks required to handle the transaction scope end processing occurring on concurrent threads.

As a result, when the session initiates some processing inside an ongoing transaction scope, no locks are applied. So, when a transaction timeout occurs concurrently to the session processing, we have no locking mechanism enabled against which we could wait for the end of the session processing.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -458,6 +456,27 @@ protected virtual void CompleteTransaction(bool isCommitted)
 				{
 					// Allow transaction completed actions to run while others stay blocked.
 					_bypassLock.Value = true;
+					// Ensure no other session processing is still ongoing. In case of a transaction timeout, the transaction is
+					// cancelled on a new thread even for non-distributed scopes. So, the session could be doing some processing,
+					// and will not be interrupted until attempting some usage of the connection. See #3355.
+					// Thread safety of a concurrent session BeginProcess is ensured by the Wait performed by BeginProcess.
+					var isProcessing = _session.IsProcessing();
```
fredericDelaporte, line 463: Calling BeginProcess was inadequate: it does not support the session being already closed, it attempt to join the current system transaction if any, ... We actually only need to check if the session is still ongoing some processing.

.editorconfig
```diff
@@ -2,6 +2,7 @@ root=true
 
 [*]
 insert_final_newline = true
+charset = utf-8
```
fredericDelaporte, line 5: A comment was corrupted due to the file being saved as ascii instead of utf-8.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -497,6 +522,15 @@ protected virtual void CompleteTransaction(bool isCommitted)
...
 				}
+
+				if (isSessionProcessing)
+				{
+					throw new HibernateException(
+						"A synchronization timeout occurred at transaction completion: the session was still processing. You may " +
+							$"raise {Cfg.Environment.SystemTransactionCompletionLockTimeout} if it is set too low. It may also " +
+							"be a limitation of the data provider, like not supporting transaction scope timeouts occurring on " +
+							"concurrent threads.");
+				}
```
fredericDelaporte, line 533: ODBC would still "succeed" the test without that throw, but it would take up to nine minutes, due to waiting the synchronization timeout on each attempt. SAP SQL Anywhere would also succeed it, a bit faster, seeming not freezing the session systematically on transaction timeouts. MySQL would sometimes succeed it, but would also be regularly left wrecked by those timeouts occurring on concurrent threads.

So, we may just log the synchronization error without throwing, but it would make it harder to understand the delay occurring on transaction scopes timeout. Not every application records NHibernate logs.

I wonder if another trnasaction factory settings should be added to let the user choose what to do here.

src/NHibernate/Engine/ISessionImplementor.cs
```diff
@@ -66,6 +66,15 @@ internal static IDisposable BeginProcess(this ISessionImplementor session)
 				: SessionIdLoggingContext.CreateOrNull(session.SessionId);
 		}
 
+		internal static bool IsProcessing(this ISessionImplementor session)
+		{
+			if (session == null)
```
hazzik, line 71: nit: you don't need to check for null. `is` will check that.

src/NHibernate/Engine/ISessionImplementor.cs
```diff
@@ -66,6 +66,15 @@ internal static IDisposable BeginProcess(this ISessionImplementor session)
...
+				return false;
+			return session is AbstractSessionImpl impl
```
hazzik, line 73: nit: i think this can be replaced by `return session is AbstractSessionImpl impl && impl.IsProcessing;` but up to you really.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -454,10 +452,37 @@ protected virtual void CompleteTransaction(bool isCommitted)
...
+							// Naïve yield.
+							Thread.Sleep(10);
+							isSessionProcessing = _session.IsProcessing();
+						}
+						if (isSessionProcessing)
+						{
+							// Throwing would give up attempting to close the session if need be, which may still succeed. So,
+							// just log an error.
+							_logger.Error(
+								"A synchronization timeout occurred at transaction completion: the session is still processing. Attempting " +
```
hazzik, line 479: nit: can you please split lines by the sentence ends?

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -497,6 +522,15 @@ protected virtual void CompleteTransaction(bool isCommitted)
 					// Dispose releases blocked threads by the way.
 					Dispose();
 				}
+
+				if (isSessionProcessing)
+				{
+					throw new HibernateException(
+						"A synchronization timeout occurred at transaction completion: the session was still processing. You may " +
```
hazzik, line 529: Same nit with sentence ends.

doc/reference/modules/configuration.xml
```diff
@@ -1050,8 +1050,8 @@ var session = sessions.OpenSession(conn);
                                 after scope disposal. This occurs when the transaction is distributed.
                                 This notably concerns <literal>ISessionImplementor.AfterTransactionCompletion(bool, ITransaction)</literal>.
                                 NHibernate protects the session from being concurrently used by the code following the scope disposal
-                                with a lock. To prevent any application freeze, this lock has a default timeout of five seconds. If the
-                                application appears to require longer (!) running transaction completion events, this setting allows to
+                                with a lock. To prevent any application freeze, this lock has a default timeout of one second. If the
```
fredericDelaporte, line 1053: Five seconds for a thread synchronization timeout is very long. One second is more reasonable.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -454,10 +484,38 @@ protected virtual void CompleteTransaction(bool isCommitted)
...
+						{
+							// Naïve yield.
+							Thread.Sleep(10);
+							isSessionProcessing = _session.IsProcessing();
+						}
+						if (isSessionProcessing)
+						{
+							// Throwing would give up attempting to close the session if need be, which may still succeed. So,
+							// just log an error.
+							_logger.Error(
```
fredericDelaporte, line 510: Maybe this should be "downgraded" to just `Warn`.

fredericDelaporte: Done, also as a way to relaunch the CI. For some reason most Teamcity builds were not triggered.

fredericDelaporte:
Rebased and forced pushed to have all the Teamcity build and maybe fix the WIP status which seems to have some failures for whatever reason.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -454,10 +484,45 @@ protected virtual void CompleteTransaction(bool isCommitted)
...
 				if (!IsInActiveTransaction)
 					return;
+				Lock();
+				// In case of a rollback due to a timeout, we may have the session disposal running concurrently
+				// to the transaction completion in a way our current locking mechanism cannot fully protect: the
+				// session disposal "BeginProcess" can go through the Wait before it is locked but flag the
+				// session as procesisng after the transaction completion has read it as not processing. To dodge
+				// that case, we consider the session as still processing initially regardless of its actual
+				// status in case of rollback.
+				var isSessionProcessing = !isCommitted || _session.IsProcessing();
```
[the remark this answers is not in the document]
fredericDelaporte: Maybe we should just ignore this unlikely concurrency possibility and remove that crutch.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -454,10 +484,45 @@ protected virtual void CompleteTransaction(bool isCommitted)
...
 				try
 				{
 					// Allow transaction completed actions to run while others stay blocked.
 					_bypassLock.Value = true;
```
hazzik, line 498: Maybe the proper fix would be to remove bypass lock?

hazzik: What happens if before Lock() we put Wait()?

fredericDelaporte: Systematic deadlock then lock timeout with thrown exception, for both questions.

If the transaction is no more considered active, `Wait` activates the lock, which will block until deactivated, whatever the thread. Wait then attempt to go through the lock unless its thread has bypass lock set. Bypass lock allows the transaction completion thread to not get blocked by the lock.

src/NHibernate/Impl/AbstractSessionImpl.cs
```diff
@@ -405,16 +410,31 @@ public bool IsClosed
...
-		/// <returns>
-		/// If not already processing, an object to dispose for restoring the previous session id.
-		/// Otherwise, <see langword="null" />.
-		/// </returns>
-		public IDisposable BeginContext()
+        }
+
+        /// <summary>
+        /// If not nested in another call to <c>BeginProcess</c> on this session, optionnaly check
+		/// and update the session status, then set its session id in context and flag it as processing.
```
hazzik, line 417: nit: some whitespace issues here.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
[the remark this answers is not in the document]
hazzik: > Maybe we should just ignore this unlikely concurrency possibility and remove that crutch.

I would agree with this.

hazzik (dismissed):

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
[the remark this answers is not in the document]
fredericDelaporte: Removed.

src/NHibernate/Impl/AbstractSessionImpl.cs
fredericDelaporte: Fixed.

hazzik (approved):

fredericDelaporte:
That is flaky. I have launched the test locally repeatedly and ended up a single time with the failure. It yielded me no additional information allowing me to ascertain if that is a teardown specific concurrency issue or still a NHibernate issue. That one time was in simple "run" mode, not in debug mode. I have not yet reproduced it in under debug.

src/NHibernate.Test/DebugSessionFactory.cs
```diff
@@ -43,14 +43,13 @@ public partial class DebugSessionFactory : ISessionFactoryImplementor
...
-			=> _debugConnectionProvider ??
-				(_debugConnectionProvider = ActualFactory.ConnectionProvider as DebugConnectionProvider);
+			=> _debugConnectionProvider ??= ActualFactory.ConnectionProvider as DebugConnectionProvider;
 		public ISessionFactoryImplementor ActualFactory { get; }
 
 		public EventListeners EventListeners => ((SessionFactoryImpl)ActualFactory).EventListeners;
 
 		[NonSerialized]
-		private readonly ConcurrentBag<ISessionImplementor> _openedSessions = new ConcurrentBag<ISessionImplementor>();
+		private readonly ConcurrentQueue<ISessionImplementor> _openedSessions = new();
```
fredericDelaporte, line 52: I finally reproduced the trouble in debug mode, but still did not get more information about it. The unclosed session was the 51st in the bag among 160, but that is an unordered collection, so that does not mean much. With a queue, it will be possible to reliably know if that is the last one of the test or not.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -264,16 +264,10 @@ public bool ShouldAutoClose
...
-		/// <remarks>
-		/// Do not call this method inside a transaction scope, use <c>Dispose</c> instead, since
-		/// Close() is not aware of distributed transactions
-		/// </remarks>
-		/// </summary>
+		/// <inheritdoc />
 		public DbConnection Close()
 		{
-			using (BeginContext())
+			using (BeginProcess(true))
```
fredericDelaporte, line 270: For consistency with Dispose. It does not make it more "aware" of system transactions. The non-awareness of this method is that it closes unconditionally, instead of delaying closing to the transaction completion if called in an ongoing transaction scope with flush from completion enabled.

src/NHibernate/Impl/SessionImpl.cs
```diff
@@ -1488,18 +1482,18 @@ public void Reconnect(DbConnection conn)
...
-				TransactionContext?.Wait();
-				// If the synchronization above is bugged and lets a race condition remaining, we may
+				// If the BeginProcess synchronization is bugged and lets a race condition remaining, we may
 				// blow here with a null ref exception after the null check. We could introduce
 				// a local variable for avoiding it, but that would turn a failure causing an exception
 				// into a failure causing a session and connection leak. So do not do it, better blow away
 				// with a null ref rather than silently leaking a session. And then fix the synchronization.
-				if (TransactionContext != null && TransactionContext.CanFlushOnSystemTransactionCompleted)
+				if (TransactionContext != null && TransactionContext.CanFlushOnSystemTransactionCompleted &&
+					TransactionContext.IsInActiveTransaction)
```
fredericDelaporte, line 1496: For consistency with Cleanup.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -266,8 +299,8 @@ protected virtual void Lock()
 			{
 				if (!_needCompletionLocking || _isDisposed)
 					return;
-				_needCompletionLocking = false;
 				_lock.Reset();
+				_needCompletionLocking = false;
```
fredericDelaporte, line 303: There was a concurrency issue possible here, allowing the session disposal to not wait where it should have. But the case I have seen may be bound to the theoretical case for which I have put initially a crutch.

I will try for a few more days to reproduce the issue with this change, to somewhat check if the session not closed issue does still occur.

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
```diff
@@ -497,14 +564,26 @@ protected virtual void CompleteTransaction(bool isCommitted)
...
+				{
+					throw new HibernateException(
+						"A synchronization timeout occurred at transaction completion: the session was still processing. " +
+							$"You may raise {Cfg.Environment.SystemTransactionCompletionLockTimeout} if it is set too low. " +
+							"It may also be a limitation of the data provider, " +
+							"like locks applied on its side while processing transaction cancellations occurring on concurrent threads, " +
+							"thus preventing the session to finish its current processing during a transaction cancellation. " +
+							$"In such case, you may enable {Cfg.Environment.IgnoreSessionSynchronizationFailuresOnSystemTransaction}, " +
+							$"and possibly lower {Cfg.Environment.SystemTransactionCompletionLockTimeout}.");
+				}
```
fredericDelaporte, line 578: In my repeated test runs, this has occurred quite a few times with PostgreSQL. Waiting longer allow the end of the session processing. So, maybe we will need to raise the transaction completion lock timeout for PostgreSQL in order to avoid some flakiness of this test.

The timeout occurring also causes some concurrency failures to become possible since transaction completion is run concurrently to the session processing, and I have seen it once. Unfortunately in such case only the NHibernate logs allows to see there was a transaction completion lock timeout, but these do not seem to be surfaced in Teamcity logs or in test run logs from Visual Studio. So, it may cause some confusion on whether there is still concurrency issues or if it was a trouble of transaction lock timeout.

hazzik (approved):

src/NHibernate/Transaction/AdoNetWithSystemTransactionFactory.cs
fredericDelaporte: The failure does still happen. I have put back in place the crutch, but only for the case which can fail : when session usage from transaction scope completion is not disabled. That is a legacy case we advise to not use.

I will again try for a few days to reproduce the issue.

fredericDelaporte: Reverting: the unclosed session issue does still occur. So, there is a remaining concurrency issue, but only for the case where using the session from transaction completion is not disabled as recommended. I think I let it lie there.

src/NHibernate.Test/SystemTransactions/SystemTransactionFixture.cs
```diff
@@ -643,6 +650,152 @@ public void CanUseSessionWithManyDependentTransaction(bool explicitFlush)
...
+		{
+			Assume.That(TestDialect.SupportsTransactionScopeTimeouts, Is.True, "The tested dialect is not supported for transaction scope timeouts.");
+			// Other special cases: ODBC and SAP SQL Anywhere succeed this test only with transaction.ignore_session_synchronization_failures
+			// enabled.
+			// They freeze the session during the transaction cancellation. To avoid the test to be very long, the synchronization
+			// lock timeout should be lowered too.
+
+			// A concurrency issue exists with the legacy setting allowing to use the session from transaction completion, which
+			// may cause session leaks. Ignore them.
+			FailOnNotClosedSession = !UseConnectionOnSystemTransactionPrepare;
```
fredericDelaporte, line 666: The PostgreSQL CI has again failed with the session leak. I am adding a condition to ignore that leak.

fredericDelaporte:
As this sometimes happens, the Oracle build had a timeout, taking more than one hour to complete. On relaunch, it had worked.

hazzik (approved):

PR #3485, opened 2024-02-06
Fix implied commits for Oracle and MySQL

### LINKED ISSUE #3474: PerformTemporaryTableDDLInIsolation seems to be wrong for oracle dialect
NHibernate Version: 5.4.2
.net framework 4.8

We observed that running ExecuteUpdate on a HQL-Query causes updates that have been done before to be committed afterwards.
```
session.BeginTransaction();
session.Merge(someEntity);
session.Flush();
session.CreateQuery("Update MultiTableEntity set property = value where id = 123").ExecuteUpdate();
```
Debugging into NHibernate I was able to track this down to `ShouldIsolateTemporaryTableDDL` at least it seems like it.
For some reason NHibernate tries to create a (temporary) table which apparently fails because it already exists but just logs the error as debug and continues.
Since creating tables in oracle is causing a commit this seems to be the issue.
I noticed that there is a function `ShouldIsolateTemporaryTableDDL` in the class `AbstractStatementExecutor` which apparently first checks the dialect for `PerformTemporaryTableDDLInIsolation` which returns false or rather null when using the oracle dialect which if I am understanding this correctly should return true because oracle has an implicit commit when doing DDL operations and therefore DDL should run in isolation.

```		
protected virtual bool ShouldIsolateTemporaryTableDDL()
		{
			bool? dialectVote = Factory.Dialect.PerformTemporaryTableDDLInIsolation();
			if (dialectVote.HasValue)
			{
				return dialectVote.Value;
			}
			return Factory.Settings.IsDataDefinitionImplicitCommit;
		}
```

fredericDelaporte:
fix #3474


src/NHibernate/Dialect/Dialect.cs
```diff
@@ -717,28 +717,28 @@ public virtual string GenerateTemporaryTableName(string baseTableName)
...
+		/// <term><see langword="true" /></term>
+		/// <description>Unequivocally, perform the temporary table DDL in isolation.</description>
+		/// </item>
+		/// <item>
+		/// <term><see langword="false" /></term>
+		/// <description>Unequivocally, do <b>not</b> perform the temporary table DDL in isolation.</description>
+		/// </item>
+		/// <item>
+		/// <term><see langword="null" /></term>
+		/// <description>Defer to <see cref="Cfg.Settings.IsDataDefinitionImplicitCommit" />.</description>
```
fredericDelaporte, line 737: That setting remains dead code, though. I considered exposing it in the XML configuration, but it seemed out of place. I guess it was meant for the Hibernate logic, querying some metadata of the provider allowing to get the information, rather than hard-coding the behavior in configuration.

src/NHibernate/Dialect/MySQLDialect.cs
```diff
@@ -451,6 +451,9 @@ public override string CreateTemporaryTableString
 			get { return "create temporary table if not exists"; }
 		}
 
+		/// <inheritdoc />
+		public override string DropTemporaryTableString => "drop temporary table";
```
fredericDelaporte, line 455: The drop works without the `temporary` keyword, but causes an implied commit.

hazzik (approved):

Issue #2407, opened 2020-05-29
LinkedHaspMap - need info


PriyaGurijala:
Hello,
Can I use `NHibernate.Util.LinkedHashMap` with .Net standard 4.5 ? Please help me.

fredericDelaporte:
As [.Net Standard](https://docs.microsoft.com/fr-fr/dotnet/standard/net-standard) latest version is 2.1, I guess you ask about using `NHibernate.Util.LinkedHashMap` with .Net Framework 4.5.

- NHibernate is not meant to be an utility library. This kind of classes should have been made internal in my opinion. In other words, we do not support external uses of it.
- Just use an old enough NHibernate version. NHibernate v5.x targets .Net Framework v4.6.1, so you cannot use it with .Net framework v4.5. NHibernate v4.x are usable with .Net Framework v4.0, so it can also be used with .Net Framework v4.5.
- You have reached the issue tracker, meant for bug reports or feature requests.
  Please post on the [nhusers group](https://groups.google.com/forum/#!forum/nhusers) for support requests (or see [Groups](https://nhibernate.info/groups.html)).

PR #3487, opened 2024-02-08
Allocation free iterators for LinkedHashMap

### LINKED ISSUE #3486: LinkedHashMap creates enormous amount of memory traffic
Because of an efficient implementation of Enumerable pattern.
It creates a significant pressure to GC.

Most of the traffic come from boxed `KeyValuePair<CollectionEntry, IPersistentCollection>` values, but Enumerator also could be made a `struct` for removing allocations further.

During a startup our application caches a lot of data.
So, the image below is for our case.
image

I work on a fix, check my solution against original using DotNetBenchmark.

wilbit:
fixes #3486


src/NHibernate/Util/LinkedHashMap.cs
```diff
@@ -182,12 +182,22 @@ private void OverrideEntry(Entry e, TValue value)
 			InsertEntry(e);
 		}
 
-		public virtual ICollection<TKey> Keys
+		public KeyCollection Keys
```
fredericDelaporte, line 185: Such changes are binary breaking, the class and property being public. So, such a change could only be merged for a new major version. For a faster release of the change, you need to ensure all changes are non-breaking. So, you would have to introduce new properties/methods and obsolete the old ones, instead of changing in a breaking way existing ones.

src/NHibernate/Util/LinkedHashMap.cs
```diff
@@ -385,7 +400,7 @@ public override string ToString()
 
 		#endregion
 
-		private class KeyCollection : ICollection<TKey>
+		public class KeyCollection : ICollection<TKey>
```
fredericDelaporte, line 403: To avoid increasing the API for which we have to avoid breaking changes, better switch it to internal instead, if possible.

src/NHibernate/Util/LinkedHashMap.cs
```diff
@@ -457,11 +477,33 @@ IEnumerator IEnumerable.GetEnumerator()
 
 			#endregion
 
-			private class Enumerator : ForwardEnumerator<TKey>
+			public struct Enumerator : IEnumerator<TKey>
```
fredericDelaporte, line 480: Better make it internal if possible.

src/NHibernate/Util/LinkedHashMap.cs
```diff
@@ -471,10 +513,19 @@ public override TKey Current
...
+				{
+					current = dictionary.header;
+				}
+
+				void IDisposable.Dispose() { }
 			}
 		}
 
-		private class ValuesCollection : ICollection<TValue>
+		public class ValuesCollection : ICollection<TValue>
```
fredericDelaporte, line 528: Same here.

src/NHibernate/Util/LinkedHashMap.cs
```diff
@@ -546,11 +602,33 @@ IEnumerator IEnumerable.GetEnumerator()
 
 			#endregion
 
-			private class Enumerator : ForwardEnumerator<TValue>
+			public struct Enumerator : IEnumerator<TValue>
```
fredericDelaporte, line 605: Same here.

src/NHibernate/Util/LinkedHashMap.cs
```diff
@@ -560,32 +638,31 @@ public override TValue Current
...
+				{
+					current = dictionary.header;
+				}
+
+				void IDisposable.Dispose() { }
 			}
 		}
 
-		private abstract class ForwardEnumerator<T> : IEnumerator<T>
+		public struct Enumerator : IEnumerator<KeyValuePair<TKey, TValue>>
```
fredericDelaporte, line 653: Same here.

src/NHibernate.Test/UtilityTest/LinkedHashMapFixture.cs
```diff
@@ -364,6 +405,11 @@ public void Performance()
 				Console.Out.WriteLine();
 			}
 		}
+
+		private static System.Type GetVariableType<T>(T instance)
+		{
+			return typeof(T);
+		}
```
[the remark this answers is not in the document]
wilbit: `.GetType` will return a real type of instance, I needed to check a type of variable (basically, a return type of chosen `.GetEnumerator`)

src/NHibernate.Test/UtilityTest/LinkedHashMapFixture.cs
```diff
@@ -161,6 +162,46 @@ public void GetEnumeratorModifyExceptionFromUpdate()
...
+
+		[Test]
+		public void EnumeratorInstanceShouldBeStruct()
+		{
+			var lhm = new LinkedHashMap<string, Player>();
+			var enumerator = lhm.GetEnumerator();
+			var enumeratorType = GetVariableType(enumerator);
+
+			Assert.IsFalse(enumeratorType.IsClass || enumeratorType.IsInterface);
```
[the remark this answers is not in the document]
wilbit: I wanted to check that it is struct, so, I would prefer to keep 1 assertion (`Type` does not have `IsStruct` property =()

src/NHibernate/Util/LinkedHashMap.cs
wilbit: According to your https://github.com/nhibernate/nhibernate-core/issues/2407#issuecomment-636494042

> NHibernate is not meant to be an utility library. This kind of classes should have been made internal in my opinion. In other words, we do not support external uses of LinkedHashMap.

So, maybe, it is even better to make `LinkedHashMap` class internal.

src/NHibernate/Util/LinkedHashMap.cs
wilbit: I cannot make it internal because it is used in a public method.
I would suggest to make `LinkedHashMap` internal.

src/NHibernate/Util/LinkedHashMap.cs
wilbit: same here

src/NHibernate/Util/LinkedHashMap.cs
wilbit: same here

src/NHibernate/Util/LinkedHashMap.cs
wilbit: same here

src/NHibernate/Util/LinkedHashMap.cs
wilbit: same here

src/NHibernate.Test/UtilityTest/LinkedHashMapFixture.cs
```diff
@@ -161,6 +162,46 @@ public void GetEnumeratorModifyExceptionFromUpdate()
...
+		}
+
+		[Test]
+		public void KeysEnumeratorInstanceShouldBeStruct()
+		{
+			var lhm = new LinkedHashMap<string, Player>();
+			var enumerator = lhm.Keys.GetEnumerator();
+			var enumeratorType = GetVariableType(enumerator);
+
+			Assert.IsFalse(enumeratorType.IsClass || enumeratorType.IsInterface);
```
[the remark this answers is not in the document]
wilbit: same here

src/NHibernate.Test/UtilityTest/LinkedHashMapFixture.cs
```diff
@@ -161,6 +162,46 @@ public void GetEnumeratorModifyExceptionFromUpdate()
...
+		}
+
+		[Test]
+		public void ValuesEnumeratorInstanceShouldBeStruct()
+		{
+			var lhm = new LinkedHashMap<string, Player>();
+			var enumerator = lhm.Values.GetEnumerator();
+			var enumeratorType = GetVariableType(enumerator);
+
+			Assert.IsFalse(enumeratorType.IsClass);
```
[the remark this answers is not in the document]
wilbit:
```suggestion
			Assert.That(enumeratorType.IsClass || enumeratorType.IsInterface, Is.False, "The enumerator should be a struct");
```

src/NHibernate/Util/LinkedHashMap.cs
wilbit: I cannot use `ObsoleteAttribute` here, because a class cannot have 2 properties with the same name.

src/NHibernate.Test/UtilityTest/LinkedHashMapFixture.cs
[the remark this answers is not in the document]
fredericDelaporte: ok

src/NHibernate/Util/LinkedHashMap.cs
fredericDelaporte: > it is even better to make LinkedHashMap class internal

Yes it is better, but that would be a breaking change, so, it would have to wait for a major release.

> I don't care about getting these changes faster.

Unless we start issuing major versions for rather minor changes, I do not think a major version will come any time soon. There are many things that should be done for an actual new major version, but I currently do not see enough contributing workforce available to do them. There are approved PR in the next major milestone lingering there since years...

> I cannot use ObsoleteAttribute here, because a class cannot have 2 properties with the same name.

But you can add it with another name. Granted, that would not be ideal.

I know `GetEnumerator` has a special treatment by the compiler for foreach statement, so for this one, using another name is not an option.

hazzik:
@wilbit @fredericDelaporte would it be feasible to create a new class and use it?

wilbit:
> @wilbit @fredericDelaporte would it be feasible to create a new class and use it?

Well, I see that `LinkedHashMap` is not used in public properties/fields/method, so, technically it is feasible.
But also, `LinkedHashMap` implements `IDeserializationCallback` interface, and I'm not really sure what it brings. If nothing, that we are good to go and the only thing which is left is a new name and/or namespace for the internal class.

Another suggestion from me:
1. keep the changes from this PR for a next major release (after some fixes and a rebase, see point 2)
2. for an upcoming minor release fix the major part of #3486 by casting `LinkedHashMap` to `IDictionary<TKey, TValue>` in `BatchFetchQueue.GetCollectionBatchAsync` (this should do a trick and `foreach` will use generic `IEnumerator<>` instead of non-generic `IEnumerator`).

fredericDelaporte:
Obsoleting LinkedHashMap entirely and adding an internal replacement seems to me as the best way to handle this.

> But also, `LinkedHashMap` implements `IDeserializationCallback` interface, and I'm not really sure what it brings.

It is for supporting binary serialization. See also the `Serializable` attribute of the class. That thing is obsoleted in .Net Core for security reasons, but it is still used by many users especially under .Net Framework. The replacement class will have to be binary serializable too, because it will serve as state of many other classes which are serializable too. This allows to serialize the whole session and restore it later, by example.

wilbit:
@fredericDelaporte
1. OK, where to put the new internal class and how to name it?
2. about serialisation: would it work if somebody serialize a session with the current `LinkedHashMap` and deserialize with internal `LinkedHashMap` (which is, basically, would be a copy of the current `LinkedHashMap` in terms of fields...)?

fredericDelaporte:
1. Why not `LinkHashMap`, same place?
2. That will not work and will be a possible breaking change, but neither a binary one nor a source one, which is acceptable for a minor. We will have to warn in the release notes about the incompatibility of states serialized by version prior to the next minor with this next minor, and vice-versa.

wilbit:
added LinkHashMap, @hazzik , @fredericDelaporte

src/NHibernate/Util/LinkedHashMap.cs
```diff
@@ -19,6 +19,8 @@ namespace NHibernate.Util
 	/// </remarks>
 	[DebuggerTypeProxy(typeof(CollectionProxy<>))]
 	[Serializable]
+	// Since 5.5
+	[Obsolete("This class has no more usages and will be removed in a future version. Replaced with internal LinkHashMap class")]
```
fredericDelaporte, line 23: It will be 5.6. No need to mention the replacement since it will not be available for external use.

```suggestion
	// Since 5.6
	[Obsolete("This class has no more usages and will be removed in a future version.")]
```

src/NHibernate/Util/LinkHashMap.cs
```diff
@@ -0,0 +1,650 @@
...
+	[Serializable]
+	internal class LinkHashMap<TKey, TValue> : IDictionary<TKey, TValue>, IDeserializationCallback
+	{
+		[Serializable]
+		protected class Entry
+		{
+			private readonly TKey key;
+			private TValue evalue;
+			private Entry next;
+			private Entry prev;
```
fredericDelaporte, line 30: To avoid the `this`, switch to `_` prefix for class members.

```suggestion
			private readonly TKey _key;
			private TValue _value;
			private Entry _next;
			private Entry _prev;
```

src/NHibernate/Util/LinkHashMap.cs
```diff
@@ -0,0 +1,650 @@
...
+			{
+				return "[" + key + "=" + evalue + "]";
+			}
+
+			#endregion
+		}
+
+		private readonly Entry header;
+		private readonly Dictionary<TKey, Entry> entries;
+		private long version;
```
fredericDelaporte, line 88:
```suggestion
		private readonly Entry _header;
		private readonly Dictionary<TKey, Entry> _entries;
		private long _version;
```

src/NHibernate/Util/LinkHashMap.cs
```diff
@@ -0,0 +1,650 @@
...
+			buf.Append(']');
+
+			return buf.ToString();
+		}
+
+		#endregion
+
+		public class KeyCollection : ICollection<TKey>
+		{
+			private readonly LinkHashMap<TKey, TValue> dictionary;
```
fredericDelaporte, line 385:
```suggestion
			private readonly LinkHashMap<TKey, TValue> _dictionary;
```

src/NHibernate/Util/LinkHashMap.cs
```diff
@@ -0,0 +1,650 @@
...
+
+			IEnumerator IEnumerable.GetEnumerator() => new Enumerator(dictionary);
+
+			#endregion
+
+			public struct Enumerator : IEnumerator<TKey>
+			{
+				private readonly LinkHashMap<TKey, TValue> dictionary;
+				private Entry current;
+				private readonly long version;
```
fredericDelaporte, line 449:
```suggestion
				private readonly LinkHashMap<TKey, TValue> _dictionary;
				private Entry _current;
				private readonly long _version;
```

src/NHibernate/Util/LinkHashMap.cs
```diff
@@ -0,0 +1,650 @@
...
+					current = dictionary.header;
+				}
+
+				void IDisposable.Dispose() { }
+			}
+		}
+
+		public class ValueCollection : ICollection<TValue>
+		{
+			private readonly LinkHashMap<TKey, TValue> dictionary;
```
fredericDelaporte, line 495:
```suggestion
			private readonly LinkHashMap<TKey, TValue> _dictionary;
```

src/NHibernate/Util/LinkHashMap.cs
```diff
@@ -0,0 +1,650 @@
...
+
+			IEnumerator IEnumerable.GetEnumerator() => new Enumerator(dictionary);
+
+			#endregion
+
+			public struct Enumerator : IEnumerator<TValue>
+			{
+				private readonly LinkHashMap<TKey, TValue> dictionary;
+				private Entry current;
+				private readonly long version;
```
fredericDelaporte, line 559:
```suggestion
				private readonly LinkHashMap<TKey, TValue> _dictionary;
				private Entry _current;
				private readonly long _version;
```

src/NHibernate/Util/LinkHashMap.cs
```diff
@@ -0,0 +1,650 @@
...
+
+				void IDisposable.Dispose() { }
+			}
+		}
+
+		public struct Enumerator : IEnumerator<KeyValuePair<TKey, TValue>>
+		{
+			private readonly LinkHashMap<TKey, TValue> dictionary;
+			private Entry current;
+			private readonly long version;
```
fredericDelaporte, line 607:
```suggestion
			private readonly LinkHashMap<TKey, TValue> _dictionary;
			private Entry _current;
			private readonly long _version;
```

src/NHibernate.Test/UtilityTest/LinkHashMapFixture.cs
```diff
@@ -0,0 +1,474 @@
...
+			IDictionary<string, Player> lhm = new LinkHashMap<string, Player>();
+			lhm["123"] = new Player("123", "yyyyyyy");
+			Assert.Throws<InvalidOperationException>(
+				() =>
+				{
+					foreach (var pair in lhm)
+					{
+						lhm["78945"] = new Player("78945", "Someone");
+					}
+				});
```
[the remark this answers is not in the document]
wilbit: Got it...

src/NHibernate/Util/LinkHashMap.cs
wilbit: I thought it was o code style :)))

src/NHibernate.Test/UtilityTest/LinkedHashMapFixture.cs
```diff
@@ -8,6 +8,8 @@
 namespace NHibernate.Test.UtilityTest
 {
 	[TestFixture]
+	// Since 5.5
+	[Obsolete("LinkedHashMap class has no more usages and will be removed in a future version.")]
 	public class LinkedHashMapFixture
```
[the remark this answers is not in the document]
wilbit: applied

src/NHibernate.Test/UtilityTest/LinkHashMapFixture.cs
```diff
@@ -0,0 +1,474 @@
...
+
+			int entries = 0;
+			foreach (KeyValuePair<string, Player> pair in lhm)
+				entries++;
+			foreach (string s in lhm.Keys)
+				entries++;
+			foreach (Player value in lhm.Values)
+				entries++;
+
+			Assert.AreEqual(0, entries, "should not have any entries in the enumerators");
```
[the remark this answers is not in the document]
wilbit: applied `Assert.That(entries, Is.Zero, "should not have any entries in the enumerators");`

src/NHibernate.Test/UtilityTest/LinkHashMapFixture.cs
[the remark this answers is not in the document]
wilbit: applied

src/NHibernate.Test/UtilityTest/LinkHashMapFixture.cs
```diff
@@ -0,0 +1,474 @@
...
+			IDictionary<string, Player> lhm = new LinkHashMap<string, Player>();
+			lhm["123"] = new Player("123", "yyyyyyy");
+			Assert.Throws<InvalidOperationException>(
+				() =>
+				{
+					foreach (var pair in lhm)
+					{
+						lhm["123"] = new Player("123", "aaaaaaa");
+					}
+				});
```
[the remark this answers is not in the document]
wilbit: applied

src/NHibernate/Util/LinkHashMap.cs
wilbit: applied

src/NHibernate/Util/LinkHashMap.cs
wilbit: applied

src/NHibernate/Util/LinkHashMap.cs
wilbit: applied

src/NHibernate/Util/LinkHashMap.cs
wilbit: applied

src/NHibernate/Util/LinkHashMap.cs
wilbit: applied

src/NHibernate.Test/UtilityTest/LinkHashMapFixture.cs
```diff
@@ -0,0 +1,474 @@
...
+			Assert.That(dlhm.Count, Is.EqualTo(6));
+			var index = 0;
+			foreach (var pair in dlhm)
+			{
+				Assert.That(pair.Key, Is.EqualTo(players[index].Id));
+				Assert.That(pair.Value, Is.EqualTo(players[index]));
+				index++;
+			}
+
+			Assert.AreEqual(6, index);
```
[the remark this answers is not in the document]
wilbit: applied

src/NHibernate/Util/LinkedHashMap.cs
wilbit: applied

src/NHibernate/Util/LinkHashMap.cs
wilbit: applied

src/NHibernate/Util/LinkHashMap.cs
wilbit: applied

src/NHibernate.Test/UtilityTest/LinkHashMapFixture.cs
```diff
@@ -0,0 +1,474 @@
...
+		}
+
+		[Test]
+		public void Contains()
+		{
+			var lhm = new LinkHashMap<string, Player>();
+			Fill(lhm);
+
+			Assert.That(lhm.Contains("12341"), Is.True);
+			Assert.That(lhm.Contains("55555"), Is.False);
```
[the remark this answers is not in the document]
wilbit: I do not change that because:
- The suggestion does not work, because `Does.Contain` checks existence against KeyValuePairs collection.
- `LinkHashMap.Contains(string)` is a target of this test.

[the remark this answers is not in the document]
fredericDelaporte: Ok.

src/NHibernate/Util/LinkHashMap.cs
fredericDelaporte: There are no [Microsoft guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields) for private fields. `camelCase` naming without `_` prefix in NHibernate is generally the result of port from Java without renaming. This causes the fields to be looking like local variables and methods parameters. The `_` prefix is deemed ugly by some, but has the merit to allow obvious recognition of fields versus local variables and method parameters.

src/NHibernate/Util/LinkHashMap.cs
```diff
@@ -0,0 +1,641 @@
...
+	/// </remarks>
+	[DebuggerTypeProxy(typeof(CollectionProxy<>))]
+	[Serializable]
+	internal class LinkHashMap<TKey, TValue> : IDictionary<TKey, TValue>, IDeserializationCallback
+	{
+		[Serializable]
+		protected class Entry
+		{
+			private readonly TKey _key;
+			private TValue _evalue;
```
fredericDelaporte, line 28: The `e` was very likely a "trick" to dodge the fact that `value` is a reserved word. (Instead of using `@` which is meant for this purpose.) There are no reason to keep it.

```suggestion
			private TValue _value;
```

wilbit: Maybe, I wasn't sure because of the serialisation...
Also one question bother me: may I
- remove fields by using auto properties
- or remove properties and make fields "public"?
what would be a safe move to clean `Entry` class?

fredericDelaporte: You can remove fields by using auto properties.

wilbit: OK, I've renamed `_evalue` to `_value`, but I did not use auto properties because generated backend fields have longer names and I realised that size of binary serialized object (stream/array) depends on:
- length of field names, 1 byte per character
- length of parent class name for nested classes, 3 bytes per character
- length of class name, 3 bytes per character
- length of namespace name, 3 bytes per character

Current size of 1 serialized object like `LinkHashMap<int,int>.Entry` is 826 bytes long.
- Using of auto properties makes it 890 bytes long (+7.75%)
- Extracting `Entry` class directly to NHibernate namespace and using field names like:
```c#
private readonly TKey k;
private TValue v;
private Entry<TKey,TValue> n;
private Entry<TKey,TValue> p;
```
makes it 759 bytes long (-8.11%)

So, I would not use auto properties and also consider to rename fields to 1 character long names at least. This will make it 810 bytes long (-1.94%).

Any thought? Because I don't really know how important it is (I don't use serialisation in NHibernate and not familiar with this feature, but it seems important to mention).

fredericDelaporte: We do not really care about the resulting size of the serialized binary. I have no recollection of this subject having ever been raised. Binary serialization is for us a legacy that we do not want to break in a minor: as long as it keeps working, that is good enough for us.

So, no need to bother about the resulting size. Code cleanliness is more important for us than the impacts it may have on the size of state binary serialized. (No one letter backing fields please!)

fredericDelaporte (dismissed):

fredericDelaporte:
Master branch protection requires the PR to be up-to-date with master. This PR has not edits for maintainers enabled, so, we cannot update it ourselves. Can you update it?

src/NHibernate/Util/LinkHashMap.cs
```diff
@@ -0,0 +1,624 @@
...
+
+			public Entry Next { get; set; }
+
+			public Entry Prev { get; set; }
+
+			#region System.Object Members
+
+			public override int GetHashCode()
+			{
+				return Key == null ? 0 : Key.GetHashCode();
```
wilbit, line 45: I've removed using of `Value` from `GetHashCode`

fredericDelaporte (commented):
Master branch protection requires the PR to be up-to-date with master. This PR has not edits for maintainers enabled, so, we cannot update it ourselves. Can you update the PR branch or enable edits for maintainers?
If you update it without enabling edits for maintainers, please also regenerate async files, as the automated task for this fails without this permission.

wilbit:
rebased, @fredericDelaporte

fredericDelaporte (approved):

PR #3489, opened 2024-02-14
Inserting multiple associations of the same entity fails

mihaicodrean:
This works in 5.3.3, for example, but not in 5.5.0.
Thoughts as to why / how to fix?

Here's the failing test message:
```
NHibernate.Hql.Ast.ANTLR.QuerySyntaxException : A recognition error occurred. [insert into Enrolment (Course, Student) select e.Course, e.Student from Enrolment e]
  ----> Antlr.Runtime.MismatchedTreeNodeException : A recognition error occurred.
```
And the stack trace:
```
ErrorCounter.ThrowQueryException() line 71
BasicExecutor.ctor(IStatement statement, IQueryable persister) line 36
QueryTranslatorImpl.BuildAppropriateStatementExecutor(IStatement statement) line 478
QueryTranslatorImpl.DoCompile(IDictionary`2 replacements, Boolean shallow, String collectionRole) line 430
QueryTranslatorImpl.Compile(IDictionary`2 replacements, Boolean shallow) line 116
ASTQueryTranslatorFactory.CreateQueryTranslators(IQueryExpression queryExpression, IASTNode ast, String queryIdentifier, String collectionRole, Boolean shallow, IDictionary`2 filters, ISessionFactoryImplementor factory) line 53
ASTQueryTranslatorFactory.CreateQueryTranslators(IQueryExpression queryExpression, String collectionRole, Boolean shallow, IDictionary`2 filters, ISessionFactoryImplementor factory) line 21
QueryExpressionPlan.CreateTranslators(IQueryExpression queryExpression, String collectionRole, Boolean shallow, IDictionary`2 enabledFilters, ISessionFactoryImplementor factory) line 37
QueryExpressionPlan.ctor(IQueryExpression queryExpression, Boolean shallow, IDictionary`2 enabledFilters, ISessionFactoryImplementor factory) line 19
QueryPlanCache.GetHQLQueryPlan(IQueryExpression queryExpression, Boolean shallow, IDictionary`2 enabledFilters) line 76
AbstractSessionImpl.GetHQLQueryPlan(IQueryExpression queryExpression, Boolean shallow) line 584
AbstractSessionImpl.CreateQuery(String queryString) line 563
HqlBulkOperations.InsertFromSelectWithMultipleAssociations() line 46
--MismatchedTreeNodeException
TreeParser.RecoverFromMismatchedToken(IIntStream input, Int32 ttype, BitSet follow)
BaseRecognizer.Match(IIntStream input, Int32 ttype, BitSet follow)
SqlGenerator.selectClause() line 2099
```
And the standard output:
```
13:24:05,570 ERROR Parser:199 - MismatchedTreeNodeException(151!=3)
13:24:05,680 ERROR Parser:199 - MismatchedTreeNodeException(3!=49)
13:24:05,683 ERROR Parser:199 - MismatchedTreeNodeException(49!=3)
```

### DIFF
--- a/src/NHibernate.Test/BulkManipulation/HQLBulkOperations.cs
+++ b/src/NHibernate.Test/BulkManipulation/HQLBulkOperations.cs
@@ -33,7 +33,7 @@ namespace NHibernate.Test.BulkManipulation
 			}
 		}
 
-		[Test, KnownBug("#3489")]
+		[Test]
 		public void InsertFromSelectWithMultipleAssociations()
 		{
 			using var s = OpenSession();
--- a/src/NHibernate.Test/BulkManipulation/HQLBulkOperations.cs
+++ b/src/NHibernate.Test/BulkManipulation/HQLBulkOperations.cs
@@ -36,6 +36,9 @@ namespace NHibernate.Test.BulkManipulation
 		[Test]
 		public void InsertFromSelectWithMultipleAssociations()
 		{
+			Assume.That(TestDialect.NativeGeneratorSupportsBulkInsertion,
+				"The dialect does not support a native generator compatible with bulk insertion.");
+
 			using var s = OpenSession();
 			using var tx = s.BeginTransaction();
 
--- a/src/NHibernate.Test/BulkManipulation/HQLBulkOperations.cs
+++ b/src/NHibernate.Test/BulkManipulation/HQLBulkOperations.cs
@@ -32,5 +32,18 @@ namespace NHibernate.Test.BulkManipulation
 				tx.Commit();
 			}
 		}
+
+		[Test, KnownBug("#3489")]
+		public void InsertFromSelectWithMultipleAssociations()
+		{
+			using var s = OpenSession();
+			using var tx = s.BeginTransaction();
+
+			s.CreateQuery("insert into Enrolment (Course, Student)" +
+				" select e.Course, e.Student from Enrolment e")
+				.ExecuteUpdate();
+
+			tx.Commit();
+		}
 	}
 }
--- a/src/NHibernate/Hql/Ast/ANTLR/Tree/SelectClause.cs
+++ b/src/NHibernate/Hql/Ast/ANTLR/Tree/SelectClause.cs
@@ -443,23 +443,22 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 			IList<FromElement> fetchedFromElements)
 		{
 			var appender = new ASTAppender(ASTFactory, this);
-			var combinedFromElements = new List<FromElement>();
 			var processedElements = new HashSet<FromElement>();
-			RenderNonScalarIdentifiers(appender, processedElements, combinedFromElements, inheritedExpressions);
+			RenderNonScalarIdentifiers(appender, processedElements, inheritedExpressions);
 			if (Walker.IsShallowQuery)
 			{
 				return;
 			}
 
 			// Append fetched elements
-			RenderFetchedNonScalarIdentifiers(appender, fetchedFromElements, processedElements, combinedFromElements);
+			RenderFetchedNonScalarIdentifiers(appender, fetchedFromElements, processedElements);
 			if (currentFromClause.IsScalarSubQuery)
 			{
 				return;
 			}
 			
 			// Generate the property select tokens.
-			foreach (var fromElement in combinedFromElements)
+			foreach (var fromElement in processedElements)
 			{
 				RenderNonScalarProperties(appender, fromElement);
 			}
@@ -482,7 +481,6 @@ namespace NHibernate.Hql.Ast.ANTLR.Tree
 		private void RenderNonScalarIdentifiers(
 			ASTAppender appender,
 			HashSet<FromElement> processedElements,
-			List<FromElement> combinedFromElements,
 			Dictionary<ISelectExpression, SelectClause> inheritedExpressions)
 		{
 			foreach (var e in NonScalarExpressions)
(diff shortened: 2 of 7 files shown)


mihaicodrean:
I have narrowed it down to this commit: 10393dfbabd7bf2f9e191ce5113621effc58bc2c for #2551.

@maca88 & @bahusoid, can I please bring your attention to this, maybe you have an idea on how to address it?

src/NHibernate.Test/BulkManipulation/HQLBulkOperations.cs
```diff
@@ -3,7 +3,7 @@
 namespace NHibernate.Test.BulkManipulation
 {
 	[TestFixture]
-	public class HqlBulkOperations: BaseFixture
+	public class HqlBulkOperations: TestCase
```
fredericDelaporte, line 6: This change is breaking this existing test case. Please revert.

src/NHibernate.Test/BulkManipulation/HQLBulkOperations.cs
```diff
@@ -32,5 +32,23 @@ public void SimpleDelete()
 				tx.Commit();
 			}
 		}
+
+		protected override string MappingsAssembly => "NHibernate.Test";
+
+		protected override string[] Mappings => new[] { "BulkManipulation.Enrolment.hbm.xml" };
```
fredericDelaporte, line 38: The BaseFixture here will include all mappings set as embedded resources in its directory. There is no need for these overrides.

mihaicodrean: Done, thanks for clarifying. I got an error for duplicate mappings after adding those (now reverted), and switching to TestCase (now reverted as well) addressed that.

mihaicodrean:
@fredericDelaporte, I see that the "NHibernate (NHibernate Core)" check step in TeamCity is now failing as expected, for the contributed test. Is the merge just not possible in this case, e.g. can't really contribute failing tests?

hazzik:
Need to mark the test with [KnownBug] attribute

fredericDelaporte:
Squashed and rebased on 5.4.x (11984a06), it does fail too.
The same on the commit prior to #2551 succeeds, see 491e6da. So, it is confirmed it is a 5.4 regression.

We should fix it on 5.4.x. As such, the test case needs to be rebased on 5.4.x and the PR needs to target the 5.4.x branch.

For fixing it, I think the troubles lies within `SelectClause`, but I am not knowledgeable about that code.

fredericDelaporte:
I have narrowed it to [this change](https://github.com/nhibernate/nhibernate-core/pull/2551/files#r1527598112), but have not yet devise a proper fix.

fredericDelaporte:
I have now overridden your branch in order to target 5.4.x for a potential fix I have added.

mihaicodrean:
> In debug, forcing the execution to jump directly to RenderNonScalarIdentifiers (so, not adding again the from to combinedFromElements) allows the test to succeed.

Thanks much for identifying the root cause!

src/NHibernate/Hql/Ast/ANTLR/Tree/SelectClause.cs
```diff
@@ -493,10 +493,14 @@ private void InitAliases(List<ISelectExpression> selectExpressions)
...
-				var node = (IASTNode) e;
-				if (processedElements.Add(fromElement))
+				var isNewFrom = processedElements.Add(fromElement);
+				if (isNewFrom)
 				{
 					combinedFromElements.Add(fromElement);
+				}
+				var node = (IASTNode) e;
+				if (isNewFrom || node.Type == SqlGenerator.DOT)
+				{
```
fredericDelaporte, line 503: Though "it works", I do not think this is correct. I need to analyze more what does the code with `combinedFromElements`, but it sounds likely to me it should be renamed to something like `combinedEntityFromElements` and fetched only with non-shallow `fromElement`, used to select the whole entity, not just its id.

src/NHibernate/Hql/Ast/ANTLR/Tree/SelectClause.cs
```diff
@@ -494,7 +494,7 @@ private void InitAliases(List<ISelectExpression> selectExpressions)
 				}
 
 				var node = (IASTNode) e;
-				if (processedElements.Add(fromElement))
+				if (Walker.IsShallowQuery && node.Type == SqlGenerator.DOT || processedElements.Add(fromElement))
```
fredericDelaporte, line 497: The `processedElements` set serves a de-duplication purpose of selected entities. But, alone, it fails accounting for associations selected in shallow queries which will be seen as having the same fromElement although being actually different entities being selected.

Shallow queries may still have duplicated entities selected and cannot be entirely ignored for the de-duplication, if we want to keep the test `HqlDuplicateEntitySelectionSubQuery` as currently is. (I fail to understand why this tests is valid, why should we support the queries in this test, but I did not get answers about this [here](https://github.com/nhibernate/nhibernate-core/pull/2551#pullrequestreview-1941571476). And anyway, it has been release in 5.4, so, it is legacy.)

The `DOT` check allows to recognize cases in shallow queries where the same fromElement may be used by different expressions selecting different things, and a such, needing to be not falsely de-duplicated.

(I thought about checking if the expressions `e` where "the same" (having same Path) instead when sharing a common fromElement, but that would imply building a dictionary of sets instead of `processedElements`. I do not consider it is worth it.)

fredericDelaporte:
Hazzik and Maca, since the fix is actually from me, I will not approve this PR myself.

src/NHibernate/Hql/Ast/ANTLR/Tree/SelectClause.cs
maca88: > I fail to understand why this tests is valid, why should we support the queries in this test

To be honest, I don't remember why I've added such test. Looking back at it, I agree that the test shouldn't be added.

Perhaps we could add a TODO for removing test `HqlDuplicateEntitySelectionSubQuery` in version `5.5` and remove the `node.Type == SqlGenerator.DOT` part from the condition.

maca88 (approved):
The fix looks good to me.

hazzik (approved):

mihaicodrean:
> @mihaicodrean, may you check this dev package solves the issue for your application, and that none other regression remains?

Thanks for the fix! I have applied the patch to my local 5.5.0 branch, and all works as expected for me.

PR #3492, opened 2024-02-26
Add NUnit analyzers

hazzik:
*  Add NUnit.Analyzers 
* Fix NUnit2007: The actual value should not be a constant 
* Fix NUnit2041: The comparison constraint always fails as the actual and the expected value are not comparable
* Fix NUnit2022: The type of the actual argument - 'IEnumerable<>' - has no property 'Count'
* Fix NUnit2023: The type of the actual argument - 'bool' - can never be null
* Fix other issues

### DIFF
--- a/src/NHibernate.Test/SqlTest/Custom/CustomSQLSupportTest.cs
+++ b/src/NHibernate.Test/SqlTest/Custom/CustomSQLSupportTest.cs
@@ -48,10 +48,10 @@ namespace NHibernate.Test.SqlTest.Custom
 			s = OpenSession();
 			t = s.BeginTransaction();
 			jboss = (Organization)s.Get(typeof(Organization), orgId);
-			Assert.AreEqual(jboss.Employments.Count, 2);
+			Assert.AreEqual(2, jboss.Employments.Count);
 			emp = (Employment)GetFirstItem(jboss.Employments);
 			gavin = emp.Employee;
-			Assert.AreEqual(gavin.Name, "GAVIN");
+			Assert.AreEqual("GAVIN", gavin.Name);
 			Assert.AreEqual(s.GetCurrentLockMode(gavin), LockMode.Upgrade);
 			emp.EndDate = DateTime.Today;
 			Employment emp3 = new Employment(gavin, jboss, "US");
@@ -64,7 +64,7 @@ namespace NHibernate.Test.SqlTest.Custom
 			IEnumerator iter = s.GetNamedQuery("allOrganizationsWithEmployees").List().GetEnumerator();
 			Assert.IsTrue(iter.MoveNext());
 			Organization o = (Organization)iter.Current;
-			Assert.AreEqual(o.Employments.Count, 3);
+			Assert.AreEqual(3, o.Employments.Count);
 
 			foreach (Employment e in o.Employments)
 			{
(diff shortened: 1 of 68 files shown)


src/NHibernate.Test/NHSpecificTest/NH901/Fixture.cs
```diff
@@ -81,7 +81,11 @@ public void ReplaceValueTypeComponentWithSameValueDoesNotMakeItDirty()
 				a.Street = "Some Street";
 				a.City = "Some City";
 				jimmy.Address = a;
-				Assert.AreNotSame(jimmy.Address, a);
+#pragma warning disable NUnit2040
+				// NUnit2040 : The SameAs constraint always fails on value types as the actual and the expected value cannot be the same reference
+				// TODO: Find out what was the intention of the test. Address is a struct.
+				Assert.That(a, Is.Not.SameAs(jimmy.Address));
+#pragma warning restore NUnit2040
```
fredericDelaporte, line 88: I think this can be removed. It would make more sense to check the structs are equal before assigning the new a to Jimmy's address.

src/NHibernate.Test/SqlTest/Custom/CustomSQLSupportTest.cs
```diff
@@ -64,7 +65,13 @@ public void HandSQL()
...
+
+			Assert.AreEqual(3, o.Employments.Count);
+
+			foreach (Employment e in o.Employments)
+			{
+				s.Delete(e);
+			}
 
 			foreach (Employment e in o.Employments)
 			{
```
fredericDelaporte, line 77: Failed copy paste? (Why the duplication of these lines of code?)

src/NHibernate.Test/TypeParameters/TypeParameterTest.cs
```diff
@@ -93,17 +94,18 @@ public void Loading()
...
 			Widget obj = (Widget) s.CreateQuery("from Widget o where o.Str = :string")
 			                      	.SetString("string", "all-normal").UniqueResult();
-			Assert.AreEqual(obj.ValueOne, 7, "Non-Default value incorrectly loaded");
-			Assert.AreEqual(obj.ValueTwo, 8, "Non-Default value incorrectly loaded");
-			Assert.AreEqual(obj.ValueThree, 9, "Non-Default value incorrectly loaded");
-			Assert.AreEqual(obj.ValueFour, 10, "Non-Default value incorrectly loaded");
+			Assert.AreEqual(7, obj.ValueOne, "Non-Default value incorrectly loaded");
+			Assert.AreEqual(8, obj.ValueTwo, "Non-Default value incorrectly loaded");
+			Assert.AreEqual(9, obj.ValueThree, "Non-Default value incorrectly loaded");
+
```
fredericDelaporte, line 100:
```suggestion
```

src/NHibernate.Test/SqlTest/Custom/CustomSQLSupportTest.cs
hazzik: Tool failed probably (this was done by a tool)

src/NHibernate.Test/NHSpecificTest/NH901/Fixture.cs
hazzik: I wanted to do minimal changes. I think I rather disable this warning in editorconfig file.

src/NHibernate.Test/TypeParameters/TypeParameterTest.cs
hazzik: I'll redo the commits so it is done by the tool rather than with (occasional) manual fixes.

fredericDelaporte (approved):

PR #3495, opened 2024-03-01
Improves the precision of the GUID Comb generator.

ashleyhuxley:
By dividing by 3.333333 but casting to a long, the apparent increase in precision was lost due to truncation. The improved code creates a 4 byte number representing the time of day at the maximum possible precision.


src/NHibernate/Id/GuidCombGenerator.cs
```diff
@@ -48,20 +50,19 @@ private Guid GenerateComb()
 
 			DateTime now = DateTime.UtcNow;
 
-			// Get the days and milliseconds which will be used to build the byte string 
-			TimeSpan days = new TimeSpan(now.Ticks - BaseDateTicks);
-			TimeSpan msecs = now.TimeOfDay;
+			// We use 2 bytes for the day (65,535 possible values) so we're good until 6/6/2079.
```
fredericDelaporte, line 53: Why would we be no more "good" after that? The days will have looped, and then what? There are 10 other bytes which remains mostly random. What trouble do you expect?

src/NHibernate/Id/GuidCombGenerator.cs
```diff
@@ -48,20 +50,19 @@ private Guid GenerateComb()
...
+			var days = new TimeSpan(now.Ticks - BaseDateTicks);
+			var daysArray = BitConverter.GetBytes(days.Days);
 
-			// Convert to a byte array 
-			// Note that SQL Server is accurate to 1/300th of a millisecond so we divide by 3.333333
```
fredericDelaporte, line 56: The aim of this is to emulate the SQL Guid Comb algorithm in order to have NHibernate Guid.Comb behaving like its SQL Server expression, `CAST(CAST(NEWID() AS BINARY(10)) 
+ CAST(GETDATE() AS BINARY(6)) AS UNIQUEIDENTIFIER)`, see the [article](https://www.informit.com/articles/article.aspx?p=25862&seqNum=7) about this. Your change break this and will cause issues to applications expecting NHibernate Guid.Comb to be equivalent to this SQL computing.

fredericDelaporte (commented):
I do no think we should change the existing Guid.Comb, meant to match a SQL Server equivalent expression.

Instead, a new generator may be proposed for optimizing that last four bytes part, but does it solve any actual issue? (Why is it worth it?)

And preferably, it should have an equivalent SQL expression allowing to generate GUID with the same "sequential" properties on the database side. Meaning GUID generated DB side at the same time with that equivalent SQL expression will have the same last six bytes than those generated with the new Guid algorithm, like Guid.Comb allows.

hazzik:
Closing as per @fredericDelaporte

PR #3502, opened 2024-03-15
Added NET5_0_OR_GREATER in every place where NETSTANDARD2_1_OR_GREATER was used

gliljas:


src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs
```diff
@@ -1,7 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-#if NETCOREAPP2_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+#if NETCOREAPP2_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER || NET5_0_OR_GREATER
```
fredericDelaporte, line 4: Am I wrong thinking `NETCOREAPP2_0_OR_GREATER` already implies `NET5_0_OR_GREATER`, causing this change and most other changes to be unneeded?

src/NHibernate/Driver/ReflectionBasedDriver.cs
```diff
@@ -46,12 +46,12 @@ protected ReflectionBasedDriver(string providerInvariantName, string driverAssem
...
-#if NETFX || NETSTANDARD2_1_OR_GREATER
+#if NETFX || NETSTANDARD2_1_OR_GREATER || NET5_0_OR_GREATER 
 				if (string.IsNullOrEmpty(providerInvariantName))
 				{
 #endif
 					throw new HibernateException(string.Format(ReflectionTypedProviderExceptionMessageTemplate, driverAssemblyName));
-#if NETFX || NETSTANDARD2_1_OR_GREATER
+#if NETFX || NETSTANDARD2_1_OR_GREATER || NET5_0_OR_GREATER
 				}
 				var factory = DbProviderFactories.GetFactory(providerInvariantName);
```
fredericDelaporte, line 56: The reason for this condition was that `DbProviderFactories`was not available in netcoreapp2.0.

Since it is available in NetStandard 2.1, it must be there too in Net Core 2.1, and the precise condition should then be:

```suggestion
#if NETFX || NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER
				}
				var factory = DbProviderFactories.GetFactory(providerInvariantName);
```

Or if it is preferred to write conditions only relative to our actual targets:

```suggestion
#if NETFX || NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER
				}
				var factory = DbProviderFactories.GetFactory(providerInvariantName);
```

src/NHibernate/Driver/ReflectionBasedDriver.cs
```diff
@@ -46,12 +46,12 @@ protected ReflectionBasedDriver(string providerInvariantName, string driverAssem
 
 			if (connectionType == null || commandType == null)
 			{
-#if NETFX || NETSTANDARD2_1_OR_GREATER
+#if NETFX || NETSTANDARD2_1_OR_GREATER || NET5_0_OR_GREATER
```
fredericDelaporte, line 49: Then as a consequence of the comment on below lines :

```suggestion
#if NETFX || NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER
```

or 

```suggestion
#if NETFX || NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER
```

src/NHibernate/Driver/ReflectionBasedDriver.cs
hazzik: > Since it is available in NetStandard 2.1, it must be there too in Net Core 2.1

No, the first version to implement net standard is .NET Core 3.0: https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1

src/NHibernate/Util/CollectionHelper.cs
```diff
@@ -733,7 +733,7 @@ public static bool BagEquals<T>(IEnumerable<T> c1, IEnumerable<T> c2, IEqualityC
 				(comparer == null ? DictionaryEquals(m1, m2, EqualityComparer<V>.Default) :
 					m1.All(kv => m2.TryGetValue(kv.Key, out var value) && comparer.Equals(kv.Value, value)));
 
-#if !NETCOREAPP2_0_OR_GREATER && !NETSTANDARD2_1_OR_GREATER
+#if !NETCOREAPP2_0_OR_GREATER && !NETSTANDARD2_1_OR_GREATER && !NET5_0_OR_GREATER
```
hazzik, line 736: Don't need it here.

PR #3503, opened 2024-03-19
Add ToDictionaryAsync and ToHashSetAsync Linq extension methods

dennis-gr:
Convenience methods to avoid the slightly awkward syntax when doing it manually after calling `ToListAsync`.

Async-Generator fails to generate tests for ToDictionaryAsync, not sure why it doesn't get recognized.

### DIFF
--- a/src/NHibernate/Linq/LinqExtensionMethods.cs
+++ b/src/NHibernate/Linq/LinqExtensionMethods.cs
@@ -2405,6 +2405,9 @@ namespace NHibernate.Linq
 		/// <summary>
 		/// Executes the query and returns its result as a <see cref="HashSet{T}"/>.
 		/// </summary>
+		/// <remarks>
+		/// Note that this method still allocates an intermediate <see cref="List{T}"/> instance from which the result is created.
+		/// </remarks>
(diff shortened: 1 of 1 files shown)


fredericDelaporte:
What is the awkward? `(await query.ToListAsync()).ToHashSet()`? Or `hashSet.UnionWith(await query.ToListAsync())`?

I do not see the need.

fredericDelaporte:
I see some harm:
 - It may lure users into thinking the thing is optimized in order to directly load the result into a hashset/dictionary instead of going through an intermediate list.
 - That is still some additional code to maintain, just for sparing a couple of parenthesis to callers.
 
 Other maintainers may see it otherwise, so, It leave it open.

dennis-gr:
> I see some harm:
> 
>     * It may lure users into thinking the thing is optimized in order to directly load the result into a hashset/dictionary instead of going through an intermediate list.

Optimizing this would be possible as soon as https://github.com/nhibernate/nhibernate-core/pull/2274 is done. I'll add a remark that calling these methods still creates an intermediate list.

>     * That is still some additional code to maintain, just for sparing a couple of parenthesis to callers.

Like I said, ATM it's a small usability improvement that I think is worth it, with the option to provide a slightly optimized version in the future. EF Core has these methods and it would be nice if NH did too.

Issue #3516, opened 2024-04-02
ObjectToSQLString implementations are problematic


gliljas:
While working with DateOnly and TimeOnly support I realized that the ObjectToSQLString method of ILiteralType is quite problematic. 

**AbstractStringType**:
`"'" + (string)value + "'"`

- SQL injection
- No N-prefix for unicode (required in SQL server when the string is outside ANSI)

**AbstractCharType**:
`'\'' + value.ToString() + '\''`

- SQL injection
- No N-prefix for unicode (required in SQL server when the string is outside ANSI)

**AbstractDateTimeType**:
`"'" + (DateTime) value + "'"`

- No guarantee that the stringified DateTime will be a supported datetime syntax

**DateTimeOffsetType**:
`"'" + ((DateTimeOffset) value) + "'"`

- No guarantee that the stringified DateTimeOffset will be a supported datetimeoffset syntax

**DateType**:
`"\'" + ((DateTime)value).ToShortDateString() + "\'"`

- No guarantee that the stringified DateTime will be a supported datetime syntax

**DecimalType**:
`value.ToString()`

- Could very well be a decimal with the wrong kind of decimal separator

And there are more. 

Since the method has access to the dialect, maybe more formatting should be moved there.

fredericDelaporte:
From my viewpoint, this method should be obsoleted and no more used. Building string fragments of literal values is a bad practice. Parameters should be used instead.

I can find three usages:
 - To handle discriminators. I have neither investigated why does-it not use parameters, nor how much work would it involve.
 - A seemingly unused `AddColumn` overload on `SqlInsertBuilder` and `SqlUpdateBuilder`: why not obsoleting these methods?

gliljas:
Using literal values in SQL does have some value. E.g. filtered indexes rely on the filtered columns **not** being supplied as parameters. And filtered indexes are really useful, e.g in scenarios where discriminators are being used. It makes it easy to create indexes tailor made for the sub classes in a class hierarchy. We use that functionality, and I'm certain others do as well.

We could probably "re-enable" such things with query rewriting if the literals were removed, but it's something to take into account. Making them into parameters is likely to cause performance degradation for some users.

fredericDelaporte:
[...]

PR #3517, opened 2024-04-02
Obsolete literal AddColumn

fredericDelaporte:
SqlInsert/UpdateBuilder AddColumn overloads taking a value have a SQL injection vulnerability, and have no usage.

### DIFF
--- a/src/NHibernate.Test/SqlCommandTest/SqlInsertBuilderFixture.cs
+++ b/src/NHibernate.Test/SqlCommandTest/SqlInsertBuilderFixture.cs
@@ -26,7 +26,9 @@ namespace NHibernate.Test.SqlCommandTest
 
 			insert.AddColumn("intColumn", NHibernateUtil.Int32);
 			insert.AddColumn("longColumn", NHibernateUtil.Int64);
+#pragma warning disable CS0618 // Type or member is obsolete
 			insert.AddColumn("literalColumn", false, (ILiteralType) NHibernateUtil.Boolean);
+#pragma warning restore CS0618 // Type or member is obsolete
 			insert.AddColumn("stringColumn", 5.ToString());
 
 			SqlCommandInfo sqlCommand = insert.ToSqlCommandInfo();
@@ -53,7 +55,9 @@ namespace NHibernate.Test.SqlCommandTest
 
 			insert.SetTableName("test_insert_builder");
 
+#pragma warning disable CS0618 // Type or member is obsolete
 			insert.AddColumn("stringColumn", "aSQLValue", (ILiteralType)NHibernateUtil.String);
+#pragma warning restore CS0618 // Type or member is obsolete
 			insert.SetComment("Test insert");
 			string expectedSql =
 	"/* Test insert */ INSERT INTO test_insert_builder (stringColumn) VALUES ('aSQLValue')";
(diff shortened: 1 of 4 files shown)


src/NHibernate/SqlCommand/SqlInsertBuilder.cs
```diff
@@ -66,6 +66,8 @@ public virtual SqlInsertBuilder AddColumn(string columnName, IType propertyType)
 		/// <param name="val">The value to set for the column.</param>
 		/// <param name="literalType">The NHibernateType to use to convert the value to a sql string.</param>
 		/// <returns>The SqlInsertBuilder.</returns>
+		// Since v5.6
+		[Obsolete("This method is unsafe and has no more usages. Use the overload with a property type and use a parameterized query.")]
 		public SqlInsertBuilder AddColumn(string columnName, object val, ILiteralType literalType)
 		{
 			return AddColumn(columnName, literalType.ObjectToSQLString(val, Dialect));
```
fredericDelaporte, line 73: As noted by @gliljas in #3516, these methods implementations are unsafe and may lead to SQL injections or invalid SQL. Since these `AddColumn` overloads have no usage, better just obsolete them. Adding a value in a SQL query as a string fragment is anyway a bad practice. Parameterization should be used instead.

src/NHibernate.Test/SqlCommandTest/SqlInsertBuilderFixture.cs
```diff
@@ -26,15 +25,13 @@ public void InsertSqlStringTest()
...
-			insert.AddColumn("literalColumn", false, (ILiteralType) NHibernateUtil.Boolean);
 			insert.AddColumn("stringColumn", 5.ToString());
 
 			SqlCommandInfo sqlCommand = insert.ToSqlCommandInfo();
 			SqlType[] actualParameterTypes = sqlCommand.ParameterTypes;
 
-		    string falseString = factoryImpl.Dialect.ToBooleanValueString(false);
 			string expectedSql =
-                "INSERT INTO test_insert_builder (intColumn, longColumn, literalColumn, stringColumn) VALUES (?, ?, " + falseString + ", 5)";
+				"INSERT INTO test_insert_builder (intColumn, longColumn, stringColumn) VALUES (?, ?, 5)";
```
hazzik, line 34: I think we should not disable the tests. Just mark portions as obsolete for now.

fredericDelaporte:
If we do that for mitigating some of the issues raised in #3516, it should target the currently released minor instead, I realized. Even if obsoleting members in a patch release is not a normal practice with SemVer, when that is a security issue mitigation, I seems reasonable.

If instead the trouble is fixed at its root in the minor release, then we could keep on targeting the next minor, with another obsolete message.

hazzik:
Let's do 5.4.x and 5.5.x

hazzik (approved):

src/NHibernate/SqlCommand/SqlInsertBuilder.cs
leifjones: I'm curious why not remove it?

leifjones: Thanks!

Issue #3500, opened 2024-03-08
RemoveUnusedCommandParameters() removing needed parameters


tskong:
Hi,

I'm running this linq statement, but it's erroring as not all the SQL parameters are being presented to SQL server, I have been debugging through the nhibernate source and the culprit is partially RemoveUnusedCommandParameters(), but I could do with some help in diagnosing it.

The C# code uses listContactGroupId twice within the linq statement, and that seems to confuse things.

```
listContactGroupId = new List<int>() { 7, 14, 21, 28 };
var res = dbObject.Count(x => (x.EnquiryContactGroupId.HasValue && 
listContactGroupId.Contains(x.EnquiryContactGroupId.Value)) || (x.ApplicationContactGroupId.HasValue && listContactGroupId.Contains(x.ApplicationContactGroupId.Value)));
```
The SQL Looks like this

```
select (count(*)) as col_0_0_ from EnquiryApplicationContactDetail_V venquiryap0_ where (venquiryap0_.EnquiryContactGroupId is not null) and (venquiryap0_.EnquiryContactGroupId in (@p0 , @p1 , @p2 , @p3)) or (venquiryap0_.ApplicationContactGroupId is not null) and (venquiryap0_.ApplicationContactGroupId in (@p4 , @p5 , @p6 , @p7))
```

But when presented to the SQL server, parameter values @p4->@P7 are missing.

The parameters go missing at the method RemoveUnusedCommandParameters(), there is a line that removes the duplicates

`var assignedParameterNames = new HashSet<string>(formatter.AssignedParameterNames); 
`
The formatter.AssignedParmetersNames, rather than being @p0, @p1, @p2, @p3,@P4, @p5, @p6, @P7 it's @p0, @p1, @p2, @p3, @p0, @p1, @p2, @p3, so it ends up with @p0, @p1, @p2, @p3

It's got something to do with nhibernate.sqlcommand.parameter.ParameterPosition which assigns the @p[n] parameter name.

Can anyone shed any light on how it generates the parameter names?

Thanks,

tskong:
I have a bit more information. It appears that "backtrack" on the parameters has something to do with it, 0-3 and 4-7 have the same backtrack id, so it think 4-7 are the same parameters and then it discards them, even though they are needed.  What is this backtrack for, and can I inhibit it?

[...]

PR #3520, opened 2024-04-04
Fix for #3500

tskong:
Please see https://github.com/nhibernate/nhibernate-core/issues/3500 for issue

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/NH3004/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH3004/Fixture.cs
@@ -1,5 +1,6 @@
-﻿using System.Collections;
+using System.Collections;
 using NHibernate.SqlCommand;
+using NHibernate.Test.TypesTest;
 using NUnit.Framework;
 
 namespace NHibernate.Test.NHSpecificTest.NH3004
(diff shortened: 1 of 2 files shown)


hazzik:
Just heads up for reviewers: I cannot reproduce the issue, so I believe this fix is not necessary/not correct.

tskong:
The unit test from hazzik (checking SqlLogSpy) is different from my unit test, I am checking the command.text sent to the SQL server, 

This issue definitely caused my code to have the problem which has now gone away.  I've checked the other unit tests and they are still passing, so this fix does no harm.

tskong:
I've run through the test that @hazzik had, and the result is different from mine, but also what they are testing for.  My test is using the SqlDriver, and the hazzik unit test is using SQlLogSpy.  I may have uncovered a new bug, that wasn't picked up in the previous tests.

Please can it be reviewed.

**Before my fix
command.commandText**
SELECT * FROM Test where value = @p0 or value=@p1

**SQLLogSpy:
NHibernate.SQL** 
    select
        (count(*)) as col_0_0_ 
    from
        Entity entity0_ 
    where
        (
            entity0_.GroupId1 is not null
        ) 
        and (
            entity0_.GroupId1 in (
                @p0 , @p1 , @p2 , @p3
            )
        ) 
        or (
            entity0_.GroupId2 is not null
        ) 
        and (
            entity0_.GroupId2 in (
                @p0 , @p1 , @p2 , @p3
            )
        );
    @p0 = 7 [Type: Int32 (0:0:0)], @p1 = 14 [Type: Int32 (0:0:0)], @p2 = 21 [Type: Int32 (0:0:0)], @p3 = 28 [Type: Int32 (0:0:0)]

**After my fix
command.commandText**
SELECT * FROM Test where value = @p0 or value=@p0

**SqlLogSpy:**
 select
        (count(*)) as col_0_0_ 
    from
        Entity entity0_ 
    where
        (
            entity0_.GroupId1 is not null
        ) 
        and (
            entity0_.GroupId1 in (
                @p0 , @p1 , @p2 , @p3
            )
        ) 
        or (
            entity0_.GroupId2 is not null
        ) 
        and (
            entity0_.GroupId2 in (
                @p0 , @p1 , @p2 , @p3
            )
        );
    @p0 = 7 [Type: Int32 (0:0:0)], @p1 = 14 [Type: Int32 (0:0:0)], @p2 = 21 [Type: Int32 (0:0:0)], @p3 = 28 [Type: Int32 (0:0:0)]

tskong:
@hazzik After some further investigation, I found a castle core interceptor that was firing, this was doing something to the NHibernate.SqlCommand.SqlString which was upsetting the parameter processing.

I am going to cancel this PR, thanks for questioning the PR, and making me find the proper source of the problem 👍

hazzik:
Glad you've found the issue @tskong

PR #3524, opened 2024-04-08
Fix setting ScaleDefined in SqlType

### LINKED ISSUE #3523: ScaleDefined is not set by SqlType ctor
```
public SqlType(DbType dbType, byte precision, byte scale)
{
	this.dbType = dbType;
	this.precision = precision;
	this.scale = scale;
	precisionDefined = true;
}

public SqlType(DbType dbType, byte scale)
{
	this.dbType = dbType;
	this.scale = scale;
	ScaleDefined = true;
}
```

As you can see, ScaleDefined is not set by the upper ctor, even though it's most definitely set.

gliljas:
Fixes #3523

### DIFF
--- a/src/NHibernate/SqlTypes/SqlType.cs
+++ b/src/NHibernate/SqlTypes/SqlType.cs
@@ -23,69 +23,46 @@ namespace NHibernate.SqlTypes
 	[Serializable]
 	public class SqlType : IEquatable<SqlType>
 	{
-		private readonly DbType dbType;
-		private readonly int length;
-		private readonly byte precision;
-		private readonly byte scale;
-		private readonly bool lengthDefined;
-		private readonly bool precisionDefined;
-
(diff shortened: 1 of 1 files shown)


hazzik:
I personally hate the concept of `xxxDefined`. This is old pre-nullable way of doing things. I think it should be done by `Nullable<X>` instead:

```
private int? length;
public int Length => length.GetValueOrDefault();
public bool LengthDefined => length.HasValue;
```

gliljas:
> I personally hate the concept of `xxxDefined`. This is old pre-nullable way of doing things and I think it should be done by `Nullable<X>` now instead:
> 
> ```
> private int? length;
> public int Length => length.GetValueOrDefault();
> public bool LengthDefined => length.HasValue;
> ```

Done

hazzik (approved):

Issue #3525, opened 2024-04-12
Querying Seconds on DateTime(Offset) can cause issues in e.g PostgreSQL


gliljas:
Again, something I stumbled upon when working with TimeOnly. Everything worked fine, until I ran the tests on PostgreSQL.

The issue is that the `seconds` HQL method yields an `extract(second from ....)` for PostgreSQL, but that SQL function returns a double, with the fractional seconds included. In other words, querying for `where seconds(prop)=3`  or `.Where(x => x.Seconds == 3)` will not match if prop actually contained `3.001`.  Just reading the value works fine, since Int32Type truncates the value.

Apparently, the HQL methods  `seconds` and  `extract(second from ....)` are explicitly defined in HQL (Java) to return the seconds with fractional precision, so they instead make sure that other dialects conform to this. E.g the SQL Server dialect defines `seconds` as `(datepart(second,?2)+datepart(nanosecond,?2)/1e9)`

I'm not sure Nhibernate's HQL has to align with Hibernate's, so an option would be to ensure that e.g the Postgres dialect aligns with SQL Server et. al. and redefines `seconds` as `extract(seconds from date_trunc('seconds', value))`. Another option would be to add a new HQL function and use that from `DateTimePropertiesHqlGenerator`.

fredericDelaporte:
Fixed by #3528/

PR #3528, opened 2024-04-19
Fix Linq .Second bug for PostgreSQL et al.

gliljas:
Fix for #3525

The actual fix was adding a new HQL function "secondtruncated", which is used if it is available, otherwise it falls back to the old "second". Another option would be to fix the "second" function in the dialects where it doesn't work, but that could be considered a more breaking change. As mentioned in the issue, Hibernate has defined "second" to actually include fractional seconds, and have adjusted the dialects where the corresponding SQL function truncates.

The most changes were in DateTimeTests.cs. I really wanted to use that test class since it already tests exactly this (but not seconds), but to get fractional seconds into the Northwind dataset was a bit icky, so I opted to create specific test data.

I used the approach of comparing the result from the LINQ query with the result from the same query run in memory. Of course that doesn't work in scenarios with null, case sensitivity etc., but here it works just fine.

The MySQL test fails on Github, for some reason. I can't make it fail locally.

### DIFF
--- a/src/NHibernate/Dialect/Function/SQLFunctionRegistry.cs
+++ b/src/NHibernate/Dialect/Function/SQLFunctionRegistry.cs
@@ -22,13 +22,10 @@ namespace NHibernate.Dialect.Function
 		/// </summary>
 		public ISQLFunction FindSQLFunction(string functionName)
 		{
-			if (!userFunctions.ContainsKey(functionName) && !dialect.Functions.ContainsKey(functionName) && _functionAliases.TryGetValue(functionName, out var sqlFunction))
+			if (!userFunctions.TryGetValue(functionName, out ISQLFunction result) && !dialect.Functions.TryGetValue(functionName, out result))
 			{
-				functionName = sqlFunction;
-			}
-			if (!userFunctions.TryGetValue(functionName, out ISQLFunction result))
-			{
-				dialect.Functions.TryGetValue(functionName, out result);
+				if (_functionAliases.TryGetValue(functionName, out var sqlFunction) && !_functionAliases.ContainsKey(sqlFunction))
+					return FindSQLFunction(sqlFunction);
 			}
 			return result;
 		}
--- a/src/NHibernate/Dialect/Function/SQLFunctionRegistry.cs
+++ b/src/NHibernate/Dialect/Function/SQLFunctionRegistry.cs
@@ -7,8 +7,8 @@ namespace NHibernate.Dialect.Function
 	{
 		private readonly Dialect dialect;
 		private readonly IDictionary<string, ISQLFunction> userFunctions;
-		//Temporary alias support
-		private static Dictionary<string, string> _functionAliases = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase) { { "secondtruncated", "second" } };
+		// Temporary alias support
+		private readonly static Dictionary<string, string> _functionAliases = new(StringComparer.OrdinalIgnoreCase) { { "secondtruncated", "second" } };
 
 		public SQLFunctionRegistry(Dialect dialect, IDictionary<string, ISQLFunction> userFunctions)
 		{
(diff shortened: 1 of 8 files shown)


src/NHibernate/Linq/Functions/DateTimePropertiesHqlGenerator.cs
```diff
@@ -33,8 +34,13 @@ public DateTimePropertiesHqlGenerator()
 
 		public override HqlTreeNode BuildHql(MemberInfo member, Expression expression, HqlTreeBuilder treeBuilder, IHqlExpressionVisitor visitor)
 		{
-			return treeBuilder.MethodCall(member.Name.ToLowerInvariant(),
+			var functionName = member.Name.ToLowerInvariant();
+			if (functionName == "second" && (visitor.SessionFactory as ISessionFactoryImplementor)?.Dialect.Functions.ContainsKey("secondtruncated") == true)
```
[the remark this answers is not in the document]
gliljas: I would love to, but the only meaningful way to do that, I think, would be to create a new kind of ISQLFunction, which aliases another function.

`RegisterFunction("secondtruncated ", new SQLFunctionAlias(this, "second"));`

hazzik:
Mysql seems to be failing because it needs to use `floor` instead of `cast`. I had this issue with my local postgres.

gliljas:
The problem isn't flooring, but rather the old MySQL version used on the TC build.

src/NHibernate/Dialect/Function/SQLFunctionRegistry.cs
```diff
@@ -20,8 +22,11 @@ public SQLFunctionRegistry(Dialect dialect, IDictionary<string, ISQLFunction> us
...
-			ISQLFunction result;
-			if (!userFunctions.TryGetValue(functionName, out result))
+			if (!userFunctions.ContainsKey(functionName) && !dialect.Functions.ContainsKey(functionName) && _functionAliases.TryGetValue(functionName, out var sqlFunction))
+			{
+				functionName = sqlFunction;
+			}
+			if (!userFunctions.TryGetValue(functionName, out ISQLFunction result))
 			{
 				dialect.Functions.TryGetValue(functionName, out result);
 			}
```
fredericDelaporte, line 32: This incurs double look-ups on the nominal case. I would rather avoid that.

```suggestion
			if (!userFunctions.TryGetValue(functionName, out ISQLFunction result) && !dialect.Functions.TryGetValue(functionName, out result))
			{
				if (_functionAliases.TryGetValue(functionName, out var sqlFunction) && !_functionAliases.ContainsKey(sqlFunction))
					return FindSQLFunction(sqlFunction);
			}
```

The ContainsKey is there to prevent an endless recursion in case of a badly setup `_functionAliases`.

src/NHibernate/Dialect/Function/SQLFunctionRegistry.cs
```diff
@@ -30,6 +35,10 @@ public ISQLFunction FindSQLFunction(string functionName)
 
 		public bool HasFunction(string functionName)
 		{
+			if (!userFunctions.ContainsKey(functionName) && !dialect.Functions.ContainsKey(functionName) && _functionAliases.TryGetValue(functionName, out var sqlFunction))
+			{
+				functionName = sqlFunction;
+			}
 			return userFunctions.ContainsKey(functionName) || dialect.Functions.ContainsKey(functionName);
```
fredericDelaporte, line 42: Same here.

```suggestion
			if (userFunctions.ContainsKey(functionName) || dialect.Functions.ContainsKey(functionName))
				return true;
			if (_functionAliases.TryGetValue(functionName, out var sqlFunction) && !_functionAliases.ContainsKey(sqlFunction))
				return HasFunction(sqlFunction);
			return false;
```

fredericDelaporte (dismissed):

fredericDelaporte:
I will likely merge this on next Sunday.

fredericDelaporte:
Rebased, and Deepsource legit concerns fixed.

fredericDelaporte (approved):

PR #3538, opened 2024-05-02
Treat all non-virtual methods of System.Object as not proxiable

### LINKED ISSUE #3536: MemberwiseClone should be virtual in dotnet 9 preview 3
I have begun testing my system with dotnet 9 Preview 3 and received a message that the method MemberwiseClone should be virtual. I have just changed the version from 8 to 9 and turned to receive the error when the system started.
I have looked for the changes between DotNet 8 and 9 and found a different signature from that method, MemberwiseClone. It has changed to protected internal and turned to be considered as a virtual declaration.

In dotnet 8 (release/8.0) - Object.CoreCLR.cs

```
// Returns a new object instance that is a memberwise copy of this
// object.  This is always a shallow copy of the instance. The method is protected
// so that other object may only call this method on themselves.  It is intended to
// support the ICloneable interface.
[Intrinsic]
protected unsafe object MemberwiseClone()
{
    object clone = RuntimeHelpers.AllocateUninitializedClone(this);

    // copy contents of "this" to the clone

    nuint byteCount = RuntimeHelpers.GetRawObjectDataSize(clone);
    ref byte src = ref this.GetRawData();
    ref byte dst = ref clone.GetRawData();

    if (RuntimeHelpers.GetMethodTable(clone)->ContainsGCPointers)
        Buffer.BulkMoveWithWriteBarrier(ref dst, ref src, byteCount);
    else
        Buffer.Memmove(ref dst, ref src, byteCount);

    return clone;
}
```

In dotnet 9 preview 3 (release/9.0-preview3) - Object.CoreCLR.cs

```
// Returns a new object instance that is a memberwise copy of this
// object.  This is always a shallow copy of the instance. The method is protected
// so that other object may only call this method on themselves.  It is intended to
// support the ICloneable interface.
[Intrinsic]
protected internal unsafe object MemberwiseClone()
{
    object clone = this;
    RuntimeHelpers.AllocateUninitializedClone(ObjectHandleOnStack.Create(ref clone));
    Debug.Assert(clone != this);

    // copy contents of "this" to the clone

    nuint byteCount = RuntimeHelpers.GetRawObjectDataSize(clone);
    ref byte src = ref this.GetRawData();
    ref byte dst = ref clone.GetRawData();

    if (RuntimeHelpers.GetMethodTable(clone)->ContainsGCPointers)
        Buffer.BulkMoveWithWriteBarrier(ref dst, ref src, byteCount);
    else
        SpanHelpers.Memmove(ref dst, ref src, byteCount);

    return clone;
}
```

I have changed the NHibernate code to consider the MemberwiseClone in a method ShouldBeProxiable at class DefaultDynamicProxyMethodCheckerExtensions. It works, but I wonder if it is the best approach.

```
public static bool ShouldBeProxiable(this MethodInfo method)
{
	// to use only for real methods (no getter/setter)
	return (method.DeclaringType != typeof (MarshalByRefObject)) &&
	       !IsFinalizeMethod(method) &&
	       (!(method.DeclaringType == typeof (object) && "MemberwiseClone".Equals(method.Name))) &&
		   (!(method.DeclaringType == typeof(object) && "GetType".Equals(method.Name))) &&
		   (!(method.DeclaringType == typeof (object) && "obj_address".Equals(method.Name))) && // Mono-specific method
	       !IsDisposeMethod(method) &&
				 (method.IsPublic || method.IsAssembly || method.IsFamilyOrAssembly);
}
```

I know DotNet 9 is in the preview stage yet, but I need to test my system using the latest version of DotNet.
Another way would be to define the validation Proxy as false, but I wouldn't like to do that.
If my approach is correct, I could send a pull request. If someone has another idea, please let me know.

hazzik:
Fixes #3536

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/ProxyValidator/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/ProxyValidator/Fixture.cs
@@ -1,7 +1,6 @@
 using System;
 using NHibernate.Proxy;
 using NUnit.Framework;
-using System.Collections.Generic;
 
 namespace NHibernate.Test.NHSpecificTest.ProxyValidator
 {
@@ -9,15 +8,6 @@ namespace NHibernate.Test.NHSpecificTest.ProxyValidator
 	public class Fixture
 	{
 		private readonly IProxyValidator pv = new DynProxyTypeValidator();
-
-		private void Validate(System.Type type)
-		{
-			ICollection<string> errors = pv.ValidateType(type);
-			if (errors != null)
-			{
-				throw new InvalidProxyTypeException(errors);
-			}
-		}
 
 		public class ValidClass
 		{
@@ -65,11 +55,34 @@ namespace NHibernate.Test.NHSpecificTest.ProxyValidator
 		}
 
 		[Test]
-		public void ValidClassTest()
+		public void ObjectIsValid()
 		{
-			Validate(typeof(ValidClass));
+			var errors = pv.ValidateType(typeof(object));
+			Assert.That(errors, Is.Null);
 		}
 
+		[Test]
+		public void ValidClassTest()
+		{
+			var errors = pv.ValidateType(typeof(ValidClass));
+			Assert.That(errors, Is.Null);
+		}
+
+		public class InvalidSealedToString : ValidClass
+		{
+			public sealed override string ToString()
+			{
+				return base.ToString();
+			}
+		}
+		
+		[Test]
+		public void SealedObjectOverride()
+		{
+			var errors = pv.ValidateType(typeof(InvalidSealedToString));
+			Assert.That(errors, Has.Count.EqualTo(1));
+		}
+		
 		public class InvalidPrivateConstructor : ValidClass
 		{
 			private InvalidPrivateConstructor()
@@ -80,7 +93,8 @@ namespace NHibernate.Test.NHSpecificTest.ProxyValidator
 		[Test]
 		public void PrivateConstructor()
 		{
-			Assert.Throws<InvalidProxyTypeException>(() => Validate(typeof(InvalidPrivateConstructor)));
+			var errors = pv.ValidateType(typeof(InvalidPrivateConstructor));
+			Assert.That(errors, Has.Count.EqualTo(1));
 		}
 
 		public class InvalidNonVirtualProperty : ValidClass
@@ -95,7 +109,8 @@ namespace NHibernate.Test.NHSpecificTest.ProxyValidator
 		[Test]
 		public void NonVirtualProperty()
 		{
-			Assert.Throws<InvalidProxyTypeException>(() => Validate(typeof(InvalidNonVirtualProperty)));
+			var errors = pv.ValidateType(typeof(InvalidNonVirtualProperty));
+			Assert.That(errors, Has.Count.EqualTo(2));
 		}
 
 		public class InvalidPublicField : ValidClass
@@ -106,7 +121,8 @@ namespace NHibernate.Test.NHSpecificTest.ProxyValidator
 		[Test]
 		public void PublicField()
 		{
-			Assert.Throws<InvalidProxyTypeException>(() => Validate(typeof(InvalidPublicField)));
+			var errors = pv.ValidateType(typeof(InvalidPublicField));
+			Assert.That(errors, Has.Count.EqualTo(1));
 		}
 
 		public class InvalidNonVirtualEvent : ValidClass
@@ -119,7 +135,8 @@ namespace NHibernate.Test.NHSpecificTest.ProxyValidator
 		[Test]
 		public void NonVirtualEvent()
 		{
-			Assert.Throws<InvalidProxyTypeException>(() => Validate(typeof(InvalidNonVirtualEvent)));
+			var errors = pv.ValidateType(typeof(InvalidNonVirtualEvent));
+			Assert.That(errors, Has.Count.EqualTo(2));
 		}
 
 		public interface ValidInterface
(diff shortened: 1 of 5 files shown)


src/NHibernate.Test/NHSpecificTest/ProxyValidator/ShouldBeProxiableTests.cs
```diff
@@ -128,24 +163,16 @@ public void MyClassLowerCaseFinalizeIsProxiable()
...
+				() =>
+				{
+					Assert.That(method.ShouldBeProxiable(), Is.True);
+					Assert.That(method.IsProxiable(), Is.True);
+				});
 		}
 
 		[Test]
-		public void MyClassFinalizeWithParametersIsProxiable()
+		public void MyClassFinalizeWithParametersShouldBeProxiable()
```
hazzik, line 175: Collapsed `XXIsProxiable` & `XXShouldBeProxiable` tests into a single test with `Assert.Multiple`.

src/NHibernate/Proxy/DefaultDynamicProxyMethodCheckerExtensions.cs
```diff
@@ -23,12 +23,10 @@ public static bool IsProxiable(this MethodInfo method)
 		public static bool ShouldBeProxiable(this MethodInfo method)
 		{
 			// to use only for real methods (no getter/setter)
-			return (method.DeclaringType != typeof (MarshalByRefObject)) &&
-			       !IsFinalizeMethod(method) &&
```
hazzik, line 27: The `!IsFinalizeMethod(method)` check is not necessary, because object destructors are protected. This method only considers `public`, `internal` and `protected internal` methods.

src/NHibernate/Proxy/DefaultDynamicProxyMethodCheckerExtensions.cs
```diff
@@ -23,12 +23,10 @@ public static bool IsProxiable(this MethodInfo method)
...
-			       (!(method.DeclaringType == typeof (object) && "GetType".Equals(method.Name))) &&
-			       (!(method.DeclaringType == typeof (object) && "obj_address".Equals(method.Name))) && // Mono-specific method
+			return method.DeclaringType != typeof(MarshalByRefObject) &&
+			       !(method.DeclaringType == typeof(object) && !method.IsVirtual) &&
```
hazzik, line 27: Non virtual methods of `System.Object` should not be marked as required to be proxiable. This would handle `GetType`, `obj_address` (not standard `internal` Mono method, that was removed [here](https://github.com/mono/mono/commit/4072c020e175e00c8598dcc0fa87953a6c203586)), `MemberwiseClone`, etc.

hazzik:
In theory we could mark all `System.Object` methods as not required to be proxiable.

hazzik:
In theory we could mark all `System.Object` methods as not required to be proxiable. With non virtual methods it is understandable - we cannot override them anyway. But with virtual (`ToString`, `GetHashCode`, `Eqauls`) it is a bit trickier: user can override and seal the method. However, when we would be analyzing the class we would receive `MyClass.ToString` method for analysis, instead of `object.ToString` and it will say that the method should be proxiable although `object.ToString` is not required to be proxiable.

src/NHibernate.Test/NHSpecificTest/ProxyValidator/Fixture.cs
```diff
@@ -183,13 +194,15 @@ public class InvalidInternalField : ValidClass
...
 
 		[Test]
 		public void InternalField()
 		{
-			Assert.Throws<InvalidProxyTypeException>(() => Validate(typeof(InvalidInternalField)));
+			var errors = pv.ValidateType(typeof(InvalidInternalField));
+			Assert.That(errors, Has.Count.EqualTo(1));
 		}
 
 		public class InvalidNonVirtualProtectedProperty : ValidClass
```
fredericDelaporte, line 208: That name is confusing, even more with the withdrawal of the assert message.

```suggestion
		public class ValidNonVirtualProtectedProperty : ValidClass
```

src/NHibernate.Test/NHSpecificTest/ProxyValidator/Fixture.cs
```diff
@@ -204,8 +217,8 @@ protected int NonVirtualProperty
 		[Test]
 		public void NonVirtualProtected()
 		{
-			Validate(typeof(InvalidNonVirtualProtectedProperty));
-			Assert.IsTrue(true, "Always should pass, protected members do not need to be virtual.");
+			var errors = pv.ValidateType(typeof(InvalidNonVirtualProtectedProperty));
```
fredericDelaporte, line 220:
```suggestion
			var errors = pv.ValidateType(typeof(ValidNonVirtualProtectedProperty));
```

src/NHibernate.Test/NHSpecificTest/ProxyValidator/Fixture.cs
hazzik: Thanks

fredericDelaporte (approved):

hazzik:
@fredericDelaporte, I have a question. Should I re-target the changes to an earlier version? If so what? (I think we need a support policy so it would be clearer in the future).

fredericDelaporte:
Since we usually take long for releasing the next minor version, yes, you are right, we should target a lower version. The currently released minor seems enough to me.

I would rather not target older versions. That is more work which gives an incentive for not upgrading to the latest minor, which I do not see as desirable. It is reasonable to let older version be broken by .net versions which were not released when the older version was the latest.

hazzik:
Ok, cool. I'll squash and re-target to 5.5.x branch then.

fredericDelaporte (approved):

Issue #3539, opened 2024-05-02
Equals is ignored for Components that contain collection


DmitryMak:
NHibernate ignores Equals implementation on Components when they contain a collection. As a result, it issues an unnecessary DELETE followed by multiple INSERTs. It basically recreates all rows that were backing collection even though the collection has not changed. 

Example:

```

<class name="Person">
    <component name="CreditInfo">
        <bag name="OldCards" cascade="all" >
            <key column="PersonId" />
                ...


class Person {
    public CreditInfo CreditInfo;
}

public class CreditInfo {
    public readonly ICollection<String> OldCards;

    public CreditInfo(IEnumerable<String> oldCards) {        
        OldCards = new List<String>(oldCards);
    }
    
    public override Boolean Equals(Object obj) {

        var other = (CreditInfo)obj;
        return OldCards.SequenceEqual(other.OldCards);  <-- No changes detected by Equals!
    }
}

 
person.CreditInfo = new CreditInfo(person.CreditInfo.OldCards);  <-- effectively cloned

// will DELETE all card rows and INSERT all of them again even though CreditInfo did not change

```

I think there are 2 counterintuitive behaviors here:

1) Why is NHibernate ignoring CreditInfo.Equals? The method tells it that the component (value type) has not changed. A very common immutable Value Object scenario. Why does it have to go 'inside' the component? 

2) Replacing NHibernate collection (backing OldCards) with a new collection with the same items seem to make NHibernate delete and reinsert all rows. Is there a way to implement value equality on collections, think SequenceEquals?

DmitryMak:
Essentially I want to implement a small immutable collection that is compared by value (SequenceEquals). It looks like my only option is to serialize the collection into a string and store it as a column.

hazzik:
@DmitryMak can you submit a test case?

DmitryMak:
@hazzik I submitted a PR with a test that explain what I meant. Though I'm not sure how to write an assertion that fails when an unexpected DML is issued.

hazzik:
You have incorrect `GetHashCode` generation. It assumes hash code of an instance of a set, instead of it's elements.

hazzik:
> I'm not sure how to write an assertion that fails when an unexpected DML is issued.

Something like this:
```csharp
using var spy = new SqlLogSpy();
t.Commit();
Assert.That(spy.GetWholeLog(), Is.Empty);
```

PR #3541, opened 2024-05-03
Equals is ignored for Components that contain collection

DmitryMak:
Test that explains the intent in #3539

### DIFF
--- a/src/NHibernate.Test/NHSpecificTest/GH3539/CardInfo.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH3539/CardInfo.cs
@@ -1,0 +1,47 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NHibernate.Test.NHSpecificTest.GH3539;
+
+public class CardInfo
+{
+	private readonly ISet<string> _oldCards;
+
+	protected CardInfo() { }
+
+	public CardInfo(params string[] cards)
+	{
+		_oldCards = cards.ToHashSet();
+	}
+
+	public virtual ISet<string> GetCardsCopy()
+	{
+		return _oldCards.ToHashSet();
+	}
+
+	public override bool Equals(object obj)
+	{
+		if (obj is null)
+		{
+			return false;
+		}
+		if (ReferenceEquals(this, obj))
+		{
+			return true;
+		}
+		if (obj.GetType() != GetType())
+		{
+			return false;
+		}
+		var other = (CardInfo) obj;
+		return _oldCards.SetEquals(other._oldCards);
+	}
+
+	public override int GetHashCode()
+	{
+		var hashCode = new HashCode();
+		hashCode.Add(_oldCards);
+		return hashCode.ToHashCode();
+	}
+}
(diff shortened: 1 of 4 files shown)


src/NHibernate.Test/NHSpecificTest/GH3539/CardInfo.cs
```diff
@@ -0,0 +1,47 @@
...
+			return false;
+		}
+		var other = (CardInfo) obj;
+		return _oldCards.SetEquals(other._oldCards);
+	}
+
+	public override int GetHashCode()
+	{
+		var hashCode = new HashCode();
+		hashCode.Add(_oldCards);
```
hazzik, line 44: You have invalid hashcode generation function. You add a collection as an object.
```suggestion
		foreach (var card in _oldCards) hashCode.Add(card);
```

src/NHibernate.Test/NHSpecificTest/GH3539/ImmutableComponentTest.cs
```diff
@@ -0,0 +1,93 @@
...
+
+			using (ISession s = Sfi.OpenSession())
+			using (ITransaction t = s.BeginTransaction())
+			{
+				var restored = s.Get<Person>(personId);
+
+				var oldCardInfo = restored.CardInfo;
+
+				Assert.That(oldCardInfo.GetCardsCopy().Contains("card1"));
+				Assert.That(oldCardInfo.GetCardsCopy().Contains("card2"));
```
hazzik, line 74:
```suggestion
				Assert.That(oldCardInfo.GetCardsCopy().Contains("card2"));
				Assert.That(oldCardInfo.GetHashCode(), Is.EqualTo(newCardInfo.GetHashCode()));
```

src/NHibernate.Test/NHSpecificTest/GH3539/CardInfo.cs
DmitryMak: Thank you, I agree. But this is not a production code of course. I don't think it changes anything regarding the issue. Would just be a performance issue on production, hash collision.

hazzik: NHibernate uses GetHashCode to determine equality. It first checks hash code (must be the same), then calls equals.

hazzik:
Ok. I've fixed tests... But they still fail. Seems legit bug.

src/NHibernate.Test/NHSpecificTest/GH3539/CardInfo.cs
DmitryMak: oh, right! It actually may produce different hash codes for what we think should be the same value. I think the best approach would have been to use HashSet<String>.CreateSetComparer().GetHashCode(_oldCards). But it is not possible because NHibernate requires _oldCards to be an ISet. It uses foreach underneath anyway.

DmitryMak:
> Ok. I've fixed tests... But they still fail. Seems legit bug.

Thank you! FWIW I tried to use `collection-type` and a custom IUserCollection with Wrap overridden to wrap the new HashSet into a PersistentGenericSet, but the result was the same.

PR #3544, opened 2024-05-09
Upgrade vulnerable test dependencies

fredericDelaporte:
If we do security patches for 5.4.x, shouldn't we upgrade vulnerable dependencies?

At least in the main library, if not also in the test library.

### DIFF
--- a/build-common/NHibernate.props
+++ b/build-common/NHibernate.props
@@ -16,6 +16,7 @@
 
     <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net48;net6.0</NhAppTargetFrameworks>
     <NhLibTargetFrameworks Condition ="$(NhLibTargetFrameworks) == ''">net461;net48;netcoreapp2.0;netstandard2.0;netstandard2.1;net6.0</NhLibTargetFrameworks>
+    <RuntimeFrameworkVersion Condition="'$(TargetFramework)' == 'netcoreapp2.0'">2.0.3</RuntimeFrameworkVersion>
     <NhNetFx>false</NhNetFx>
     <NhNetFx Condition="$(TargetFramework.StartsWith('net4'))">true</NhNetFx>
     <!-- Visual Basic requires to use coma (,) as a separator, other project types - semicolon (;) -->
--- a/src/NHibernate.Test/NHibernate.Test.csproj
+++ b/src/NHibernate.Test/NHibernate.Test.csproj
@@ -56,11 +56,11 @@
   <ItemGroup>
     <PackageReference Include="log4net" Version="2.0.15" />
     <PackageReference Include="Microsoft.AspNetCore.OData" Version="7.5.12" />
-    <PackageReference Include="Microsoft.Data.SqlClient" Version="3.0.0" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="3.1.5" />
     <PackageReference Include="NHibernate.Caches.CoreDistributedCache.Memory" Version="5.8.0" />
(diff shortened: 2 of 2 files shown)


build-common/NHibernate.props
```diff
@@ -16,6 +16,7 @@
 
     <NhAppTargetFrameworks Condition ="$(NhAppTargetFrameworks) == ''">net48;net6.0</NhAppTargetFrameworks>
     <NhLibTargetFrameworks Condition ="$(NhLibTargetFrameworks) == ''">net461;net48;netcoreapp2.0;netstandard2.0;netstandard2.1;net6.0</NhLibTargetFrameworks>
+    <RuntimeFrameworkVersion Condition="'$(TargetFramework)' == 'netcoreapp2.0'">2.0.3</RuntimeFrameworkVersion>
```
fredericDelaporte, line 19: That is the minimal non vulnerable version of netcoreapp2.0.

The main effect of this is to cause the vulnerability warning to go away. It may prevent usage of the library with a lower vulnerable version, but that is not clear if that applies to libraries or only applications. See [here][https://learn.microsoft.com/en-us/dotnet/core/versions/selection#self-contained-deployments-include-the-selected-runtime).

> For framework-dependent applications, the RuntimeFrameworkVersion specifies the minimum required runtime framework version.

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -56,17 +56,17 @@
...
     <PackageReference Include="System.Data.SQLite.Core" Version="1.0.115.5" />
-    <PackageReference Include="System.Linq.Dynamic.Core" Version="1.2.12" />
+    <PackageReference Include="System.Linq.Dynamic.Core" Version="1.3.14" />
     <PackageReference Include="NSubstitute" Version="4.4.0" />
     <PackageReference Include="NUnit" Version="3.13.2" />
     <PackageReference Include="NUnit3TestAdapter" Version="4.4.2" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
     <PackageReference Include="FirebirdSql.Data.FirebirdClient" Version="8.5.2" />
-    <PackageReference Include="Npgsql" Version="6.0.6" />
+    <PackageReference Include="Npgsql" Version="8.0.3" />
```
fredericDelaporte, line 69: Npgsql has a vulnerability in all versions up to 8.0.2. SO I have backport the fix for Npgsql 8.

src/NHibernate.Test/NHibernate.Test.csproj
```diff
@@ -56,17 +56,17 @@
   <ItemGroup>
     <PackageReference Include="log4net" Version="2.0.15" />
     <PackageReference Include="Microsoft.AspNetCore.OData" Version="7.5.12" />
-    <PackageReference Include="Microsoft.Data.SqlClient" Version="3.0.0" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="3.1.5" />
```
fredericDelaporte, line 59: For test dependencies vulnerabilities, I have upgraded to the lowest fixed minor but highest patch of that minor.

src/NHibernate.Test/NHibernate.Test.csproj
hazzik: I think we need to split out the Npgsql upgrade to another PR to have it separated in the change-log.

fredericDelaporte: Done

build-common/NHibernate.props
hazzik: All version of .NET Core App 2.0 runtime are vulnerable. But for some reason not everything triggers a warning.

PR #3547, opened 2024-05-12
Handle SQL injection vulnerabilities within ObjectToSQLString

### LINKED ISSUE #3516: ObjectToSQLString implementations are problematic

fredericDelaporte:
Fix SQL injections noticed in #3516.

It does not fix wholly #3516. Bugs noted in #3516 that do not cause a security issue are not fixed.


fredericDelaporte:
That is currently handling only the string case, and tests need to be added.

src/NHibernate/Dialect/DB2Dialect.cs
```diff
@@ -296,6 +298,30 @@ public override string ForUpdateString
...
+			if (isUnicode)
+				literal.Replace(@"\", @"\\");
+
+			literal
+				.Replace("'", "''")
+				.Insert(0, '\'')
+				.Append('\'');
+
+			if (isUnicode)
+				literal.Insert(0, "U&");
```
fredericDelaporte, line 320: Other databases supports that escaping, but only DB2 documentation seemed to imply Unicode string literals were supported only by using it.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2076,6 +2069,55 @@ public virtual string ConvertQuotesForCatalogName(string catalogName)
...
+			var literal = new StringBuilder(value);
+			if (EscapeBackslashInStrings)
+				literal.Replace(@"\", @"\\");
+
+			literal
+				.Replace("'", "''")
+				.Insert(0, '\'')
+				.Append('\'');
+
+			if (UseNPrefixForUnicodeStrings && type.DbType == DbType.String || type.DbType == DbType.StringFixedLength)
```
fredericDelaporte, line 2114: Supporting Unicode "by the way" is not needed for fixing the vulnerability. But that is needed to justify the method signature: otherwise the `type` parameter would have no usage. So, I did this to avoid introducing a method without that parameter, which we would need to obsolete once we would support Unicode.

src/NHibernate.Test/NHSpecificTest/GH3516/FixtureByCode.cs
```diff
@@ -0,0 +1,54 @@
...
+		{
+			using var session = OpenSession();
+			using var transaction = session.BeginTransaction();
+			session.CreateQuery("delete from System.Object").ExecuteUpdate();
+
+			transaction.Commit();
+		}
+
+		[Test]
+		public void SqlInjectionInStrings()
```
fredericDelaporte, line 43: Cherry-pîcked on 5.4.x without the fix, that does wreck the Entity table.

src/NHibernate.Test/NHSpecificTest/GH3516/Entity.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace NHibernate.Test.NHSpecificTest.GH3516
+{
+	public class Entity
+	{
+		public virtual Guid Id { get; set; }
+		public virtual string Name { get; set; }
+
+		public const string NameWithSingleQuote = "'; drop table Entity; --";
+
+		public const string NameWithEscapedSingleQuote = @"\'; drop table Entity; --";
```
fredericDelaporte, line 12: Some databases support backslash escaping on top of doubling escaping. So if the fix was only doubling the quote without also escaping the backslash, there would still be an injection vulnerability, because the first quote would be escaped and not considered by the database.

deAtog:
I don't like this fix. In my opinion, every dialect should have a method to escape a string value for use in a SQL statement. This allows the mechanism of how strings are escaped to vary from dialect to dialect. Additionally, this alleviates the need for a flag as the default implementation of this function in the Dialect class can simply return the value as is.

hazzik:
I would agree with @deAtog that it would be better to delegate the escape to the dialect instead of introducing the configuration option.

src/NHibernate/Dialect/IngresDialect.cs
```diff
@@ -62,6 +62,11 @@ public IngresDialect()
 		/// <inheritdoc />
 		public override int MaxAliasLength => 32;
 
+		/// <inheritdoc />
+		/// <remarks><see langword="true" /> by default for Ingres,
+		/// <see href="https://docs.actian.com/ingres/11.0/index.html#page/SQLRef/String_Literals.htm#ww110572" />.</remarks>
+		protected override bool UseNPrefixForUnicodeStrings { get; set; } = true;
```
hazzik, line 68: Can we have this as computed please?

fredericDelaporte:
I do not really understand your comment, deAtog.

> every dialect should have a method to escape a string value for use in a SQL statement

But that is what this PR does, since the base implementation is overridable. Any dialect can define its own implementation if needed. To minimize the likeliness of leaving a dialect with a vulnerability, we should have the most common escaping implemented in the base dialect. All dialects in NHibernate support doubling the quotes. Quite many also supports backslash escaping as an alternative, which forces to escapes backslashes too for them. That is why the base implementation handles doubling single quotes and provides an option to easily enable backslash escapes.

But a dialect whose needs would require a different implementation can still provide it, as done in the DB2 case.

About additional escapes that may be required for database also handling a backslash escape, I may be wrong but it seemed to me the backslash escaping was enough. Or would not escaping line feeds and the like still open injection vulnerabilities for them?

I am not attempting to implement completely all the various escaping of all the supported databases. The subject is to remove a corner cases injection vulnerability. (There is very few uses of literal injection in NHibernate: discriminators, which are normally static values in mapping files, not application user data ; HQL queries referencing static fields, which is a very little known and used feature, and some sql builder helpers overloads that NHibernate does not use and that we are going to obsolete.)

src/NHibernate/Dialect/PostgreSQL91Dialect.cs
```diff
@@ -0,0 +1,16 @@
+namespace NHibernate.Dialect
+{
+	/// <summary>
+	/// An SQL dialect for PostgreSQL 9.1 and above.
+	/// </summary>
+	/// <remarks>
+	/// PostgreSQL 9.1 disables backslash escape sequences by default.
+	/// </remarks>
+	public class PostgreSQL91Dialect : PostgreSQL83Dialect
```
hazzik, line 9: Not sure that we need to introduce new dialect. According to this: https://endoflife.date/postgresql support for Postgres 9.1 ended more than 7 years ago. I think we can enable/disable this option by default.

src/NHibernate/Dialect/Dialect.cs
deAtog: `a && b || c` evaluates as `(a && b) || c` Did you intend to always add the N prefix for DbType.StringFixedLength or maybe the line should be:
```C#
if (UseNPrefixForUnicodeStrings && (type.DbType == DbType.String || type.DbType == DbType.StringFixedLength))
```

src/NHibernate/Cfg/Environment.cs
```diff
@@ -119,6 +119,12 @@ public static string Version
 		/// <summary> Enable formatting of SQL logged to the console</summary>
 		public const string FormatSql = "format_sql";
 
+		/// <summary>
+		/// Indicates if the database needs to have backslash escaped in string literals.
+		/// </summary>
+		/// <remarks>The default value is dialect dependent.</remarks>
+		public const string EscapeBackslashInStrings = "escape_backslash_in_strings";
```
deAtog, line 126: This should not be a user configurable option in my opinion. The dialect should perform the all appropriate escaping where necessary. For example, your implementation for DB2 ignores this setting while always escaping back slashes for Unicode strings.

src/NHibernate/Dialect/DB2Dialect.cs
deAtog: It would be nice if this code avoided multiple iterations over the initial value. The longer the value and more escaping added, the more expensive this operation could become. I would recommend similar changes elsewhere.

Pfmawlaw:
On Monday, May 20, 2024 at 09:35:45 AM PDT, deAtog ***@***.***> wrote:  
 
 


@deAtog commented on this pull request.

In src/NHibernate/Dialect/DB2Dialect.cs:
> +			if (type == null)
+				throw new ArgumentNullException(nameof(value));
+
+			// See https://www.ibm.com/docs/en/db2/11.5?topic=elements-constants#r0000731__title__7
+			var literal = new StringBuilder(value);
+			var isUnicode = type.DbType == DbType.String || type.DbType == DbType.StringFixedLength;
+			if (isUnicode)
+				literal.Replace(@"\", @"\\");
+
+			literal
+				.Replace("'", "''")
+				.Insert(0, '\'')
+				.Append('\'');
+
+			if (isUnicode)
+				literal.Insert(0, "U&");

It would be nice if this code avoided multiple iterations over the initial value. The longer the value and more escaping added, the more expensive this operation could become. I would recommend similar changes elsewhere.

deAtog:
> I do not really understand your comment, deAtog.
> 
> > every dialect should have a method to escape a string value for use in a SQL statement

My point here is that the base dialect shouldn't try to do anything. In my opinion this should probably be an abstract method that all dialects are required to implement. I personally do not like the idea of having a base implementation for something that is largely database dependent.

> 
> But that is what this PR does, since the base implementation is overridable. Any dialect can define its own implementation if needed. To minimize the likeliness of leaving a dialect with a vulnerability, we should have the most common escaping implemented in the base dialect. All dialects in NHibernate support doubling the quotes. Quite many also support backslash escaping as an alternative, which forces to escapes backslashes too for them. That is why the base implementation handles doubling single quotes and provides an option to easily enable backslash escapes.
> 
> Furthermore, dialects known to support backslash escapes are enabling the option by default. I still let open the possibility to enable/disable it by configuration due to cases like PostgreSQL, which has varied about its backslash escape handling.
> 
> A dialect whose needs would require a different implementation can still provide it, as done in the DB2 case.
> 
> About additional escapes that may be required for database also handling a backslash escape, I may be wrong but it seemed to me the backslash escaping was enough. Or would not escaping line feeds and the like still open injection vulnerabilities for them?

Personally, I don't know. I can see cases where certain control characters may cause problems. The handling of these characters is solely database/library dependent. Some may have issues, others may not. For example, one might be able to insert a backspace character into a string to delete previous characters or insert a null character to early terminate the string.

Also, while there _may_ be a standard for escaping single quotes within a single quoted string literal, not every database escapes them according to the standard. MySQL, for instance, recommends escaping single quotes with a backslash, but supports the use of doubling the number of single quotes within a single quoted string literal.

> 
> I am not attempting to implement completely all the various escaping of all the supported databases. The subject is to remove a corner cases injection vulnerability. (There is very few uses of literal injection in NHibernate: discriminators, which are normally static values in mapping files, not application user data; HQL queries referencing static fields, which is a very little known and used feature; and some sql builder helpers overloads that NHibernate does not use and that we are going to obsolete.)

I'll admit, I don't know the full scope of this issue. Most of the above seem like cases where the injection would have to be purposeful. However, if there is a case where a user-provided value can be directly inserted into a query then I wouldn't leave anything to chance. In my experience, it's better to err on the side of caution than to assume a partial implementation corrects all the issues.

fredericDelaporte:
> In my opinion this should probably be an abstract method that all dialects are required to implement.

That is a breaking change which would mandate a new major version. This cannot be done in a patch release. (A throwing base implementation would not be acceptable either for a patch release.)

> This should not be a user configurable option in my opinion.

See the PostgreSQL case: it was supporting backslash escapes by default up to its 9.1 version, with an option to disable it. Then it has ceased supporting them by default, but it still allows to re-enable them with a configuration setting. That case can only be handled through a configuration setting on our side too, since we cannot assume which configuration will have the database, whatever its version. The user needs to be able to enable/disable the escape for PostgreSql according to its database configuration, if he uses one of the features relying on literal injection.

> For example, your implementation for DB2 ignores this setting while always escaping back slashes for Unicode strings.

Because its Unicode string encoding mandates backslash escapes, no matter what. So, that would be nonsense to follow the configuration settings. Such a setting is meant for cases where the user can know better what is the correct course of action.

> It would be nice if this code avoided multiple iterations over the initial value.

Sure. But I see this as just a nice to have, since the implied features are either very seldomly used, or, in the case of discriminators, usually very short.

My first course of resolution was to suggest banning all literal injections in favor of parameterization, which is the best way to avoid injection vulnerabilities. But some discriminator use cases where this could significantly lower performances were raised. So, now, I attempt a "good enough" fix instead.

src/NHibernate/Dialect/Dialect.cs
fredericDelaporte: Missing parenthesis, good spot.

src/NHibernate/Dialect/PostgreSQL91Dialect.cs
fredericDelaporte: So, the current state of supported PostgerSQL databases is to have backslash escapes disabled by default. I undo any PostegreSQL dialect related changes to align on this.

deAtog:
> > In my opinion this should probably be an abstract method that all dialects are required to implement.
> 
> That is a breaking change which would mandate a new major version. This cannot be done in a patch release. (A throwing base implementation would not be acceptable either for a patch release.)

A base implementation that does nothing but return the value provided would be best in my opinion. This would allow the method to be changed to abstract in a future major release. This also guarantees that the base implementation does not break some obscure dialect. As is, there is no guarantee that the current implementation doesn't introduce a breaking change for **all** dialects.

> 
> > This should not be a user configurable option in my opinion.
> 
> See the PostgreSQL case: it was supporting backslash escapes by default up to its 9.1 version, with an option to disable it. Then it has ceased supporting them by default, but it still allows to re-enable them with a configuration setting. That case can only be handled through a configuration setting on our side too, since we cannot assume which configuration will have the database, whatever its version. The user needs to be able to enable/disable the escape for PostgreSql according to its database configuration, if he uses one of the features relying on literal injection.

There's no doubt that a configuration flag is needed here for PostgreSQL. The problem here is that you've found a case where the dialect breaks the base implementation and have introduced a global flag to avoid a dialect specific issue, hence my recommendation above. Dialects are provided all configuration properties when the Dialect.Configure is called. My recommendation would be to only interpret the escape_backslash_in_strings property from within the PostgreSQL dialect. There is no convention on where to define dialect specific properties, but I don't think defining it with all the other global properties is appropriate as it implies that it affects all dialects alike. I have seen many cases of cache specific properties that use a similar approach. Such properties are usually only mentioned in the documentation, but it would be good to have constants defined for them and a convention of where to find them.

> 
> > For example, your implementation for DB2 ignores this setting while always escaping back slashes for Unicode strings.
> 
> Because its Unicode string encoding mandates backslash escapes, no matter what. So, that would be nonsense to follow the configuration settings. Such a setting is meant for cases where the user can know better what is the correct course of action.

That's exactly my point. From the context provided, the user has no idea whether or not the setting will affect the dialect they're using.

> 
> > It would be nice if this code avoided multiple iterations over the initial value.
> 
> Sure. But I see this as just a nice to have, since the implied features are either very seldomly used, or, in the case of discriminators, usually very short.
> 
> My first course of resolution was to suggest banning all literal injections in favor of parameterization, which is the best way to avoid injection vulnerabilities. But some discriminator use cases where this could significantly lower performances were raised. So, now, I attempt a "good enough" fix instead.

Parameters would undoubtedly be better in my opinion, regardless of performance. There are many opportunities where performance could be improved, but now is not the time to address that concern.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2076,6 +2069,55 @@ public virtual string ConvertQuotesForCatalogName(string catalogName)
...
+				throw new ArgumentNullException(nameof(value));
+
+			var literal = new StringBuilder(value);
+			if (EscapeBackslashInStrings)
+				literal.Replace(@"\", @"\\");
+
+			literal
+				.Replace("'", "''")
+				.Insert(0, '\'')
+				.Append('\'');
```
deAtog, line 2112: There's no guarantee that the implementation of this method won't break an obscure dialect that does not support doubling the number of single quotes inside a single quoted string to escape a single quote. While this is the ANSI standard, not all dialects adhere to that standard.

The only non-breaking thing to do here is to simply return the value passed.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2076,6 +2069,55 @@ public virtual string ConvertQuotesForCatalogName(string catalogName)
...
+		/// <remarks><see langword="false" /> by default in the base dialect.</remarks>
+		protected virtual bool UseNPrefixForUnicodeStrings => false;
+
+		/// <summary>The SQL string literal value to which this database maps string values.</summary>
+		/// <param name="value">The string value.</param>
+		/// <param name="type">The SQL type of the string value.</param>
+		/// <returns>The appropriate SQL string literal.</returns>
+		/// <exception cref="ArgumentNullException">Thrown if <paramref name="value"/> or
+		/// <paramref name="type"/> is <see langword="null" />.</exception>
+		public virtual string ToStringLiteral(string value, SqlType type)
```
deAtog, line 2098: I'm not sold on necessity of passing the SqlType here. The primary purpose of this value is to determine whether or not the value to be escaped should result in a Unicode value. I would suggest one of the following:
1. Create a separate method called ToUnicodeStringLiteral.
2. Pass a boolean value indicating whether or not to return a Unicode string literal.

Admittedly, I prefer the former over the latter as it allows dialects to have different implementations if needed without a conditional on the boolean value provided.

src/NHibernate/Dialect/Dialect.cs
fredericDelaporte: Fixing a bug of lack of escapism is anyway behavior breaking for those which were compensating it by escaping values themselves (if anyone).

And from a security standpoint, better take the risk of breaking some dialects not natively supported by NHibernate rather than leaving the most common vulnerability (lack of doubling the quote) open for all not natively supported dialects.

Fixing bug is anyway always somewhat behavior breaking. That is normal and accepted for patch releases. That is binary breaking changes or source breaking changes which are banned by SemVer for a patch or minor release.

Mores specifically, from https://semver.org/ :
> Patch version Z (x.y.Z | x > 0) MUST be incremented if only backward compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.

(So, a new throwing base method causing previously "working in most cases" features to cease working is neither a binary breaking change nor a source one. But undoubtedly it does not qualify as "an internal change that fixes incorrect behavior".)

fredericDelaporte:
I have added the fix for the char type. I have reordered and cleaned the commits by the way, putting tests first to facilitate checking them without the fix.

src/NHibernate/Type/AbstractCharType.cs
```diff
@@ -51,9 +51,7 @@ public override void Set(DbCommand cmd, object value, int index, ISessionImpleme
 		}
 
 		public override string ObjectToSQLString(object value, Dialect.Dialect dialect)
-		{
-			return '\'' + value.ToString() + '\'';
```
fredericDelaporte, line 55: ToString instead of cast does allow for some type mismatch in queries using static fields. That allows comparing a char entity property to a string instead of a char. The reverse fails due to the string type having a cast instead.

So, we have inconsistent behavior here. As that is not the subject of this PR, I leave it as is.

src/NHibernate/Dialect/Dialect.cs
deAtog: If I understand correctly, you are saying that in the case of a bug fix, it's okay to throw an exception for changes that correct behavior. Admittedly, I do not have a problem with that so long as the exception is thrown in the appropriate location. The changes here would cause an exception to be thrown once the statement is executed, not when the method is called. Any dialects broken by this change would therefore not know the root cause of the exception. Therefore, if you'd like to throw an exception in the base case instead of leaving external dialects vulnerable, that would be acceptable.

Alternatively, as already discussed, NHibernate could use query parameters to avoid escaping any values. It may not be as performant as the solution here, but it would undoubtedly be more maintainable and secure. It would also not introduce any breaking changes.

src/NHibernate/Dialect/Dialect.cs
```diff
@@ -2076,6 +2069,55 @@ public virtual string ConvertQuotesForCatalogName(string catalogName)
...
+		{
+			if (value == null)
+				throw new ArgumentNullException(nameof(value));
+			if (type == null)
+				throw new ArgumentNullException(nameof(value));
```
hazzik, line 2103:
```suggestion
				throw new ArgumentNullException(nameof(type));
```

hazzik:
Ok, thanks for the clarification

deAtog:
@hazzik Even if this were only for strings this is not ready for merging in my opinion. I have reservations about breaking non-ANSI compliant dialects with the changes proposed here. It would be far better to switch to parameters, regardless of performance impacts, than use what has been implemented here. NHibernate already suffers from quite a few performance related issues and the impact of using parameters here is irrelevant given their security benefits.

hazzik:
@deAtog you're welcome to submit a PR with your suggestions.

src/NHibernate/Type/EnumCharType.cs
```diff
@@ -171,8 +171,6 @@ public override object FromStringValue(string xml)
 		}
 
 		public override string ObjectToSQLString(object value, Dialect.Dialect dialect)
-		{
-			return '\'' + GetValue(value).ToString() + '\'';
-		}
+			=> dialect.ToStringLiteral(GetValue(value).ToString(), SqlType);
```
fredericDelaporte, line 174: That one is a bit wicked as it does cast the enum underlying integral value as a character, allowing for injecting any character. So, it could abnormally terminate a query by injecting a quote, by example. With a single character injection, it looks like not much harm may be done. But if the query gets batched with others queries, who knows?

So, I have to add a test case for this one.

src/NHibernate/Type/UriType.cs
```diff
@@ -84,9 +84,7 @@ public override object FromStringValue(string xml)
 		}
 
 		public string ObjectToSQLString(object value, Dialect.Dialect dialect)
-		{
-			return "'" + ((Uri)value).OriginalString + "'";
-		}
+			=> dialect.ToStringLiteral(((Uri) value).OriginalString, SqlType);
```
fredericDelaporte, line 87: I have to check if `Uri` may accept an original value allowing SQL injections (very likely yes), and if yes, add another test case.

src/NHibernate.Test/NHSpecificTest/GH3516/FixtureByCode.cs
```diff
@@ -0,0 +1,288 @@
...
+			else
+			{
+				using var session = OpenSession();
+				var all = session.CreateQuery("from Entity e").List<Entity>();
+				Assert.That(all, Has.Count.GreaterThan(0));
+			}
+		}
+
+		[Test]
+		public void SqlInjectionInStringDiscriminator()
```
fredericDelaporte, line 166: Checking the trouble with string discriminator. Other discriminator types do not have the trouble, because the convert the mapping value to their type, and would fail there in case of an injection attempt.

src/NHibernate/Type/DecimalType.cs
```diff
@@ -66,7 +66,7 @@ public override object FromStringValue(string xml)
 
 		public override string ObjectToSQLString(object value, Dialect.Dialect dialect)
 		{
-			return value.ToString();
+			return Convert.ToDecimal(value).ToString();
```
fredericDelaporte, line 69: That is a minimal fix, causing the invalid query injected SQL to be rejected when defined. It could be a direct cast too, but that may break working cases where the value was already a valid string representation of the type.

This fix is minimal, especially for decimal or floating types, because:
 - It does not fix the possible troubles triggered when running under some specific cultures.
 - Some databases have different literal representations than what .Net generates for these types.

fredericDelaporte:
Still WIP: I have to review all the other types to check if they allow some SQL injection.

src/NHibernate.Config.Templates/SapSQLAnywhere.cfg.xml
```diff
@@ -15,7 +15,7 @@ for your own use before compiling tests in Visual Studio.
 		<property name="connection.connection_string">
 			UID=DBA;PWD=sql;Server=localhost;DBN=nhibernate;DBF=c:\nhibernate.db;ASTOP=No;Enlist=false;
 		</property>
-		<property name="dialect">NHibernate.Dialect.SybaseSQLAnywhere12Dialect</property>
+		<property name="dialect">NHibernate.Dialect.SapSQLAnywhere17Dialect</property>
```
fredericDelaporte, line 18: We cannot test it locally if staying on the old dialect. That is only the template, so, the trouble can be fixed locally. But why not fixing that here?

src/NHibernate.Test/NHSpecificTest/GH3516/FixtureByCode.cs
```diff
@@ -0,0 +1,343 @@
...
+				Assert.That(() => list = session.CreateQuery("from Entity e").List<Entity>(), Throws.Nothing);
+				Assert.That(list, Has.Count.GreaterThan(0));
+			}
+		}
+
+		[Test]
+		public void SqlInjectionWithDatetime()
+		{
+			var wickedCulture = new CultureInfo("en-US");
+			wickedCulture.DateTimeFormat.ShortDatePattern = "yyyy-MM-ddTHH:mm:ss\\'\"; drop table Entity; --\"";
```
fredericDelaporte, line 328: Yes, that does drop the table when comparing a datetime to a static property...

src/NHibernate/Type/AbstractDateTimeType.cs
```diff
@@ -176,7 +176,7 @@ public override object FromStringValue(string xml)
 		public override object DefaultValue => BaseDateValue;
 
 		/// <inheritdoc />
-		public override string ObjectToSQLString(object value, Dialect.Dialect dialect) =>
-			"'" + (DateTime) value + "'";
+		public override string ObjectToSQLString(object value, Dialect.Dialect dialect)
+			=> dialect.ToStringLiteral(((DateTime) value).ToString(), SqlTypeFactory.GetAnsiString(50));
```
fredericDelaporte, line 180: That is still far from being a correct datetime representation for many SQL vendors and cultures. But at least, it does no more allow SQL injections. Granted, injecting through a very custom culture is a hard to exploit attack vector, even more with the unlikeliness of the vulnerable features to be used by applications. But after all, who knows how corrupted a computer culture might be?

src/NHibernate/Type/DecimalType.cs
fredericDelaporte: That is a still vulnerable fix, because the culture can be altered in order to cause injection through the NUmberFormatInfo properties likes NegativeSignor DecimalSeparator...

fredericDelaporte (commented):
So, still WIP

src/NHibernate.Test/NHSpecificTest/GH3516/FixtureByCode.cs
hazzik: Niiice

src/NHibernate/Type/DecimalType.cs
deAtog: I was just about to comment the same. You cannot call Convert.To* or <type>.ToString() safely without specifying an IFormatProvider. Doing so will result in inconsistent results whenever the culture changes. Additionally, some dialects may allow the use of a comma as a decimal separator instead of a period as stated in the DB2 documentation here: https://www.ibm.com/docs/en/db2-for-zos/12?topic=1-decimal-point-is-field-decimal-decp-value Needless to say, none of this would matter if parameters were used.

src/NHibernate/Type/Int32Type.cs
```diff
@@ -114,7 +115,7 @@ public override object DefaultValue
 
 		public override string ObjectToSQLString(object value, Dialect.Dialect dialect)
 		{
-			return value.ToString();
+			return Convert.ToInt32(value).ToString(CultureInfo.InvariantCulture);
```
hazzik, line 118: Can we do downcast instead? `return ((Int32)value).ToString(CultureInfo.InvariantCulture);`?

fredericDelaporte: https://github.com/nhibernate/nhibernate-core/pull/3547#discussion_r1632428617

> It could be a direct cast too, but that may break working cases where the value was already a valid string representation of the type.

For many types, the static field injection in queries features could work with non-matching types. The convert can already be a possible breaking change. The cast seems to me as possibly breaking more cases.

There are also a bunch of types who do cast to their expected type. So, for consistency, it will be better for all types to the same.

Still, here, we are in a patch PR, so, I tend to favor minimize the possibilities of introducing possible breaking changes for some cases. But the possibly affected cases are also corner uses of a corner case feature. So, why not.

hazzik: @fredericDelaporte can you explain why you've changed your opinion?

src/NHibernate.Test/NHSpecificTest/GH3516/FixtureByCode.cs
```diff
@@ -0,0 +1,388 @@
...
+
+			mapper.Class<BaseClass>(rc =>
+			{
+				rc.Id(x => x.Id, m => m.Generator(Generators.GuidComb));
+				rc.Discriminator(x => x.Column("StringDiscriminator"));
+				rc.Property(x => x.Name);
+				rc.Abstract(true);
+			});
+			mapper.Subclass<Subclass1>(rc => rc.DiscriminatorValue(Entity.NameWithSingleQuote));
+			mapper.Subclass<Subclass2>(rc => rc.DiscriminatorValue(Entity.NameWithEscapedSingleQuote));
```
hazzik, line 78: It feels like we need to try to query these

src/NHibernate/Type/Int32Type.cs
fredericDelaporte: Isn't what does my previous comment? In other words, I have no strong opinion on that subject. Both options have advantages and drawbacks.

src/NHibernate.Test/NHSpecificTest/GH3516/FixtureByCode.cs
fredericDelaporte: Isn't it done at line 222?

hazzik: Thanks. Missed that

hazzik (approved):

PR #3554, opened 2024-05-29
Fix support for nullable struct components

### LINKED ISSUE #854: NH-1284 - Initialization of nullable structs component
**McJoe Mayorga** created an [issue](https://nhibernate.jira.com/browse/NH-1284) — <time datetime="2008-04-07T15:35:54.581Z" title="2008-04-07T15:35:54.581Z">7th April 2008, 15:35:54</time>:
> The fix to [NH-901](https://nhibernate.jira.com/browse/NH-901) breaks the following prior `ComponentType` behavior (from reference doc):
> 
> > Like all value types, components do not support shared references. The null value semantics of a component are ad hoc. When reloading the containing object, NHibernate will assume that if all component columns are null, then the entire component is null. This should be okay for most purposes.
> 
> 
> As of 1.2.1, the component will no longer be null.  Any value type struct member T will now get initialized to default(T), which may produce unwanted behavior.  Sample test attached.
> 
> 
> 
---
**Karl Chu** added a comment — <time datetime="2008-04-23T18:34:11.221Z" title="2008-04-23T18:34:11.221Z">23rd April 2008, 18:34:11</time>:
> What is "unwanted behavior" is debatable.  If one uses a struct, one wants value semantics.  What would be the reasons to use a "nullable struct" instead of making it a class?
> 
---
**McJoe Mayorga** added a comment — <time datetime="2008-05-07T16:28:25.786Z" title="2008-05-07T16:28:25.786Z">7th May 2008, 16:28:25</time>:
> The example above is a bit contrived.  A better one would be something like LocalizedPrice, which contains a price and a currency.  This is pretty naturally a struct.  We need to differentiate between null values (payment pending, say) and defaults (voided payment in default currency) .
> 
---
**Fabio Maulo** added a comment — <time datetime="2008-07-12T22:34:19.204Z" title="2008-07-12T22:34:19.204Z">12th July 2008, 22:34:19</time>:
> The problem is something more deep than appear.
> With NH-901 we are allowing the use of value type like component.
> In NH-1284 the user want use a component in a nullable property.
> The real underlining type of  "Address?" is not "Address" but "`System.Nullable<Address>`".
> If the user map the Type "Address" (a struct) NH don't have a way to know that it is nullable.
> If the user don't map the type NH can't find "`System.Nullable<Address>`".
> More than a bug it appear like a limitation so the issue can be converted to improvement.
> 
> What we need is to catch the real type of the property (in this case "`System.Nullable<Address>`") and create mapping info for underlining type.
> We can't fix it for NH2.0.0.
> 
---
**Fabio Maulo** added a comment — <time datetime="2008-07-12T22:47:50.305Z" title="2008-07-12T22:47:50.305Z">12th July 2008, 22:47:50</time>:
> Added test to the trunk.
> 

----
* [NH901regression.zip](https://github.com/nhibernate/nhibernate-core/files/15477182/NH901regression.zip)

[NH-901]: https://nhibernate.jira.com/browse/NH-901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

hazzik:
Fixes #854


src/NHibernate/Cfg/XmlHbmBinding/PropertiesBinder.cs
```diff
@@ -402,30 +386,27 @@ private void BindCollectionProperty(ICollectionPropertiesMapping collectionMappi
 			property.Cascade = collectionMapping.Cascade ?? mappings.DefaultCascade;
 		}
 
-		private Property CreateProperty(IEntityPropertyMapping propertyMapping, string propertyOwnerClassName, IValue value, IDictionary<string, MetaAttribute> inheritedMetas)
```
hazzik, line 405: In all cases the caller was passing `className` as `propertyOwnerClassName` argument.

fredericDelaporte (approved):

PR #3557, opened 2024-06-03
Fix configuration order in tests

hazzik:
Configuration could influence mappings (via `INamingStrategy`, for example), so it should be called before `AddMappings`.

Also remove calls to empty `base.Configuration(...)` and fix usage of `configuration` parameter.


src/NHibernate.Test/ConnectionTest/AsyncLocalSessionContextFixture.cs
```diff
@@ -17,8 +17,7 @@ protected override ISession GetSessionUnderTest()
 
 		protected override void Configure(Configuration configuration)
```
fredericDelaporte, line 18: Should rename too, to avoid masking the base property.

src/NHibernate.Test/Linq/QueryCacheableTests.cs
```diff
@@ -14,7 +14,6 @@ protected override void Configure(Configuration cfg)
 		{
 			cfg.SetProperty(Environment.UseQueryCache, "true");
 			cfg.SetProperty(Environment.GenerateStatistics, "true");
-			base.Configure(cfg);
```
fredericDelaporte, line 17: Same.

hazzik: Done

fredericDelaporte (approved):

PR #3559, opened 2024-06-05
Add new DB2NetDriver and obsolete DB2CoreDriver

hazzik:
* Added new driver for IBM DB2 to be used with [Net5.IBM.Data.Db2](https://www.nuget.org/packages/Net5.IBM.Data.Db2) (.NET 5) and [Net.IBM.Data.Db2](https://www.nuget.org/packages/Net.IBM.Data.Db2) (.NET 6-8).
* Obsolete driver for .NET Core 2-3.1
---
_**NB:**_ on Windows set `DB2LANG=en_US` environment variable.

### DIFF
--- a/src/NHibernate/Driver/DB2CoreDriver.cs
+++ b/src/NHibernate/Driver/DB2CoreDriver.cs
@@ -1,27 +1,16 @@
-using System.Data.Common;
-using NHibernate.SqlTypes;
+using System;
 
 namespace NHibernate.Driver
 {
 	/// <summary>
 	/// A NHibernate Driver for using the IBM.Data.DB2.Core DataProvider.
 	/// </summary>
-	public class DB2CoreDriver : DB2DriverBase
-	{		
+	// Since v5.6
+	[Obsolete("Please use DB2NetDriver")]
+	public class DB2CoreDriver : DB2NetDriver
+	{
 		public DB2CoreDriver() : base("IBM.Data.DB2.Core")
 		{
-		}
-		
-		public override bool UseNamedPrefixInSql => true;
-
-		public override bool UseNamedPrefixInParameter => true;
-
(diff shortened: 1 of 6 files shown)


src/NHibernate/Driver/DB2CoreDriver.cs
```diff
@@ -1,27 +1,16 @@
...
 namespace NHibernate.Driver
 {
 	/// <summary>
 	/// A NHibernate Driver for using the IBM.Data.DB2.Core DataProvider.
 	/// </summary>
-	public class DB2CoreDriver : DB2DriverBase
-	{		
+	// Since v5.6
+	[Obsolete("Please use DB2NetDriver")]
+	public class DB2CoreDriver : DB2NetDriver
```
fredericDelaporte, line 10: Obsoleting a driver is not user source noticeable in most cases, since it is usually setup through string value configuration.

We should issue a log warning to increase the likeliness it might be seen.

src/NHibernate/Driver/DB2CoreDriver.cs
```diff
@@ -1,27 +1,16 @@
...
+	{
 		public DB2CoreDriver() : base("IBM.Data.DB2.Core")
 		{
 		}
```
fredericDelaporte, line 14:
```suggestion
		private static readonly INHibernateLogger Log = NHibernateLogger.For(typeof(DB2CoreDriver));
		
		public DB2CoreDriver() : base("IBM.Data.DB2.Core")
		{
		    Log.Warn("DB2CoreDriver is obsolete, please use DB2NetDriver instead.");
		}
```

hazzik: Thanks

fredericDelaporte (approved):

PR #3560, opened 2024-06-07
Fix parameter name of Check() in ClassCustomizer

dennis-gr:
Should be _check_ instead of _tableName_.


src/NHibernate/Mapping/ByCode/Impl/CustomizersImpl/ClassCustomizer.cs
```diff
@@ -117,9 +117,9 @@ public void Table(string tableName)
 			CustomizersHolder.AddCustomizer(typeof(TEntity), (IClassMapper m) => m.Table(tableName));
 		}
 
-		public void Check(string tableName)
+		public void Check(string check)
```
fredericDelaporte, line 120: That is source breaking, so, next major change.

fredericDelaporte (approved):
Thanks.

PR #3568, opened 2024-06-29
A possible bug

tohidemyname:
A join query returns an empty set.


tohidemyname:
@hazzik Thanks a lot.

PR #3601, opened 2024-08-26
Encourage use of .NET 9.0's System.Threading.Lock

MarkCiliaVincenti:
Will improve performance once the library starts targeting .NET 9.0 as locking on System.Threading.Lock is ~25% more performant over locking on an object. Newly-added micro dependency allows backported use without affecting performance.

### DIFF
--- a/build-common/NHibernate.props
+++ b/build-common/NHibernate.props
@@ -6,7 +6,7 @@
     <VersionPatch  Condition="'$(VersionPatch)'  == ''">0</VersionPatch>
     <!-- Clear VersionSuffix for making release and set it to dev for making development builds -->
     <VersionSuffix Condition="'$(VersionSuffix)' == ''">dev</VersionSuffix>
-    <LangVersion Condition="'$(MSBuildProjectExtension)' != '.vbproj'">12.0</LangVersion>
+    <LangVersion Condition="'$(MSBuildProjectExtension)' != '.vbproj'">preview</LangVersion>
 
     <VersionPrefix Condition="'$(VersionPrefix)' == ''">$(NhVersion).$(VersionPatch)</VersionPrefix>
     <VersionSuffix Condition="'$(VersionSuffix)' != '' AND '$(BuildNumber)' != ''">$(VersionSuffix).$(BuildNumber)</VersionSuffix>
--- a/src/NHibernate.Test/NHSpecificTest/NH2030/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH2030/Fixture.cs
@@ -15,7 +15,7 @@ namespace NHibernate.Test.NHSpecificTest.NH2030
 		[Test]
 		public void GetTypeWithLenShouldBeThreadSafe()
 		{
-			Lock sync = new Lock();
+			object sync = new object();
 			List<Exception> exceptions = new List<Exception>();
 
 			ManualResetEvent startEvent = new ManualResetEvent(false);
--- a/src/NHibernate.Test/NHSpecificTest/NH2030/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH2030/Fixture.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Text;
@@ -15,7 +15,7 @@ namespace NHibernate.Test.NHSpecificTest.NH2030
 		[Test]
 		public void GetTypeWithLenShouldBeThreadSafe()
 		{
-			object sync = new object();
+			Lock sync = new Lock();
 			List<Exception> exceptions = new List<Exception>();
 
 			ManualResetEvent startEvent = new ManualResetEvent(false);
--- a/src/NHibernate.Test/NHSpecificTest/NH2192/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH2192/Fixture.cs
@@ -45,7 +45,7 @@ namespace NHibernate.Test.NHSpecificTest.NH2192
 		[Test]
 		public void HqlIsThreadsafe_UsingThreads()
 		{
-			object sync = new object();
+			Lock sync = new Lock();
 			List<int> results = new List<int>();
 			List<Exception> exceptions = new List<Exception>();
 
--- a/src/NHibernate.Test/NHSpecificTest/NH2192/Fixture.cs
+++ b/src/NHibernate.Test/NHSpecificTest/NH2192/Fixture.cs
@@ -45,7 +45,7 @@ namespace NHibernate.Test.NHSpecificTest.NH2192
 		[Test]
 		public void HqlIsThreadsafe_UsingThreads()
 		{
-			Lock sync = new Lock();
+			object sync = new object();
 			List<int> results = new List<int>();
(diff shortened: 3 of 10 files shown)


build-common/NHibernate.props
```diff
@@ -6,7 +6,7 @@
     <VersionPatch  Condition="'$(VersionPatch)'  == ''">0</VersionPatch>
     <!-- Clear VersionSuffix for making release and set it to dev for making development builds -->
     <VersionSuffix Condition="'$(VersionSuffix)' == ''">dev</VersionSuffix>
-    <LangVersion Condition="'$(MSBuildProjectExtension)' != '.vbproj'">12.0</LangVersion>
+    <LangVersion Condition="'$(MSBuildProjectExtension)' != '.vbproj'">preview</LangVersion>
```
fredericDelaporte, line 9: So, it needs language version 14.0? We will likely wait for it before merging your proposal, if we accept it.

src/NHibernate.Test/NHSpecificTest/NH2030/Fixture.cs
```diff
@@ -15,7 +15,7 @@ public class Fixture
 		[Test]
 		public void GetTypeWithLenShouldBeThreadSafe()
 		{
-			object sync = new object();
+			Lock sync = new Lock();
```
fredericDelaporte, line 18: What is the point of changing the locking mechanism of existing tests?

src/NHibernate/Cache/SyncCacheLock.cs
```diff
@@ -10,28 +10,23 @@ class SyncCacheLock : ICacheLock
 
 		class MonitorLock : IDisposable
```
fredericDelaporte, line 11: That is no more a Monitor based lock with your change. Better change its name. `InternalLock`, maybe?

src/NHibernate.Test/NHSpecificTest/NH2192/Fixture.cs
```diff
@@ -45,7 +45,7 @@ protected override void OnTearDown()
 		[Test]
 		public void HqlIsThreadsafe_UsingThreads()
 		{
-			object sync = new object();
+			Lock sync = new Lock();
```
fredericDelaporte, line 48: Same question.

build-common/NHibernate.props
fredericDelaporte: I have not spotted what does require such a change. Why this change?

src/NHibernate/NHibernate.csproj
```diff
@@ -73,6 +73,10 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
   </ItemGroup>
 
+  <ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))">
+    <PackageReference Include="Backport.System.Threading.Lock" Version="1.1.6" />
```
fredericDelaporte, line 77: So, that is [this Nuget package](https://www.nuget.org/packages/Backport.System.Threading.Lock), coming from [your repository](https://github.com/MarkCiliaVincenti/Backport.System.Threading.Lock).

fredericDelaporte (commented):
bUnit-dev/bUnit#1532 has a discussion about this which is relevant for NHibernate too.

For the simple usages we currently have (excepted the Monitor one which would have to be slightly different), using something like:
```c#
#if NET_9_0_OR_GREATER
private readonly Lock syncLock = new();
#else
private readonly object syncLock = new();
#endif
```

Seems enough and would avoid adding a new dependency for users to investigate if they want it in their project or not.

build-common/NHibernate.props
MarkCiliaVincenti: It's 13.0 as such, and comes with .NET 9.0, but for now you need to set the LangVersion to preview on .NET 9.0 in order to lock on a System.Threading.Lock object. The expectation is that this changes as soon as .NET 9.0 is released.

src/NHibernate.Test/NHSpecificTest/NH2030/Fixture.cs
MarkCiliaVincenti: Consistency maybe; it also helps in case something breaks with regards to the lock object. Can revert if you want.

src/NHibernate/NHibernate.csproj
MarkCiliaVincenti: Yes, that is correct.

src/NHibernate.Test/NHSpecificTest/NH2192/Fixture.cs
MarkCiliaVincenti: Reverted

src/NHibernate/Cache/SyncCacheLock.cs
MarkCiliaVincenti: Done

src/NHibernate.Test/NHSpecificTest/NH2030/Fixture.cs
MarkCiliaVincenti: Reverted

MarkCiliaVincenti:
> Thanks for the proposal,
> 
> [bUnit-dev/bUnit#1532](https://github.com/bUnit-dev/bUnit/pull/1532) has a discussion about this which is relevant for NHibernate too.
> 
> For the simple usages we currently have (excepted the Monitor one which would have to be slightly different), using something like:
> 
> ```cs
> #if NET_9_0_OR_GREATER
> private readonly Lock syncLock = new();
> #else
> private readonly object syncLock = new();
> #endif
> ```
> 
> Seems enough and would avoid adding a new dependency for users to investigate if they want it in their project or not.

Yes, it is very relevant.
https://github.com/bUnit-dev/bUnit/pull/1532#issuecomment-2323294989 in particular

Using the code you pasted above will give "[CS9216](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/lock-semantics?f1url=%3FappId%3Droslyn%26k%3Dk(CS9216)#lock-warning): A value of type System.Threading.Lock converted to a different type will use likely unintended monitor-based locking in lock statement." when you try using it in `MonitorLock` / `InternalLock` if you do it globally, plus you won't get the performance benefits of .NET 9.0 on that class if you keep using `Monitor`.

MarkCiliaVincenti:
As for your concern regarding the additional dependency, I understand, but what we can do is copy the code into NHibernate and just put in a comment at the top giving credit. It's not ideal, but if the dependency is a deal-breaker for you we can do that instead.

hazzik:
Thanks for the contribution. It is unlikely we'll be targeting .NET 9 version as it is not LTS. Also, we would like to avoid an external dependency.

MarkCiliaVincenti:
> Thanks for the contribution. It is unlikely we'll be targeting .NET 9 version as it is not LTS. Also, we would like to avoid an external dependency.

We could copy the code in. But if you don't intend on targeting .NET 9 that's another story because this would have to wait until .NET 10.

MarkCiliaVincenti:
@fredericDelaporte a new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency. If this new development makes you reconsider, I will amend this PR accordingly.

PR #3603, opened 2024-09-01
Allow deriving from DynamicMapEntityTuplizer

### LINKED ISSUE #3602: Unable to Implement Custom Tuplizer Due to Internal Constructor in DynamicMapEntityTuplizer
In the NHibernate documentation (Chapter 4.5. Tuplizers), it's mentioned that it's possible to extend NHibernate.Tuple.Entity.DynamicMapEntityTuplizer and create a custom tuplizer. However, when attempting to do so, I found that the constructor for this class is marked as internal, which prevents users from implementing a custom tuplizer if NHibernate is used as a NuGet package.

The constructor of DynamicMapEntityTuplizer should be public, allowing developers to extend this class and implement custom tuplizers as described in the documentation.

fredericDelaporte:
fix #3602

### DIFF
--- a/src/NHibernate/Tuple/Entity/DynamicMapEntityTuplizer.cs
+++ b/src/NHibernate/Tuple/Entity/DynamicMapEntityTuplizer.cs
@@ -13,7 +13,7 @@ namespace NHibernate.Tuple.Entity
 	{
 		private static readonly INHibernateLogger log = NHibernateLogger.For(typeof(PocoEntityTuplizer));
 
-		internal DynamicMapEntityTuplizer(EntityMetamodel entityMetamodel, PersistentClass mappingInfo)
+		public DynamicMapEntityTuplizer(EntityMetamodel entityMetamodel, PersistentClass mappingInfo)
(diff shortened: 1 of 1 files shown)


src/NHibernate/Tuple/Entity/DynamicMapEntityTuplizer.cs
```diff
@@ -13,7 +13,7 @@ public class DynamicMapEntityTuplizer : AbstractEntityTuplizer
 	{
 		private static readonly INHibernateLogger log = NHibernateLogger.For(typeof(PocoEntityTuplizer));
 
-		internal DynamicMapEntityTuplizer(EntityMetamodel entityMetamodel, PersistentClass mappingInfo)
+		public DynamicMapEntityTuplizer(EntityMetamodel entityMetamodel, PersistentClass mappingInfo)
```
fredericDelaporte, line 16: All other tuplizers, including the dynamic component one, have a public constructor. Let's fix this inconsistency.

hazzik (approved):

PR #3613, opened 2024-09-29
Fix ByCode OneToOne XML serialization

### LINKED ISSUE #3607: Invalid ByCode serialization to XML for OneToOne mappings
We've come across this issue today whilst upgrading from 5.3.7 to the latest version (5.5.2 at time of writing).  It's a crash error when using code like the following, if those mappings included any one-to-one relationships between entities.  The conversion-to-XML process appears to generate invalid XML which then crashes during XML schema validation.

```csharp
var mapper = new ModelMapper();
mapper.AddMappings(/* ... */);
var hbmMappings = mapper.CompileMappingForAllExplicitlyAddedEntities();
config.AddXml(hbmMappings.AsString());
```

Importantly, this affects the viability of the (AFAIK) only known workaround to #2307, which we are using.

I've created [a minimal self-contained reproduction case, hosted at this repository](https://github.com/craigfowler/NHibernate.XmlConversionBug).  I'll also reproduce the README for that repo below, as that contains a little more analysis.

---

Reproduction case for an NHibernate crash issue when converting Mapping By Code (MBC) mappings into XML, when those MBC mappings include any **One to one** mappings.
The `optimistic-lock` element/attribute is incorrectly serialized to XML and subsequently causes an XML schema validation error.

## Versions affected

I tried this out against a variety of NHibernate versions.

* 5.3.7 is unaffected
* 5.3.20 seems unaffected
* 5.4.0 seems unaffected
* 5.4.1 _reproduces the problem_
* 5.5.0 _reproduces the problem_
* _The problem is still present in 5.5.2_, which is the latest version at the time of writing

So, it would certainly seem that **5.4.1 is the first affected version**.

## Sample reproduction case

There are two unit tests in the test project.  One demonstrates the behavior using 'pure MBC' and the other uses the 'convert MBC mappings to XML' technique.
The remainder of the project is a minimal NHibernate entity & mapping setup with 3 entities, mappings and a small helper to create a session factory for them.
It uses SQLite in-memory driver/dialect so as to be self-contained.
As far as I can tell the DB driver/dialect is irrelevant though, because we reproduced this in an MS SQL Server environment.

1. Clone out the repository
2. Build with .NET 8 or higher: `dotnet build`
3. Run the unit tests: `dotnet test`
    * Optional, to run the tests against a specific NHibernate version set the `NhVersion` property to the desired NHibernate version
    * For example: `dotnet test /p:NhVersion=5.4.9`

### Expected behaviour

The tests for both techniques should pass (should not throw an exception).

### Actual behaviour

The 'pure MBC' test passes, but the test for the technique which converts to XML fails with an exception:

```text
Expected: No Exception to be thrown
  But was:  <NHibernate.MappingException: (string)(25,8): XML validation error: The element 'one-to-one' in namespace 'urn:nhibernate-mapping-2.2' has invalid child element 'OptimisticLock' in namespace 'urn:nhibernate-mapping-2.2'. List of possible elements expected: 'meta, formula' in namespace 'urn:nhibernate-mapping-2.2'.
 ---> System.Xml.Schema.XmlSchemaValidationException: The element 'one-to-one' in namespace 'urn:nhibernate-mapping-2.2' has invalid child element 'OptimisticLock' in namespace 'urn:nhibernate-mapping-2.2'. List of possible elements expected: 'meta, formula' in namespace 'urn:nhibernate-mapping-2.2'.
   at NHibernate.Cfg.Configuration.LogAndThrow(Exception exception)
   at NHibernate.Cfg.Configuration.ValidationHandler(Object o, ValidationEventArgs args)
   at System.Xml.Schema.XmlSchemaValidator.ValidateElementContext(XmlQualifiedName elementName, Boolean& invalidElementInContext)
   at System.Xml.Schema.XmlSchemaValidator.ValidateElement(String localName, String namespaceUri, XmlSchemaInfo schemaInfo, String xsiType, String xsiNil, String xsiSchemaLocation, String xsiNoNamespaceSchemaLocation)
   at System.Xml.XsdValidatingReader.ProcessElementEvent()
   at System.Xml.XsdValidatingReader.Read()
   at System.Xml.XmlLoader.LoadNode(Boolean skipOverWhitespace)
   at System.Xml.XmlLoader.LoadDocSequence(XmlDocument parentDoc)
   at System.Xml.XmlDocument.Load(XmlReader reader)
   at NHibernate.Cfg.Configuration.LoadMappingDocument(XmlReader hbmReader, String name)
   at NHibernate.Cfg.Configuration.AddXmlReader(XmlReader hbmReader, String name)
   at NHibernate.Cfg.Configuration.AddXml(String xml, String name)
   at NHibernate.Cfg.Configuration.AddXml(String xml)
   at NHibernate.XmlConversionBug.SessionFactoryCreator.GetSessionFactoryUsingXmlConversion() in C:\repos\NHibernate.XmlConversionBug\NHibernate.XmlConversionBug\SessionFactoryCreator.cs:line 21
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
--- End of stack trace from previous location ---
   at NUnit.Framework.Internal.ExceptionHelper.Rethrow(Exception exception)
   at NUnit.Framework.Internal.Reflect.DynamicInvokeWithTransparentExceptions(Delegate delegate)
   at NUnit.Framework.Internal.ExceptionHelper.RecordException(Delegate parameterlessDelegate, String parameterName)>

   at NHibernate.XmlConversionBug.Tests.SessionFactoryCreatorTests.CreatingASessionFactoryWithMbcConvertedToXmlShouldNotThrow() in C:\repos\NHibernate.XmlConversionBug\NHibernate.XmlConversionBug.Tests\SessionFactoryCreatorTests.cs:line 15
```

## Related info

This affects the viability of a workaround for another NHibernate bug; [issue 2307 on the NHibernate issue tracker](https://github.com/nhibernate/nhibernate-core/issues/2307).
That workaround to issue 2307 (which affects the capabilities of MBC mappings) is to use the 'convert MBC mappings to XML' technique.
That workaround is no longer viable for mappings which include any one to one relationships, because that would now cause a crash error.

fredericDelaporte:
fix #3607

### DIFF
--- a/src/NHibernate/Cfg/MappingSchema/HbmOneToOne.cs
+++ b/src/NHibernate/Cfg/MappingSchema/HbmOneToOne.cs
@@ -18,9 +18,13 @@ namespace NHibernate.Cfg.MappingSchema
 			get { return access; }
 		}
 
+		// 6.0 Todo : remove XmlIgnore after removing the setter. See #3607 fix.
+		[XmlIgnore]
 		public bool OptimisticLock
 		{
 			get => optimisticlock;
+			// Since v5.4.10
+			[Obsolete("Providing a setter for OptimisticLock was unintended and will be removed.")]
 			set => optimisticlock = value;
 		}
 
--- a/src/NHibernate.Test/NHSpecificTest/GH3607/FixtureByCode.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH3607/FixtureByCode.cs
@@ -1,0 +1,31 @@
+using NHibernate.Cfg;
+using NHibernate.Cfg.MappingSchema;
+using NHibernate.Mapping.ByCode;
+using NUnit.Framework;
+
+namespace NHibernate.Test.NHSpecificTest.GH3607
+{
+	/// <summary>
+	/// By code mapping serialization failure since v5.4.1. Adapted from <see href="https://github.com/craigfowler/NHibernate.XmlConversionBug" />.
+	/// </summary>
+	[TestFixture]
+	public class FixtureByCode : TestCaseMappingByCode
+	{
+		protected override HbmMapping GetMappings()
+		{
+			var mapper = new ModelMapper();
+			mapper.AddMappings(new[] { typeof(OrderMapping), typeof(LineItemMapping), typeof(LineItemDataMapping) });
+			return mapper.CompileMappingForAllExplicitlyAddedEntities();
+		}
+
+		[Test]
+		public void SerializeMappingToXml()
+		{
+			var mapping = GetMappings();
+			string serialized = "";
+			Assert.That(() => serialized = mapping.AsString(), Throws.Nothing, "Mapping serialization failure");
+			var config = new Configuration();
+			Assert.That(() => config.AddXml(serialized), Throws.Nothing, "Configuration with serialized mapping has failed");
+		}
+	}
+}
--- a/src/NHibernate.Test/NHSpecificTest/GH3607/LineItem.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH3607/LineItem.cs
@@ -1,0 +1,15 @@
+namespace NHibernate.Test.NHSpecificTest.GH3607
+{
+	public class LineItem
+	{
+		public virtual int Id { get; set; }
+
+		public virtual Order ParentOrder { get; set; }
+
+		public virtual string ItemName { get; set; }
+
+		public virtual decimal Amount { get; set; }
+
+		public virtual LineItemData Data { get; set; }
+	}
+}
--- a/src/NHibernate.Test/NHSpecificTest/GH3607/LineItemData.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH3607/LineItemData.cs
@@ -1,0 +1,9 @@
+namespace NHibernate.Test.NHSpecificTest.GH3607
+{
+	public class LineItemData
+	{
+		public virtual LineItem LineItem { get; set; }
+
+		public virtual string Data { get; set; }
+	}
+}
--- a/src/NHibernate.Test/NHSpecificTest/GH3607/LineItemDataMapping.cs
+++ b/src/NHibernate.Test/NHSpecificTest/GH3607/LineItemDataMapping.cs
@@ -1,0 +1,14 @@
+using NHibernate.Mapping.ByCode.Conformist;
+
+namespace NHibernate.Test.NHSpecificTest.GH3607
+{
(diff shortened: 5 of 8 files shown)


src/NHibernate/Cfg/MappingSchema/HbmOneToOne.cs
```diff
@@ -18,9 +18,13 @@ public string Access
...
 		}
 
+		// 6.0 Todo : remove XmlIgnore after removing the setter. See #3607 fix.
+		[XmlIgnore]
 		public bool OptimisticLock
 		{
 			get => optimisticlock;
+			// Since v5.4.10
+			[Obsolete("Providing a setter for OptimisticLock was unintended and will be removed.")]
 			set => optimisticlock = value;
```
fredericDelaporte, line 28: Providing a setter cause the property to be serialized to XML, which causes #3607. And there are no usage for a setter here.

fredericDelaporte: @bahusoid, that setter was added in #3216. Was there any reason for it I am missing?

hazzik (approved):

src/NHibernate/Cfg/MappingSchema/HbmOneToOne.cs
```diff
@@ -18,9 +18,13 @@ public string Access
 			get { return access; }
 		}
 
+		// 6.0 Todo : remove XmlIgnore after removing the setter. See #3607 fix.
```
hazzik, line 21: Do we actually care if it stays here or removed?

src/NHibernate/Cfg/MappingSchema/HbmOneToOne.cs
bahusoid: Nope. It's some mistake.

src/NHibernate/Cfg/MappingSchema/HbmOneToOne.cs
fredericDelaporte: For consistency with other mapping classes like `HbmManyToOne`, I would rather remove it.

src/NHibernate/Cfg/MappingSchema/HbmOneToOne.cs
fredericDelaporte: Ok, thanks.

PR #3629, opened 2024-11-27
Don't join fkey table when linq comparison with composite id

### LINKED ISSUE #3676: When referencing a composite-id in Linq, the fkey table is wrongly joined
This behavior is inconsistent with HQL and results in a mild performance degradation when Linq is used over HQL. I've submitted a patch along with unit tests (#3629) and am just waiting for someone to approve.

jeff-hagen:
When referencing a foreign key composite-id in Linq, don't add a join with the fkey table to the resulting SQL (make behavior consistent with the same query in HQL). Fixes #3676


jeff-hagen:
@fredericDelaporte would you mind taking a look at my PR?

jeff-hagen:
OK. Thank you. 

> About *.sql files, file encoding changes are not encouraged. They are harder to review and break somewhat the history when looking for significant changes.
> 
> This said, as the encoding was not consistent with the 2012 files, let it be.
> 
> LGTM.

PR #3632, opened 2024-12-10
Fix version update on dirty collection with another property having optimistic-lock false

### LINKED ISSUE #3631: Missing version update on dirty collection with another property having optimistic-lock false
I have an entity with ordinary properties, a property with optimistic-lock="false" and a set (implicitly with optimistic-lock="true" as it's the default value).

If I just add a child entity to the collection, the version is incremented correctly but if I change a property without optimistic lock at the same time, the version is not incremented.

The version must be incremented if at least one property (including collections) with optimistic-lock="true" is changed.

The reason is that ```FlushEntityEvent.HasDirtyCollections``` is not filled in case there dirty properties but this doesn't take into consideration whether the dirty properties lead to a version increment. Calling ```DefaultFlushEntityEventListener.HasDirtyCollections``` is necessary because of the side effect it has on the ```Event```.

wvopt:
Incrementing Version on parent entity for a collection change works correctly even with optimistic-lock="false" on another property of the same entity. Call to HasDirtyCollections has a side effect thus must be called independently from the existence of other dirty properties.

Fixes #3631

### DIFF
--- a/src/NHibernate.Test/Async/VersionTest/Db/DbVersionFixture.cs
+++ b/src/NHibernate.Test/Async/VersionTest/Db/DbVersionFixture.cs
@@ -52,6 +52,7 @@ namespace NHibernate.Test.VersionTest.Db
 			admin = await (s.GetAsync<Group>(admin.Id));
 			guy.Groups.Add(admin);
 			admin.Users.Add(guy);
+			guy.NoOptimisticLock = "changed";
 			await (t.CommitAsync());
 			s.Close();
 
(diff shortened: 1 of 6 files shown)


fredericDelaporte:
Sorry for the long delay,,

LGTM, but why is it another account opening the pull request than the fix author?

fredericDelaporte (approved):

csharper2010:
Thanks for your time!

> Sorry for the long delay,,

No problem, We are working with a patched version until the fix is officially available.

> LGTM, but why is it another account opening the pull request than the fix author?

I mixed up two accounts I need for different purposes.

fredericDelaporte:
As a result, the merged commit will have the other account as an author. It seems you need to open a new PR if you want to avoid that. Let us know.

csharper2010:
> As a result, the merged commit will have the other account as an author. It seems you need to open a new PR if you want to avoid that. Let us know.

No need to change that... go ahead.

Issue #3618, opened 2024-10-18
ISQLQuery.ExecuteUpdate produces an exception in certain cases


dimako:
NHibernate v5.5.2, .Net Framework 4.8

Table definition:
`sql CREATE TABLE JustText(c nvarchar(max))`

Failing test
```c#
        [TestMethod]
        public void ItShouldInsertTwoRecords()
        {
            var config = new Configuration();
            config.SetProperty(NHibernate.Cfg.Environment.ConnectionString, "Server=localhost;Database=Sandbox;Trusted_Connection=True;TrustServerCertificate=True");
            config.SetProperty(NHibernate.Cfg.Environment.ConnectionDriver, "NHibernate.Driver.SqlClientDriver");
            config.SetProperty(NHibernate.Cfg.Environment.Dialect, "NHibernate.Dialect.MsSql2012Dialect, NHibernate");
            using (var factory = config.BuildSessionFactory())
            {
                using (var session = factory.OpenSession())
                {
                    session.CreateSQLQuery(@"begin
INSERT INTO JustText (c)VALUES (N'
--');
INSERT INTO JustText (c)VALUES (N'
P:
');
end;").ExecuteUpdate();
                }
            }

        }
```

The test is failing with the exception:

```
NHibernate.Exceptions.GenericADOException: could not execute native bulk manipulation query:begin
INSERT INTO JustText (c)VALUES (N'
--');
INSERT INTO JustText (c)VALUES (N'
P:
');
end;[SQL: SQL not available] ---> System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.

  Stack Trace: 
ThrowHelper.ThrowKeyNotFoundException()
Dictionary`2.get_Item(TKey key)
NamedParameterSpecification.SetEffectiveType(QueryParameters queryParameters)
ParametersBackTrackExtensions.ResetEffectiveExpectedType(IEnumerable`1 parameterSpecs, QueryParameters queryParameters)
NativeSQLQueryPlan.PerformExecuteUpdate(QueryParameters queryParameters, ISessionImplementor session)
--- End of inner exception stack trace ---
NativeSQLQueryPlan.PerformExecuteUpdate(QueryParameters queryParameters, ISessionImplementor session)
SessionImpl.ExecuteNativeUpdate(NativeSQLQuerySpecification nativeQuerySpecification, QueryParameters queryParameters)
SqlQueryImpl.ExecuteUpdate()
```

[...]

PR #3637, opened 2025-01-02
GH3618 - Failed recognition of SQL comments

gliljas:
#3618

### DIFF
--- a/src/NHibernate/Engine/Query/ParameterParser.cs
+++ b/src/NHibernate/Engine/Query/ParameterParser.cs
@@ -48,47 +48,47 @@ namespace NHibernate.Engine.Query
 
 			int stringLength = sqlString.Length;
 			bool inQuote = false;
-			bool afterNewLine = false;
 			for (int indx = 0; indx < stringLength; indx++)
 			{
 				int currentNewLineLength;
 
-				// check comments
-				if (indx + 1 < stringLength && sqlString.Substring(indx,2) == "/*")
+				// check comments, unless in quote
+				if (!inQuote)
 				{
-					var closeCommentIdx = sqlString.IndexOf("*/", indx + 2, StringComparison.Ordinal);
-					recognizer.Other(sqlString.Substring(indx, (closeCommentIdx- indx)+2));
-					indx = closeCommentIdx + 1;
-					continue;
-				}
-				
-				if (afterNewLine && (indx + 1 < stringLength) && sqlString.Substring(indx, 2) == "--")
-				{
-					var closeCommentIdx = sqlString.IndexOfAnyNewLine(indx + 2, out currentNewLineLength);
-						
-					string comment;
-					if (closeCommentIdx == -1)
+					if (indx + 1 < stringLength && sqlString.Substring(indx, 2) == "/*")
 					{
-						closeCommentIdx = sqlString.Length;
-						comment = sqlString.Substring(indx);
+						var closeCommentIdx = sqlString.IndexOf("*/", indx + 2, StringComparison.Ordinal);
+						recognizer.Other(sqlString.Substring(indx, (closeCommentIdx - indx) + 2));
+						indx = closeCommentIdx + 1;
+						continue;
 					}
-					else
+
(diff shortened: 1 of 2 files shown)


src/NHibernate/Engine/Query/ParameterParser.cs
```diff
@@ -48,47 +48,47 @@ public static void Parse(string sqlString, IRecognizer recognizer)
...
 			bool inQuote = false;
-			bool afterNewLine = false;
 			for (int indx = 0; indx < stringLength; indx++)
 			{
 				int currentNewLineLength;
 
-				// check comments
-				if (indx + 1 < stringLength && sqlString.Substring(indx,2) == "/*")
+				// check comments, unless in quote
+				if (!inQuote)
```
fredericDelaporte, line 56: We do this look ahead up to two times, it may be worth introducing a local here and guarding furthermore the test.

```suggestion
				// check comments, unless in quote or at end of string
				if (!inQuote && indx + 1 < stringLength)
```

src/NHibernate/Engine/Query/ParameterParser.cs
```diff
@@ -48,47 +48,47 @@ public static void Parse(string sqlString, IRecognizer recognizer)
...
-					continue;
-				}
-				
-				if (afterNewLine && (indx + 1 < stringLength) && sqlString.Substring(indx, 2) == "--")
-				{
-					var closeCommentIdx = sqlString.IndexOfAnyNewLine(indx + 2, out currentNewLineLength);
-						
-					string comment;
-					if (closeCommentIdx == -1)
+					if (indx + 1 < stringLength && sqlString.Substring(indx, 2) == "/*")
```
fredericDelaporte, line 58:
```suggestion
					var candidateOpenCommentToken = sqlString.Substring(indx, 2);
					if (candidateOpenCommentToken == "/*")
```

src/NHibernate/Engine/Query/ParameterParser.cs
```diff
@@ -48,47 +48,47 @@ public static void Parse(string sqlString, IRecognizer recognizer)
...
-						closeCommentIdx = sqlString.Length;
-						comment = sqlString.Substring(indx);
+						var closeCommentIdx = sqlString.IndexOf("*/", indx + 2, StringComparison.Ordinal);
+						recognizer.Other(sqlString.Substring(indx, (closeCommentIdx - indx) + 2));
+						indx = closeCommentIdx + 1;
+						continue;
 					}
-					else
+
+					if ((indx + 1 < stringLength) && sqlString.Substring(indx, 2) == "--")
```
fredericDelaporte, line 66:
```suggestion
					if (candidateOpenCommentToken == "--")
```

src/NHibernate/Engine/Query/ParameterParser.cs
```diff
@@ -48,47 +48,47 @@ public static void Parse(string sqlString, IRecognizer recognizer)
...
+				// check comments, unless in quote
+				if (!inQuote)
 				{
-					var closeCommentIdx = sqlString.IndexOf("*/", indx + 2, StringComparison.Ordinal);
-					recognizer.Other(sqlString.Substring(indx, (closeCommentIdx- indx)+2));
-					indx = closeCommentIdx + 1;
-					continue;
-				}
-				
-				if (afterNewLine && (indx + 1 < stringLength) && sqlString.Substring(indx, 2) == "--")
```
fredericDelaporte, line 65: So, this parser only accounted for single line comments starting at the beginning of the line. Any thought on why was that? Just a false assumption of who has coded that? Or some db actually handling it that way?

fredericDelaporte (approved):
Some old redundant code could be slightly optimized by the way, but LGTM.

src/NHibernate/Engine/Query/ParameterParser.cs
gliljas: I think it was a false assumption.

fredericDelaporte (approved):

Issue #3643, opened 2025-01-22
Query cache + entity with Enum id + child collection => NullReferenceException when loading from second level cache


cwatzl:
After upgrading an old legacy project from NH 5.2.7 to 5.5.2 I encountered a peculiar regression bug:

- Root Entity has an Id property having an enum type, and a one-to-many relationship with another entity
- Second level cache and query cache is used for querying the Entity; child collection is fetched eagerly in the query in question
- First execution of the query (with unpopulated cache) works; subsequent executions fail with `NullReferenceException` thrown in `TypeHelper`.

Example model:

```c#
	class Entity
	{
		private readonly ICollection<ChildEntity> _children = new List<ChildEntity>();
		public virtual EntityId Id { get; protected set; }
		public virtual IEnumerable<ChildEntity> Children => _children.AsEnumerable();
	}

	class ChildEntity
	{
		public virtual int Id { get; protected set; }
	}

	enum EntityId
	{
		Id1,
		Id2
	}
```

Mapping:

```c#
mapper.Class<Entity>(
	rc =>
	{
		rc.Id(x => x.Id);
		rc.Bag(
			x => x.Children,
			m =>
			{
				m.Access(Accessor.Field);
				m.Key(k => k.Column("EntityId"));
			},
			r => r.OneToMany());
		rc.Cache(
			cm =>
			{
				cm.Include(CacheInclude.All);
				cm.Usage(CacheUsage.ReadWrite);
			});
	});

mapper.Class<ChildEntity>(
	rc =>
	{
		rc.Id(x => x.Id);
		rc.Cache(
			cm =>
			{
				cm.Include(CacheInclude.All);
				cm.Usage(CacheUsage.ReadWrite);
			});
	});
```

Query:

```c#
session
    .Query<Entity>()
    .FetchMany(x => x.Children)
    .WithOptions(opt => opt.SetCacheable(true))
    .ToList();
```

Exception thrown on second and all subsequent query executions:

```
NHibernate.Exceptions.GenericADOException : Could not execute query[SQL: SQL not available]
  ----> System.NullReferenceException : Object reference not set to an instance of an object.
   at NHibernate.Impl.SessionImpl.List(IQueryExpression queryExpression, QueryParameters queryParameters, IList results, Object filterConnection) in C:\dev\nhibernate-core\src\NHibernate\Impl\SessionImpl.cs:line 563
[...]

PR #3644, opened 2025-01-22
Provide test case for #3643

cwatzl:
Test case for #3643.


fredericDelaporte:
Thanks for the test case. It has been put into #3693.

PR #3653, opened 2025-02-24
AbstractDateTimeType.Round is a Floor not a Round

johnboyerkarbonhq:
I was messing around with `AbstractDateTimeType` and couldn't help but notice that the method `Round` is intended to Round but it actually is doing a floor. Thought I might quickly pop a PR in.


hazzik:
I think this was intentional. I don't think it is semantically correct to round time up, because it potentially could produce a future time.

Issue #3657, opened 2025-03-08
SessionFactoryObjectFactory Early Cache Population Causes Threading Issue


mjkent:
In https://github.com/nhibernate/nhibernate-core/issues/2190 the time that the `SessionFactoryObjectFactory` cache for an instance was moved up before the `SessionFactory` is built. This can mean that during the `ISessionFactory` constructor, `SessionFactoryObjectFactory.GetNamedInstance` can return a partially constructed `ISessionFactory`

In the case where application start wasn't the right time to create the `ISessionFactory` and double checked locking was recommended. An example:
```
private static ISessionFactory GetSessionFactory(sessionFactoryName)
{
    var sessionFactory = SessionFactoryObjectFactory.GetNamedInstance(sessionFactoryName);
    if (sessionFactory == null)
    {
        lock (_sessionFactoryApplicationLock)
        {
            sessionFactory = SessionFactoryObjectFactory.GetNamedInstance(sessionFactoryName);
            if (sessionFactory == null)
            {
                sessionFactory = GetConfig(sessionFactoryName).BuildSessionFactory();
            }
        }
    }
    return sessionFactory;
}
```

**When multiple threads hit this at the same time, please imagine:**
Thread 1 comes in, `SessionFactoryObjectFactory.GetNamedInstance` returns null.
Thread 1 acquires the lock the second `SessionFactoryObjectFactory.GetNamedInstance` returns null.
Thread 1 starts building the session factory. Early in the `ISessionFactory` [the cache is populated with a not yet built `SessionFactory`](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Impl/SessionFactoryImpl.cs#L242)
Thread 2 calls `SessionFactoryObjectFactory.GetNamedInstance` and gets the partially built `ISessionFactory`
Bad behavior: Thread 1 is writing to the various dictionaries while thread 2 is reading them. We've seen `entityPersisters` not get fully populated(which populates one at a time in a loop on a normal dictionary) while `classMetadata` is populated(which populates a temporary dictionary and then uses a readonly dictionary). Dictionaries are thread safe if multiple threads are reading from them at once, but if any writing is going on at that time, the behavior is undefined and will break

**Suggested improvement**
[...]

PR #3658, opened 2025-03-08
Make Most of the Dictionaries in SessionFactoryImpl ReadOnly Improves #3657

mjkent:
Requesting comments on this PR that improves a regression documented in #3657

The core of this is to make `SessionFactoryImpl` dictionary's read only - this makes the behavior easier to understand(and safer) for those who accidentally do it(in the issue I provided a case that a change to the NH source caused this to be a new behavior on the existing use of the API in a recommended/best practice way). It doesn't fully fix the "issue" - because `SessionFactoryImpl` is meant to not be created multiple at once, so it's not an issue in that way, but giving more deterministic and easier to debug behavior on a regression is an improvement here. It also means that instead of having a half constructed object(that has some Dictionary's populated and others half populated, it will either be null or set correctly).

There is also an unlike case of writing and reading from `[SessionFactoryObjectFactory.cs`'s dictionary's at the same time, which is documented as an undefined behavior - in practice it can cause very odd issues like dictionaries losing values.

I was sure what the NH team has set for rules on creating tests around these. Somewhat tough to create a failing test because of the non-deterministic behavior.

It would be appreciated if this could be in v5.5.y? I paid attention to keep the public API's the same and keep all existing behavior through the public API's.

Note: I did read through contributing and ran all the SqlServer tests. I have some failing local on both master and this branch that are the same(GH3516) - I assume it's an environment issue.

### DIFF
--- a/src/NHibernate/Impl/SessionFactoryImpl.cs
+++ b/src/NHibernate/Impl/SessionFactoryImpl.cs
@@ -265,7 +265,7 @@ namespace NHibernate.Impl
 			#region Persisters
 
 			var caches = new Dictionary<Tuple<string, string>, ICacheConcurrencyStrategy>();
-			entityPersisters = new Dictionary<string, IEntityPersister>();
+			var tmpEntityPersisters = new Dictionary<string, IEntityPersister>();
 			implementorToEntityName = new Dictionary<System.Type, string>();
 
 			Dictionary<string, IClassMetadata> classMeta = new Dictionary<string, IClassMetadata>();
@@ -280,7 +280,7 @@ namespace NHibernate.Impl
 					model.EntityName,
 					caches);
 				var cp = PersisterFactory.CreateClassPersister(model, cache, this, mapping);
-				entityPersisters[model.EntityName] = cp;
+				tmpEntityPersisters[model.EntityName] = cp;
 				classMeta[model.EntityName] = cp.ClassMetadata;
 
 				if (model.HasPocoRepresentation)
@@ -289,6 +289,8 @@ namespace NHibernate.Impl
 				}
 			}
 
+			entityPersisters = new ReadOnlyDictionary<string, IEntityPersister>(tmpEntityPersisters);
+
 			entityPersistersSpaces = entityPersisters
 				.SelectMany(x => x.Value.QuerySpaces.Select(y => new { QuerySpace = y, Persister = x.Value }))
 				.ToLookup(x => x.QuerySpace, x => x.Persister);
@@ -296,7 +298,7 @@ namespace NHibernate.Impl
 			classMetadata = new ReadOnlyDictionary<string, IClassMetadata>(classMeta);
 
 			Dictionary<string, ISet<string>> tmpEntityToCollectionRoleMap = new Dictionary<string, ISet<string>>();
-			var collPersisters = new Dictionary<string, ICollectionPersister>();
+			var tmpCollectionPersisters = new Dictionary<string, ICollectionPersister>();
 			foreach (Mapping.Collection model in cfg.CollectionMappings)
 			{
 				var cache = GetCacheConcurrencyStrategy(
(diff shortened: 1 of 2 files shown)


src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -130,27 +130,27 @@ public void HandleEntityNotFound(string entityName, string propertyName, object
...
+		private readonly IReadOnlyDictionary<System.Type, string> implementorToEntityName;
 
 		[NonSerialized]
 		private readonly EventListeners eventListeners;
 
 		[NonSerialized]
 		private readonly Dictionary<string, FilterDefinition> filters;
 		[NonSerialized]
-		private readonly Dictionary<string, IIdentifierGenerator> identifierGenerators;
+		private readonly IReadOnlyDictionary<string, IIdentifierGenerator> identifierGenerators;
```
fredericDelaporte, line 141: Causes later code to no more compile.

src/NHibernate/Impl/SessionFactoryObjectFactory.cs
```diff
@@ -21,8 +22,8 @@ public static class SessionFactoryObjectFactory
 	{
 		private static readonly INHibernateLogger log;
 
-		private static readonly IDictionary<string, ISessionFactory> Instances = new Dictionary<string, ISessionFactory>();
-		private static readonly IDictionary<string, ISessionFactory> NamedInstances = new Dictionary<string, ISessionFactory>();
+		private static readonly IDictionary<string, ISessionFactory> Instances = new ConcurrentDictionary<string, ISessionFactory>();
+		private static readonly IDictionary<string, ISessionFactory> NamedInstances = new ConcurrentDictionary<string, ISessionFactory>();
```
fredericDelaporte, line 26: This looks to me legit, while all other changes look superfluous to me. May you isolate this in a standalone PR?

src/NHibernate/Impl/SessionFactoryImpl.cs
```diff
@@ -105,23 +105,23 @@ public void HandleEntityNotFound(string entityName, string propertyName, object
 			new ConcurrentDictionary<string, CacheBase>();
 
 		[NonSerialized]
-		private readonly IDictionary<string, IClassMetadata> classMetadata;
+		private readonly IReadOnlyDictionary<string, IClassMetadata> classMetadata;
```
fredericDelaporte, line 108: All theses changes look superfluous to me, because this will never be enough to get the build of a session factory threadsafe with concurrent uses of the factory being built. The only reasonable fix is to not end up using a factory that is concurrently being built.

src/NHibernate/Impl/SessionFactoryObjectFactory.cs
fredericDelaporte: Well, actually no, these changes are superfluous too, because all methods of the type are synchronized, meaning no concurrent access can occur.

fredericDelaporte:
These changes do not actually fix the reported issue and do not improve the robustness of the code:
 - A session factory is no meant to be used concurrently to its own build. No change should be made attempting to support this.
 - Changing a dictionary for a concurrent one while using it through a non concurrent interface does not improve concurrency support.
 - Using a concurrent dictionary in an already fully synchronized class (all methods being synchronized) is redundant and unneeded.

mjkent:
@fredericDelaporte I'm hoping you'll reconsider.

For the changes to `src/NHibernate/Impl/SessionFactoryImpl.cs` - I think there is some benefit and low risk. 
One major benefit is that future developers don't accidentally do something to change the values in the List's that the changes protect. Since use(and not construction) of the `ISessionFactory` is often multi-threaded, explicitly using data structures that protect this capability has value. Once built, these `Dictionary`'s should not be added to/changed.

It's a pretty low cost change for this. I'd be happy to provide a PR with just those changes.


On the changes to `SessionFactoryObjectFactory` - I see your meaning. I missed the attribute on the methods contained.

PR #3662, opened 2025-03-13
Run tests using .NET 10

hazzik:


cremor:
Idea: Add tests to see how NHibernate handles the new APIs [Queryable.LeftJoin](https://learn.microsoft.com/en-us/dotnet/api/system.linq.queryable.leftjoin?view=net-10.0) and [Queryable.RightJoin](https://learn.microsoft.com/en-us/dotnet/api/system.linq.queryable.rightjoin?view=net-10.0) that were added in .NET 10.

`RightJoin` will most likely throw an unsupported exception, so maybe that is not relevant.
But `LeftJoin` is already available as an extension method in NHibernate itself. What happens when targeting .NET 10, is it still using the NHibernate method, or is it picking up the .NET 10 one (if both namespaces are used)?

ValResnick:
Hi,

is there an estimated release date for a version that will be compatible with .NET 10?
We would love to upgrade our project to .NET10.

fredericDelaporte:
I have push a commit to install .net 10 on Appveyor, but now realize you had previously some script for doing that too and reverted it. Still, AppVeyor images do not have it installed by default as of now (https://github.com/appveyor/ci/issues/3984, https://www.appveyor.com/docs/windows-images-software/#net-framework).

What was the reason for the revert?

fredericDelaporte (dismissed):

fredericDelaporte:
The macos 13 check will never be fulfilled. There are no more runners for macos 13, it seems.
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

hazzik:
> What was the reason for the revert?

it was installing preview version. I should have checked if appveyor have a release version of .net10 before removing it completely

hazzik:
I've reordered and reorganized commits removing unnecessary churn

fredericDelaporte (dismissed):

PR #3640, opened 2025-01-19
Some string.Intern's for minimizing duplicate strings count

AlexanderKot:
After startup application that I develop now has ~9% of duplicate strings from all allocated objects.
1st consumer is ASP.NET MVC routing,  NHib related strings seems on 2d place
image
and many others

1.
| Value | Wasted | Count | CountAfter |
|---|---|---|---|
| System.String, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e | 491,36 KB | 2420 | 2 |
| System.Boolean, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e | 281,78 KB | 1375 | 2 |
| NHibernate.Envers.Entities.RevisionTypeType, NHibernate.Envers, Version=5.5.2.0, Culture=neutral, PublicKeyToken=e2c5b946037fb7f8 | 212,9 KB | 846 |
| Int64 | 8,41 KB	 | 862 | 8 |
| Int32 | 6,97 KB	 | 715 | 6 |

image
Minimized by interning **SimpleValue.typeName**

2.
| Value | Wasted | Count | CountAfter |
|---|---|---|---|
| NN.Core.BL.NhibTune.IdGenerators.LongIdGenerator, Core.BL, Version=9.0.0.1, Culture=neutral, PublicKeyToken=null | 61,25 KB | 281 | - |

image
Minimized by interning **SimpleValue.identifierGeneratorStrategy**

3.
| Value | Wasted | Count | CountAfter |
|---|---|---|---|
| originalId.REV | 55,21 KB | 2020 | - |
| originalId.REV.id | 16,77 KB | 506 | 2 |
| originalId.REV.Id | 16,73 KB | 505 | - |

image
image
image
Minimized by interning in 
**AbstractPropertyMapping .ExtendPath
EntityMetamodel.MapIdentifierPropertyTypes
AbstractEntityPersister. InitSubclassPropertyAliasesMap**

4.
| Value | Wasted | Count | CountAfter |
|---|---|---|---|
| 0_ | 31,14 KB | 7973 | - |
| 1_ | 2,75 KB | 704 | 6 |

image
image
image
Minimized by generating suffix with interning
**StringHelper.GenerateSuffix**

5.
| Value | Wasted | Count | CountAfter |
|---|---|---|---|
| Id | 8,57 KB | 2195 | 1867 | 
[...]

PR #3663, opened 2025-03-18
Make StringHelper.GenerateSuffix internal

hazzik:


src/NHibernate/Util/StringHelper.cs
```diff
@@ -872,7 +872,7 @@ public static bool IsAnyNewLine(this string str, int index, out int newLineLengt
 			return false;
 		}
 
-		public static string GenerateSuffix(int index)
+		internal static string GenerateSuffix(int index)
```
fredericDelaporte, line 875: Introduced in #3640, not yet released, so, ok.

fredericDelaporte (approved):

src/NHibernate/Util/StringHelper.cs
hazzik: Thanks. I've actually done the change in the PR, but forgot to push :-/

PR #3668, opened 2025-04-06
Add GetOrBuild on the session factory factory

### LINKED ISSUE #3657: SessionFactoryObjectFactory Early Cache Population Causes Threading Issue

fredericDelaporte:
fix #3657

### DIFF
--- a/src/NHibernate/Impl/SessionFactoryObjectFactory.cs
+++ b/src/NHibernate/Impl/SessionFactoryObjectFactory.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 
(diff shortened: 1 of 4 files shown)


src/NHibernate/Impl/SessionFactoryObjectFactory.cs
```diff
@@ -78,16 +79,46 @@ public static void RemoveInstance(string uid, string name, IDictionary<string, s
...
 		/// <returns>An instantiated ISessionFactory.</returns>
+		/// <remarks>
+		/// <para>It is the caller responsibility to ensure <paramref name="instanceBuilder"/>
+		/// will add and yield a session factory of the requested <paramref name="name"/>.</para>
+		/// <para>If the session factory instantiation is performed concurrently outside of a
+		/// <c>GetOrAddNamedInstance</c> call, this method may yield an instance of it still being
+		/// built, which may lead to threading issues.</para>
+		/// </remarks>
+		[MethodImpl(MethodImplOptions.Synchronized)]
+		public static ISessionFactory GetOrBuildNamedInstance(string name, Func<ISessionFactory> instanceBuilder)
```
fredericDelaporte, line 97: To allow threadsafe use of the factory, a get or build is needed. Some applications do use it in a multi-thread context. They were always exposed to threading bugs due to the get being able of yielding a not yet fully built instance in such case, but this was worsen by #2190. With this `GetOrBuild`, provided the application only build factories through it, fully built instances will be yielded.

hazzik (approved):

Issue #2005, opened 2019-02-06
Implement LockMode.UpgradeSkipLocked


beppler:
Hibernate has an additional lock mode called UPGRADE_SKIPLOCKED described as:

The lock acquisition request skips the already locked rows. It uses a SELECT …​ FOR UPDATE SKIP LOCKED in Oracle and PostgreSQL 9.5, or SELECT …​ with (rowlock, updlock, readpast) in SQL Server.

This kind of lock is very usefull quen there are many consumers for a table that is used as a queue.

Documentation for Hibernate [version 5](http://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#locking-LockMode)

Documentation for Hibernate [version 4](https://docs.jboss.org/hibernate/orm/4.3/devguide/en-US/html/ch05.html#d5e1372)

beppler:
First implemented on Hibernate [here](https://github.com/hibernate/hibernate-orm/commit/e0cfc6bf2ef614c5dc762a508205cc4812122989#diff-f39934fc5ee42d01adf51d64771b5738)

PR #3681, opened 2025-07-21
Add LockMode.UpgradeSkipLocked (#2005)

PleasantD:
* Added a new PostgreSQL dialect
* Couldn't find an early enough Transact-SQL reference for READPAST, so added it to `MsSql2005Dialect`


src/NHibernate/nhibernate-mapping.xsd
```diff
@@ -1637,6 +1637,7 @@
 			<xs:enumeration value="read" />
 			<xs:enumeration value="upgrade" />
 			<xs:enumeration value="upgrade-nowait" />
+			<xs:enumeration value="upgrade-skiplocked" />
```
hazzik, line 1640: Adding this in the middle renumbers the enum values

PR #3684, opened 2025-08-13
Add Check(string) to IClassAttributesMapper<TEntity>

nats:
In NHibernate.Mapping.ByCode, the class Check() method is implemented but not exposed for use.

Sample code:

```cs
            mapper.Class<Entity1>(m => {
                m.Check("arbitrary sql check");
            });
```

This commit adds the Check() method to correct interface for above sample to work.

### DIFF
--- a/src/NHibernate/Mapping/ByCode/IClassMapper.cs
+++ b/src/NHibernate/Mapping/ByCode/IClassMapper.cs
@@ -58,6 +58,7 @@ namespace NHibernate.Mapping.ByCode
 		void Discriminator(Action<IDiscriminatorMapper> discriminatorMapping);
 		void DiscriminatorValue(object value);
 		void Table(string tableName);
+		void Check(string check);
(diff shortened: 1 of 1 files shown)


src/NHibernate/Mapping/ByCode/IClassMapper.cs
```diff
@@ -58,6 +58,7 @@ public interface IClassAttributesMapper<TEntity> : IEntityAttributesMapper, IEnt
 		void Discriminator(Action<IDiscriminatorMapper> discriminatorMapping);
 		void DiscriminatorValue(object value);
 		void Table(string tableName);
+		void Check(string check);
```
hazzik, line 61: This technically changes the interface and would be a binary breaking change. Could you please instead create an extension method for it? Also, please add some tests.

hazzik (changes requested):

PR #3685, opened 2025-08-14
Fix filtering sub-queries on property-ref

### LINKED ISSUE #3609: Fitering with a subquery on a many-to-one with property-ref generates invalid SQL
I have found this issue whilst attempting to upgrade from NHibernate 5.3.7 to 5.5.2.  This issue has caused us to abort the upgrade, as it causes a number of our app's queries to return incorrect results.

I have created a minimal reproduction case for the issue [and published it in this git repository](https://github.com/craigfowler/NHibernate.PropertyRefBug).  I have also reproduced the README for it below, as that contains details about the issue.

---
# NHibernate HBM `property-ref` bug

This is a reproduction case for an NHibernate bug whereby a `property-ref` attribute on a `<many-to-one />` XML HBM mapping is incorrectly used when generating SQL.
This appears to affect only queries which would cause an `EXISTS( ... )`-style SQL query to be generated by NHibernate.

The issue differs a little depending upon whether or not the `<many-to-one />` mapping uses the `not-found="ignore"` attribute.

The incorrect SQL appears within that subquery. It relates to the column from the table which is the main subject of the subquery, corresponding to the foreign key column of the table which has the many-to-one mapping.  NHibernate generates SQL using the primary key column, where it should instead use the column corresponding to the property identified by the `property-ref` attribute.

## Versions affected

I have tried this out against a variety of NHibernate versions.
In our apps, where we are affected by this issue, we have also been using `not-found="ignore"` on the same mappings.
Whilst investigating/creating the reproduction case, I discovered that if I remove the `not-found` attribute from the mappings _then I can reproduce this problem even in version 5.3.7_.

But - for the scenario which affects us, when we are using `not-found="ignore"`:

* 5.3.7 is unaffected
* 5.5.2 _reproduces the problem_ (this is current latest version at time of writing)
* 5.4.1 seems unaffected
* 5.4.3 _reproduces the problem_
* 5.3.15 seems unaffected
* 5.3.17 _reproduces the problem_

It seems like **5.3.17** along with 5.4.3 are the first affected versions.
Muddying the picture a little, _there is another related issue, predating this, affecting versions 5.3.16 & 5.4.2 only_.  See below for more info.
This means that if we wish to see a sample of 'good' SQL for the query, we must use 5.3.15 or 5.4.1, or earlier.

### Another related issue, predating this

There is a separate-but-related issue to this which affects NH 5.3.16 and 5.4.2.
[That issue might be #3269](https://github.com/nhibernate/nhibernate-core/issues/3269) but I can't be 100% certain just from a read of the ticket.
The related issue means that the test case included in this repository will pass on those two versions, _but NHibernate would still generate incorrect SQL, causing other problems_.

It's best to consider _the "last good" versions_ to be 5.3.15 & 5.4.1 with regard to the issue that this repo demonstrates.

## Sample reproduction case

There are two mapped classes in this solution, including a mapping which reproduces this issue:

* It uses `not-found="ignore"`
* It uses `property-ref="..."`

One of the tests included in the solution executes a query which triggers generation of a SQL `EXISTS(...)` subquery.

This solution uses SQLite in-memory driver/dialect so as to be self-contained.
As far as I can tell the DB driver/dialect is irrelevant though, because we reproduced this in an MS SQL Server environment.

To run this reproduction case:

1. Clone the repository
2. Run the tests using .NET 8 or higher: `dotnet test`
    * Optional, to run the tests against a specific NHibernate version set the `NhVersion` property to the desired NHibernate version
    * For example: `dotnet test /p:NhVersion=5.4.9`

The mappings include an interceptor which will capture and emit the SQL which is generated to STDERR.
This is to help illustrate/diagnose the problem.

### Expected behaviour

Both unit tests should pass; the queries performed by the tests are querying against known data which has been added to the in-memory DB.

### Actual behaviour

The unit test for the subquery fails.

### Further info/analysis

Note that in passing scenarios (NH version 5.3.15) the SQL emitted by the interceptor is as follows.

```sql
select cast(count(*) as INTEGER) as col_0_0_
from LineItem lineitem0_
where exists (
    select order1_.Id
    from TheOrder order1_, TheOrder order2_
    where
        lineitem0_.OrderId=order2_.UniqueId
        and order1_.CreatedDate>?
        and order1_.Id=order2_.Id
    )
```

However in failing scenarios (NH >= 5.3.17) the SQL emitted by the interceptor is as follows.  The problematic area is pointed out in a comment.

```sql
select cast(count(*) as INTEGER) as col_0_0_
from LineItem lineitem0_
where exists (
    select order1_.Id
    from TheOrder order1_
    where
        order1_.CreatedDate>?
        and order1_.Id=lineitem0_.OrderId
--                  ^^
--      This criterion should use order1_.UniqueId and not order1_.Id as
--      the column on the joined table.  That's because the property-ref attribute
--      of the many-to-one indicates that UniqueId is the property to use, instead
--      of the primary key value.
    )
```

fredericDelaporte:
Fix #3609.

The trouble for #3609 is, when comparing an entity from a direct select to a property-ref association on the entity, the primary key of the entity ends up compared to the property-ref of the association.

This fix forces a join when comparing entities through a property-ref association, replacing the property-ref in the comparison by the primary key.

Ideally, it would be better to instead replace the primary key in the comparison by the referenced property of the compared entity without adding a join. But I cannot find how to do that. And in case of comparisons between two association, one with a property-ref, the other with the primary key, the join is mandatory.

The fix had to take into account [NH-892](https://nhibernate.jira.com/browse/NH-892).
In the case of NH-892, a property-ref association is compared to an entity parameter. In this case, the parameter ends up fetched with the referenced property, and so, should be directly compared to the property-ref of the association. So, this gets broken by this fix. The fix had to further change the `DotNode` type to no more be the property-ref association type but the entity-type, in order for the parameter value to be fetched with the primary key value instead of the property-ref value.

(Actually NH-892 was already half-broken: changing the test case association to use `not-found="ignore"` triggers the same failure, because this forces a join replacing the property-ref in the comparison, too. That is likely a regression from #2081.)

Overall, this fix adds a join which in many cases could theoretically be optimized out for comparing directly references property values, but that is tedious to do and bug prone. So, I settle for a simpler, less optimal solution, but more robust. For me, the first performance metric is correctness. Speed comes second.

[NH-892]: https://nhibernate.jira.com/browse/NH-892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
```diff
@@ -426,13 +429,14 @@ private void DereferenceEntity(EntityType entityType, bool implicitJoin, string
...
 				joinIsNeeded = generateJoin || (Walker.IsInSelect && !Walker.IsInCase) || (Walker.IsInFrom && !Walker.IsComparativeExpressionClause)
-				               || comparisonWithNullableEntity;
+				               || comparisonWithNullableEntityOrThroughPropertyRef;
 			}
 
 			if ( joinIsNeeded )
 			{
-				DereferenceEntityJoin(classAlias, entityType, implicitJoin, parent, comparisonWithNullableEntity);
-				if (comparisonWithNullableEntity)
+				var forceLeftJoin = comparisonWithNullableEntityOrThroughPropertyRef && !IsCorrelatedSubselect;
```
fredericDelaporte, line 437: CorrelatedSubselect forces a theta-style join, which is not compatible with left joins. That was the cause for #3306 fixed by #3312 by ceasing to generate joins for nullable entities when in a subselect. It looks like a crutch actually. And it remains one.

src/NHibernate/Hql/Ast/ANTLR/Tree/DotNode.cs
```diff
@@ -426,15 +430,18 @@ private void DereferenceEntity(EntityType entityType, bool implicitJoin, string
...
 			{
-				DereferenceEntityJoin(classAlias, entityType, implicitJoin, parent, comparisonWithNullableEntity);
-				if (comparisonWithNullableEntity)
+				// Subselect queries use theta style joins, which cannot be forced to left outer joins.
+				var forceLeftJoin = comparisonWithNullableEntity && !IsCorrelatedSubselect;
+				DereferenceEntityJoin(classAlias, entityType, implicitJoin, parent, forceLeftJoin);
+				if (comparisonWithNullableEntity || comparisonThroughPropertyRef)
 				{
 					_columns = FromElement.GetIdentityColumns();
+					DataType = FromElement.EntityPersister.EntityMetamodel.EntityType;
```
fredericDelaporte, line 444: This fixes NH-892 regression.

fredericDelaporte:
This is currently based on master.

As a v5.3, it should target v5.3.x, but that is so old now, I do not think it is worth it, since that would mean a new version with the recent CVE not fixed in it.

I will see to target v5.4 instead

hazzik (approved):

Issue #3622, opened 2024-11-07
ISession.Refresh does not update lazy properties


ValResnick:
Hi, 

we have noticed a problem in versions 4.4.9 and 5.5.2. In version 5.2.1 we did not have this problem, I suspect it is related to the lazy property changes in version 5.3 (https://github.com/nhibernate/nhibernate-core/commit/116398d6d16b1f1415285936aab07c98508d2149).

The following scenario: 
We have opened a session with objects and also open another session and edit an object there, which is already loaded in the first session. We call Refresh so that this object is up-to-date in the previous session. Unfortunately, in the versions above, the lazy properties are no longer updated, but contain the previous and therefore old value.

Reproduction:

```
[Test] // In Class FetchLazyPropertiesFixture
public void Bug()
{
	using (var outerSession = OpenSession())
	{
		const string query = "from Person fetch Image where Id = 1";
		const string namePostFix = "_MODIFIED";
		const int imageLength = 4711;
		
		Person outerPerson = outerSession.CreateQuery(query).UniqueResult<Person>();
		
		Assert.That(outerPerson.Name.EndsWith(namePostFix), Is.False); // Normal property 
		Assert.That(outerPerson.Image.Length, Is.EqualTo(1)); // Lazy Property
		
		// Changing the properties of the person in a different sessions
		using (var innerSession = OpenSession())
		{
			var transaction = innerSession.BeginTransaction();
		
			Person innerPerson = innerSession.CreateQuery(query).UniqueResult<Person>();
			innerPerson.Image = new byte[imageLength];
			innerPerson.Name += namePostFix;
			innerSession.Update(innerPerson);
			
			transaction.Commit();
		}
		
		// Refreshing the person in the outer session
		outerSession.Refresh(outerPerson);
		
		Assert.That(outerPerson.Name.EndsWith(namePostFix), Is.True); // Value has changed
		Assert.That(outerPerson.Image.Length, Is.EqualTo(imageLength)); // This is still the old value
	}                                                                                                 
}
```

Best regards, 
Thomas

ValResnick:
Hello again,

I found out that if I always inject a new interceptor in the method PocoEntityTuplizer.AfterInitialize, my test runs through.
This will certainly not be the solution, but I hope it can help?

```
public override void AfterInitialize(object entity, ISessionImplementor session)
		{
			if (IsInstrumented)
			{
				// var interceptor = _enhancementMetadata.ExtractInterceptor(entity);
[...]

PR #3697, opened 2025-09-08
Fix ISession.Refresh not updating initialized lazy properties

ValResnick:
Fixes https://github.com/nhibernate/nhibernate-core/issues/3622


src/NHibernate/Intercept/AbstractFieldInterceptor.cs
```diff
@@ -21,7 +22,8 @@ public abstract class AbstractFieldInterceptor : IFieldInterceptor
 		private readonly HashSet<string> loadedUnwrapProxyFieldNames = new HashSet<string>();
 		private readonly string entityName;
 		private readonly System.Type mappedClass;
-
+		private readonly string[] originalUninitializedFields;
```
hazzik, line 25: It is the same as `uninitializedFieldsReadOnly`

src/NHibernate/Intercept/IFieldInterceptor.cs
```diff
@@ -41,6 +41,8 @@ public interface IFieldInterceptor
 
 		/// <summary> Get the MappedClass (field container).</summary>
 		System.Type MappedClass { get; }
+		
+		void ClearInitializedLazyFields();
```
hazzik, line 45: This would be a binary breaking change that I would like to be avoided

src/NHibernate/Intercept/AbstractFieldInterceptor.cs
```diff
@@ -33,7 +35,8 @@ protected internal AbstractFieldInterceptor(ISessionImplementor session, ISet<st
 			this.unwrapProxyFieldNames = unwrapProxyFieldNames ?? new HashSet<string>();
 			this.entityName = entityName;
 			this.mappedClass = mappedClass;
-			this.uninitializedFieldsReadOnly = uninitializedFields != null ? new ReadOnlySet<string>(uninitializedFields) : null;
+			this.uninitializedFieldsReadOnly = uninitializedFields != null ? new ReadOnlySet<string>(new HashSet<string>(uninitializedFields)) : null;
```
hazzik, line 38: There was a bug that ReadOnlySet is not really read-only

It was possible to update content of read-only set by updating underlying collection.

Some tests are depend on that behavior.

src/NHibernate/Intercept/IFieldInterceptor.cs
ValResnick: What do you mean by that?

hazzik: Please see [here](https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes#binary-breaking-change) and [here](https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules#modifications-to-the-public-contract).

src/NHibernate/Intercept/AbstractFieldInterceptor.cs
ValResnick: This is definitely an improvement, but I'd like to know more before I change it. So I added an extra variable.

PR #3707, opened 2025-09-25
Register `localtimestamp` as `current_timestamp` in Oracle9iDialect

hazzik:
In Oracle `current_timestamp` and `systimestamp` return `TIMESTAMP WITH TIME ZONE` type, which in Oracle ManagedDataAccess v23 turns into `DateTimeOffset` of expected `DateTime`, so [some tests fail](https://github.com/nhibernate/nhibernate-core/actions/runs/17997173671/job/51198821791).

Switching it to `localtimestamp` to fix the problem.

### DIFF
--- a/src/NHibernate/Dialect/Oracle9iDialect.cs
+++ b/src/NHibernate/Dialect/Oracle9iDialect.cs
@@ -10,14 +10,7 @@ namespace NHibernate.Dialect
 		public override string CurrentTimestampSelectString =>
 			$"select {CurrentTimestampSQLFunctionName} from dual";
 
-		public override string CurrentTimestampSQLFunctionName
-		{
-			get
-			{
-				// the standard SQL function name is current_timestamp...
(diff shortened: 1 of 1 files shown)


hazzik:
Not sure on this front. I’ll try to add some verifications/test. 

If that change is not acceptable, we would need to add workaround to `AbstractDateTimeType` to be able to read from `DateTimeOffset`.

hazzik:
To the above: my server is in UTC, and session UTC+10

hazzik:
So, I would say that this change is pretty safe and should not be breaking. However, we can say a few words in a probably breaking changes section if we want. 

In any way, I would like this to be included in 5.6, especially if we want to treat this as a breaking change

fredericDelaporte (approved):
I thought the union might cause some additional type conversion before inserting and checked with separated insert queries, same results.

PR #3714, opened 2025-10-29
Fixes query cache result transformer aliases issue

### LINKED ISSUE #3713: Query cache breaks result transformers which need aliases
# Summary

If you use query caching on a query with a result transformers which use the aliases array (e.g. `AliasToBeanResultTransformer`), when the data comes from cache, `TransformTuple` receives an empty `aliases` array, thus preventing the transformer from functioning correctly.

# How to reproduce

This is a test you can add to `src/NHibernate.Test/CacheTest/QueryCacheFixture.cs` in order to reproduce the problem:

```cs
		[Test]
		public void QueryCacheWithAliasToBeanTransformer()
		{
			using (var s = OpenSession())
			{
				const string query = "select s.id_ as Id from Simple as s";
				var result1 = s
					.CreateSQLQuery(query)
					.SetCacheable(true)
					.SetResultTransformer(Transformers.AliasToBean<SimpleDTO>())
					.UniqueResult();

				Assert.That(result1, Is.InstanceOf<SimpleDTO>());
				var dto1 = (SimpleDTO)result1;
				Assert.That(dto1.Id, Is.EqualTo(1));

				// Run a second time, just to test the query cache
				var result2 = s
					.CreateSQLQuery(query)
					.SetCacheable(true)
					.SetResultTransformer(Transformers.AliasToBean<SimpleDTO>())
					.UniqueResult();

				Assert.That(result2, Is.InstanceOf<SimpleDTO>());
				var dto2 = (SimpleDTO)result2;
				Assert.That(dto2.Id, Is.EqualTo(1)); // <-- the Id is 0 because the transformer failed
			}
		}

		private class SimpleDTO
		{
			public long Id { get; set; }
			public string Address { get; set; }
		}
```
# Investigation

What seems to happen is that:

A `CustomLoader` instance is used. Its `GetResultList` method relies on `ReturnAliasesForTransformer`/`transformerAliases` to get the alias array. `transformerAliases` could be filled either in the constructor (where in this case it is not filled because `customQuery.CustomQueryReturns`) and also in `AutoDiscoverTypes`.

In `Loader.ListUsingQueryCache`, when the data is not found the cache, `DoList` is called, which ends up calling  `AutoDiscoverTypes` on the result transformer somewhere inside. However, when the data is found in the cache, nothing calls `AutoDiscoverTypes` on the result transformer and it remains the empty array created in the constructor.

As `AutoDiscoverTypes` uses the metadata returned by the database, and when using the cache, we are not doing a database query, the answer is probably to add the aliases to the cached data and read them out from there. And indeed they are stored in the cache and also read out by `StandardQueryCache` via `GetResultsMetadata` and then put into the `QueryKey`'s `CacheableResultTransformer` by calling its `SupplyAutoDiscoveredParameters` method.

However those aliases never makes it back into the `CustomLoader`'s `transformerAliases`, which is what the user-code-supplied transformer (in this case, the `AliasToBeanResultTransformer`) eventually gets.

# Additional considerations

I encountered it with `CreateSQLQuery`/`ISQLQuery`, and I'm not sure if it happens with other types of queries.

delta-emil:
I don't know if this is the most elegant way to solve the issue, but it does solve it, at least for the case I found.
I'm not sure if other Loader sub-classes also need something similar.

fix #3713

### DIFF
--- a/src/NHibernate.Test/Async/CacheTest/QueryCacheFixture.cs
+++ b/src/NHibernate.Test/Async/CacheTest/QueryCacheFixture.cs
@@ -108,7 +108,7 @@ namespace NHibernate.Test.CacheTest
 		{
 			using (var s = OpenSession())
 			{
-				const string query = "select s.id_ as Id from Simple as s";
+				const string query = "select s.id_ as Id from Simple as s;";
 				var result1 = await (s
 					.CreateSQLQuery(query)
 					.SetCacheable(true)
--- a/src/NHibernate.Test/Async/CacheTest/QueryCacheFixture.cs
+++ b/src/NHibernate.Test/Async/CacheTest/QueryCacheFixture.cs
@@ -11,6 +11,7 @@
 using System.Collections;
 using NHibernate.Cfg;
 using NHibernate.DomainModel;
+using NHibernate.Transform;
 using NUnit.Framework;
 using Environment = NHibernate.Cfg.Environment;
 
@@ -100,6 +101,41 @@ namespace NHibernate.Test.CacheTest
 
 				Assert.That(result, Is.EqualTo(200012), "Unexpected cached result");
 			}
+		}
+
+		[Test]
+		public async Task QueryCacheWithAliasToBeanTransformerAsync()
+		{
+			using (var s = OpenSession())
+			{
+				const string query = "select s.id_ as Id from Simple as s";
+				var result1 = await (s
+					.CreateSQLQuery(query)
+					.SetCacheable(true)
+					.SetResultTransformer(Transformers.AliasToBean<SimpleDTO>())
+					.UniqueResultAsync());
+
+				Assert.That(result1, Is.InstanceOf<SimpleDTO>());
+				var dto1 = (SimpleDTO)result1;
+				Assert.That(dto1.Id, Is.EqualTo(1));
+
+				// Run a second time, just to test the query cache
+				var result2 = await (s
+					.CreateSQLQuery(query)
+					.SetCacheable(true)
+					.SetResultTransformer(Transformers.AliasToBean<SimpleDTO>())
+					.UniqueResultAsync());
+
+				Assert.That(result2, Is.InstanceOf<SimpleDTO>());
+				var dto2 = (SimpleDTO)result2;
+				Assert.That(dto2.Id, Is.EqualTo(1));
+			}
+		}
+
+		private class SimpleDTO
+		{
+			public long Id { get; set; }
+			public string Address { get; set; }
 		}
 	}
 }
--- a/src/NHibernate.Test/Async/CacheTest/QueryCacheFixture.cs
+++ b/src/NHibernate.Test/Async/CacheTest/QueryCacheFixture.cs
@@ -107,7 +107,10 @@ namespace NHibernate.Test.CacheTest
 		[Test]
 		public async Task QueryCacheWithAliasToBeanTransformerAsync()
 		{
-			const string query = "select s.id_ as Id from Simple as s;";
+			if (!TestDialect.HasActualIntegerTypes)
+				Assert.Ignore("Database does not support properly integers.");
+
+			const string query = "select s.id_ as Id from Simple s;";
 			Sfi.Statistics.Clear();
 
 			using (var s = OpenSession())
--- a/src/NHibernate.Test/Async/CacheTest/QueryCacheFixture.cs
+++ b/src/NHibernate.Test/Async/CacheTest/QueryCacheFixture.cs
@@ -43,35 +43,48 @@ namespace NHibernate.Test.CacheTest
 		{
 			using var s = OpenSession();
 			using var t = s.BeginTransaction();
-				s.Delete("from Simple");
+			s.Delete("from Simple");
 			t.Commit();
 		}
 
 		[Test]
 		public async Task QueryCacheWithNullParametersAsync()
 		{
+			const string query = "from Simple s where s = :s or s.Name = :name or s.Address = :address";
+			Sfi.Statistics.Clear();
+			object simple;
 			using (var s = OpenSession())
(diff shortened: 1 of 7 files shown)


fredericDelaporte:
Thanks for the contribution,

This issues would mean #3169 was not fixed for all cases.

Looking at its fix, #3184, it might be due to some paths not using an adequate overload, like here:
https://github.com/nhibernate/nhibernate-core/blob/d8cde71481d006c5473cdfa648dd44c996c48115/src/NHibernate/Cache/IQueryCache.cs#L172-L179

It looks worth checking, but I have done it yet.

src/NHibernate.Test/Async/CacheTest/QueryCacheFixture.cs
```diff
@@ -101,5 +102,40 @@ public async Task QueryCacheWithScalarReturnAsync()
 				Assert.That(result, Is.EqualTo(200012), "Unexpected cached result");
 			}
 		}
+
+		[Test]
+		public async Task QueryCacheWithAliasToBeanTransformerAsync()
+		{
+			using (var s = OpenSession())
+			{
+				const string query = "select s.id_ as Id from Simple as s";
```
fredericDelaporte:
```suggestion
				const string query = "select s.id_ as Id from Simple as s;";
```

fredericDelaporte (commented):
This PR targets the 5.5.x branch. Fixes for bugs which are not regressions introduced in a supported version (5.5.x and 5.6.x as of now) should usually target the master branch.

It looks like this bug dates back to #2943, a 5.3.x change.

This said, the change is very simple and looks safe, so why not still putting this in 5.5.4, if other maintainers agree.

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
```diff
@@ -90,5 +91,48 @@ public void QueryCacheWithScalarReturn()
 				Assert.That(result, Is.EqualTo(200012), "Unexpected cached result");
 			}
 		}
+
+		[Test]
+		public void QueryCacheWithAliasToBeanTransformer()
+		{
+			const string query = "select s.id_ as Id from Simple as s;";
```
fredericDelaporte: This causes the test to fail for Oracle, which does not support statements ending with a semi-colon. Can you rewrite the test with a HQL query instead? It would let NHibernate handles this.

fredericDelaporte: Using a HQL query does not test the case anymore because it does not use auto-discovered aliases.

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
```diff
@@ -14,51 +15,65 @@ public class QueryCacheFixture : TestCase
...
+
 		protected override void OnSetUp()
 		{
-			using (var s = OpenSession())
-			{
-				s.Save(new Simple(), 1L);
-				s.Flush();
-			}
+			using var s = OpenSession();
+			using var t = s.BeginTransaction();
```
fredericDelaporte, line 26: Not using transaction on the setup cause the cache space for `Simple` to remain in "pre-invalidated" state. Queries detected as using this space then cannot hit the cache.

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
```diff
@@ -14,51 +15,65 @@ public class QueryCacheFixture : TestCase
...
+					.SetEntity("s", simple)
 					.SetString("name", null)
 					.SetString("address", null)
 					.SetCacheable(true)
 					.UniqueResult();
 
 				Assert.That(result, Is.Not.Null);
-				Assert.That(s.GetIdentifier(result), Is.EqualTo(1));
+				Assert.That(s.GetIdentifier(result), Is.EqualTo(SimpleId));
+				Assert.That(Sfi.Statistics.QueryCacheHitCount, Is.EqualTo(1));
```
fredericDelaporte, line 75: Make sure the cache is actually used and tested. The previous lack of transactions in the test would have caused this to fail with the use of two sessions.

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
```diff
@@ -73,22 +88,82 @@ public void QueryCacheWithScalarReturn()
...
 				Assert.That(result, Is.EqualTo(200012), "Unexpected cached result");
 			}
 		}
+
+		[Test]
+		public void QueryCacheWithAliasToBeanTransformer()
+		{
+			if (!TestDialect.HasActualIntegerTypes)
+				Assert.Ignore("Database does not support properly integers.");
```
fredericDelaporte, line 124: The SQL query returns a decimal with Oracle, which causes a typing failure. The `QueryCacheWithScalarReturn` has actually the same trouble, dodged by casting to decimal in the query for every database. Then it disables the test for SQLite due to it not having a decimal type...

Here I do the opposite, I disable the test for Oracle.

src/NHibernate.Test/CacheTest/QueryCacheFixture.cs
```diff
@@ -73,22 +88,82 @@ public void QueryCacheWithScalarReturn()
...
+
+			const string query = "select s.id_ as Id from Simple s;";
```
fredericDelaporte, line 126: Some databases (Oracle again) does not support the `as` keyword for table alias. I have not found a reference for it, but it seems not using it is more widely supported than using it. So, I removed it. For columns alias, that is the opposite, better use it.

About the semicolon, it is not supported by the Oracle engine too. (It is only supported by its tooling.) But as the test will not support Oracle, I have not left something in place for handling this.

fredericDelaporte (approved):
I have rebased this PR due to some fixes from upstream being needed for running the CI.

PR #3715, opened 2025-10-29
Enable use of Testcontainers

gliljas:
Just something that you can opt in to. By specifying a connectionstring like "testcontainers=oracle", you can run tests on an Oracle database without the hassle of setting one up.

There's certainly room for more extensibility and possibly a "keep-alive" feature, but this works reasonably well.

### DIFF
--- a/src/NHibernate.Test.VisualBasic/Issues/GH3716/Entity.vb
+++ b/src/NHibernate.Test.VisualBasic/Issues/GH3716/Entity.vb
@@ -1,0 +1,7 @@
+Namespace Issues.GH3716
+    Public Class Entity
+        Public Overridable Property Id As Guid
+
+        Public Overridable Property Date1 As Date?
+    End Class
+End Namespace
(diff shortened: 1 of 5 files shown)


src/NHibernate.Test/TestContainerSetup.cs
```diff
@@ -0,0 +1,76 @@
...
+	using Testcontainers.MariaDb;
+	using Testcontainers.MsSql;
+	using Testcontainers.MySql;
+	using Testcontainers.Oracle;
+	using Testcontainers.PostgreSql;
+
+	[SetUpFixture]
+	public class TestContainerSetup
+	{
+		private static IDatabaseContainer _container;
```
fredericDelaporte: When using double check locking, better set the tested variable as "volatile". Depending on the target architecture, it may be needed for preventing threading issues.

```suggestion
		private static volatile IDatabaseContainer _container;
```

fredericDelaporte (commented):
Failures seems unrelated.

Oracle ones are its usual recurring timeout.

AppVeyor looks like some temporary outage.

I have relaunched these builds.

fredericDelaporte (approved):

PR #3721, opened 2025-11-12
Linq: replace MemoryExtensions.Contains with Enumerable.Contains

### LINKED ISSUE #3651: Add support for "first-class span types" C# language feature
Since the latest Visual Studio version (17.13)/C# version some of our code started to fail at runtime.
The problem is that Microsoft implemeted the "First-class Span types" language feature (https://github.com/dotnet/csharplang/blob/main/proposals/first-class-span-types.md)
I'm using the LangVersion=preview settings (because of the "field" keyword)

For example we have the following code:
```
        var ids = new Guid[] { guid1, guid2... };
        foreach (var r in session.Query<MyClass>().Where(x => ids.Contains(x.Id)))
        {
            ...
        }
 
```

Earlier (before 11.02.2025) this code worked, but now it throws the following exception:
```
NHibernate.HibernateException
  HResult=0x80131500
  Message=Evaluation failure on op_Implicit(value(System.Guid[]))
  Source=NHibernate
  StackTrace:
   at NHibernate.Linq.Visitors.NhPartialEvaluatingExpressionVisitor.Visit(Expression expression)
....
Inner Exception 1:
NotSupportedException: Specified method is not supported.
 
```

Because earlier the predicate expression in the Where method was calling the `Enumerable.Contains` extension method, but now it calls the `MemoryExtensions.Contains` method.

There is a workaround:
```
        var ids = new Guid[] { guid1, guid2... };
        foreach (var r in session.Query<MyClass>().Where(x => ((IEnumerable<Guid>)ids).Contains(x.Id)))
        {
            ...
        }

```

but it needs to change our code everywhere. And hard to find every places since there is no build error, only a runtime exception.

hazzik:
Fixes #3651


ValResnick (approved):
The draft looks good. Let me know if I can help with anything.
It looks like the .NET 10 topic has been moved to a lower priority.
However, this issue is currently blocking our upgrade to .NET 10, so it’s a critical blocker for us at the moment.

ValResnick (approved):
Looks fine to me

fredericDelaporte:
Is there a reason for this PR to stay as a draft?

hazzik:
@fredericDelaporte no reason

fredericDelaporte (approved):

PR #3722, opened 2025-11-12
Use global.json dotnet version for actions

hazzik:
Otherwise Generate Async code action fails here: https://github.com/nhibernate/nhibernate-core/pull/3662


andreas-eriksson (approved):

fredericDelaporte:
Shouln't this be backported to supported older branches?

hazzik:
@fredericDelaporte not really, the issue only for the cases when the .net version in the target branch is different to the version in a PR branch. I don’t think we plan .net version upgrades in the patch versions.

PR #3726, opened 2025-11-17
Fix #3725: DbBatchBatcher empty batch execution bug

### LINKED ISSUE #3725: DbBatchBatcher may fail on empty batch execution
## Description

  `DbBatchBatcher` (available in NHibernate 5.6.0) throws `InvalidOperationException` when attempting to execute an empty
  batch, while other batchers (`GenericBatchingBatcher` and `SqlClientBatchingBatcher`) handle this scenario correctly.

  ## Error

  System.InvalidOperationException: ExecuteNonQuery: CommandText property has not been initialized
     at Microsoft.Data.SqlClient.SqlCommand.ValidateCommand(Boolean isAsync, String method)
     at Microsoft.Data.SqlClient.SqlCommand.InternalExecuteNonQuery(TaskCompletionSource`1 completion, Boolean
  sendToPipe, Int32 timeout, Boolean& usedCache, Boolean asyncWrite, Boolean inRetry, String methodName)
     at Microsoft.Data.SqlClient.SqlCommand.ExecuteNonQuery()
     at Microsoft.Data.SqlClient.SqlBatch.ExecuteNonQuery()
     at NHibernate.AdoNet.DbBatchBatcher.DoExecuteBatch(DbCommand ps)

  ## Root Cause

  `DbBatchBatcher.DoExecuteBatch()` and `DoExecuteBatchAsync()` lack empty batch validation that exists in other
  batchers. When `ExecuteBatch()` is called with an empty batch (`_currentBatch.BatchCommands.Count == 0`), the code
   attempts to execute a `DbBatch` with no commands, causing the `CommandText` to be uninitialized.

  **File:** `src/NHibernate/AdoNet/DbBatchBatcher.cs`
  **Methods:** `DoExecuteBatch()` (line 116) and `DoExecuteBatchAsync()` (line 145)

  ## Comparison with Other Batchers

  **GenericBatchingBatcher** (working correctly):
  ```csharp
  protected override void DoExecuteBatch(DbCommand ps)
  {
      if (_currentBatch.CountOfCommands == 0)
      {
          Expectations.VerifyOutcomeBatched(_totalExpectedRowsAffected, 0, ps);
          return;
      }
      // ... proceed with execution
  }
```
  DbBatchBatcher (missing check):
  ```csharp
  protected override void DoExecuteBatch(DbCommand ps)
  {
      try
      {
          // Missing empty batch check here
          rowsAffected = _currentBatch.ExecuteNonQuery(); // Throws on empty batch
```
  Proposed Fix

  Add empty batch check at the beginning of both methods:
  ```csharp
  protected override void DoExecuteBatch(DbCommand ps)
  {
      if (_currentBatch.BatchCommands.Count == 0)
      {
          Expectations.VerifyOutcomeBatched(_totalExpectedRowsAffected, 0, ps);
          return;
      }
      // ... existing code
  }
```
Environment
  - NHibernate version: 5.6.0
  - .NET version: 8
  - Database: SQL Server

 Impact

  This causes runtime exceptions in scenarios where ExecuteBatch() is triggered without any commands being added to
  the batch, which can occur during normal session lifecycle operations.

### PR DESCRIPTION
**Pull request by ja-kar**
Fixes #3725

  ## Description

  `DbBatchBatcher` throws `InvalidOperationException` when attempting to execute an empty batch, while
  `GenericBatchingBatcher` and `SqlClientBatchingBatcher` handle this scenario correctly.

  ## Root Cause

  `DbBatchBatcher.DoExecuteBatch()` and `DoExecuteBatchAsync()` lack empty batch validation. When `ExecuteBatch()`
  is called with an empty batch (`_currentBatch.BatchCommands.Count == 0`), the code attempts to execute a `DbBatch`
   with no commands, causing:

  System.InvalidOperationException: ExecuteNonQuery: CommandText property has not been initialized

  ## Changes

  Added empty batch check at the beginning of both methods, matching the pattern used in `GenericBatchingBatcher`:

  ```csharp
  if (_currentBatch.BatchCommands.Count == 0)
  {
      Expectations.VerifyOutcomeBatched(_totalExpectedRowsAffected, 0, ps);
      return;
  }
```
  Modified:
  - src/NHibernate/AdoNet/DbBatchBatcher.cs - Added guard clauses to DoExecuteBatch() and DoExecuteBatchAsync()

  Added:
  - Test cases in src/NHibernate.Test/Ado/BatcherFixture.cs verifying the fix
  - Generated async test counterparts

### DIFF
--- a/src/NHibernate.Test/Ado/BatcherFixture.cs
+++ b/src/NHibernate.Test/Ado/BatcherFixture.cs
@@ -306,44 +306,54 @@ namespace NHibernate.Test.Ado
 		}
 
 		[Test]
-		[Description("The batcher should handle empty batch execution without throwing exceptions.")]
-		public void EmptyBatchShouldNotThrowException()
+		[Description("Inserting exactly BatchSize entities should not throw on commit. See GH-3725.")]
+		public void InsertExactlyBatchSizeEntitiesShouldNotThrowOnCommit()
 		{
-			// This test verifies that batchers handle empty batches correctly
-			// DbBatchBatcher had a bug where ExecuteBatch was called on an empty batch,
-			// causing InvalidOperationException: CommandText property has not been initialized
-			// See GH-3725
+			// This test verifies that DbBatchBatcher handles empty batches correctly.
+			// The bug (GH-3725): When inserting exactly BatchSize entities, the batch auto-executes
+			// when full (via ExecuteBatchWithTiming), which clears _currentBatch but NOT _batchCommand.
+			// On commit, ExecuteBatch() is called, sees _batchCommand is set, and calls DoExecuteBatch
+			// on an empty _currentBatch, causing InvalidOperationException.
 
-			using var session = OpenSession();
-			using var transaction = session.BeginTransaction();
+			// BatchSize is configured as 10 in this fixture
+			const int batchSize = 10;
 
-			// Execute queries that don't add to the batch
-			_ = session.Query<VerySimple>().FirstOrDefault();
+			using (var session = OpenSession())
+			using (var transaction = session.BeginTransaction())
+			{
+				// Insert exactly BatchSize entities - this fills the batch and triggers auto-execution
+				for (int i = 0; i < batchSize; i++)
+				{
+					session.Save(new VerySimple { Id = 1000 + i, Name = $"Test{i}", Weight = i * 1.1 });
+				}
 
-			// Prepare a new command which triggers ExecuteBatch on any existing batch
-			// If the previous command didn't add anything to the batch, this would fail
-			// before the fix with InvalidOperationException
-			_ = session.Query<VerySimple>().FirstOrDefault();
+				// Commit triggers ExecuteBatch() which would fail on empty batch without the fix
+				transaction.Commit();
+			}
 
-			// Test passes if no exception is thrown
-			transaction.Commit();
+			Cleanup();
 		}
 
 		[Test]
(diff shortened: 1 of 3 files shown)


gliljas:
Good catch and thanks for the fix!

My very personal reflection is that your tests should be included in the normal batcher tests, rather than as a NHSpecificTest.

ja-kar:
Good point! I'll move the tests

fredericDelaporte:
I am trying the test without the fix, and I do not get a failure. The test must demonstrate the trouble.

Test run with PostgreSQL, after having reset mixed the commit, and undone the change on NHibernate to only keep the change on NHibernate.Test.
image

ja-kar:
I have updated the test to be more like the scenario I encountered the bug in and verified that the tests now fail without the fix on SQL Server

fredericDelaporte:
I have tried with your test update and SQL Server: under net8, the test is ignored. Under net4.8, there is still no failure without your fix. With PostgreSQL, still no failures either.

So, I am now launching the whole CI on your test to check if anything starts failing, see #3733.

fredericDelaporte (approved):

ja-kar:
Did you use NHibernate.Driver.MicrosoftDataSqlClientDriver as the driver on .NET8? Wasn't ignored for me. On net4.8 it wont fail though as this bug only exists in the DbBatchBatcher afaik. Maybe it should be locked down to only run for .NET8 then.

Issue #3738, opened 2026-03-13
Incompatible unicode string literals with DB2400Dialect


staikfrite:
The DB2400Dialect inherits from DB2Dialect and at some point, the DB2Dialect.ToStringLiteral method was updated to handle unicode, especially the part 

```C#
if (isUnicode)
	literal.Insert(0, "U&");
```

However, the IBM .Net drivers for the "DB2 for i" database can't cope with the "U&" prefix, which leads on some occasions (amongst them the usage of DiscriminatorValue in the mapping) to DB-failures.

I suggest to re-override the ToStringLiteral in the DB2400Dialect class, this time without the "isUnicode" part

See [Pull request](https://github.com/nhibernate/nhibernate-core/pull/3739)

fredericDelaporte:
Fixed by #3740.

PR #3739, opened 2026-03-13
See issue #3738

staikfrite:
https://github.com/nhibernate/nhibernate-core/issues/3738

### DIFF
--- a/src/NHibernate/Dialect/DB2400Dialect.cs
+++ b/src/NHibernate/Dialect/DB2400Dialect.cs
@@ -1,7 +1,10 @@
 using System;
+using System.Data;
 using System.Data.Common;
+using System.Text;
 using NHibernate.Dialect.Schema;
 using NHibernate.SqlCommand;
+using NHibernate.SqlTypes;
 
 namespace NHibernate.Dialect
 {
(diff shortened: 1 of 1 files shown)


src/NHibernate/Dialect/DB2400Dialect.cs
```diff
@@ -68,5 +71,23 @@ public override bool SupportsVariableLimit
...
 		}
+
+		public override string ToStringLiteral(string value, SqlType type)
+		{
+			if (value == null)
+				throw new System.ArgumentNullException(nameof(value));
+			if (type == null)
+				throw new System.ArgumentNullException(nameof(value));
+
+			// See https://www.ibm.com/docs/en/db2/11.5?topic=elements-constants#r0000731__title__7
```
fredericDelaporte, line 82: For DB2 on iSeries | OS/400 | for i | or whatever they rebrand it to, the relevant documentation is https://www.ibm.com/docs/en/i/7.6.0?topic=constants-character-string

It does not support the `U&` literal string encoding, indeed. But these strings does not seem to support Unicode either, which seems to be called ["graphic string"](https://www.ibm.com/docs/en/i/7.6.0?topic=constants-graphic-string#rbafzgraphicconst__title__3). And there, it seems we should use the `N` prefix for Unicode strings.

So, the Unicode logic should not be removed but changed to use the `N` prefix instead.

May you check with a custom dialect if that works? I do not know people having access to this DB2 for i database for testing this.

fredericDelaporte:
DB2 for i supports Unicode in a fashion similar to many other DB, and should use the Dialect base implementation, actually. So, I have proposed another fix, see #3740.

src/NHibernate/Dialect/DB2400Dialect.cs
staikfrite: Hi Frederic, 

thank you for you suggestion: it is much cleaner to call the Dialect base method that uses the N prefix anyway.

PR #3740, opened 2026-03-22
Fix string literals for DB2 on i

### LINKED ISSUE #3738: Incompatible unicode string literals with DB2400Dialect

fredericDelaporte:
fix #3738

### DIFF
--- a/src/NHibernate/Dialect/DB2Dialect.cs
+++ b/src/NHibernate/Dialect/DB2Dialect.cs
@@ -27,7 +27,7 @@ namespace NHibernate.Dialect
 	/// </list>
 	/// </para>
 	/// <para>
-	/// Two BD2 engines exist. One for Linux, UNIX and Windows, and another one for the IBM i system
+	/// Two DB2 engines exist. One for Linux, UNIX and Windows, and another one for the IBM i system
 	/// (formerly OS/400). For this other engine, see <see cref="DB2400Dialect" />.
 	/// </para>
 	/// </remarks>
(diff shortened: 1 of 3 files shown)


fredericDelaporte:
May replace #3739.

src/NHibernate/Dialect/DB2Dialect.cs
```diff
@@ -297,9 +303,18 @@ public override string ForUpdateString
...
+		/// <summary>
+		/// Indicates if the base <see cref="Dialect.ToStringLiteral(string, SqlType)" /> should be used instead
+		/// of the DB2 specific <see cref="ToStringLiteral(string, SqlType)" />.
+		/// </summary>
+		protected virtual bool UseDialectBaseToStringLiteral => false;
+
 		/// <inheritdoc />
 		public override string ToStringLiteral(string value, SqlType type)
 		{
+			if (UseDialectBaseToStringLiteral)
```
bahusoid: Do we really need separate flag? It seems this check can be replaced with `UseNPrefixForUnicodeStrings` as current implementation doesn't support it anyway.

fredericDelaporte:
By lack of contributor for validating the PR, I intend to merge this next week (from 2026-05-03).

Issue #3608, opened 2024-09-19
Memory leak in NHibernate 5.5.0 compared to 5.4.9


Bogatinov:
Hi friends, I'm writing this issue to report a Memory leak on NHibernate objects when upgrading from v5.4.9 to v5.5.0 (and v5.5.2)

## Configuration
- dotNET Framework 4.8.1
- NHibernate v5.4.9 and v.5.5.2
- FluentNHibernate v3.4.0 with a binding redirect for NHibernate 5.5.0
- AMD Ryzen 9 5950X with 32GB 3600Mhz RAM/Samsung 980 1TB
- SQL Server 16.0.1000.6 and 15.0.4385.2 databases
- We are using `Session.Query<T>`, `Session.QueryOver<T>` and `Session.CreateSQLQuery("raw")`
- around 270 mapping classes for tables with `FluentNHibernate`

I haven't tested to build a custom FluentNHibernate DLL with the 5.5.2 package and see if that resolves the issue.

## Story
Comparing NHibernate v5.4.9 and v5.5.0 there seem to be +200MB after 16min of usage of our app. I tested this on a Staging environment, and after 2 weeks, our testers reported that the VM was running slow. 

Checked that the RAM was full with _Application Insights Metrics_, while the app RAM usage was around 6GB. This looked weird since +200MB is a small difference, but the more users request for SQL Server 2016/2019 data, the more this increased.

I reverted to v5.4.9 and the app stayed on around 2GB for the last 2 months.

## Repository
TBD - currently setting up...

do let me know in which direction should I test it, since I'm noticing that most types with high Delta difference are part of common usages in NHibernate APIs. .

## Benchmarks
It's a project I can not share because of NDA reasons. The benchmark is applying automatic 10 actions in 16 min usage, by 1 user.

### v5.4.9
![368199764-0f5b3033-bd66-4604-bc87-e70d9af831c0](https://github.com/user-attachments/assets/46a6911c-231f-48eb-8234-12f09296f464)

### v5.5.2
![368199625-7a0fb175-26c6-4718-b18d-bb7be551f8ca](https://github.com/user-attachments/assets/4efdf63a-65f0-48df-b99f-8223679d2a6c)

### Comparison in objects
For the same benchmark test case, v5.5.2 generated 150k more objects (~15MB more)
[...]

PR #3742, opened 2026-04-17
Refactor Command parts to avoid duplicate characters. Reduces memory footprint

tschu-schleupen:
Adresses #3608 
Related to and idea to fix it from #3640 

We noticed ", " dominated memory consumption after running our services for a while: 
image

After applying the changes made in this PR: 
image


src/NHibernate/SqlCommand/SqlString.cs
```diff
@@ -1064,7 +1065,14 @@ private struct Part : IEquatable<Part>
...
 				SqlIndex = sqlIndex;
-				Content = content;
+				Content = content.Length switch
+				{
+					1 when content == StringHelper.ClosedParen => StringHelper.ClosedParen,
+					1 when content == StringHelper.OpenParen   => StringHelper.OpenParen,
+					1 when content == StringHelper.Comma       => StringHelper.Comma,
+					2 when content == StringHelper.CommaSpace  => StringHelper.CommaSpace,
+					_ => content
+				};
```
fredericDelaporte, line 1086: A comment should explain why this is useful, over simply storing the content as is.

fredericDelaporte, line 1086: Or, this would be more explicit:

```suggestion
				Content = content.Length <= 2 ? string.IsInterned(content) ?? content : content;
```

The four StringHelper members the proposed code handles are interned, as they are literal constants. So, this code should work too. (I have not tested it.) And it would handle all cases of one to two characters strings.

About the execution cost, I think the intern lookup should be more performant than up to three string comparisons for one character length strings. But this would have to be benchmarked.

fredericDelaporte (commented):
As this is not a regression fix, it should target master.

src/NHibernate/SqlCommand/SqlString.cs
tschu-schleupen: Good point — I was uncertain myself, so I built a small BenchmarkDotNet project to measure both variants side by side. Each scenario processes 1000 strings per invocation, with inputs freshly allocated via `new string(char[])` so reference equality with the compile-time-interned literals doesn't short-circuit the measurement:

- **SwitchHits** — only the four strings the switch targets (the hot path)
- **ShortNonHits** — short strings (`=`, `?`, `*`, `!=`, `<>`, …) the switch misses but `IsInterned` might catch
- **LongStrings** — typical SQL tokens (`SELECT`, `:p0`, `customer_id`, …) that both approaches short-circuit via the length check
- **RealisticMix** — ~60% switch hits, ~10% short non-hits, ~30% longs

Results on an AMD Ryzen 9 3900X, across .NET 10, .NET 8 and .NET Framework 4.8 (mean per 1000-string invocation; ratio is `IsInterned` vs. switch, >1 means the switch is faster):

| Scenario       | Runtime    | Switch         | IsInterned     | Ratio  |
|----------------|------------|---------------:|---------------:|-------:|
| SwitchHits     | .NET 10    |    1,418 ns    |   62,241 ns    |  43.9× |
| SwitchHits     | .NET 8     |    1,268 ns    |   53,638 ns    |  42.3× |
| SwitchHits     | .NET Fx 4.8|    6,178 ns    |   84,441 ns    |  13.7× |
| ShortNonHits   | .NET 10    |    1,266 ns    |   65,076 ns    |  51.4× |
| ShortNonHits   | .NET 8     |    1,486 ns    |   55,912 ns    |  37.6× |
| ShortNonHits   | .NET Fx 4.8|    7,814 ns    |   88,466 ns    |  11.3× |
| RealisticMix   | .NET 10    |   25,068 ns    |  470,986 ns    |  20.2× |
| RealisticMix   | .NET 8     |   79,073 ns    |  408,742 ns    |   5.2× |
| RealisticMix   | .NET Fx 4.8|  135,326 ns    |  629,083 ns    |   4.7× |
| LongStrings    | .NET 10    |      679 ns    |      639 ns    |   0.94×|
| LongStrings    | .NET 8     |      624 ns    |      600 ns    |   0.96×|
| LongStrings    | .NET Fx 4.8|    1,940 ns    |      743 ns    |   0.38×|

Allocations are zero for both implementations in all scenarios.

The intern-lookup variant only wins on the `LongStrings` path, where both are in the low-ns-per-call range and the difference doesn't matter in practice (and interestingly it's only on .NET Framework that the gap is meaningful there — modern RyuJIT seems to fold the switch down to essentially the same cost). On everything else, the global intern table's hashing and synchronisation overhead dominates the length-gated comparisons against the compile-time constants, which likely reduce to reference equality thanks to literal interning.

So I'd lean toward keeping the switch as-is, and I'll add a comment explaining the intent (addressing your other review thread at the same time). If `IsInterned` would still be preferred for coverage reasons — e.g. catching other short tokens a tokenizer might produce — I can extend the switch with additional arms instead; that keeps the per-call cost flat and lets us deduplicate exactly the tokens we know are worth it.

Happy to share the benchmark project if you'd like to sanity-check the setup, and I'll retarget the PR to `master` as you requested.

fredericDelaporte: Ok, thanks, go ahead with keeping the switch.

fredericDelaporte (approved):

fredericDelaporte:
This will wait #3743, in order to be able to validate the PR through the CI.

hazzik:

tschu-schleupen:
> Can you please enable [Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) option on this PR?

I would if I could. But the fork was created in an organisation account. In this setup the feature in PRs does not exist. Is there any change you can describe, so that I can apply it?

hazzik:
Need to merge the master

tschu-schleupen:
merge is done

PR #3743, opened 2026-04-26
Fix vulnerable packages blocking the test build

fredericDelaporte:


fredericDelaporte:
I intend to merge this on next week-end.

src/NHibernate/NHibernate.csproj
```diff
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <Description>An object persistence library for relational databases.</Description>
     <TargetFrameworks>$(NhLibTargetFrameworks)</TargetFrameworks>
-    <NoWarn>$(NoWarn);3001;3002;3003;3005;1591;419</NoWarn>
+    <NoWarn>$(NoWarn);3001;3002;3003;3005;1591;419;AC0105;NETSDK1138</NoWarn>
```
fredericDelaporte, line 7: The AppVeyor build seems to no more tolerate the HQL warning AC0105 "no lexer rule corresponding to token", no idea why.
I suppress the "NETSDK1138: The target framework 'netcoreapp2.0' is out of support" too, as it might be causing troubles too, with TreatWarningsAsErrors.

fredericDelaporte:
Having done some other changes, I postpone to the next weekend merging this, if no review happens.

src/NHibernate/NHibernate.csproj
fredericDelaporte: It does not work. The warning is still seen, as "warning 105" now, and it still seems to cause the build to fail.

Locally, I do reproduce the trouble by running `.\build.ps1 -TaskList Build` with Nhibernate bin & obj folders removed.

psake.ps1
```diff
@@ -121,7 +121,7 @@ Task Build {
     Exec { 
         dotnet build ./src/NHibernate/NHibernate.csproj -m:1 -c Release
         dotnet build ./src/NHibernate.sln -c Release
-    }
+    } -MaxRetries 1
```
fredericDelaporte, line 124: That is a crutch, but locally, it allows to work around the failure. I still leave the warning suppressions in place, it gives a cleaner build output.

fredericDelaporte: With it, the AppVeyo build works again.

deAtog:
Any chance of merging this? I see that all builds for pull requests are failing due to this issue.

Issue #2603, opened 2020-11-11
BinaryFormatter is not supported in .net5.0


danielegiallonardo:
Net5.0 applications won't start since BinaryFormatter use is prohibited, as stated [here](https://docs.microsoft.com/it-it/dotnet/core/compatibility/3.1-5.0#binaryformatter-serialization-methods-are-obsolete-and-prohibited-in-aspnet-apps).
Please consider replacing BinaryFormatter with something Net5.0-friendly.

bahusoid:
> It's something AspNet Core related, since .net5.0 console applications don't have such an issue.

Well according to docs you mentioned you can still use it with ASP.NET. You just need to update your csproj file:
>To continue using BinaryFormatter in ASP.NET apps, you can re-enable it in the project file. 
```xml
<PropertyGroup>
  <TargetFramework>net5.0</TargetFramework>
  <!-- Warning: Setting the following switch is *NOT* recommended in web apps. -->
  <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
</PropertyGroup>
```

danielegiallonardo:
Yes, they've introduced that tag in order to legacy applications to still run until all the references to BinaryFormatter in the depedencies are replaced, but it's still something they strongly DON'T recommend, since BinaryFormatter has security issues and it has been marked as deprecated in .net5+. 
In order to the library to be fully .Net5+-compliant, all the references to BinaryFormatter should be replaced with one of the suggested alternatives.

hazzik:
> BinaryFormatter should be replaced with one of the suggested alternatives.

It _could_ be replaced with one of the alternatives.

NHibernate uses BinaryFormatter in just two places: `SerializableType` and `SerializationHelper`. It is totally safe to enable `BinaryFormatter` if you are not using any of these features.

danielegiallonardo:
I know that a workaround exists, and I'm currently forced to adopt it, but are you suggesting that I should disable for a long-term an application-wide block of a high-risk OWASP reported vulnerability, implemented in a core class that is reported as deprecated (and I expect it to be removed in future versions of the .net sdk)?
Anyway, I'm not asking for a hot-fix, I'm just reporting an existing issue about NHibernate compatibiliy with .net5+.

fredericDelaporte:
Do we actually need to disable that block for running NHibernate on .Net 5? I think it is not even needed.

[...]

PR #3745, opened 2026-05-13
Add pluggable serialization strategy and make BinaryFormatter opt-in

ValResnick:
Introduce ISerializationStrategy and SerializationConfiguration.
Default to a throwing strategy unless explicitly configured, and add NHibernate property serialization.strategy_class for strategy setup.
Keep test bootstrap on BinaryFormatter for backward-compatible test behavior.

Refs #2603

## Questions for reviewers
- BinaryFormatterSerializationStrategy still needs to be moved out so that BinaryFormatter is no longer used in the core project.
Suggestions for the best target location are very welcome.

### DIFF
--- a/src/NHibernate.Test/UtilityTest/SerializationConfigurationFixture.cs
+++ b/src/NHibernate.Test/UtilityTest/SerializationConfigurationFixture.cs
@@ -32,9 +32,9 @@ namespace NHibernate.Test.UtilityTest
 		}
 
 		[Test]
-		public void CanUseBinaryFormatterStrategy()
+		public void CanSerialization()
 		{
-			TestConfigurationHelper.UseBinaryFormatterSerialization();
+			TestConfigurationHelper.UseTestSerialization();
 			var value = new[] { 1, 2, 3 };
 			var bytes = SerializationHelper.Serialize(value);
 			var copy = (int[]) SerializationHelper.Deserialize(bytes);
@@ -52,7 +52,7 @@ namespace NHibernate.Test.UtilityTest
 				[Environment.Dialect] = typeof(Dialect.GenericDialect).AssemblyQualifiedName,
 				[Environment.ConnectionDriver] = typeof(Driver.SqlClientDriver).AssemblyQualifiedName,
 				[Environment.ConnectionString] = "Server=(localdb)\\MSSQLLocalDB;Database=master;Integrated Security=SSPI",
-				[Environment.SerializationStrategy] = typeof(BinaryFormatterSerializationStrategy).AssemblyQualifiedName
+				[Environment.SerializationStrategy] = typeof(TestSerializationStrategy).AssemblyQualifiedName
 			};
 
 			var settingsFactory = new SettingsFactory();
@@ -68,7 +68,7 @@ namespace NHibernate.Test.UtilityTest
 
 			var sessionFactoryConfiguration = new SessionFactoryConfigurationBase();
 			sessionFactoryConfiguration.Properties[Environment.SerializationStrategy] =
-				typeof(BinaryFormatterSerializationStrategy).AssemblyQualifiedName;
+				typeof(TestSerializationStrategy).AssemblyQualifiedName;
 
 			var hibernateConfiguration = new StubHibernateConfiguration
 			{
@@ -79,7 +79,7 @@ namespace NHibernate.Test.UtilityTest
 
 			Environment.InitializeGlobalProperties(hibernateConfiguration);
 
-			Assert.That(SerializationConfiguration.Strategy, Is.TypeOf<BinaryFormatterSerializationStrategy>());
+			Assert.That(SerializationConfiguration.Strategy, Is.TypeOf<TestSerializationStrategy>());
 		}
 
 		private sealed class StubHibernateConfiguration : IHibernateConfiguration
--- a/src/NHibernate.Test/UtilityTest/SerializationConfigurationFixture.cs
+++ b/src/NHibernate.Test/UtilityTest/SerializationConfigurationFixture.cs
@@ -1,0 +1,64 @@
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+using NHibernate.Cfg;
+using NHibernate.Util;
+using NUnit.Framework;
+
+namespace NHibernate.Test.UtilityTest
+{
+	[TestFixture]
+	[NonParallelizable]
+	public class SerializationConfigurationFixture
+	{
+		private ISerializationStrategy _initialStrategy;
+
+		[SetUp]
+		public void SetUp()
+		{
+			_initialStrategy = SerializationConfiguration.Strategy;
+		}
+
+		[TearDown]
+		public void TearDown()
+		{
+			SerializationConfiguration.Strategy = _initialStrategy;
+		}
+
+		[Test]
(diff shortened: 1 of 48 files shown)


phillipc:
I like the direction of introducing a pluggable serialization strategy. But please make the default behavior fail fast, so nhibernate-code has no direct BinaryFormatter dependency.

The next step can be stricter: BinaryFormatter should be removed completely from the NHibernate core assembly.
In this draft, core still ships a BinaryFormatter-based strategy, which keeps the obsolete API in the production package.

I suggest this split:

1. Keep only the abstraction and secure default in core (strategy interface + throwing default).
2. Move any BinaryFormatter fallback implementation to NHibernate tests only, just to preserve test coverage and legacy test scenarios. _(It can be also use as template for legacy projects)_
3. Do not provide a built-in BinaryFormatter strategy for all users.

If projects still need legacy BinaryFormatter behavior, they should implement and register their own strategy in their .NET project and explicitly opt in there. That keeps NHibernate core clean and avoids endorsing BinaryFormatter in runtime code.

ValResnick:
I like this proposal a lot. It keeps the change simple, and since BinaryFormatter is effectively dead anyway, making it opt-in and moving it out of the core path feels like the right direction.

I’ll update this PR accordingly.

M-Kirchhoff:
The direction looks right. Removing BinaryFormatter from the core assembly and moving it into the test project is a clean step, given that BinaryFormatter is deprecated and a known security risk.
One thing worth ensuring before this lands: as this is a **breaking change**, it should be accompanied by a major version bump and clear release notes with a migration guide, so that existing users know they may need to implement and register their own ISerializationStrategy.

src/NHibernate/Cfg/SettingsFactory.cs
```diff
@@ -358,6 +359,25 @@ public Settings BuildSettings(IDictionary<string, string> properties)
 			return settings;
 		}
 
+		private static void ConfigureSerializationStrategy(IDictionary<string, string> properties)
```
phillipc: @ValResnick thx for your changes! Only a small review finding:

Is a process-global serialization strategy intended, or should strategy be tied to SessionFactory settings/context? I would prefer a process-global configuration, see `NHibernate.Cfg.Environment.InitializeGlobalProperties`

phillipc (approved):

src/NHibernate.Test/UtilityTest/SerializationConfigurationFixture.cs
```diff
@@ -0,0 +1,94 @@
...
+		public void ThrowsByDefaultWhenConfiguredWithThrowingStrategy()
+		{
+			SerializationConfiguration.Strategy = new ThrowingSerializationStrategy();
+			Assert.That(() => SerializationHelper.Serialize(new object()), Throws.TypeOf<SerializationException>());
+		}
+
+		[Test]
+		public void CanUseBinaryFormatterStrategy()
+		{
+			TestConfigurationHelper.UseBinaryFormatterSerialization();
```
fredericDelaporte: With #3662, tests are now run with .net10 and .net Framework.
So, binary serialization can no more work at all in tests when run under .net (core).
https://learn.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter.serialize#exceptions

Tests like this one would need some compilation directive like:
```c#
#if NET9_0_OR_GREATER
	// Check it throws PlatformNotSupportedException
	// ...
#else
	// Check it works.
	// ...
#endif
```

src/NHibernate/Util/SerializationConfiguration.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+
+namespace NHibernate.Util
+{
+	/// <summary>
+	/// Global serialization configuration used by NHibernate serialization features.
+	/// </summary>
+	public static class SerializationConfiguration
+	{
+		private static ISerializationStrategy _strategy = new ThrowingSerializationStrategy();
```
fredericDelaporte: This is a breaking change for .net8 and lower. (Nh is build targeting: .Net Fx 4.6.1, 4.8; .net core 2.0; .net std 2.0, 2.1; .net 6.0, 8.0, 10.0)

For merging this in the next minor, it should be:

```suggestion
		private static ISerializationStrategy _strategy =
#if NET9_0_OR_GREATER
			new ThrowingSerializationStrategy();
#else
			new BinaryFormatterSerializationStrategy();
#endif
```

The `BinaryFormatterSerializationStrategy` class would have to be added in the lib, but encapsulated wholly in a `#if !NET9_0_OR_GREATER`.

That way, it could be in the next minor, while still being removed out of .net10 builds.

src/NHibernate.Test/UtilityTest/SerializationConfigurationFixture.cs
fredericDelaporte: But still the .net 10 CI succeed on this test. Shouldn't they fail?

fredericDelaporte: Well, it works thanks to the NuGet package System.Runtime.Serialization.Formatters which adds back the functionality.

Then defaulting to throwing even with net10 remains a breaking change, since there is still a way to get it back with .net10.

Anyway, in the current state of the tests, no need to change anything here.

src/NHibernate/Util/SerializationConfiguration.cs
fredericDelaporte: It is actually still possible to use binary serialization with .net 10, as do the NH tests, through the [System.Runtime.Serialization.Formatters NuGet package](https://www.nuget.org/packages/system.runtime.serialization.formatters/).

So, on principle, we should default to BinaryFormatterSerializationStrategy even for net10.

But to better align with the change in net10, we could maybe still do something similar to .net10: removing it when targeting .net10, document this a a "possible breaking change" in the release, explaining how to restore the feature.

phillipc: I believe NHibernate should drop BinaryFormatter support by default to prevent vulnerabilities. While removing it is a breaking change for projects relying on binary serialization for complex types, using ORM mappings this way is generally considered an anti-pattern.

If a project absolutely requires this functionality, it can easily be re-enabled by configuring an external dependency to the BinaryFormatter package. This approach aligns also in .NET 8 and .NET 9.

ValResnick:
As suggested, BinaryFormatterSerializationStrategy is back in the core library, fully wrapped in _#if !NET9_0_OR_GREATER_ .
I verified that the compiled .NET 10 assembly contains no BinaryFormatter type or reference (decompiler-clean).

The default strategy is now framework-dependent: ThrowingSerializationStrategy on .NET 9+, BinaryFormatter on earlier targets to preserve backward compatibility.
  
To keep the test suite working on every Framework Version, the test
project has its own TestSerializationStrategy (BinaryFormatter-backed) instead of depending on the library type,  which is absent on .NET 9+.

fredericDelaporte (approved):
Thanks,

Possible breaking change for the release note:

> The NHibernate features using binary serialization will now throw a SerializationException for applications targeting .NET 10 or greater. These features can be restored by configuring a serialization strategy with your own custom strategy. (See Cfg.Environment.SerializationStrategy or set NHibernate.Util.SerializationConfiguration.Strategy.) See #3745.

PR #3754, opened 2026-05-24
Switch macos-13 tests to macos-26-intel

fredericDelaporte:
macos-13 has been removed out of GitHub actions.

### DIFF
--- a/.github/workflows/NetCoreTests.yml
+++ b/.github/workflows/NetCoreTests.yml
@@ -8,8 +8,9 @@ jobs:
       fail-fast: false
       matrix:
         DB: [SQLite]
-        # macos: SQLite.Core 1.x lacks an arm64 interop lib and so requires an intel macos image. Switching to SQLite 2
-        # (without .Core suffix) might allow to run on arm64. See https://system.data.sqlite.org
+        # macos: SQLite.Core 1.x lacks an arm64 interop lib and so requires an intel macos image. There are no latest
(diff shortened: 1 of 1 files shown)


.github/workflows/NetCoreTests.yml
```diff
@@ -8,7 +8,10 @@ jobs:
       fail-fast: false
       matrix:
         DB: [SQLite]
-        OS: [ubuntu-latest, windows-latest, macos-13]
+        # macos: System.Data.SQLite.Core 1.x lacks an arm64 interop lib and so requires an intel macos image. There are
+        # no latest such images, excepted -large but they require a paying GitHub account. https://github.com/actions/runner-images
+        # Switching to System.Data.SQLite 2 (without .Core suffix) might allow to run on arm64. See https://system.data.sqlite.org
+        OS: [ubuntu-latest, windows-latest, macos-26-intel]
```
fredericDelaporte, line 14: It seems there will never be a `-latest` image for macos-intel, because GitHub would be phasing out macOS intel images.

https://github.com/actions/runner-images/issues/13885#issuecomment-4191578408

>  Intel support on macOS is being phased out

So at some point we will have to switch to System.Data.SQLite 2 if we want to test macOS with GitHub, which will likely require a new NHibernate driver.

hazzik: The assembly name is the same across System.Data.SQLite 1, System.Data.SQLite 2, and System.Data.SQLite.Core

hazzik (approved):

Issue #3761, opened 2026-07-29
Change IdentityMap.EntryList to use IDictionaryEnumerator & return a generic IList


amilia-nabrari:
If we change `IdentityMap.EntryList` from `IList EntryList` to `IList<DictionaryEntry> EntryList` and use `IDictionaryEnumerator` (to get access to `.Entry` which avoids boxing) we can speed up the method and reduce allocations.

We could then change `IdentityMap.Entries` and `IdentityMap.ConcurrentEntries` to return a generic `IList<DictionaryEntry>` too (as they are the only usages I can find within NH itself).

If we consider this change in signature a breaking change, maybe we could implement internal typed methods with the new signatures.

The results look promising in my benchmark:

### With `IdentityMap.InstantiateSequenced`
| Method                | EntryCount | Mean         | Error       | StdDev      | Ratio | RatioSD | Gen0    | Gen1    | Gen2    | Allocated | Alloc Ratio |
|---------------------- |----------- |-------------:|------------:|------------:|------:|--------:|--------:|--------:|--------:|----------:|------------:|
| Old_EntryListSnapshot | 100        |   1,598.3 ns |    31.08 ns |    30.52 ns |  1.00 |    0.03 |  0.5836 |  0.0076 |       - |   7.17 KB |        1.00 |
| New_EntryListSnapshot | 100        |     415.9 ns |     8.15 ns |    14.49 ns |  0.26 |    0.01 |  0.1354 |  0.0005 |       - |   1.66 KB |        0.23 |
|                       |            |              |             |             |       |         |         |         |         |           |             |
| Old_EntryListSnapshot | 1000       |  15,885.6 ns |   311.12 ns |   332.90 ns |  1.00 |    0.03 |  5.7373 |  0.7324 |       - |  70.45 KB |        1.00 |
| New_EntryListSnapshot | 1000       |   4,292.4 ns |    78.21 ns |   109.63 ns |  0.27 |    0.01 |  1.2817 |  0.0687 |       - |  15.73 KB |        0.22 |
|                       |            |              |             |             |       |         |         |         |         |           |             |
| Old_EntryListSnapshot | 10000      | 182,198.7 ns | 3,522.40 ns | 4,325.82 ns |  1.00 |    0.03 | 57.3730 | 24.4141 |       - | 703.27 KB |        1.00 |
[...]

PR #3762, opened 2026-07-31
Change IdentityMap.EntryList to use IDictionaryEnumerator & return a generic IList

Fube:
As we discussed in https://github.com/nhibernate/nhibernate-core/issues/3761

### DIFF
--- a/src/NHibernate.Test/UtilityTest/IdentityMapFixture.cs
+++ b/src/NHibernate.Test/UtilityTest/IdentityMapFixture.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections;
-using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using NHibernate.Util;
 using NUnit.Framework;
@@ -59,20 +58,22 @@ namespace NHibernate.Test.UtilityTest
 		}
 
 		/// <summary>
-		/// Verify that GetEntries returns an ICollection that contains the same
+		/// Verify that ConcurrentEntities returns an ICollection that contains the same
 		/// Keys/Values as originally added into the IdentityMap.
 		/// </summary>
 		[Test]
-		public void GetEntries()
+		public void ConcurrentEntries()
 		{
-			var map = GetIdentityMap();
+			IDictionary map = GetIdentityMap();
 
 			map.Add(noHashCode1, value1);
 			map.Add(noHashCode2, value2);
 
-			// call GetEntries and verify it doesn't use the HashCode to build the 
+			// call ConcurrentEntries and verify it doesn't use the HashCode to build the 
 			// new list.
-			var concurrent = IdentityMap.GetEntries(map);
+#pragma warning disable CS0618 // Type or member is obsolete
+			ICollection concurrent = IdentityMap.ConcurrentEntries(map);
+#pragma warning restore CS0618 // Type or member is obsolete
 
 			Assert.AreEqual(2, concurrent.Count, "There are two elements in concurrent Map");
 			foreach (DictionaryEntry de in concurrent)
@@ -87,10 +88,37 @@ namespace NHibernate.Test.UtilityTest
 		}
 
 		/// <summary>
-		/// Tests that it is safe to modify the IdentityMap while iterating through GetEntries.
+		/// Verify that GetEntries returns an ICollection that contains the same
+		/// Keys/Values as originally added into the IdentityMap.
 		/// </summary>
 		[Test]
-		public void GetEntriesModification()
+		public void GetEntries()
+		{
+			var map = GetIdentityMap();
+
+			map.Add(noHashCode1, value1);
+			map.Add(noHashCode2, value2);
+
+			var entries = IdentityMap.GetEntries(map);
+
+			Assert.AreEqual(2, entries.Count, "There are two elements in the IdentityMap");
+			foreach (var de in entries)
+			{
+				var noCode = (NoHashCode) de.Key;
+				var noCodeValue = de.Value;
+
+				Assert.IsTrue(map.Contains(noCode), "The Key in the entries should have been in the original map's Keys");
+				Assert.IsTrue(noCodeValue == map[noCode],
+				              "The Value identified by the Key in entries should be the same as the IdentityMap");
+			}
+		}
+
+		/// <summary>
+		/// Tests that it is safe to modify the IdentityMap while iterating through the
+		/// ConcurrentEntities.
+		/// </summary>
+		[Test]
+		public void ConcurrentEntitiesModification()
 		{
 			NoHashCode noHashCode3 = new NoHashCode();
 			object value3 = new object();
@@ -98,11 +126,13 @@ namespace NHibernate.Test.UtilityTest
 			NoHashCode noHashCode4 = new NoHashCode();
 			object value4 = new object();
 
-			var map = GetIdentityMap();
+			IDictionary map = GetIdentityMap();
 			map.Add(noHashCode1, value1);
 			map.Add(noHashCode2, value2);
 
-			var concurrent = IdentityMap.GetEntries(map);
+#pragma warning disable CS0618 // Type or member is obsolete
+			ICollection concurrent = IdentityMap.ConcurrentEntries(map);
+#pragma warning restore CS0618 // Type or member is obsolete
 
 			for (int i = 0; i < concurrent.Count;)
 			{
@@ -111,6 +141,36 @@ namespace NHibernate.Test.UtilityTest
 
 				i++;
 				Assert.AreEqual(2, concurrent.Count, "Should still be 2 items in the concurrent ICollection");
+				Assert.AreEqual(2 + i, map.Count, "Should be " + (2 + i) + " items in the IdentityMap");
+			}
+		}
+
+		/// <summary>
+		/// Tests that it is safe to modify the IdentityMap while iterating through the
+		/// GetEntries snapshot.
+		/// </summary>
+		[Test]
+		public void GetEntriesModification()
+		{
+			var noHashCode3 = new NoHashCode();
+			var value3 = new object();
+
+			var noHashCode4 = new NoHashCode();
+			var value4 = new object();
+
+			var map = GetIdentityMap();
+			map.Add(noHashCode1, value1);
+			map.Add(noHashCode2, value2);
+
+			var entries = IdentityMap.GetEntries(map);
+
+			for (var i = 0; i < entries.Count;)
+			{
+				if (i == 0) map.Add(noHashCode3, value3);
+				if (i == 1) map.Add(noHashCode4, value4);
+
+				i++;
+				Assert.AreEqual(2, entries.Count, "Should still be 2 items in the GetEntries ICollection");
 				Assert.AreEqual(2 + i, map.Count, "Should be " + (2 + i) + " items in the IdentityMap");
 			}
 		}
--- a/src/NHibernate.Test/UtilityTest/IdentityMapFixture.cs
+++ b/src/NHibernate.Test/UtilityTest/IdentityMapFixture.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using NHibernate.Util;
 using NUnit.Framework;
@@ -69,9 +70,9 @@ namespace NHibernate.Test.UtilityTest
 			map.Add(noHashCode1, value1);
 			map.Add(noHashCode2, value2);
 
-			// call ConcurrentEntries and verify it doesn't use the HashCode to build the 
+			// call ConcurrentEntriesList and verify it doesn't use the HashCode to build the 
 			// new list.
-			ICollection concurrent = IdentityMap.ConcurrentEntries(map);
+			IList<DictionaryEntry> concurrent = IdentityMap.ConcurrentEntriesList(map);
 
 			Assert.AreEqual(2, concurrent.Count, "There are two elements in concurrent Map");
 			foreach (DictionaryEntry de in concurrent)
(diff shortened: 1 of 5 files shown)


src/NHibernate.Test/UtilityTest/IdentityMapFixture.cs
```diff
@@ -69,9 +70,9 @@ public void ConcurrentEntries()
 			map.Add(noHashCode1, value1);
 			map.Add(noHashCode2, value2);
 
-			// call ConcurrentEntries and verify it doesn't use the HashCode to build the 
+			// call ConcurrentEntriesList and verify it doesn't use the HashCode to build the 
 			// new list.
-			ICollection concurrent = IdentityMap.ConcurrentEntries(map);
+			IList<DictionaryEntry> concurrent = IdentityMap.ConcurrentEntriesList(map);
```
fredericDelaporte: Code practices can by modernised when modified.


```suggestion
			var concurrent = IdentityMap.ConcurrentEntriesList(map);
```

src/NHibernate.Test/UtilityTest/IdentityMapFixture.cs
```diff
@@ -102,7 +103,7 @@ public void ConcurrentEntitiesModification()
 			map.Add(noHashCode1, value1);
 			map.Add(noHashCode2, value2);
 
-			ICollection concurrent = IdentityMap.ConcurrentEntries(map);
+			IList<DictionaryEntry> concurrent = IdentityMap.ConcurrentEntriesList(map);
```
fredericDelaporte:
```suggestion
			var concurrent = IdentityMap.ConcurrentEntriesList(map);
```

src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
```diff
@@ -90,7 +90,7 @@ protected virtual void FlushCollections(IEventSource session)
 		{
 			log.Debug("Processing unreferenced collections");
 
-			ICollection list = IdentityMap.Entries(session.PersistenceContext.CollectionEntries);
+			IList<DictionaryEntry> list = IdentityMap.EntriesList(session.PersistenceContext.CollectionEntries);
```
fredericDelaporte:
```suggestion
			var list = IdentityMap.EntriesList(session.PersistenceContext.CollectionEntries);
```

src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
```diff
@@ -147,7 +147,7 @@ protected virtual void FlushEntities(FlushEvent @event)
 			// It is safe because of how IdentityMap implements entrySet()
 			IEventSource source = @event.Session;
 
-			ICollection list = IdentityMap.ConcurrentEntries(source.PersistenceContext.EntityEntries);
+			IList<DictionaryEntry> list = IdentityMap.ConcurrentEntriesList(source.PersistenceContext.EntityEntries);
```
fredericDelaporte:
```suggestion
			var list = IdentityMap.ConcurrentEntriesList(source.PersistenceContext.EntityEntries);
```

src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
```diff
@@ -174,7 +174,7 @@ protected virtual void PrepareCollectionFlushes(ISessionImplementor session)
 			// and reset reached, doupdate, etc.
 			log.Debug("dirty checking collections");
 
-			ICollection list = IdentityMap.Entries(session.PersistenceContext.CollectionEntries);
+			IList<DictionaryEntry> list = IdentityMap.EntriesList(session.PersistenceContext.CollectionEntries);
```
fredericDelaporte:
```suggestion
			var list = IdentityMap.EntriesList(session.PersistenceContext.CollectionEntries);
```

src/NHibernate/Event/Default/AbstractFlushingEventListener.cs
```diff
@@ -189,7 +189,7 @@ protected virtual void PrepareEntityFlushes(IEventSource session)
 			log.Debug("processing flush-time cascades");
 
 			var anything = Anything;
-			ICollection list = IdentityMap.ConcurrentEntries(session.PersistenceContext.EntityEntries);
+			IList<DictionaryEntry> list = IdentityMap.ConcurrentEntriesList(session.PersistenceContext.EntityEntries);
```
fredericDelaporte:
```suggestion
			var list = IdentityMap.ConcurrentEntriesList(session.PersistenceContext.EntityEntries);
```

src/NHibernate/Util/IdentityMap.cs
```diff
@@ -60,16 +60,42 @@ public static IDictionary InstantiateSequenced(int size)
 		/// </summary>
 		/// <param name="map">The IDictionary to get the enumeration safe list.</param>
 		/// <returns>A Collection of DictionaryEntries</returns>
+		[Obsolete("Use ConcurrentEntriesList instead.")]
```
fredericDelaporte: We add a comment for contributors to know easily since when it is obsolete. It would be 5.8.

For what this method is doing, I do not think there is a value in providing an alternative for it. It is easy enough to handle by user code if need be.

```suggestion
		// Since v5.8.
		[Obsolete("This method has no more usage in NHibernate and will be removed in a future version.")]
```

src/NHibernate/Util/IdentityMap.cs
```diff
@@ -60,16 +60,42 @@ public static IDictionary InstantiateSequenced(int size)
...
 		public static ICollection ConcurrentEntries(IDictionary map)
 		{
 			return ((IdentityMap)map).EntryList;
 		}
 
+		[Obsolete("Use EntriesList instead.")]
```
fredericDelaporte:
```suggestion
		// Since v5.8.
		[Obsolete("This method has no more usage in NHibernate and will be removed in a future version.")]
```

src/NHibernate/Util/IdentityMap.cs
```diff
@@ -60,16 +60,42 @@ public static IDictionary InstantiateSequenced(int size)
...
+		}
+
+		/// <summary>
+		/// Return the Dictionary Entries (as instances of <c>DictionaryEntry</c> in a list
+		/// that is safe from concurrent modification).  Ie - we may safely add new instances
+		/// to the underlying <c>IDictionary</c> during enumeration of the <c>Values</c>.
+		/// </summary>
+		/// <param name="map">The IDictionary to get the enumeration safe list.</param>
+		/// <returns>A typed list of DictionaryEntries, avoiding boxing when built and enumerated.</returns>
+		public static IList<DictionaryEntry> EntriesList(IDictionary map)
```
fredericDelaporte: For what does this utility method, better make it internal.

```suggestion
		internal static IList<DictionaryEntry> GetEntries(IDictionary map)
```

src/NHibernate/Util/IdentityMap.cs
```diff
@@ -60,16 +60,42 @@ public static IDictionary InstantiateSequenced(int size)
...
+		/// that is safe from concurrent modification).  Ie - we may safely add new instances
+		/// to the underlying <c>IDictionary</c> during enumeration of the <c>Values</c>.
+		/// </summary>
+		/// <param name="map">The IDictionary to get the enumeration safe list.</param>
+		/// <returns>A typed list of DictionaryEntries, avoiding boxing when built and enumerated.</returns>
+		public static IList<DictionaryEntry> ConcurrentEntriesList(IDictionary map)
+		{
+			return ((IdentityMap)map).EntryListTyped;
+		}
+
```
fredericDelaporte: As `Entries` and `ConcurrentEntriesList` are actually the same, I see no use to keep both. Better keep only one.


```suggestion
```

src/NHibernate/Util/IdentityMap.cs
```diff
@@ -220,15 +246,37 @@ public void CopyTo(Array array, int i)
 		/// 
 		/// Contains a copy (not that actual instance stored) of the DictionaryEntries in a List.
 		/// </summary>
+		[Obsolete("Use EntryListTyped instead.")]
```
fredericDelaporte:
```suggestion
		// Since v5.8.
		[Obsolete("Use GetEntries instead.")]
```

src/NHibernate/Util/IdentityMap.cs
```diff
@@ -220,15 +246,37 @@ public void CopyTo(Array array, int i)
...
 		public IList EntryList
+		{
+			get { return (IList) EntryListTyped; }
+		}
```
fredericDelaporte: Can be modernised as:

```suggestion
public IList EntryList => GetEntries();
```

The cast can be avoided with a further suggestion, changing the name, the return type and the member nature.

src/NHibernate/Util/IdentityMap.cs
fredericDelaporte: By convention in .net, properties should do little to no work. That is not the case of this one, so, we would better change it to a method.

That would allow to name it simply `GetEntries`, which will be supported as an overload of the static one that I also suggest to rename likewise.

Furthermore, I see no value here in hiding the concrete type with the `IList<>` interface. It was required to use the old untyped interface, but since we are going for the typed one, and since we yield a "standard" `List<>`, nothing custom, I think we would have better yielding it as-is.

```suggestion
		public List<DictionaryEntry> GetEntries
		{
			var list = new List<DictionaryEntry>(map.Count);
			if (map.Count == 0)
				return list;

			var enumerator = map.GetEnumerator();
			while (enumerator.MoveNext())
			{
				list.Add(enumerator.Entry);
			}

			return list;
		}
```

fredericDelaporte (commented):
Thanks for your contribution.

It looks good to me. I suggest some minor improvements.

(My suggestions are not complete. Some of them imply additional changes, the renaming ones, which you may have better to do with the IDE.)

src/NHibernate/Util/IdentityMap.cs
Fube: I will then also change `internal static IList<DictionaryEntry> GetEntries` to return a `List`
https://github.com/nhibernate/nhibernate-core/pull/3762#discussion_r3699802645

src/NHibernate/Util/IdentityMap.cs
```diff
@@ -220,19 +233,36 @@ public void CopyTo(Array array, int i)
...
-				foreach (DictionaryEntry de in map)
-				{
-					DictionaryEntry newEntry = new DictionaryEntry(de.Key, de.Value);
-					list.Add(newEntry);
-				}
-
 				return list;
 			}
+
+			var enumerator = map.GetEnumerator();
```
hazzik, line 259: Can we use `new List<DE>(map)` or even `map.ToList()`?

Also, I think

Fube: `new List<DE>(map)` does not work because `IDictionary` does not implement `IEnumerable<T>`.
`Enumerable.ToList()` does not work for the same reason.

src/NHibernate.Test/UtilityTest/IdentityMapFixture.cs
```diff
@@ -58,20 +59,20 @@ public void AddNoHashCode()
 		}
 
 		/// <summary>
-		/// Verify that ConcurrentEntities returns an ICollection that contains the same
+		/// Verify that GetEntries returns an ICollection that contains the same
 		/// Keys/Values as originally added into the IdentityMap.
 		/// </summary>
 		[Test]
-		public void ConcurrentEntries()
+		public void GetEntries()
```
hazzik: I think we still need original tests.

Fube: I'll revert the changes and add identical ones for `GetEntries`.

src/NHibernate/Util/IdentityMap.cs
```diff
@@ -220,19 +233,36 @@ public void CopyTo(Array array, int i)
...
+		/// You can safely iterate over this VIEW and modify the actual IdentityMap because the
+		/// VIEW is a copy of the contents, not a reference to the existing Map.
+		/// 
+		/// Contains a copy (not the actual instance stored) of the DictionaryEntries in a List.
+		/// </summary>
+		/// <remarks>
+		/// This uses <see cref="IDictionaryEnumerator"/> directly, which avoids boxing the
+		/// <see cref="DictionaryEntry"/> structs when building the snapshot list.
+		/// </remarks>
+		public List<DictionaryEntry> GetEntries()
```
hazzik, line 251: Can we have method `public List<KeyValuePair<TK,TV>> GetEntries<TK,TV>()`?

For a long time I wanted to replace IdentityMap with a generic version, but did not have enough resources do do so.

Fube: I would be happy to take a look and see what's possible.
I also wonder if we can replace it with something that does not pay such a heavy price for snapshotting state, maybe an implementation with a `ImmutableDict`.

fredericDelaporte: Beware of not extending too much the PR. Doing it in a subsequent PR could be preferable, if that addition causes a lot of changes.

Fube: By itself, having a `List<KeyValuePair<TK, TV>> GetEntries<TK, TV>()` just moves the casting inside `GetEntries` and the caller has to supply the types for `TK` and `TV`.

As @fredericDelaporte points out, actually making `IdentityMap` generic (`IdentityMap<TK, TV>`) is much more work and would incur many more changes.
@hazzik Do you still want me to add `GetEntries<TK, TV>` or do you want me to defer it to a future PR?

Fube: On my comment about replacing the backing data structure with something that does not pay a heavy price for snapshotting, we definitely can implement something, but it will not be trivial.

None of the built-in std lib collection types that I can find correspond to what we are looking for (`ImmutableDictionary` is the closest, but its performance, both in CPU & memory, is horrendous for this use case).

I made a copy-on-write-when-snapshot-active dictionary and it performs much better than the existing eager snapshot solution, but it makes other operations more expensive.
If taking on this complexity is something the library is willing to do, I can continue investigating this for a future PR.

hazzik (approved):

PR #3771, opened 2026-08-08
Move Firebird Windows tests from AppVeyor to GitHub Actions

hazzik:
Windows runners cannot run the Linux Firebird image used by the Linux jobs, so the server is installed from its chocolatey package instead.


hazzik:
AppVeyor builds are unaffordable slow now.

fredericDelaporte (approved):

PR #3774, opened 2026-08-09
Keep the Firebird test database loaded between test fixtures

hazzik:
The test suite clears the Firebird connection pool after each test fixture. Then the database has no connections, and the server removes the database from memory. The next fixture reads the database again from the disk. A test run does this approximately 2500 times.

LINGER tells the server to keep the database in memory for 3 seconds after the last connection closes.


hazzik:
Works opposite of what was expected

PR #3794, opened 2026-08-13
Support for `for update nowait` and `for update of` locks in the MySQL 8 dialect

hazzik:
Supports `for update nowait` and `for update of` with MySQL 8. MySQL supports both clauses from 8.0.1.

Run tests with MySQL 8.4 and the MySQL 8 dialect. The MySQL 8 saves GUID values in `char(36)` columns, so removed `Old Guids=True` from the connection strings.

Installs the MySQL from the archives, because the chocolatey package downloads from the CDN, which keeps only the newest patch release.

### DIFF
--- a/src/NHibernate.Test/Linq/QueryLock.cs
+++ b/src/NHibernate.Test/Linq/QueryLock.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Linq;
 using System.Transactions;
-using NHibernate.Dialect;
 using NHibernate.DomainModel.Northwind.Entities;
 using NHibernate.Driver;
 using NHibernate.Engine;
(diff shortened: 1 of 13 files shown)


src/NHibernate.Test/Linq/QueryLock.cs
```diff
@@ -190,6 +205,11 @@ orderby e.CompanyName
 
 		private void AssertSeparateTransactionIsLockedOut(string customerId)
 		{
+			Assume.That(
+				TestDialect.SupportsNoWaitLock || !TestDialect.HasBrokenQueryTimeoutOnLockWait,
```
fredericDelaporte, line 209: So, supporting this feature in MySql is only partially tested, due to this limitation.

fredericDelaporte (approved):

fredericDelaporte (approved):

src/NHibernate.Test/Linq/QueryLock.cs
hazzik: 5.7 does not support `nowait` and any provider 8.0.31+ has `CommandTimeout` broken for async case.

8+ supports `nowait` but `CommandTimeout` on the provider is still broken. But the test works, because the query aborts at 0sec without waiting.

PR #3812, opened 2026-08-16
Add test results job to GitHub Actions workflow

hazzik:
I'm planning to make this check required in a PR/merge queue gate.


fredericDelaporte (commented):
I do not understand what is the added value of this check.

If the tests fail in some build, won't the build be failed too?

hazzik:
It fails if one of the test in the matrix fails, and passes only if all are green. The benefit is that we can add only one gate to the merge queue instead of adding all ~20

hazzik:
Also, it waits for all tests to pass. Currently we have only SQLite tests in the gate, and they are quite fast (~8 minutes each). It means there is a possibility that PR is merged before all tests passed. Previously slow single-runner AppVeyor was that gate that added latency.

fredericDelaporte (approved):

PR #3823, opened 2026-08-18
Linq: call RemoveRedundantCast in pre transform

### LINKED ISSUE #3802: NHibernate 5.7.0 fails to translate `string[].Contains()` calls in LINQ queries
A fix was recently introduced for issues caused by calling `.Contains()` on certain types in 5.7.0 when running on dotnet 10

https://github.com/nhibernate/nhibernate-core/pull/3721

I would have expected this fix to work when `Contains()` is called on a simple `string[]`, but it does not. A reproduction of the issue is here: 

https://github.com/heathprovost/NHibernateIssueReproduction

Oddly, it DOES work if you just use `new[]` instead of `new string[]`. Im not sure I understand how or why that would be the case.

ps. 5.6.1 fails in BOTH cases, so the fix does work in some scenarios, but not all (or at least not in cases I would have expected it to work)

Fube:
Fixes #3802

### DIFF
--- a/src/NHibernate.Test/Linq/WhereTests.cs
+++ b/src/NHibernate.Test/Linq/WhereTests.cs
@@ -487,6 +487,96 @@ namespace NHibernate.Test.Linq
 
 			Assert.That(query.Count, Is.EqualTo(2));
 		}
+		
+		[Test(Description = "GH-3802")]
+		public void UsersWithTypedArrayContains()
+		{
+			var names = new string[] { "ayende", "rahien" };
+
+			var query = (from user in db.Users
+			             where names.Contains(user.Name)
+			             select user).ToList();
+
+			Assert.That(query.Count, Is.EqualTo(2));
+		}
+		
+		[Test(Description = "GH-3802")]
+		public void UsersWithTypedArrayContainsWithExplicitCast()
+		{
+			var names = new string[] { "ayende", "rahien" };
+
+			var query = db.Users.Where(user => ((string[])names).Contains(user.Name)).ToList();
+
+			Assert.That(query.Count, Is.EqualTo(2));
+		}
+		
+		[Test(Description = "GH-3802")]
+		public void UsersWithTypedArrayContainsWithExplicitAs()
+		{
+			var names = new string[] { "ayende", "rahien" };
+
+			var query = db.Users.Where(user => (names as string[]).Contains(user.Name)).ToList();
+
+			Assert.That(query.Count, Is.EqualTo(2));
+		}
+		
+		[Test(Description = "GH-3802")]
+		public void UsersWithTypedArrayEnumerableContainsWithExplicitCast()
+		{
+			var names = new string[] { "ayende", "rahien" };
+
+			var query = db.Users.Where(user => Enumerable.Contains((string[])names, user.Name)).ToList();
+
+			Assert.That(query.Count, Is.EqualTo(2));
+		}
+		
+
+		[Test(Description = "GH-3802")]
+		public void UsersWithTypedArrayEnumerableContainsWithExplicitAs()
+		{
+			var names = new string[] { "ayende", "rahien" };
+
+			var query = db.Users.Where(user => Enumerable.Contains((names as string[]), user.Name)).ToList();
+
+			Assert.That(query.Count, Is.EqualTo(2));
+		}
+
+#nullable enable
+		[Test(Description = "GH-3802")]
+		public void UsersWithArrayContainsNullableEnabled()
+		{
+			var names = new[] { "ayende", "rahien" };
+
+			var query = (from user in db.Users
+			             where names.Contains(user.Name)
+			             select user).ToList();
+
+			Assert.That(query.Count, Is.EqualTo(2));
+		}
+
+		// The compiler casts the array to its own type before it converts it to a span, because an expression
+		// tree cannot hold the nullable annotations. The cast must not hide the evaluated array from the
+		// generator.
+		[Test(Description = "GH-3802")]
+		public void UsersWithTypedArrayContainsNullableEnabled()
+		{
+			var names = new string[] { "ayende", "rahien" };
+
+			var query = db.Users.Where(user => names.Contains(user.Name)).ToList();
+
+			Assert.That(query.Count, Is.EqualTo(2));
+		}
+		
+		[Test(Description = "GH-3802")]
+		public void UsersWithTypedArrayEnumerableContainsNullableEnabled()
+		{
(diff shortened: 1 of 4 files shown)


src/NHibernate.Test/Linq/WhereTests.cs
```diff
@@ -487,6 +487,100 @@ where names.Contains(user.Name)
...
+		}
+		
+		[Test(Description = "GH-3802")]
+		public void UsersWithTypedArrayEnumerableContainsNullableEnabled()
+		{
+			var names = new string[] { "ayende", "rahien" };
+
+			var query = new List<User>();
+			var res = db.Users.Where(user => Enumerable.Contains(names, user.Name));
+			foreach (var user1 in res) query.Add(user1);
```
fredericDelaporte: Enumerating directly the `Queryable` prevents the generation of the async counterpart test.

And for clarity, I prefer `foreach` one-liners to be avoided.
```suggestion
			var res = db.Users.Where(user => Enumerable.Contains(names, user.Name)).ToList();
			foreach (var user1 in res)
				query.Add(user1);
```

src/NHibernate.Test/Linq/WhereTests.cs
```diff
@@ -487,6 +487,100 @@ where names.Contains(user.Name)
...
+		// tree cannot hold the nullable annotations. The cast must not hide the evaluated array from the
+		// generator.
+		[Test(Description = "GH-3802")]
+		public void UsersWithTypedArrayContainsNullableEnabled()
+		{
+			var names = new string[] { "ayende", "rahien" };
+
+			var query = new List<User>();
+			var res = db.Users.Where(user => names.Contains(user.Name));
+			foreach (var user1 in res) query.Add(user1);
```
fredericDelaporte:
```suggestion
			var res = db.Users.Where(user => names.Contains(user.Name)).ToList();
			foreach (var user1 in res)
				query.Add(user1);
```

fredericDelaporte (dismissed):

hazzik:
I had a 2 questions for the RemoveRedundantCast, when thinking about fixing this:
- Can we have RemoveRedundantCast only in one place in pre-transform?
- Are we sure RemoveRedundantCast removes only unnecessary casts?

hazzik (dismissed):

Fube:
> I had a 2 questions for the RemoveRedundantCast, when thinking about fixing this:
> - Can we have RemoveRedundantCast only in one place in pre-transform?
> - Are we sure RemoveRedundantCast removes only unnecessary casts? 

I will investigate and get back to you.

Fube:
> Are we sure RemoveRedundantCast removes only unnecessary casts?

@hazzik The first `if` does not remove an important cast, but the second one (the one that tries to simplify casts) can remove important casts.
It can:
1. [Remove a cast that should throw](https://dotnetfiddle.net/h0czvf)
2. Drop the inner node's explicit conversion method if one is defined
3. Change null check semantics by not checking if the outer node is a `TypeAs`

I think only the 2nd issue can affect SQL generation, and even then it is contrived.

Here you can see the issue can only be triggered by manually constructing an expression tree and attaching a conversion method that the CLR would not be able to discover by itself
```csharp
	private struct Meters : IEquatable<Meters>
	{
		public int Value;

		public static Meters FromCentimeters(int centimeters) => new Meters { Value = centimeters / 100 };

		public bool Equals(Meters other) => Value == other.Value;
	}
	
	private struct Meters2 : IEquatable<Meters2>
	{
		public int Value;

		public static explicit operator Meters2(int centimeters) => new Meters2 { Value = centimeters / 100 };

		public bool Equals(Meters2 other) => Value == other.Value;
	}
	
        // This test passes
	[Test]
	public void DoubleCastWithExplicitConversionMethodBreaksPreTransform()
	{
		var conversionMethod = typeof(Meters).GetMethod(nameof(Meters.FromCentimeters));

		var param = Expression.Parameter(typeof(NumericEntity), "o");
		var integerProperty = Expression.Property(param, nameof(NumericEntity.Integer));

		// (Meters?)(Meters) o.Integer
		// The inner cast carries an explicit conversion Method
		var innerConvert = Expression.Convert(integerProperty, typeof(Meters), conversionMethod);
		var outerConvert = Expression.Convert(innerConvert, typeof(Meters?));
		var selector = Expression.Lambda<Func<NumericEntity, Meters?>>(outerConvert, param);

		// Evaluates correctly on its own.
		var compiled = selector.Compile();
		Assert.That(compiled(new NumericEntity { Integer = 250 }), Is.EqualTo(new Meters { Value = 2 }));

		var query = db.NumericEntities.Select(selector);

		var ex = Assert.Throws<InvalidOperationException>(() =>
			                                                  NhRelinqQueryParser.PreTransform(
				                                                  query.Expression,
				                                                  new PreTransformationParameters(QueryMode.Select, Sfi)));

		Assert.That(ex.Message, Does.Contain("coercion operator"));
	}
	
	// This test fails
	[Test]
	public void DoubleCastWithExplicitConversionMethodBreaksPreTransform2()
	{
		Expression<Func<NumericEntity, Meters2?>> selector = o => (Meters2?) (Meters2) o.Integer;
		
		// Evaluates correctly on its own.
		var compiled = selector.Compile();
		Assert.That(compiled(new NumericEntity { Integer = 250 }), Is.EqualTo(new Meters2 { Value = 2 }));

		var query = db.NumericEntities.Select(selector);

		var ex = Assert.Throws<InvalidOperationException>(() =>
			                                                  NhRelinqQueryParser.PreTransform(
				                                                  query.Expression,
				                                                  new PreTransformationParameters(QueryMode.Select, Sfi)));

		Assert.That(ex.Message, Does.Contain("coercion operator"));
	}
```

I do not think it is entirely impossible to end up in a situation where a non-discoverable conversion method is attached or maybe a situation where the outer convert has another explicitly defined conversion method that is different from the inner's conversion method (like adding `public static explicit operator Meters2?(int centimeters) => throw new InvalidOperationException();`), but I was unable to construct such a situation.

[The PR](https://github.com/nhibernate/nhibernate-core/commit/7360394438bc1397e11d60a2d54d413cfbd786c0) that adds the second `if` references https://github.com/nhibernate/nhibernate-core/issues/2490

Fube:
> Can we have RemoveRedundantCast only in one place in pre-transform?

@hazzik I am not sure.
Although `NhRelinqQueryParser.QueryParser` is private, it is exposed publicly through `NhRelinqQueryParser.Parse`.
Removing `RemoveRedundantCast` from `NhRelinqQueryParser`'s static initializer would mean introducing a breaking change for users that invoke `Parse`.

hazzik (approved):

PR #3836, opened 2026-08-28
Map Firebird UDF functions to built-in functions

hazzik:
Map the HQL function names that come from the `ib_udf` and `fbudf` external libraries to the equivalent Firebird built-in functions. The `sqtr` name was a spelling error and now calls `sqrt`.

These libraries are not installed by default, and Firebird 4 does not permit external functions. The names could not work on a standard installation.

### DIFF
--- a/src/NHibernate/Dialect/FirebirdDialect.cs
+++ b/src/NHibernate/Dialect/FirebirdDialect.cs
@@ -456,7 +456,7 @@ namespace NHibernate.Dialect
 		{
 			OverrideStandardHQLFunctions();
 			RegisterFirebirdServerEmbeddedFunctions();
-			RegisterExternalFbAndIbStandardUDFs();
+			RegisterBuiltInFunctions();
 		}
 
 		private void OverrideStandardHQLFunctions()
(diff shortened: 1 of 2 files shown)


src/NHibernate/Dialect/FirebirdDialect.cs
```diff
@@ -519,7 +519,7 @@ private void RegisterMathematicalFunctions()
 			RegisterFunction("rand", new NoArgSQLFunction("rand", NHibernateUtil.Double));
 			RegisterFunction("random", new NoArgSQLFunction("rand", NHibernateUtil.Double));
 			RegisterFunction("sign", new StandardSQLFunction("sign", NHibernateUtil.Int32));
-			RegisterFunction("sqtr", new StandardSQLFunction("sqtr", NHibernateUtil.Double));
```
fredericDelaporte, line 522: Wasn't this not working at all, since the spelling error was in the native name, too?

In such case, why keep the spelling error in HQL, if it was not functional? I mean, register `sqrt` in HQL if it is not already in the base dialect, and remove `sqtr` if it was not working at all.

PR #3837, opened 2026-08-28
Register more Firebird built-in functions

hazzik:
Register these Firebird built-in functions:

- `exp`
- `hash`
- `datediff`
- `maxvalue`, `minvalue` — native names aliased by `greatest` and `least`
- `uuid_to_char` — a native name aliased by `strguid`
- `gen_uuid` — a native name aliased by `new_uuid`

### DIFF
--- a/src/NHibernate/Dialect/FirebirdDialect.cs
+++ b/src/NHibernate/Dialect/FirebirdDialect.cs
@@ -470,12 +470,16 @@ namespace NHibernate.Dialect
 			// Modulo does not throw for decimal parameters but they are casted to int by Firebird, which produces unexpected results
 			RegisterFunction("mod", new ModulusFunction(false, false));
 			RegisterFunction("greatest", new StandardSQLFunction("maxvalue"));
+			RegisterFunction("maxvalue", new StandardSQLFunction("maxvalue"));
 			RegisterFunction("least", new StandardSQLFunction("minvalue"));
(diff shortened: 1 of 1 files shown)


src/NHibernate/Dialect/FirebirdDialect.cs
```diff
@@ -470,12 +470,16 @@ private void OverrideStandardHQLFunctions()
 			// Modulo does not throw for decimal parameters but they are casted to int by Firebird, which produces unexpected results
 			RegisterFunction("mod", new ModulusFunction(false, false));
 			RegisterFunction("greatest", new StandardSQLFunction("maxvalue"));
+			RegisterFunction("maxvalue", new StandardSQLFunction("maxvalue"));
 			RegisterFunction("least", new StandardSQLFunction("minvalue"));
+			RegisterFunction("minvalue", new StandardSQLFunction("minvalue"));
```
fredericDelaporte, line 475: I do not see the value of registering the Firebird native name in HQL, instead of only having the standard name registered for HQL.

For me, NHibernate is about abstracting away vendor specificities. So, why should we expose their specific function names in HQL?

fredericDelaporte: It seems some dialect already does this. I guess it is for convenience.

fredericDelaporte (approved):

PR #3842, opened 2026-08-28
Increase Firebird 4 id length and decimal precision limits

hazzik:
Raise the maximum alias length of the Firebird 4 dialect to 63 characters, and the maximum precision of a decimal column to 29 digits.

Firebird 4 increases the identifier length limit from 31 to 63 characters, and the precision of exact numeric types from 18 to 38 digits. 29 digits is the maximum that a .NET decimal can hold. Before this change, a mapping with a precision above 18 silently became `DECIMAL(18, 5)`.

This changes the generated DDL and the generated aliases of an application that uses the Firebird 4 dialect.


hazzik:
Split into #3843 (identifier length) and #3844 (decimal precision).

PR #3843, opened 2026-08-28
Increase Firebird 4 identifier length limit to 63 characters

hazzik:
Increase the maximum alias length of the Firebird 4 dialect from 30 to 63 characters. Firebird 4 increases the identifier length limit from 31 to 63 characters.

This changes the aliases that an application generates with the Firebird 4 dialect.


fredericDelaporte (dismissed):

fredericDelaporte (dismissed):
It will likely get conflicted again by the other PR.

fredericDelaporte (approved):

PR #3850, opened 2026-08-28
Publish dev NuGet package from GitHub Actions

### LINKED ISSUE #3278: Migrate 'Publish Dev NuGet Package' build from TeamCity to GitHub Actions
Running TC build agent cost me money (~$65 per month), so I would like it to be decommissioned over time.

hazzik:
Add a `Publish dev build` workflow that packs `NHibernate` and pushes the development NuGet package to the Cloudsmith feed. 

It runs on Windows after the `.NET Core` workflow succeeds on a push to `master` or a `*.x` branch, so the package keeps its strong-name signatures and only ships once the tests pass. The `dev` suffix and build number come from the workflow, replacing the TeamCity build and its `NHibernate.dev.props` file.

Fixes #3278

### DIFF
--- a/.github/workflows/PublishDevBuild.yml
+++ b/.github/workflows/PublishDevBuild.yml
@@ -1,0 +1,41 @@
+name: Publish dev build
+
+on:
+  workflow_run:
+    workflows: [".NET Core"]
+    types: [completed]
+    branches: [master, '*.x']
+
+concurrency:
+  group: publish-dev-build
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    if: github.event.workflow_run.event == 'push' && github.event.workflow_run.conclusion == 'success'
+    # Windows produces real strong-name signatures; other platforms only public-sign the assemblies.
+    runs-on: windows-latest
+    name: Publish
+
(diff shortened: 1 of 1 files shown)


.github/workflows/PublishDevBuild.yml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  publish:
+    if: github.event.workflow_run.event == 'push' && github.event.workflow_run.conclusion == 'success'
+    # Windows produces real strong-name signatures; other platforms only public-sign the assemblies.
+    runs-on: windows-latest
+    name: Publish
+
+    env:
+      VersionSuffix: dev
+      BuildNumber: ${{ github.event.workflow_run.run_number }}
```
bahusoid, line 22: Shouldn't you  add last used number to it (5060 or more) so new published builds won't be considered a lower version?

hazzik: Not really, the minor/patch versions have shifted, so it wouldn’t overlap

fredericDelaporte (approved):

Issue #3661, opened 2025-03-11
DbDataReader left open when cancellation occurs in AdvanceAsync after ExecuteReaderAsync


drakke:
## Description
There's a risk `DbDataReader` won't be disposed of in `GetResultSetAsync` method within the `src\NHibernate\Async\Loader\Loader.cs` class. When a cancellation token is canceled after the `DbDataReader` is successfully obtained from `session.Batcher.ExecuteReaderAsync()` but during the `AdvanceAsync` method call, the reader is not closed.

The issue occurs due to the catch block for `OperationCanceledException` that throws without closing the reader:

```csharp
catch (OperationCanceledException) { throw; }
catch (Exception sqle)
{
    ADOExceptionReporter.LogExceptions(sqle);
    session.Batcher.CloseCommand(st, rs);
    throw;
}
```

## Steps to Reproduce
I manually changed the NH source code to force the error to happen. Adding the following right after the DataReader initially gets set:
```csharp
rs = await (session.Batcher.ExecuteReaderAsync(st, cancellationToken)).ConfigureAwait(false);

// Added code on row 1064 to force OperationCanceledException:
CancellationTokenSource cts = new CancellationTokenSource();
cts.Cancel();
cts.Token.ThrowIfCancellationRequested();
```

Then I added this test to `src\NHibernate.Test\Async\Legacy\SQLLoaderTest.cs`:
```csharp
[Test]
public void ExpiredCancellationTokenThrowsOperationCanceledException()
{
    if (Dialect is Oracle8iDialect)
    {
        return;
    }

    Assert.ThrowsAsync<OperationCanceledException>(() => InnerExpiredCancellationTokenThrowsOperationCanceledException());
}

public async Task InnerExpiredCancellationTokenThrowsOperationCanceledException()
{
    using ISession session = OpenSession();
    using ITransaction txn = session.BeginTransaction();

    IQuery q = session.CreateSQLQuery(@"
        SELECT TOP 100000 a.*, b.* 
        FROM sys.objects a
        CROSS JOIN sys.objects b
        ORDER BY a.name");
    await (q.ListAsync());
}
```

This leads to an unexpected `InvalidOperationException` while rolling back the transaction rather than the expected `OperationCanceledException`.

```
  Message: 
      Expected: <System.OperationCanceledException>
      But was:  <System.InvalidOperationException: There is already an open DataReader associated with this Command which must be closed first.
[...]

PR #3852, opened 2026-08-31
Fixed race condition leaking resources

### LINKED ISSUE #3835: DbDataReader leak after cancellation that even survives ISession.Dispose(), e.g. on SQLite this stalls the whole application until the garbage collector happens to run
# Summary

DbDataReader leak after cancellation that can even survive ISession.Dispose(). For example, on SQLite this can stall the entire application until the garbage collector happens to run, especially when eagerly loaded data structures take time to load and cancellation occurs during the database operation.

This issue is similar to, but not exactly the same as, issue #3661 

This issue caused real-world bug reports about "database locked" (SQLITE_BUSY) in our SQLite-based application, because SQLite only allows one writer lock at a time and, after cancellation, there were suddenly two locks involved (one leaked from the previously cancelled and disposed ISession).

The issue was tracked to the NHibernate 5 source code by Claude Code Opus 5, which also provided a bug fix.

The only currently known workaround is to disable, i.e. not use, CancellationToken on NHibernate statements, because any of those statements may leak resources and subsequently block databases such as SQLite or exhaust the connection pool of other database providers until the Garbage Collector finalizes and disposes the leaked DbDataReader instances.

A description generated by Claude Code Opus 5 is provided below, along with a .patch that fixes the issue and a .patch containing a unit test that reproduces the problem and can be used to verify the fix.

# DESCRIPTION By Claude Code Opus 5 including Suggestion for Fix 

Tested with NHibernate 5.6.0, release commit `2b0e1f1d6e3bad8d53a53e0b51622fde57c6bd12`.

If a query is cancelled after its `DbDataReader` has already been created, the reader can stay open. There
are two places where this happens. They look similar but they are not the same bug, because in one of them
the reader is still known to the batcher and in the other it is not.

## Window A: `AbstractBatcher.DoExecuteReader(Async)`, not reported before

```csharp
var reader = await cmd.ExecuteReaderAsync(cancellationToken);      // (A) reader now exists
if (reader == null) throw new InvalidOperationException(...);
return _factory.ConnectionProvider.Driver.SupportsMultipleOpenReaders
    ? reader                                                       // no further token check
    : await NHybridDataReader.CreateAsync(reader, cancellationToken);   // (B)
```

`NHybridDataReader.CreateAsync` only checks the token twice and then wraps the reader:

```csharp
public static Task<NHybridDataReader> CreateAsync(DbDataReader reader, CancellationToken cancellationToken)
{
    if (cancellationToken.IsCancellationRequested)
        return Task.FromCanceled<NHybridDataReader>(cancellationToken);   // reader is dropped
    return CreateAsync(reader, false, cancellationToken);                 // note inMemory: false
}

public static async Task<NHybridDataReader> CreateAsync(DbDataReader reader, bool inMemory, CancellationToken cancellationToken)
{
    cancellationToken.ThrowIfCancellationRequested();                     // reader is dropped
    ...
}
```

`inMemory` is `false` on this path, so `NDataReader` is never called. That is the only method here with a
`finally { reader.Close(); }`, so nothing cleans up. The `OperationCanceledException` then travels out of
`ExecuteReaderAsync` before `_readersToClose.Add(reader)` runs, so the reader is not registered anywhere.
`_readersToClose` is private, so at that point no public API can reach the reader any more:

* `ISession.Dispose()` cannot close it, because `CloseCommands()` only closes readers it knows about.
* `IBatcher.HasOpenResources` cannot even tell you it happened. It is
  `_commandsToClose.Count > 0 || _readersToClose.Count > 0`, and the cancelled command is still in the
  command set, so it returns `true` either way.

Only the garbage collector can still release the reader.

This is not a tight race that is hard to hit. With a driver that has no real async ADO.NET support
(System.Data.SQLite is one), `cmd.ExecuteReaderAsync` runs `ExecuteReader()` synchronously and returns an
already completed task. The `await` at (A) never suspends, so the window is as long as preparing the
statement and fetching the first row takes. We measured 15 to 17 ms for one eager collection load, and a
normal `Task.Delay` hits it.

### What this does on SQLite

`SQLite20Driver.SupportsMultipleOpenReaders` is `false`, so SQLite always takes branch (B).

The leaked reader keeps its `sqlite3_stmt` alive. `sqlite3_close_v2` therefore cannot finish and leaves the
connection as a zombie (`eOpenState == SQLITE_STATE_ZOMBIE`) that still holds a SHARED lock on the database
file. We found exactly this in a process dump, and the surviving statement was at `errCode == SQLITE_ROW`,
so it had been abandoned in the middle of a result set. That fits this window, because `ExecuteReader()`
already steps once to fetch the first row.

The result is not one failed query. It stops the application:

* The next write cannot get its EXCLUSIVE lock, so it escalates to a PENDING lock. In our dump the writer
  sat there with `winFile.locktype == 3`.
* While a PENDING lock is held, SQLite does not grant any new SHARED lock. So new reads are refused too,
  not just writes.
* System.Data.SQLite keeps retrying for the whole command timeout, so the call does not return quickly
  either.
* If the application serialises access to the database file, everything else queues up behind that stuck
  write.

This lasts until a garbage collection happens to run and finalizes the abandoned reader. That is why the
problem looks random and shows up far away from the code that caused it.

On a provider that pools connections, the leaked reader keeps its connection busy for the same unpredictable
amount of time. We did not test whether that is enough to exhaust a pool, so we are only describing the
mechanism here.

DbDataReader leak after cancellation that even survives ISession.Dispose(), e.g. on SQLite this stalls the
whole application until the garbage collector happens to run. It is especially likely when eagerly loaded
data structures take time to load and the cancellation happens in the middle of the database operation.

This issue caused real bug reports about "database is locked" in our SQLite based application. What users
see is a `SQLiteException` with `SQLiteErrorCode.Busy` (`SQLITE_BUSY`, result code 5) and the message
`database is locked`, thrown by a `COMMIT` after System.Data.SQLite has spun in its busy retry loop for the
whole command timeout. The cause is that the leaked reader keeps a SHARED lock on the database file, and
SQLite cannot grant the EXCLUSIVE lock that a write needs while any SHARED lock is still held. Those locks
are per file handle, not per process, so the application ends up blocking itself: the leaked reader from an
already cancelled and disposed ISession blocks every later write from the same application.

Raising `BusyTimeout`, which is the usual advice for `SQLITE_BUSY`, does not reliably help here. With normal
contention the lock goes away when the other transaction finishes, so waiting longer works. In this case the
lock is only released when the garbage collector finalizes the abandoned reader, and nothing says when that
will be. If the process is not allocating much at the time, no collection may run for a long while, so any
finite timeout can expire first. Waiting longer only makes the stall longer.

## Window B: `Loader.GetResultSet(Async)`, this one is #3661

`GetResultSetAsync` closed the reader and the command only in `catch (Exception sqle)`. The separate
`catch (OperationCanceledException) { throw; }` did not clean anything up, so a reader that was already
created earlier in the same `try` stayed open when a cancellation surfaced after that point.

## Why this is not a duplicate of #3661

#3661 is Window B. This report adds Window A, and the two differ where it counts:

| | #3661 / Window B | Window A |
|---|---|---|
| reader is in `_readersToClose` | yes | no |
| `ISession.Dispose()` closes it | yes | no |
| worst case | cleanup is late | database unusable until the GC runs |

Fixing #3661 on its own does not fix Window A, and fixing Window A does not fix #3661.

The patch below also does not use the fix suggested in the #3661 thread
(`catch (OperationCanceledException) { session.Batcher.CloseReader(rs); throw; }`). Two reasons:

* `CloseReader(rs)` only closes the reader and leaves the `DbCommand` in `_commandsToClose`, while the
  existing general `catch` called `CloseCommand(st, rs)`. Doing the cleanup in a `finally` keeps both paths
  the same.
* A `finally` with a success flag covers every way out of the method instead of listing exception types, so
  it still works if someone adds another `catch` later.

Please note the patch moves the existing `CloseCommand(st, rs)` out of `catch (Exception sqle)` and into the
`finally`. Behaviour on the general exception path is unchanged, and cancellation is now covered as well.

## Suggested fix

Both changes are in the synchronous files. The `Async/` files in the patches are generated output, not hand
written (see "What we checked").

`src/NHibernate/AdoNet/AbstractBatcher.cs`, `DoExecuteReader`:

```diff
+			DbDataReader reader = null;
 			try
 			{
-				var reader = cmd.ExecuteReader();
+				reader = cmd.ExecuteReader();
 				if (reader == null)
 				{
 					throw new InvalidOperationException("The query execution has yielded a null reader. (Has it been canceled?)");
 				}
-				return _factory.ConnectionProvider.Driver.SupportsMultipleOpenReaders
+
+				var result = _factory.ConnectionProvider.Driver.SupportsMultipleOpenReaders
 					? reader
 					: NHybridDataReader.Create(reader);
+				reader = null;
+				return result;
 			}
 			catch (Exception e)
 			{
 				e.Data["actual-sql-query"] = cmd.CommandText;
 				Log.Error(e, "Could not execute query: {0}", cmd.CommandText);
 				throw;
 			}
+			finally
+			{
+				reader?.Dispose();
+			}
```

The local is set to `null` only once something else has taken over the reader, so the `finally` disposes it
on every other way out. It disposes the raw reader directly instead of going through the batcher, because
the batcher never registered it.

`src/NHibernate/Loader/Loader.cs`, `GetResultSet`:

```diff
 			DbDataReader rs = null;
+			var success = false;
 			try
 			{
 				...
+				success = true;
 				return rs;
 			}
 			catch (Exception sqle)
 			{
 				ADOExceptionReporter.LogExceptions(sqle);
-				session.Batcher.CloseCommand(st, rs);
 				throw;
 			}
+			finally
+			{
+				if (!success)
+				{
+					session.Batcher.CloseCommand(st, rs);
+				}
+			}
```

The `success` flag is needed because `rs` is already set well before the method succeeds, for example while
`Advance` or `AutoDiscoverTypes` are still running. Closing it at that point would be just as wrong as
leaking it.

No public API changes.

## How to reproduce

`ReaderLeakOnCancellationFixture` in patch 0002 reproduces both windows against a real SQLite database file.
A self contained query does not reproduce it, because SQLite takes no file lock for a query that never
touches a real table, so the fixture joins a table with one row in it.

The cancellation is not timed, it is deterministic. The fixture cancels from inside System.Data.SQLite's own
`SQLiteConnection.Changed` / `NewDataReader` event, which is raised in the `SQLiteDataReader` constructor
before its `NextResult()`. So the token is guaranteed to be cancelled while `ExecuteReader()` is still
running.

`AppliesTo` requires `factory.Dialect is SQLiteDialect && !Driver.SupportsMultipleOpenReaders`. The dialect
check is needed because the second condition is also false for SqlClient, MySql and Npgsql, while the SQL in
the fixture is SQLite specific.

Result: before the fix both tests fail, after the fix both pass in about a second, on `net48` and `net8.0`.
The probe commands use a 2 second timeout so that a future regression fails in seconds instead of sitting in
SQLite's retry loop.

## What we checked

* The async files were generated, not hand written. We ran `dotnet async-generator` (from
  `.config/dotnet-tools.json`, config `src/AsyncGenerator.yml`) against the patched synchronous files and it
  produced no further changes, so the `Async/` files in the patches are exactly what the tool produces.
* No regressions in the full suite. Baseline at 5.6.0 was 14835 tests, 12976 passed, 4 failed, 16 warnings,
  85 inconclusive, 1754 skipped. After the fix: 14839 tests (the 4 new ones), 12978 passed (+2), the same 4
  failures compared by full test name (`FooBarTest(Async).PersistCollections(Async)` and
  `GH1180.ByCodeFixture(Async).DecimalTypes(Async)`, both already failing before and unrelated), 16
  warnings, 85 inconclusive, 1756 skipped (+2 `[Explicit]`).

## Attached patches

Based on `2b0e1f1d6` (tag 5.6.0), apply with `git am`. We checked that they apply cleanly to a fresh
checkout of that tag.

1. `0001-Fix-DbDataReader-leak-...`: the fix. 4 files, 80 insertions, 6 deletions.
[0001-Fix-DbDataReader-leak-when-a-query-is-cancelled-or-f.patch](https://github.com/user-attachments/files/31521080/0001-Fix-DbDataReader-leak-when-a-query-is-cancelled-or-f.patch)
2. `0002-Add-regression-test-...`: `ReaderLeakOnCancellationFixture`.
[0002-Add-regression-test-for-the-reader-leak-on-cancellat.patch](https://github.com/user-attachments/files/31521088/0002-Add-regression-test-for-the-reader-leak-on-cancellat.patch)
3. `0003-Demonstrate-a-pre-fix-production-workaround-...`: not meant to be merged. It is an `[Explicit]`   fixture with a batcher that users can apply themselves until a fixed version is released, plus the   measured cost of doing so (about 1%, which is inside the noise between runs, because   System.Data.SQLite's `ExecuteReaderAsync` never sees the token mid query anyway). Included only as
   context for anyone hitting this on a released version.
[0003-Demonstrate-a-pre-fix-production-workaround-and-meas.patch](https://github.com/user-attachments/files/31521096/0003-Demonstrate-a-pre-fix-production-workaround-and-meas.patch)

## Related, not fixed

`Loader.PrepareQueryCommandAsync` has the same `catch (OperationCanceledException) { throw; }` pattern, but
it runs while the command is being prepared, before a reader exists. So it would only leak a `DbCommand`. We
have no symptom for it and did not change it, just mentioning it.

Fube:
Fixes #3835, #3661


Fube:
@fredericDelaporte Tests inspired from https://github.com/nhibernate/nhibernate-core/issues/3661#issuecomment-2984661640 to avoid rerunning the tests many times in a row, but I am unsure I like how they turned out.

src/NHibernate.Test/Ado/CancellationTokenFixture.cs
```diff
@@ -0,0 +1,209 @@
...
+	[SetUp]
+	public void Setup()
+	{
+		TestDriver.CancellationTokenSource = null;
+		TestDriver.CommandWasDisposed = false;
+		TestDriver.Reader = null;
+		TestDriver.SupportsMultipleOpenReaders = false;
+	}
+
+
```
fredericDelaporte: Nit: no double blank lines.


```suggestion

```

src/NHibernate.Test/Ado/CancellationTokenFixture.cs
```diff
@@ -0,0 +1,209 @@
...
+[TestFixture]
+public class CancellationTokenFixture : TestCase
+{
+	protected override string MappingsAssembly => "NHibernate.Test";
+
+	protected override string[] Mappings => ["Ado.VerySimple.hbm.xml"];
+
+	protected override void Configure(Configuration configuration)
+	{
+		base.Configure(configuration);
```
fredericDelaporte: The base `Configure` is no-op, our usual practice is to not call it. (Only one test does call it currently.)


```suggestion
```

src/NHibernate.Test/Ado/CancellationTokenFixture.cs
```diff
@@ -0,0 +1,209 @@
...
+		configuration.SetProperty("connection.driver_class", typeof(TestDriver).AssemblyQualifiedName);
```
fredericDelaporte, line 22: These tests assume SQL Server, as they replace the driver by one derived from `MicrosoftDataSqlClientDriver`.

An `Assert.Ignore` should be emitted if the currently tested database is not compatible.

And to avoid interfering with the old SqlClientDriver tests, we should check here the test is configured to use the `MicrosoftDataSqlClientDriver`, if yes adjust the configuration, if no, `Assert.Ignore`.

src/NHibernate/AdoNet/AbstractBatcher.cs
```diff
@@ -250,24 +250,39 @@ public virtual DbDataReader ExecuteReader(DbCommand cmd)
...
 			}
 			catch (Exception e)
 			{
 				e.Data["actual-sql-query"] = cmd.CommandText;
 				Log.Error(e, "Could not execute query: {0}", cmd.CommandText);
 				throw;
 			}
+			finally
+			{
+				if (!success)
```
fredericDelaporte, line 281: I would comment on this.


```suggestion
				// The cleanup must be done in a finally statement, because the async generator adds another catch statement.
				if (!success)
```

src/NHibernate/Loader/Loader.cs
```diff
@@ -1556,14 +1558,22 @@ protected DbDataReader GetResultSet(
...
 			}
 			catch (Exception sqle)
 			{
 				ADOExceptionReporter.LogExceptions(sqle);
-				session.Batcher.CloseCommand(st, rs);
 				throw;
 			}
+			finally
+			{
+				if (!success)
```
fredericDelaporte, line 1572:
```suggestion
				// The cleanup must be done in a finally statement, because the async generator adds another catch statement.
				if (!success)
```

fredericDelaporte:
I do not see a better way to test, and that looks still better than no test at all.

src/NHibernate.Test/Ado/CancellationTokenFixture.cs
fredericDelaporte: Or maybe just test the dialect is derived from `MsSql2000Dialect` in an `AppliesTo` override. That means these test would still run and use the newer client for tests run on the older. But the trouble with testing directly the configured driver is, we may at some point no more use this one and then these tests would no more run at all.

It is a lot less likely that we would cease testing with a dialect derived from `MsSql2000Dialect`.

```
		protected override bool AppliesTo(Dialect.Dialect dialect) => dialect is MsSql2000Dialect;
```

src/NHibernate.Test/Ado/CancellationTokenFixture.cs
Fube: Fixed

src/NHibernate.Test/Ado/CancellationTokenFixture.cs
Fube: Removed

src/NHibernate.Test/Ado/CancellationTokenFixture.cs
Fube: Added the `AppliesTo` override

src/NHibernate/AdoNet/AbstractBatcher.cs
Fube: Added comment

src/NHibernate/Loader/Loader.cs
Fube: Added comment

Fube:
@fredericDelaporte Do you want me to address all the cases identified in https://github.com/nhibernate/nhibernate-core/issues/3835#issuecomment-5465449752 in this PR or in a separate one?

fredericDelaporte (commented):
LGTM.

Is-it draft because you would intend to fix all the other cases you have listed in #3835?

Fube:
> LGTM.
> 
> Is-it draft because you would intend to fix all the other cases you have listed in #3835?

Yes I left it in draft until we decide the scope of this PR.

src/NHibernate.Test/Ado/CancellationTokenFixture.cs
```diff
@@ -0,0 +1,210 @@
...
+	protected override string MappingsAssembly => "NHibernate.Test";
+
+	protected override string[] Mappings => ["Ado.VerySimple.hbm.xml"];
+	
+	protected override void Configure(Configuration configuration)
+	{
+		configuration.SetProperty("connection.driver_class", typeof(TestDriver).AssemblyQualifiedName);
+	}
+
+	protected override bool AppliesTo(Dialect.Dialect dialect) => dialect is MsSql2000Dialect;
```
fredericDelaporte, line 25: It still fails for the old SqlClientDriver, because these tests do not set `TrustServerCertificate=true` in the connection string. The old driver does not mandate it, the new does.

I am adding it. I see no reason to test without it.

.github/workflows/NetCoreTests.yml
```diff
@@ -31,7 +31,7 @@ jobs:
               docker run -e "ACCEPT_EULA=Y" -e "SA_PASSWORD=P@ssw0rd" -e "MSSQL_PID=Express" -p 1433:1433 -d --name sqlexpress mcr.microsoft.com/mssql/server:2019-latest;
           # The install creates the SQLEXPRESS instance and gives the sysadmin role to the user that runs the tests.
           - DB: SqlServer2008
-            CONNECTION_STRING: 'Server=(local)\SQLEXPRESS;initial catalog=nhibernate;Integrated Security=SSPI;packet size=4096;Connection Timeout=60;'
+            CONNECTION_STRING: 'Server=(local)\SQLEXPRESS;initial catalog=nhibernate;Integrated Security=SSPI;packet size=4096;TrustServerCertificate=true;Connection Timeout=60;'
```
fredericDelaporte, line 34: Required for the new tests, which use a driver wrapper around the Microsoft.Data.SqlClient driver. They run for all SQL Server tests, because I consider that testing the driver for ignoring the test may cause them to be too easily ignored and no more run at all in case of a future driver change.

fredericDelaporte:
If you are willing to handle all the other cases, I think it is fine to do it here.

Fube:
@fredericDelaporte For `DbDataReader.Dispose` and `IBatcher.CloseCommand` the stance was that they are not supposed to throw so handling them in a `finally` without a nested `try-catch` was fine.

But for the other items on the list, do we want to take the same stance?
For example, is it acceptable for `ISession.Close` or `IBatcher.AbortBatch` to throw?

fredericDelaporte:
From your list:
- https://github.com/nhibernate/nhibernate-core/blob/fbf41a6adcb8d538f895b65c6c24803ad57385ff/src/NHibernate/Async/Connection/DriverConnectionProvider.cs#L41-L46
  That `Dispose` may throw, so a nested try/catch would be required.
- https://github.com/nhibernate/nhibernate-core/blob/5b3ed3426aacea050d7d340d3b0bdec12427ab3c/src/NHibernate/Async/Context/ThreadLocalSessionContext.cs#L41-L46
  `orphan` is an `ISession`, and its `Close` can throw. It would required a nested try/catch.
- https://github.com/nhibernate/nhibernate-core/blob/858b0f50cbdeddf49d646be2b5ae78270d9d002f/src/NHibernate/Async/Loader/Loader.cs#L1000-L1011
  Case already discussed, no need for a nested try/catch on this one.
- https://github.com/nhibernate/nhibernate-core/blob/8b74774d1fde5c9cc93ba1704d50b4b393f1345c/src/NHibernate/Async/Persister/Collection/AbstractCollectionPersister.cs#L242-L250
  Unclear we need to something. A cancellation will not abort the batch, meaning previously batched commands may remain in the wait list for execution depending on where the cancellation occurred, and the currently batched command will stay in the batch and will not be closed. But its reference will not get lost in the process and it will still be released later, on batch cleanup during session closing or whatever. Looking at the built-in batchers code, all seem to stay in a consistent state.
  So, it does not look to me we need to do something here.
  It may cause some inconsistent behavior if the session is used later on instead of being closed: sometimes, after a cancellation there, previously batched commands would have been cancelled by the way, other times, they will still be there and executed in case of a batch flush. But that is of no concern to me, as it is documented that a session having thrown should be discarded and not used any more.
- https://github.com/nhibernate/nhibernate-core/blob/8b74774d1fde5c9cc93ba1704d50b4b393f1345c/src/NHibernate/Async/Persister/Collection/AbstractCollectionPersister.cs#L406
  Same as previous case.
- https://github.com/nhibernate/nhibernate-core/blob/8b74774d1fde5c9cc93ba1704d50b4b393f1345c/src/NHibernate/Async/Persister/Collection/AbstractCollectionPersister.cs#L548
  Same as previous case.
- https://github.com/nhibernate/nhibernate-core/blob/3adcfa70bd75933b60b50fd7ae7f884fe07d1739/src/NHibernate/Async/Persister/Collection/BasicCollectionPersister.cs#L109
  Same as previous case.
- https://github.com/nhibernate/nhibernate-core/blob/3adcfa70bd75933b60b50fd7ae7f884fe07d1739/src/NHibernate/Async/Persister/Collection/OneToManyPersister.cs#L91
  Same as previous case.
- https://github.com/nhibernate/nhibernate-core/blob/3adcfa70bd75933b60b50fd7ae7f884fe07d1739/src/NHibernate/Async/Persister/Collection/OneToManyPersister.cs#L159
  Same as previous case.
- https://github.com/nhibernate/nhibernate-core/blob/bbe7fc9d4229cef44dd3be27daa3315222033564/src/NHibernate/Async/Persister/Entity/AbstractEntityPersister.cs#L657
  Same as previous case.
- https://github.com/nhibernate/nhibernate-core/blob/bbe7fc9d4229cef44dd3be27daa3315222033564/src/NHibernate/Async/Persister/Entity/AbstractEntityPersister.cs#L796
  Same as previous case.
- https://github.com/nhibernate/nhibernate-core/blob/bbe7fc9d4229cef44dd3be27daa3315222033564/src/NHibernate/Async/Persister/Entity/AbstractEntityPersister.cs#L926
  Same as previous case.

Side note: my analysis of batchers `CloseCommand` not needing a try catch guard is based on the `AbstractBatcher` implementation, which does not allow overrides which could break that. But a custom batcher not deriving from `AbstractBatcher`, implementing directly `IBatcher`, could break this. We can consider that would be a bug to fix in the custom batcher. (But as `IBatcher` does not document these closing methods should never throw, that remains debatable.)

Fube:
@fredericDelaporte thank you for the detailed analysis. I have should have the PR ready in a couple weeks.

PR #3854, opened 2026-09-06
Fix AutoJoinTransaction lost in serialization

### LINKED ISSUE #3782: A deserialized session does not join the current system transaction
`ConnectionManager.ShouldAutoJoinTransaction` is always `false` after deserialization. A deserialized session does not join the current system transaction (`Transaction.Current`). If you use the session in a `TransactionScope`, the session stays out of that scope.

`ConnectionManager.GetObjectData` writes `ownConnection`, `session`, `connectionReleaseMode`, `interceptor` and `connectionAccess`. It does not write `ShouldAutoJoinTransaction`. The property has only a getter (`src/NHibernate/AdoNet/ConnectionManager.cs:128`), and the constructor sets it. The deserialization constructor does not set it. The property keeps the default value `false`.

`AdoNetWithSystemTransactionFactory.ShouldAutoJoinSystemTransaction` reads the property. If the value is `false`, NHibernate does not enlist the session in `Transaction.Current`. The operations of the session stay out of the system transaction, and the applicable data is not in the scope of that transaction.

This test fails on master (b6b13fe14). The two first asserts are correct. The last assert fails.

```csharp
[Test]
public void AutoJoinTransactionSurvivesSerialization()
{
	using var session = Sfi.OpenSession();
	var before = ((ISessionImplementor) session).ConnectionManager.ShouldAutoJoinTransaction;

	var deserialized = (ISession) SerializationHelper.Deserialize(SerializationHelper.Serialize(session));
	var after = ((ISessionImplementor) deserialized).ConnectionManager.ShouldAutoJoinTransaction;

	Assert.That(Sfi.Settings.AutoJoinTransaction, Is.True, "factory setting");
	Assert.That(before, Is.True, "before round trip");
	Assert.That(after, Is.True, "after round trip");
}
```

fredericDelaporte:
Fixes #3782

That is a 5.0 regression. Targeting the earliest version still supported.

Possible breaking change: serialized sessions with previous versions will not be deserializable.

### DIFF
--- a/src/NHibernate.Test/SystemTransactions/AutoJoinSettingFixture.cs
+++ b/src/NHibernate.Test/SystemTransactions/AutoJoinSettingFixture.cs
@@ -1,5 +1,7 @@
 using System.Transactions;
 using NHibernate.Cfg;
+using NHibernate.Engine;
+using NHibernate.Util;
 using NUnit.Framework;
 
 namespace NHibernate.Test.SystemTransactions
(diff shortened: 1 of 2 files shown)


hazzik:
Target the 5.7 and we can backport to 5.6 via margify

hazzik (approved):

